PR #4318, opened 2020-09-14
50544 openldap migration

Firstyear:
Work in progress of the openldap migration tooling. Not yet ready for review.


Firstyear:
@droideck If you get some time do you mind having a quick look at this? I'm sure there is plenty to comment on, but I don't want this to get "too large" too. 

I'm going to start designing what the CLI for this should look like today to wire it all together now.

dirsrvtests/tests/suites/openldap_2_389/migrate_test.py
```diff
@@ -0,0 +1,141 @@
...
+    inst = topology_st.standalone
+    config_path = os.path.join(DATADIR1, 'slapd.d')
+    config = olConfig(config_path)
+    ldifs = {
+        "dc=example,dc=com": os.path.join(DATADIR1, 'example_com.slapcat.ldif'),
+    }
+
+    # 1.3.6.1.4.1.5322.13.1.1 is namedObject, so check that isn't there
+
+    migration = Migration(config, inst, ldifs
```
droideck: A typo - no coma at the end of the line

dirsrvtests/tests/suites/openldap_2_389/migrate_test.py
```diff
@@ -0,0 +1,141 @@
...
+        skip_oids=['1.3.6.1.4.1.5322.13.1.1'],
```
droideck: I think it should be `skip_schema_oids` (as in the __init__ class definition)

src/lib389/cli/openldap_to_ds
```diff
@@ -0,0 +1,188 @@
...
+        help="Prompt for password for the bind DN"
+    )
+parser.add_argument('-y', '--pwdfile',
+        help="Specifies a file containing the password for the binddn",
+        default=None
+    )
+parser.add_argument('-Z', '--starttls',
+        help="Connect with StartTLS",
+        default=False, action='store_true'
+    )
```
droideck: Probably, worth specifying in the description for which server the credentials are: OpenLDAP or 389-ds-base (I understand that we process an LDIF file for OpenLDAP, mostly, but it's better to be explicit than implicit in such docs, IMHO)

droideck (changes requested):
Sorry for the late review but I'm finally here. :)

Besides the comments below, there are a couple of things.
Probably, you've already thought about that but still, there are things that will be nice to have (if you'll have time):

- Tests with actual OpenLDAP installation (with some pytest.skipif() flag like this - `openldap_installation_ack = pytest.mark.skipif(not os.environ.get('OPENLDAP_INSTALLATION_ACK', False), reason="You should have these packages installed")`) - may be even for a different versions;
- CLI tests;

The rest looks good(!) at the first glance, I'll be looking in more details on the way of the development. :)

Firstyear:
> Sorry for the late review but I'm finally here. :)

Better late than never :) I appreciate the review I will certainly apply these comments

> 
> Besides the comments below, there are a couple of things.
> Probably, you've already thought about that but still, there are things that will be nice to have (if you'll have time):
> 
> * Tests with actual OpenLDAP installation (with some pytest.skipif() flag like this - `openldap_installation_ack = pytest.mark.skipif(not os.environ.get('OPENLDAP_INSTALLATION_ACK', False), reason="You should have these packages installed")`) - may be even for a different versions;

So the way this will work is always to read the *offline* content of an openldap install, aka the slapd.d directory and the ldif export. So the examples in the PR are generated from real openldap servers, so they are "real" installations :) 

Part of the reason to structure it this way is that it guarantees no destructive actions against the existing openldap install (good for production risk minimising) but also means we can MUCH more easily generate test cases! 

> * CLI tests;

Yep for sure. Most of the logic should be behind in the migration tooling though so we should be able to test it pretty effectively without needing to go through the cli, but a basic cli smoke test is a good idea. 

> 
> The rest looks good(!) at the first glance, I'll be looking in more details on the way of the development. :)

Thank you! I'll submit more later, I just got some approval from thierry to continue with a syncrepl issue so I want to finish that first :)

Firstyear:
@droideck Thanks mate, I've made the changes you suggested, but also in the mean time I've effectively completed the majority of the tool. A few command line options aren't wired in yet, but I just stepped on a bug in memberof so I need to fix that first. :)

Firstyear:
@droideck Okay, I think this is probably 95% complete now, and I've tested it pretty thoroughly. As mentioned there are test suites with this. I'll probably also get some testing from people at SUSE too about the functionality.

Firstyear:
@droideck, @mreynolds389 review reminder :)

mreynolds389 (approved):
LGTM

droideck:
I have the issue while building RPM using mock:
```
Error: Installed (but unpackaged) file(s) found:
   /usr/sbin/openldap_to_ds
   /usr/share/man/man8/openldap_to_ds.8.gz
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/sbin/openldap_to_ds
   /usr/share/man/man8/openldap_to_ds.8.gz
```

Firstyear:
Thanks @mreynolds389 and @droideck. I have rebased and updated the spec file, can you confirm it works for you @droideck? 

PS: Yes, I will squash the commits to a single change when I merge :)

dirsrvtests/tests/suites/openldap_2_389/migrate_test.py
```diff
@@ -0,0 +1,141 @@
...
+@pytest.mark.skipif(ds_is_older('1.4.3'), reason="Not implemented")
+def test_parse_openldap_slapdd():
+    """Test parsing an example openldap configuration. We should be able to
+    at least determine the backends, what overlays they have, and some other
+    minimal amount.
+
+    :id: b0061ab0-fff4-45c6-b6c6-171ca3d2dfbc
+    :setup: Data directory with an openldap config directory.
+    :steps:
+        1.
```
droideck: Could you please put some steps here? I think some automatic test-doc systems may complain if the fields are empty.

droideck (approved):
Besides the comment, looks very good! Ack.
Thanks for the big work and let's test it more ;)

Firstyear:
@droideck done, and commit description updated/squashed. And I have already lined up some more testing of the feature to occur soon so more improvements to come I hope!

Firstyear:
@droideck Double checking the ack is still good :)

droideck:
Yes, looks good!

Firstyear:
Thank you @droideck! I will continue to add more tests to this and I have some community members who want to test it too. really appreciate the review!

PR #4323, opened 2020-09-14
Issue 4319 - Performance search rate: listener may be erroneously wak…

### LINKED ISSUE #4319: Performance search rate: listener may be erroneously waken up
When a worker completed an operation, on a given connection, it can wake up the listener on specific events (timeout, replication, maxthread). If this happens it wakes up the listener but the keeps flag set. So this worker will systematically wakeup the listener for others following operations (even on others connections or even if the specific events do not happen).
The result is extra signals (write) and noise on worker.

It can be detect with perf tracing worker syscalls, without the fix the # writes are ~sendto, else it is half

    Without fix:
    
       syscall            calls    total       min       avg       max      stddev
                                   (msec)    (msec)    (msec)    (msec)        (%)
       --------------- -------- --------- --------- --------- ---------     ------
       futex              83704  1021.638     0.000     0.012   100.113     23.99%
       sendto             30268   245.164     0.002     0.008     0.041      0.22%
       write              30301   123.068     0.002     0.004     0.033      0.23%
       recvfrom           15222    65.530     0.002     0.004     0.034      0.28%
       poll                  40     1.146     0.003     0.029     0.098     12.61%
 
 
    With fix
 
       syscall            calls    total       min       avg       max      stddev
                                   (msec)    (msec)    (msec)    (msec)        (%)
       --------------- -------- --------- --------- --------- ---------     ------
       futex             137730  1299.397     0.001     0.009   100.147     18.88%
       sendto             44004   338.978     0.003     0.008     0.054      0.15%
       recvfrom           22145   104.284     0.002     0.005     0.031      0.15%
       poll                  52   101.674     0.004     1.955   100.139     98.46%
       write              22093    83.288     0.002     0.004     0.029      0.21%

It improves throughput ~2%

tbordaz:
…en up

Bug description:
	A worker thread usually wakes up the listener when it has completed
        reading the operation from the operation.
	In addition upon exceptional event (timeout while reading op or
	max thread per connection), it sets a local flag (need_wakeup) and
	wakes the listener.
	The problem is that it does not reset the flag after wake up. So
	for any further operation (on any operation) it will trigger this
	additional wake up.
	This triggers a write syscall and wakes up listener for nothing.
	This impacts througput by ~2%

Fix description:
	reset the need_wakeup after signal_listner

relates: https://github.com/389ds/389-ds-base/issues/4319

Reviewed by: ?

Platforms tested: F31, RHEL8.3

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -1467,7 +1467,6 @@ connection_threadmain()
     Connection *conn = NULL;
     Operation *op;
     ber_tag_t tag = 0;
-    int need_wakeup = 0;
     int thread_turbo_flag = 0;
     int ret = 0;
     int more_data = 0;
@@ -1817,6 +1816,8 @@ connection_threadmain()
             }
             if (!more_data) {
                 if (!thread_turbo_flag) {
+                    int32_t need_wakeup = 0;
+
                     /*
                      * Don't release the connection now.
                      * But note down what to do.
@@ -1840,10 +1841,13 @@ connection_threadmain()
                     }
                     conn->c_threadnumber--;
                     connection_release_nolock(conn);
-                    /* Call signal_listner after releasing the
-                     * connection if required. */
+                    /* If need_wakeup, call signal_listner once.
+                     * Need to release the connection (refcnt--)
+                     * before that call.
+                     */
                     if (need_wakeup) {
                         signal_listner();
+                        need_wakeup = 0;
                     }
                 } else if (1 == is_timedout) {
                     /* covscan reports this code is unreachable  (2019/6/4) */


ldap/servers/slapd/connection.c
```diff
@@ -1482,9 +1482,10 @@ connection_threadmain()
 #endif
 
     while (1) {
-        int is_timedout = 0;
+        int is_timedout;
         time_t curtime = 0;
 
+        is_timedout = 0;
```
Firstyear: Why move this assignment?

Firstyear:
As an alternate idea, why not move need_wakeup from line 1470 to line 1819? I think your patch is correct, that need_wakeup as a variable isn't reseting each loop, so past settings to need_wakeup are leaking between listener operations. So if that's the case, shouldn't we limit the scope of the variable instead of just reseting it? It would mean we don't need the need_wakeup = 0 on line 1832 too. 

Thanks!

ldap/servers/slapd/connection.c
tbordaz: It is a leftover of a previous tentative fix. I was initially thinking that the extra signal_listener were triggered by is_timeout not being reset at each loop. I had a doubt how compiler processed this initial setting and to be sure of that I moved this line

tbordaz:
@Firstyear thanks for your review. I agree with the move of the definition of need_wakeup to the block in 1819. I will do it. 
I prefer to keep resetting the flag once signal_listner has been triggered. It makes it clear that purpose of need_wake is limited to this.

ldap/servers/slapd/connection.c
Firstyear: Okay, if you undo this bit (ie change back to `int is_timedout = 0;` on a single, line then you have my ack :) Thanks!

mreynolds389:
LGTM

Firstyear:
> LGTM

There is still an outstanding comment in my review section :)

tbordaz:
Thanks @mreynolds389 , @Firstyear for this review. I have updated the patch according to your comments.

PR #4325, opened 2020-09-15
Issue 4241 - Add CI tests

sgouvern:
Automated tests to verify that
- db2ldif exits properly when the ldif file path provided cannot be accessed
- a usefull error message is displayed as output when the ldif file cannot be accessed

Relates: https://github.com/389ds/389-ds-base/issues/4241
Relates: https://github.com/389ds/389-ds-base/issues/4278

Reviewed by: bsimonova, Firstyear. Thanks !

### DIFF
--- a/dirsrvtests/tests/suites/export/export_test.py
+++ b/dirsrvtests/tests/suites/export/export_test.py
@@ -1,0 +1,155 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+
+import os
+import pytest
+import subprocess
+
+from lib389.topologies import topology_st as topo
+from lib389._constants import DEFAULT_SUFFIX, DEFAULT_BENAME
+from lib389.utils import *
+from lib389.paths import Paths
+from lib389.cli_base import FakeArgs
+from lib389.cli_ctl.dbtasks import dbtasks_db2ldif
+
+pytestmark = pytest.mark.tier1
+
+def run_db2ldif_and_clear_logs(topology, instance, backend, ldif, output_msg, encrypt=False, repl=False):
+    args = FakeArgs()
+    args.instance = instance.serverid
+    args.backend = backend
+    args.encrypted = encrypt
+    args.replication = repl
+    args.ldif = ldif
+     
+    dbtasks_db2ldif(instance, topology.logcap.log, args)
+
+    log.info('checking output msg')
+    if not topology.logcap.contains(output_msg):
+        log.error('The output message is not the expected one')
+        assert False
+
+    log.info('Clear the log')
+    topology.logcap.flush()
+
+
+@pytest.mark.bz1806978
+@pytest.mark.ds51188
+@pytest.mark.skipif(ds_is_older("1.3.10", "1.4.2"), reason="Not implemented")
+def test_dbtasks_db2ldif_with_non_accessible_ldif_file_path(topo):
+    """Export with dsctl db2ldif, giving a ldif file path which can't be accessed by the user (dirsrv by default)
+
+    :id: ca91eda7-27b1-4750-a013-531a63d3f5b0
+    :setup: Standalone Instance - entries imported in the db
+    :steps:
+        1. Stop the server
+        2. Launch db2ldif with an non accessible ldif file path 
+        3. Catch the reported error code
+        4. check the error reported in the errors log
+    :expected results:
+        1. Operation successful
+        2. Operation properly fails, without crashing
+        3. An error code different from 139 (segmentation fault) should be reported
+        4. 'ERR - bdb_db2ldif - db2ldif: userRoot: can't open file' should be reported
+    """
+    export_ldif = '/tmp/nonexistent/export.ldif'
+
+    log.info("Stopping the instance...")
+    topo.standalone.stop()
+
+    log.info("Performing an offline export to a non accessible ldif file path - should fail properly")
+    expected_output="db2ldif failed"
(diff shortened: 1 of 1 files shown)


sgouvern:
Coming back to this PR after a long time, sorry it will not ease the review :(
I included comments from @bsimonova and @Firstyear (see #4278) 
I split the test into two tests, as it concerns two different bz.
I also used a direct cli call, because I couldn't reproduced the crash using lib389 db2ldif dbtask.
Thanks to review.

bsimonova:
@sgouvern 
The docstring is now in correct format, thanks!

dirsrvtests/tests/suites/export/export_test.py
```diff
@@ -0,0 +1,101 @@
...
+        4. 'ERR - bdb_db2ldif - db2ldif: userRoot: can't open file' should be reported
+    """
+    export_ldif = '/tmp/nonexistent/export.ldif'
+
+    log.info("Stopping the instance...")
+    topo.standalone.stop()
+
+    log.info("Performing an offline export to a non accessible ldif file path - should fail properly")
+    try:
+        subprocess.check_call(['db2ldif', '-Z', topo.standalone.serverid, '-s', DEFAULT_SUFFIX, '-a', export_ldif])
```
Firstyear: I think you should use the topo db2ldif call instead. The reason is that if you are a developer (like myself) then I install into a prefix (/opt/dirsrv) and db2ldif won't be on my shell path, so subprocess here would fail. topo.db2ldif is aware of prefixs, and knows about the server id etc, so you should call `topo.standalone.db2ldif( ... ) ` instead.

Firstyear: Or if that's not possible, at least use `os.path.join(self.ds_paths.sbin_dir, 'db2ldif')` to account for prefixs.

sgouvern:
Thanks @bsimonova and @Firstyear for your review. 
@Firstyear I've updated the test according to your comments, thanks to review !

dirsrvtests/tests/suites/export/export_test.py
```diff
@@ -0,0 +1,103 @@
...
+        3. Catch the reported error code
+        4. check the error reported in the errors log
+    :expected results:
+        1. Operation successful
+        2. Operation properly fails, without crashing
+        3. An error code different from 139 (segmentation fault) should be reported
+        4. 'ERR - bdb_db2ldif - db2ldif: userRoot: can't open file' should be reported
+    """
+    export_ldif = '/tmp/nonexistent/export.ldif'
+    db2ldif_cmd = os.path.join(topo.standalone.ds_paths.sbin_dir, 'db2ldif')
```
Firstyear, line 102: So it turns out my comment before was mostly correct. This won't work on 1.4.x because db2ldif no longer exists so you need to call the  topo.standalone.db2ldif() api call which will handle this correctly. Sorry about this mistake :( :(

sgouvern: The point is that one of the tests was to verify a fix for a crash produced by db2ldif in 1.4.2. Even if not supported, it is still delivered in 389-ds-base-legacy-tools package, and a fix has been done in 1.4.3.
So In the 1st test, I still use the direct cli call to db2ldif for verification of the fix in the 1.4.3 version, and use the lib389 dbtasks_db2ldif for test run in future versions, and for tests with other scope.
Does it makes sense ?

Firstyear: Yes it does, but db2ldif is a shell script that wraps ns-slapd db2ldif. This is also what dsctl instance db2ldif does, and what they python call does. so if you have a crash in db2ldif, then you should be able to produce it with the call to the python API which is able to test on 1.3.x and 1.4.x. I think it's more important to have a test that will work for 1.4.x and future versions, than just 1.3.x and lower.

sgouvern: I also thought in the 1st place that using python API call would produce the crash, which for some reason is not the case, maybe it could be investigated. 
I agree that its important to have tests that will work on future versions, and they have to work too on lower versions, for regression detection purpose. And I think too it's worth to have a test verifying that ns-slapd db2ldif won't crash anymore in future versions, as long as we deliver 389-ds-base-legacy-tools package. 
So I separated the ns-slapd db2ldif test so that it can be run only on relevant versions, and other tests can be run on all versions, older and future.
I hope that you'll find this satisfying enough :)

Firstyear: I think the rest of the team will attest to how difficult I am to satisfy :P 

It's curious that the python call does *not* cause the crash to occur, because underneath db2ldif the shell script and the python call both lower to a call to ns-slapd db2ldif <options>. So perhaps the options the python version emits causes a difference, but they both "should" cause the crash because it's the same call in the end. 

With the ds_is_older flag yes, this limits the scope, so I think this is probably okay in that case.

sgouvern: @Firstyear , thanks :)
@bsimonova, could you please merge it ? Thanks !

sgouvern: @Firstyear, @vashirov, it seems that Barbora has not the rights to merge. 
When you have time, can you merge it ? Thanks !

sgouvern: Sorry, please forget my previous request,  I didn't see that @Firstyear already merge it.
Thanks, @Firstyear !!

Firstyear: All good, @sgouvern and @bsimonova if you need anything merged just ping/email me as I think I'm pretty responsive/active :)

bsimonova: @Firstyear I had recently acquired the rights for pagure before migration, but not for github yet. Thanks!

PR #4331, opened 2020-09-18
Issue 4176 - import ldif2cl task should not close all changelogs

### LINKED ISSUE #4176: move CL followup (3) - import ldif2cl task should not close all changelogs
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/51123
- Created at 2020-05-28 14:08:38 by [lkrispen](https://pagure.io/user/lkrispen) (@elkris)
- Assigned to [mreynolds](https://pagure.io/user/mreynolds) (@mreynolds389)
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=1843524

---

replica_execute_ldif2cl_task() wraps the cl import between 
cl5Close and cl5Open calls, but these close and reopen the changelog for ALL replicas, it should be sufficient to Set/Unset cldb for the changelog imported.

NOTE: how are replication agreements affected/ notified ?
I have the impression that this code is not and was not safe.

mreynolds389:
Bug Description:  With the new per-backend replication changelog, the
                  ldif2cl task would incorrectly close all the backends.

Fix Description:  First, the global changelog struct (s_cl5Desc) was
                  completely removed and merged with the replica changelog
                  db handle struct.  The dbState variable is used to
                  synchronize access to the changelog db struct during
                  shutdown, or ldif2cl tasks.

                  The CLI was updated to handle setting changelog encryption,
                  and importing/restoring a changelog ldif.

                  The UI was updated to handle the new per-backlend changelog
                  and its configuration.  Also added the option to
                  export/import the changelog and its various forms.

Fixes: https://github.com/389ds/389-ds-base/issues/4176

### DIFF
--- a/dirsrvtests/tests/suites/replication/changelog_encryption_test.py
+++ b/dirsrvtests/tests/suites/replication/changelog_encryption_test.py
@@ -1,0 +1,81 @@
+import logging
+import pytest
+import os
+import time
+from lib389._constants import DEFAULT_SUFFIX,  DN_CHANGELOG,  DN_USERROOT_LDBM
+from lib389.topologies import topology_m1c1 as topo
+from lib389.dseldif import DSEldif
+from lib389.utils import ds_supports_new_changelog
+from lib389.replica import Replicas
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+
+def test_cl_encryption_setup_process(topo):
+    """Take an already working replication deployment, and setup changelog
+    encryption
+
+    :id: 1a1b7d29-69f5-4f0e-91c4-e7f66140ff17
+    :setup: Master Instance, Consumer Instance
+    :steps:
+        1. Enable TLS for the server
+        2. Export changelog
+        3. Enable changelog encryption
+        4. Import changelog
+        5. Verify replication is still working
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+    """
+
+    supplier = topo.ms['master1']
+    consumer = topo.cs['consumer1']
+
+    # Enable TLS
+    log.info('Enable TLS ...')
+    supplier.enable_tls()
+    consumer.enable_tls()
+
+    # Export changelog
+    log.info('Export changelog ...')
+    replicas = Replicas(supplier)
+    replica = replicas.get(DEFAULT_SUFFIX)
+    replica.begin_task_cl2ldif()
+    replica.task_finished()
+
+    # Enable changelog encryption
+    log.info('Enable changelog encryption ...')
+    dse_ldif = DSEldif(supplier)
+    supplier.stop()
+    if ds_supports_new_changelog():
+        changelog = 'cn=changelog,{}'.format(DN_USERROOT_LDBM)
+    else:
+        changelog = DN_CHANGELOG
+    dse_ldif.replace(changelog, 'nsslapd-encryptionalgorithm', 'AES')
+    if dse_ldif.get(changelog, 'nsSymmetricKey'):
+        dse_ldif.delete(changelog, 'nsSymmetricKey')
+    supplier.start()
+
+    # Import changelog
+    log.info('Import changelog ...')
+    replica.begin_task_ldif2cl()
+    replica.task_finished()
+
+    # Verify replication is still working
+    log.info('Test replication is still working ...')
+    assert replica.test_replication([consumer])
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/dirsrvtests/tests/suites/replication/multiple_changelogs_test.py
+++ b/dirsrvtests/tests/suites/replication/multiple_changelogs_test.py
@@ -1,0 +1,175 @@
+import ldap
+import logging
+import pytest
+import os
+import threading
+import time
+from lib389._constants import *
+from lib389.topologies import topology_m1c1 as topo
+from lib389.idm.directorymanager import DirectoryManager
+from lib389.idm.domain import Domain
+from lib389.backend import Backend
+from lib389.replica import Replicas, ReplicationManager
+
+log = logging.getLogger(__name__)
+
+SECOND_SUFFIX = 'dc=second_suffix'
+MOD_COUNT = 50
+
+class DoMods(threading.Thread):
+    """modify the suffix entry"""
+    def __init__(self, inst, task):
+        """
+        Initialize the thread
+        """
+        threading.Thread.__init__(self)
+        self.daemon = True
+        self.inst = inst
+        self.name = inst.serverid
+        self.task = task
+
+    def run(self):
+        """
+        Start adding users
+        """
+        idx = 0
+        conn = DirectoryManager(self.inst).bind()
+        domain = Domain(conn, DEFAULT_SUFFIX)
+        while idx < MOD_COUNT:
+            try:
+                domain.replace('description', str(idx))
+            except:
+                if self.task == "import":
+                    # Failures are expected during an import
+                    pass
+                else:
+                    # export, should not fail
+                    log.fatal('Updates should not fail during an export')
+                    assert False
+            idx += 1
+
+
+def test_multiple_changelogs(topo):
+    """Test the multiple suffixes can be replicated with the new per backend
+    changelog.
+
+    :id: eafcdb57-4ea2-4887-a0a8-9e4d295f4f4d
+    :setup: Master Instance, Consumer Instance
+    :steps:
+        1. Create s second suffix
+        2. Enable replication for second backend
+        3. Perform some updates on both backends and make sure replication is
+           working for both backends
+
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+    """
+    supplier = topo.ms['master1']
+    consumer = topo.cs['consumer1']
+
+    # Create second suffix dc=second_backend on both replicas
+    for inst in [supplier, consumer]:
+        # Create the backends
+        props = {'cn': 'secondRoot', 'nsslapd-suffix': SECOND_SUFFIX}
+        be = Backend(inst)
+        be.create(properties=props)
+        be.create_sample_entries('001004002')
+
+    # Setup replication for second suffix
+    repl = ReplicationManager(SECOND_SUFFIX)
+    repl.create_first_master(supplier)
+    repl.join_consumer(supplier, consumer)
+
+    # Test replication works for each backend
+    for suffix in [DEFAULT_SUFFIX, SECOND_SUFFIX]:
+        replicas = Replicas(supplier)
+        replica = replicas.get(suffix)
+        log.info("Testing replication for: " + suffix)
+        assert replica.test_replication([consumer])
+
+
+def test_multiple_changelogs_export_import(topo):
+    """Test that we can export and import the replication changelog
+
+    :id: b74fcaaf-a13f-4ee0-98f9-248b281f8700
+    :setup: Master Instance, Consumer Instance
+    :steps:
+        1. Create s second suffix
+        2. Enable replication for second backend
+        3. Perform some updates on a backend, and export the changelog
+        4. Do an export and import while the server is idle
+        5. Do an import while the server is under load
+
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+    """
+    SECOND_SUFFIX = 'dc=second_suffix'
+    supplier = topo.ms['master1']
+    consumer = topo.cs['consumer1']
+    supplier.config.set('nsslapd-errorlog-level', '0')
+    # Create second suffix dc=second_backend on both replicas
+    for inst in [supplier, consumer]:
+        # Create the backends
+        props = {'cn': 'secondRoot', 'nsslapd-suffix': SECOND_SUFFIX}
+        be = Backend(inst)
+        try:
+            be.create(properties=props)
+            be.create_sample_entries('001004002')
+        except ldap.UNWILLING_TO_PERFORM:
+            pass
+
+    # Setup replication for second suffix
+    try:
+        repl = ReplicationManager(SECOND_SUFFIX)
+        repl.create_first_master(supplier)
+        repl.join_consumer(supplier, consumer)
+    except ldap.ALREADY_EXISTS:
+        pass
+
+    # Put the replica under load, and export the changelog
+    replicas = Replicas(supplier)
+    replica = replicas.get(DEFAULT_SUFFIX)
+    doMods1 = DoMods(supplier, task="export")
+    doMods1.start()
+    replica.begin_task_cl2ldif()
+    doMods1.join()
+    replica.task_finished()
+
+    # allow some time to pass, and test replication
+    time.sleep(1)
+    assert replica.test_replication([consumer])
+
+    # While idle, go an export and import, and make sure replication still works
+    log.info("Testing idle server with CL export and import...")
+    replica.begin_task_cl2ldif()
+    replica.task_finished()
+    replica.begin_task_ldif2cl()
+    replica.task_finished()
+    assert replica.test_replication([consumer])
+
+    # stability test, put the replica under load, import the changelog, and make
+    # sure server did not crash.
+    log.info("Testing busy server with CL import...")
+    doMods2 = DoMods(supplier, task="import")
+    doMods2.start()
+    replica.begin_task_ldif2cl()
+    doMods2.join()
+    replica.task_finished()
+    # Replication will be broken so no need to test it.  This is just make sure
+    # the import works, and the server is stable
+    assert supplier.status()
+    assert consumer.status()
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -31,7 +31,7 @@
 #include "cl_crypt.h"
 #include "plhash.h"
 #include "plstr.h"
-
+#include <pthread.h>
 #include "db.h"
 #include "cl5_clcache.h" /* To use the Changelog Cache */
 #include "repl5.h"       /* for agmt_get_consumer_rid() */
@@ -39,6 +39,7 @@
 #define GUARDIAN_FILE "guardian" /* name of the guardian file */
 #define VERSION_FILE "DBVERSION" /* name of the version file  */
 #define V_5 5                    /* changelog entry version */
+#define V_6 6                    /* changelog entry version that includes encrypted flag */
 #define CHUNK_SIZE 64 * 1024
 #define DBID_SIZE 64
 #define FILE_SEP "_" /* separates parts of the db file name */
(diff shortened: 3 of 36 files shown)


ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -111,17 +111,19 @@ typedef unsigned short u_int16_t;
...
 #define BDB_RDNFORMAT         "rdn-format"   /* Subtree rename enabled */
 #define BDB_RDNFORMAT_VERSION "3"            /* rdn-format version (by default, 0) */
 #define BDB_DNFORMAT          "dn-4514"      /* DN format RFC 4514 compliant */
 #define BDB_DNFORMAT_VERSION  "1"            /* DN format version */
-#define BDB_CL_FILENAME       "changelog.db"
+#define BDB_CL_FILENAME       "replication_changelog.db"
 
 #define LMDB_IMPL             "lmdb"
-#define LMDB_CL_FILENAME      "changelog.mdb"
+#define LMDB_CL_FILENAME      "replication_changelog.mdb"
```
elkris: several comments on this:
- why change the name "changelog", without automatically update filenames it could break existing deployments
- the bdb/lmdb specific definitions shoulod't be at lmdb level, only known and needed in the impl layer
- the filename for lmdb doesn't make sense, in lmdb there exist only one filename for the backend (or even all backends) and "changelog" is just one database inside this opaque file

mreynolds389: Well I used "replication_changelog" because I was worried that "changelog" could easily be a custom attribute a customer might use.  Making the name more unique seemed like a safer approach.  Right now there are no releases that have your original changelog work in them, so no one is using changelog.db at the moment.  Figured now was the best time to change it before we get some huge customer that is already using "changelog" as an index and then replication or index breaks.

As for LMDB, I don't know how the database looks and is laid out.  I was tryingto be proactive in this patch :-)   I didn't know the entire database was one file.  There are very little docs out there about such things.  So if it's not needed I can remove that then

Firstyear: I agree with the naming, making it more specific is a good thing here.

Firstyear:
I don't see anything obviously wrong, but I am not an expert in this area, so it would be good for @tbordaz to look too. I am assuming this does pass with ASAN + The full replication test suite?

mreynolds389:
> I don't see anything obviously wrong, but I am not an expert in this area, so it would be good for @tbordaz to look too. I am assuming this does pass with ASAN + The full replication test suite?

Yes it passed ASAN and the replication test suite :-)

dirsrvtests/tests/suites/replication/changelog_encryption_test.py
```diff
@@ -0,0 +1,81 @@
...
+    # Enable TLS
+    log.info('Enable TLS ...')
+    supplier.enable_tls()
+    consumer.enable_tls()
+
+    # Export changelog
+    log.info('Export changelog ...')
+    replicas = Replicas(supplier)
+    replica = replicas.get(DEFAULT_SUFFIX)
+    replica.begin_task_cl2ldif()
```
Firstyear, line 51: Is there a task.wait()  you could use here?

dirsrvtests/tests/suites/replication/multiple_changelogs_test.py
```diff
@@ -0,0 +1,178 @@
...
+    replica.begin_task_cl2ldif()
+    doMods1.join()
+
+    # allow some time to pass, and test replication
+    time.sleep(1)
+    assert replica.test_replication([consumer])
+
+    # While idle, go an export and import, and make sure replication still works
+    log.info("Testing idle server with CL export and import...")
+    replica.begin_task_cl2ldif()
```
Firstyear, line 150: Similar here, is there a task.wait() ?

src/lib389/lib389/replica.py
```diff
@@ -1084,7 +1088,22 @@ def set_max_age(self, value):
...
         """
+        self.replace('nsslapd-changelogmaxage', value)
 
+    def set_encrypt(self, value):
+        """Set the changelog encryption
+
+        :param value:
+        :type value: str
+        """
+        self.replace('nsslapd-changelogmaxage', value)
```
Firstyear: This looks wrong? Should this be maxage here?

src/lib389/lib389/replica.py
```diff
@@ -1666,6 +1685,22 @@ def begin_task_ldif2cl(self):
         """
         self.replace('nsds5task', 'ldif2cl')
 
+    def task_finished(self):
```
Firstyear, line 1712: Is this ever called outside of a testing context?

Firstyear (changes requested):

src/lib389/lib389/replica.py
mreynolds389: Oh, yeah, copy and paste error. Nice catch!

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -773,28 +667,24 @@ cl5ImportLDIF(const char *clDir, const char *ldifFile, Replica *replica)
...
-{
-    return s_cl5Desc.dbState;
+    /* All done, reset the clcache, state, and mode */
+    pthread_mutex_lock(&(cldb->stLock));
+
+    clcache_destroy();
+    clcache_init();
+    clcache_set_config();
+    cldb->dbState = CL5_STATE_OPEN;
+    cldb->dbOpenMode = CL5_OPEN_NORMAL;
```
tbordaz, line 682: Shouldn't we restore RUV and entrycount before stating it is open in normal mode ?
Like it is done in cldb_SetReplicaDB ?

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -104,17 +104,15 @@ typedef enum {
     CL5_OPEN_RESTORE_RECOVER, /* restore from archive and recover */
```
tbordaz: It looks to me that CL5_OPEN_RESTORE* and CL5_OPEN_CLEAN* are no longer in use.
Would you remove them ?

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -634,28 +514,42 @@ cl5ImportLDIF(const char *clDir, const char *ldifFile, Replica *replica)
...
+    ruv_obj = replica_get_ruv(replica);
+
+    pthread_mutex_lock(&(cldb->stLock));
+    slapi_back_get_info(be, BACK_INFO_DBENV_CLDB, (void **)&pDB);
+    cldb->db = pDB;
+    cldb->dbOpenMode = CL5_OPEN_LDIF2CL;
+    slapi_ch_free_string(&cldb->ident);
+    cldb->ident = ruv_get_replica_generation((RUV*)object_get_data (ruv_obj));
+    if (_cldb_CheckAndSetEnv(be, cldb) != CL5_SUCCESS) {
+        object_release(ruv_obj);
```
tbordaz, line 541: Difficult to know what to do at this step. 
the dbState is IMPORT and openMode is LDIF2CL.
Would move dbState to OPEN and OpenMode to NORMAL.

ldap/servers/plugins/replication/cl5_config.c
```diff
@@ -314,6 +314,12 @@ cldb_config_modify(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *entryAfter, in
                     /* Storing the encryption symmetric key */
                     /* no need to change any changelog configuration */
                     goto done;
+
+
+                } else if (strcasecmp(config_attr, CONFIG_CHANGELOG_ENCRYPTION_ALGORITHM) == 0) {
+                    /* We should allow the operation to succeed but it requires
+                     * a restart to take effect. */
```
tbordaz, line 319: Is it missing ?
slapi_ch_free_string(&config.encryptionAlgorithm);
config.encryptionAlgorithm = slapi_ch_strdup(config_attr_value);

ldap/servers/plugins/replication/cl5_config.c
```diff
@@ -314,6 +314,12 @@ cldb_config_modify(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *entryAfter, in
                     /* Storing the encryption symmetric key */
                     /* no need to change any changelog configuration */
                     goto done;
```
tbordaz, line 316: I do not think this 'goto done' is valid, it should continue reading the remaining mods and then doing the tests after the loop.

ldap/servers/plugins/replication/cl5_config.c
```diff
@@ -314,6 +314,12 @@ cldb_config_modify(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *entryAfter, in
...
+
+
+                } else if (strcasecmp(config_attr, CONFIG_CHANGELOG_ENCRYPTION_ALGORITHM) == 0) {
+                    /* We should allow the operation to succeed but it requires
+                     * a restart to take effect. */
+                    goto done;
```
tbordaz, line 320: Same thing, I do not understand the 'goto done'

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -472,13 +362,12 @@ cl5GetUpperBoundRUV(Replica *r, RUV **ruv)
     }
 
     /* changelog is not initialized */
-    if (s_cl5Desc.dbState == CL5_STATE_NONE) {
-        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl, "cl5GetUpperBoundRUV - "
-                                                           "Changelog is not initialized\n");
+    if (cldb->dbState == CL5_STATE_CLOSED) {
```
tbordaz, line 364: I think it is more (cldb->dbState !- CL5_STATE_OPEN) -> BAD STATE
If it is closed or imported we are in transient state where maxruv is not valid

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -128,18 +126,23 @@ struct cl5DBFileHandle
...
-    int flags;	/* currently used to mark the file or as initialized */
-    RUV *purgeRUV;  /* ruv to which the file has been purged */
-    RUV *maxRUV;         /* ruv that marks the upper boundary of the data */
-    CL5Config clConf;      /* trimming and encryption config                                */
+    DB *db;                 /* db handle to the changelog file */
+    DB_ENV *dbEnv;          /* db environment shared by all db files */
+    char *ident;            /* identifier for changelog, used in error messages */
+    int entryCount;         /* number of entries in the file  */
+    CL5State dbState;       /* changelog current state */
+    pthread_mutex_t stLock; /* lock that controls access to the changelog flags */
```
tbordaz: It would be good to say in the comment which fields are protected by the lock.
I think dbState should be protected but may be at least 'db' 'dbEnv'.

ldap/servers/plugins/replication/cl5_api.c
tbordaz, line 364: In my opinion access to dbState requires 'stLock'. It is the same in cl5ImportLDIF, cl5ConfigTrimming, cl5GetOperationCount

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -914,16 +804,26 @@ cl5WriteOperationTxn(cldb_Handle *cldb, const slapi_operation_parameters *op, vo
...
+    if (cldb->dbState != CL5_STATE_OPEN) {
+        if (cldb->dbState == CL5_STATE_IMPORT) {
+            /* this is expected, no need to flood the logs with the same msg */
+            slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl,
+                    "cl5WriteOperationTxn - Changelog is currently being initialized and can not be updated\n");
+        } else {
+            slapi_log_err(SLAPI_LOG_INFO, repl_plugin_name_cl,
+                    "cl5WriteOperationTxn - Changelog is not initialized\n");
+        }
+        slapi_log_err(SLAPI_LOG_INFO, repl_plugin_name_cl,
```
tbordaz: This log is useless

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -4571,31 +4421,28 @@ void
...
 {
     cleanruv_purge_data *purge_data = (cleanruv_purge_data *)arg;
+    Replica *replica = purge_data->replica;
+    cldb_Handle *cldb = replica_get_cl_info(replica);
 
+    pthread_mutex_lock(&(cldb->stLock));
     /* Make sure we have a change log, and we aren't closing it */
-    if (s_cl5Desc.dbState == CL5_STATE_CLOSED ||
-        s_cl5Desc.dbState == CL5_STATE_CLOSING) {
+    if (cldb->dbState == CL5_STATE_CLOSED) {
```
tbordaz: Shouldn't it be  'ldb->dbState != CL5_STATE_OPEN

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -1274,51 +1275,32 @@ replica_execute_ldif2cl_task(Replica *replica, char *returntext)
...
+                  "All replication updates will be rejected until the import is complete.\n",
                   replica_get_name(replica));
-    imprc = cl5ImportLDIF(clDir, fName, replica);
-    if (CL5_SUCCESS == imprc) {
+
+    rc = cl5ImportLDIF(clDir, fName, replica);
+    if (CL5_SUCCESS == rc) {
         slapi_log_err(SLAPI_LOG_INFO, repl_plugin_name,
                       "replica_execute_ldif2cl_task - Finished changelog import of replica \"%s\"\n",
                       replica_get_name(replica));
```
tbordaz, line 1289: rc = LDAP_SUCCESS so that replica_execute_ldif2cl_task returns LDAP rc

src/lib389/lib389/cli_conf/replication.py
```diff
@@ -1208,24 +1244,52 @@ def create_parser(subparsers):
...
     repl_set_per_backend_cl.add_argument('--trim-interval', help="The interval to check if the replication changelog can be trimmed")
+    repl_set_per_backend_cl.add_argument('--encrypt', action='store_true', help="Set the replication changelog to use encryption.  You must export & import the changelog after setting this.")
+    repl_set_per_backend_cl.add_argument('--disable-encrypt', action='store_true', help="Set the replication changelog to not use encryption.  You must export & import the changelog after setting this.")
 
     repl_get_per_backend_cl = repl_subcommands.add_parser('get-changelog', help='Display replication changelog attributes.')
     repl_get_per_backend_cl.set_defaults(func=get_per_backend_cl)
     repl_get_per_backend_cl.add_argument('--suffix', required=True, help='The suffix that uses the changelog')
 
-    repl_dump_cl = repl_subcommands.add_parser('dump-changelog', help='Decode Directory Server replication change log and dump it to an LDIF')
+    repl_dump_cl = repl_subcommands.add_parser('dump-changelog', help='Dump/export the Directory Server replication changelog to an LDIF')
```
tbordaz: If it is not too late we have to decide the naming of CL operation: dump/restore or export/import.
I tend to prefer export/import

src/lib389/lib389/cli_conf/replication.py
```diff
@@ -1208,24 +1244,52 @@ def create_parser(subparsers):
...
+    repl_dump_cl.add_argument('-o', '--output-file', required=True, help="Path name for the final result.")
+    repl_dump_cl.add_argument('-r', '--replica-root', required=True,
+                              help="Specify replica root whose changelog you want to export.")
+
+    repl_def_dump_cl = dump_subcommands.add_parser('default', help='Export the replication changelog to the server\'s default LDIF directory.')
+    repl_def_dump_cl.set_defaults(func=dump_def_cl)
+    repl_def_dump_cl.add_argument('-r', '--replica-root', required=True,
+                                  help="Specify replica root whose changelog you want to export.")
+
+    repl_restore_cl = repl_subcommands.add_parser('restore-changelog',
```
tbordaz: idem 'restore-changelog' or 'import-changelog'

tbordaz (commented):
Very nice patch @mreynolds389 !!

ldap/servers/plugins/replication/cl5_config.c
mreynolds389: No it is not.  So what this does, and the comment above about the "goto", is so we can set this over LDAP without it being rejected.  But it requires a server restart to taker effect.  Previously any attempt to change the config while the server was running resulted in an err 53.  Meaning you were forced to edit the dse.ldif while the server was stopped.  Now you can set it via ldapmodfy, and it won't take effect until the server is restarted.

ldap/servers/plugins/replication/cl5_config.c
mreynolds389: See comment below...

ldap/servers/plugins/replication/cl5_config.c
mreynolds389: See above..

src/lib389/lib389/cli_conf/replication.py
mreynolds389: I do like export/import better, but in RHEL 8 we already documented "dump-changelog" and "restore-changelog" from when Simon did the cl-dump port to python.  So I was trying to create the least amount of doc disturbance :-)  We can still change it though since it's not going to land in RHEL 8.

ldap/servers/plugins/replication/cl5_config.c
tbordaz: Okay. So it would probably be the same for CONFIG_CHANGELOG_SYMMETRIC_KEY (few lines above). Allow ldapmodify update  but do not set it in config.symmetricKey until the next restart.

mreynolds389: Would a user set this attribute?   I thought the "server" was the only thing that set this attribute.  If a user would add this value then I can add the exception

dirsrvtests/tests/suites/replication/changelog_encryption_test.py
mreynolds389: For replica tasks there is "task_finished()" so I used that...

src/lib389/lib389/replica.py
mreynolds389: It is used by the CLI, and now in the CI tests as well.

ldap/servers/plugins/replication/cl5_api.c
mreynolds389: This is already done in the same function a little higher up.  Am I missing something?

ldap/servers/plugins/replication/cl5_api.c
mreynolds389: done

ldap/servers/plugins/replication/cl5_api.c
mreynolds389: Yes, hard to say what we should do.  So I set the dbState to CL5_STATE_CLOSED, and the dbEnv to CL5_OPEN_NONE

ldap/servers/plugins/replication/cl5_config.c
mreynolds389: > Okay. So it would probably be the same for CONFIG_CHANGELOG_SYMMETRIC_KEY (few lines above). Allow ldapmodify update but do not set it in config.symmetricKey until the next restart.

Looks like this was always allowed.  So I'm going to leave it as it was.

ldap/servers/plugins/replication/cl5_api.c
mreynolds389: nice catch

ldap/servers/plugins/replication/repl5_replica_config.c
mreynolds389: This is fine.  CL5_SUCCESS does directly map to LDAP_SUCCESS.  If its not CL5_SUCCESS then we reset the rc to LDAP_OPERATIONS_ERROR

dirsrvtests/tests/suites/replication/changelog_encryption_test.py
mreynolds389: Fixed

dirsrvtests/tests/suites/replication/multiple_changelogs_test.py
mreynolds389: Fixed

mreynolds389:
Changes made, all the tests are passing.

I updated the changelog record (v6) to include a "encrypted" flag.  This allows for the exporting of changelogs that have a mix of encrypted and non-encrypted changes making it much for robust for exporting/importing/encryption/cert-renewal situtations

Please review...

ldap/servers/plugins/replication/cl5_api.c
tbordaz: Yes but shouldn't maxRUV, purgeRUv and enryCount be reread from the imported CL (so after the import loop) ?

ldap/servers/plugins/replication/cl5_api.c
tbordaz: thanks

ldap/servers/plugins/replication/cl5_api.c
tbordaz: Still think that trimming is not valid during import.

tbordaz:
Except the minor comment in trigger_cl_purging_thread to trim only if the state is CL5_STATE_OPEN. The patch looks very nice to me.

ACK

ldap/servers/plugins/replication/cl5_api.c
mreynolds389: I'll look into that...

tbordaz (approved):
Overall patch is good to me. I think there is a pending comment from @Firstyear

PR #4332, opened 2020-09-21
Issue 4322 - Fix a source link

### LINKED ISSUE #4322: Replace Pagure references with GitHub ones
The repo has a lot of references to pagure.io that should be replaced with github.com.
The list includes but is not limited to:

  - man pages;
  - lib389 docs;
  - cockpit-389-ds code comments;
  - core DS code comments;
  - dirsrvtests 'skipif' fixture;

droideck:
Description: Source0 should point to a local file instead of
a remote URL. We use it for testing/development only so
there is no need in external links.

Reviewed by: ?

Fixes: #4322

### DIFF
--- a/rpm/389-ds-base.spec.in
+++ b/rpm/389-ds-base.spec.in
@@ -177,7 +177,7 @@ Requires:         cracklib-dicts
 # Picks up our systemd deps.
 %{?systemd_requires}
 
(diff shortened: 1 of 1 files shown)


Firstyear:
Agreed, ack from me.

PR #4335, opened 2020-09-22
Issue4209

mreynolds389:


ldap/servers/plugins/replication/repl5_connection.c
```diff
@@ -1060,50 +1056,96 @@ conn_connect(Repl_Connection *conn)
         PR_Unlock(conn->lock);
         return return_value;
     }
+    return_value = conn_connect_with_bootstrap(conn, PR_FALSE);
+    if (return_value != CONN_OPERATION_SUCCESS &&
+        (conn->last_ldap_error == LDAP_INVALID_CREDENTIALS ||
```
tbordaz, line 1061: any reason to be restrictive ? For example if bootstrap is over LDAP and regular is over LDAPS, if secure port is disabled on consumer bootstrap will not be use.

tbordaz:
Except one question, the patch looks good to me

mreynolds389:
> Except one question, the patch looks good to me

What is the question?

ldap/servers/plugins/replication/repl5_connection.c
mreynolds389: There could be timeout errors, server down, etc, but we only need/should bootstrap if authentication fails.  

For conflicting configs, that is on the Admin IMO.  Both replicas should be configured the same in regards to TLS.

tbordaz: Okay thanks

tbordaz:
The patch is good. ACK

PR #4343, opened 2020-09-24
Issue 4342 - UI - additional fixes for creation instance modal

### LINKED ISSUE #4342: UI - fix issues with create instance validation
**Issue Description**

In the instance creation modal there is an incorrect warning about the port number range.  It should stats valid port numbers are between 1 and 65535.  The root DN character validation allows non ascii as the first characters after the "=".  And we are not forewarning about the instance name length if it is greater than 80 characters

mreynolds389:
Description: In the instance creation modal there is an incorrect warning
             about the port number range. It should state valid port numbers
             are between 1 and 65535. The root DN character validation allows
             non ascii as the first characters after the "=". And we are not
             forewarning about the instance name length if it is greater than
             80 characters

Fixes: https://github.com/389ds/389-ds-base/issues/4342

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/dsModals.jsx
+++ b/src/cockpit/389-console/src/dsModals.jsx
@@ -89,9 +89,11 @@ export class CreateInstanceModal extends React.Component {
         if (dn.endsWith(",")) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


droideck (approved):
Looks good to me!

PR #4344, opened 2020-09-28
4224 openldap entryuuid confusion

### LINKED ISSUE #4313: Syncrepl may send changelog entries twice causing client data corruption
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/51260
- Created at 2020-09-07 05:15:32 by [firstyear](https://pagure.io/user/firstyear) (@Firstyear)

---

#### Issue Description
While investigating 51171 and expanding the test cases, I noticed that it was possible for certain syncrepl events to be skipped causing client data corruption. This is when the following order of operations occurs

```
       server modrdn out of scope
client sync to state X
       server modrdn into scope
client sync state to X + 1
```

Further investigation showed that we would send previously seen changelog entries, and due to how these are processed while optimising out to remove entries that were added/deleted in succession, this would cause the sync repl events to not be sent. This leads to a state where clients would recieve the delete, but not the subsequent re-add as the server deems the delete followed by add as a no-op.

Firstyear:
Bug Description: OpenLDAP server as a syncrepl consumed enforces
    the condition that syncUUID in ldap messages must match the entryuuid
    of the entry. This is not in the RFC but it affects this one situation.

    Fix Description: To resolve this, we enforce that entryuuid is a
    requirement to the openldap syncrepl mode. Only entries with an
    entryuuid can be sent to openldap. Additionally, this mode is disabled
    by default by a configuration parameter "syncrepl-allow-openldap" in
    the content sync plugin config.

NOTE: It looks like the commits from 51260 which were previously merged seem to be missing after the migration, so this branch includes those. 

Thanks! 

Fixes: #4313 #4224

### DIFF
--- a/dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
+++ b/dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
@@ -144,7 +144,7 @@ def syncstate_assert(st, sync):
     # Find the primary uuid we expect to see in syncrepl.
     # This will be None if not present.
     acc_uuid = account.get_attr_val_utf8('entryuuid')
-    if acc_uuid is None:
+    if not sync.openldap:
         nsid = account.get_attr_val_utf8('nsuniqueid')
         # nsunique has a diff format, so we change it up.
         # 431cf081-b44311ea-83fdb082-f24d490e
@@ -173,7 +173,7 @@ def syncstate_assert(st, sync):
     assert 1 == len(sync.entries.keys())
     assert 1 == len(sync.present)
     ####################################
-    ## assert sync.present == [acc_uuid]
+    assert sync.present == [acc_uuid]
     assert 0 == len(sync.delete)
     if sync.openldap:
         assert True == sync.refdel
@@ -191,7 +191,7 @@ def syncstate_assert(st, sync):
     assert 1 == len(sync.entries.keys())
     assert 1 == len(sync.present)
     ####################################
-    ## assert sync.present == [acc_uuid]
+    assert sync.present == [acc_uuid]
     assert 0 == len(sync.delete)
     if sync.openldap:
         assert True == sync.refdel
@@ -210,12 +210,16 @@ def syncstate_assert(st, sync):
     assert 1 == len(sync.entries.keys())
     assert 1 == len(sync.present)
     ####################################
-    ## assert sync.present == [acc_uuid]
+    assert sync.present == [acc_uuid]
     assert 0 == len(sync.delete)
     if sync.openldap:
         assert True == sync.refdel
     else:
         assert False == sync.refdel
+
+    # import time
+    # print("attach now ....")
+    # time.sleep(45)
 
     ## Modrdn (out of scope, then back into scope)
     account.rename('uid=test1_modrdn', newsuperior=DEFAULT_SUFFIX)
@@ -228,11 +232,14 @@ def syncstate_assert(st, sync):
     log.debug(f"sd: {sync.delete}, sp: {sync.present} sek: {sync.entries.keys()}")
     assert 0 == len(sync.entries.keys())
     assert 0 == len(sync.present)
-    assert 0 == len(sync.delete)
+    ## WARNING: This test MAY FAIL here if you do not have a new enough python-ldap
+    # due to an ASN.1 parsing bug. You require at least python-ldap 3.3.1
+    assert 1 == len(sync.delete)
+    assert sync.delete == [acc_uuid]
     if sync.openldap:
         assert True == sync.refdel
     else:
-        assert True == sync.refdel
+        assert False == sync.refdel
 
     # Put it back
     account.rename('uid=test1_modrdn', newsuperior=OU_PEOPLE)
@@ -244,7 +251,7 @@ def syncstate_assert(st, sync):
     assert 1 == len(sync.entries.keys())
     assert 1 == len(sync.present)
     ####################################
-    ## assert sync.present == [acc_uuid]
+    assert sync.present == [acc_uuid]
     assert 0 == len(sync.delete)
     if sync.openldap:
         assert True == sync.refdel
@@ -253,10 +260,6 @@ def syncstate_assert(st, sync):
 
     ## Delete
     account.delete()
-
-    # import time
-    # print("attach now ....")
-    # time.sleep(45)
 
     # Check
     log.debug("*test* del")
@@ -270,11 +273,11 @@ def syncstate_assert(st, sync):
     log.debug(f"sd: {sync.delete}, sp: {sync.present} sek: {sync.entries.keys()}")
     assert 0 == len(sync.entries.keys())
     assert 0 == len(sync.present)
-    assert 0 == len(sync.delete)
+    assert 1 == len(sync.delete)
+    assert sync.delete == [acc_uuid]
     ####################################
-    ## assert sync.delete == [acc_uuid]
     if sync.openldap:
         assert True == sync.refdel
     else:
-        assert True == sync.refdel
+        assert False == sync.refdel
 
--- a/dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
+++ b/dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
@@ -21,6 +21,8 @@ from lib389._constants import *
 
 log = logging.getLogger(__name__)
 
+OU_PEOPLE = "ou=people,%s" % DEFAULT_SUFFIX
+
 class ISyncRepl(DirSrv, SyncreplConsumer):
     """
     This implements a test harness for checking syncrepl, and allowing us to check various actions or
@@ -28,6 +30,11 @@ class ISyncRepl(DirSrv, SyncreplConsumer):
     later to ensure that syncrepl worked as expected.
     """
     def __init__(self, inst, openldap=False):
+        ### 🚧 WARNING 🚧
+        # There are bugs with python ldap sync repl in ALL VERSIONS below 3.3.1.
+        # These tests WILL FAIL unless you have version 3.3.1 or higher!
+        assert ldap.__version__ >= '3.3.1'
+
         self.inst = inst
         self.msgid = None
 
@@ -55,6 +62,7 @@ class ISyncRepl(DirSrv, SyncreplConsumer):
         self.delete = []
         self.present = []
         self.entries = {}
+        self.refdel = False
         self.next_cookie = None
         # Start the sync
         # If cookie is none, will call "get_cookie" we have.
@@ -89,6 +97,9 @@ class ISyncRepl(DirSrv, SyncreplConsumer):
 
     def syncrepl_present(self, uuids, refreshDeletes=False):
         log.debug(f'=====> refdel -> {refreshDeletes} uuids -> {uuids}')
+        if refreshDeletes:
+            # Indicate we recieved a refdel in the process.
+            self.refdel = True
         if uuids is not None:
             self.present = self.present + uuids
 
@@ -105,8 +116,9 @@ class ISyncRepl(DirSrv, SyncreplConsumer):
 
 def syncstate_assert(st, sync):
     # How many entries do we have?
+    # We setup sync under ou=people so we can modrdn out of the scope.
     r = st.search_ext_s(
-        base=DEFAULT_SUFFIX,
+        base=OU_PEOPLE,
         scope=ldap.SCOPE_SUBTREE,
         filterstr='(objectClass=*)',
         attrsonly=1,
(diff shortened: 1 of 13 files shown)


ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -499,16 +556,29 @@ sync_read_entry_from_changelog(Slapi_Entry *cl_entry, void *cb_data)
...
+                      "Check 'cn=Retro Changelog Plugin,cn=plugins,cn=config' contains 'nsslapd-attribute: nsuniqueid:targetUniqueId'\n");
         return (1);
     }
+
+    /* If we were requested to do openldap mode, get the targetEntryuuid too */
+    if (cb->openldap_compat) {
+        entryuuid = sync_get_attr_value_from_entry(cl_entry, CL_ATTR_ENTRYUUID);
+        if (entryuuid == NULL) {
+            slapi_log_err(SLAPI_LOG_ERR, SYNC_PLUGIN_SUBSYSTEM,
+                          "sync_read_entry_from_changelog - Retro Changelog does not provide nsuniquedid."
```
tbordaz: .."Retro Changelog does not provide entryuuid."

ldap/servers/plugins/sync/sync_init.c
```diff
@@ -168,12 +168,36 @@ sync_start(Slapi_PBlock *pb)
...
+        /* Do we allow openldap sync? */
+        Slapi_Attr *chattr = NULL;
+        if (slapi_entry_attr_find(e, SYNC_ALLOW_OPENLDAP_COMPAT, &chattr) == 0) {
+            Slapi_Value *sval = NULL;
+            slapi_attr_first_value(chattr, &sval);
+
+            const struct berval *value = slapi_value_get_berval(sval);
+            if (NULL != value && NULL != value->bv_val && '\0' != value->bv_val[0]) {
+                if (strcasecmp(value->bv_val, "on") == 0) {
+                    allow_openldap_compat = true;
```
tbordaz: Please use PR_TRUE

ldap/servers/plugins/sync/sync_init.c
```diff
@@ -168,12 +168,36 @@ sync_start(Slapi_PBlock *pb)
 {
     int argc;
     char **argv;
+    Slapi_Entry *e = NULL;
+    bool allow_openldap_compat = false;
```
tbordaz: PRBool allow_openldap_compat = PR_FALSE;

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -893,9 +893,10 @@ sync_send_results(void *arg)
                     break;
                 }
                 ectrls = (LDAPControl **)slapi_ch_calloc(2, sizeof(LDAPControl *));
-                if (req->req_cookie)
+                if (req->req_cookie) {
                     sync_cookie_update(req->req_cookie, ec);
-                sync_create_state_control(ec, &ectrls[0], chg_type, req->req_cookie);
+                }
+                sync_create_state_control(ec, &ectrls[0], chg_type, req->req_cookie, false);
```
tbordaz: In persist mode, shouldn't we copy the client or session cookie and call sync_create_state_control with 'req->req_cookie->openldap_compat ?
Else the state control will contain syncuuid=nsuniqueid instead of entryuuid

ldap/servers/plugins/sync/sync_util.c
```diff
@@ -141,9 +174,26 @@ sync_create_state_control(Slapi_Entry *e, LDAPControl **ctrlp, int type, Sync_Co
 
     *ctrlp = NULL;
 
-    slapi_entry_attr_find(e, SLAPI_ATTR_UNIQUEID, &attr);
-    slapi_attr_first_value(attr, &val);
-    uuid = sync_nsuniqueid2uuid(slapi_value_get_string(val));
+    if (openldap_compat) {
```
tbordaz, line 177: There is something I miss here. Is this flag coming from the request cookie or from the global allow_openldap_compat ?

ldap/servers/plugins/sync/sync_util.c
```diff
@@ -608,7 +658,7 @@ sync_cookie_update(Sync_Cookie *sc, Slapi_Entry *ec)
 }
 
 Sync_Cookie *
-sync_cookie_parse(char *cookie, bool *cookie_refresh)
+sync_cookie_parse(char *cookie, bool *cookie_refresh, bool *allow_openldap_compat)
```
tbordaz: Please prefer PRBool, PR_TRUE, PR_FALSE

ldap/servers/plugins/sync/sync_util.c
```diff
@@ -625,6 +675,10 @@ sync_cookie_parse(char *cookie, bool *cookie_refresh)
 
     sc = (Sync_Cookie *)slapi_ch_calloc(1, sizeof(Sync_Cookie));
     if (strncmp(cookie, "rid=", 4) == 0) {
+        if (*allow_openldap_compat != true) {
+            slapi_log_err(SLAPI_LOG_ERR, SYNC_PLUGIN_SUBSYSTEM, "sync_cookie_parse - An openldap sync request was made, but " SYNC_ALLOW_OPENLDAP_COMPAT " is false\n");
+            goto error_return;
```
tbordaz, line 681: Indicate what need to be done (set syncrepl-allow-openldap in plugin entry)

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -322,10 +331,35 @@ sync_refresh_update_content(Slapi_PBlock *pb, Sync_Cookie *client_cookie, Sync_C
...
+     * if 5 was an "add" of the entry, and 6 was a "delete" of the same entry then sync
+     * would over-optimise and remove the sync value because it things the add/delete was
+     * in the same operation so we'd never send it. But the client HAD seen the add, and
+     * now we'd never send the delete so this would be a bug. This created some confusion
+     * for me in the tests, but the sync repl tests now correctly work and reflect the behaviour
+     * expected.
+     */
+    if (server_cookie->openldap_compat) {
+        /* In openldap compat we only want items that have an entryuuid, else we can't sync them */
+        filter = slapi_ch_smprintf("(&(changenumber>=%lu)(changenumber<=%lu)(targetEntryUUID=*))",
```
tbordaz: Use CL_ATTR_ENTRYUUID

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -606,27 +696,36 @@ sync_read_entry_from_changelog(Slapi_Entry *cl_entry, void *cb_data)
...
         if (prev == -1) {
+            slapi_log_err(SLAPI_LOG_PLUGIN, SYNC_PLUGIN_SUBSYSTEM, "sync_read_entry_from_changelog - %s LDAP_REQ_DELETE\n", uniqueid);
             cb->cb_updates[index].upd_chgtype = LDAP_REQ_DELETE;
             cb->cb_updates[index].upd_uuid = uniqueid;
+            cb->cb_updates[index].upd_euuid = entryuuid;
             cb->cb_updates[index].upd_e = sync_deleted_entry_from_changelog(cl_entry);
         } else {
             /* if it was added since last cookie state, we
-                 * can ignoere it */
+             * can ignoere it */
```
tbordaz: type 'ignore'

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -641,12 +740,19 @@ sync_send_deleted_entries(Slapi_PBlock *pb, Sync_UpdateNode *upd, int chg_count,
...
+    PR_ASSERT(cookie);
+
     syncUUIDs[0] = NULL;
     for (size_t index = 0; index < chg_count; index++) {
-        if (upd[index].upd_chgtype == LDAP_REQ_DELETE &&
-            upd[index].upd_uuid) {
+        if (upd[index].upd_chgtype == LDAP_REQ_DELETE && upd[index].upd_uuid) {
             if (uuid_index < SYNC_MAX_DELETED_UUID_BATCH) {
-                syncUUIDs[uuid_index] = sync_nsuniqueid2uuid(upd[index].upd_uuid);
+                if (upd[index].upd_euuid) {
```
tbordaz, line 756: Shouldn't we test  cookie->openldap_compat to select uuid or nsuniqueid ?

tbordaz (commented):
@Firstyear the overall patch looks very nice. I have minor comments. However I would appreciate a design (port389) describing the solution and the choices

ldap/servers/plugins/sync/sync_init.c
Firstyear: This is not a PRBOOL type, it's a C bool type, that's why it uses "true" not PR_TRUE. I can change it to PR_BOOL if you prefer.

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: upd_euuid is only set if openldap_compat is true

ldap/servers/plugins/sync/sync_util.c
Firstyear: openldap_compat in the cookie is *only* true if the initial request is openldap format (rid=) AND syncrepl-allow-openldap is true.

ldap/servers/plugins/sync/sync_init.c
Firstyear: Ahh you did req the change to PR bool, okay I'll change it.

ldap/servers/plugins/sync/sync_persist.c
Firstyear: Yep, I think you're right. I think for now it may be easier to actually disable persistent with openldap mode instead.

Firstyear:
@tbordaz Thanks for your review, I have updated all the areas you commented on, and I have updated https://www.port389.org/docs/389ds/design/openldap2ds.html with details about the syncrepl changes.

Firstyear:
@tbordaz, @mreynolds389  Review reminder :)

Firstyear:
Ping @tbordaz and @mreynolds389 :)

Firstyear:
Ping @tbordaz and @mreynolds389,

mreynolds389:
@tbordaz has been doing a lot of work around sync repl lately.  I'd prefer him to do this review as it could impact what he is working on

dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
```diff
@@ -115,49 +122,157 @@ def syncstate_assert(st, sync):
...
+    assert 1 == len(sync.present)
+    ####################################
+    assert sync.present == [acc_uuid]
+    assert 0 == len(sync.delete)
+    if sync.openldap:
+        assert True == sync.refdel
+    else:
+        assert False == sync.refdel
+
+    ## ModRdn (remain in scope)
```
tbordaz, line 201: In my test MODRDN (in/out scope) are both failing

dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
```diff
@@ -115,49 +122,157 @@ def syncstate_assert(st, sync):
...
 
     ## Delete
     account.delete()
 
     # Check
     log.debug("*test* del")
-    sync.syncrepl_search()
+    sync.syncrepl_search(base=OU_PEOPLE)
     sync.syncrepl_complete()
     # In a delete, the cookie isn't updated (?)
```
tbordaz, line 268: Indeed the cookie should be updated and it was related to another bug.

dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
```diff
@@ -115,49 +122,157 @@ def syncstate_assert(st, sync):
...
     sync.check_cookie()
     log.debug(f'{sync.entries.keys()}')
     log.debug(f'{sync.present}')
     log.debug(f'{sync.delete}')
+    log.debug(f"sd: {sync.delete}, sp: {sync.present} sek: {sync.entries.keys()}")
     assert 0 == len(sync.entries.keys())
     assert 0 == len(sync.present)
     assert 1 == len(sync.delete)
```
tbordaz, line 276: How can I debug that ?
Sync_repl sends an ID_SET containing the deleted entry. But sync.delete does not contain it. How should I debug this ?

Firstyear:
Yeah, that's fair @mreynolds389. I'll poke him some more then about the matter :)

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -705,13 +815,14 @@ sync_send_entry_from_changelog(Slapi_PBlock *pb, int chg_req __attribute__((unus
...
-        if (entries)
+        if (entries) {
             db_entry = *entries; /* there can only be one */
+        }
     }
 
     if (db_entry && sync_is_entry_in_scope(pb, db_entry)) {
         LDAPControl **ctrl = (LDAPControl **)slapi_ch_calloc(2, sizeof(LDAPControl *));
-        sync_create_state_control(db_entry, &ctrl[0], chg_type, NULL);
+        sync_create_state_control(db_entry, &ctrl[0], chg_type, NULL, cookie->openldap_compat);
```
tbordaz: I think you should catch the case of a failure of sync_create_state_control and skip that entry
At this point the entry had an entryuuid in the *changelog*. But if for some reason the entry lost its entryuuid in the DB then the state control will be empty and the sent entry has no entryuuid.

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -89,7 +90,7 @@ sync_srch_refresh_pre_search(Slapi_PBlock *pb)
              * when using their changelog mode. As a result, we parse the cookie to handle this
              * shenangians to determine if this is valid.
              */
-            client_cookie = sync_cookie_parse(cookie, &cookie_refresh);
+            client_cookie = sync_cookie_parse(cookie, &cookie_refresh, &allow_openldap_compat);
```
tbordaz, line 93: Did you test with a server that disallow a openldap client doing content_sync search ?
In such case the client_cookie will be NULL but we will send entries without taking care of entryuuid and it can confuse the client. Should not we stop the search ?

In addition if client_cookie is NULL, some paths of code can dereference it.

tbordaz:
@Firstyear except those two last comments/questions, the overall patch looks nice to me

Firstyear:
Hey mate, I don't see the two comments/questions you are referring to? All the comments I can see are your older ones which have been addressed?

tbordaz:
@Firstyear , definitely something I do not understand with PR. Some comments can remain "hidden" and some others are public.. I do not know what to do to make all my comments public. Can you see my two comments (sync_create_state_control, null cookie) ?

Firstyear:
> @Firstyear , definitely something I do not understand with PR. Some comments can remain "hidden" and some others are public.. I do not know what to do to make all my comments public. Can you see my two comments (sync_create_state_control, null cookie) ?

If you start a review, your comments are "staged" in your browser, and only if you click "complete review" in the top right do they become visible to me and others. Maybe that's what is going on?

dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
Firstyear: This test *may* fail due to a bug in python LDAP. Can you confirm your python version by doing:

```
python3
import ldap
help(ldap)
```

Then look for "VERSION" at the foot of the help? It should be 3.3.1 or greater.

dirsrvtests/tests/suites/syncrepl_plugin/__init__.py
Firstyear: See my other comment. This is likely the python-ldap bug in syncrepl.

Firstyear:
@tbordaz I think your python tests are failing because of this https://github.com/python-ldap/python-ldap/pull/351 

I indicated in a comment how to check your python ldap version for this :)

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: I'll check this, I was pretty sure that there is no NULL derefs. Which code paths can this happen in?

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: I'm trying to understand this comment, but maybe I can't quite see what you are thinking of. Are you saying that if someone were to LDAPMOD and del the entryuuid attr that would cause a problem?

tbordaz: Sorry for this not clear remark.

sync_create_state_control can fails (if the db entry has no SLAPI_ATTR_ENTRYUUID). In such case we should not send the entry. So the suggestion is to test the returned code of sync_create_state_control to skip the entry.

tbordaz:
> @tbordaz I think your python tests are failing because of this [python-ldap/python-ldap#351](https://github.com/python-ldap/python-ldap/pull/351)
> 
> I indicated in a comment how to check your python ldap version for this :)

Exactly !! you can not imagine the amount of time I spent to identify that the failure was due to incorrect python-ldap failing to decode syncInfo. As I am still testing with prefix_install I was using an old version of python-ldap. Thanks for having fixed that.
This python-ldap bug impacted test_syncrepl_basic but others sync_repl tests were failing because of invalid testcase. Now all is fixed upstream

ldap/servers/plugins/sync/sync_refresh.c
tbordaz: If sync_cookie_parse is called with a cookie containing 'rid=' and allow_openldap_compat==False, then client_cookie is NULL. Later if this is a refreshPersist, it will dereferenced few lines after '_if (client_cookie->openldap_compat == PR_TRUE) .._'

ldap/servers/plugins/sync/sync_util.c
```diff
@@ -625,16 +675,21 @@ sync_cookie_parse(char *cookie, bool *cookie_refresh)
 
     sc = (Sync_Cookie *)slapi_ch_calloc(1, sizeof(Sync_Cookie));
     if (strncmp(cookie, "rid=", 4) == 0) {
+        if (*allow_openldap_compat != PR_TRUE) {
```
tbordaz, line 678: It looks that the detection of openldap cookie is a bit fragile.
If the cookie is '_host:389#uid=foo,nsServerID=123,dc=com:dc=com:(objectclass=*)#4321_' then '_nsServerID_' will erroneously match the condition.

ldap/servers/plugins/sync/sync_refresh.c
tbordaz: Indeed, running the sync_repl tests shows this deref:
<pre>
#0  0x00007fdc9e741494 in sync_srch_refresh_pre_search (pb=0x7fdbe4000b60) at ldap/servers/plugins/sync/sync_refresh.c:109
109	                if (client_cookie->openldap_compat == PR_TRUE) {
[Current thread is 1 (Thread 0x7fdc06ffd700 (LWP 1329659))]
(gdb) where
#0  0x00007fdc9e741494 in sync_srch_refresh_pre_search (pb=0x7fdbe4000b60) at ldap/servers/plugins/sync/sync_refresh.c:109
#1  0x00007fdca21c44ad in plugin_call_func (list=0x24d73d0, operation=403, pb=0x7fdbe4000b60, call_one=0) at ldap/servers/slapd/plugin.c:2002
#2  0x00007fdca21c4316 in plugin_call_list (list=0x24bbef0, operation=403, pb=0x7fdbe4000b60) at ldap/servers/slapd/plugin.c:1945
#3  0x00007fdca21c10a3 in plugin_call_plugins (pb=0x7fdbe4000b60, whichfunction=403) at ldap/servers/slapd/plugin.c:414
#4  0x00007fdca21b2076 in op_shared_search (pb=0x7fdbe4000b60, send_result=1) at ldap/servers/slapd/opshared.c:588
#5  0x00000000004372e8 in do_search (pb=0x7fdbe4000b60) at ldap/servers/slapd/search.c:388
#6  0x0000000000418ea9 in connection_dispatch_operation (conn=0x33c7588, op=0x24968d0, pb=0x7fdbe4000b60) at ldap/servers/slapd/connection.c:659
#7  0x000000000041af0f in connection_threadmain () at ldap/servers/slapd/connection.c:1770
#8  0x00007fdca1ca4b34 in _pt_root () from /lib64/libnspr4.so
#9  0x00007fdca1c374e2 in start_thread () from /lib64/libpthread.so.0
#10 0x00007fdca1b0b6a3 in clone () from /lib64/libc.so.6
(gdb) print client_cookie
$1 = (Sync_Cookie *) 0x0
</pre>

dirsrvtests/tests/suites/syncrepl_plugin/openldap_test.py
```diff
@@ -45,18 +46,25 @@ def test_syncrepl_openldap(topology):
         1. Success
     """
     st = topology.standalone
+    # Ensure entryuuid is setup
+    plug = EntryUUIDPlugin(st)
+    task = plug.fixup(DEFAULT_SUFFIX)
```
tbordaz, line 52: Note that this testcase fails for me. The entryUUID tasks container does not exist. 

<pre>
[04/Dec/2020:10:09:14.190071590 +0100] conn=1 op=5 ADD dn="cn=entryuuid_fixup_2020-12-04T10:09:14.188531,cn=entryuuid task,cn=tasks,cn=config"
[04/Dec/2020:10:09:14.192727883 +0100] conn=1 op=5 RESULT err=32 tag=105 nentries=0 wtime=0.000286806 optime=0.002672704 etime=0.002953848

</pre>

Firstyear: Have you set --enable-rust? EntryUUID only works if you have rust ... 

Saying that I should check that the tests in this suite *only* operate if enable_rust is true, so that's a good point.

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: Interesting, I'll check this out :)

ldap/servers/plugins/sync/sync_util.c
Firstyear: Good catch, I wonder if there is some way to bound the strncmp here ...

Firstyear: Re-reading, strncmp will only compare the literal first for characters, so your nsServerid=123 example won't work because it will actually compary `rid=` to `host`.

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: I'm not sure about this. If the entry has lost it's entryUUID, then that would indicate db corruption or tampering, so the  `sync_create_state_control` should fail - and does, with "sync_create_state_control - Some entries are missing entryUUID. Unable to proceed. You may need to re-run the entryuuid fixup". If the entry *had* a entryUUID and it was removed, this also indicates tampering so the error is valid. It's worth noting the syntax of entryuuid is: `SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation` which prevents user modification of the value, and prevents users *removing* the value. So if the entryUUID goes missing, then this is an error and we SHOULD fail the whole sync repl operation for openldap because it means we could potentially replicate and corrupt data silently. So this is the correct solution to fail fast and hard to prevent replicating corrupted data (especially as openldap's replication processing for incoming data is very fragile).

Firstyear:
@tbordaz Okay, I've updated this - I've fixed the null deref, and added flags to the tests to prevent them running if python ldap is too old, as well as checkingf for enable-rust which gives the entry uuid support. I've run all the sync repl tests and they all pass.

ldap/servers/plugins/sync/sync_refresh.c
tbordaz: I agree, if it detects that entryUUID is missing (so sync_create_state_control fails) it should fails. This is what I wanted to say in my initial comment.
The decision to just skip this update or to stop the full sync_repl search is up to you. I think it makes sense to break the search as the missing entryUUID indicates 389-DS is not longer able top sync with openldap.

tbordaz:
> @tbordaz Okay, I've updated this - I've fixed the null deref, and added flags to the tests to prevent them running if python ldap is too old, as well as checkingf for enable-rust which gives the entry uuid support. I've run all the sync repl tests and they all pass.

Sorry, I can not see the changes in the patch. Am I missing something ?

Firstyear:
> Sorry, I can not see the changes in the patch. Am I missing something ?

Nope, I made a mistake in my push - I pushed the branch to the upstream project, rather than my fork. No damage done to upstream since it was a unique branch name, so I have deleted the mistake, and now pushed to the correct branch. You should now see the changes. Sorry about that :(

ldap/servers/plugins/sync/sync_refresh.c
Firstyear: Ahhhhhh I see what you are saying now! Because the `rc` of `sync_create_state_control` isn't checked the entry is skipped rather than the search terminated. So I think here, we should have an `rc = sync_create_state_control(...)` and then if `!= 0`, we return the rc similar to `sync_srch_refresh_pre_entry`

Firstyear: Add rc check :) thank you!

tbordaz (approved):
Thanks for the patch. LGTM. Ack

Firstyear:
Thanks @tbordaz I might ask @mreynolds389 getting a second look over as well. Really appreciate your help :)

mreynolds389:
LGTM

PR #4346, opened 2020-09-28
Ticket 4345 - import self sign cert doc comment

### LINKED ISSUE #4345: Improve self-sign-cert doc in template.inf
It was raised that the message in template.inf may be misleading, that it doesn't indicate that disabling this *also* disable nsslapd-security which causes later import of certs to fail. We should improve this to help users, but annoyingly, it's an edge case we can't do much about beside document it.

Firstyear:
Bug Description: It was raised that the doc comment with TLS
and self sign cert could be confusing and if disabled it was
not clear how to enable TLS later.

Fix Description: Improve the doc comment with examples.

fixes: #4345

Review by: ???


droideck (approved):
LGTM!

PR #4349, opened 2020-09-29
Issue 4348 - Add tests for dsidm

### LINKED ISSUE #4348: Add tests for dsidm
I am creating this issue to keep track of commits for dsidm tests I am working on.
dsidm didn't get enough testing unlike dsconf/dsctl, so we have to fill a blank spot in our automated tests.

bsimonova:
Description:
Created tests for dsidm client_config option and enhanced
the src/lib389/lib389/cli_idm/client_config.py so the output gets caught
to a log file and can be compared.
Also modified the dbgen_test.py, because the check_value_in_log_and_reset() function
provided a fake value in the log file. So in case something failed and the original value
was not in the output, the test would find the value in "Check that {} is present" line
and therefore it would not fail.

Fixes: https://github.com/389ds/389-ds-base/issues/4348

Reviewed by: ???


dirsrvtests/tests/suites/clu/dsidm_config_test.py
```diff
@@ -0,0 +1,216 @@
...
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import time
+import subprocess
+import pytest
+
+from lib389.cli_idm.client_config import *
+from lib389.tasks import *
+from lib389.utils import *
```
droideck: Could you, please, specify the exact imports here? 

`import *` is too broad and it's a bad coding habit we better get rid of from our source :)

droideck (commented):
The rest looks good to me!

src/lib389/lib389/cli_idm/client_config.py
```diff
@@ -269,6 +282,23 @@ def display(inst, basedn, log, args):
         group_member='member',
     ))
 
+    log.info(DISPLAY_TEMPLATE.format(
```
Firstyear: Are you sure about log.info here? IIRC info as a level is always printed from the CLI so maybe this should be debug?

bsimonova:
@Firstyear @droideck 
I changed the log.info to log.debug and specified exact imports.
Please review, thanks!

droideck (approved):
Looks good to me!

bsimonova:
@droideck @Firstyear 
Could you please review again? 
I just did a small change with checking "ldap_access_filter" according to these changes https://github.com/389ds/389-ds-base/pull/4354

droideck:
> @droideck @Firstyear
> Could you please review again?
> I just did a small change with checking "ldap_access_filter" according to these changes #4354

Sure, but, please, avoid 'force-push' tactic... It's impossible to see what was changed if you `did a small change` :)
Just push a new commit to the branch.

bsimonova:
Ahh, sorry, I am still using commands like for pagure out of habit :)

droideck (approved):
LGTM!

PR #4352, opened 2020-10-01
Ticket 4347 - log when server requires a restart for a plugin to become active

### LINKED ISSUE #4347: MemberOf Task object isn't created until after a restart
After enabling the memberof plugin live, the task area is not created until after a restart. This means you can't run a memberof fixup until after you restart the directory server at least once.

Firstyear:
Bug Description: When testing another feature, user confusion was exprierenced
while enabling a plugin. This was due to a misunderstanding about dynamic
plugins and when and how they take effect.

Fix Description: When dynamic plugins are NOT enabled, and an attempt
is made to enable a plugin, clearly log that we require a restart of
the server for this functionality to become active.

fixes: #4347

Review by: ???

### DIFF
--- a/dirsrvtests/tests/suites/dynamic_plugins/notice_for_restart_test.py
+++ b/dirsrvtests/tests/suites/dynamic_plugins/notice_for_restart_test.py
@@ -1,0 +1,51 @@
+
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 William Brown <william@blackhats.net.au>
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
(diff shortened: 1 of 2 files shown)


droideck (changes requested):
Looks reasonable and the implementation is good!

One tiny thing, can we have a small test case for the `nsslapd-dynamic-plugins=off` option which checks the message?

Firstyear:
Sure, that should be easy to do.

Firstyear:
@droideck Test provided in update, already checked that it works as intended :)

droideck (approved):
Looks great! Thanks!

mreynolds389:
LGTM

Firstyear:
Double ack! It must be the good code then :)

PR #4353, opened 2020-10-01
Ticket 4350 - dsrc should warn when tls_cacertdir is invalid

### LINKED ISSUE #4350: Lib389 dsrc should warn when tls cacertdir provided is a file, not a directory.
It's common for TLS cacert options to list a single file *or* a directory. This can confuse people because the naming is subtle, so in the case we are provided with a file (probably a ca.pem) then we should warn the user and provide the steps to rectify this ( change to dir, run c_rehash).

Firstyear:
Bug Description: When the cacertdir is not a directory
or does not exist we should warn that this is not valid
and provide recification steps.

Fix Description: Check if the path exists or is a directory
and report this, along with steps on how to run c_rehash

fixes: #4350

Review by: ???

### DIFF
--- a/src/lib389/lib389/cli_base/dsrc.py
+++ b/src/lib389/lib389/cli_base/dsrc.py
@@ -137,6 +137,12 @@ def dsrc_to_ldap(path, instance_name, log):
         raise Exception("%s [%s] saslmech must be one of EXTERNAL or PLAIN" % (path, instance_name))
 
     dsrc_inst['tls_cacertdir'] = config.get(instance_name, 'tls_cacertdir', fallback=None)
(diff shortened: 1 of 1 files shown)


droideck (approved):
LGTM!

Firstyear:
PS: @droideck in the "review by" do you prefer me to keep using spichugi or do you want me to use your github profile name?

droideck:
> PS: @droideck in the "review by" do you prefer me to keep using spichugi or do you want me to use your github profile name?

I've started to use `@username` in `reviewed by:` as it increases discoverability.
So feel free to use my username.:)

PR #4354, opened 2020-10-01
Ticket 4351 - improve generated sssd.conf output

### LINKED ISSUE #4351: Improvements to dsidm
Some minor issues were raised:

* When group is set to none, ldap_access_filter = None is set which is not valid. We should emita  comment in this case instead. 
* user status has no attribute is_locked. (fixed in master already)
* warn on sssd config if uri is ldapi

Firstyear:
Bug Description: There were some subtle issues in the sssd.conf
generator. When no group was specified, we'd generate an invalid
config. When the config used ldapi, it may not work on remote
servers.

Fix Description: When the uri is ldapi, emit a warning for
this parameter to be reviewed. When ldap filter is none
provide the example as commented out.

fixes: #4351

Review by: ???


droideck (approved):
LGTM!

PR #4355, opened 2020-10-01
Issue 4297- On ADD replication URP issue internal searches with filte…

tbordaz:
…r containing unescaped chars

Bug description:
	In MMR a consumer receiving a ADD has to do some checking based on basedn.
	It checks if the entry was a tombstone or if the conflicting parent entry was a tombstone.

	To do this checking, URP does internal searches using basedn.
	A '*' (ASTERISK) is valid in a RDN and in a DN. But using a DN in an assertionvalue of a filter, the ASTERISK needs to be escaped else the server will interprete the filtertype to be a substring. (see
	https://tools.ietf.org/html/rfc4515#section-3)

	The problem is that if a added entry contains an ASTERISK in the DN, it will not be escaped in internal search and trigger substring search (likely unindexed).

Fix description:
	escape the DN before doing internal search in URP

https://github.com/389ds/389-ds-base/issues/4297 (ex. https://pagure.io/389-ds-base/issue/51244)

Reviewed by:  Mark Reynolds (thanks !)

Platforms tested: F31

### DIFF
--- a/dirsrvtests/tests/suites/replication/acceptance_test.py
+++ b/dirsrvtests/tests/suites/replication/acceptance_test.py
@@ -7,6 +7,7 @@
 # --- END COPYRIGHT BLOCK ---
 #
 import pytest
+import logging
 from lib389.replica import Replicas
 from lib389.tasks import *
 from lib389.utils import *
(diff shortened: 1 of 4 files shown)


Firstyear:
Ack from me too, this looks good!

droideck (approved):
One small thing - instead of 
`https://github.com/389ds/389-ds-base/issues/4297 (ex. https://pagure.io/389-ds-base/issue/51244)` in the commit message, could you please use 

`Fixes: #4297`

It will automatically link and close the original issue (and the issue has a link to the old Pagure issue so there is no need in specifying it here).

tbordaz:
Thanks for the reviews

PR #4356, opened 2020-10-05
Issue 4329 - Sync repl - if a serie of updates target the same entry …

### LINKED ISSUE #4329: Sync repl - if a serie of updates target the same entry then the cookie get wrong changenumber
**Issue Description**
If several updates target the same returned entry, then cookie keeps the changenumber of the first update. 

regression introduced pagure ticket 51190

 - Platform: all

**Steps to Reproduce**
create a user with mep plugin creating a managed group

**Expected results**
The changenumber of the cookie should change at each notified updates

tbordaz:
…then the cookie get wrong changenumber

Bug description:
            In persist mode, sync_repl sends a matching updated entry with a sync state control
            containing a cookie. The cookie contains the changenumber related to the updated entry.
	    If several consecutive updates targets the same entry, sync_repl will send for each
            update the same changenumber (the first of the set of updates).
            changenumber will resync as soon as another entry is sent.
            The reason why sync_repl sends several time the same entry is that the internal
            search looks for '(changenumber >= cookie_changenumber)' rather than
            '(changenumber > cookie_changenumber)'.

Fix description:
            Change the filter to look for the next changenumber

Fixes: #4329

Reviewed by:  ?

Platforms tested: F31, F33

### DIFF
--- a/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
+++ b/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
@@ -14,12 +14,13 @@ from ldap.syncrepl import SyncreplConsumer
 from ldap.ldapobject import ReconnectLDAPObject
 import pytest
 from lib389 import DirSrv
+from lib389.idm.organizationalunit import OrganizationalUnits, OrganizationalUnit
 from lib389.idm.user import nsUserAccounts, UserAccounts
 from lib389.idm.group import Groups
 from lib389.topologies import topology_st as topology
 from lib389.paths import Paths
 from lib389.utils import ds_is_older
-from lib389.plugins import RetroChangelogPlugin, ContentSyncPlugin, AutoMembershipPlugin, MemberOfPlugin, MemberOfSharedConfig, AutoMembershipDefinitions
+from lib389.plugins import RetroChangelogPlugin, ContentSyncPlugin, AutoMembershipPlugin, MemberOfPlugin, MemberOfSharedConfig, AutoMembershipDefinitions, MEPTemplates, MEPConfigs, ManagedEntriesPlugin, MEPTemplate
 from lib389._constants import *
 
 from . import ISyncRepl, syncstate_assert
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/sync/sync_util.c
```diff
@@ -487,7 +487,7 @@ sync_cookie_get_change_number(int lastnr, const char *uniqueid)
     LDAPControl **ctrls = NULL;
     
     ctrls = (LDAPControl **)slapi_ch_calloc(2, sizeof(LDAPControl *));
-    char *filter = slapi_ch_smprintf("(&(changenumber>=%d)(targetuniqueid=%s))", lastnr, uniqueid);
+    char *filter = slapi_ch_smprintf("(&(changenumber>=%d)(targetuniqueid=%s)(!(changenumber=%d)))", lastnr, uniqueid, lastnr);
```
Firstyear: Instead of adding a not condition, why not use lastnr + 1?

Firstyear:
Beside that comment, looks good :) Good find, I missed the persistent version in my previous fix :)

ldap/servers/plugins/sync/sync_util.c
tbordaz: Initially I was thinking doing this. In fact one of the patch you did also modified a filter with nr+1.
It works but it makes some assumption on the changenumber (numeric and increased by one). I have no strong opinion on one option or the other. 

One consideration is that we realized that using retroCL to implement sync_repl had some drawbacks and that possibly in the future sync_repl will rely on the replication changelog. In such case, update ID will no longer be numeric/incremented by one.

But so many things will need to be changed in sync_repl if we move from retroCL to replication changelog that I agree with you that it is better to follow your proposal. I will update the patch.

dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
```diff
@@ -147,6 +148,91 @@ def run(self):
         log.info("ZZZ result = %s" % self.result)
         self.conn.unbind()
 
+def test_sync_repl_mep(topology, request):
+    """Test sync repl with MEP plugin that triggers several
+    updates on the same entry
+
+    :id: 9b87493b-0493-46f9-8364-6099d0e5d806
```
droideck: I see that ID is not unique across the tests.
I found two more:
dirsrvtests/tests/suites/plugins/acceptance_test.py
dirsrvtests/tests/suites/plugins/managed_entry_test.py

Could you please generate a new one here? Ideally with `python3 -c "import uuid; print(uuid.uuid4())"` command. :)

droideck (approved):
The test looks good to me! Ack for that part!

ldap/servers/plugins/sync/sync_util.c
Firstyear: Yeah I think it would be a lot of work to move sync repl to the main changelog. It would effectively be a complete rewrite I think. So then the decision we make here won't affect a future change like that. Thanks for updating this!

Firstyear:
With that, Ack from me (provided the test case uuid's are updated as per @droideck's comment). Thanks!

PR #4358, opened 2020-10-06
Ticket 4350 - One line, fix invalid type error in tls_cacertdir check

### LINKED ISSUE #4350: Lib389 dsrc should warn when tls cacertdir provided is a file, not a directory.

Firstyear:
Bug Description: When the tls_cacertdir parameter was not
present os.path fails with None not a str.

Fix Description: Check if the path is None

fixes: #4350

Review by:


mreynolds389:
ack

PR #4359, opened 2020-10-06
Ticket 4334 - Task timeout may cause larger dataset imports to fail

### LINKED ISSUE #4334: Python task timeout updates.
**Issue Description**
Importing a file with 1 million users generates an import task failed message.

 - Platform: Fedora 32 - 5.6.6-300.fc32.x86_64 (Cloud Edition)
 - Package and version: 389-ds-base-1.4.3.12-1.fc32.x86_64

**Steps to Reproduce**
1. Increase the default import cache size from 16M to 100M. Importing the 1 M user file failed when using the default import cache size.
  
    image

   ldapmodify -D "cn=dm"  -w password  -f update.ldif

2. Verify updates.
   ldapsearch -D cn=dm  -b "cn=bdb,cn=config,cn=ldbm database,cn=plugins,cn=config" -w password objectclass=* nsslapd-import-cachesize

    
    image

3. Generate an ldif file containing 1 million users.
sudo dsctl inst01 ldifgen users --number 1000000 --suffix=dc=example,dc=com --ldif-file users.ldif

4. Import file into instance backend.
sudo dsconf inst01 backend import userroot users.ldif

    image

5. Request a search for everything from the server, works as expected.
ldapsearch -D cn=dm -w password  objectclass=*

    image

**Expected results**
It was expected that a search of the server DB would fail following the import task error. This error message could prevent some user automation maybe.

**Additional context**

Firstyear:
Bug Description: The task.wait() function had a hardcoded timeout
and no method to "disable" that check. This could cause very large
databases to fail to import.

Fix Description: Support timeout=None, which allows the task to
take 'infinite' time. Additionally, this provides a warning that
this is occuring.

fixes: #4334

Review by: ???

### DIFF
--- a/src/lib389/lib389/cli_conf/backend.py
+++ b/src/lib389/lib389/cli_conf/backend.py
@@ -241,7 +241,7 @@ def backend_import(inst, basedn, log, args):
     task = mc.import_ldif(ldifs=args.ldifs, chunk_size=args.chunks_size, encrypted=args.encrypted,
                           gen_uniq_id=args.gen_uniq_id, only_core=args.only_core, include_suffixes=args.include_suffixes,
                           exclude_suffixes=args.exclude_suffixes)
-    task.wait()
+    task.wait(timeout=None)
     result = task.get_exit_code()
 
     if task.is_complete() and result == 0:
@@ -269,7 +269,7 @@ def backend_export(inst, basedn, log, args):
                           encrypted=args.encrypted, min_base64=args.min_base64, no_dump_uniq_id=args.no_dump_uniq_id,
                           replication=args.replication, not_folded=args.not_folded, no_seq_num=args.no_seq_num,
                           include_suffixes=args.include_suffixes, exclude_suffixes=args.exclude_suffixes)
-    task.wait()
+    task.wait(timeout=None)
     result = task.get_exit_code()
 
     if task.is_complete() and result == 0:
--- a/src/lib389/lib389/tasks.py
+++ b/src/lib389/lib389/tasks.py
@@ -81,7 +81,10 @@ class Task(DSLdapObject):
         """Wait until task is complete."""
 
         count = 0
-        while count < timeout:
+        if timeout is None:
+            self._log.debug("No timeout is set, this may take a long time ...")
+
+        while timeout is None or count < timeout:
             if self.is_complete():
                 break
             count = count + 1


Firstyear:
cc @jchapma who reported this issue originally.

mreynolds389:
LGTM

Firstyear:
Thanks mate, I'll let @droideck have a look too since he was involved in the discussion on the ticket :) :)

src/lib389/lib389/cli_conf/backend.py
```diff
@@ -241,7 +241,7 @@ def backend_import(inst, basedn, log, args):
     task = mc.import_ldif(ldifs=args.ldifs, chunk_size=args.chunks_size, encrypted=args.encrypted,
                           gen_uniq_id=args.gen_uniq_id, only_core=args.only_core, include_suffixes=args.include_suffixes,
                           exclude_suffixes=args.exclude_suffixes)
-    task.wait()
+    task.wait(timeout=None)
```
droideck, line 244: I was thinking, maybe, we should log at least something here if the user SIGINT at that point.
(we can overwrite signal_handler here, I think)

droideck (approved):
Besides this small verbosity comment, that I am not 100% sure of - looks good to me!

src/lib389/lib389/cli_conf/backend.py
Firstyear: When you say sigint you mean they want to cancel the task? Or you want the sigint to show that the task is still waiting?

droideck: > When you say sigint you mean they want to cancel the task? Or you want the sigint to show that the task is still waiting?

I think just 'log that the task is waiting' will be enough.
But we can go even further and ask user a question like: 'Do you want to cancel task? (Yes - Cancel, No - Keep it running in the background)'.
I am good with both options :)

Firstyear: It's a cool idea. I think maybe a more intutive way is on the first sigint (ctrl-c) we print the current status of the task and if they press ctrl-c again within 1 second then we cancel?

mreynolds389: > It's a cool idea. I think maybe a more intutive way is on the first sigint (ctrl-c) we print the current status of the task and if they press ctrl-c again within 1 second then we cancel?

I really like this idea, although you can only get so much information from the task entry.  Regardless think this would be a great enhancement, that being said perhaps it should be done in a different ticket with a wider scope.

Firstyear: Yep, that sounds reasonable. In that case it's okay to merge this as is and I'll create a new ticket for the interupt handling?

droideck: Agree. Let's do this in a separate ticket. For this one, maybe you can log at least something but it's up to you.
You have my ack.

Issue #4366, opened 2020-10-08
lib389 entry inactivity checks are wrong


mreynolds389:
**Issue Description**

The lib389 function that that converts a generalized time to epoch seconds was flawed, and it causes the account status inactivity checks to be incorrect.  So "dsidm INST account status DN" would not always report the correct status

**Package Version and Platform:**
 - Platform: [e.g. Fedora]
 - Package and version: 389-ds-base-1.4.4.2

**Steps to Reproduce**

1. Enable account policy plugin, setting lastLoginAttr and inactivity limit
2. Run dsidm on an existing user:  dsidm INST account status DN
3. The "Time Until Inactive" is wrong.  This can cause the tool to report the wrong status of an entry in regards to the inactivity limit.

mreynolds389:
Reproducible in existing test case: plugins/accpol_test.py

mreynolds389:
7cb9a635a..aaaaaf508  master     -> origin/master
dc8fc3152..ca35b78c6  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
11d224059..988f30d5a  389-ds-base-1.4.2 -> 389-ds-base-1.4.2

PR #4367, opened 2020-10-08
Issue 4366 - lib389 - Fix account status inactivity checks

mreynolds389:
Bug Description:  

When we converted the entries lastLoginAttr to epoch seconds the function was not converting it correctly, and the value was off by quite a bit.  This caused the CLI tools to potentially report the wrong status of the entry.

Fix Description:

First the times from the entry are gmtime, not local.  So instead of grabbing the current local time, we need to grab the current gmtime.  Second, the function that converts a generalized time to epoch seconds is not working.  So that was reworked to generate the correct epoch value.

relates: https://github.com/389ds/389-ds-base/issues/4366


Firstyear:
Ack :)

PR #4374, opened 2020-10-12
Issue 4372 BUG - Chaining DB did not validate bind mech parameters

### LINKED ISSUE #4372: dsconf chaining --bind-mech may create invalid bind mech
```
 --bind-mech BIND_MECH
                        Sets the authentication method to use to authenticate to the remote server: <leave empty for LDAP/LDAPS>, EXTERNAL, DIGEST-MD5, or GSSAPI
```

However bind-mech is "required" and when set to "", this creates an invalid configuration. This should be set to "SIMPLE".

Firstyear:
Bug Description: Chaining DB did not validate the content of bind mech.
When combined with an ambiguous help string, this caused users to set
blank/empty strings into the chaining db config, that would not auth
correctly to the tarcet.

Fix Description: The chaining DB should strictly enforce the incoming
values that are set. The help in dsconf should be explicit about what
values are supported and how they are set.

fixes: #4372

Review by: ???

### DIFF
--- a/ldap/servers/plugins/chainingdb/cb_instance.c
+++ b/ldap/servers/plugins/chainingdb/cb_instance.c
@@ -1459,6 +1459,19 @@ cb_instance_bindmech_set(void *arg, void *value, char *errorbuf, int phase, int 
         }
     }
 
+    if (rc == LDAP_SUCCESS) {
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
LGTM

Issue #4373, opened 2020-10-12
dsconf chaining may create invalid config when parent suffix does not exist


Firstyear:
It is possible that someone wishes to create a chaining config, where no parent suffix already exists. Consider chaining say for dc=example,dc=com and the existing DB is for dc=example,dc=org. In this case, dsconf creates the mapping tree entry with a parent suffix of dc=com, even though no mapping tree entry for dc=com exists.

When the chaining entry is created, the mapping tree must be examined to determine if a parent *does* exist and what it is along the DN path.

Firstyear:
@mreynolds389 Checking with you that it's okay for this to be backported to 1.4.3 and 1.4.4.

mreynolds389:
Yes it can be backported!

Firstyear:
I reran the mapping tree and basic tests on both branches to be 100% sure :) 
```
   9bde84277..66f1a75eb  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
   b79419a44..43bcb5616  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

Thanks again @mreynolds389!

mreynolds389:
Reopening.  This causes a malloc error in the server right after instance creation:

```
	389-Directory/2.0.1.20201209gitd644ebaa9 B2020.344.2303
	localhost.localdomain:389 (/etc/dirsrv/slapd-localhost)

[10/Dec/2020:11:56:29.862159667 -0500] - ERR - memory allocator - cannot calloc 0 elements;
trying to allocate 0 or a negative number of elements is not portable and
gives different results on different platforms.
[10/Dec/2020:11:56:29.903738247 -0500] - INFO - main - 389-Directory/2.0.1.20201209gitd644ebaa9 B2020.344.2303 starting up
[10/Dec/2020:11:56:29.906456781 -0500] - INFO - main - Setting the maximum file descriptor limit to: 524288
[10/Dec/2020:11:56:29.910923818 -0500] - ERR - allow_operation - Component identity is NULL
[10/Dec/2020:11:56:30.649917222 -0500] - INFO - PBKDF2_SHA256 - Based on CPU performance, chose 2048 rounds
```

Firstyear:
@mreynolds389 Interesting, good catch. I don't know how I missed that ... anyway, I'll fix this today.

PR #4375, opened 2020-10-13
Issue 4373 - BUG - Mapping Tree nodes can be created that are invalid

Firstyear:
Bug Description: The mapping tree is built and arranged based on
    the content of the nsslapd-parent-suffix attribute. However, it is
    possible that this value is invalid pointing at a non-existant
    suffix, or that it could be pointing at a suffix that is invalid
    in the suffix hierarchy that mapping trees expect.

    https://www.port389.org/docs/389ds/design/mapping_tree_assembly.html

    Fix Description: Rather than build the mapping tree by arranging
    nodes through the nsslapd-parent-suffix value, we should sort and build
    them through the known and defined suffix values in cn (which we already)
    rely upon to be correct. This allows stable ordering and avoids potential
    user and developer errors.

    fixes: #4373

    Review by: ???

### DIFF
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -45,6 +45,16 @@ struct mt_node
     void *mtn_extension;          /* plugins can extend a mapping tree node */
 };
 
+/*
+ * A temporary value used to sort the order of mapping tree nodes
+ * and how they should be built into a suffix.
+ */
+struct mt_suffix_ord
+{
+    Slapi_DN *mtn_subtree;
+    size_t index;
+};
+
 #define BE_LIST_INIT_SIZE 10
 #define BE_LIST_INCREMENT 10
 
@@ -106,7 +116,6 @@ static int extension_type = -1; /* type returned from the factory */
 
 /* Note: This DN is no need to be normalized. */
 #define MAPPING_TREE_BASE_DN "cn=mapping tree,cn=config"
-#define MAPPING_TREE_PARENT_ATTRIBUTE "nsslapd-parent-suffix"
 
 void mtn_wlock(void);
 void mtn_lock(void);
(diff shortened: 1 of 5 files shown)


Firstyear:
This is not ready for merge yet, the test cases are provided to start a discussion about how we should resolve the mapping tree construction issues that exist.

Firstyear:
Relates #4373

Firstyear:
@mreynolds389 I don't think this is ready for merge yet (there is a single failing case I need to understand but I think it's an environment issue as it fails on git master too). But since this relates closely to your current work we probably want your MT changes to be merged first and I'll rebase onto that :)

Firstyear:
Okay, this is good to go, please check @mreynolds389 and @tbordaz, (I will wait until #4378 is merged to rebase and retest of course)

Firstyear:
@mreynolds389 This update adds extended testing as discussed on the mailing list, including deeper nesting, broad and many suffixes, different lengths, and some extra ideas I thought of too. They all pass as well :)

Firstyear:
@mreynolds389, @progier389, @tbordaz, @droideck review reminder - we have a customer who has been running this patch for a month with no issues. Would be good to have in 1.4.3 / 1.4.4 so they can stop running hotfix builds.

ldap/servers/slapd/mapping_tree.c
```diff
@@ -1355,17 +1317,31 @@ mapping_tree_entry_add_callback(Slapi_PBlock *pb __attribute__((unused)),
...
 
-    /* WARNING
-     * for adds we don't need to grab the mapping tree global lock,
-     * because the add operation in the tree is atomic because
-     * only one pointer is updated in the tree.
-     * Should the mapping tree stucture change, this  would have to
-     * be checked again
+    /*
+     * We absolutely do need the lock here because we are *changing things*
+     * and pointer updates are not magically atomic ...
```
progier389: A very minor point: IMHO this comment is a bit cryptic and should be removed.
         It is a response to the erroneous comment you removed rather than a clear message that stands by itself.
         It will be puzzling when someone will read it later on without the removed message context.
         IMHO: You should just remove that comment:
                There is no need to justify that you lock the mapping tree when modifying its data.

Apart this minor point, LGTM

progier389 (approved):

ldap/servers/slapd/mapping_tree.c
Firstyear: That's a good point about the comment. I'll reword it to be clearer, and as much as I trust your review, I think that @mreynolds389 may want to have a look at this issue too for his ack as well 👍 

Thanks for your time!

mreynolds389:
LGTM

ldap/servers/slapd/mapping_tree.c
```diff
@@ -889,99 +851,118 @@ mtn_create_extension(mapping_tree_node *node)
...
-    for (x = 0; entries[x] != NULL; x++) {
+    /*
+     * Sort the entries by their suffix. We do this by making a temp
+     * array, and putting in a copy of the suffix and it's index to the
+     * entries. We then use this sorted suffix array to then offset through
+     * all the entries.
+     */
+    int ent_count = 0;
+    slapi_pblock_get(pb, SLAPI_NENTRIES, (void *)&ent_count);
+    struct mt_suffix_ord *ordered_suffixes = (struct mt_suffix_ord *)slapi_ch_calloc(ent_count, sizeof(struct mt_suffix_ord));
```
mreynolds389, line 921: This might be the calloc of size 0 we see after creating an instance...

PR #4378, opened 2020-10-14
Issue 2526 - suffix management in backends incorrect

mreynolds389:
Description:  

Previously the server used to support multiple suffixes per backend and the server had to maintain and check a be list of suffixes.  However, this is no longer supported, so all of this code can be cleaned up to support a single suffix per backend.

Also added a check that when creating a mapping tree entry, that the backend entry must already exist and match the suffix.

Relates: https://github.com/389ds/389-ds-base/issues/2526

### DIFF
--- a/dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
+++ b/dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
@@ -1,0 +1,65 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import ldap
+import logging
+import pytest
+import os
+from lib389._constants import *
+from lib389.topologies import topology_st as topo
+from lib389.mappingTree import MappingTrees
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+
+def test_invalid_mt(topo):
+    """Test that you can not add a new suffix/mapping tree
+    that does not already have the backend entry created.
+
+    :id: caabd407-f541-4695-b13f-8f92af1112a0
+    :setup: Standalone Instance
+    :steps:
+        1. Create a new suffix that specifies an existing backend which has a
+           different suffix.
+        2. Create a suffix that has no backend entry at all.
+    :expectedresults:
+        1. Should fail with UNWILLING_TO_PERFORM
+        1. Should fail with UNWILLING_TO_PERFORM
+    """
+
+    bad_suffix = 'dc=does,dc=not,dc=exist'
+    mts = MappingTrees(topo.standalone)
+    
+    properties = {
+        'cn': bad_suffix,
+        'nsslapd-state': 'backend',
+        'nsslapd-backend': 'userroot',
+    }
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        mts.create(properties=properties)
+
+    properties = {
+        'cn': bad_suffix,
+        'nsslapd-state': 'backend',
+        'nsslapd-backend': 'notCreatedRoot',
+    }
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        mts.create(properties=properties)
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/ldap/servers/plugins/chainingdb/cb_instance.c
+++ b/ldap/servers/plugins/chainingdb/cb_instance.c
@@ -1662,9 +1662,9 @@ cb_instance_search_config_callback(Slapi_PBlock *pb __attribute__((unused)),
     char buf[CB_BUFSIZE];
     struct berval val;
     struct berval *vals[2];
-    int i = 0;
     cb_backend_instance *inst = (cb_backend_instance *)arg;
     cb_instance_config_info *config;
+    const Slapi_DN *aSuffix;
 
     vals[0] = &val;
     vals[1] = NULL;
@@ -1673,25 +1673,17 @@ cb_instance_search_config_callback(Slapi_PBlock *pb __attribute__((unused)),
 
     slapi_rwlock_rdlock(inst->rwl_config_lock);
 
-    {
-        const Slapi_DN *aSuffix;
-        i = 0;
-        if (inst->inst_be) {
-            while ((aSuffix = slapi_be_getsuffix(inst->inst_be, i))) {
-                val.bv_val = (char *)slapi_sdn_get_dn(aSuffix);
-                val.bv_len = strlen(val.bv_val);
-                if (val.bv_len) {
-                    if (i == 0)
-                        slapi_entry_attr_replace(e, CB_CONFIG_SUFFIX, (struct berval **)vals);
-                    else
-                        slapi_entry_attr_merge(e, CB_CONFIG_SUFFIX, (struct berval **)vals);
-                }
-                i++;
+    if (inst->inst_be) {
+        if ((aSuffix = slapi_be_getsuffix(inst->inst_be))) {
+            val.bv_val = (char *)slapi_sdn_get_dn(aSuffix);
+            val.bv_len = strlen(val.bv_val);
+            if (val.bv_len) {
+                slapi_entry_attr_replace(e, CB_CONFIG_SUFFIX, (struct berval **)vals);
             }
         }
     }
 
-    for (i = 0; inst->chaining_components && inst->chaining_components[i]; i++) {
+    for (size_t i = 0; inst->chaining_components && inst->chaining_components[i]; i++) {
         val.bv_val = inst->chaining_components[i];
         val.bv_len = strlen(val.bv_val);
         if (val.bv_len) {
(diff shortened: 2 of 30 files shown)


dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
```diff
@@ -0,0 +1,65 @@
...
+#
+import ldap
+import logging
+import pytest
+import os
+from lib389._constants import *
+from lib389.topologies import topology_st as topo
+from lib389.mappingTree import MappingTrees
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
```
Firstyear, line 17: I don't think this boiler plate is needed anymore? Isn't this setup in topologies?

ldap/servers/plugins/chainingdb/cb_instance.c
```diff
@@ -1673,25 +1673,17 @@ cb_instance_search_config_callback(Slapi_PBlock *pb __attribute__((unused)),
...
+            val.bv_val = (char *)slapi_sdn_get_dn(aSuffix);
+            val.bv_len = strlen(val.bv_val);
+            if (val.bv_len) {
+                slapi_entry_attr_replace(e, CB_CONFIG_SUFFIX, (struct berval **)vals);
             }
         }
     }
 
-    for (i = 0; inst->chaining_components && inst->chaining_components[i]; i++) {
+    for (size_t i = 0; inst->chaining_components && inst->chaining_components[i]; i++) {
```
Firstyear, line 1686: 😍

ldap/servers/slapd/mapping_tree.c
```diff
@@ -569,7 +569,7 @@ free_mapping_tree_node_arrays(backend ***be_list, char ***be_names, int **be_sta
  * tree node (guaranteed to be non-NULL).
  */
 static int
-mapping_tree_entry_add(Slapi_Entry *entry, mapping_tree_node **newnodep)
+mapping_tree_entry_add(Slapi_Entry *entry, mapping_tree_node **newnodep, PRBool check_be)
```
Firstyear, line 572: Is there a situation where we would want check_be to be false?

Firstyear (commented):
Generally I really like this change, I was looking at the mapping tree myself and was planning to add validation that the BE existed for the suffix too, so I'm glad you beat me to it! I think it's a good clean up. I'll let you respond to the comments but it's pretty close to an ack from me :)

ldap/servers/slapd/mapping_tree.c
```diff
@@ -605,6 +605,37 @@ mapping_tree_entry_add(Slapi_Entry *entry, mapping_tree_node **newnodep)
         return lderr;
     }
 
+    /* Verify there is a matching backend for this suffix */
```
Firstyear, line 608: Actually, I forgot a point here. When we are doing the check of the BE, we should either assert/validate that the suffix from the MT == BE suffix *or*, we should actually just ignore the suffix from the MT config entry, and we should rely on the suffix value from the BE instead. What do you think?

dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
mreynolds389: Sure I can remove that, it's being generated by create_test.py...

ldap/servers/slapd/mapping_tree.c
mreynolds389: Yes, when we start the server we load the entries from dse.ldif out of order, so we skip the check in this case - otherwise the server fails to start :-)

ldap/servers/slapd/mapping_tree.c
mreynolds389: > Actually, I forgot a point here. When we are doing the check of the BE, we should either assert/validate that the suffix from the MT == BE suffix 

Won't that accomplish the exact same thing this code is already doing?

> or, we should actually just ignore the suffix from the MT config entry, and we should rely on the suffix value from the BE instead. What do you think?  

Why would we ignore the invalid configuration in the MT entry?  Sorry I am probably missing your point :-/

ldap/servers/slapd/mapping_tree.c
Firstyear: As in we load the MT entries out of order? Or we load the MT before the BE's?

dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
Firstyear: we should probably clean up create_test.py then too :)

ldap/servers/slapd/mapping_tree.c
mreynolds389: Correct, the mapping tree entry is processed before the backend entry.  So technically you can still screw things up if you edit the dse.ldif manually.  But we only support doing these updates over LDAP so I'm not that worried about it to be honest.

ldap/servers/slapd/mapping_tree.c
Firstyear: When we create the MT node, we use the subtree (suffix) from get_subtree_from_entry which uses the value of cn in the mapping tree. But we never check that cn == backend suffix.

So with your change, we now assert that a backend for MT nsslapd-backend also exists, but we never validate that the backend suffix == mt suffix.

IMO we should either check these are the same value *or* even better, we only read the suffix from the BE because the BE knows what suffix it provides, so we remove the possibility of inconsistency from the MT entries.

dirsrvtests/tests/suites/mapping_tree/acceptance_test.py
mreynolds389: Sorry, that was implied by my previous comment :-p

Firstyear: Ahh great, lets do that then (probably should be a seperate PR if we need to backport this change ...)

ldap/servers/slapd/mapping_tree.c
Firstyear: Really? For the mapping tree node to exist it has to have the BE pointer that it's going to serve, so the BE has to be loaded first. That's what get_backend_from_attr relies on in the entry_add call.

ldap/servers/slapd/mapping_tree.c
mreynolds389: > When we create the MT node, we use the subtree (suffix) from get_subtree_from_entry which uses the value of cn in the mapping tree. But we never check that cn == backend suffix.
> 
> So with your change, we now assert that a backend for MT nsslapd-backend also exists, but we never validate that the backend suffix == mt suffix.

Yes it does:

    slapi_sdn_compare(subtree, be->be_suffix) == 0)

The suffixes must match.  Sorry, I'm still missing your point here :-(

ldap/servers/slapd/mapping_tree.c
mreynolds389: If it didn't happen I would not have had to add the boolean check :-)

ldap/servers/slapd/mapping_tree.c
Firstyear: With the MT are there types where we don't have an associated backend? Referrals maybe? We probably should just check in the case where the MT has a backend that suffix's are equal I think.

ldap/servers/slapd/mapping_tree.c
Firstyear: Oh I believe you but I'm confused. We rely on the backend list being populated at start up, else we'll get the error "The nsslapd-backend attribute has no value for the mapping tree", so the backend list *has* to be populated first, else no backends could possibly be loaded in the mapping tree without a subsequent second pass over the configs. So I'm curious now to see how the situation exists where the PR_FALSE flag is needed here because in my mind it just doesn't compute, and probably indicates some other issue ....

ldap/servers/slapd/mapping_tree.c
Firstyear: Ohhhh okay, I missed that line. IN that case, ignore me, this is all good then :D

ldap/servers/slapd/mapping_tree.c
Firstyear: Oh I just found the comment in get_backends_from_attr that says "we'll fix it at runtime". Yeesh. I wonder if there is a cleaner or better way to handle this situation then, because it would be good to have these checks able to always run to assert the relevant backend does exist and matches.

mreynolds389: So in the startup case, we indeed init the mapping trees before starting the backend plugin.  The default backends (config, schema, rootdse) are inited, but not the ldbm backends:

main() --> setup_internal_backends() -->  mapping_tree_init() --> mapping_tree_node_get_children() --> mapping_tree_entry_add()

Firstyear: All good, thanks for clearing that up. In that case Ack from me :)

Firstyear:
@mreynolds389 Reminder of ack :)

mreynolds389:
> @mreynolds389 Reminder of ack :)

I know, sorry, I wanted some old TET chaining tests performed on it before pushing.  QE is going to do this, but I don't know the progress on this task.  I'll check in with them tomorrow...

Firstyear:
Ahh didn't know that, I thought maybe you missed my notification since it was buried in the comment chain. Let me know how the tests go, it'll be good to have them run through with the other mapping tree PR I have in later too :)

mreynolds389:
> Ahh didn't know that, I thought maybe you missed my notification since it was buried in the comment chain. Let me know how the tests go,

Yeah sorry I should have said something in the ticket, but I guess I was hoping the tests would have been done by now.  Next time I won't leave out such details ;)

abbra:
This change to `slapi_be_getsuffix()` is breaking ABI and changing API for external plugins, in FreeIPA and slapi-nis.
Is it possible to do it without incurring a lot of trouble to existing plugins? They have to work against old versions of 389-ds too.

abbra:
Please see https://pagure.io/freeipa/issue/8556

mreynolds389:
> This change to `slapi_be_getsuffix()` is breaking ABI and changing API for external plugins, in FreeIPA and slapi-nis.
> Is it possible to do it without incurring a lot of trouble to existing plugins? They have to work against old versions of 389-ds too.

Fixed!

abbra:
Thank you, Mark! It would be great to improve the API but we really stuck with the function signature at this point, so ignoring the second argument is the right way to do.

Issue #4391, opened 2020-10-21
DSE config modify does not call be_postop 


tbordaz:
**Issue Description**
During a DSE modify, be_preop callback are called. But be_postop callback are called at the condition dse_call_callback is different that SLAPI_DSE_CALLBACK_DO_NOT_APPLY.

This should systematically call be_postop if be_preop were called. In addition  postop_modify_config_dse returning an invalid rc, systematically prevents DSE modify to call be_postop

**Package Version and Platform:**
 - Platform: all
 - Package and version: since 2010

**Steps to Reproduce**
Under debugger, update any config attribute (ie. nsslapd-auditlog-logging-enabled) and check that be_postop are not called

**Expected results**
If be_preop are called, be_postop should be called

Firstyear:
Yeah, I agree, we should call the postops too to balance this.

tbordaz:
5b0cbddc4..bc92c17bc    master
6e1b7b9d2..f1285d78a    1.4.4
d2dcf85c3..5ea53ebd7    1.4.3

mreynolds389:
@tbordaz can this be closed?

tbordaz:
opss.. forgot to close this one

PR #4394, opened 2020-10-22
Issue 4391 - DSE config modify does not call be_postop

tbordaz:
Bug description:
	During a DSE modify, be_preop callback are called. But be_postop callback are called at the condition
	dse_call_callback is different that SLAPI_DSE_CALLBACK_DO_NOT_APPLY.

	This should systematically call be_postop if be_preop were called.
	In addition postop_modify_config_dse returning an invalid rc, systematically prevents DSE modify to call be_postop

Fix description:
        The required bug fix is that dse_callback need to return SLAPI_DSE_CALLBACK* not ldap rc.
	Also in case of vlv config (SLAPI_DSE_CALLBACK_DO_NOT_APPLY) if preop were called
        it requires to call the postop.

	In dse_modify, rc is used for dse_call_callback() (returns SLAPI_DSE_CALLBACK*)
        but also for plugin_call_plugin (returns SLAPI_PLUGIN_*). Those rc are not compatible
	and although the code works to help maintenance use 'plugin_rc' instead of 'rc'.

relates: https://github.com/389ds/389-ds-base/issues/4391

Reviewed by:

Platforms tested: F31

### DIFF
--- a/ldap/servers/slapd/dse.c
+++ b/ldap/servers/slapd/dse.c
@@ -1866,31 +1866,33 @@ dse_modify(Slapi_PBlock *pb) /* JCM There should only be one exit point from thi
     /* give the dse callbacks the first crack at the modify */
     rc = dse_call_callback(pdse, pb, SLAPI_OPERATION_MODIFY, DSE_FLAG_PREOP, ec, ecc, &returncode, returntext);
     if (SLAPI_DSE_CALLBACK_OK == rc) {
+        int plugin_rc;
+
         /* next, give the be plugins a crack at it */
         slapi_pblock_set(pb, SLAPI_RESULT_CODE, &returncode);
         slapi_pblock_set(pb, SLAPI_MODIFY_EXISTING_ENTRY, ecc);
-        rc = plugin_call_plugins(pb, SLAPI_PLUGIN_BE_PRE_MODIFY_FN);
+        plugin_rc = plugin_call_plugins(pb, SLAPI_PLUGIN_BE_PRE_MODIFY_FN);
         need_be_postop = 1; /* if the be preops were called, have to call the be postops too */
         if (!returncode) {
             slapi_pblock_get(pb, SLAPI_RESULT_CODE, &returncode);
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/dse.c
```diff
@@ -2055,12 +2057,28 @@ dse_modify(Slapi_PBlock *pb) /* JCM There should only be one exit point from thi
...
+            plugin_call_plugins(pb, SLAPI_PLUGIN_BE_POST_MODIFY_FN);
+            if (!returncode) {
+                slapi_pblock_get(pb, SLAPI_RESULT_CODE, &returncode);
+            }
+        }
+    } else {
+        /* It should not happen but just be paranoiac, do not
+         * forget to call the postop if needed
+         */
+        if (need_be_postop) {
```
Firstyear, line 2069: Is there a situation where this is false?

ldap/servers/slapd/dse.c
```diff
@@ -2055,12 +2057,28 @@ dse_modify(Slapi_PBlock *pb) /* JCM There should only be one exit point from thi
...
+            plugin_call_plugins(pb, SLAPI_PLUGIN_BE_TXN_POST_MODIFY_FN);
```
Firstyear, line 2070: Don't these need plugin_rc = plugin_call_plugins?

ldap/servers/slapd/dse.c
tbordaz: Yes there are several "early" 'goto done;' in the function where the dse_modify fails before preop are called. For example if the modified entry does not exist (LDAP_NO_SUCH_OBJECT)

ldap/servers/slapd/dse.c
tbordaz: Right, nice catch

droideck (approved):
The change looks good to me too! (if `plugin_rc = plugin_call_plugins` part William mentioned is fixed)

Issue #4363, opened 2020-10-07
Sync repl: per thread structure was incorrectly initialized


tbordaz:
**Issue Description**
The system to detect nested entry is broken for example, this direct update is assumed to be nested
<pre>
[02/Oct/2020:12:48:42.407064204 +0000] conn=11 op=100 ADD dn="ipk11UniqueID=9a0baa52-04ad-11eb-a0dc-525400ea4c12,cn=keys,cn=sec,cn=dns,dc=ipa,dc=test"
[02/Oct/2020:12:48:42.504063075 +0000] conn=11 op=100 RESULT err=0 tag=105 nentries=0 wtime=0.000369948 optime=0.097011474 etime=0.097377063 csn=5f7721ab000800040000

[02/Oct/2020:12:48:42.427542634 +0000] - DEBUG - content-sync-plugin - sync_update_persist_betxn_pre_op - nested operation targets "ipk11UniqueID=9a0baa52-04ad-11eb-a0dc-525400ea4c12,cn=keys,cn=sec,cn=dns,dc=ipa,dc=test" (0x7fe7734bc800)

</pre>

Also the way pending updates are flushed rely on betxn_preop - betxn_postop balance. Unfortunately some op seems to miss the postop, so operation remain pending

<pre>
[02/Oct/2020:12:48:17.366968904 +0000] conn=28 op=1 MOD dn="cn=config"
[02/Oct/2020:12:48:17.374753804 +0000] conn=28 op=0 RESULT err=0 tag=97 nentries=0 wtime=0.009704399 optime=0.066472741 etime=0.076171658 dn="cn=directory manager"

[02/Oct/2020:12:48:17.392523825 +0000] - DEBUG - content-sync-plugin - sync_update_persist_betxn_pre_op - primary operation targets "cn=config" (0x7fe7734bc800)
#post op is not called

</pre>
**Package Version and Platform:**
 - Platform: since #4243
 - Package and version: [e.g. 389-ds-base-1.4.4.4-20200721git5d41dc5a4.fc32.x86_64]

**Steps to Reproduce**
With freeipa DNSSEC tests (no a systematic failure)

**Expected results**
Sync_repl pending should be more robust

tbordaz:
Related to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1887449

tbordaz:
[...]

PR #4395, opened 2020-10-22
Issue 4363 - Sync repl: per thread structure was incorrectly initialized

tbordaz:
Bug description:
	A per thread structure should be allocated once, either on get/set.
        Currently it is allocated on the primary operation and free when
        the primary operation is completed.

Fix description:
	The per thread structure is now a HEAD structure.
        The HEAD is the where the primary operation is referenced when
        the operation starts and where it is reset when the primary operation ends
	(pushed to the sync_repl thread)

relates: https://github.com/389ds/389-ds-base/issues/4363

Reviewed by:

Platforms tested: F31, F33


mreynolds389 (approved):
LGTM

PR #4399, opened 2020-10-23
Issue 2058: create keep alive entry after on-line initialization

progier389:
Bug description:
Keep alive entry is not created on target master after on line initialization, and its RUVelement stays empty until a direct update is issued on that master 

Fix description:
The patch allows a consumer (configured as a master) to create (if it did not exist before) the consumer's keep alive entry. It creates it at the end of a replication session at a time we are sure the changelog exists and will not be reset. It allows a consumer to have RUVelement with csn in the RUV at the first incoming replication session.

That is basically lkrispen's proposal with an associated pytest testcase 

Reviewed by:

Platforms tested: F32

relates: #[2058](https://github.com/389ds/389-ds-base/issues/2058)

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_test.py
@@ -109,6 +109,30 @@ def _move_ruv(ldif_file):
         for dn, entry in ldif_list:
             ldif_writer.unparse(dn, entry)
 
+def _remove_replication_data(ldif_file):
+    """ Remove the replication data from ldif file:
+        db2lif without -r includes some of the replica data like 
+        - nsUniqueId
+        - keepalive entries
+        This function filters the ldif fil to remove these data
+    """
+
+    with open(ldif_file) as f:
+        parser = ldif.LDIFRecordList(f)
+        parser.parse()
+
+        ldif_list = parser.all_records
+        # Iterate on a copy of the ldif entry list
+        for dn, entry in ldif_list[:]:
+            if dn.startswith('cn=repl keep alive'):
+                ldif_list.remove((dn,entry))
+            else:
+                entry.pop('nsUniqueId')
+    with open(ldif_file, 'w') as f:
+        ldif_writer = ldif.LDIFWriter(f)
+        for dn, entry in ldif_list:
+            ldif_writer.unparse(dn, entry)
+
 
 @pytest.fixture(scope="module")
 def topo_with_sigkill(request):
@@ -922,6 +946,112 @@ def test_moving_entry_make_online_init_fail(topology_m2):
     log.info("m2entry count - %d", len(m2entries))
 
     assert len(m1entries) == len(m2entries)
+
+
+def get_keepalive_entries(instance,replica):
+    # Returns the keep alive entries that exists with the suffix of the server instance
+    try:
+        entries = instance.search_s(replica.get_suffix(), ldap.SCOPE_ONELEVEL,
+                    "(&(objectclass=ldapsubentry)(cn=repl keep alive*))",
+                    ['cn', 'nsUniqueId', 'modifierTimestamp'])
+    except ldap.LDAPError as e:
+        log.fatal('Failed to retrieve keepalive entry (%s) on instance %s: error %s' % (dn, instance, str(e)))
+        assert False
+    # No error, so lets log the keepalive entries
+    if log.isEnabledFor(logging.DEBUG):
+        for ret in entries:
+            log.debug("Found keepalive entry:\n"+str(ret));
+    return entries
+
+def verify_keepalive_entries(topo, expected):
+    #Check that keep alive entries exists (or not exists) for every masters on every masters
+    #Note: The testing method is quite basic: counting that there is one keepalive entry per master.
+    # that is ok for simple test cases like test_online_init_should_create_keepalive_entries but
+    # not for the general case as keep alive associated with no more existing master may exists
+    # (for example after: db2ldif / demote a master / ldif2db / init other masters)
+    # ==> if the function is somehow pushed in lib389, a check better than simply counting the entries
+    # should be done.
+    for masterId in topo.ms:
+        master=topo.ms[masterId]
+        for replica in Replicas(master).list():
+            if (replica.get_role() != ReplicaRole.MASTER):
+               continue
+            replica_info = f'master: {masterId} RID: {replica.get_rid()} suffix: {replica.get_suffix()}'
+            log.debug(f'Checking keepAliveEntries on {replica_info}')
+            keepaliveEntries = get_keepalive_entries(master, replica);
+            expectedCount = len(topo.ms) if expected else 0
+            foundCount = len(keepaliveEntries)
+            if (foundCount == expectedCount):
+                log.debug(f'Found {foundCount} keepalive entries as expected on {replica_info}.')
+            else:
+                log.error(f'{foundCount} Keepalive entries are found '
+                          f'while {expectedCount} were expected on {replica_info}.')
+                assert False
+
+
+def test_online_init_should_create_keepalive_entries(topo_m2):
+    """Check that keep alive entries are created when initializinf a master from another one
+
+    :id: d5940e71-d18a-4b71-aaf7-b9185361fffe
+    :setup: Two masters replication setup
+    :steps:
+        1. Generate ldif without replication data
+        2  Init both masters from that ldif
+        3  Check that keep alive entries does not exists
+        4  Perform on line init of master2 from master1
+        5  Check that keep alive entries exists
+    :expectedresults:
+        1. No error while generating ldif
+        2. No error while importing the ldif file
+        3. No keepalive entrie should exists on any masters
+        4. No error while initializing master2
+        5. All keepalive entries should exist on every masters
+
+    """
+
+    repl = ReplicationManager(DEFAULT_SUFFIX)
+    m1 = topo_m2.ms["master1"]
+    m2 = topo_m2.ms["master2"]
+    # Step 1: Generate ldif without replication data
+    m1.stop()
+    m2.stop()
+    ldif_file = '%s/norepl.ldif' % m1.get_ldif_dir()
+    m1.db2ldif(bename=DEFAULT_BENAME, suffixes=[DEFAULT_SUFFIX],
+               excludeSuffixes=None, repl_data=False,
+               outputfile=ldif_file, encrypt=False)
+    # Remove replication metadata that are still in the ldif
+    _remove_replication_data(ldif_file)
+
+    # Step 2: Init both masters from that ldif
+    m1.ldif2db(DEFAULT_BENAME, None, None, None, ldif_file)
+    m2.ldif2db(DEFAULT_BENAME, None, None, None, ldif_file)
+    m1.start()
+    m2.start()
+
+    """ Replica state is now as if CLI setup has been done using:
+        dsconf master1 replication enable --suffix "${SUFFIX}" --role master
+        dsconf master2 replication enable --suffix "${SUFFIX}" --role master
+        dsconf master1 replication create-manager --name "${REPLICATION_MANAGER_NAME}" --passwd "${REPLICATION_MANAGER_PASSWORD}"
+        dsconf master2 replication create-manager --name "${REPLICATION_MANAGER_NAME}" --passwd "${REPLICATION_MANAGER_PASSWORD}"
+        dsconf master1 repl-agmt create --suffix "${SUFFIX}"
+        dsconf master2 repl-agmt create --suffix "${SUFFIX}"
+    """
+
+    # Step 3: No keepalive entrie should exists on any masters
+    verify_keepalive_entries(topo_m2, False)
+
+    # Step 4: Perform on line init of master2 from master1
+    agmt = Agreements(m1).list()[0]
+    agmt.begin_reinit()
+    (done, error) = agmt.wait_reinit()
+    assert done is True
+    assert error is False
+
+    # Step 5: All keepalive entries should exists on every masters
+    #  Verify the keep alive entry once replication is in sync
+    # (that is the step that fails when bug is not fixed)
+    repl.wait_for_ruv(m2,m1)
+    verify_keepalive_entries(topo_m2, True);
 
 
 if __name__ == '__main__':
--- a/ldap/servers/plugins/replication/repl_extop.c
+++ b/ldap/servers/plugins/replication/repl_extop.c
@@ -1175,6 +1175,10 @@ multimaster_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
                 */
                 if (replica_is_flag_set(r, REPLICA_LOG_CHANGES) && cldb_is_open(r)) {
                     replica_log_ruv_elements(r);
+                    /* now that the changelog is open and started, we can alos cretae the
+                     * keep alive entry without risk that db and cl will not match
+                     */
+                    replica_subentry_check(replica_get_root(r), replica_get_rid(r));
                 }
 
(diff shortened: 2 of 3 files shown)


mreynolds389:
Just a quick note, and not your fault, but we are moving away from adding tests to dirsrvtests/tests/ticket directory.  Please add this test to an "appropriate" test in dirsrvtests/tests/suite/replication/
Thanks!

dirsrvtests/tests/tickets/ticket2058_test.py
```diff
@@ -0,0 +1,131 @@
+import logging
+import pytest
+import os
+import subprocess
+import time
+import ldap
+from lib389._constants import *
```
droideck: Please, try to avoid mass imports. It makes code less readable and less explicit (which is python coding best practices).

dirsrvtests/tests/tickets/ticket2058_test.py
```diff
@@ -0,0 +1,131 @@
...
+from lib389.topologies import topology_i2 as topo
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
```
droideck: You can skip this part. It's in lib389 now :)

dirsrvtests/tests/tickets/ticket2058_test.py
```diff
@@ -0,0 +1,131 @@
...
+  rc = subprocess.run(cmd.split(), capture_output=True)
+  if (rc.returncode != 0):
+      log.error(f"Command {cmd} failed: {rc}")
+      assert False
+  return rc
+
+def check_single_keep_alive(topo,instance,dn,expected):
+    # check that keep alive entry does not exists on master1
+    try:
+        ent = topo.ins[instance].getEntry(dn, ldap.SCOPE_BASE, "(objectclass=ldapsubentry)", ['nsUniqueId', 'modifierTimestamp'])
```
droideck: Please, use direct python-ldap calls if you need something specific.
https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#ldap.LDAPObject.search_s

We plan to deprecate 'Entry' object some time soon. 

P.S. we also can add an interface for getting 'keep alive' entries to lib389 and use it from there.

dirsrvtests/tests/tickets/ticket2058_test.py
```diff
@@ -0,0 +1,131 @@
...
+    check_single_keep_alive(topo, "standalone1", f"cn=repl keep alive 1, {SUFFIX}", expected)
+    check_single_keep_alive(topo, "standalone1", f"cn=repl keep alive 2, {SUFFIX}", expected)
+    check_single_keep_alive(topo, "standalone2", f"cn=repl keep alive 1, {SUFFIX}", expected)
+    check_single_keep_alive(topo, "standalone2", f"cn=repl keep alive 2, {SUFFIX}", expected)
+
+
+def test_ticket2058(topo):
+    """Checks that replication keep alive entry are created after on line initialisations
+       (without the fix, only the entry from master1 is created)
+
```
droideck: Also, add additional field here - `:id: UUID`.
You can get UUID like this: `python3 -c "import uuid; print(uuid.uuid4())"`

dirsrvtests/tests/tickets/ticket2058_test.py
```diff
@@ -0,0 +1,131 @@
...
+        f"--replica-id 2 --bind-dn {REPLICATION_MANAGER_DN} "
+        f"--bind-passwd {REPLICATION_MANAGER_PASSWORD}")
+    run(f"dsconf standalone1 replication create-manager --name {REPLICATION_MANAGER_NAME} --passwd {REPLICATION_MANAGER_PASSWORD}")
+    run(f"dsconf standalone2 replication create-manager --name {REPLICATION_MANAGER_NAME} --passwd {REPLICATION_MANAGER_PASSWORD}")
+    run(f"dsconf standalone1 repl-agmt create --suffix {SUFFIX} "
+        f"--conn-protocol LDAP --host {master2.host} --port {master2.port} "
+        f"--bind-method SIMPLE --bind-dn {REPLICATION_MANAGER_DN} --bind-passwd {REPLICATION_MANAGER_PASSWORD}  agmt")
+    run(f"dsconf standalone2 repl-agmt create --suffix {SUFFIX} "
+        f"--conn-protocol LDAP --host {master1.host} --port {master1.port} "
+        f"--bind-method SIMPLE --bind-dn {REPLICATION_MANAGER_DN} --bind-passwd {REPLICATION_MANAGER_PASSWORD}  agmt")
```
droideck: It's more efficient to use direct lib389 calls. It makes the code easier to debug and modify.
I see that you want to create a topology and init it later.
Check the code from this test case:
https://github.com/389ds/389-ds-base/blob/master/dirsrvtests/tests/suites/replication/regression_test.py#L195
Or the code from the link Mark posted:
https://github.com/389ds/389-ds-base/blob/master/src/lib389/lib389/cli_conf/replication.py

And feel free to ping me or other team members.
lib389 documentation is a bit poor now...

droideck (changes requested):
Otherwise, it's a very good code! Thanks!
And yes, as @mreynolds389  mentioned, please, put the code to `suites`.
Probably, this path will be the best place:
`dirsrvtests/tests/suites/replication/regression_test.py`

dirsrvtests/tests/tickets/ticket2058_test.py
Firstyear: @droideck This is probably from the create test template, I think @mreynolds389 has done a cleanup from it recently....

dirsrvtests/tests/tickets/ticket2058_test.py
Firstyear: Well, if we have a keepalive we want to be able to check, we probably could add a lib389 dsldapobject type for it to make this a bit clearer...

dirsrvtests/tests/tickets/ticket2058_test.py
Firstyear: Yeah, I was going to say, don't you want a topology fixture here? There are some examples in the replication suite ....

ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1175,6 +1175,10 @@ multimaster_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
                 */
                 if (replica_is_flag_set(r, REPLICA_LOG_CHANGES) && cldb_is_open(r)) {
                     replica_log_ruv_elements(r);
+                    /* now that the changelog is open and started, we can alos cretae the
+                     * keep alive entry without risk that db and cl will not match
+                     */
+                    replica_subentry_check(replica_get_root(r), replica_get_rid(r));
```
Firstyear, line 1181: When is this call normally made? What's the problem that this resolves? It would be good to understand what this fix is changing.

dirsrvtests/tests/tickets/ticket2058_test.py
droideck: I think `create_topology` function is built the way so it does init during the creation implicitly.
We can write another topology fixture which doesn't do that but I think the case is very rare... So it's better to setup replication here as Pierre has done (but make it with lib389 tools as I mentioned above).

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: It was also created by create_test template, 
 Anyway fixing mreynolds389's remark about avoiding to create new testcase in ticket directory will take care of that ...

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: I get DeprecationWarning when using search_s  ...
DsLdapObject seems the way to go these days. that said creating a class for keepalive entries seems an overkill,
 as for now we only check their presence (and I am not sure we should really expose them in lib389 API)
My feeling is that we should rather provide a generic way to get a set of entries ...

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: Thanks !
I now understand that these :id:  are random uuid.  Ok I will gerate them using  uuid -m -v 4

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: Finally I found an easier method: using 2 master topology then reinitialize the replicas from an ldif without any replication metadata.  I already tried this method without success, because generating the ldif is not so easy (as exporting ldif without the replication data still export them) but seeing one of the testcase in regression_test have shown me how to easily filter the ldif file. So exporting the db then filter ldif to remove nsUniqueId and keepalive entries bring the replica state
 to the same state than the CLI (i.e: replication role and agreements are enabled but replication is stalled because master
  has different generationIds)

ldap/servers/plugins/replication/repl_extop.c
progier389: I will try to clarify what lkrispenz and tbordaz explained in the issue:
  When the keep alive entry feature was originally added, tbordaz added a call to replica_subentry_check
    in replica_enable_replication to insure that keep alive entry is already there.
 (Another place it is called is on supplier side when initializing a replica agremeent)

In ticket [48445](https://pagure.io/389-ds-base/issue/48445) the creation of a keep alive entry after online init was 
 removed because of race condition with the changelog handling but ludwig
And some time later this issue was created because keep alive entry is not created on the master (that are init by another master) and lkrispenz proposed to add it when ending the online init session once changelog is fully started

If your question is what keep alive entries are for and what is their impact, then here is the answer:
  As I understand it, these entries exist only to trigger a change that get replicated over the topology:
  Their main purpose is to generate records in the changelog and they are updated from time to time by fractional
  replication  to insure that at least a change must be replicated by FR after a great number of not replicated changes are
  found in the changelog. The interest is that the fractional RUV get then updated so less changes need to be walked
    in the changelog when searching for the first change to send)

So creating a keep alive entry has little impact as far as the replication is fully started (including the changelog) 
    (As tbordaz mentioned, the main difference is that the master will appear in the RUV ))

IMHO the main interests of this fix are:
   - it is a good exercise for me to discover the 389 the open development process
   - more importantly, it will avoid having people getting puzzled by the fact that a master does not appear in the RUV 
     and spending time to understand why (I suspect that it what happened to lkrispenz and why he opened the issue -;) )

dirsrvtests/tests/tickets/ticket2058_test.py
Firstyear: Mark recently did a change to the template create that removes this too :)

dirsrvtests/tests/tickets/ticket2058_test.py
Firstyear: Yeah, it's been asked about before, to have a way to get a single entry by dn, or to do an objectClass=* style search. Maybe I should let go of my ideals and create it :)

ldap/servers/plugins/replication/repl_extop.c
Firstyear: That makes sense, and sounds good to me. I wonder if this (really good) explanation could be in the code comments somewhere? That way it's preserved for future-us to remember. Thank you!

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: Getting a single object sound fine but beware to the filter:
   subentries are hidden if (objectclass=subentry) is not an explicit part of the filter
   so the filter should rather be:   (|(objectclass=*)(objectclass=subentry)) 

[progier@linux-3 slapd]$ ldapsearch -LLL -Y EXTERNAL -H ldapi://$S1 -b "cn=repl keep alive 1,dc=example,dc=com"  "(objectclass=*)" dn 2>/dev/null
[progier@linux-3 slapd]$ ldapsearch -LLL -Y EXTERNAL -H ldapi://$S1 -b "cn=repl keep alive 1,dc=example,dc=com"  "(|(objectclass=*)(objectclass=ldapsubentry))" dn 2>/dev/null
dn: cn=repl keep alive 1,dc=example,dc=com

[progier@linux-3 slapd]$

ldap/servers/plugins/replication/repl_extop.c
progier389: Hum, That seems a good idea: 
 the part about explaining the keep alive entries role could be set in comment just before the definition of the
  replica_subentry_* functions that creates, checks and updates the keep alive entries.

dirsrvtests/tests/tickets/ticket2058_test.py
progier389: Sorry, In the previous message I meant objectclass=ldapsubentry rather than objectclass=subentry !

Firstyear:
Looks great, thanks for adding that comment explanation. Ack from me, :)

dirsrvtests/tests/suites/replication/regression_test.py
```diff
@@ -924,6 +948,109 @@ def test_moving_entry_make_online_init_fail(topology_m2):
...
+    m2 = topo_m2.ms["master2"]
+    # Step 1: Generate ldif without replication data
+    m1.stop()
+    m2.stop()
+    ldif_file = '%s/norepl.ldif' % m1.get_ldif_dir()
+    m1.db2ldif(bename=DEFAULT_BENAME, suffixes=[DEFAULT_SUFFIX],
+               excludeSuffixes=None, repl_data=False,
+               outputfile=ldif_file, encrypt=False)
+    # Remove replication metadata that are still in the ldif
+    _remove_replication_data(ldif_file)
```
droideck, line 1023: One out of three runs has failed here because it didn't find the `ldif_file`.
Maybe, it's worth to set a small `time.wait()` after `db2ldif` task.

droideck (approved):
Besides the small issue, you have my ack!
Really great job with the Python and lib389!

dirsrvtests/tests/suites/replication/regression_test.py
progier389: Strange:   As far as I remember the off line export is supposed to be synchronous.
  (That is not an ldap task that need monitoring) but it is a ns-slapd instance in db2ldif mode)  
So I am not sure that adding a delay will help.
I am running the test in loop to try to reproduce the failure

progier389: So far I did not yet reproduced the issue but I hit another one: step 5 is sometime failing because the keep alive entry was not yet replicated +=> Must wait that master1 is in sync with master2 before verifying the entries.
Restarted the tests after adding a wait_for_ruv(m2,m1)

progier389: Test run in loop without error since 2 hours. I also double checked and indeed off-line export is synchronous.
So I do not know what happened in your environment (At this point: either the file was not created or it has been removed, both scenario let me think that something unusual happened in your environment (maybe the instance was been cleared by another test or some cleanup task in background ... ))

droideck: > Test run in loop without error since 2 hours. I also double checked and indeed off-line export is synchronous.
> So I do not know what happened in your environment (At this point: either the file was not created or it has been removed, both scenario let me think that something unusual happened in your environment (maybe the instance was been cleared by another test or some cleanup task in background ... ))

Looks like this.
I test on an isolated environment (freshly created libvirt machine) but still, something could happen there.

If the test will fail on our CI, we will see. The failure was purely test-related (and, probably, env-related) so no worries :)

You have my ack! And I think we can merge.

PR #4402, opened 2020-10-26
Issue 4262 - Remove legacy tools subpackage (final cleanup)

mreynolds389:
Description:  Found a few more cleanup issues with removing perl.  Also hardened the topologies teardown (it's ok if the bind fails)

Relates: https://github.com/389ds/389-ds-base/issues/4262


droideck (approved):
LGTM!

vashirov:
Sorry, I'm a bit late for review. There is an incorrectly set var `enable_perl` in `defaults.inf`:
```
enable_perl = @enable_perl@
```
Also, some tests (for example, basic_test.py) fail to teardown the instance at the very end. I'm not sure what's going on there.
```
(Pdb) topology.standalone.state
5    
(Pdb) topology.standalone.whoami_s()
*** ldap.SERVER_DOWN: {'desc': "Can't contact LDAP server"}
(Pdb) topology.standalone.simple_bind_s(DN_DM, PASSWORD)
*** ldap.SERVER_DOWN: {'desc': "Can't contact LDAP server"}
```
Even though it reports that the server is running (5 == DIRSRV_STATE_ONLINE), it fails to connect to it.

Instead of rebinding as DM we can stop the instance in the finalizer as an instance doesn't need to be online to be removed.

mreynolds389:
> Sorry, I'm a bit late for review. There is an incorrectly set var `enable_perl` in `defaults.inf`:
> 
> ```
> enable_perl = @enable_perl@

Errrrrr, damn, ok, I'll fix this.  I ran the basic test suite, although I didn't wait for it to finish all the tests.

> ```
> 
> Also, some tests (for example, basic_test.py) fail to teardown the instance at the very end. I'm not sure what's going on there.
> 
> ```
> (Pdb) topology.standalone.state
> 5    
> (Pdb) topology.standalone.whoami_s()
> *** ldap.SERVER_DOWN: {'desc': "Can't contact LDAP server"}
> (Pdb) topology.standalone.simple_bind_s(DN_DM, PASSWORD)
> *** ldap.SERVER_DOWN: {'desc': "Can't contact LDAP server"}
> ```
Hmm this patch should ignore those errors, I added try's and except's around the bind, but yes switching it to stopping the server sounds much more robust \.

PR #4405, opened 2020-10-27
Issue 4262 - more perl removal cleanup

mreynolds389:
Description:  Removed the remaining "enabled_perl" code.

Relates: https://github.com/389ds/389-ds-base/issues/4262

Passes "basic" test suite.


mreynolds389:
All the perl stuff should be fixed now.  As for the new errors in changelog, it is because of the backend/mapping tree checks I added from a different issue (#2526)

vashirov (approved):
I applied this on top of master with #2526 included and ran tests under `suites`. All teardown errors are gone now. There are other failures, but they are not related to perl removal.

PR #4406, opened 2020-10-27
Issue 2526 - retrocl backend created out of order

mreynolds389:
Bug Description:  

A recent change verified that you do not create a mapping tree entry before the backend entry was created.  The server created the retrocl backend in the opposite order which broke the retrocl.

Fix Description:  

Create the retrocl backend entry before creating the mapping tree entry.

Relates: https://github.com/389ds/389-ds-base/issues/2526

Reviewed by: ?


vashirov (approved):
`dirsrvtests/tests/suites/replication/changelog_test.py` is passing now, ack from me.

PR #4408, opened 2020-10-29
Issue 4403 RFE - OpenLDAP pw hash migration tests

### LINKED ISSUE #4403: OpenLDAP migration - password hash support/tests
Add tests for various types of openldap password hash types, and implement any needed support code to make them function. This will help make openldap to 389 migrations smoother.

Firstyear:
Bug Description: As we want to support openldap to 389 password migration,
we should check if we allow accounts to continue to bind. This involves
testing different openldap authentication schemes to determine if they
work.

Fix Description: Add tests for different password and contrib password
types that are supported in openldap.

fixes: #4403

Review by: ???

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1205,6 +1205,14 @@ libslapd_la_SOURCES = ldap/servers/slapd/add.c \
 
 libslapd_la_CPPFLAGS = $(AM_CPPFLAGS) $(DSPLUGIN_CPPFLAGS) $(SASL_CFLAGS) @db_inc@ $(KERBEROS_CFLAGS) $(PCRE_CFLAGS) $(SDS_CPPFLAGS) $(SVRCORE_INCLUDES)
 libslapd_la_LIBADD = $(LDAPSDK_LINK) $(SASL_LINK) $(NSS_LINK) $(NSPR_LINK) $(KERBEROS_LIBS) $(PCRE_LIBS) $(THREADLIB) $(SYSTEMD_LIBS) libsds.la libsvrcore.la
+# If asan is enabled, it creates special libcrypt interceptors. However, they are
+# detected by the first load of libasan at runtime, and what is in the linked lib
+# so we need libcrypt to be present as soon as libasan is loaded for the interceptors
+# to function. Since ns-slapd links libslapd, this is pulled at startup, which allows
+# pwdstorage to be asan checked with libcrypt.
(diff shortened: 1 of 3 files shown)


Firstyear:
@mreynolds389 This also resolves a linking issue with asan that causes crypt to have a false negative in tests, but only under asan. :)

src/lib389/lib389/__init__.py
```diff
@@ -1090,6 +1090,7 @@ def start(self, timeout=120, post_open=True):
                 self.log.info("INFO: ASAN options will be copied from your environment")
                 env['ASAN_SYMBOLIZER_PATH'] = "/usr/bin/llvm-symbolizer"
                 env['ASAN_OPTIONS'] = "symbolize=1 detect_deadlocks=1 log_path=%s/ns-slapd-%s.asan" % (self.ds_paths.run_dir, self.serverid)
+                env['LD_BIND_NOW'] = "yes"
```
progier389: Have you tried to use  LD_PRELOAD=libcrypt.so instead of LD_BIND_NOW=yes ?

My concern is that LD_BIND_NOW has a noticeable impact on ns-slapd start time 
  that will be painful when running all the pytests to check for non regression.
 ( FYI: on my machine when using LD_BIND_NOW=yes start time increased by 2.5 seconds. )
That said, IMHO if we do not find any better solution, LD_BIND_NOW is still acceptable.

mreynolds389: At a 2.5 second increase per restart it would increase the testing time by at least 1.5 to 2 hours, so if there are other options like LD_PRELOAD they should be explored.

droideck (approved):
The test looks good to me!

src/lib389/lib389/__init__.py
Firstyear: I can remove the LD_BIND_NOW, I think the change to linking in libslapd is actually enough to resolve the issue.

Firstyear:
Okay, @progier389 and @mreynolds389 I have been able to confirm that removing LD_BIND_NOW works provided that -lcrypt exists for libslapd when asan is enabled. Hopefully that should resolve the concerns re startup performance :)

progier389 (approved):
Thanks. 
LD_BIND_NOW was my only concern, since it is removed. It is ok for me ...

Issue #4415, opened 2020-11-02
Console fails to load schema if there are additional parentheses


mreynolds389:
Steps to Reproduce:
1. Stop the RHDS instance
2. Add a schema line containing parentheses in the X-ORIGIN information.
For instance:

attributetypes: ( 2.16.840.1.113730.3.1.604 NAME 'testName' DESC 'Sun ONE defined attribute type' EQUALITY 2.5.13.2 SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation X-ORIGIN ( 'Sun Java(TM)' 'user defined' ) )

3. Start the instance
4. Click on the Schema tab in the Console. Tables are empty.
    4-1.  Or use dsconf -j INST schema list
5. Things work fine if 'Sun Java(TM)' is changed to 'Sun Java'

Actual results:
Console doesn't load schema data.

Expected results:
Schema data should be loaded.

Additional info:
# dsconf --verbose -j ldapi://%2fvar%2frun%2fslapd-<INSTANCE>.socket schema list
DEBUG: The 389 Directory Server Configuration Tool
...
DEBUG: cn=schema get_attr_vals('objectClasses')
DEBUG: cn=schema get_attr_vals('attributeTypes')
DEBUG: list index out of range
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 134, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.6/site-packages/lib389/cli_conf/schema.py", line 32, in list_all
    attributetype_elems = schema.get_attributetypes(json=json)
  File "/usr/lib/python3.6/site-packages/lib389/schema.py", line 390, in get_attributetypes
    return self._get_schema_objects(AttributeType, json=json)
  File "/usr/lib/python3.6/site-packages/lib389/schema.py", line 149, in _get_schema_objects
    obj_i['x_origin'] = vals[0]
IndexError: list index out of range
{"desc": "list index out of range"}#
#

mreynolds389:
c4b1be456..264f173ae  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
f3317020b..c5d7865a6  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
4a8de7537..53d37df7f  389-ds-base-1.4.2 -> 389-ds-base-1.4.2

mreynolds389:
[...]

PR #4416, opened 2020-11-02
Issue 4415 - unable to query schema if there are extra parenthesis

mreynolds389:
Bug Description:  

When a client does a schema lookup in lib389 asking for the result in JSON, the X-ORIGIN is not correctly parsed if it contains an extra parenthesis

Fix Description:  

When parsing between the X-ORIGIN encapsulating parenthesis find the right most match, not the first match.

Relates: https://github.com/389ds/389-ds-base/issues/4415

### DIFF
--- a/dirsrvtests/tests/suites/clu/schema_test.py
+++ b/dirsrvtests/tests/suites/clu/schema_test.py
@@ -1,0 +1,61 @@
+import logging
+import pytest
+import os
+from lib389.topologies import topology_st as topo
+from lib389.schema import Schema
+
+log = logging.getLogger(__name__)
+
+
+def test_origins_with_extra_parenthesis(topo):
+    """Test the custom schema with extra parenthesis in X-ORIGIN can be parsed
+    into JSON
+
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/clu/schema_test.py
```diff
@@ -0,0 +1,57 @@
...
+        'equality': None,
+        'substr': None,
+        'ordering': None,
+        'usage': None,
+        'sup': None
+    }
+    schema.add_attributetype(parameters)
+
+    # Search for attribute with JSON option
+    schema.query_attributetype(ATTR_NAME, json=True)
```
droideck: Probably, it needs some kind of an assertion here that `x_origin` is really correct.

droideck (approved):
Besides this minor comment, the rest looks good!

dirsrvtests/tests/suites/clu/schema_test.py
mreynolds389: Changes made, please review.

droideck:
Ack! Thanks!

PR #4417, opened 2020-11-02
Issue 4176 - CL trimming causes high CPU

mreynolds389:
Bug Description:  

The changelog trimming switched to using pthread_cond_timedwait() instead of NSPR, but the relative time was used for the wait time instead of the absolute time.  This caused it to basically not wait at all and consume all the CPU.

Fix Description:  

Use the absolute(monotomic) time for the condition wait time.

Relates: https://github.com/389ds/389-ds-base/issues/4176

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/replication/changelog_trimming_test.py
+++ b/dirsrvtests/tests/suites/replication/changelog_trimming_test.py
@@ -87,7 +87,7 @@ def test_max_age(topo, setup_max_age):
         3. Trimming occurs
 
     """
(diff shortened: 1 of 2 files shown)


progier389 (approved):
ACK. 
cldb->clCvar is associated with monotonic clock: pthread_condattr_setclock(&(cldb->clCAttr), CLOCK_MONOTONIC);
So providing monotonic time in pthread_cond_timedwait makes sense.

Firstyear:
Ack from me too :)

PR #4423, opened 2020-11-09
Issue 4281 - dsidm user status fails with Error: 'nsUserAccount' obje…

bsimonova:
…ct has no attribute 'is_locked'

Description:
Created a test to verify bz1862971, because the status,lock and unlock options
were moved from dsidm user to dsidm account. The rest of the tests for dsidm will soon follow
so I have created helper functions for next tests.

Relates: https://github.com/389ds/389-ds-base/issues/4281
Relates: https://github.com/389ds/389-ds-base/issues/4348

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsidm_account_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_account_test.py
@@ -1,0 +1,132 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import time
+import subprocess
+import pytest
+import logging
+import os
+
+from lib389 import DEFAULT_SUFFIX
+from lib389.cli_idm.account import list, get_dn, lock, unlock, delete, modify, rename, entry_status, \
+    subtree_status, reset_password, change_password
+from lib389.topologies import topology_st
+from lib389.cli_base import FakeArgs
+from lib389.utils import ds_is_older
+from lib389.idm.user import nsUserAccounts
+
+pytestmark = pytest.mark.tier0
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+def check_value_and_reset(topology, content_list=None, content_list2=None, check_value=None, check_value_not=None):
+        if content_list2 is not None:
+            log.info('Check if content is present in output')
+            for item in content_list + content_list2:
+                assert topology.logcap.contains(item)
+
+        if content_list is not None:
+            log.info('Check if content is present in output')
+            for item in content_list:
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/clu/dsidm_account_test.py
```diff
@@ -0,0 +1,148 @@
...
+         6. Success
+    """
+
+    standalone = topology_st.standalone
+    users = nsUserAccounts(standalone, DEFAULT_SUFFIX)
+    test_user = users.get('test_user_1000')
+
+    entry_list = ['Entry DN: {}'.format(test_user.dn),
+                    'Entry Creation Date',
+                    'Entry Modification Date']
```
droideck: Small indentation issue

dirsrvtests/tests/suites/clu/dsidm_account_test.py
```diff
@@ -0,0 +1,148 @@
...
+         4. Run dsidm account entry status
+         5. Run dsidm account unlock
+         6. Run dsidm account entry status
+    :expectedresults:
+         1. Success
+         2. Success
+         3. Success
+         4. Success
+         5. Success
+         6. Success
```
droideck: Probably it makes sense to mention here - what exactly do you expect:

```
state_lock = 'Entry State: directly locked through nsAccountLock'
or
state_unlock= 'Entry State: activated'
```

droideck (approved):
Besides these two small issues - looks good to me! Ack!

dirsrvtests/tests/suites/clu/dsidm_account_test.py
bsimonova: Thanks for pointing that out, I fixed the indentation and messages in expected results, please review.

dirsrvtests/tests/suites/clu/dsidm_account_test.py
```diff
@@ -0,0 +1,148 @@
...
+from lib389.cli_idm.account import list, get_dn, lock, unlock, delete, modify, rename, entry_status, \
+    subtree_status, reset_password, change_password
+from lib389.topologies import topology_st
+from lib389.cli_base import FakeArgs
+from lib389.utils import ds_is_older
+from lib389.idm.user import nsUserAccounts
+
+pytestmark = pytest.mark.tier0
+
+LOG_FILE = '/tmp/dsidm_account.log'
```
Firstyear: I'm a bit worried about this use of a log file. Can't python logging log into a in memory buffer?

Firstyear: Actually, since your using FakeArgs and the cli test framework, there is a log capturing type. See /src/lib389/lib389/tests/cli and how it captures the output and tests it.

bsimonova: Thanks for reminding me about that, I changed the log capture accordingly. Please review.

Firstyear:
No problem, that looks much better!

Firstyear:
Ack from me too :)

PR #4424, opened 2020-11-10
Issue 4316 - performance search rate: useless poll on network send ca…

tbordaz:
…llback

Bug description:
	When sending back result/entries, DS first poll the connection to check
        it is able to write data on the socket. Then it writes the data.
	The purpose of the poll is to handle ioblocktimeout.
	The problem is that most of the time, the socket will process the write
	without any issue so it is useless to poll before the write.

Fix description:
	The fix is try write first. It polls for ioblocktimeout
        only if the write fails

relates: https://github.com/389ds/389-ds-base/issues/4316

Reviewed by:

Platforms tested: F31


Firstyear:
I've just read the code in daemon.c and after quite a bit of thought, I agree this change is correct :) 

Ack!

PR #4427, opened 2020-11-10
Issue 2054 - BUG - do not add referrals for masters with different data generation

progier389:
Bug description:
The problem is that some operation mandatory in the usual cases are
also performed when replication cannot take place because the
database set are differents (i.e: RUV generation ids are different)

One of the issue is that the csn generator state is updated when
starting a replication session (it is a problem when trying to
reset the time skew, as freshly reinstalled replicas get infected
by the old ones)

A second issue is that the RUV got updated when ending a replication session
(which may add replica that does not share the same data set,
then update operations on consumer retun referrals towards wrong masters

Fix description:
The fix checks the RUVs generation id before updating the csn generator
and before updating the RUV.

Related issue: [2054](https://github.com/389ds/389-ds-base/issues/2054)

Reviewed by:

Platforms tested: F32

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_test.py
@@ -13,6 +13,7 @@ from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts
 from lib389.pwpolicy import PwPolicyManager
 from lib389.utils import *
 from lib389.topologies import topology_m2 as topo_m2, TopologyMain, topology_m3 as topo_m3, create_topology, _remove_ssca_db, topology_i2 as topo_i2
+from lib389.topologies import topology_m2c2 as topo_m2c2
 from lib389._constants import *
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccount
@@ -22,6 +23,7 @@ from lib389.idm.directorymanager import DirectoryManager
 from lib389.replica import Replicas, ReplicationManager, Changelog5, BootstrapReplicationManager
 from lib389.agreement import Agreements
 from lib389 import pid_from_file
+from lib389.dseldif import *
 
 
 pytestmark = pytest.mark.tier1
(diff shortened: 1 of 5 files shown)


dirsrvtests/tests/suites/replication/regression_test.py
```diff
@@ -1054,6 +1056,294 @@ def test_online_init_should_create_keepalive_entries(topo_m2):
...
+
+    # Step 6: Perform update on both masters
+    repl.test_replication(m1, c1)
+    repl.test_replication(m2, c2)
+
+    # Step 7: Check that c1 has no time skew
+    # Stop server to insure that dse.ldif is uptodate
+    c1.stop()
+    c1_nsState = DSEldif(c1).readNsState(DEFAULT_SUFFIX)[0]
+    c1_timeSkew = int(c1_nsState['time_skew'])
```
Firstyear, line 1307: With time_skew, is it possible that it could occur given some other system clock change during the test?

Firstyear: Ahhh you set the time_skew manually. I still would be curious if it could be affected by other system affects though.

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -829,6 +829,37 @@ replica_set_ruv(Replica *r, RUV *ruv)
...
+        local_gen = ruv_get_replica_generation(local_ruv);
+        object_release(local_ruv_obj);
+    }
+    if (NULL == remote_gen || NULL == local_gen || strcmp(remote_gen, local_gen) != 0) {
+        return_value = PR_FALSE;
+    } else {
+        return_value = PR_TRUE;
+    }
+    slapi_ch_free((void **)&remote_gen);
+    slapi_ch_free((void **)&local_gen);
```
mreynolds389: This is very minor, but we try and use slapi_ch_free_string() when freeing char pointers (instead of slapi_ch_free()).  I realize you were just moving the code around, but it's still nice to do this clean up whenever we can.

ldap/servers/plugins/replication/repl5_inc_protocol.c
```diff
@@ -2161,26 +2162,12 @@ examine_update_vector(Private_Repl_Protocol *prp, RUV *remote_ruv)
...
-        if (NULL != local_ruv_obj) {
-            local_ruv = (RUV *)object_get_data(local_ruv_obj);
-            PR_ASSERT(local_ruv);
-            local_gen = ruv_get_replica_generation(local_ruv);
-            object_release(local_ruv_obj);
-        }
-        if (NULL == remote_gen || NULL == local_gen || strcmp(remote_gen, local_gen) != 0) {
-            return_value = EXAMINE_RUV_GENERATION_MISMATCH;
-        } else {
+	if (replica_check_generation(prp->replica, remote_ruv)) {
```
mreynolds389: Please be careful not to use tabs for indentation.  It should only be 4 spaces per "tab"

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -1315,6 +1346,11 @@ replica_update_csngen_state_ext(Replica *r, const RUV *ruv, const CSN *extracsn)
 
     PR_ASSERT(r && ruv);
 
+    if (!replica_check_generation(r, ruv)) /* ruv has wrong generation - we are done */
+    {
+       return 0;
```
mreynolds389: Indentation is off - one more space please :-)

mreynolds389:
Besides those minor issues I mentioned, this looks really good - great job on the test case!

dirsrvtests/tests/suites/replication/regression_test.py
progier389: As the test is based on time skew, yes changing  the system time while the test is running would affect the result.
   ( moving the time backwards may increase the final time skew and moving it forward may decrease it) 
That said with a timeSkewMargin of 5 minutes on a test that spends less than 2 minutes even a poorly written NTP software should not break the test.
And I suspect that quite a number of tests would be broken by a significant system time change.

dirsrvtests/tests/suites/replication/regression_test.py
```diff
@@ -1054,6 +1056,294 @@ def test_online_init_should_create_keepalive_entries(topo_m2):
...
+        7. Perform on line init from master1 to master2
+        8. Perform update on master2
+        9. Check that c1 RUV does contains URL towards m2
+    :expectedresults:
+        1. No error while generating ldif
+        2. No error while importing the ldif file
+        3. No error and Initialization done.
+        4. No error
+        5. master2 replicaid should not be in the consumer1 RUV
+        5. master2 replicaid should be in the consumer2 RUV
```
vashirov: Numbering is off

dirsrvtests/tests/suites/replication/regression_test.py
vashirov: We have pytest-libfaketime integration that can be used for time change shenanigans (see `dirsrvtests/tests/suites/healthcheck/health_security_test.py` for example).
And for the server part I had an unfinished integration for libfaketime with lib389 somewhere. It was editing the systemd drop-in config file to load libfaketime using LD_PRELOAD and control it using resource file. Something like here: https://bugzilla.redhat.com/show_bug.cgi?id=1843157#c4 
It has issues with SELinux though, so it needs some work to become a part of our tests.
I need to find it and submit.

dirsrvtests/tests/suites/replication/regression_test.py
progier389: Good catch !

dirsrvtests/tests/suites/replication/regression_test.py
progier389: Thank you for the information. that is interesting to know. 
Using libfaketime to change the result of time syscall within master2 is indeed another way to increase the csn generator time skew. 
That said it has exactly the same behavior than tweaking the csngen state (i.e test result may be wrong if global system time change by more than 5 minutes).
And due to the facts that lib389 already deals with csngen state internal to compute the time skew and that there are still issue around libfaketime solution, I propose to kept the csngen state tweak solution.

vashirov: Yes, it's just for information. I agree with your approach, it looks good to me.

dirsrvtests/tests/suites/replication/regression_test.py
progier389: fixed in latest force push

ldap/servers/plugins/replication/repl5_replica.c
progier389: fixed in latest force push

ldap/servers/plugins/replication/repl5_inc_protocol.c
progier389: fixed in latest force push

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -829,6 +829,37 @@ replica_set_ruv(Replica *r, RUV *ruv)
...
 }
 
+/*
+ * Check if replica generation is the same than the remote ruv one
+ */
+int
+replica_check_generation(Replica *r, const RUV *remote_ruv)
+{
+    int return_value;
+
```
mreynolds389: :-)  Please review this blank line.  Then ACK from me!

mreynolds389 (approved):
Besides that blank line this is good for me, thanks!

Issue #4429, opened 2020-11-11
NULL dereference in revert_cache()


mreynolds389:
**Issue Description**

There is a potential for a NULL pointer dereference in revert_cache(), we should add a check...

```
 void
 revert_cache(ldbm_instance *inst, struct timespec *start_time)
 {
+    if (inst == NULL) {
+        return;
+    }
     flush_hash(&inst->inst_cache, start_time, ENTRY_CACHE);
     flush_hash(&inst->inst_dncache, start_time, DN_CACHE);
 }
```

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=1896850

mreynolds389:
f1285d78a..32f30f268  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
5ea53ebd7..147b1df37  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
01a576bc4..d2f2e5074  389-ds-base-1.4.2 -> 389-ds-base-1.4.2

PR #4430, opened 2020-11-11
Issue 4429 - NULL dereference in revert_cache()

mreynolds389:
Bug Description:  During a delete, if the DN (with an escaped leading space)
                  of an existing entry fail to parse the server will revert
                  the entry update.  In this case it will lead to a crash
                  becuase ther ldbm inst struct is not set before it attempts
                  the cache revert.

Fix Description:  Check the the ldbm instance struct is not NULL before
                  dereferencing it.

Relates: https://github.com/389ds/389-ds-base/issues/4429

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/syntax/acceptance_test.py
+++ b/dirsrvtests/tests/suites/syntax/acceptance_test.py
@@ -6,12 +6,14 @@
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 
+import ldap
 import logging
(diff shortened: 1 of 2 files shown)


droideck (approved):
The C code change looks good to me! You have my ack.

P.S. the server doesn't crash during the test even if the change is not present. But maybe it's not 100% reproducible and happens not every time.

Firstyear:
Ack from me too

Issue #4383, opened 2020-10-18
group rdn with leading space char and add fails error 21 invalid syntax and delete fails error 32


mreynolds389:
Ticket was cloned from Red Hat Bugzilla (product _Red Hat Enterprise Linux 8_): [Bug 1888863](https://bugzilla.redhat.com/show_bug.cgi?id=1888863)

```
Description of problem:

deleting the DN of a group fails if there is a leading space char in the RDN
fails with error 32

user entries have a valid DN syntax with

dn: uid=\ leadingspaceuser1,cn=users,dc=example,dc=test
dn: uid=\20leadingspaceuser1,cn=users,dc=example,dc=test

dn: cn=\ leadingspaceuser1,cn=users,dc=example,dc=test
dn: cn=\20leadingspaceuser1,cn=users,dc=example,dc=test


but group DNs have a different behavior:

dn: cn=\ specialgroupname1,cn=groups,dc=example,dc=test
dn: cn=\20specialgroupname1,cn=groups,dc=example,dc=test

then, if such a group is in the database, it cannot be deleted, there is an
error 32

why do the special chars cannot be escaped for groups?


Version-Release number of selected component (if applicable):
RHEL-8.2
389-ds-base-1.4.x

but same on any versions of 389-ds-base


How reproducible:
on demand

Steps to Reproduce:
1. have some data

echo \ specialgroupname1 | base64
IHNwZWNpYWxncm91cG5hbWUxCg==

echo IHNwZWNpYWxncm91cG5hbWUxCg== | base64 -d
 specialgroupname1

cat << EOF > ~/dn.group.cn.with.leading.space.cn.users.ldif
dn: cn=users,dc=example,dc=test
objectClass: top
objectClass: nsContainer
cn: users

dn: uid=user1,cn=users,dc=example,dc=test
givenName: Test
sn: User1
uid: user1
cn: Test User1
displayName: Test User1
initials: TU
gecos: Test User1
homeDirectory: /home/user1
uidNumber: 969400001
gidNumber: 969400001
objectClass: person
objectClass: organizationalperson
objectClass: posixaccount
objectClass: inetorgperson
objectClass: inetuser

dn: uid=user2,cn=users,dc=example,dc=test
givenName: Test
sn: User2
uid: user2
cn: Test User2
displayName: Test User2
initials: TU
gecos: Test User2
homeDirectory: /home/user2
uidNumber: 969400002
gidNumber: 969400002
objectClass: person
[...]

PR #4431, opened 2020-11-11
Issue 4383 - Do not normalize escaped spaces in a DN

mreynolds389:
Bug Description:  

Adding an entry with an escaped leading space leads to many problems.  Mainly id2entry can get corrupted during an import of such an entry, and the entryrdn index is not updated correctly

Fix Description:  In slapi_dn_normalize_ext() leave an escaped space intact.

Relates: https://github.com/389ds/389-ds-base/issues/4383

### DIFF
--- a/dirsrvtests/tests/suites/syntax/acceptance_test.py
+++ b/dirsrvtests/tests/suites/syntax/acceptance_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2019 Red Hat, Inc.
+# Copyright (C) 2020 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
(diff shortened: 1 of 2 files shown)


Firstyear:
Seems good to me :)

progier389 (commented):
I would have liked to see some other tests like space in the middle and trailing spaces:
  'cn': 'another test' versus "cn=another\\20test,..." 
  'cn': 'third test ' versus "cn=third test\\20,..." 
Otherwise it looks good.

tbordaz:
The fix looks good to me as well.

mreynolds389:
> I would have liked to see some other tests like space in the middle and trailing spaces:
> 'cn': 'another test' versus "cn=another\20test,..."
> 'cn': 'third test ' versus "cn=third test\20,..."

Not a problem, I'll add some more tests...

mreynolds389:
There is a limitation in lib389 on adding escaped spaces in the middle of value using DSLdapObject.create(properties=props), but I was able to add trailing spaces.  So I updated the CI test for checking leading and trailing spaces.

We should probably create a new DN syntax test suite/file to handle other various DN's with different kinds of escaped values, etc, but that should be done via a new ticket.

progier389:
That sound fair.
So it is ok for me !

progier389 (commented):
LGTM

Issue #4432, opened 2020-11-11
After a failed online import the next imports are very slow


mreynolds389:
**Issue Description**

If you attempt to do an online import and entert he wrong LDIF name that import will fail.  When you try it again with the correct LDIF name the import becomes very slow. 

Reproducible steps:

[1]  Attempt an import with a typo in the LDIF file name so that the import fails:

         # dsconf slapd-localhost backend import userroot 1mil-typo.ldif 

[2] Fix the typo and do the import:

         # dsconf localhost backend import userroot 1mil.ldif 

[02/Nov/2020:16:44:06.244158418 -0500] - INFO - import_monitor_threads - import userroot: Processed 58 entries -- average rate 2.9/sec, recent rate 2.9/sec, hit ratio 0%
[02/Nov/2020:16:44:26.281630749 -0500] - INFO - import_monitor_threads - import userroot: Processed 108 entries -- average rate 2.7/sec, recent rate 2.7/sec, hit ratio 100%
[02/Nov/2020:16:44:46.314696042 -0500] - INFO - import_monitor_threads - import userroot: Processed 158 entries -- average rate 2.6/sec, recent rate 2.5/sec, hit ratio 100%
...

[3]  Restart the server, and try the import again:

[02/Nov/2020:16:51:19.259511417 -0500] - INFO - import_monitor_threads - import userroot: Processed 85095 entries -- average rate 4254.8/sec, recent rate 4254.7/sec, hit ratio 0%
[02/Nov/2020:16:51:39.516911186 -0500] - INFO - import_monitor_threads - import userroot: Processed 153041 entries -- average rate 3825.9/sec, recent rate 3825.9/sec, hit ratio 97%
[02/Nov/2020:16:51:59.708787576 -0500] - INFO - import_monitor_threads - import userroot: Processed 217130 entries -- average rate 3618.8/sec, recent rate 3300.9/sec, hit ratio 96%

mreynolds389:
9c7d59020..6233c0415  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
174b3cd5e..ef8241ee8  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=1837982

PR #4433, opened 2020-11-11
Issue 4432 - After a failed online import the next imports are very slow

mreynolds389:
Bug Description:  

When an online import fails the entry and DN caches are "reset", but we use the wrong "new maxsize" which was setting the entry cache maxsize to zero which killed the import performance.

Fix Description:  When resetting the caches use the previous cache maxsize.

Relates: https://github.com/389ds/389-ds-base/issues/4432

### DIFF
--- a/dirsrvtests/tests/suites/import/import_test.py
+++ b/dirsrvtests/tests/suites/import/import_test.py
@@ -13,9 +13,7 @@ Will test Import (Offline/Online)
 import os
 import pytest
 import time
-import shutil
 import glob
-import os
 from lib389.topologies import topology_st as topo
 from lib389._constants import DEFAULT_SUFFIX
 from lib389.dbgen import dbgen_users
@@ -25,8 +23,8 @@ from lib389.monitor import Monitor
 from lib389.backend import Backends
 from lib389.config import LDBMConfig
 from lib389.utils import ds_is_newer
-from lib389.idm.user import UserAccount, UserAccounts
-from lib389.idm.account import Accounts, Account
+from lib389.idm.user import UserAccount
+from lib389.idm.account import Accounts
 
 pytestmark = pytest.mark.tier1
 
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/import/import_test.py
```diff
@@ -357,6 +355,48 @@ def test_entry_with_escaped_characters_fails_to_import_and_index(topo, _import_c
...
+    bad_import_ldif = ldif_dir + '/perf_import_typo.ldif'
+
+    # Build LDIF file
+    dbgen_users(topo.standalone, 30000, import_ldif, DEFAULT_SUFFIX)
+
+    # Online import which fails
+    import_task = ImportTask(topo.standalone)
+    import_task.import_suffix_from_ldif(ldiffile=bad_import_ldif, suffix=DEFAULT_SUFFIX)
+    import_task.wait()
+
```
Firstyear, line 388: Does the cache size show up in either the error log or the cn=config that we can check and assert?

mreynolds389: It would be under the backend cn=monitor entry as "maxentrycachesize", but after the failed import the backend is disabled and you can not get any cn=monitor data on it.

Firstyear: Ahhh okay. Maybe in errorlog? I think we do emit some cache values there ... It'd be good to just have a way to assert the cache number really sticks in the process I think for the test.

mreynolds389: We do not "log" the entry cache any where.  The monitor has this info, but in this case its not useful since the backend is offline.  We do report the import cache, but in this case it's the backend entry cache that is the problem.  I'm just a little weary about adding more error logging especially since this (entry cache maxsize) is not something that typically needs to be monitored or exposed in the errors log.  I think this was just an oversight with some of the new backend rework that was done in 1.4.3. 

I know you want to simplify the test case, but I don't want to tweak the server code just to make this corner case bug more easily/reliably detectable.

Firstyear: Yeah that's fair. I think I just wanted the test case to "really assert" the fix was correct to prevent regressions, as much as I trust your work. So if you're happy with this "as is" then I think we can look past this.

progier389 (approved):
LGTM

PR #4437, opened 2020-11-13
Issue 4315: performance search rate: nagle triggers high rate of sets…

tbordaz:
…ocketopt

Bug description:
	When a socket is set with NO_DELAY=0 (nagle), written pdu are buffered
	until buffer is full or tcp_cork is set. This reduce network traffic when
        the application writes partial pdu.
        DS write complete pdu (results/entries/..) so it gives low benefit for DS.
	In addition nagle being 'on' by default, DS sets/unset socket tcp_cork to send
	immediately results/entries at each operation. This is an overhead of syscalls.

Fix description:
	Disable nagle by default

relates: https://github.com/389ds/389-ds-base/issues/4315

Reviewed by: ?

Platforms tested:  F33

### DIFF
--- a/ldap/servers/slapd/libglobs.c
+++ b/ldap/servers/slapd/libglobs.c
@@ -1639,12 +1639,11 @@ FrontendConfig_init(void)
 #endif /* USE_SYSCONF */
 
     init_accesscontrol = cfg->accesscontrol = LDAP_ON;
-#if defined(LINUX)
-    /* On Linux, by default, we use TCP_CORK so we must enable nagle */
-    init_nagle = cfg->nagle = LDAP_ON;
-#else
+
+    /* nagle triggers set/unset TCP_CORK setsockopt per operation
+     * as DS only sends complete PDU there is no benefit of nagle/tcp_cork
+     */
     init_nagle = cfg->nagle = LDAP_OFF;
-#endif
     init_security = cfg->security = LDAP_OFF;
     init_ssl_check_hostname = cfg->ssl_check_hostname = LDAP_ON;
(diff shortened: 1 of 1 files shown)


Firstyear:
I'm not sure this change works as intended. Looking at ./ldap/servers/slapd/daemon.c:2823, ifdef LINUX we always set enable_nagle to true on all connections regardless of the state of config_get_nagle. I think you need to remove those lines as well, so that we are relying on the value in libglobs correctly. 

Beside that, I agree to remove nagles algorithm by default in the server.

mreynolds389:
> I'm not sure this change works as intended. Looking at ./ldap/servers/slapd/daemon.c:2823, ifdef LINUX we always set enable_nagle to true on all connections regardless of the state of config_get_nagle. I think you need to remove those lines as well, so that we are relying on the value in libglobs correctly.

Actually that code is in configure_ns_socket() which is dead code.  So Thierry's patch is still valid, but the patch might as well remove all references to configure_ns_socket(): daemon.c & fe.h

Ack from me!

Firstyear:
Ahhh good point. In that case Ack as well :)

Firstyear:
(Since @tbordaz has been busy do we want to merge this for him @mreynolds389 ? )

Firstyear:
Merged for @tbordaz since I think he's busy, and there are two positive reviews.

tbordaz:
Oppsss. Thanks @Firstyear , I forgot it was not merged :(

Firstyear:
> Oppsss. Thanks @Firstyear , I forgot it was not merged :(

No problem mate, here to help out :)

PR #4438, opened 2020-11-13
Fix pytest test collection

vashirov:
Bug Description:
Some tests were missing tier0 and tier1 marks, didn't have _test postfix in the filename.
Because of this some tests were not collected and not executed.

Fix Description:

* Add missing pytest marks for tier0 and tier1
* Rename test modules to have _test in the filename.


droideck (approved):
I am not sure if the work is still going but the change looks good to me in the current state!

vashirov:
I was planning to add more fixes, but I think it can be done separately as they are related to the tests themselves, not to the collection issue.

PR #4441, opened 2020-11-15
Issue 4112 - Added a CI test

aadhikar:
Bug Description: If the dbhome directory is set, eg to /dev/shm/instance
then an online backup fails because it looks for the log.000000x file
in the wring diretcory.

Relates: https://github.com/389ds/389-ds-base/issues/4112

Reviewed by: ??

### DIFF
--- a/dirsrvtests/tests/suites/backups/backup_test.py
+++ b/dirsrvtests/tests/suites/backups/backup_test.py
@@ -7,6 +7,9 @@ from lib389.properties import BACKEND_SAMPLE_ENTRIES, TASK_WAIT
 from lib389.topologies import topology_st as topo
 from lib389.backend import Backend
 from lib389.tasks import BackupTask, RestoreTask
+from lib389.config import BDB_LDBMConfig
+from lib389 import DSEldif
+from lib389.utils import ds_is_older
 
 pytestmark = pytest.mark.tier1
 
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/backups/backup_test.py
```diff
@@ -64,7 +67,34 @@ def test_missing_backend(topo):
...
+        3. Check for the correct errors in the log
+    :expectedresults:
+        1. Success
+        2. Failure
+        3. Success
+    """
+    bdb_ldbmconfig = BDB_LDBMConfig(topo.standalone)
+    dseldif = DSEldif(topo.standalone)
+    topo.standalone.stop(timeout=10)
+    dseldif.replace(bdb_ldbmconfig.dn, 'nsslapd-db-home-directory', '/tmp/test')
```
Firstyear: It's not a good idea to hardcode in directory names like this because /tmp/test is something I even use and dump values into. You should look at a mktemp or other way of generating randomised folder names.

dirsrvtests/tests/suites/backups/backup_test.py
```diff
@@ -64,7 +67,34 @@ def test_missing_backend(topo):
...
+    topo.standalone.start(timeout=10)
```
Firstyear: Don't hardcode timeout in start/stop because on a highly loaded CI system, 10 seconds may not be enough. There is a default value here for a reason. Please remove the timeout.

Firstyear (changes requested):

aadhikar:
@Firstyear hey I have made the changes thanks for a quick review:)

droideck (approved):
Looks good to me!

dirsrvtests/tests/suites/backups/backup_test.py
```diff
@@ -64,7 +68,37 @@ def test_missing_backend(topo):
...
+        3. Check for the correct errors in the log
+    :expectedresults:
+        1. Success
+        2. Failure
+        3. Success
+    """
+    bdb_ldbmconfig = BDB_LDBMConfig(topo.standalone)
+    dseldif = DSEldif(topo.standalone)
+    topo.standalone.stop()
+    cmd = ['mktemp', '-d']
```
Firstyear: https://docs.python.org/3/library/tempfile.html

you don't need subprocess. Whenever possible, subprocess is a last resort, always look for a python api first.

Firstyear (changes requested):

aadhikar:
@Firstyear hey I have made the changes, I was on leave so the delay.

Firstyear:
Looks much better @aadhikar, I'll let @droideck have a look too just in case, but otherwise Ack from me.

Issue #4440, opened 2020-11-13
Using ldifgen with --start-idx option fails with unsupported operand


progier389:
Description of problem:
Using ldifgen with --start-idx option fails with ERROR: Error: unsupported operand type(s) for +=: 'int' and 'str'

Version-Release number of selected component (if applicable):
389-ds-base-1.4.3.8-2.module+el8.3.0+6591+ebfc9766.x86_64

How reproducible:
Always

Steps to Reproduce:
1. Create DS instance
2.

dsctl -v slapd-localhost ldifgen users --suffix dc=example,dc=com --parent ou=people,dc=example,dc=com --number 100000 --generic --start-idx=50
DEBUG: The 389 Directory Server Administration Tool
DEBUG: Inspired by works of: ITS, The University of Adelaide
DEBUG: Called with: Namespace(func=<function dbgen_create_users at 0x7feec90632f0>, generic=True, instance='slapd-localhost', json=False, ldif_file='users.ldif', list=False, localize=False, number='100000', parent='ou=people,dc=example,dc=com', rdn_cn=False, remove_all=False, start_idx='50', suffix='dc=example,dc=com', verbose=True)
DEBUG: Allocate <class 'lib389.DirSrv'> with None
DEBUG: Allocate <class 'lib389.DirSrv'> with b'localhost.localdomain':636
DEBUG: Allocate <class 'lib389.DirSrv'> with b'localhost.localdomain':636
DEBUG: Instance allocated
INFO:
Generating LDIF with the following options:
INFO: - number=100000
INFO: - suffix=dc=example,dc=com
INFO: - parent=ou=people,dc=example,dc=com
INFO: - generic=True
INFO: - start-idx=50
INFO: - rdn-cn=False
INFO: - localize=False
INFO: - ldif-file=/var/lib/dirsrv/slapd-localhost/ldif/users.ldif
INFO:
Writing LDIF ...
DEBUG: unsupported operand type(s) for +=: 'int' and 'str'
Traceback (most recent call last):
File "/usr/sbin/dsctl", line 138, in <module>
result = args.func(inst, log, args)
File "/usr/lib/python3.6/site-packages/lib389/cli_ctl/dbgen.py", line 195, in dbgen_create_users
dbgen_users(inst, args.number, args.ldif_file, args.suffix, generic=args.generic, parent=args.parent, startIdx=args.start_idx, rdnCN=False, pseudol10n=args.localize)
[...]

PR #4444, opened 2020-11-16
Issue 4440 - BUG - ldifgen with --start-idx option fails with unsuppo…

progier389:
…rted operand

Bug description:
Got TypeError exception when usign:
  dsctl -v slapd-localhost ldifgen users --suffix
     dc=example,dc=com --parent ou=people,dc=example,dc=com
     --number 100000 --generic --start-idx=50
The reason is that by default python parser provides
 value for numeric options:
  as an integer if specified by "--option value" or
  as a string if specified by "--option=value"

Fix description:
convert the numeric parameters to integer when using it.
 options impacted are:
  - in users subcommand:   --number ,  --start-idx
  - in mod-load subcommand:   --num-users, --add-users,
               --del-users, --modrdn-users, --mod-users

FYI: An alternative solution would have been to indicate the
parser that these values are an integer. But two reasons
 leaded me to implement the first solution:
 - first solution fix the problem for all users while the
   second one fixes only dsctl command.
 - first solution is easier to test:
    I just added a new test file generated by a script
      that duplicated existing ldifgen test, renamed the
       test cases and replaced the numeric arguments by
       strings.
   Second solution would need to redesign the test framework
    to be able to test the parser.

relates: https://github.com/389ds/389-ds-base/issues/4440

Reviewed by:

Platforms tested: F32


src/lib389/lib389/dbgen.py
```diff
@@ -220,6 +220,9 @@ def dbgen_users(instance, number, ldif_file, suffix, generic=False, entry_name="
     """
     Generate an LDIF of randomly named entries
     """
+    #Lets insure that integer parameters are not string
```
mreynolds389: Please add a space between the # and the comment :-)  Other than that, ack!

Firstyear:
Ack from me, awesome work!

progier389:
Commit: 3c3e1f30cdb046a1aabb93aacebcf261a76a0892
33279b75e..3c3e1f30c    Master
ea6e4a84e..5eacadd1e   389-ds-base-1.4.4
cda9c3e93..3b6cd3fb3    389-ds-base-1.4.3

Issue #4384, opened 2020-10-19
PR_WaitCondVar() issue causes replication delay when clock jumps backwards


tiran:
**Issue Description**

See issue is explained in detailed on FreeeIPA upstream bug https://pagure.io/freeipa/issue/8544. 

The root cause of the problem is NSPR's ``PR_WaitCondVar()`` implementation. On pthread platforms the function is built on top of ``pthread_cond_timedwait``. The function takes an absolute time. Both ``pthread_cond_timedwait`` and ``PR_WaitCondVar`` internal helper ``pt_TimedWait`` default to  realtime clock which is subject to clock changes. ``pthread_condattr_setclock`` with ``CLOCK_MONOTONIC`` could be used to change the clock source from real to monotonic clock. This fix would require changes internal APIs or even a new API in NSPR.

Summary:

* FreeIPA server with DNS role use local BIND DNS instance as only DNS server.
* DNS facilities are not available until 389-DS, KDC, and BIND DNS are up and working.
* chronyd cannot synchronize clocks until DNS is up.
* Kerberos / GSSAPI doesn't work correctly when clocks of two hosts are way out of sync.
* When chronyd eventually syncs time, a backward jump confuses the sleep timer of the thread sync condition. This can extend a timed condition wait from a few seconds to several hours.

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: 389-ds-base-1.4.3.12-1.fc32.x86_64

**Steps to Reproduce**
1. Install IPA cluster with multiple replications and DNS on each server
2. change hardware clock on one server so chronyd has to adjust clock by 2 hours on boot
3. reboot server

**Expected results**
Replication session backing off timer is not affected by clock jumps. For example a wait timeout of 60 seconds is not shortened or prolonged by real time clock changes. Instead it uses a monotonic, stable clock source.

**Logs with replication debug information**

Replication connection fails with GSSAPI auth problem because remote server does not like our Kerboers ticket. This is expected behavior. 389-DS retries a couple of times with increasing back off time:

```
[18/Oct/2020:20:05:57.131884318 +0200] - DEBUG - NSMMReplicationPlugin - bind_and_check_pwp - agmt="cn=meTovm-178.ipa.example" (vm-178:389) - Replication bind with GSSAPI auth failed: LDAP error 49 (Invalid credentials) (SASL(-13): authentication failure: GSSAPI Failure: gss_accept_sec_context)
[...]

PR #4450, opened 2020-11-20
Issue 4384 - Use MONOTONIC clock for all timing events and conditions

mreynolds389:
Bug Description:  

All of the server's event handling and replication were based on REALTIME clocks, which can be influenced by the system time changing.  This could causes massive delays, and simply cause unexpected behavior.

Fix Description:  

Move all condition variables to use pthread instead of NSPR functions.  Also make sure we use MONOTONIC clocks when we get the current time when checking for timeouts and other timed events.

All CI tests that could be affected by this code pass.

Relates: https://github.com/389ds/389-ds-base/issues/4384

### DIFF
--- a/dirsrvtests/tests/suites/plugins/entryusn_test.py
+++ b/dirsrvtests/tests/suites/plugins/entryusn_test.py
@@ -6,9 +6,11 @@
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 #
+import os
 import ldap
 import logging
 import pytest
+import time
 from lib389._constants import DEFAULT_SUFFIX
 from lib389.config import Config
 from lib389.plugins import USNPlugin, MemberOfPlugin
@@ -211,6 +213,7 @@ def test_entryusn_after_repl_delete(topology_m2):
         user_usn = user_1.get_attr_val_int('entryusn')
 
         user_1.delete()
+        time.sleep(1)  # Gives a little time for tombstone creation to complete
 
         ts = tombstones.get(user_rdn)
         ts_usn = ts.get_attr_val_int('entryusn')
--- a/ldap/servers/plugins/replication/repl5_replica_config.c
+++ b/ldap/servers/plugins/replication/repl5_replica_config.c
@@ -1,6 +1,6 @@
 /** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
- * Copyright (C) 2005 Red Hat, Inc.
+ * Copyright (C) 2020 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -31,14 +31,17 @@
 #define CLEANALLRUVLEN 11
 #define REPLICA_RDN "cn=replica"
 
+#define CLEANALLRUV_MAX_WAIT 7200 /* 2 hours */
+#define CLEANALLRUV_SLEEP 5
+
 int slapi_log_urp = SLAPI_LOG_REPL;
 static ReplicaId cleaned_rids[CLEANRID_BUFSIZ] = {0};
 static ReplicaId pre_cleaned_rids[CLEANRID_BUFSIZ] = {0};
 static ReplicaId aborted_rids[CLEANRID_BUFSIZ] = {0};
 static PRLock *rid_lock = NULL;
 static PRLock *abort_rid_lock = NULL;
-static PRLock *notify_lock = NULL;
-static PRCondVar *notify_cvar = NULL;
+static pthread_mutex_t notify_lock;
+static pthread_cond_t notify_cvar;
 static PRLock *task_count_lock = NULL;
 static int32_t clean_task_count = 0;
 static int32_t abort_task_count = 0;
@@ -106,6 +109,9 @@ dont_allow_that(Slapi_PBlock *pb __attribute__((unused)),
 int
 replica_config_init()
 {
+    int rc = 0;
+    pthread_condattr_t condAttr;
+
     s_configLock = PR_NewLock();
 
     if (s_configLock == NULL) {
@@ -135,18 +141,31 @@ replica_config_init()
                       PR_GetError());
         return -1;
     }
-    if ((notify_lock = PR_NewLock()) == NULL) {
-        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name, "replica_config_init - "
-                                                       "Failed to create notify lock; NSPR error - %d\n",
-                      PR_GetError());
+    if ((rc = pthread_mutex_init(&notify_lock, NULL)) != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "replica_config_init",
+                      "Failed to create notify lock: error %d (%s)\n",
+                      rc, strerror(rc));
         return -1;
     }
-    if ((notify_cvar = PR_NewCondVar(notify_lock)) == NULL) {
-        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name, "replica_config_init - "
-                                                       "Failed to create notify cond var; NSPR error - %d\n",
-                      PR_GetError());
+    if ((rc = pthread_condattr_init(&condAttr)) != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "replica_config_init",
+                      "Failed to create notify new condition attribute variable. error %d (%s)\n",
+                      rc, strerror(rc));
         return -1;
     }
+    if ((rc = pthread_condattr_setclock(&condAttr, CLOCK_MONOTONIC)) != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "replica_config_init",
+                      "Cannot set condition attr clock. error %d (%s)\n",
+                      rc, strerror(rc));
+        return -1;
+    }
+    if ((rc = pthread_cond_init(&notify_cvar, &condAttr)) != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "replica_config_init",
+                      "Failed to create new condition variable. error %d (%s)\n",
+                      rc, strerror(rc));
+        return -1;
+    }
+    pthread_condattr_destroy(&condAttr);
 
     /* config DSE must be initialized before we get here */
     slapi_config_register_callback(SLAPI_OPERATION_ADD, DSE_FLAG_PREOP, CONFIG_BASE, LDAP_SCOPE_SUBTREE,
@@ -1755,9 +1774,13 @@ replica_cleanallruv_thread(void *arg)
          * to startup timing issues, we need to wait before grabbing the replica obj, as
          * the backends might not be online yet.
          */
-        PR_Lock(notify_lock);
-        PR_WaitCondVar(notify_cvar, PR_SecondsToInterval(10));
-        PR_Unlock(notify_lock);
+        struct timespec current_time = {0};
+        clock_gettime(CLOCK_MONOTONIC, &current_time);
+        current_time.tv_sec += 10;
+
+        pthread_mutex_lock(&notify_lock);
+        pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+        pthread_mutex_unlock(&notify_lock);
         data->replica = replica_get_replica_from_dn(data->sdn);
         if (data->replica == NULL) {
             cleanruv_log(data->task, data->rid, CLEANALLRUV_ID, SLAPI_LOG_ERR, "Unable to retrieve repl object from dn(%s).", data->sdn);
@@ -1801,15 +1824,18 @@ replica_cleanallruv_thread(void *arg)
     ruv_obj = replica_get_ruv(data->replica);
     ruv = object_get_data(ruv_obj);
     while (data->maxcsn && !is_task_aborted(data->rid) && !is_cleaned_rid(data->rid) && !slapi_is_shutting_down()) {
+        struct timespec current_time = {0};
         if (csn_get_replicaid(data->maxcsn) == 0 ||
             ruv_covers_csn_cleanallruv(ruv, data->maxcsn) ||
             strcasecmp(data->force, "yes") == 0) {
             /* We are caught up, now we can clean the ruv's */
             break;
         }
-        PR_Lock(notify_lock);
-        PR_WaitCondVar(notify_cvar, PR_SecondsToInterval(5));
-        PR_Unlock(notify_lock);
+        clock_gettime(CLOCK_MONOTONIC, &current_time);
+        current_time.tv_sec += CLEANALLRUV_SLEEP;
+        pthread_mutex_lock(&notify_lock);
+        pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+        pthread_mutex_unlock(&notify_lock);
     }
     object_release(ruv_obj);
     /*
@@ -1877,18 +1903,20 @@ replica_cleanallruv_thread(void *arg)
         /*
          *  need to sleep between passes
          */
-        cleanruv_log(data->task, data->rid, CLEANALLRUV_ID, SLAPI_LOG_NOTICE, "Not all replicas have received the "
-                                                                        "cleanallruv extended op, retrying in %d seconds",
+        cleanruv_log(data->task, data->rid, CLEANALLRUV_ID, SLAPI_LOG_NOTICE,
+                     "Not all replicas have received the cleanallruv extended op, retrying in %d seconds",
                      interval);
         if (!slapi_is_shutting_down()) {
-            PR_Lock(notify_lock);
-            PR_WaitCondVar(notify_cvar, PR_SecondsToInterval(interval));
-            PR_Unlock(notify_lock);
+            struct timespec current_time = {0};
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_unlock(&notify_lock);
         }
-        if (interval < 14400) { /* 4 hour max */
-            interval = interval * 2;
-        } else {
-            interval = 14400;
+        interval *= 2;
+        if (interval >= CLEANALLRUV_MAX_WAIT) {
+            interval = CLEANALLRUV_MAX_WAIT;
         }
     }
     /*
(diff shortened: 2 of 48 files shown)


elkris:
Hi Mark,
didn't check all the changes, but two comments.
1] The functions in slapi2nspr.c no longer are 2nspr, shouldn't the file name change as well ?
2] What usage is left of NSPR ? With this unconditional move away from PR_* the multiple platform support is gone, which is ok, think it didn't work on windows for quite some time. Isn't it time to remove it completely, eg for file system access and so on. ?

mreynolds389:
> Hi Mark,

Hi Ludwig!

> didn't check all the changes, but two comments.
> 1] The functions in slapi2nspr.c no longer are 2nspr, shouldn't the file name change as well ?

I actuality did think about that, but there are other similar (non-nspr) functions in that same file like slapi_new_rwlock_prio().  So I just left it as is.  Perhaps we should just rename the file to something more accurate?  

> 2] What usage is left of NSPR ? With this unconditional move away from PR_* the multiple platform support is gone, which is ok, think it didn't work on windows for quite some time. Isn't it time to remove it completely, eg for file system access and so on. ?

Yes we want to ditch NSPR completely, but that work was out of the scope for this specific issue (this patch is already very large).  I will open a new ticket to investigate completely removing NSPR...

elkris:
> > 1] The functions in slapi2nspr.c no longer are 2nspr, shouldn't the file name change as well ?
> 
> I actuality did think about that, but there are other similar (non-nspr) functions in that same file like slapi_new_rwlock_prio(). So I just left it as is. Perhaps we should just rename the file to something more accurate?
> 
yes, that was the idea. It was slapi 2 nspr where nspr is the netscape portable runtime, if it is no longer portable (and ns) so what remains is slapi2runtime - or something like this

mreynolds389:
> > > 1] The functions in slapi2nspr.c no longer are 2nspr, shouldn't the file name change as well ?
> > 
> > 
> > I actuality did think about that, but there are other similar (non-nspr) functions in that same file like slapi_new_rwlock_prio(). So I just left it as is. Perhaps we should just rename the file to something more accurate?
> 
> yes, that was the idea. 

Sorry I read your first comment too quickly (trying to get too much done a Friday I suppose), I originally thought you wanted me to move the new functions to a brand new file.  I wasn't trying to steal your suggestion :-)  Anyway I will rename that file...

Firstyear:
I'm not sure about replacing all of these in a single large sweep. If there are any time values in use that are sent to other servers, then we *can not* use monotonic. Monotonic is an arbitrary local value only useful for asserting the progress of time forwards on a system. So I think here my concern would be that replacing all these in a "big bang" we might find some usage of our clock that was not contained to a single server. 

Would it be possible to target these clock changes to per-subsystem instead? 

Another aspect too is that related to #4384 is that it's not just the clock, but the design of the queue that is also fundamentally flawed so that needs to be considered also.

mreynolds389:
> I'm not sure about replacing all of these in a single large sweep. If there are any time values in use that are sent to other servers, then we _can not_ use monotonic. Monotonic is an arbitrary local value only useful for asserting the progress of time forwards on a system. So I think here my concern would be that replacing all these in a "big bang" we might find some usage of our clock that was not contained to a single server.

Every clock I changed was for timeouts and timing events, not the actual of time of day.  I did not blindly replace *all* the time functions (slapi_current_utc_time).  I did run all the CI tests we have that would exercise this code and found no problems.  The other issue is that these are all connected.  I'm not sure I can really break it up into meaningful smaller chunks.  The replication code, which is what we are trying to fix, needs the eventq changes, and once the eventq changed we needed to change all the code that uses it, etc.  It's kind of all or nothing...

> 
> Would it be possible to target these clock changes to per-subsystem instead?
> 
> Another aspect too is that related to #4384 is that it's not just the clock, but the design of the queue that is also fundamentally flawed so that needs to be considered also.

Right, there are many problems mentioned in the Issue, but this PR is just for the clock issue (which will address the main problem reported in the issue).  The other issues will be addressed in future PRs.

Firstyear:
> Every clock I changed was for timeouts and timing events, not the actual of time of day. I did not blindly replace _all_ the time functions (slapi_current_utc_time). I did run all the CI tests we have that would exercise this code and found no problems. The other issue is that these are all connected. I'm not sure I can really break it up into meaningful smaller chunks. The replication code, which is what we are trying to fix, needs the eventq changes, and once the eventq changed we needed to change all the code that uses it, etc. It's kind of all or nothing...
> 

Ahhh I see. Sorry for the mistake, I'll review this a bit more carefully now then :) 

> Right, there are many problems mentioned in the Issue, but this PR is just for the clock issue (which will address the main problem reported in the issue). The other issues will be addressed in future PRs.

Cool :)

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -1801,15 +1821,18 @@ replica_cleanallruv_thread(void *arg)
...
             ruv_covers_csn_cleanallruv(ruv, data->maxcsn) ||
             strcasecmp(data->force, "yes") == 0) {
             /* We are caught up, now we can clean the ruv's */
             break;
         }
-        PR_Lock(notify_lock);
-        PR_WaitCondVar(notify_cvar, PR_SecondsToInterval(5));
-        PR_Unlock(notify_lock);
+        clock_gettime(CLOCK_MONOTONIC, &current_time);
+        current_time.tv_sec += 5;
```
Firstyear: hardcoded value, should this be a #define in the file?

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -1877,13 +1900,16 @@ replica_cleanallruv_thread(void *arg)
...
-            PR_WaitCondVar(notify_cvar, PR_SecondsToInterval(interval));
-            PR_Unlock(notify_lock);
+            struct timespec current_time = {0};
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_unlock(&notify_lock);
         }
         if (interval < 14400) { /* 4 hour max */
```
Firstyear: Similar here with 14400

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
```diff
@@ -1429,21 +1429,19 @@ import_free_job(ImportJob *job)
          * To avoid freeing fifo queue under bulk_import_queue use
          * job lock to synchronize
          */
-        if (job->wire_lock)
-            PR_Lock(job->wire_lock);
+        if (&job->wire_lock)
```
Firstyear: missing { }

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -2036,14 +2042,17 @@ bdb_pre_close(struct ldbminfo *li)
...
             PRIntervalTime before = PR_IntervalNow();
             /* There are 3 ways to wake up from this WaitCondVar:
                1) The last database thread exits and calls NotifyCondVar - thread_count
                should be 0 in this case
                2) Timeout - in this case, thread_count will be > 0 - bad
                3) A bad error occurs - bad - will be reported as a timeout
             */
-            PR_WaitCondVar(pEnv->bdb_thread_count_cv, cvwaittime);
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += 25; /* cvwaittime but in seconds */
```
Firstyear: hardcoded value could be #defined?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -3566,20 +3582,24 @@ log_flush_threadmain(void *param)
...
+                    struct timespec current_time = {0};
                     if (BDB_CONFIG(li)->bdb_stop_threads)
                         break;
                     if (PR_IntervalNow() - last_flush > interval_flush) {
                         do_flush = 1;
                         break;
                     }
-                    PR_WaitCondVar(sync_txn_log_do_flush, interval_wait);
+                    clock_gettime(CLOCK_MONOTONIC, &current_time);
+                    current_time.tv_sec += trans_batch_txn_max_sleep * 1000;
```
Firstyear: Why times 1000?

ldap/servers/slapd/slapi2runtime.c
```diff
@@ -44,47 +46,49 @@
 Slapi_Mutex *
 slapi_new_mutex(void)
```
Firstyear, line 47: Could consider inline attributes here.

ldap/servers/slapd/slapi2runtime.c
```diff
@@ -44,47 +46,49 @@
...
+ * Description: behaves just like pthread_mutex_unlock().
  * Returns:
  *    non-zero if mutex was successfully unlocked.
  *    0 if mutex is NULL or is not locked by the calling thread.
  */
 int
 slapi_unlock_mutex(Slapi_Mutex *mutex)
 {
-    if (mutex == NULL || PR_Unlock((PRLock *)mutex) == PR_FAILURE) {
+    if (mutex == NULL || pthread_mutex_unlock((pthread_mutex_t *)mutex) != 0) {
```
Firstyear, line 92: PR_ASSERT mutex != NULL? It seems like a condition we should never have ...?

Firstyear (changes requested):
Mostly really good, few small things. :)

mreynolds389:
> > Every clock I changed was for timeouts and timing events, not the actual of time of day. I did not blindly replace _all_ the time functions (slapi_current_utc_time). I did run all the CI tests we have that would exercise this code and found no problems. The other issue is that these are all connected. I'm not sure I can really break it up into meaningful smaller chunks. The replication code, which is what we are trying to fix, needs the eventq changes, and once the eventq changed we needed to change all the code that uses it, etc. It's kind of all or nothing...
> 
> Ahhh I see. Sorry for the mistake, I'll review this a bit more carefully now then :)

No worries,  It's a big change, and I'm not saying it's not without risk.  But I did do a lot of testing with this PR, but you never know something can always slip through the cracks...

Firstyear:
> > > Every clock I changed was for timeouts and timing events, not the actual of time of day. I did not blindly replace _all_ the time functions (slapi_current_utc_time). I did run all the CI tests we have that would exercise this code and found no problems. The other issue is that these are all connected. I'm not sure I can really break it up into meaningful smaller chunks. The replication code, which is what we are trying to fix, needs the eventq changes, and once the eventq changed we needed to change all the code that uses it, etc. It's kind of all or nothing...
> > 
> > 
> > Ahhh I see. Sorry for the mistake, I'll review this a bit more carefully now then :)
> 
> No worries, It's a big change, and I'm not saying it's not without risk. But I did do a lot of testing with this PR, but you never know something can always slip through the cracks...

Yeah I just didn't realise how it was all pretty interconnected. Once I read it, I can see how it escalated quickly for you.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: It should be a divide!  :-)  And since its dealing with 50 milliseconds this also needs special care for nanosecond overflow.  I'll work on this tomorrow...

ldap/servers/slapd/slapi2runtime.c
mreynolds389: Not sure what you mean here, can you elaborate?

mreynolds389:
Most changes applied @Firstyear please review..

dirsrvtests/tests/suites/plugins/entryusn_test.py
```diff
@@ -209,6 +211,7 @@ def test_entryusn_after_repl_delete(topology_m2):
         user_usn = user_1.get_attr_val_int('entryusn')
 
         user_1.delete()
+        time.sleep(1)
```
Firstyear: Could be worth commenting why this time.sleep exists.

Firstyear (commented):
Example 

```
static inline void __attribute__((always_inline))
_pblock_assert_pb_task(Slapi_PBlock *pblock) { ... }
```

Because these are small functions that just immediately call something else, it can be beneficial to just wrap them in inline attributes to hint to the compiler to inline rather than function call them.

Firstyear:
Anyway, only those two comments left, otherwise ack. :) Could be wise to have @tbordaz check too if he's around since it's a larger change.

mreynolds389:
Changes applied, @tbordaz would you mind looking this over?

tbordaz:
It remains use of PR_Now repl5_schedule.c (schedule replication Linux/Window) that should use monotonic

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -1877,18 +1903,21 @@ replica_cleanallruv_thread(void *arg)
...
-        }
-        if (interval < 14400) { /* 4 hour max */
+            struct timespec current_time = {0};
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_unlock(&notify_lock);
+        }
+        if (interval < CLEANALLRUV_MAX_WAIT) {
```
tbordaz: It looks interval can go above CLEANALLRUV_MAX_WAIT
should rather do
interval *= 2;
if (interval >= CLEANALLRUV_MAX_WAIT) {
   interval = CLEANALLRUV_MAX_WAIT
}

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -1938,18 +1967,21 @@ replica_cleanallruv_thread(void *arg)
...
-            PR_Unlock(notify_lock);
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_unlock(&notify_lock);
         }
 
-        if (interval < 14400) { /* 4 hour max */
+        if (interval < CLEANALLRUV_MAX_WAIT) {
```
tbordaz: idem

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -2162,15 +2194,18 @@ check_replicas_are_done_cleaning(cleanruv_data *data)
...
-            PR_Unlock(notify_lock);
+            struct timespec current_time = {0};
+            clock_gettime(CLOCK_MONOTONIC, &current_time);
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_lock(&notify_lock);
         }
 
-        if (interval < 14400) { /* 4 hour max */
```
tbordaz, line 2170: idem

ldap/servers/plugins/replication/repl5_replica_config.c
```diff
@@ -2271,14 +2306,17 @@ check_replicas_are_done_aborting(cleanruv_data *data)
...
+            current_time.tv_sec += interval;
+            pthread_mutex_lock(&notify_lock);
+            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
+            pthread_mutex_unlock(&notify_lock);
+        }
+        if (interval < CLEANALLRUV_MAX_WAIT) {
             interval = interval * 2;
         } else {
-            interval = 14400;
+            interval = CLEANALLRUV_MAX_WAIT;
```
tbordaz, line 2314: idem and several others below

ldap/servers/plugins/roles/roles_cache.c
```diff
@@ -343,7 +343,7 @@ roles_cache_create_suffix(Slapi_DN *sdn)
 
     slapi_lock_mutex(new_suffix->create_lock);
     if (new_suffix->is_ready != 1) {
-        slapi_wait_condvar(new_suffix->suffix_created, NULL);
+        slapi_wait_condvar_pt(new_suffix->suffix_created, new_suffix->create_lock, NULL);
```
tbordaz, line 346: Is it correct ?
suffix_created is Slapi_Condvar no pthread_cond_t
create_lock is Slapi_Mutex not pthread_mutex_t

ldap/servers/plugins/roles/roles_cache.c
```diff
@@ -384,7 +384,7 @@ roles_cache_wait_on_change(void *arg)
             test roles_def->keeprunning before
             going to sleep.
         */
-        slapi_wait_condvar(roles_def->something_changed, NULL);
+        slapi_wait_condvar_pt(roles_def->something_changed, roles_def->change_lock, NULL);
```
tbordaz, line 387: idem

ldap/servers/slapd/eventq.c
```diff
@@ -423,12 +451,22 @@ eq_stop()
...
+
+        pthread_mutex_lock(&(eq->eq_lock));
+        pthread_cond_broadcast(&(eq->eq_cv));
+        pthread_mutex_unlock(&(eq->eq_lock));
+
+        pthread_mutex_lock(&ss_lock);
+        clock_gettime(CLOCK_MONOTONIC, &current_time);
+        if (current_time.tv_nsec + 100000000 > 1000000000) {
+            /* ns will overflow, just bump the seconds */
+            current_time.tv_sec++;
```
tbordaz, line 464: shouldn't we set current_time.tv_nsec = 1s - current_time.tv_nsec

ldap/servers/slapd/house.c
```diff
@@ -84,9 +99,16 @@ void
 housekeeping_stop()
 {
     if (housekeeping_tid) {
-        PR_Lock(housekeeping_mutex);
-        PR_NotifyCondVar(housekeeping_cvar);
-        PR_Unlock(housekeeping_mutex);
+        /* Notify the thread */
+        pthread_mutex_lock(&housekeeping_mutex);
+        pthread_cond_broadcast(&housekeeping_cvar);
```
tbordaz: should be pthread_cond_signal

ldap/servers/slapd/psearch.c
```diff
@@ -101,21 +101,26 @@ void
 ps_init_psearch_system()
 {
     if (!PS_IS_INITIALIZED()) {
+        int rc = 0;
```
tbordaz: int32_r ;)

mreynolds389:
> It remains use of PR_Now repl5_schedule.c (schedule replication Linux/Window) that should use monotonic

The schedule needs to use the REALTIME.  For example, MONOTONIC can *not* be used to see how many seconds we are away from midnight.  Yes the schedule is impacted by a system time change, but switching to MONOTONIC would completely break the schedule :-)

ldap/servers/slapd/house.c
Firstyear: If there is only one housekeeping thread, there is no difference between broadcast and signal?

mreynolds389: @firstyear yes that's true, but I'm going to change to a signal anyway so we have a proper example in the code of how to use these functions.

ldap/servers/plugins/roles/roles_cache.c
mreynolds389: Slapi_Condvar and Slapi_Mutex was basically wrappers, this is correct.

ldap/servers/slapd/eventq.c
mreynolds389: I don't think we can do that because we need to move the clock forward, this could move it backwards because we are not checking the current value of nsecs.  nsecs could already be at 999999999, so changing nano seconds to 1 second minus the delay would move the clock backwards and that would negate the timed wait.  

Now the current code is not perfect, it's not always going to sleep 100 ms if there is an overflow of nanoseconds.  I will revise this...

tbordaz: Yes the change looks good. I had a doubt regarding  the lost of nsec in case of nsec overflow.

tbordaz (approved):
The patch looks good to  me. ack

PR #4453, opened 2020-11-20
Issue 3657 - Add options to dsctl for dsrc file

mreynolds389:
Description:  

Add options to create, delete, and display the .dsrc CLI tool shortcut file.

Relates: https://github.com/389ds/389-ds-base/issues/3657

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsrc_test.py
+++ b/dirsrvtests/tests/suites/clu/dsrc_test.py
@@ -1,0 +1,136 @@
+import logging
+import pytest
+import os
+from os.path import expanduser
+from lib389.cli_base import FakeArgs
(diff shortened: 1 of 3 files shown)


Firstyear:
I like the intent of this, but perhaps you should consider using https://docs.python.org/3/library/configparser.html rather than manually writing out the INF content. This way you can also then update a .dsrc with section-awareness later when new instances are added.

mreynolds389:
Okay, made a bunch of changes, and added a CI test.  Now the tool focuses on the concept of a "config section" verses a "file".  So now you can add/modify/delete config sections.  However, the display option prints *all* the sections in the file.

Firstyear:
That's awesome, I really like this much better! Ack :)

PR #4455, opened 2020-11-23
Issue 4454 - RFE - fix version numbers to allow object caching

### LINKED ISSUE #4454: BUILDNUM/VERSION_DATE prevent object caching during compiling
It's common to want to use things like ccache to cache C objects for re-compilation to speed up recompiles. A similar tool is sccache from mozilla, which in addition to C/C++, can cache Rust. 

When experimenting with sccache I noticed that DS was unable to be cached, and would always be recompiled. This shows up as cache misses. Here is a list of the stats after a clean DS build, with a hot cache:

```
#24 258.4 Compile requests                    900
#24 258.4 Compile requests executed           648
#24 258.4 Cache hits                          186
#24 258.4 Cache hits (C/C++)                  107
#24 258.4 Cache hits (Rust)                    79
#24 258.4 Cache misses                        455
#24 258.4 Cache misses (C/C++)                455
...
#24 258.4 Cache location                  Redis: redis://172.24.20.4:6379
#24 258.4 Cache size                          680 MiB
#24 258.4 Max cache size                        2 GiB
```

There are 455 Cache misses, all in C/C++.

This occurs as sccache relies on the executed compilation command to be the same, however our passing of BUILDNUM as a -D switch breaks this. Additionally the VERSION_DATE also changes some header content which also causes recompiles of some files.

Setting both BUILDNUM and VERSION_DATE to 0 values allows the cache to work:

```
#24 234.5 Compile requests                    900
#24 234.5 Compile requests executed           648
#24 234.5 Cache hits                          641
#24 234.5 Cache hits (C/C++)                  562
#24 234.5 Cache hits (Rust)                    79
#24 234.5 Cache misses                          0
...
#24 234.5 Cache location                  Redis: redis://172.24.20.4:6379
#24 234.5 Cache size                          982 MiB
#24 234.5 Max cache size                        2 GiB

```

The difference approaches 2 to 3 minutes reduction in compilation time on my i7 system. 

My proposal is that when --enable-debug is true, than BUILDNUM and VERSION_DATE be set to 0 values. Since --enable-debug is only used by us developers, this doesn't break the usage of these values for packaging and releases.

cc @vashirov who probably has thoughts about this.

Firstyear:
Bug Description: ccache and sccache are unable to cache object
files in 389-ds due to the use of BUILDNUM that takes a current
time including minutes, and VERSION.sh using the current
date and git commit which can change between branches.

Fix Description: When using --enable-debug, BUILDNUM and VERSION
are set to 0 or "DEVELOPER BUILD". Since this is now static
object caching can now occuring, reducing developer recompile
times and allowing incremental compilation to work correctly.

fixes: #4454

Review by: ???

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,12 @@ QUOTE := $(NULLSTRING)"# a double quote"
 #
 
 PYTHON := python3
+if DEBUG
+# This allows sccache to work correctly with C files.
+BUILDNUM := "\"0000.000.0000\""
+else
 BUILDNUM := $(shell $(srcdir)/buildnum.py)
+endif
 NQBUILDNUM := $(subst \,,$(subst $(QUOTE),,$(BUILDNUM)))
 DEBUG_DEFINES = @debug_defs@
 DEBUG_CFLAGS = @debug_cflags@
--- a/VERSION.sh
+++ b/VERSION.sh
@@ -47,7 +47,7 @@ PACKAGE_VERSION=$VERSION_MAJOR.$VERSION_MINOR.${VERSION_MAINT}${VERSION_PREREL}
 PACKAGE_TARNAME=${brand}-ds-base
 # url for bug reports
 PACKAGE_BUGREPORT="${PACKAGE_BUGREPORT}enter_bug.cgi?product=$brand"
-PACKAGE_STRING="$PACKAGE_TARNAME $PACKAGE_VERSION"
+# PACKAGE_STRING="$PACKAGE_TARNAME $PACKAGE_VERSION"
 # the version of the ds console package that this directory server
 # is compatible with
 # console .2 is still compatible with 389 .3 for now
--- a/configure.ac
+++ b/configure.ac
@@ -7,16 +7,7 @@ AC_CONFIG_HEADERS([config.h])
 # include the version information
 . $srcdir/VERSION.sh
 AC_MSG_NOTICE(This is configure for $PACKAGE_TARNAME $PACKAGE_VERSION)
-AC_DEFINE_UNQUOTED([DS_PACKAGE_VERSION], "$PACKAGE_VERSION", [package version])
-AC_DEFINE_UNQUOTED([DS_PACKAGE_TARNAME], "$PACKAGE_TARNAME", [package tarball name])
-AC_DEFINE_UNQUOTED([DS_PACKAGE_BUGREPORT], "$PACKAGE_BUGREPORT", [package bug report url])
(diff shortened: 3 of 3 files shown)


Firstyear:
It looks like this change breaks ds_is_older, so I probably need to rethink the approach a little bit.

Firstyear:
Update fixes ds_is_older behaviour.

mreynolds389:
I need to do do more testing on Fedora, but on my laptop the build number is always the DEVELOPER BUILD  even if I do an optimized build.  Could something specific to my system so I need to try a Fedora build.  I'll have results tomorrow...

Firstyear:
If you have --enable-debug it makes it DEVELOPER BUILD. There are also some cases where certain files don't get refreshed even if you change config options, so make clean might clear this up for you.

Firstyear:
Yep, after a make clean it works for me - from config.log. Saying that now I'm not sure if I've botched the formatting there with the 2.0.1 not having a space to the date ... was that always the case? 

```
| #define DS_PACKAGE_VERSION "2.0.1.20201125git426fd157b"
| #define DS_PACKAGE_TARNAME "389-ds-base"
| #define DS_PACKAGE_BUGREPORT "http://bugzilla.redhat.com/enter_bug.cgi?product=389"
| #define DS_PACKAGE_STRING "389-ds-base 2.0.1.20201125git426fd157b"
| #define VERSION "2.0.1.20201125git426fd157b"
| #define PACKAGE "389-ds-base"
```

mreynolds389:
Yeah I blew away my build dir, and ran:

RANLIB=gcc-ranlib AR=gcc-ar CFLAGS='-flto -pipe -Wall -fPIC -fexceptions -fno-common -fstack-protector --param=ssp-buffer-size=4  -m64 -mtune=generic' CXXFLAGS='-fno-common -pipe -Wall -O2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic' ../389-ds-base/configure --enable-autobind --with-selinux --with-openldap --with-tmpfiles-d=/etc/tmpfiles.d --with-systemdsystemunitdir=/usr/lib/systemd/system --with-systemdsystemconfdir=/etc/systemd/system  --with-systemdgroupname=dirsrv.target --with-fhs --libdir=/usr/lib64 --with-systemd

But it still showed developer builld, but the Fedora build was correct, so ACK from me.

Firstyear:
@mreynolds389 I wonder if there is some build cache or even in your output directory something going on there too? I have a "make nuke" command that removes my /opt/dirsrv and build directory to avoid this, and that seemed to work for me ... .

PR #4466, opened 2020-11-26
Issue 4243 - Fix CI: SyncRepl plugin provides a wrong cookie

tbordaz:
Bug description:
	Individual testcase run fine but they fails when
	run in a raw

Fix description:
	Each testcase needs to do cleanup (at the end) or
        make initialization more robust

relates: https://github.com/389ds/389-ds-base/issues/4243

Reviewed by:

Platforms tested:  F31

### DIFF
--- a/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
+++ b/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
@@ -30,6 +30,85 @@ pytestmark = pytest.mark.tier1
 
 log = logging.getLogger(__name__)
 
+@pytest.fixture(scope="function")
+def init_sync_repl_plugins(topology, request):
+    """Prepare test environment (retroCL/sync_repl/
+    automember/memberof) and cleanup at the end of the test
+      1.: enable retroCL
+      2.: configure retroCL to log nsuniqueid as targetUniqueId
+      3.: enable content_sync plugin
+      4.: enable automember
+      5.: create (2) groups. Few groups can help to reproduce the concurrent updates problem.
+      6.: configure automember to provision those groups with 'member'
+      7.: enable and configure memberof plugin
+      8.: enable plugin log level
+      9.: restart the server
+      """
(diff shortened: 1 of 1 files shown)


Firstyear:
@droideck thoughts? I'm wondering how to structure this as a fixture but I think I'm not seeing it?

droideck:
I think this part can be a fixture:
https://github.com/389ds/389-ds-base/blob/db95b40b77790d7813b1edc03172399b31fb9664/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py#L271-L323

If I understand correctly.
But only for the last three tests:

- test_sync_repl_cookie
- test_sync_repl_cookie_add_del
- test_sync_repl_cookie_with_failure

tbordaz:
Thanks @droideck for your comments. If this part of code becomes a fixture, the cleanup as no longer access to group or am_config. How should it (cleanup) proceed ? Deleting all group/am_config returned by a search ?

droideck:
> Thanks @droideck for your comments. If this part of code becomes a fixture, the cleanup as no longer access to group or am_config. How should it (cleanup) proceed ? Deleting all group/am_config returned by a search ?

It's okay, you can add another finalizer to the fixture and it will be executed after the test case:

```
    def fin():
        for am_config in am_configs_cleanup:
            am_config.delete()
        for g in group:
            try:
                g.delete()
            except:
                pass

    request.addfinalizer(fin)
```

And the test case will have the finalizer for user only:

```
    def fin():
        inst.restart()
        for user in users_set:
            try:
                user.delete()
            except:
                pass

    request.addfinalizer(fin)
```

tbordaz:
Thanks @droideck for your help and advices. I moved initialization of the 3 testcases (test_sync_repl_cookie,test_sync_repl_cookie_add_del and  test_sync_repl_cookie_with_failure) into a fixture.

droideck (approved):
Looks good to me! Thank you!

PR #4467, opened 2020-11-27
Issue 4243 - Fix test: SyncRepl plugin provides a wrong cookie

tbordaz:
Bug description:
	The test_sync_repl_cookie_with_failure test case was incorrect.
	During a refreshPersistent search, a cookie is sent
	with the intermediate message that indicates the end of the refresh phase.
	Then a second cookie is sent on the updated entry (group10)
	I believed this test was successful some time ago but neither python-ldap
	nor sync_repl changed (intermediate sent in post refresh).
	So the testcase was never successful :(

Fix description:
	The fix is just to take into account the two expected cookies

relates: https://github.com/389ds/389-ds-base/issues/4243

Reviewed by:

Platforms tested: F31


mreynolds389:
LGTM

PR #4471, opened 2020-12-01
Issue 4284 - dsidm fails to delete an organizationalUnit entry

bsimonova:
Description:
Created test for dsidm organizationalunit delete and moved the function
check_value_in_log_and_reset() to __init__.py, because it will be used for
other dsidm tests.
Also modified the delete() function in lib389 to be able to delete the entry
without warning message for test purposes.

Relates: https://github.com/389ds/389-ds-base/issues/4284

Reviewed by: ???


droideck (approved):
LGTM!

Issue #4460, opened 2020-11-25
python3-lib389 v1.4.3 seems to ignore the configuration parameters from /etc/openldap/ldap.conf


aivanov389:
**Issue Description**
python3-lib389 v1.4.3 seems to ignore the configuration parameters from /etc/openldap/ldap.conf. python3-lib389 v1.4.3 works fine.

**Package Version and Platform:**
 - Platform: CentOS 8.2.2004
 - Package and version: python3-lib389-1.4.3.13-1.module_el8+10475+b74bca99.noarch

**Steps to Reproduce**
1. Configure 389ds of any version (tested with 1.3.x and 1.4.x) with SSL enabled, disable non-secure binds (nsslapd-require-secure-binds: on). Let's take this 389ds server's  certificate signed with a self-signed CA my_ca.crt. Let's call this server ldap.example.com

2. Make a clean install of CentOS 8 + updates (resulting in updated CentOS 8.2.2004) on another server (management.example.com).

**_All the following actions are performed only on management.example.com (we do not touch anything on ldap.example.com):_**

3. [root@management ~]# dnf -y install openldap-clients

4. Do not touch the file /etc/openldap/ldap.conf (empty config from rpm) or put the following to force the usage of system pem bundle:
TLS_CACERT     /etc/pki/tls/cert.pem
TLS_REQCERT     demand

5. Install 389ds python libs and management tools v1.4.2:
[root@management ~]# dnf -y module enable 389-directory-server:stable
[root@management ~]# dnf -y install 389-ds-base
[root@management ~]# rpm -qi python3-lib389 | grep Version
----->Version     : 1.4.2.16

[root@management ~]# ldapsearch -LLL -x -H ldaps://ldap.example.com:636 -D "cn=Directory Manager" -w mypassword -b '' -s base defaultnamingcontext
----->ldap_sasl_bind(SIMPLE): Can't contact LDAP server (-1)
_#This is ok since the ldap.example.com is signed by self-signed CA_

[root@management ~]# dsconf ldaps://ldap.example.com:636 -D "cn=Directory Manager" -w mypassword security get
----->Error: Can't contact LDAP server - error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)
_#This is ok since the ldap.example.com is signed by self-signed CA_
 
_#Now, install the self-signed CA to system bundle and check both connections (with ldapsearch and with dsconf) once again:_
[...]

PR #4472, opened 2020-12-03
Issue 4460 - BUG - add machine name to subject alt names in SSCA

Firstyear:
Bug Description: During SSCA creation, the server cert did not have
the machine name, which meant that the cert would not work without
reqcert = never.

Fix Description: Add the machine name as an alt name during SSCA
creation. It is not guaranteed this value is correct, but it
is better than nothing.

relates: https://github.com/389ds/389-ds-base/issues/4460

Review by: ???

### DIFF
--- a/src/lib389/lib389/instance/setup.py
+++ b/src/lib389/lib389/instance/setup.py
@@ -888,7 +888,7 @@ class SetupDs(object):
                         tlsdb_inst = NssSsl(dbpath=os.path.join(etc_dirsrv_path, dir))
                         tlsdb_inst.import_rsa_crt(ca)
 
-            csr = tlsdb.create_rsa_key_and_csr()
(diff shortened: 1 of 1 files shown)


mreynolds389:
LGTM since it was my fix :-)

droideck (approved):
Ack from me too:)

Firstyear:
> LGTM since it was my fix :-)

What are you saying, It took me ages to come up with this idea and write the code :P 

Anyway, jokes aside, thank you both for the quick review :)

Firstyear:
@mreynolds389 Should we apply this to anything in the 1.4.x series btw?

mreynolds389:
> @mreynolds389 Should we apply this to anything in the 1.4.x series btw?

Go all the way down to 1.4.2 please

PR #4474, opened 2020-12-04
Issue 4229 - RFE - Improve rust linking and build performance

### LINKED ISSUE #4229: Rust - clean up libsds linking with librslapd
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/51176
- Created at 2020-06-23 04:01:56 by [firstyear](https://pagure.io/user/firstyear) (@Firstyear)
- Assigned to nobody

---

#### Issue Description
We may no longer require librsds as it can be part of libslapd instead. This removes one extra .so and makes linking simpler, preferring the rust module system.

Firstyear:
Bug Description: Due to changes in how we approach rust in our
make system, we can improve this significantly to reduce complexity
in our linking, and to remove a large quantity of deadcode that
is no longer needed.

Fix Description: Remove older parts of sds (removed in favour
of rust datastructures and concread), and remove lfds which
is no longer used from nunc-stans

fixes: https://github.com/389ds/389-ds-base/issues/4229

Review by: ???


Firstyear:
This would be "nice to have" for 1.4.4 for us since we are going to enable rust in 1.4.4 at suse, but I can also understand if you don't want to back port this to that version and we leave it as 2.0.0 only :)

mreynolds389:
1.4.4 should be fine

lslebodn:
@Firstyear there is a following code at the end of Makefile. Is it still valid? Could it be removed?
```
# The rm in man3 is to remove files like: _home_william_development_389ds_libsds_src_.3
# If there is a way to ignore this in doxygen I'm all ears ...

doxyfile.stamp:
	cd $(srcdir); $(DOXYGEN) $(abs_top_builddir)/docs/slapi.doxy
	rm -f $(abs_top_builddir)/man/man3/_*
	touch doxyfile.stamp
```

Firstyear:
> @Firstyear there is a following code at the end of Makefile. Is it still valid? Could it be removed?
> 
> ```
> # The rm in man3 is to remove files like: _home_william_development_389ds_libsds_src_.3
> # If there is a way to ignore this in doxygen I'm all ears ...
> 
> doxyfile.stamp:
> 	cd $(srcdir); $(DOXYGEN) $(abs_top_builddir)/docs/slapi.doxy
> 	rm -f $(abs_top_builddir)/man/man3/_*
> 	touch doxyfile.stamp
> ```

Honestly, I think we could probably remove doxygen from the build because none of us use it's output ... but that may be up to @mreynolds389 and the rest of the team to have a say in ...

Firstyear:
> 1.4.4 should be fine

Not sure if this is an ack, so assuming you haven't reviewed yet :)

Firstyear:
```
   5d77fe605..60fb9fa82  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

PR #4476, opened 2020-12-04
Issue 4419 - Warn users of skipped entries during ldif2db online import

### LINKED ISSUE #4419: ldif2db - online. Warn the user of skipped entries
**Issue Description**
ldif2db is used to import directory entries from an ldif file to a directory server instance. During the import operation, entries might not conform to various constraints (attribute syntax, schema violation, etc) and be skipped. On completion of the import operation the user receives a success message, with no indication of a skipped entry, apart from a message in the error logs. 

 - Platform: 5.6.6-300.fc32.x86_64
 - Package and version: 389-Directory/1.4.4.4

**Steps to Reproduce**
1. Verify syntax checking is enabled (default)
2. Generate sample ldif file
3. Modify the ldif file to violate a syntax checking rule (An empty fax number)
4. Import ldif file, notice import success message
5. Check the server error logs:
```c
ERR - slapi_entry_syntax_check - "uid=OJeleniewski1000,ou=product development,dc=example,dc=com": (facsimileTelephoneNumber) value #0 invalid per syntax
WARN - import_producer - import userroot: Skipping entry "uid=OJeleniewski1000,ou=product development,dc=example,dc=com" which violates attribute syntax,
```

**Expected results**
The user should get a warning message saying some entries were skipped and the that they should check the logs for more info.

**Additional context**
Parent bug: https://bugzilla.redhat.com/show_bug.cgi?id=1627645
ldif2db can operate in both offline and online modes, the parent bug will be divided into offline and online updates that each will require a different approach.

jchapma:
Bug Description:  During an online ldif2db import entries that do not
                  conform to various constraints will be skipped and
                  not imported. On completition of an import with skipped
                  entries, the server responds with a success message
                  and logs the skipped entry detail to the error logs.
                  The success messgae could lead the user to believe
                  that all entries were successfully imported.

Fix Description:  If a skipped entry occurs during import, the import
                  will continue and a warning message will be displayed.
                  The schema is extended with a nsTaskWarning attribute
                  which is used to capture and retrieve any task
                  warnings.

                  CLI tools for online import updated.

                  Test added to generate an incorrect ldif entry and perform an
                  online import.

Fixes: https://github.com/389ds/389-ds-base/issues/4419

Reviewed by: tbordaz, mreynolds389  (Thanks)

### DIFF
--- a/dirsrvtests/tests/suites/import/import_test.py
+++ b/dirsrvtests/tests/suites/import/import_test.py
@@ -65,6 +65,9 @@ def _import_clean(request, topo):
         import_ldif = ldif_dir + '/basic_import.ldif'
         if os.path.exists(import_ldif):
             os.remove(import_ldif)
+        syntax_err_ldif = ldif_dir + '/syntax_err.dif'
+        if os.path.exists(syntax_err_ldif):
+            os.remove(syntax_err_ldif)
 
     request.addfinalizer(finofaci)
 
@@ -141,17 +144,19 @@ def _create_bogus_ldif(topo):
 
 def _create_syntax_err_ldif(topo):
     """
-    Create an incorrect ldif entry that violates syntax check
+    Create an ldif file, which contains an entry that violates syntax check
     """
     ldif_dir = topo.standalone.get_ldif_dir()
     line1 = """dn: dc=example,dc=com
 objectClass: top
 objectClass: domain
 dc: example
+
 dn: ou=groups,dc=example,dc=com
 objectClass: top
 objectClass: organizationalUnit
 ou: groups
+
 dn: uid=JHunt,ou=groups,dc=example,dc=com
 objectClass: top
 objectClass: person
@@ -201,6 +206,34 @@ def test_import_with_index(topo, _import_clean):
     assert f'{place}/userRoot/roomNumber.db' in glob.glob(f'{place}/userRoot/*.db', recursive=True)
 
 
+def test_online_import_with_warning(topo, _import_clean):
+    """
+    Import an ldif file with syntax errors, verify skipped entry warning code
+
+    :id: 5bf75c47-a283-430e-a65c-3c5fd8dbadb8
+    :setup: Standalone Instance
+    :steps:
+        1. Create standalone Instance
+        2. Create an ldif file with an entry that violates syntax check (empty givenname)
+        3. Online import of troublesome ldif file
+    :expected results:
+        1. Successful import with skipped entry warning
+        """
+    topo.standalone.restart()
+
+    import_task = ImportTask(topo.standalone)
+    import_ldif1 = _create_syntax_err_ldif(topo)
+
+    # Importing  the offending ldif file - online
+    import_task.import_suffix_from_ldif(ldiffile=import_ldif1, suffix=DEFAULT_SUFFIX)
+
+    # There is just  a single entry in this ldif
+    import_task.wait(5)
+
+    # Check for the task nsTaskWarning attr, make sure its set to skipped entry code
+    assert import_task.present('nstaskwarning')
+    assert TaskWarning.WARN_SKIPPED_IMPORT_ENTRY == import_task.get_task_warn()
+
 def test_crash_on_ldif2db(topo, _import_clean):
     """
     Delete the cn=monitor entry for an LDBM backend instance. Doing this will
@@ -246,7 +279,7 @@ def test_ldif2db_allows_entries_without_a_parent_to_be_imported(topo, _import_cl
     topo.standalone.start()
 
 
-def test_ldif2db_syntax_check(topo):
+def test_ldif2db_syntax_check(topo, _import_clean):
     """ldif2db should return a warning when a skipped entry has occured.
     :id: 85e75670-42c5-4062-9edc-7f117c97a06f
     :setup:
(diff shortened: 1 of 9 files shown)


dirsrvtests/tests/suites/import/import_test.py
```diff
@@ -246,7 +277,7 @@ def test_ldif2db_allows_entries_without_a_parent_to_be_imported(topo, _import_cl
     topo.standalone.start()
 
 
-def test_ldif2db_syntax_check(topo):
+def test_ldif2db_syntax_check(topo, _import_clean):
```
jchapma, line 282: Added _import_clean  function to clean up generated syntax error ldif file used for ldif2db offline test.

droideck:
I see that the test case fails with this PR:

```
    def test_fast_slow_import(topo, _toggle_private_import_mem, _import_clean):
...
        # Let's set nsslapd-db-private-import-mem:on, nsslapd-import-cache-autosize: 0
        config = LDBMConfig(topo.standalone)
        # Measure offline import time duration total_time1
        total_time1 = _import_offline(topo, 20)
        # Now nsslapd-db-private-import-mem:off
        config.replace('nsslapd-db-private-import-mem', 'off')
        accounts = Accounts(topo.standalone, DEFAULT_SUFFIX)
        for i in accounts.filter('(uid=*)'):
            UserAccount(topo.standalone, i.dn).delete()
        # Measure offline import time duration total_time2
        total_time2 = _import_offline(topo, 20)
        # total_time1 < total_time2
>       assert total_time1 < total_time2
E       assert 2.4483795166015625 < 2.287022590637207
```

It is run without ASAN and I've repeated the test suite a few times on my VM and it was 100% reproducible. 
389-ds-base: 2.0.1-20201207git33d4ffae7.fc32
py.test -s -v dirsrvtests/tests/suites/import/import_test.py

mreynolds389:
> I see that the test case fails with this PR:
>
>    def test_fast_slow_import(topo, _toggle_private_import_mem, _import_clean):> 
> ...
> 
> It is run without ASAN and I've repeated the test suite a few times on my VM and it was 100% reproducible.
> 389-ds-base: 2.0.1-20201207git33d4ffae7.fc32
> py.test -s -v dirsrvtests/tests/suites/import/import_test.py

I don't see how this PR could impact import performance, weird...

Firstyear:
> I see that the test case fails with this PR:
> 
> ```
>     def test_fast_slow_import(topo, _toggle_private_import_mem, _import_clean):
> ...
>         # Let's set nsslapd-db-private-import-mem:on, nsslapd-import-cache-autosize: 0
>         config = LDBMConfig(topo.standalone)
>         # Measure offline import time duration total_time1
>         total_time1 = _import_offline(topo, 20)
>         # Now nsslapd-db-private-import-mem:off
>         config.replace('nsslapd-db-private-import-mem', 'off')
>         accounts = Accounts(topo.standalone, DEFAULT_SUFFIX)
>         for i in accounts.filter('(uid=*)'):
>             UserAccount(topo.standalone, i.dn).delete()
>         # Measure offline import time duration total_time2
>         total_time2 = _import_offline(topo, 20)
>         # total_time1 < total_time2
> >       assert total_time1 < total_time2
> E       assert 2.4483795166015625 < 2.287022590637207
> ```
> 
> It is run without ASAN and I've repeated the test suite a few times on my VM and it was 100% reproducible.
> 389-ds-base: 2.0.1-20201207git33d4ffae7.fc32
> py.test -s -v dirsrvtests/tests/suites/import/import_test.py

It's probably not a great idea to measure timing in tests anyway, because any other system noise *could* cause easy false negatives here. 

> I don't see how this PR could impact import performance, weird...

```
if((job) && (job->skipped)) {
```

I wonder if it's to do with this line. `job->skipped` is a counter, so the moment it's non-zero, we'll always attempt to set the `WARN_SKIPPED_IMPORT_ENTRY` even if it's already set. That involves a deref + fn call + OR + store, so if we get the warning early followed by many other entries we'll be hitting this function every iteration, even if on *this* iteration we didn't actually need to add the warning.

jchapma:
This PR checks the job->skipped counter after the ldif file parsing has complete, this counter is accessed once per ldif import, so it shouldn't add too much overhead.

Firstyear:
> This PR checks the job->skipped counter after the ldif file parsing has complete, this counter is accessed once per ldif import, so it shouldn't add too much overhead.

Ahh my bad, I must have misunderstood something then :)

jchapma:
I would like to know the reason for the test time differences though, let me have a look at it .

jchapma:
The test_fast_slow_import() test compares the duration of ldif import times when the **nsslapd-db-private-import-mem** config is enabled and disabled, the enabled setting being the most efficient. 

On my setup this test fails 40% of the time. This test uses dbgen_users() to create a random ldif file for each import, each with the same number of users, but with differing user details, family name, given name, etc. I noticed that the test fails when the ldif file used for the first import is larger than the ldif file used for the second import. I guess the overhead of parsing a larger ldif file could be the responsible for the mysterious time difference.

Firstyear:
> The test_fast_slow_import() test compares the duration of ldif import times when the **nsslapd-db-private-import-mem** config is enabled and disabled, the enabled setting being the most efficient.
> 
> On my setup this test fails 40% of the time. This test uses dbgen_users() to create a random ldif file for each import, each with the same number of users, but with differing user details, family name, given name, etc. I noticed that the test fails when the ldif file used for the first import is larger than the ldif file used for the second import. I guess the overhead of parsing a larger ldif file could be the responsible for the mysterious time difference. Sometimes the generated ldif files differ by 400 bytes.

Would it be worth re-using the same ldif instead to make this more reliable?

jchapma:
I reckon it would, this test is not part of this PR so I should create another PR for it I guess.

Firstyear:
Yep, agreed. In that case is there any reason to not-merge this @droideck ?

droideck (approved):
Seems reasonable, you have my ack then! Thanks!:)

jchapma:
Thanks for your feedback everyone.

Issue #4381, opened 2020-10-15
LDAPI autobind: authenticate as target DN


tiran:
**Is your feature request related to a problem? Please describe.**

**UPDATE** tl;dr I would like to have the uid/gid mapping in a local, non-replicated DB and the effective user in a shared, replicated DIT. This improvement would allow FreeIPA to use autobind for services without pre-allocated and reserved uids across all servers in a cluster. See FreeIPA ticket https://pagure.io/freeipa/issue/8544#comment-696374 for more information.

**Describe the solution you'd like**

LDAPI autobind allows to authenticate by mapping the effective UID and GID of a process to an LDAP entry. Some UIDs and GIDs for system services are hard-coded and preallocated. Others are assigned randomly when a service is installed The Fedora project is reluctant to grant UIDs any more. The range of allocated UIDs is very limited. UID/GID may also be different on others OSes (e.g. Debian).

Therefore I like to request an extension to [autobind feature](https://directory.fedoraproject.org/docs/389ds/FAQ/ldapi-and-autobind.html). When a search for ``(&(uidNumber=<uid>)(gidNumber=<gid>))`` return one entry and the entry contains a specific attribute of syntax type DN, the bind as the value of that attribute.

Example:

```
dn: cn=config
nsslapd-ldapientrysearchbase: cn=autobind,cn=config
nsslapd-ldapiauthenticateasdntype: authenticateAsDN

dn: cn=bind-dyndb-ldap,cn=autobind,cn=config
cn: bind-dyndb-ldap
uidNumber: 25
gidNumber: 25
authenticateAsDN: krbprincipalname=DNS/server.ipa.example@IPA.EXAMPLE,cn=services,cn=accounts,dc=ipa,dc=example
```

An LDAPI connect for ``25:25`` would bind as user ``krbprincipalname=DNS/server.ipa.example@IPA.EXAMPLE,cn=services,cn=accounts,dc=ipa,dc=example`` instead of ``cn=bind-dyndb-ldap,cn=autobind,cn=config``. If the target entry does not exist, then authentication would fail. If an entry does not have ``authenticateAsDN`` then bind as the direct entry instead.

**Describe alternatives you've considered**
Alternative would be proper support for referential integrity and memberof for cn=config backend. I assume that this would be harder to implement for you *and* less useful for FreeIPA.

**Additional context**
n/a

Firstyear:
" Others are assigned randomly when a service is installed"

[...]

PR #4477, opened 2020-12-05
Issue 4381 - RFE - LDAPI authentication DN rewritter

mreynolds389:
Description:  

Create a new LDAPI configuration attribute to specify a DN syntax attribute that can be added to an LDAP entry that will cause the BIND to use the DN set in that attribute.

For more information see:

https://www.port389.org/docs/389ds/design/ldapi-auto-auth-dn-design.html

Relates: https://github.com/389ds/389-ds-base/issues/4381

### DIFF
--- a/dirsrvtests/tests/suites/ldapi/ldapi_test.py
+++ b/dirsrvtests/tests/suites/ldapi/ldapi_test.py
@@ -1,0 +1,154 @@
+import logging
+import pytest
+import os
+import subprocess
+from lib389._constants import DEFAULT_SUFFIX, DN_DM
+from lib389.idm.user import UserAccounts
+from lib389.ldapi import LDAPIMapping, LDAPIFixedMapping
+from lib389.topologies import topology_st as topo
+from lib389.tasks import LDAPIMappingReloadTask
+
+
+def test_ldapi_authdn_attr_rewrite(topo, request):
+    """Test LDAPI Authentication DN mapping feature
+
+    :id: e8d68979-4b3d-4e2d-89ed-f9bad827718c
+    :setup: Standalone Instance
+    :steps:
+        1. Set LDAPI configuration
+        2. Create LDAP user
+        3. Create OS user
+        4. Create entries under cn=config for auto bind subtree and mapping entry
+        5. Do an LDAPI ldapsearch as the OS user
+        6. OS user was mapped expected LDAP entry
+        7. Do search using root & LDAPI
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Success
+        7. Success
+    """
+
+    LINUX_USER = "ldapi_test_lib389_user"
+    LINUX_USER2 = "ldapi_test_lib389_user2"
+    LINUX_USER3 = "ldapi_test_lib389_user3"
+    LINUX_PWD = "5ecret_137"
+    LDAP_ENTRY_DN = "uid=test_ldapi,ou=people,dc=example,dc=com"
+    LDAP_ENTRY_DN2 = "uid=test_ldapi2,ou=people,dc=example,dc=com"
+    LDAP_ENTRY_DN3 = "uid=test_ldapi3,ou=people,dc=example,dc=com"
+    LDAPI_AUTH_CONTAINER = "cn=auto_bind,cn=config"
+
+    def fin():
+        # Remove the OS users
+        for user in [LINUX_USER, LINUX_USER2, LINUX_USER3]:
+            try:
+                subprocess.run(['userdel', '-r', user])
+            except:
+                pass
+    request.addfinalizer(fin)
+
+    # Must be root
+    if os.geteuid() != 0:
+        return
+
+    # Perform config tasks
+    topo.standalone.config.set('nsslapd-accesslog-logbuffering', 'off')
+    topo.standalone.config.set('nsslapd-ldapiDNMappingBase', 'cn=auto_bind,cn=config')
+    topo.standalone.config.set('nsslapd-ldapimaptoentries', 'on')
+    topo.standalone.config.set('nsslapd-ldapiuidnumbertype', 'uidNumber')
+    topo.standalone.config.set('nsslapd-ldapigidnumbertype', 'gidNumber')
+    ldapi_socket_raw = topo.standalone.config.get_attr_val_utf8('nsslapd-ldapifilepath')
+    ldapi_socket = ldapi_socket_raw.replace('/', '%2F')
+
+    # Create LDAP users
+    users = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    user_properties = {
+        'uid': 'test_ldapi',
+        'cn': 'test_ldapi',
+        'sn': 'test_ldapi',
+        'uidNumber': '2020',
+        'gidNumber': '2020',
+        'userpassword': 'password',
+        'description': 'userdesc',
+        'homeDirectory': '/home/test_ldapi'}
+    users.create(properties=user_properties)
+
+    user_properties = {
+        'uid': 'test_ldapi2',
+        'cn': 'test_ldapi2',
+        'sn': 'test_ldapi2',
+        'uidNumber': '2021',
+        'gidNumber': '2021',
+        'userpassword': 'password',
+        'description': 'userdesc',
+        'homeDirectory': '/home/test_ldapi2'}
+    users.create(properties=user_properties)
+
+    user_properties = {
+        'uid': 'test_ldapi3',
+        'cn': 'test_ldapi3',
+        'sn': 'test_ldapi3',
+        'uidNumber': '2023',
+        'gidNumber': '2023',
+        'userpassword': 'password',
+        'description': 'userdesc',
+        'homeDirectory': '/home/test_ldapi3'}
+    users.create(properties=user_properties)
+
+    # Create OS users
+    subprocess.run(['useradd', '-u', '5001', '-p', LINUX_PWD, LINUX_USER])
+    subprocess.run(['useradd', '-u', '5002', '-p', LINUX_PWD, LINUX_USER2])
+
+    # Create some mapping entries
+    ldapi_mapping = LDAPIMapping(topo.standalone, LDAPI_AUTH_CONTAINER)
+    ldapi_mapping.create_mapping(name='entry_map1', username='dummy1',
+                                 ldap_dn='uid=dummy1,dc=example,dc=com')
+    ldapi_mapping.create_mapping(name='entry_map2', username=LINUX_USER,
+                                 ldap_dn=LDAP_ENTRY_DN)
+    ldapi_mapping.create_mapping(name='entry_map3', username='dummy2',
+                                 ldap_dn='uid=dummy3,dc=example,dc=com')
+
+    # Restart server for config to take effect, and clear the access log
+    topo.standalone.deleteAccessLogs(restart=True)
+
+    # Bind as OS user using ldapsearch
+    ldapsearch_cmd = f'ldapsearch -b \'\' -s base -Y EXTERNAL -H ldapi://{ldapi_socket}'
+    os.system(f'su {LINUX_USER} -c "{ldapsearch_cmd}"')
+
+    # Check access log
+    assert topo.standalone.ds_access_log.match(f'.*AUTOBIND dn="{LDAP_ENTRY_DN}".*')
+
+    # Bind as Root DN just to make sure it still works
+    assert os.system(ldapsearch_cmd) == 0
+    assert topo.standalone.ds_access_log.match(f'.*AUTOBIND dn="{DN_DM}".*')
+
+    # Create some fixed mapping
+    ldapi_fixed_mapping = LDAPIFixedMapping(topo.standalone, LDAPI_AUTH_CONTAINER)
+    ldapi_fixed_mapping.create_mapping("fixed", "5002", "5002", ldap_dn=LDAP_ENTRY_DN2)
+    topo.standalone.deleteAccessLogs(restart=True)
+
+    # Bind as OS user using ldapsearch
+    os.system(f'su {LINUX_USER2} -c "{ldapsearch_cmd}"')
+
+    # Check access log
+    assert topo.standalone.ds_access_log.match(f'.*AUTOBIND dn="{LDAP_ENTRY_DN2}".*')
+
+    # Add 3rd user, and test reload task
+    subprocess.run(['useradd', '-u', '5003', '-p', LINUX_PWD, LINUX_USER3])
+    ldapi_fixed_mapping.create_mapping("reload", "5003", "5003", ldap_dn=LDAP_ENTRY_DN3)
+
+    reload_task = LDAPIMappingReloadTask(topo.standalone).create()
+    reload_task.wait(timeout=20)
+
+    os.system(f'su {LINUX_USER3} -c "{ldapsearch_cmd}"')
+    assert topo.standalone.ds_access_log.match(f'.*AUTOBIND dn="{LDAP_ENTRY_DN3}".*')
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
--- a/ldap/schema/30ns-common.ldif
+++ b/ldap/schema/30ns-common.ldif
@@ -56,6 +56,8 @@ attributeTypes: ( nsClassname-oid NAME 'nsClassname' DESC 'Netscape defined attr
 attributeTypes: ( 2.16.840.1.113730.3.1.2337 NAME 'nsCertSubjectDN' DESC 'An x509 DN from a certificate used to map during a TLS bind process' SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 X-ORIGIN '389 Directory Server Project' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2342 NAME 'nsSshPublicKey' DESC 'An nsSshPublicKey record' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 X-ORIGIN '389 Directory Server Project' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2343 NAME 'legalName' DESC 'An individuals legalName' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN '389 Directory Server Project' )
+attributeTypes: ( 2.16.840.1.113730.3.1.2376 NAME 'nsslapd-authenticateAsDN' DESC 'LDAPI mapping DN' SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.2377 NAME 'nsslapd-ldapiUsername' DESC 'LDAPI mapping system username' SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 objectClasses: ( nsAdminDomain-oid NAME 'nsAdminDomain' DESC 'Netscape defined objectclass' SUP organizationalUnit MAY ( nsAdminDomainName ) X-ORIGIN 'Netscape' )
 objectClasses: ( nsHost-oid NAME 'nsHost' DESC 'Netscape defined objectclass' SUP top MUST ( cn ) MAY ( serverHostName $ description $ l $ nsHostLocation $ nsHardwarePlatform $ nsOsVersion ) X-ORIGIN 'Netscape' )
 objectClasses: ( nsAdminGroup-oid NAME 'nsAdminGroup' DESC 'Netscape defined objectclass' SUP top MUST ( cn ) MAY ( nsAdminGroupName $ description $ nsConfigRoot $ nsAdminSIEDN ) X-ORIGIN 'Netscape' )
@@ -75,3 +77,5 @@ objectClasses: ( 2.16.840.1.113730.3.2.334 NAME 'nsOrgPerson' DESC 'A representa
   mobile $ o $ pager $ photo $ roomNumber $ uid $ userCertificate $ telephoneNumber $
   x500uniqueIdentifier $ userSMIMECertificate $ userPKCS12 )
   X-ORIGIN '389 Directory Server Project' )
+objectClasses: ( 2.16.840.1.113730.3.2.338 NAME 'nsLDAPIAuthMap' DESC 'LDAPI authentication DN mapping' SUP top MUST ( cn $ nsslapd-ldapiUsername $ nsslapd-authenticateAsDN ) MAY ( description ) X-ORIGIN '389 Directory Server' )
+objectClasses: ( 2.16.840.1.113730.3.2.339 NAME 'nsLDAPIFixedAuthMap' DESC 'LDAPI fixed authentication DN mapping' SUP top MUST ( cn $ uidNumber $ gidNumber $ nsslapd-authenticateAsDN ) MAY ( description ) X-ORIGIN '389 Directory Server' )
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -56,10 +56,6 @@
 #include <ssl.h>
 #include "fe.h"
 
-#if defined(ENABLE_LDAPI)
-#include "getsocketpeer.h"
-#endif /* ENABLE_LDAPI */
-
 #if defined(LDAP_IOCP)
 #define SLAPD_WAKEUP_TIMER 250
 #else
@@ -988,6 +984,7 @@ slapd_daemon(daemon_ports_t *ports)
             }
             listeners++;
         }
+        initialize_ldapi_auth_dn_mappings(LDAPI_STARTUP);
     }
 #endif /* ENABLE_LDAPI */
 
@@ -1135,6 +1132,11 @@ slapd_daemon(daemon_ports_t *ports)
      */
     connection_table_free(the_connection_table);
     the_connection_table = NULL;
+
+#if defined(ENABLE_LDAPI)
+    /* Free LDAPI mappings */
+    free_ldapi_auth_dn_mappings(LDAPI_SHUTDOWN);
+#endif
 
     if (!in_referral_mode) {
         /* Close SNMP collator after the plugins closed...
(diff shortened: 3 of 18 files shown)


dirsrvtests/tests/suites/ldapi/ldapi_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Create OS user
+    subprocess.run(['useradd', '-p', LINUX_PWD, LINUX_USER ])
+
+    # Get UID GID
+    uid = getpwnam(LINUX_USER)[2]
+    gid = grp.getgrnam(LINUX_USER)[2]
+
+    # Create subtree to store LDAPI mappings under cn=config
+    config = DSLdapObject(topo.standalone, dn="cn=auto_bind,cn=config")
```
Firstyear: If this is could to be an auto_bind uid/gid mapping type we should probably give it a real objectClass and a dsldapobject subclass.

dirsrvtests/tests/suites/ldapi/ldapi_test.py
```diff
@@ -0,0 +1,114 @@
...
+                  basedn="cn=config",
+                  properties={'cn': 'auto_bind'})
+
+    # Create mapping entry
+    config = DSLdapObject(topo.standalone, dn="cn=entry_map1,cn=auto_bind,cn=config")
+    config._create_objectclasses = ['top', 'extensibleObject']
+    config.create(rdn="cn",
+                  basedn="cn=auto_bind,cn=config",
+                  properties={'cn': 'entry_map1',
+                              'uidNumber': str(uid),
```
Firstyear: This may have issues if you run the test under root depending on the order of mapping precedence?

ldap/servers/slapd/daemon.c
```diff
@@ -1901,21 +1896,27 @@ slapd_bind_local_user(Connection *conn)
...
-                            auth_dn);
+                        char *auth_dn = NULL;
+                        char *authDNAtttr = config_get_ldapi_map_dn_attr();
+
+                        if (authDNAtttr) {
+                            /*
+                             * We configured a dn mapping attribute, now check
+                             * for it in the entry
+                             */
+                            auth_dn = slapi_entry_attr_get_charptr(entries[0], authDNAtttr);
```
Firstyear: Doesn't this rely on our ldapi search base being all backends including the config be's?

Firstyear (changes requested):
One thing I'm worried about here in this design is systemd dynamic users. Right now we assume that a dn maps to a uid/gid number, but really it actually maps a dn to a local system uid which we should translate to a uidNumber. It's critical to understand in this system - dynamic users do NOT guarantee the stability of a uid/gid number assignment to a uid. As a result, this means that with this design, any service that would change to dynamic users would potentially not function if the uidnumber changed, causing a different service to be able to impersonate them.

As a result, I think what needs to occur here is that we only configure the uid (name) of the account we want to remap into the associated dn. At start up we read all the auto bind config entries, and resolve the uid (name) to it's uidnumber and primary gid number. We can then refer to this map for the rest of execution.

This also resolves the issue where if ldapisearchbase is *not* set to all backends, that the auto bind maps would not be read. IE the ldapisearchbase could be set to the cn=account container in IPA, but the autobind maps would still be able to function since the auto bind maps would NOT be in entries, but in an in memory mapping. 

For more see http://0pointer.net/blog/dynamic-users-with-systemd.html

ldap/servers/slapd/daemon.c
mreynolds389: No, it always/only uses the configured LDAPI search base.  As it has always been, you have to set a single search base if you want to map entries.  It never checked "all" backends, you have to set a single backend.  So the code uses the search base to find entries based on the LDAP configuration (nsslapd-ldapiEntrySearchBase, nsslapd-ldapiUIDNumberType, etc).  Once we find entries we can change the DN to whatever backend we want (if nsslapd-ldapiauthdnattr is set), and everything works correctly.

Now your concern about dynamic users is valid, but we don't support them today.  Your proposal, which I like very much, would require a complete redesign of LDAPI entry mapping.  LDAPI, as of today, is based on static/hardcoded uidNumber/gidNumbers in entries.  This PR is still compliant with that approach of mapping uid/gid to some entry, but to do what you are suggesting would require rewriting LDAPI mapping code.  

IMHO this PR is very lightweight, it's back-portable, and does not change the behavior of how LDAPI works or is configured.  So what I think we need is an "LDAPI V2" that has its own/extended configuration, because we can't break/change how LDAPI V1 works today for customers who are actually using it and expect it to behave a certain way.

So...  I think this PR is still valid and is a good solution for the use case (which is needed now).  Then in 2.1.0 we can add dynamic user support (LDAPI-V2), and deprecate LDAPI-V1.  This allows customers to migrate to the newer design when they are ready.  

Thoughts?

dirsrvtests/tests/suites/ldapi/ldapi_test.py
mreynolds389: Actually the test requires you run it as root :-)  Anyway I don't see a need for mapping precedence, since the server is doing an internal search that should only return a single entry.  If I am missing your point, please elaborate...

dirsrvtests/tests/suites/ldapi/ldapi_test.py
mreynolds389: Perhaps the Config class could be extended so we can add entries to it instead of creating a new class?  

And yes, I can add a new objectclass/attributes for these config entries under "cn=auto_bind,cn=config".

mreynolds389:
Added the schema, and did some cleanup....

dirsrvtests/tests/suites/ldapi/ldapi_test.py
Firstyear: Nah, we need a new dsldapobject class so we can expose search/list/create/get from the cli. And that also is part of why an oc is needed to gen the correct filters for this.

ldap/schema/30ns-common.ldif
```diff
@@ -75,3 +76,4 @@ objectClasses: ( 2.16.840.1.113730.3.2.334 NAME 'nsOrgPerson' DESC 'A representa
   mobile $ o $ pager $ photo $ roomNumber $ uid $ userCertificate $ telephoneNumber $
   x500uniqueIdentifier $ userSMIMECertificate $ userPKCS12 )
   X-ORIGIN '389 Directory Server Project' )
+objectClasses: ( 2.16.840.1.113730.3.2.338 NAME 'nsLDAPIAuthMap' DESC 'LDAPI authentication DN mapping' SUP top MUST ( uidNumber $ gidNumber $ nsslapd-authenticateAsDN ) MAY ( cn ) X-ORIGIN '389 Directory Server' )
```
Firstyear: shouldn't this be must cn for the rdn in cn=config?

dirsrvtests/tests/suites/ldapi/ldapi_test.py
Firstyear: Well I think my point is more that in the test it "breaks" root ldapi, but I think our test env doesn't actually rely on ldapi for root as we use the DM pw and ldap via localhost, so it's probably actually a non-concern.

Firstyear (commented):
Github not letting me add the response to the conversation here since it's "outdated" so moving here. 

> No, it always/only uses the configured LDAPI search base. As it has always been, you have to set a single search base if you want to map entries. It never checked "all" backends, you have to set a single backend. So the code uses the search base to find entries based on the LDAP configuration (nsslapd-ldapiEntrySearchBase, nsslapd-ldapiUIDNumberType, etc). Once we find entries we can change the DN to whatever backend we want (if nsslapd-ldapiauthdnattr is set), and everything works correctly.

So line 1902 you get `CONFIG_LDAPI_AUTH_DN_ATTRIBUTE` from the entry in `entries`. entries is populated by the search on 1871, which uses `base_dn` from `config_get_ldapi_search_base_dn`. This would mean that if the ldapi search base *doesn't* include cn=config, then your mapping entries would never be found and used in the process. And part of the "value" of this change is "per-server" uid/gid maps for local/os specific uid bindings. I think that's what my concern here is. 

If in your test case, you define the ldapisearchbase to DEFAULT_SUFFIX does it still pass? 

> 
> Now your concern about dynamic users is valid, but we don't support them today. Your proposal, which I like very much, would require a complete redesign of LDAPI entry mapping. LDAPI, as of today, is based on static/hardcoded uidNumber/gidNumbers in entries. This PR is still compliant with that approach of mapping uid/gid to some entry, but to do what you are suggesting would require rewriting LDAPI mapping code.

Yes and no - I think that the current ldapi mapping code actually is not a problem. Because the users that exist in the LDAP database also is the source of truth *for* those users with regard to to their uid/gidnumber. The point of this PR for IPA is about users who exist *outside* of our source of truth, as in system service accounts which can be per server or per distro, or even per-invocation of the systemd unit file.

I'm suggesting that *only* your nsLDAPIAuthMap configurations do this dynamic mapping because this PR really only exists to allow a dynamic mapping for per-server uids to an ldap entry. That limits the amount of change to this PR and still keeps it backportable without changing the main ldapi mapping code (which honestly, it quite fine as it currently exists because as mentioned, for those users we *are* the source of truth). 

Worth remembering too that os services *may* adopt this systemd dynamic user scheme without our or IPA's knowledge, which means this could silently/subtely break on us, so this really is a case of "do it right once" else we'll have to come back and do it again.

EDIT: Another use case is uid/gids that are assigned on package install. This could cause the ldap config on a restore from backup to break. For example, imagine you install the server where the bind dns user is uid 30, then we set the cn=config ldapi map to uidNumber: 30. We have to restore from a backup, and when we re-install the host the bind dns user is assigned uid 40. Now when we restore the mapping is broken. So either then the restore process has to do a fixup on every uidNumber/gidNumber map, or that situation is overlooked (and it probably has been) and that means that on a restore the system  just is hosed in silently brilliant ways. Where as if the map took it's info from uid and resolved it once on startup, because the config map has "uid: binddns" or whatever it is, it would get the right uidNumber/gidNumber at start up even if those values have changed. 

> 
> IMHO this PR is very lightweight, it's back-portable, and does not change the behavior of how LDAPI works or is configured. So what I think we need is an "LDAPI V2" that has its own/extended configuration, because we can't break/change how LDAPI V1 works today for customers who are actually using it and expect it to behave a certain way.
> 
> So... I think this PR is still valid and is a good solution for the use case (which is needed now). Then in 2.1.0 we can add dynamic user support (LDAPI-V2), and deprecate LDAPI-V1. This allows customers to migrate to the newer design when they are ready.
> 
> Thoughts?

Many, as always :)

mreynolds389:
Okay rewrote the fix to use a in-memory linked list of system users to ldap DN mappings.  Also created a lib389 class for LDAPIMappings.  Waiting to hear if the new design will work for IPA, but this is ready for review.

Updated design doc:

https://www.port389.org/docs/389ds/design/ldapi-auto-auth-dn-design.html

ldap/servers/slapd/daemon.c
```diff
@@ -1778,6 +1789,83 @@ daemon_register_connection()
 }
 
 #if defined(ENABLE_LDAPI)
```
Firstyear: Unrelated to this commit, but we really should just remove --enable-ldapi because there is no reason no have it disabled ...

ldap/servers/slapd/daemon.c
```diff
@@ -1778,6 +1789,83 @@ daemon_register_connection()
...
+                    } else {
+                        mappings->next = new_mapping;
+                        mappings = mappings->next;
+                    }
+                } else {
+                    slapi_log_err(SLAPI_LOG_WARNING, "initialize_ldapi_auth_dn_mappings",
+                            "System username (%s) in entry (%s) was not found and will be ignored.\n",
+                            username, slapi_entry_get_dn(entries[i]));
+                }
+            }
```
Firstyear: Should there be a matching warning if username is null?

src/lib389/lib389/ldapi.py
```diff
@@ -0,0 +1,50 @@
...
+        @param instance - a DirSrv instance
+        @param container_dn - the subtree where the mapping entries are located
+        """
+        super(LDAPIMapping, self).__init__(instance, dn)
+        self._dn = dn
+        self._rdn_attribute = 'cn'
+        self._filterattrs = ['cn']
+        self._container_dn = container_dn
+
+    def createMappingContainer(self):
```
Firstyear: Shouldn't we just use a generic container type rather than a helper like this?

src/lib389/lib389/ldapi.py
```diff
@@ -0,0 +1,50 @@
...
+        self._dn = None
+        self.create(basedn=basedn, properties={'cn': rdn_val})
+
+    def createMappingEntry(self, name, username, ldap_dn):
+        """
+        @param name - name for the container entry.  This will be the RDN value.
+        @param username - the system username
+        @param ldap_dn - the LDAP DN that the system user should be mapped to
+        """
+        self._must_attributes = ['nsslapd-ldapiusername', 'nsslapd-authenticateasdn']
```
Firstyear: Shouldn't these must attributes be defined in __init__ so that .create works without this helper call?

Firstyear (changes requested):

Firstyear:
> Okay rewrote the fix to use a in-memory linked list of system users to ldap DN mappings. Also created a lib389 class for LDAPIMappings. 

Thanks mate, this is *much* better and will work in a much broader set of environments. Besides some python nitpicking the C code looks really good. 

> Waiting to hear if the new design will work for IPA, but this is ready for review.

Well I don't really want to accept the "previous design" so I hope that this works for them given it solves a broader set of actual issues that exist ....

tiran:
The updated design no longer solves my feature request #4381 and introduces new potential issues.

* ``nsslapd-ldapiusername`` cannot map local uid/gid to correct service account. For example on host ``server1`` the user ``named`` has to map to ``DNS/server1``. The same user account should map to ``DNS/server2`` on host ``server2``.
* In IPA case UID/GID lookup might introduce a lookup loop ``DS -> glibc -> NSS -> sssd -> DS``. This may cause issues and even dead locks.

I also don't see an index on ``nsslapd-ldapiusername``. How would the new attribute scale with large LDAP database?

tiran:
> One thing I'm worried about here in this design is systemd dynamic users. Right now we assume that a dn maps to a uid/gid number, but really it actually maps a dn to a local system uid which we should translate to a uidNumber. It's critical to understand in this system - dynamic users do NOT guarantee the stability of a uid/gid number assignment to a uid. As a result, this means that with this design, any service that would change to dynamic users would potentially not function if the uidnumber changed, causing a different service to be able to impersonate them.

The term **dynamic user** is overloaded and is used in different contexts. My feature request is about dynamically allocated UIDs and GIDs and not related to systemd's dynamic users. We cannot get statically allocated UIDs for some of our services. Users and groups are statically created in ``/etc/passwd`` on installation time. systemd dynamic users (aka ephemeral user accounts) are out of scope for my feature request.

ldap/servers/slapd/daemon.c
```diff
@@ -1778,6 +1789,83 @@ daemon_register_connection()
...
+    char *dn;
+    uid_t uid;
+    gid_t gid;
+    struct ldapi_mapping *next;
+};
+
+static struct ldapi_mapping *ldapi_mappings = NULL;
+
+static void
+initialize_ldapi_auth_dn_mappings(void)
```
tiran: This code introduces a static lookup cache from uid/gid to account. However you cannot safely assume that UID/GID are fixed during the lifetime of a 389-DS process, especially in the context of systemd dynamic user feature. systemd only *tries* to keep the uid and gid of a dynamic user stable. But it does not and cannot guarantee that it will always allocate the same UID/GID for ``DynamicUser``. This fact is explained in the *Persistent Data* section of http://0pointer.net/blog/dynamic-users-with-systemd.html. The cached uid may become invalid or map to a totally different user. This is a potential security issue.

Further more the static cache also won't work for services with dynamic users that are started after 389-DS. The dynamic user only exists when the service is running:

```
# systemctl start dynuser
# id dynuser
uid=62382(dynuser) gid=62382(dynuser) groups=62382(dynuser)
# systemctl stop dynuser
# id dynuser
id: ‘dynuser’: no such user
```

mreynolds389:
> The updated design no longer solves my feature request #4381 and introduces new potential issues.
> 
>     * `nsslapd-ldapiusername` cannot map local uid/gid to correct service account. For example on host `server1` the user `named` has to map to `DNS/server1`. The same user account should map to `DNS/server2` on host `server2`.

Okay in that case this approach will not work if each username needs to map to a different ldap entry on each replica.

> 
>     * In IPA case UID/GID lookup might introduce a lookup loop `DS -> glibc -> NSS -> sssd -> DS`. This may cause issues and even dead locks.
So getpwnam() invokes NSS -> sssd -> DS?
> 
> 
> I also don't see an index on `nsslapd-ldapiusername`. How would the new attribute scale with large LDAP database?

This attribute does not need an index.  It is never searched.   The only searching the server does is on the new objectclass which is indexed.  But I guess that doesn't matter if this patch does not fix your issue.

ldap/servers/slapd/daemon.c
mreynolds389: Just to play devils advocate, wouldn't this also break the original design?  I was previously hardcoding the UID/GID, but your argument is that it can change during the lifetime of the DS process.  So that would break as well, am I missing something?

tiran:
> Okay in that case this approach will not work if each username needs to map to a different ldap entry on each replica.

Sorry, I did not state this requirement explicitly in my ticket and only implied it with the example ``krbprincipalname=DNS/server.ipa.example@IPA.EXAMPLE,cn=services,cn=accounts,dc=ipa,dc=example``. I should have been more explicit and added a second example.

> So getpwnam() invokes NSS -> sssd -> DS?

In this context "NSS" stands for Name Service Switch, not for Netscape Security Service, https://man7.org/linux/man-pages/man5/nsswitch.conf.5.html. 

getpwnam() is a glibc function. Internally glibc uses the provider plugins from /etc/nsswitch.conf, e.g. ``files`` for a file database in ``/etc`` like ``/etc/passwd``. SSSD is a NSS provider for ``passwd`` (users) and ``group`` databases. getpwnam() calls into SSS NSS provider, which then talks to sssd daemon, which will perform an LDAP query to lookup the user record.

ldap/servers/slapd/daemon.c
tiran: Yes, it could break in theory. In practice there are rules, packaging guidelines, and best practices in place to prevent breakage. Of course applications can ignore the rules and shoot themselves in the foot.

System accounts are typically created once and never removed. RPM packaging guidelines state that users and groups created by packages should never be removed as well, https://fedoraproject.org/wiki/Packaging:UsersAndGroups . My PoC has additional precautions and updates UID/GID mappings on ``ipa-restore``.

Firstyear: > The updated design no longer solves my feature request #4381 and introduces new potential issues.
> 
> * `nsslapd-ldapiusername` cannot map local uid/gid to correct service account. For example on host `server1` the user `named` has to map to `DNS/server1`. The same user account should map to `DNS/server2` on host `server2`.

You can set the search base to cn=config for per server maps.

> * In IPA case UID/GID lookup might introduce a lookup loop `DS -> glibc -> NSS -> sssd -> DS`. This may cause issues and even dead locks.

Only if the source of your uid/gid is also DS. You should not be allocated systemd uid/gids in DS. So the actual flow for say bind is:

DS -> glibc -> nss -> sssd -> files

EDIT: As well, even if you did store the system id in DS then three scenarioes are possible. SSSD will fail over to a different ds server and the resolve will succeed, sssd will serve the value from it's cache, or sssd will fail to connect to the localhost DS and give no result. Either way, the point stands - it's extraordinarily bad practice to store system uid/gid in LDAP, and this feature is for mapping system uid/gid to a principal on systems where that ID can change between servers. So because of the above, this is still the correct design. 

> 
> I also don't see an index on `nsslapd-ldapiusername`. How would the new attribute scale with large LDAP database?

Because it's started on a specific subtree, probably not tooooo bad, but it should be indexed.

Firstyear: > This code introduces a static lookup cache from uid/gid to account. However you cannot safely assume that UID/GID are fixed during the lifetime of a 389-DS process,

The original design DOES assume they are fixed, because it literally hardcodes the uid/gid.

tiran:
For my FreeIPA use case I'm only interested in ``nsslapd-authenticateAsDN``. Since we have full control and reign over the lifetime of our system users, we do not need ``nsslapd-ldapiUsername``. To avoid any potential issues with NSS/SSSD/KDC/LDAP I prefer to stick to the existing uidNumber/gidNumber mapping.

I do see the benefits of mapping SO_PEERCRED uid to account name for other use cases. In my opinion, the proposed implementation is flawed and too complex. Instead of caching name -> uid in a linked list, I suggest that an implementation should rather lookup user name by uid in each authentication request. ``getpwuid()`` is blazing fast. On my local system and test system the lookup is in the magnitude of 1μs to 10μs in common cases and 100μs in worst cases. The slowest case is ``getpwuid(0)`` because it always falls back to parsing ``/etc/passwd``.

```
$ python3 -m timeit -s 'from pwd import getpwuid' 'getpwuid(0)'
10000 loops, best of 5: 26.6 usec per loop
$ python3 -m timeit -s 'from pwd import getpwuid' 'getpwuid(1)'
200000 loops, best of 5: 1.3 usec per loop
```

## Proposal

* if ``SO_PEERCRED`` uid == 0, bind as DM
* search for ``(&(uidNumber=$UID)(gidNumber=$GID))``
* if search results no result and new config option ``cn=config nsslapd-ldapilookupbyname == TRUE``
  * translate uid to name with ``getpwuid_r`` (**not** ``getpwuid``!)
  * search for ``(nsslapd-ldapiUsername=pwd->pw_name)``
* perform optional ``nsslapd-authenticateAsDN`` translation

mreynolds389:
@tiran Thanks for your continued investigation!   I actually already have a clean solution to this that works in all cases.  I will be updating the design doc later today, and I will update this ticket once it is ready for review.

mreynolds389:
Ok, third rewrite of the feature.  The design doc has all the relevant information, but I basically created two LDAPI mapping config entries.  For for static UID/GID assign for LDAP DN mapping, and one where the UID/GID is dynamically obtained at server startup.  So it handles both scenarios now.

src/lib389/lib389/ldapi.py
Firstyear: This comment still needs addressing :)

ldap/servers/slapd/daemon.c
```diff
@@ -1778,6 +1789,99 @@ daemon_register_connection()
...
+    slapi_search_internal_pb(search_pb);
+    slapi_pblock_get(search_pb, SLAPI_PLUGIN_INTOP_RESULT, &result);
+    if (result == LDAP_SUCCESS) {
+        slapi_pblock_get(search_pb, SLAPI_PLUGIN_INTOP_SEARCH_ENTRIES, &entries);
+        for (size_t i = 0; entries && entries[i]; i++) {
+            const char *username = slapi_entry_attr_get_ref(entries[i], CONFIG_LDAPI_AUTH_USERNAME_ATTRIBUTE);
+            uid_t user_uid = slapi_entry_attr_get_uint(entries[i], "uidNumber");
+            gid_t user_gid = slapi_entry_attr_get_uint(entries[i], "gidNumber");
+
+            if (user_uid && user_gid) {
```
Firstyear: If we are using this logic, perhaps it would be worth it being a single objectClass and we require the uid to be set, but then you can force-override a uid/gid number if you want?

Firstyear (changes requested):
Only some smaller comments this time, other wise looks good.

ldap/servers/slapd/daemon.c
mreynolds389: This could be improved, because I wrote the schema to enforce that the needed values are set.  Either uid & gid are both NULL or they are both set (we don't need to check both are not NULL).  It's one or the other, and the code does not represent that very well.  So I'd prefer to keep the two objectclasses, and let me revise this code a bit...

ldap/servers/slapd/daemon.c
mreynolds389: Yeah I was thinking about that as well.  There is another defined feature LDAPI  ENABLE_AUTO_DN_SUFFIX.  That needs to be looked at as well.

ldap/servers/slapd/daemon.c
Firstyear: That's fine, I'm not too stubborn about this detail.

tiran:
The scratch build from 20201214 no longer sets the effective DN to the value of ``nsslapd-authenticateasdn``:

## configuration

```
dn: cn=named,cn=autobind,cn=config
objectClass: nsLDAPIFixedAuthMap
objectClass: top
cn: named
uidNumber: 25
gidNumber: 25
nsslapd-authenticateAsDN:
krbprincipalname=DNS/ipaserver.ipa.test@IPA.TEST,cn=services,cn=accounts,dc=ipa,dc=test
```

## whoami

```
# sudo -u named ldapwhoami -H ldapi://%2Frun%2Fslapd-IPA-TEST.socket -Y EXTERNAL
SASL/EXTERNAL authentication started
SASL username: gidNumber=25+uidNumber=25,cn=peercred,cn=external,cn=auth
SASL SSF: 0
dn: cn=named,cn=autobind,cn=config
```

## access log

```
[15/Dec/2020:12:14:10.004502612 +0100] conn=61 AUTOBIND dn="cn=named,cn=autobind,cn=config"
```

## expected dn

``krbprincipalname=DNS/ipaserver.ipa.test@IPA.TEST,cn=services,cn=accounts,dc=ipa,dc=test``

mreynolds389:
@tiran Yup there was bug in the mapping.  After fixing it I do see the expected results:

# sudo -u ldapi_test_lib389_user2 ldapwhoami -H ldapi://%2Fvar%2Frun%2Fslapd-standalone1.socket -Y EXTERNAL
SASL/EXTERNAL authentication started
SASL username: gidNumber=1002+uidNumber=1002,cn=peercred,cn=external,cn=auth
SASL SSF: 0
dn: uid=test_ldapi2,ou=people,dc=example,dc=com

Access log also shows correct user.  A new build is in the works and I will let you know when it is ready

dirsrvtests/tests/suites/ldapi/ldapi_test.py
```diff
@@ -0,0 +1,160 @@
...
+from lib389._constants import DEFAULT_SUFFIX, DN_DM
+from lib389.idm.user import UserAccounts
+from lib389.ldapi import LDAPIMapping
+from lib389.ldap_objs import DSContainer
+from lib389.topologies import topology_st as topo
+from lib389.tasks import LDAPIMappingReloadTask
+
+log = logging.getLogger(__name__)
+
+
```
Firstyear, line 11: Test needs a marker to state that it adds/removes users, especially because your useradd command uses 1001 and 1002 as uids which as low numbers *may* already exist on the system.

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,409 @@
...
+ * END COPYRIGHT BLOCK **/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <unistd.h>
+#include "slap.h"
+
+#if defined(ENABLE_LDAPI)
```
Firstyear, line 17: Could be above the includes as we don't need anything unless ldapi is enabled (but also, we want to remove this define so probably not a nitpick that matters ....)

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,409 @@
...
+
+struct ldapi_mapping {
+    char *dn;
+    uid_t uid;
+    gid_t gid;
+    struct ldapi_mapping *next;
+};
+
+static struct ldapi_mapping *ldapi_mappings = NULL;
+static pthread_mutex_t dn_mapping_lock;
```
Firstyear: Consider using a rwlock here instead. If we use a mutex, with heavy threads we'll serialise on the lock, but with a rwlock we wont block and allows parallel ops to continue. Also Ensure if you use a rwlock to set it to "writer preferring" so that updates to this do not stall forever which is a problem on the default linux "reader preferring".

ldap/servers/slapd/slap.h
```diff
@@ -2740,4 +2742,15 @@ typedef enum _upgrade_status {
 upgrade_status upgrade_server(void);
 PRBool upgrade_plugin_removed(char *plg_libpath);
 
+/* ldapi.c */
+#if defined(ENABLE_LDAPI)
+#define LDAPI_STARTUP 0
+#define LDAPI_RELOAD 1
```
Firstyear: Use an enum rather than defines.

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,409 @@
...
+
+    pthread_mutex_unlock(&dn_mapping_lock);
+
+    if (shutdown) {
+        pthread_mutex_destroy(&dn_mapping_lock);
+    }
+}
+
+int32_t
+slapd_identify_local_user(Connection *conn)
```
Firstyear, line 137: Is there a way to PR_ASSERT that the connection lock is held here?

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,409 @@
...
+    /* bind at all? */
+    if (config_get_ldapi_bind_switch()) {
+        /* map users to a dn root may also map to an entry */
+        /* require real entry? */
+        if (config_get_ldapi_map_entries()) {
+            /*
+             * First check if the uid/gid maps to one of our LDAPI auth mappings
+             */
+        	pthread_mutex_lock(&dn_mapping_lock);
+            for (struct ldapi_mapping *mapping = ldapi_mappings; mapping; mapping = mapping->next) {
```
Firstyear, line 182: Indentation?

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,409 @@
...
+                        0, /* pb not req */
+                        e,
+                        0, /* no response control */
+                        0, /* don't check password policy */
+                        0  /* don't send ldap result */
+                        );
+
+                    if (1 == ret)
+                        /* sorry root,
+                         * just not cool enough
```
Firstyear, line 342: Consider adding "😎" to this comment for impact.

src/lib389/lib389/ldapi.py
Firstyear: Same comment :)

ldap/servers/slapd/libglobs.c
```diff
@@ -827,6 +827,10 @@ static struct config_get_and_set
      NULL, 0,
      (void **)&global_slapdFrontendConfig.ldapi_search_base_dn,
      CONFIG_STRING, NULL, SLAPD_DEFAULT_LDAPI_SEARCH_BASE, NULL},
+    {CONFIG_LDAPI_AUTH_MAP_BASE_ATTRIBUTE, config_set_ldapi_mapping_base_dn,
```
Firstyear, line 836: Should this have a default value?

Firstyear (changes requested):
Really getting there mate, i really like the break out to an ldapi files here :)

ldap/servers/slapd/ldapi.c
mreynolds389: I did not add that comment (wish I had) :-)

dirsrvtests/tests/suites/ldapi/ldapi_test.py
mreynolds389: I changed it to  500[1-3]

ldap/servers/slapd/ldapi.c
Firstyear: Nothing stopping you from improving the comment for future generations :)

ldap/servers/slapd/ldapi.c
mreynolds389: I will take care of the defines in a separate ticket.

ldap/servers/slapd/ldapi.c
mreynolds389: Done.

ldap/servers/slapd/ldapi.c
mreynolds389: Hmmm, I don't think so, but this function itself, or how it is called, was not changed.  Never had a problem before...

ldap/servers/slapd/libglobs.c
mreynolds389: The ldapi code uses NULL to check if anything is set, so a default "string" is not needed.  Or did you mean something else?

ldap/servers/slapd/slap.h
mreynolds389: Had a feeling you might request enums on this one, but got side tracked.  So sure I'll switch this around.

src/lib389/lib389/ldapi.py
mreynolds389: No, because they are reset in the fixed mapping function.  I guess that means it should really be two classes I suppose :-/

ldap/servers/slapd/ldapi.c
mreynolds389: Fair enough

dirsrvtests/tests/suites/ldapi/ldapi_test.py
Firstyear: I still think it needs a "ack" marker, similar to what @droideck did recently for the dd/log test since this will change/affect your system. It also probably needs a "are we root" check too.

ldap/servers/slapd/ldapi.c
```diff
@@ -0,0 +1,408 @@
...
+#if defined(ENABLE_AUTO_DN_SUFFIX)
+        if (ret) {
+            /* create phony auth dn? */
+            char *base = config_get_ldapi_auto_dn_suffix();
+            if (base) {
+                char *tpl = "gidNumber=%u+uidNumber=%u,";
+                int32_t len =
+                    strlen(tpl) +
+                    strlen(base) +
+                    51 /* uid,gid,null,w/padding */
```
Firstyear: Does this account for 32bit uid/gid values rather than just 16 bit ones?

ldap/servers/slapd/libglobs.c
Firstyear: I think I'm suggesting that we set a default to cn=auto_bind_map,cn=config here so that admins "don't" need to configure anything beside adding their maps, then IPA can use/tweak this value if they need to customise it. In the current form the maps are "disabled by default" with this always being NULL.

ldap/servers/slapd/slap.h
Firstyear: Trust your feelings mark, you know william will always complain in a review :)

src/lib389/lib389/ldapi.py
Firstyear: Yeah i think it's two classes.

Firstyear (changes requested):
It's pretty close to done mate :) Thanks for your patience with my reviews.

ldap/servers/slapd/ldapi.c
mreynolds389: No idea, this is all copy & paste code.  I will need to look into it...

Firstyear: Yeah, I think we should do our due diligence here :)

mreynolds389:
@firstyear all changes made please review...

dirsrvtests/tests/suites/ldapi/ldapi_test.py
Firstyear: Bump on this one.

ldap/servers/slapd/slap.h
Firstyear: ^---

ldap/servers/slapd/ldapi.c
Firstyear: ^---

Firstyear (changes requested):
I think there are three previous comments that still needed follow up? Unless I missed something ....

dirsrvtests/tests/suites/ldapi/ldapi_test.py
mreynolds389: I'll remove the log...

ldap/servers/slapd/slap.h
mreynolds389: Yeah there were so many comments in this PR I got lost in what was being asked for.  Still being used to github I suppose...

mreynolds389:
@Firstyear - okay I think I got it all sorted out :-)  Please review...

Firstyear:
Ack from me!

PR #4481, opened 2020-12-08
Issue 1795 - RFE - Enable logging for libldap and libber in error log

### LINKED ISSUE #1795: enable logging for third party libs in error log
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/48680
- Created at 2016-02-18 18:21:12 by [lkrispen](https://pagure.io/user/lkrispen) (@elkris)
- Assigned to [lkrispen](https://pagure.io/user/lkrispen) (@elkris)

---

libraries like libldap, libber, libsasl do error and debug logging, but it is not available in the DS logs. 

There is a mechanism to provide a logging function and define debug levels for these libs and get the messages logged in the DS error log.

droideck:
Description: Libraries like libldap, libber do error and debug
logging, but it is not available in the DS logs.

Provide a way to enable the third party logging in DS.
Add nsslapd-external-libs-debug-enabled attribute to 'cn=config'
which will enable all of the levels abailable in libldap and libber.
The setting should be used only for debugging purposes as
it prints all of the operataions with a great verbosity.

The code for log_external_libs_debug_print() and
log_external_libs_debug_set_log_fn() functions are provided
by a former Red Hat employee - Ludwig Krispenz.

Fixes: #1795

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_test.py
+++ b/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_test.py
@@ -138,6 +138,40 @@ def test_verify_operation_when_disk_monitoring_is_off(topo, setup, reset_logs):
 
 
 @disk_monitoring_ack
+def test_enable_external_libs_debug_log(topo, setup, reset_logs):
+    """Check that OpenLDAP logs are successfully enabled and disabled when
+    disk threshold is reached
+
+    :id: 121b2b24-ecba-48e2-9ee2-312d929dc8c6
+    :setup: Standalone instance
+    :steps: 1. Set nsslapd-external-libs-debug-enabled to "on"
+            2. Go straight below 1/2 of the threshold
+            3. Verify that the external libs debug setting is disabled
+            4. Go back above 1/2 of the threshold
+            5. Verify that the external libs debug setting is enabled back
+    :expectedresults: 1. Success
+                      2. Success
+                      3. Success
+                      4. Success
+                      5. Success
+    """
+    try:
+        # Verify that verbose logging was set to default level
+        assert topo.standalone.config.set('nsslapd-disk-monitoring', 'on')
+        assert topo.standalone.config.set('nsslapd-disk-monitoring-logging-critical', 'off')
+        assert topo.standalone.config.set('nsslapd-external-libs-debug-enabled', 'on')
+        assert topo.standalone.config.set('nsslapd-errorlog-level', '8')
+        topo.standalone.restart()
+        subprocess.call(['dd', 'if=/dev/zero', 'of={}/foo'.format(topo.standalone.ds_paths.log_dir), 'bs=1M', 'count={}'.format(HALF_THR_FILL_SIZE)])
+        # Verify that logging is disabled
+        _withouterrorlog(topo, "topo.standalone.config.get_attr_val_utf8('nsslapd-external-libs-debug-enabled') != 'off'", 31)
+    finally:
+        os.remove('{}/foo'.format(topo.standalone.ds_paths.log_dir))
+        _withouterrorlog(topo, "topo.standalone.config.get_attr_val_utf8('nsslapd-external-libs-debug-enabled') != 'on'", 31)
+        assert topo.standalone.config.set('nsslapd-external-libs-debug-enabled', 'off')
+
+
+@disk_monitoring_ack
 def test_free_up_the_disk_space_and_change_ds_config(topo, setup, reset_logs):
     """Free up the disk space and change DS config
 
@@ -734,3 +768,4 @@ def test_valid_operations_are_permitted(topo, setup, reset_logs):
 if __name__ == '__main__':
     CURRENT_FILE = os.path.realpath(__file__)
     pytest.main("-s -v %s" % CURRENT_FILE)
+
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -409,6 +409,7 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
     int using_accesslog = 0;
     int using_auditlog = 0;
     int using_auditfaillog = 0;
+    int using_external_libs_debug = 0;
     int logs_disabled = 0;
     int grace_period = 0;
     int first_pass = 1;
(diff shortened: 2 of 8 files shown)


droideck:
The naming and the setting itself is up for discussion, of course :)

Also, I'll add a test for disk monitoring tomorrow.

ldap/servers/slapd/log.c
```diff
@@ -441,6 +441,23 @@ log_set_logging(const char *attrname, char *value, int logtype, char *errorbuf,
     return LDAP_SUCCESS;
 }
 
+void
+log_external_libs_debug_print(char *buffer)
+{
+    slapi_log_error(SLAPI_LOG_FATAL, "OpenLDAP Client", "%s", buffer);
```
Firstyear: Shouldn't this be "external lib -> " rather than "openldap" since we could have many types of external lib?

Firstyear: Log level should be TRACE or DEBUG maybe too?

ldap/servers/slapd/log.c
```diff
@@ -441,6 +441,23 @@ log_set_logging(const char *attrname, char *value, int logtype, char *errorbuf,
...
+}
+
+int
+log_external_libs_debug_set_log_fn(void)
+{
+    int rc = ber_set_option(NULL, LBER_OPT_LOG_PRINT_FN, log_external_libs_debug_print);
```
Firstyear: How do you s-et the print fn for ldap? or does it just use stderr?

ldap/servers/slapd/daemon.c
```diff
@@ -493,6 +499,13 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
                     if (using_auditfaillog) {
                         config_set_auditfaillog_enabled(LOGGING_ON);
                     }
+                    if (using_external_libs_debug) {
+                        if (config_set_external_libs_debug_enabled(CONFIG_EXTERNAL_LIBS_DEBUG_ENABLED,
```
Firstyear, line 503: What's the intent for these calls to set the external lib debug enabled? Is it so you don't have to enable it via dse.ldif?

Firstyear (changes requested):

ldap/servers/slapd/log.c
droideck: > Shouldn't this be "external lib -> " rather than "openldap" since we could have many types of external lib?

If we won't mention the library name then log messages like this one won't make sense:

```
[07/Dec/2020:10:43:58.042403434 -0500] - ERR - OpenLDAP Client - attempting to connect:
```

But I think you have a good point and I should give this function a more accurate name - `log_external_libs_debug_print` - probably something like `log_external_libs_debug_openldap_print`?

> Log level should be TRACE or DEBUG maybe too?

My idea of the usage is like this:
1. You have an issue;
2. You set the debug level you need (maybe only replication?);
3. You think that the problem happens somewhere in the middle of replica auth;
4. You enable nsslapd-external-libs-debug-enabled attribute;
5. You see only OpenLDAP and Replication level output (you don't need any other noise from TRACE or DEBUG)

The same way you can 'inject' the info in any other debug level.

So I think it makes sense to have it distinctive. So it's the reason I put it to the minimal level (FATAL).
As an option, we can put it to the separate loglevel, maybe?
What do you think?

ldap/servers/slapd/log.c
droideck: It uses the LBER_OPT_LOG_PRINT_FN so it is set only one time (if I understand it correctly)
https://git.openldap.org/openldap/openldap/-/blob/master/libraries/libldap/print.c#L60

ldap/servers/slapd/daemon.c
droideck: It is for disk_monitoring daemon only.
It disables the logging setting if the disk threshold was crossed (and it will set it back to the previous state after the space is available once again).

ldap/servers/slapd/log.c
Firstyear: well I think fatal is just "too intense" - I think our default levels always display warn and error, so it could be better to set this to warn instead. 

But yes, now having read the other comments I agreed that the name should be "libldap/libber" instead of openldapclient in this, and that the function should be renamed to suit :)

ldap/servers/slapd/daemon.c
Firstyear: OHHHHHh I see. That's okay then :)

droideck:
Seems reasonable!
I've implemented the requested changes and I've added a test for disk monitoring.

Please, review:)

droideck:
Also, I'd like to have someone from Red Hat team to check too as this PR adds a new setting we all will use for debugging.

@mreynolds389 @tbordaz @jchapma @progier389 what do you thing about the design?

dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_test.py
```diff
@@ -731,6 +731,38 @@ def test_valid_operations_are_permitted(topo, setup, reset_logs):
     assert topo.standalone.config.set('nsslapd-disk-monitoring-logging-critical', 'on')
 
 
+def test_enable_external_libs_debug_log(topo, setup, reset_logs):
```
Firstyear: IIRC this test should have a marker or warning of some kind about how it will use disk space. Also some filesystems will see a dev/zero dd and just make a sparse file and can over-allocate ... there are also risks with corruption on some fs's (btrfs famous enospc issues).

droideck: Good catch! Fixed:)

Firstyear:
Great, ack from me :)

Firstyear (approved):

ldap/servers/slapd/log.c
```diff
@@ -441,6 +441,23 @@ log_set_logging(const char *attrname, char *value, int logtype, char *errorbuf,
     return LDAP_SUCCESS;
 }
 
+void
```
tbordaz: I think it could be defined as static

tbordaz (approved):
Except this minor comment, the patch looks good to me. ACK

PR #4484, opened 2020-12-09
Issue 3522 - Remove DES to AES conversion code

### LINKED ISSUE #3522: Remove DES to AES conversion code
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/50465
- Created at 2019-06-24 16:35:20 by [mreynolds](https://pagure.io/user/mreynolds) (@mreynolds389)
- Assigned to nobody
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=1859296

---

#### Issue Description

Back in 389-ds-base-1.3.3 we created a process to convert the DES reversible password plugin to AES.  This was to aid customer's doing an "in-place" migration from RHEL 6 to RHEL 7.    In 389-ds-base-1.4.x (RHEL 8) that is no "in-place' migration from RHEL 6 to RHEL 8.  We also suggest using the export/import or replication migration methods as described in the installation/migration guide, which both of these methods bypass this conversion code & upgrade perl script.

I think it is safe to remove the upgrade script, the CI test, and the conversion code in daemon.c convert_pbe_des_to_aes().

#### Steps to reproduce

1.
2.
3.

#### Expected results

mreynolds389:
Description: 

remove the reversible password storage scheme upgrade code.  This was only needed for people moving from 1.2.10, which has not been supported for years.

Fixes: https://github.com/389ds/389-ds-base/issues/3522


Firstyear:
Ack from me, but as an extra detail, should we also refuse to decrypt any remaining DES passwords that may exist and remove it as a choice from the PWD storage mech?

mreynolds389:
> Ack from me, but as an extra detail, should we also refuse to decrypt any remaining DES passwords that may exist and remove it as a choice from the PWD storage mech?

Well I'm still allowing the decoding of it, but I am removing the DES config entry by default, the slapi_conversion task, and added upgrade code to remove the config entry.  Since it's on by default we can not just fully strip it out without startup problems.  So this is the first phase.  Then in 3.x we will remove all code related to DES.

Firstyear:
Sure, that's fine to me. - perhaps we could take this chance that when ever a bind with DES is used we log a warning?

Otherwise, ack from me, thank you!

droideck (approved):
Besides the point @Firstyear mentioned (warning with which I tend to agree) - **Looks good to me too!**

mreynolds389:
> Sure, that's fine to me. - perhaps we could take this chance that when ever a bind with DES is used we log a warning?

Note - DES reversible password encryption is only used for legacy replication.  So it's not something that can be used by "users", but I will add warning...

Issue #4480, opened 2020-12-08
Unexpected info returned to ldap request


tbordaz:
**Issue Description**
A ldap result can contain additional information. Such information should not allow a client application to guess if an entry exists or not

**Package Version and Platform:**
This bug impacts all release after 1.4.2.3

**Steps to Reproduce**
to be provided with an automatic testcase

**Expected results**
A ldap request should not provide any tips if an entry exists or not

tbordaz:
0b08e6f35..cc0f69283  master
730b30f08..38b97faef  389-ds-base-1.4.4
fe0409325..c8e77b811  389-ds-base-1.4.3

mreynolds389:
@tbordaz - does this also apply to 1.3.10 (RHEL 7.9)?

tbordaz:
@mreynolds389 no it does not apply to 1.3.10 (RHEL 7.9) branch.
A side effect of #49476 (backend refactoring) create this bug. backend refactoring was not backported in 1.3.10

carnil:
@tbordaz 

> @mreynolds389 no it does not apply to 1.3.10 (RHEL 7.9) branch.
> A side effect of #49476 (backend refactoring) create this bug. backend refactoring was not backported in 1.3.10

Is there reference #49476 above correct.

Trying to check which Debian releases are currently affected by this issue.

tbordaz:
@carnil, this issue was a side effect of https://github.com/389ds/389-ds-base/issues/2535 (former pagure #49476). #2535 was applied on 1.4.x branch only. So #4480 does not impact 1.3.x nor 1.2.11 branches.

carnil:
> @carnil, this issue was a side effect of #2535 (former pagure #49476). #2535 was applied on 1.4.x branch only. So #4480 does not impact 1.3.x nor 1.2.11 branches.

@tbordaz: thank you! So we (in Debian) need to only additionally check the 1.4.0.21-1 version we have. Older suites have 1.3.x based and the current unstable was updated to 1.4.4.10 which has the fix.

Firstyear:
@carnil will the 1.4.4.x updates be put into current stable? It would be good to see debian updated here. Thanks,

tbordaz:
b2c67646e..16d9020a8 1.4.2

PR #4491, opened 2020-12-10
Issue 4480 - Unexpected info returned to ldap request

tbordaz:
Bug description:
	If the bind entry does not exist, the bind result info
        reports that 'No such entry'. It should not give any
        information if the target entry exists or not

Fix description:
	Does not return any additional information during a bind

relates: https://github.com/389ds/389-ds-base/issues/4480

Reviewed by:

Platforms tested:  F31

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -1477,6 +1477,36 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
         else:
             assert True
 
+def test_bind_invalid_entry(topology_st):
+    """Test the failing bind does not return information about the entry
+
+    :id: 5cd9b083-eea6-426b-84ca-83c26fc49a6f
+
+    :setup: Standalone instance
+
+    :steps:
+    1: bind as non existing entry
+    2: check that bind info does not report 'No such entry'
+
+    :expectedresults:
+    1: pass
+    2: pass
+    """
+
+    topology_st.standalone.restart()
+    INVALID_ENTRY="cn=foooo,%s" % DEFAULT_SUFFIX
+    try:
+        topology_st.standalone.simple_bind_s(INVALID_ENTRY, PASSWORD)
+    except ldap.LDAPError as e:
+        log.info('test_bind_invalid_entry: Failed to bind as %s (expected)' % INVALID_ENTRY)
+        log.info('exception description: ' + e.args[0]['desc'])
+        if 'info' in e.args[0]:
+            log.info('exception info: ' + e.args[0]['info'])
+        assert e.args[0]['desc'] == 'Invalid credentials'
+        assert 'info' not in e.args[0]
+        pass
+
+    log.info('test_bind_invalid_entry: PASSED')
 
 if __name__ == '__main__':
     # Run isolated
--- a/ldap/servers/slapd/back-ldbm/ldbm_config.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_config.c
@@ -1243,7 +1243,7 @@ ldbm_config_search_entry_callback(Slapi_PBlock *pb __attribute__((unused)),
             if (attrs) {
                 for (size_t i = 0; attrs[i]; i++) {
                     if (ldbm_config_moved_attr(attrs[i])) {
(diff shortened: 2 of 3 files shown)


Firstyear:
I see nothing wrong with the code and the patch here, that's all good to me. I just thing this is solving a problem that doesn't exist, and in exchange makes it harder for an admin or user to identify if they have made a mistake in their command by removing the difference between "wrong pw" and "no such dn".

Knowledge of if an entry/dn exists or not, does *not* create a risk, as in most deployments anonymous bind and read is enabled by default. If you really wanted to be pedantic you could even do timing attacks to work out if this existed or not, because it's effectively impossible to do constant time search/responses. 

Now, to prove this, here is redhats login page showing the difference between "account does not exist" and "attempt the pw". 

I think that we should reject this issue, because it's not a security risk, and this pattern is accepted the world over in many authentication systems. 

Screen Shot 2020-12-11 at 10 50 56

Screen Shot 2020-12-11 at 10 51 07

Firstyear:
Just to really nail this one shut, here is google exhibiting the same behaviour - google who is a huge security target, and hires some of the best security people in the world, and they have a ui that says if the account exists or does not.

As a result, I really firmly believe this is a non issue and should be closed/rejected. 

Screen Shot 2020-12-11 at 10 56 04

vashirov:
There was a [CVE](https://bugzilla.redhat.com/show_bug.cgi?id=1347760) that was fixed 4 years ago. Recently it was reintroduced. So this PR fixes this regression. 

> Now, to prove this, here is redhats login page showing the difference between "account does not exist" and "attempt the pw".

Here's a SUSE login page that **doesn't show** a difference between "account does not exist" and "attempt the pw"

(yes, I actually went and created an account)
image
image


> google who is a huge security target, and hires some of the best security people in the world, and they have a ui that says if the account exists or does not.

I think it is a faulty generalization and this argument should be rejected :)

tbordaz:
I have not strong opinion whether it is security concern or not to return some internal information.
But [CVE-2016-4992](https://bugzilla.redhat.com/show_bug.cgi?id=1347760) resulted into a big investigation and [fix](https://github.com/389ds/389-ds-base/commit/0b932d4b926d46ac5060f02617330dc444e06da1).
Part of that fix was to send "generic" ldap result and log "specific" info in access logs when the information can be sensitive.

This part of the fix got broken by a recent [change](https://github.com/389ds/389-ds-base/commit/94c74015e529c08d59cc3e856199888867bd6449). I think it got broken by mistake. The intention was to return notification to a client that a [ldbm tuning moved to bdb](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/slapd/back-ldbm/ldbm_config.c#L1246). For this it uses SLAPI_PB_RESULT_TEXT in the pblock and send it back during result. 

The problem being that CVE-2016-4992 used that field to **only** record info in access logs.
The my original patch is too limited. #49476 changed CVE fix regarding all operations (ADD,DEL, BIND...).
I will rework the patch

Firstyear:
> > google who is a huge security target, and hires some of the best security people in the world, and they have a ui that says if the account exists or does not.
> 
> I think it is a faulty generalization and this argument should be rejected :)

It should never have been a CVE in the first place because there is provably *zero risk* in this behaviour existing. Just because SUSE "agrees" with the behaviour, does not discount that both rh/google allow it. It's not a faulty generalisation. 


> 
> This part of the fix got broken by a recent [change](https://github.com/389ds/389-ds-base/commit/94c74015e529c08d59cc3e856199888867bd6449). I think it got broken by mistake. The intention was to return notification to a client that a [ldbm tuning moved to bdb](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/slapd/back-ldbm/ldbm_config.c#L1246). For this it uses SLAPI_PB_RESULT_TEXT in the pblock and send it back during result.
> 
> The problem being that CVE-2016-4992 used that field to **only** record info in access logs.
> The my original patch is too limited. #49476 changed CVE fix regarding all operations (ADD,DEL, BIND...).
> I will rework the patch

I think it should be a tunable setting, and off by default - if customers want to "believe" this is a security risk and lower usability at the same time, they should have to enable it.

tbordaz:
Updated patch to cover all operations

mreynolds389 (approved):
LGTM

Issue #4492, opened 2020-12-10
Changelog cache can upload updates from a wrong starting point (CSN)


tbordaz:
**Issue Description**
Changelog cache is the mechanism that uploads a bulk of updates from the changelog.
The updates are consecutive (CSN order) updates from a given starting point.
The problem is that the bulk of updates can start from the beginning of the changelog even if the starting point CSN exists in the changelog.

The consequence is that the changelog cache will iterate through the all changelog to retrieve the next update to send. If the changelog is large it will delay the sent updates and create replication lag

**Package Version and Platform:**
Likely since 1.2.12

**Steps to Reproduce**
Hammering a large topology with small updates.
Accelerators are
	  Replica:
	 	nsds5ReplicaReleaseTimeout:30
	  RA: 		
		nsds5replicaBusyWaitTime: 3
		nsds5replicaSessionPauseTime: 5

**Expected results**
Replication session should start before the initial starting point

tbordaz:
32413b5b1..f629f8f5a  master
18184e5e8..260128fa1  389-ds-base-1.4.4
305b75e56..82f51a608  389-ds-base-1.4.3

tbordaz:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1898541

PR #4493, opened 2020-12-10
Issue 4492 - Changelog cache can upload updates from a wrong starting…

tbordaz:
… point (CSN)

Bug description:
          When a replication session starts, a starting point is computed
          according to supplier/consumer RUVs.
	  from the starting point the updates are bulk loaded from the CL.
          When a bulk set have been fully evaluated the server needs to bulk load another set.
	  It iterates until there is no more updates to send.
          The bug is that during bulk load, it recomputes the CL cursor position
          and this computation can be wrong. For example if a new update on
          a rarely updated replica (or not known replica) the new position will
          be set before the inital starting point

Fix description:
          Fixing the invalid computation is a bit risky (complex code resulting from
          years of corner cases handling) and a fix could fail to address others flavor
          with the same symptom
          The fix is only (sorry for that) safety checking fix that would end a replication session
          if the computed cursor position goes before the initial starting point.

relates: https://github.com/389ds/389-ds-base/issues/4492

Reviewed by:

Platforms tested: F31

### DIFF
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -152,6 +152,7 @@ struct cl5replayiterator
     ReplicaId consumerRID;  /* consumer's RID */
     const RUV *consumerRuv; /* consumer's update vector */
     Object *supplierRuvObj; /* supplier's update vector object */
+    char starting_csn[CSN_STRSIZE];
 };
 
 typedef struct cl5iterator
@@ -1030,7 +1031,7 @@ cl5GetNextOperationToReplay(CL5ReplayIterator *iterator, CL5Entry *entry)
         return CL5_BAD_DATA;
     }
 
-    rc = clcache_get_next_change(iterator->clcache, (void **)&key, &keylen, (void **)&data, &datalen, &csn);
+    rc = clcache_get_next_change(iterator->clcache, (void **)&key, &keylen, (void **)&data, &datalen, &csn, iterator->starting_csn);
 
     if (rc == DB_NOTFOUND) {
         /*
@@ -3940,7 +3941,7 @@ _cl5PositionCursorForReplay(ReplicaId consumerRID, const RUV *consumerRuv, Repli
     if (rc != 0)
         goto done;
 
-    rc = clcache_load_buffer(clcache, &startCSN, continue_on_missing);
+    rc = clcache_load_buffer(clcache, &startCSN, continue_on_missing, NULL);
 
     if (rc == 0) {
         haveChanges = PR_TRUE;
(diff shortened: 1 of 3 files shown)


mreynolds389:
I think the logging should be SLAP_LOG_REPL, instead of SLAPI_LOG_WARNING.  Customers might see the warning as an error, as in something went wrong that needs fixing.  If the server is self mending it should not log something by default.

So in your testing, when you encounter the earlier CSN and abort that bulk load, does it then corrects itself on the next bulk load?  Or does it keep aborting every bulk load after that?

tbordaz:
> I think the logging should be SLAP_LOG_REPL, instead of SLAPI_LOG_WARNING. Customers might see the warning as an error, as in something went wrong that needs fixing. If the server is self mending it should not log something by default.

I agree. Could it be SLAPI_LOG_INFO ? The interesting info here is the size of the jump. If it is few minutes it is not a big deal but a large jump could be log.
Also if it becomes frequent it will be reported as annoying message but also will show us that we have something to fix here.
> 
> So in your testing, when you encounter the earlier CSN and abort that bulk load, does it then corrects itself on the next bulk load? Or does it keep aborting every bulk load after that?

When the earlier CSN is detected it is before the bulk load. When it occurs clcache_get_next_change stops doing bulk load and returns DB_NOTFOUND and eventually UPDATE_NO_MORE_UPDATES. This will end smoothly the replication session.
A next replication session will start with  a brand new starting_point and bulk load mechanism

mreynolds389:
> > I think the logging should be SLAP_LOG_REPL, instead of SLAPI_LOG_WARNING. Customers might see the warning as an error, as in something went wrong that needs fixing. If the server is self mending it should not log something by default.
> 
> I agree. Could it be SLAPI_LOG_INFO ? The interesting info here is the size of the jump. If it is few minutes it is not a big deal but a large jump could be log.

SLAPI_LOG_INFO is default logging.  I don't think it should be logged by default IMHO.  I really feel like it belongs in REPL debug logging as it's only going to cause customer complaints/inquiries about why this new error message is popping up (even if its an info message).  There is nothing an Admin can do about it, so it should be hidden IMO.  This particular customer is going to see these messages pop up quite often, and they will probably want it "fixed".  I just don't want to create additional headaches for support.

mreynolds389:
> > > I think the logging should be SLAP_LOG_REPL, instead of SLAPI_LOG_WARNING. Customers might see the warning as an error, as in something went wrong that needs fixing. If the server is self mending it should not log something by default.
> > 
> > 
> > I agree. Could it be SLAPI_LOG_INFO ? The interesting info here is the size of the jump. If it is few minutes it is not a big deal but a large jump could be log.
> 
> SLAPI_LOG_INFO is default logging. I don't think it should be logged by default IMHO. I really feel like it belongs in REPL debug logging as it's only going to cause customer complaints/inquiries about why this new error message is popping up (even if its an info message). There is nothing an Admin can do about it, so it should be hidden IMO. This particular customer is going to see these messages pop up quite often, and they will probably want it "fixed". I just don't want to create additional headaches for support.

That being said, if you want to add logging for very jumps that's fine.

Firstyear:
I think I agree with @mreynolds389, that having this as LOG_REPL seems correct, but I also like @tbordaz's idea to LOG_WARNING for very long jumps.

tbordaz:
@mreynolds389 , @Firstyear thanks for you reviews. I have updated the patch so that logs are done at SLAP_LOG_REPL level. An exception is in case of big (24h) jump behind the starting point, it logs SLAP_LOG_WARNING

mreynolds389 (approved):
LGTM, thanks!!

PR #4495, opened 2020-12-10
Backend redesign phase 3 - bdb dependency removal

progier389:
This is a partial commit about the backend refactoring (issue 4469). 
  it compiles but I have not done any tests yet.

Design documentation is in [https://www.port389.org/docs/389ds/design/backend-redesign-phase3.html](https://www.port389.org/docs/389ds/design/backend-redesign-phase3.html)

Still to do: 
   tests  
   replication changelog change
   db.h removal within dbimpl.h 
   handling of special case (like the fact that we are checking db region to detect that an import is in progress)


dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
```diff
@@ -301,11 +301,11 @@ def test_sync_repl_mep(topology, request):
 
     # checking that the cookie are in increasing and in an acceptable range (0..1000)
     assert len(cookies) > 0
-    prev = 0
+    prev = -1
```
Firstyear, line 306: This looks like it shouldn't be in this PR

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -502,7 +497,7 @@ struct attrinfo
     DataList *ai_idlistinfo;             /* fine grained id list */
 };
 
-#define MAXDBCACHE 20
+#define MAXdbi_cursor_tACHE 20
```
Firstyear: case?

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -653,14 +648,14 @@ struct ldbminfo
...
-#define LI_DEFAULT_IMPL_FLAG  LI_BDB_IMPL /* the default is BDB for now */
+#define LI_DEFAULT_IMPL_FLAG  LI_BDB_IMPL /* the default is Bdbi_db_t for now */
 
 /* Structure used to hold stuff for the lifetime of an LDAP transaction */
 /* If we do clever stuff like LDAP transactions, we'll need a stack of TXN ID's */
 typedef struct back_txn back_txn;
 struct back_txn
 {
-    DB_TXN *back_txn_txn; /* Transaction ID for the database */
+    dbi_txn_t *back_txn_txn; /* Transaction ID for the database */
```
Firstyear, line 656: We need to distinguish between read txn and write txn as these matter greatly to LMDB.

src/lib389/lib389/instance/setup.py
```diff
@@ -888,7 +888,7 @@ def _install_ds(self, general, slapd, backends):
                         tlsdb_inst = NssSsl(dbpath=os.path.join(etc_dirsrv_path, dir))
                         tlsdb_inst.import_rsa_crt(ca)
 
-            csr = tlsdb.create_rsa_key_and_csr()
+            csr = tlsdb.create_rsa_key_and_csr(alt_names=[general['full_machine_name']])
```
Firstyear, line 891: Also doesn't belong in this pr?

Firstyear (commented):
Impressive start mate, I think you may need to rebase to latest master though.

dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
progier389: I messed when trying to rebase to the master and 3 unwanted commits were pushed in the original pull request.
I rebased to remove them this morning so the extra files are gone. 
I will wait until the work is nearly done before re merging with the master (as the merge may be painful)

ldap/servers/slapd/back-ldbm/back-ldbm.h
progier389: in fact the #define is no more used and should be removed ...
Seeing your remark I searched for "dbi_cursor_t[^ ]" and "[a-zA-Z0-9_]dbi_cursor_t" and the good new is that there is no other unwanted substitution.

ldap/servers/slapd/back-ldbm/back-ldbm.h
progier389: I just added a section about read txn in the open points of the design document.
In short the read txn support is not something that I plan to do in phase 3.
The only thing is that we should have an architecture that could easily evolve to support it (and that is the case as we can add attribute to back_txn struct or within the  dbi_txn_t one.)

mreynolds389:
So far this is very impressive!  I'm noticing indentation issues, especially in ldap/servers/slapd/back-ldbm/dbimpl.h, but I'm going to hold off on singling them out for now as this is still a "work in progress".

progier389:
Pushed a few fixes - some basic functions (add several level of entries  and subtree search) are now OK - tests are continuing.

progier389:
Pushed a few fixes about import, modrdn and delete operation.
Now tests/suites/basic/basic_test.py complete successfully.

libasan result are quite good: so once I removed the still water leaks (about suffix handling in the startup phase - and this seems unrelated to this PR) , I only saw a memory leak reported about delete operation (that I am investigating)

mreynolds389:
> libasan result are quite good: so once I removed the still water leaks (about suffix handling in the startup phase - and this seems unrelated to this PR)

I think @jchapma was looking into this leak, just an FYI.

progier389:
Committed the fixes about range search and vlv. now the vlv py.tests are pass.
During my test libasan reported a buffer read overflow in filtercmp.c 
 while calling memcmp (when comparing keys of different size)
 I fixed the issue but as it is unrelated to the backend changes, I will open a new issue to track that one

progier389:
Looks like this one got involuntary merged in 389ds:i4469 branch while I tried to understand why the rebase added 250 files in the review (it was probably because the PR was targeted toward a specific branch rather than master)
==> will delete then create the branch then a new PR to solve the issue.

PR #4496, opened 2020-12-11
Issue 4373 - BUG - calloc of size 0 in MT build

### LINKED ISSUE #4373: dsconf chaining may create invalid config when parent suffix does not exist

Firstyear:
Bug Description: In some cases it's possible for there to be
no mapping trees which causes a warning of a calloc of size
0.

Fix Description: In these cases, we can skip the attempt to calloc
and build.

fixes: https://github.com/389ds/389-ds-base/issues/4373

Review by: ???

### DIFF
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -918,6 +918,14 @@ mapping_tree_node_build_tree()
      */
     int ent_count = 0;
     slapi_pblock_get(pb, SLAPI_NENTRIES, (void *)&ent_count);
+    if (ent_count == 0) {
+        /*
+         * It is invalid to do a calloc of 0, so if we have no entries. we just return.
+         */
+        result = 0;
+        goto build_tree_done;
+    }
+
     struct mt_suffix_ord *ordered_suffixes = (struct mt_suffix_ord *)slapi_ch_calloc(ent_count, sizeof(struct mt_suffix_ord));
     /* Assert the last value is null, and that we don't sigsegv */
     PR_ASSERT(entries[ent_count] == NULL);


mreynolds389:
LGTM

Firstyear:
```
   5a80dd90b..18184e5e8  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
   ccc33ecf9..305b75e56  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
```

ldap/servers/slapd/mapping_tree.c
```diff
@@ -918,6 +918,14 @@ mapping_tree_node_build_tree()
      */
     int ent_count = 0;
     slapi_pblock_get(pb, SLAPI_NENTRIES, (void *)&ent_count);
+    if (ent_count == 0) {
```
tbordaz, line 921: Just for curiosity, at the point "_Slapi_Entry **entries_" is not NULL, in which case ent_count could it be 0 ?

Firstyear: @tbordaz Good catch, looking at search_internal_pb then entries is always set to entries size + 1, but ent_count could be 0 so I think I need to do a change like:

```
diff --git a/ldap/servers/slapd/mapping_tree.c b/ldap/servers/slapd/mapping_tree.c
index 2cc4bf82c..6df27163f 100644
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -936,8 +936,8 @@ mapping_tree_node_build_tree()
     }
     slapi_ch_free((void **)&ordered_suffixes);

-    slapi_free_search_results_internal(pb);
 build_tree_done:
+    slapi_free_search_results_internal(pb);
     slapi_pblock_destroy(pb);
     slapi_ch_free((void **)&filter);
     return result;
```

Firstyear: See #4502 @tbordaz

PR #4502, opened 2020-12-15
Issue 4373 - BUG - one line cleanup, free results in mt if ent 0

### LINKED ISSUE #4373: dsconf chaining may create invalid config when parent suffix does not exist

Firstyear:
Bug Description: We had a free on the wrong line which could
lead to a memory leak during server setup.

Fix Description: Free results if ent count is 0

fixes: https://github.com/389ds/389-ds-base/issues/4373

Review by: ???

### DIFF
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -967,8 +967,8 @@ mapping_tree_node_build_tree()
     }
     slapi_ch_free((void **)&ordered_suffixes);
 
-    slapi_free_search_results_internal(pb);
 build_tree_done:
+    slapi_free_search_results_internal(pb);
     slapi_pblock_destroy(pb);
     slapi_ch_free((void **)&filter);
     return result;


mreynolds389:
LGTM

Firstyear:
Thanks @mreynolds389, I'll let @tbordaz have a check too.

tbordaz:
Good catch. ACK

Firstyear:
@tbordaz You're the one who caught it :) Thank you both!

Firstyear:
```
   925f3e19f..730b30f08  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
   55b21d069..fe0409325  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
```

PR #4503, opened 2020-12-15
Issue 4498 - BUG - entryuuid replication may not work

### LINKED ISSUE #4498: ipa-replica-install fails with 'single-valued attribute "entryUUID" has multiple values' (master branch)
**Issue Description**
When using 389ds/389-ds-base-nightly copr repo, ipa-replica-install fails with an error related to `single-valued attribute "entryUUID" has multiple values`.

 - Platform: Fedora32
 - Package and version: 389-ds-base-2.0.1-20201212git429c2f85a.fc32.x86_64

**Steps to Reproduce**
1. install ipa on the first server: ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarder -a $PASSWORD -p $PASSWORD -U
2. install ipa client on the replica: ipa-client-install --domain ipa.test --realm IPA.TEST --server server.ipa.test --principal admin --password $PASSWORD -U
3. promote client to replica: kinit admin; ipa-replica-install -U
4. See error:
```
[...]
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] Configuring the web interface (httpd)
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [1/21]: stopping httpd
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [2/21]: backing up ssl.conf
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [3/21]: disabling nss.conf
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [4/21]: configuring mod_ssl certificate paths
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [5/21]: setting mod_ssl protocol list
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [6/21]: configuring mod_ssl log directory
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [7/21]: disabling mod_ssl OCSP
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [8/21]: adding URL rewriting rules
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [9/21]: configuring httpd
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [10/21]: setting up httpd keytab
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [11/21]: configuring Gssproxy
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [12/21]: setting up ssl
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58]   [error] RuntimeError: Certificate issuance failed (CA_UNREACHABLE: Server at https://master.ipa.test/ipa/json failed request, will retry: 4205 (single-valued attribute "entryUUID" has multiple values).)
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] Certificate issuance failed (CA_UNREACHABLE: Server at https://master.ipa.test/ipa/json failed request, will retry: 4205 (single-valued attribute "entryUUID" has multiple values).)
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] Your system may be partly configured.
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] Run /usr/sbin/ipa-server-install --uninstall to clean up.
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] 
[ipatests.pytest_ipa.integration.host.Host.replica0.cmd58] Exit code: 1
```

**Expected results**
ipa-replica-install should succeed.

**Additional context**
The logs are available in PR#[587](https://github.com/freeipa-pr-ci2/freeipa/pull/587), for instance for the test customized_ds_config_install: [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e9ccc288-3ce8-11eb-be0c-fa163ecad8b9/report.html), [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/e9ccc288-3ce8-11eb-be0c-fa163ecad8b9/test_integration-test_customized_ds_config_install.py-TestCustomDSConfigInstall-test_customized_ds_install_replica/).

Companion issue opened against freeipa: https://pagure.io/freeipa/issue/8619

Firstyear:
Bug Description: EntryUUID can be duplicated in replication

Fix Description: Add a test case to determine how this occurs

fixes: https://github.com/389ds/389-ds-base/issues/4498

Review by: ???

### DIFF
--- a/dirsrvtests/tests/suites/entryuuid/replicated_test.py
+++ b/dirsrvtests/tests/suites/entryuuid/replicated_test.py
@@ -1,0 +1,77 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 William Brown <william@blackhats.net.au>
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import ldap
+import pytest
+import logging
+from lib389.topologies import topology_m2 as topo_m2
+from lib389.idm.user import nsUserAccounts
+from lib389.paths import Paths
+from lib389.utils import ds_is_older
+from lib389._constants import *
+from lib389.replica import ReplicationManager
+
+default_paths = Paths()
+
+pytestmark = pytest.mark.tier1
+
+@pytest.mark.skipif(not default_paths.rust_enabled or ds_is_older('1.4.2.0'), reason="Entryuuid is not available in older versions")
+
+def test_entryuuid_with_replication(topo_m2):
+    """ Check that entryuuid works with replication
+
+    :id: a5f15bf9-7f63-473a-840c-b9037b787024
+
+    :setup: two node mmr
+
+    :steps:
+        1. Create an entry on one server
+        2. Wait for replication
+        3. Assert it is on the second
+
+    :expectedresults:
+        1. Success
+        1. Success
+        1. Success
+    """
+
+    server_a = topo_m2.ms["master1"]
+    server_b = topo_m2.ms["master2"]
+    server_a.config.loglevel(vals=(ErrorLog.DEFAULT,ErrorLog.TRACE))
+    server_b.config.loglevel(vals=(ErrorLog.DEFAULT,ErrorLog.TRACE))
+
+    repl = ReplicationManager(DEFAULT_SUFFIX)
+
+    account_a = nsUserAccounts(server_a, DEFAULT_SUFFIX).create_test_user(uid=2000)
+    euuid_a = account_a.get_attr_vals_utf8('entryUUID')
+    print("🧩 %s" % euuid_a)
+    assert(euuid_a is not None)
+    assert(len(euuid_a) == 1)
+
+    repl.wait_for_replication(server_a, server_b)
+
+    account_b = nsUserAccounts(server_b, DEFAULT_SUFFIX).get("test_user_2000")
+    euuid_b = account_b.get_attr_vals_utf8('entryUUID')
+    print("🧩 %s" % euuid_b)
+
+    server_a.config.loglevel(vals=(ErrorLog.DEFAULT,))
+    server_b.config.loglevel(vals=(ErrorLog.DEFAULT,))
+
+    assert(euuid_b is not None)
+    assert(len(euuid_b) == 1)
+    assert(euuid_b == euuid_a)
+
+    account_b.set("description", "update")
+    repl.wait_for_replication(server_b, server_a)
+
+    euuid_c = account_a.get_attr_vals_utf8('entryUUID')
+    print("🧩 %s" % euuid_c)
+    assert(euuid_c is not None)
+    assert(len(euuid_c) == 1)
+    assert(euuid_c == euuid_a)
+
--- a/src/plugins/entryuuid/src/lib.rs
+++ b/src/plugins/entryuuid/src/lib.rs
@@ -30,6 +30,16 @@ slapi_r_search_callback_mapfn!(entryuuid, entryuuid_fixup_cb, entryuuid_fixup_ma
 fn assign_uuid(e: &mut EntryRef) {
     let sdn = e.get_sdnref();
 
+    // 🚧 safety barrier 🚧
+    if e.contains_attr("entryUUID") {
+        log_error!(
+            ErrorLevel::Trace,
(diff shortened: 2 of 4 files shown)


dirsrvtests/tests/suites/entryuuid/replicated_test.py
```diff
@@ -0,0 +1,59 @@
...
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import ldap
+import pytest
+import logging
+from lib389.topologies import topology_m2 as topo_m2
+from lib389.idm.user import nsUserAccounts, UserAccounts
```
mreynolds389: You don't need to import userAccounts

dirsrvtests/tests/suites/entryuuid/replicated_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+    account_a = nsUserAccounts(server_a, DEFAULT_SUFFIX).create_test_user(uid=2000)
+    euuid = account_a.get_attr_val_utf8('entryUUID')
+    assert(euuid is not None)
+
+    repl.wait_for_replication(server_a, server_b)
+
+    account_b = nsUserAccounts(server_b, DEFAULT_SUFFIX).get("test_user_2000")
+    euuid = account_b.get_attr_val_utf8('entryUUID')
+    assert(euuid is not None)
```
mreynolds389: Shouldn't this be checking for multiple entryUUID attributes?

Firstyear: Yeah, I can check that too. Actually, I guess if there were multiple then any future mods of the entry would fail too?

mreynolds389: IDK, it allows two attributes.  Schema checks "probably" happen before backend post op plugins are called.

Firstyear: The schema doesn't allow two attrs, and the plugin itself also checks if the value already exists. So I'm going to expand this test a bit today.

dirsrvtests/tests/suites/entryuuid/replicated_test.py
Firstyear: Good point.

dirsrvtests/tests/suites/entryuuid/replicated_test.py
Firstyear: Right expanding the test I have a reproducer. Looking at it now.

Firstyear:
@mreynolds389 This fixes the issue and the test case was able to reproduce.

mreynolds389:
Well shouldn't the entryuuid plugin simply ignore replication updates?  Many of our current plugins already do this.  There is an operation flag you can check in the pblock (SLAPI_IS_REPLICATED_OPERATION).  I think this is the correct way to fix the problem.  Because the plugin is always going to add an extra attribute for every replicated add operation, so that logging you added will occur constantly which makes it somewhat pointless (even though it is the "trace function call" log level).

Firstyear:
Thinking about it a bit, you are absolutely correct. I will change it. 

To justify why you are correct:

If we had say a 1.4.3 server, and TWO 1.4.4 servers, we do a tot-init from 1.4.3 to the 1.4.4 servers, then the pair of 1.4.4 servers will generate disparate entryuuids that would confictl and recreate this problem. Then if we join the two 1.4.4 servecr together, boom. 

So only doing this on non-repl ops is correct because we rely on the fixup task to add the uuid on a single node.

I'll update the PR. 

I did the "check if exists" because I was thinking about the two node scenario but thinking more and more about what you said, it's correct. Thanks mate :)

Firstyear:
@mreynolds389 Updated :)

mreynolds389 (approved):
Thanks!  Ack!

Firstyear:
```
   26c6a0b78..c5dfc280d  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

PR #4505, opened 2020-12-16
Issue #4504 - Fix pytest test_dsconf_replication_monitor

### LINKED ISSUE #4504: Test tests/suites/clu/repl_monitor_test.py is failing
Using master branch on fedora, the test: test_dsconf_replication_monitor is failing:
    assertion failure in step 3 when checking that the aliased string is in the result.

> check_value_in_log_and_reset(content_list, alias_content) suites/clu/repl_monitor_test.py:189:
... 
> assert item in file_content E AssertionError: assert 'Supplier: M1 (someHostName.someDomainName:39001)' in 'Run replication monitor with aliases option\ndsrc path: /root/.dsrc\ndsrc container path: /data/config/container.inf\...t Init Status: unavailable\nReap Active: 0\nReplication Status: In Synchronization\nReplication Lag Time: 00:00:00\n\n'

It is a pure testcase issue:
  The hostnames used to generate the aliases and the expected strings are based on the topology (while dsconf replication monitor output is not using these)
and because of some changes in lib389 frameworks master1 nows ends up with localhost.localdomain while the test expects a FQDN.
(Note: it is not the case for master2 because name is got from replication agreement while the first supplier name is computed from dsconf parameters)

So to check the alias functionality it would be more reliable to get the host names from the step 2 output
 (to be sure that the alias will match the host) then compute the alias string and the expected string

 - Platform: Fedora
 - Package and version: master 

1.  Install test framework
2. Run:   1minutetip -p PYTEST_TEST=suites/clu/repl_monitor_test.py fedora
3. See error in report (pytest is failing)

Test should pass

progier389:
Bug Description: pytest test_dsconf_replication_monitor fails.
   On steps 3 the test case check that aliases are rightly resolved. 
    The test fails because the strings to generate the aliases and the strings to verify the result are built from 
      the topology hostname. But due to change in the lib389 replication monitor now displays localhost.localdomain for the 
       initial supplier

Fix Description: Get hostname from Step 2 replication monitor output (that does not use aliases) then generate the strings
    using these hostname.

fixes: [#4504](https://github.com/389ds/389-ds-base/issues/4504)

Author: progier389

Review by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/repl_monitor_test.py
+++ b/dirsrvtests/tests/suites/clu/repl_monitor_test.py
@@ -9,6 +9,7 @@
 import time
 import subprocess
 import pytest
+import re
 
 from lib389.cli_conf.replication import get_repl_monitor_info
 from lib389.tasks import *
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
LGTM

droideck (approved):
Ack from me too! Thanks!

PR #4508, opened 2020-12-17
Issue 4507 - Improve csngen testing task

### LINKED ISSUE #4507: Improve csngen testing thread
**Issue Description**
A task was added to test csn generation with time skew.  This was just for testing, but it caused the CI test to run over 10 minutes,when it should really only takes less than 30 seconds to run.

The csngen task should be checking for server shutdown so it doesn't hang the server.

mreynolds389:
Description:  Once the csngen testing task is created, it will not stop for 10 minutes
              even if you attempt to stop the server.  This is adding 10 minutes to
              the CI testing runs.

              Improved this task to check for the server shutdown, an moved the csngen
              test to the bottom of the file so it is executed last so it does not
              interfere with other tests

Fixes: https://github.com/389ds/389-ds-base/issues/4507

Reviewed by: ?


Firstyear:
No problems for me, looks good. @tbordaz may want to check too though

progier389 (approved):
LGTM

tbordaz:
LGTM

Issue #4506, opened 2020-12-17
SASL GSSAPI & GSS-SPNEGO performance issue


tiran:
**Issue Description**

Today I noticed again that LDAP SASL bind with GSSAPI and GSS-SPNEGO are sometimes slow. The issue is similar to https://pagure.io/freeipa/issue/6656. In about half the cases SASL bind is faster. In other cases the operation is more than a magnitude slower. I'm seeing the same performance spikes with GSS-SPNEGO. In fast case GSS-SPNEGO is a bit faster than GSSAPI. In slow cases it degrades like GSSAPI.

**Package Version and Platform:**
 - Platform: Fedora 32
 - Package and version: 389-ds-base-2.0.1-20201215gitae2da1018.fc32.x86_64 and 1.4.3.16-1.fc32, cyrus-sasl-2.1.27-4.fc32.x86_64, krb5-workstation-1.18.2-29.fc32.x86_64

**Steps to Reproduce**
* Install FreeIPA server
* ``kinit -kt /etc/named.keytab DNS/$(hostname)``
* run ``time ldapwhoami -H ldapi://%2Frun%2Fslapd-IPA-TEST.socket -Y GSSAPI`` multiple times

**Expected results**
Command finishes in around 20ms consistently.

**Actual result**

Wall clock time for the command is all over the board. Sometimes the command finishes in less than 20ms, sometimes it's around 250 and every know and then it takes more than 500ms.

```
real    0m0,030s
real    0m0,018s
real    0m0,017s
real    0m0,266s
real    0m0,267s
real    0m0,265s
real    0m0,017s
real    0m0,266s
real    0m0,264s
real    0m0,513s
real    0m0,266s
real    0m0,267s
real    0m0,017s
real    0m0,264s
real    0m0,266s
real    0m0,017s
real    0m0,266s
real    0m0,018s
real    0m0,264s
real    0m0,265s
real    0m0,016s
real    0m0,265s
real    0m0,266s
real    0m0,267s
real    0m0,267s
real    0m0,516s
real    0m0,018s
real    0m0,017s
real    0m0,266s
```

**Additional context**

For testing I have stop all IPA services except DS, disabled and removed all IPA plugins that may affect 389-DS performance, and restarted 389-DS. It did not affect performance.

fast bind
```
    [17/Dec/2020:12:18:24.059835167 +0100] conn=286 fd=109 slot=109 connection from local to /var/run/slapd-IPA-TEST.socket
[...]

PR #4510, opened 2020-12-18
Trace more SASL calls

tiran:
Add log trace for sasl_server_start() and sasl_server_step() to
investigate performance issues with SASL bind.

See: https://github.com/389ds/389-ds-base/issues/4506
Signed-off-by: Christian Heimes <cheimes@redhat.com>

### DIFF
--- a/ldap/servers/slapd/saslbind.c
+++ b/ldap/servers/slapd/saslbind.c
@@ -752,6 +752,54 @@ ids_sasl_server_new(Connection *conn)
 }
 
 /*
+ * start a sasl server connection
+ */
+int
+ids_sasl_server_start(Connection *conn, const char *mech,
+                      struct berval *cred,
+                      const char **sdata, unsigned int *slen)
+{
+    int rc;
+    sasl_conn_t *sasl_conn;
+
+    slapi_log_err(SLAPI_LOG_TRACE, "ids_sasl_server_start", "=> (mech=%s)\n",
+                  mech);
+
+    sasl_conn = (sasl_conn_t *)conn->c_sasl_conn;
+    rc = sasl_server_start(sasl_conn, mech,
+                           cred->bv_val, cred->bv_len,
+                           sdata, slen);
+
+    slapi_log_err(SLAPI_LOG_TRACE, "ids_sasl_server_start", "<= (rc=%s)\n",
+                  sasl_errstring(rc, NULL, NULL));
+
+    return rc;
+}
+
+/*
+ * perform a sasl server step
+ */
+int
+ids_sasl_server_step(Connection *conn, struct berval *cred,
+                     const char **sdata, unsigned int *slen)
+{
+    int rc;
+    sasl_conn_t *sasl_conn;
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/saslbind.c
```diff
@@ -945,9 +993,7 @@ ids_sasl_check_bind(Slapi_PBlock *pb)
             goto sasl_start;
         }
 
-        rc = sasl_server_step(sasl_conn,
-                              cred->bv_val, cred->bv_len,
-                              &sdata, &slen);
+        rc = ids_sasl_server_step(pb_conn, cred, &sdata, &slen);
```
Firstyear, line 996: Depending on what we want to do here, we could add these in #ifdef DEBUG.

mreynolds389:
This is somewhat off topic and directed at the Dev team, but we have a problem with Trace Function Call logging level.  It basically makes the server unusable.  The server start up will fail because it times in when this logging level is enabled, and same for other processes.  It's been over-used in the source, and I do not what to use that logging level in this PR.  

@tiran maybe in this case connection logging is the next best choice (SLAPI_LOG_CONNS).

For Dev, we either need new logging levels or we need to change/remove some of the current "trace function call" logging statements.  They need to be reduced, one way or another...

Firstyear:
@mreynolds389 Well, we need to do a lot related to logging in general, but this is a good point for certain.

mreynolds389:
@tiran if you change the log level from SLAPI_LOG_TRACE to SLAPI_LOG_CONNS you get my ack.  

@Firstyear I don't think we should #ifdef this.  This logging could be useful in customer situations, so it should be available with the standard build.

ldap/servers/slapd/saslbind.c
```diff
@@ -751,6 +751,54 @@ ids_sasl_server_new(Connection *conn)
...
+int
+ids_sasl_server_start(Connection *conn, const char *mech,
+                      struct berval *cred,
+                      const char **sdata, unsigned int *slen)
+{
+    int rc;
+    sasl_conn_t *sasl_conn;
+
+    slapi_log_err(SLAPI_LOG_TRACE, "ids_sasl_server_start", "=> (mech=%s)\n",
+                  mech);
```
mreynolds389, line 766: I think we should also log the connection ID in the logging (before and after):

    slapi_log_err(SLAPI_LOG_CONNS, "ids_sasl_server_start", "=> (conn=%" PRIu64 " mech=%s)\n",
                  conn->c_connid, mech);
    ...
    slapi_log_err(SLAPI_LOG_CONNS, "ids_sasl_server_start", "<= (conn=%" PRIu64 " rc=%s)\n",
                  conn->c_connid, sasl_errstring(rc, NULL, NULL));

mreynolds389:
I took this PR and created a new one with more logging improvements:  https://github.com/389ds/389-ds-base/pull/4749

I'm going to close this PR in favor of the new one.  Thanks @tiran !

Issue #4513, opened 2020-12-22
Fix CI test failures so suites pass 100%


droideck:
**Issue Description**
`389-ds-base/dirsrvtests/tests/suites/` test cases are failing and we need to fix them.

**Package Version and Platform:**
 - Platform: Fedora 33
 - Package and version: 2.0.1-20201221git0f38410ae.fc33

droideck:
The list of tests that FAILED (there are still a bunch of tests that were skipped and should be checked separately).

```
dirsrvtests/tests/suites/acl/keywords_part2_test.py::test_connectin_from_an_unauthorized_network FAILED
dirsrvtests/tests/suites/basic/basic_test.py::test_bind_invalid_entry FAILED
dirsrvtests/tests/suites/entryuuid/basic_test.py::test_entryuuid_indexed_import_and_search FAILED
dirsrvtests/tests/suites/entryuuid/basic_test.py::test_entryuuid_unindexed_import_and_search FAILED
dirsrvtests/tests/suites/entryuuid/basic_test.py::test_entryuuid_generation_on_add FAILED
dirsrvtests/tests/suites/entryuuid/basic_test.py::test_entryuuid_fixup_task FAILED
dirsrvtests/tests/suites/filter/rfc3673_all_oper_attrs_test.py::test_search_basic[uid=all_attrs_test,ou=people,dc=example,dc=com-False-oper_attr_list4-] FAILED
dirsrvtests/tests/suites/filter/rfc3673_all_oper_attrs_test.py::test_search_basic[uid=all_attrs_test,ou=people,dc=example,dc=com-False-oper_attr_list4-objectClass] FAILED
dirsrvtests/tests/suites/filter/rfc3673_all_oper_attrs_test.py::test_search_basic[uid=all_attrs_test,ou=people,dc=example,dc=com-True-oper_attr_list5-] FAILED
dirsrvtests/tests/suites/filter/rfc3673_all_oper_attrs_test.py::test_search_basic[uid=all_attrs_test,ou=people,dc=example,dc=com-True-oper_attr_list5-objectClass] FAILED
dirsrvtests/tests/suites/import/import_test.py::test_fast_slow_import FAILED
dirsrvtests/tests/suites/lib389/config_compare_test.py::test_config_compare FAILED
dirsrvtests/tests/suites/lib389/idm/user_compare_i2_test.py::test_user_compare_i2 FAILED
[...]

PR #4514, opened 2020-12-23
Issue 4513 - Fix schema test and lib389 task module

droideck:
Description: Fix the asssertion in schema_test.py.
Make sure that all of the tasks are up to date
with the recent changes in the task API.

Relates: #4513

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/clu/schema_test.py
+++ b/dirsrvtests/tests/suites/clu/schema_test.py
@@ -32,7 +32,7 @@ def test_origins_with_extra_parenthesis(topo):
         'names': [ATTR_NAME],
         'oid': '1.1.1.1.1.1.1.22222',
         'desc': 'Test extra parenthesis in X-ORIGIN',
-        'x_origin': X_ORG_VAL,
+        'x_origin': [X_ORG_VAL],
         'syntax': '1.3.6.1.4.1.1466.115.121.1.15',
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/clu/schema_test.py
```diff
@@ -52,12 +52,11 @@ def test_origins_with_extra_parenthesis(topo):
     attr_result = schema.query_attributetype(ATTR_NAME, json=True)
 
     # Verify the x-origin value is correct
-    assert attr_result['at']['x_origin'][0] == X_ORG_VAL
+    assert attr_result['at']['x_origin'][0][0] == X_ORG_VAL
```
mreynolds389: This just seems wrong to me.  I recently added this test case to test the fix for the embedded parenthesis, and to have this double nested subscription does not look right.  What does the dict look like to require this?  I'm just wondering if this is something that needs to be fixed in lib389 instead (maybe a regression was introduced in lib389)?

droideck: Right... I am not sure what made me do the fix like this. 
Please, check another fix :)

Firstyear:
I think it looks fine to me now, but I'll let @mreynolds389 check to be sure.

mreynolds389 (approved):
LGTM

PR #4516, opened 2021-01-04
Issue 4506 - Temporary fix for io issues

### LINKED ISSUE #4506: SASL GSSAPI & GSS-SPNEGO performance issue

Firstyear:
Bug Description: Due to how we polled accept and connections
we would interleave these causing possible delays to new
accepts, or delays to connections awaiting work.

Fix Description: Split accept into it's own thread.

fixes: https://github.com/389ds/389-ds-base/issues/4506

Review by: ???

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -58,11 +58,14 @@
 
 #if defined(LDAP_IOCP)
 #define SLAPD_WAKEUP_TIMER 250
+#define SLAPD_ACCEPT_WAKEUP_TIMER 250
 #else
 #define SLAPD_WAKEUP_TIMER 250
+#define SLAPD_ACCEPT_WAKEUP_TIMER 250
 #endif
 
 int slapd_wakeup_timer = SLAPD_WAKEUP_TIMER; /* time in ms to wakeup */
+int slapd_accept_wakeup_timer = SLAPD_ACCEPT_WAKEUP_TIMER; /* time in ms to wakeup */
 #ifdef notdef                                /* GGOODREPL */
 /*
  * time in secs to do housekeeping:
@@ -77,6 +80,7 @@ static int readsignalpipe = SLAPD_INVALID_SOCKET;
 #define FDS_SIGNAL_PIPE 0
 
 static PRThread *disk_thread_p = NULL;
+static PRThread *accept_thread_p = NULL;
 static pthread_cond_t diskmon_cvar;
 static pthread_mutex_t diskmon_mutex;
 
@@ -95,6 +99,7 @@ typedef struct listener_info
 
 static size_t listeners = 0;                /* number of listener sockets */
 static listener_info *listener_idxs = NULL; /* array of indexes of listener sockets in the ct->fd array */
+static PRFileDesc *tls_listener = NULL; /* Stashed tls listener for get_ssl_listener_fd */
 
 #define SLAPD_POLL_LISTEN_READY(xxflagsxx) (xxflagsxx & PR_POLL_READ)
 
@@ -109,7 +114,9 @@ static PRFileDesc **createprlistensockets(unsigned short port,
                                           int local);
 static const char *netaddr2string(const PRNetAddr *addr, char *addrbuf, size_t addrbuflen);
 static void set_shutdown(int);
-static void setup_pr_read_pds(Connection_Table *ct, PRFileDesc **n_tcps, PRFileDesc **s_tcps, PRFileDesc **i_unix, PRIntn *num_to_read);
+static void setup_pr_ct_firsttime_pds(Connection_Table *ct);
+static PRIntn setup_pr_accept_pds(PRFileDesc **n_tcps, PRFileDesc **s_tcps, PRFileDesc **i_unix, struct POLL_STRUCT **fds);
+static PRIntn setup_pr_read_pds(Connection_Table *ct);
 
 #ifdef HPUX10
 static void *catch_signals();
@@ -127,7 +134,7 @@ static int
 accept_and_configure(int s __attribute__((unused)), PRFileDesc *pr_acceptfd, PRNetAddr *pr_netaddr, int addrlen __attribute__((unused)), int secure, int local, PRFileDesc **pr_clonefd)
 {
     int ns = 0;
-    PRIntervalTime pr_timeout = PR_MillisecondsToInterval(slapd_wakeup_timer);
+    PRIntervalTime pr_timeout = PR_MillisecondsToInterval(slapd_accept_wakeup_timer);
 
     (*pr_clonefd) = PR_Accept(pr_acceptfd, pr_netaddr, pr_timeout);
     if (!(*pr_clonefd)) {
@@ -755,27 +762,98 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
 }
 
 static void
-handle_listeners(Connection_Table *ct)
+handle_listeners(struct POLL_STRUCT *fds)
 {
+    Connection_Table *ct = the_connection_table;
     size_t idx;
     for (idx = 0; idx < listeners; ++idx) {
         int fdidx = listener_idxs[idx].idx;
         PRFileDesc *listenfd = listener_idxs[idx].listenfd;
         int secure = listener_idxs[idx].secure;
         int local = listener_idxs[idx].local;
-        if (fdidx && listenfd) {
-            if (SLAPD_POLL_LISTEN_READY(ct->fd[fdidx].out_flags)) {
+        if (listenfd) {
+            PR_ASSERT(fds != NULL);
+            PR_ASSERT(listenfd == fds[fdidx].fd);
+            if (SLAPD_POLL_LISTEN_READY(fds[fdidx].out_flags)) {
                 /* accept() the new connection, put it on the active list for handle_pr_read_ready */
                 int rc = handle_new_connection(ct, SLAPD_INVALID_SOCKET, listenfd, secure, local, NULL);
                 if (rc) {
                     slapi_log_err(SLAPI_LOG_CONNS, "handle_listeners", "Error accepting new connection listenfd=%d\n",
                                   PR_FileDesc2NativeHandle(listenfd));
                     continue;
+                } else {
+                    /* Wake up the main event loop to handle this immediately. */
+                    signal_listner();
                 }
             }
         }
     }
     return;
+}
+
+void
+accept_thread(void *vports)
+{
+    daemon_ports_t *ports = (daemon_ports_t *)vports;
+    Connection_Table *ct = the_connection_table;
+    PRIntn num_poll = 0;
+    struct POLL_STRUCT *fds = NULL;
+    int select_return = 0;
+    PRErrorCode prerr;
+    int last_accept_new_connections = -1;
+    PRIntervalTime pr_timeout = PR_MillisecondsToInterval(slapd_accept_wakeup_timer);
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+
+    PRFileDesc **n_tcps = NULL;
+    PRFileDesc **s_tcps = NULL;
+    PRFileDesc **i_unix = NULL;
+    n_tcps = ports->n_socket;
+    s_tcps = ports->s_socket;
+#if defined(ENABLE_LDAPI)
+    i_unix = ports->i_socket;
+#endif /* ENABLE_LDAPI */
+
+    num_poll = setup_pr_accept_pds(n_tcps, s_tcps, i_unix, &fds);
+
+    while (!g_get_shutdown()) {
+        /* Do we need to accept new connections? */
+        int accept_new_connections = ((ct->size - g_get_current_conn_count()) > slapdFrontendConfig->reservedescriptors);
+        last_accept_new_connections = accept_new_connections;
+        if (!accept_new_connections) {
+            if (last_accept_new_connections) {
+                slapi_log_err(SLAPI_LOG_ERR, "accept_thread",
+                              "Not listening for new connections - too many fds open\n");
+            }
+            /* Need a sleep delay here. */
+            PR_Sleep(pr_timeout);
+            continue;
+        } else {
+            /* Log that we are now listening again */
+            if (!last_accept_new_connections && last_accept_new_connections != -1) {
+                slapi_log_err(SLAPI_LOG_ERR, "accept_thread",
+                              "Listening for new connections again\n");
+            }
+        }
+
+        select_return = POLL_FN(fds, num_poll, pr_timeout);
+        switch (select_return) {
+        case 0: /* Timeout */
+            break;
+        case -1: /* Error */
+            prerr = PR_GetError();
+            slapi_log_err(SLAPI_LOG_TRACE, "accept_thread", "PR_Poll() failed, " SLAPI_COMPONENT_NAME_NSPR " error %d (%s)\n",
+                          prerr, slapd_system_strerror(prerr));
+            break;
+        default: /* a new connection */
+            handle_listeners(fds);
+            break;
+        }
+    }
+
+    /* free the listener indexes */
+    slapi_ch_free((void **)&listener_idxs);
+    slapd_sockets_ports_free(ports);
+    slapi_ch_free((void **)&fds);
 }
 
 void
@@ -993,8 +1071,24 @@ slapd_daemon(daemon_ports_t *ports)
     /* Now we write the pid file, indicating that the server is finally and listening for connections */
     write_pid_file();
 
+    /* Prepare the CT for first use */
+    setup_pr_ct_firsttime_pds(the_connection_table);
+
     /* The server is ready and listening for connections. Logging "slapd started" message. */
     unfurl_banners(the_connection_table, ports, n_tcps, s_tcps, i_unix);
+
+    /* Create a thread to accept new connections */
+    accept_thread_p = PR_CreateThread(PR_SYSTEM_THREAD,
+                                    (VFP)(void *)accept_thread, (void*)ports,
+                                    PR_PRIORITY_NORMAL, PR_GLOBAL_THREAD,
+                                    PR_JOINABLE_THREAD,
+                                    SLAPD_DEFAULT_THREAD_STACKSIZE);
+    if (NULL == accept_thread_p) {
+        PRErrorCode errorCode = PR_GetError();
+        slapi_log_err(SLAPI_LOG_EMERG, "slapd_daemon", "Unable to fd accept thread - Shutting Down (" SLAPI_COMPONENT_NAME_NSPR " error %d - %s)\n",
+                      errorCode, slapd_pr_strerror(errorCode));
+        g_set_shutdown(SLAPI_SHUTDOWN_EXIT);
+    }
 
 #ifdef WITH_SYSTEMD
     sd_notifyf(0, "READY=1\n"
@@ -1008,7 +1102,7 @@ slapd_daemon(daemon_ports_t *ports)
         int select_return = 0;
         PRErrorCode prerr;
 
-        setup_pr_read_pds(the_connection_table, n_tcps, s_tcps, i_unix, &num_poll);
+        num_poll = setup_pr_read_pds(the_connection_table);
         select_return = POLL_FN(the_connection_table->fd, num_poll, pr_timeout);
         switch (select_return) {
         case 0: /* Timeout */
@@ -1018,9 +1112,10 @@ slapd_daemon(daemon_ports_t *ports)
             slapi_log_err(SLAPI_LOG_TRACE, "slapd_daemon", "PR_Poll() failed, " SLAPI_COMPONENT_NAME_NSPR " error %d (%s)\n",
                           prerr, slapd_system_strerror(prerr));
             break;
-        default: /* either a new connection or some new data ready */
+        default: /* some new data ready */
             /* handle new connections from the listeners */
-            handle_listeners(the_connection_table);
+            // Now done in accept_thread
+            // handle_listeners(the_connection_table);
             /* handle new data ready */
             handle_pr_read_ready(the_connection_table, connection_table_size);
             clear_signal(the_connection_table->fd);
@@ -1039,11 +1134,6 @@ slapd_daemon(daemon_ports_t *ports)
     if (!in_referral_mode) {
         ps_stop_psearch_system(); /* stop any persistent searches */
     }
-
-    /* free the listener indexes */
-    slapi_ch_free((void **)&listener_idxs);
-
-    slapd_sockets_ports_free(ports);
 
     op_thread_cleanup();
     housekeeping_stop(); /* Run this after op_thread_cleanup() logged sth */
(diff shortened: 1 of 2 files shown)


Firstyear:
@tiran here is a hacky-test fix where I split accept to a seperate thread. Can you test it with your SASL test? I've already run the basic tests and confirmed it works.

tiran:
The patch does not help. It also slows down 389-DS a lot. My test script takes 7 times (16s -> 107s) more time to complete 200 connections.

``389-ds-base-2.0.1-20210104git0f38410ae.fc32.x86_64`` without patch:

```
6ms     3
7ms     3
8ms     74
9ms     48
10ms    7
11ms    4
12ms    2
13ms    2
51ms    1
69ms    1
109ms   2
116ms   1
132ms   1
255ms   7
256ms   7
257ms   21
258ms   10
259ms   4
504ms   1
508ms   1
Total: 16s
```

``389-ds-base-2.0.1-20210104git0f38410ae.fc32.x86_64`` with #4516 patch

```
7ms     5
8ms     45
9ms     77
10ms    21
11ms    1
12ms    4
15ms    1
255ms   2
256ms   5
257ms   20
258ms   13
259ms   1
260ms   1
261ms   1
505ms   2
507ms   1
Total: 107s
```

Firstyear:
@tiran and @tbordaz can you check this again? This makes the accept thread in a seperate thread, and I can see an approx 10% speed up with ldclt -e bindeach (new connections each search). Hopefully this also allows us to identify which half of the socket parts are a problem (or it will just resolve it :) ) 

@progier389 cc because you probably will be interested as it relates a bit to the connection bound check,

Passes with asan, having issues with latest suse to run lsan so I'll need to check that later.

tiran:
My unscientific test script takes roughly 5% more time to finish with your PR. I also noticed that the distribution looks different with 2.0.1. There is another peak at multiple of 100ms.

## latest PR 389-ds-base-2.0.1-20210108gitad13f0ded

```
6ms     7
7ms     14
8ms     835
9ms     390
10ms    50
11ms    10
12ms    1
13ms    3
15ms    2
16ms    3
17ms    1
25ms    1
37ms    1
62ms    1
105ms   1
106ms   8
107ms   67
108ms   98
109ms   20
110ms   3
112ms   1
114ms   2
115ms   1
159ms   1
164ms   1
174ms   1
184ms   1
187ms   1
206ms   1
207ms   4
254ms   8
255ms   72
256ms   18
257ms   174
258ms   119
259ms   16
260ms   5
261ms   2
355ms   4
356ms   12
357ms   8
358ms   5
359ms   1
409ms   1
503ms   8
504ms   8
507ms   4
508ms   2
604ms   2
605ms   1
Total: 168s
```

## nightly build 389-ds-base-2.0.1-20210108git65678bb3b

```
6ms     9
7ms     18
8ms     916
9ms     322
10ms    44
11ms    10
13ms    3
15ms    2
16ms    1
18ms    1
25ms    1
88ms    1
105ms   3
106ms   9
107ms   95
108ms   111
109ms   35
110ms   1
113ms   1
206ms   1
207ms   2
208ms   4
209ms   1
254ms   10
255ms   50
256ms   22
257ms   148
258ms   95
259ms   17
260ms   4
262ms   1
264ms   1
265ms   2
354ms   1
355ms   1
356ms   10
357ms   11
358ms   7
359ms   3
360ms   2
503ms   2
504ms   12
505ms   2
507ms   2
508ms   1
607ms   2
753ms   2
754ms   1
Total: 159s
```

progier389:
Several comments here:
   [1] About the changes:
      If you split the polling threads you need to be careful with the signal_pipe handling. 
      It should be polled in the original thread (to interrupt the poll when polled fds must be added/removed) 
      I am not sure whether a second signal_pipe is needed for the accept poll (depends if the listen fd list may
         dynamically change or not)
    [2] About the issue itself.
           I can think about two possible reasons that could explain that the timout change helps:
           [a]:  That PR_Accept spent long time ( why ? tcpip tuning issue ? ) 
           [b]:   That somewhere in the code a call to signal_listner has been forgotten and some change
                    is taken in account only after the poll timeout expiration
                     (b is nor very plausible because it is a stress test and the poll always get triggered in less than a few ms anyway)
           [c]:   I am not really surprised that splitting the poll thread leads to better performance, 
                       (even without splitting the accept/versus read event)

Firstyear:
> My unscientific test script takes roughly 5% more time to finish with your PR. I also noticed that the distribution looks different with 2.0.1. There is another peak at multiple of 100ms.

That's fine - but now if you want, you can adjust:

```
64 #define SLAPD_WAKEUP_TIMER 250
65 #define SLAPD_ACCEPT_WAKEUP_TIMER 250
```

Because these are now independent, you can adjust this to help bisect which poll we are delaying on. I'd suggest turning ACCEPT_WAKEUP higher, maybe 1000 and WAKEUP down to say 25. That way you can tell if it gets faster, we know it's in the WAKEUP poll, if it gets MUCH slower, then we know it's in the ACCEPT part. 

> Several comments here:
> [1] About the changes:
> If you split the polling threads you need to be careful with the signal_pipe handling.
> It should be polled in the original thread (to interrupt the poll when polled fds must be added/removed)
> I am not sure whether a second signal_pipe is needed for the accept poll (depends if the listen fd list may
> dynamically change or not)

Yep, I was pretty careful about this. I checked all the locations where the signal pipe is woken from, and it's all about existing connections, not accept, which is why I have left the signal pipe as fd[0] in the ct->fd array, so that it still correctly wakes the main event loop when required. I also added a called to wake the signal pipe from the accept thread so that the connection can immediately be read from. 

> [2] About the issue itself.
> I can think about two possible reasons that could explain that the timout change helps:
> [a]: That PR_Accept spent long time ( why ? tcpip tuning issue ? )
> [b]: That somewhere in the code a call to signal_listner has been forgotten and some change
> is taken in account only after the poll timeout expiration

Yep these are all valid possibilities! Additional possibilities:

* We only add a connection fd to be polled if we can successfully lock (we use try lock), so the conn may be locked causing it's initial read to be skipped.
* It could be a subsequent read/partial read where we time out won't re-poll again because trylock
* Because this test is effectively single threaded client, we don't see this in normal conditions as we rarely hit poll timeout as "other events" are waking us and masking the delay.

So this is another benefit to the split, is as mentioned earlier in my reply, we can now tune the poll timeouts separately to see which half has the problem (accept or connection). From there, the next step becomes additional logging of the connection state I think once we know which side it's on. 

> (b is nor very plausible because it is a stress test and the poll always get triggered in less than a few ms anyway)
> [c]: I am not really surprised that splitting the poll thread leads to better performance,
> (even without splitting the accept/versus read event)

Yes, but it's better multi thread / multiple connection performance :) it's been on our TODO for a long time to split accept from connection events. So I think even if this doesn't solve the problem directly, it's going to make it easier to isolate the problem *AND* it causes better performance in the general case :) 

Thanks for your review!

progier389:
> > Several comments here:
> > [1] About the changes:
> > If you split the polling threads you need to be careful with the signal_pipe handling.
> > It should be polled in the original thread (to interrupt the poll when polled fds must be added/removed)
> > I am not sure whether a second signal_pipe is needed for the accept poll (depends if the listen fd list may
> > dynamically change or not)
> 
> Yep, I was pretty careful about this. I checked all the locations where the signal pipe is woken from, and it's all about existing connections, not accept, which is why I have left the signal pipe as fd[0] in the ct->fd array, so that it still correctly wakes the main event loop when required. I also added a called to wake the signal pipe from the accept thread so that the connection can immediately be read from.
OK: My mistake, I reacted because I did not saw the signal pipe initialization in setup_pr_accept_pds as I expected but I did not realized it is now initilazed in setup_pr_ct_firsttime_pds (which is fine) 
> 
> > [2] About the issue itself.
> > I can think about two possible reasons that could explain that the timeout change helps:
> > [a]: That PR_Accept spent long time ( why ? tcpip tuning issue ? )
> > [b]: That somewhere in the code a call to signal_listner has been forgotten and some change
> > is taken in account only after the poll timeout expiration
> 
> Yep these are all valid possibilities! Additional possibilities:
> 
> * We only add a connection fd to be polled if we can successfully lock (we use try lock), so the conn may be locked causing it's initial read to be skipped.
> * It could be a subsequent read/partial read where we time out won't re-poll again because trylock
Interesting remark! 
If one of these two points is causing the problem, that would mean that a call to signal_listner is missing when the connection lock is released. 
   I wonder if we should not then call signal_listner whenever releasing the connection lock ...
   But if we do that we should improve signal_listner (should probably use a variable and atomic operation) to insure 
      that the pipe is written only once per poll loop (otherwise we may get in trouble with the pipe flow control)
> * Because this test is effectively single threaded client, we don't see this in normal conditions as we rarely hit poll timeout as "other events" are waking us and masking the delay.
> 
> So this is another benefit to the split, is as mentioned earlier in my reply, we can now tune the poll timeouts separately to see which half has the problem (accept or connection). From there, the next step becomes additional logging of the connection state I think once we know which side it's on.
> 
> > (b is nor very plausible because it is a stress test and the poll always get triggered in less than a few ms anyway)
> > [c]: I am not really surprised that splitting the poll thread leads to better performance,
> > (even without splitting the accept/versus read event)
> 
> Yes, but it's better multi thread / multiple connection performance :) it's been on our TODO for a long time to split accept from connection events. So I think even if this doesn't solve the problem directly, it's going to make it easier to isolate the problem _AND_ it causes better performance in the general case :)
I fully agree that your patch moves things in the right direction. ([c] is just meaning that having several listening threads should 
  increase the performance (whether splitting the accept from the read events or not). That said, having a dedicated thread
    for accept events has the merit of making the code easier to understand. So although I believe that the root cause of this
     problem is not related to the mix of listening and read events, IMHO we should keep and commit your patch. )
> 
> Thanks for your review!

progier389:
About signal_pipe fd handling:
OK, that is my mistake! I reacted because I did not saw the signal pipe initialization in setup_pr_accept_pds as I expected but I did not realized it is now initilazed in setup_pr_ct_firsttime_pds (which is fine)

About the additional possibilities:
Interesting remark!
The fact that fd are skiped if try lock fails means that a call to signal_listner is missing when the connection lock is released.
But if we fix that we should improve signal_listner (should probably use a variable and atomic operation) to insure
that the pipe is written only once per poll loop (otherwise we may fill up the pipe and get in trouble with the pipe flow control)

About
`Yes, but it's better multi thread / multiple connection performance :) it's been on our TODO for a long time to split accept from connection events. So I think even if this doesn't solve the problem directly, it's going to make it easier to isolate the problem _AND_ it causes better performance in the general case :)`
I fully agree that your patch moves things in the right direction. ([c] is just meaning that having several listening threads should
increase the performance (whether splitting the accept from the read events or not). That said, having a dedicated thread
for accept events has the merit of making the code easier to understand. So although I believe that the root cause of this
problem is not related to the mix of listening and read events, IMHO we should keep and commit your patch. )

Firstyear:
> About signal_pipe fd handling:
> OK, that is my mistake! I reacted because I did not saw the signal pipe initialization in setup_pr_accept_pds as I expected but I did not realized it is now initilazed in setup_pr_ct_firsttime_pds (which is fine)

I've improved the comment in setup_pr_read_pds to highlight how the signal pipe fd is setup.

> 
> About the additional possibilities:
> Interesting remark!
> The fact that fd are skiped if try lock fails means that a call to signal_listner is missing when the connection lock is released.
> But if we fix that we should improve signal_listner (should probably use a variable and atomic operation) to insure
> that the pipe is written only once per poll loop (otherwise we may fill up the pipe and get in trouble with the pipe flow control)

It took me a while to think about this one. In some ways I completely agree, in some I don't. I think that if we were to call the signal_listener after each connection unlock, there are 95 locations in the code base we would need to adjust for this to occur correctly. We could target specific unlock calls, especially focussed around IO paths so that we call the listener on the unlock. connection_threadmain would probably be the primary focus for this, I think. That way when we have an incomplete PDU we prepare the connection immediately to be re-polled if not in turbo mode (though this could make turbo mode un-needed?). I suspect this would be from ret == CONN_TIMEOUT around line 1617 in connection.c. But I'd need to check more thoroughly to be sure. Saying that, even with this change, it doesn't guarantee that when the main event loop checks the connection, that it will be unlocked because another worker may have then taken the lock to do a write. 

Saying this, this also does bring a "risk" that we would be continually waking and rebuilding the fd table as connections complete their reads, so we would then potentially add excess load into the event loop thread. It's a tricky balance ... and that's where my thought to disagree comes in, that perhaps the "simpler" solution is just to reduce the main event loop timeout from 250ms down to 25 (?) since that would have the "same effect" as the event-driven solution you are proposing. 

If I had infinite time I would rewrite connection handling from scratch, but I do not have infinite time :) 

> I fully agree that your patch moves things in the right direction. ([c] is just meaning that having several listening threads should
> increase the performance (whether splitting the accept from the read events or not). That said, having a dedicated thread
> for accept events has the merit of making the code easier to understand. So although I believe that the root cause of this
> problem is not related to the mix of listening and read events, IMHO we should keep and commit your patch. )

Thanks, I appreciate the vote of confidence! I think it would be good to get some feedback from @tiran once he has some time to retest with adjusted wakeup times so we can help narrow down this search, and to give a bit of time for @tbordaz and @mreynolds389 to check this since connection handling is a very important area of the code. The more eyes the better :)

Firstyear:
@mreynolds389 @progier389 Can you please review this - Talking with @progier389 we think this change is valuable regardless of the possible benefit to @tiran, and it does give them more options for testing to isolate the problematic wait timer. It's now in a state I think we can merge it. Thanks!

ldap/servers/slapd/daemon.c
```diff
@@ -1202,133 +1292,117 @@ clear_signal(struct POLL_STRUCT *fds)
...
+        for (fdesc = s_tcps; fdesc && *fdesc; fdesc++, count++) { }
+    }
+#if defined(ENABLE_LDAPI)
+    if (i_unix != NULL) {
+        PRFileDesc **fdesc = NULL;
+        for (fdesc = i_unix; fdesc && *fdesc; fdesc++, count++) { }
+    }
+#endif
 
+    struct POLL_STRUCT *myfds = (struct POLL_STRUCT *)slapi_ch_calloc(1, (count + 1) * sizeof(struct POLL_STRUCT));
```
progier389: A minor point: Could you move myfds declaration to the top of the function (its scope covers the whole function)

progier389 (approved):
ACK for me once the minor point get fixed.

mreynolds389:
please remove java style comments ;-)

Firstyear:
> please remove java style comments ;-)

I was *literally* writing the cleanup of the comments as you sent that msg mate. On the other hand ... almost snuck them past you :P

ldap/servers/slapd/daemon.c
```diff
@@ -993,9 +1071,25 @@ slapd_daemon(daemon_ports_t *ports)
...
 
+    /* Prepare the CT for first use */
+    setup_pr_ct_firsttime_pds(the_connection_table);
+
     /* The server is ready and listening for connections. Logging "slapd started" message. */
     unfurl_banners(the_connection_table, ports, n_tcps, s_tcps, i_unix);
 
+    /* Create a thread to accept new connections */
+    accept_thread_p = PR_CreateThread(PR_SYSTEM_THREAD,
+                                    (VFP)(void *)accept_thread, (void*)ports,
```
mreynolds389: sorry, indentation

ldap/servers/slapd/daemon.c
```diff
@@ -1202,133 +1294,119 @@ clear_signal(struct POLL_STRUCT *fds)
...
-                slapi_log_err(SLAPI_LOG_HOUSE,
-                              "setup_pr_read_pds", "Listening for LDAPI connections on %d\n", socketdesc);
-            }
-        } else {
-            ct->fd[count].fd = NULL;
-            count++;
+    if (i_unix != NULL) {
+        PRFileDesc **fdesc = NULL;
+        for (fdesc = i_unix; fdesc && *fdesc; fdesc++, count++) {
+            myfds[count].fd = *fdesc;
```
mreynolds389, line 1366: indentation here

ldap/servers/slapd/daemon.c
mreynolds389: besides that, ack!

PR #4518, opened 2021-01-05
Issue 4517 - BUG: Multiple systemd pin warnings

### LINKED ISSUE #4517: systemd pin warning appears multiple times
During startup we show a systemd-ask-pass warning for each entry that is a child of `cn=encryption,cn=config`. This can create a lot of log noise that may not be relevant.

The log message should be outside of this loop and only emitted once.

Firstyear:
Bug Description: When multiple entries exist under
cn=encryption,cn=config then we log a warning for each
entry that systemd ask pass may be needed. This creates noise
when the warning is needed once.

Fix Description: Move the warning to outside the loop.

fixes: https://github.com/389ds/389-ds-base/issues/4517

Review by: ???

### DIFF
--- a/ldap/servers/slapd/ssl.c
+++ b/ldap/servers/slapd/ssl.c
@@ -1251,6 +1251,11 @@ slapd_ssl_init()
         slapd_extract_cert(entry, PR_TRUE);
     }
 
+#ifdef WITH_SYSTEMD
+    slapd_SSL_warn("Sending pin request to SVRCore. You may need to run"
+        " systemd-tty-ask-password-agent to provide the password if pin.txt does not exist.");
+#endif
+
     if ((family_list = getChildren(configDN))) {
         char **family;
         char *token;
@@ -1305,11 +1310,7 @@ slapd_ssl_init()
                 slapi_ch_free((void **)&token);
                 return -1;
             }
-/* authenticate */
-#ifdef WITH_SYSTEMD
-            slapd_SSL_warn("Sending pin request to SVRCore. You may need to run"
-                           " systemd-tty-ask-password-agent to provide the password.");
-#endif
+            /* authenticate */
             if (slapd_pk11_authenticate(slot, PR_TRUE, NULL) != SECSuccess) {
                 errorCode = PR_GetError();
                 slapi_log_err(SLAPI_LOG_ERR, "Security Initialization",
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/ssl.c
```diff
@@ -1251,6 +1251,11 @@ slapd_ssl_init()
         slapd_extract_cert(entry, PR_TRUE);
     }
 
+#ifdef WITH_SYSTEMD
+    slapd_SSL_warn("Sending pin request to SVRCore. You may need to run"
+        " systemd-tty-ask-password-agent to provide the password if pin.txt does not exist.");
+#endif
+
```
mreynolds389, line 1258: Perhaps this should be moved under this condition (before the for loop):

    if ((family_list = getChildren(configDN))) {

Because now it might get logged if SSL is not enabled?

mreynolds389 (approved):
Looks like my concern is invalid, security must be enabled to hit this code.

rohitpc363 (approved):

ldap/servers/slapd/ssl.c
```diff
@@ -1305,11 +1310,7 @@ slapd_ssl_init()
...
             }
-/* authenticate */
-#ifdef WITH_SYSTEMD
-            slapd_SSL_warn("Sending pin request to SVRCore. You may need to run"
-                           " systemd-tty-ask-password-agent to provide the password.");
-#endif
+            /* authenticate */
             if (slapd_pk11_authenticate(slot, PR_TRUE, NULL) != SECSuccess) {
                 errorCode = PR_GetError();
                 slapi_log_err(SLAPI_LOG_ERR, "Security Initialization",
```
rohitpc363, line 1316: Hi

rohitpc363 (approved):

rohitpc363 (approved):

droideck (approved):
Looks good to me too.

P.S. some weird spammer - rohitpc363 - got here... I've removed the message but I am not sure how to remove the 'approved' flood...

Firstyear:
> Looks good to me too.
> 
> P.S. some weird spammer - rohitpc363 - got here... I've removed the message but I am not sure how to remove the 'approved' flood...

Maybe we should check the project settings about "who" can approve commits?

Firstyear:
Updated the patch to remove the warning as we also called this during client-init which mean if you use tls client certs for replication authentication, it would log this message frequently.

mreynolds389 (approved):
LGTM

Issue #4414, opened 2020-10-29
SIGFPE crash in rhds disk monitoring routine


mreynolds389:
Ticket was cloned from Red Hat Bugzilla: [Bug 1890118](https://bugzilla.redhat.com/show_bug.cgi?id=1890118)

_Please note that this Bug is private and may not be accessible as it contains confidential Red Hat customer information._

```
Description of problem:

A customer reports a crash in RHDS that occurs every day at same time

Version-Release number of selected component (if applicable):

RHDS 11

How reproducible:

Not reproduced in-house. Happens regularly on customer site

Actual results:

Crash happening every morning around 5am

Expected results:

RHDS should not crash

Additional info:

This is the stackframe of the crash

Thread 1 (Thread 0x7f8605bfe700 (LWP 1956012)):
#0  0x000055c19b7fd48f in monitor_disk_info (pb=<optimized out>, e=0x7f860407c5c0, entryAfter=<optimized out>, returncode=0x7f8605bf730c, returntext=<optimized out>, arg=<optimized out>) at /usr/include/bits/stdio2.h:67
        i = <optimized out>
        rc = 0
        dirs = 0x7f8538a14680
        buf = "partition=\"/var/log/dirsrv\" size=\"32201768960\" used=\"6259421184\" available=\"25942347776\" use%=\"19\"\000\247\205\177\000\000\364\372\017&\206\177\000\000\240:\024\247\205\177\000\000\003\000\000\000\003", '\000' <repeats 11 times>, "\200:\016+\206\177\000\000\000\b", '\000' <repeats 14 times>, "@4\024\247\205\177\000\000\033\000\000\000\033", '\000' <repeats 20 times>, "\b\000\000\000\000\000\000"...
        val = {bv_len = 98, bv_val = 0x7f8605bf5220 "partition=\"/var/log/dirsrv\" size=\"32201768960\" used=\"6259421184\" available=\"25942347776\" use%=\"19\""}
        vals = {0x7f8605bf5200, 0x0}
        total_space = 0
        avail_space = 0
        used_space = 0
[...]

PR #4519, opened 2021-01-06
Issue 4414 - disk monitoring - prevent division by zero crash

mreynolds389:
Bug Description:  

If a disk mount has zero total space or zero used space then a division by zero can occur and the server will crash.

It has also been observed that sometimes a system can return the wrong disk entirely, and when that happens the incorrect disk also has zero available space which triggers the disk monitoring thread to immediately shut the server down.

Fix Description:  

Check the total and used space for zero and do not divide, just ignore it.  As a pre-emptive measure ignore disks from /dev, /proc, /sys (except /dev/shm). Yes it's a bit hacky, but the true underlying cause is not known yet.  So better to be safe than sorry.

Relates: https://github.com/389ds/389-ds-base/issues/4414

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -222,7 +222,27 @@ disk_mon_get_mount_point(char *dir)
         }
         if (s.st_dev == dev_id) {
             endmntent(fp);
-            return (slapi_ch_strdup(mnt->mnt_dir));
+
+            if ((strncmp(mnt->mnt_dir, "/dev", 4) == 0 && strncmp(mnt->mnt_dir, "/dev/shm", 8) != 0) ||
+                strncmp(mnt->mnt_dir, "/proc", 4) == 0 ||
+                strncmp(mnt->mnt_dir, "/sys", 4) == 0)
+            {
+                /*
+                 * Ignore "mount directories" starting with /dev (except
+                 * /dev/shm), /proc, /sys  For some reason these mounts are
+                 * occasionally/incorrectly returned.  Only seen this at a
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/daemon.c
```diff
@@ -222,7 +222,27 @@ disk_mon_get_mount_point(char *dir)
         }
         if (s.st_dev == dev_id) {
             endmntent(fp);
-            return (slapi_ch_strdup(mnt->mnt_dir));
+            if ((strncmp(mnt->mnt_dir, "/dev", 4) == 0 ||
```
Firstyear: I think the logic here looks wrong? Shouldn't it be:
```
if (
    (strncmp(mnt->mnt_dir, "/dev", 4) == 0 && strncmp(mnt->mnt_dir, "/dev/shm", 8) != 0) ||
    strncmp(mnt->mnt_dir, "/proc", 4) == 0 ||
    strncmp(mnt->mnt_dir, "/sys", 4) == 0)

```

Firstyear (changes requested):
Beside that one logic comment, I think this is good.

ldap/servers/slapd/daemon.c
mreynolds389: You get the same result, but yours is clearer.

Firstyear: Cool, lets go for clarity then, otherwise, ack :)

mreynolds389: Agreed, changes made, and merging...

PR #4520, opened 2021-01-07
Issue 4506 - BUG -

Firstyear:
Bug Description: While investigating 4506 it was noticed that
it was possible to exceed the capacity of the connection table
fd array if you had many listeners and a large number of
connections. The number of connections required and in the
correct state to cause this is in the thousands and would
be infeasible in reality, but it is still worth defending
from this.

Fix Description: Add the correct bound on the while loop
setting up the fd for polling.

relates: https://github.com/389ds/389-ds-base/issues/4506

Review by: ???

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -1334,7 +1334,7 @@ setup_pr_read_pds(Connection_Table *ct, PRFileDesc **n_tcps, PRFileDesc **s_tcps
      * is no longer in use, we should remove it from the linked
      * list. */
     c = connection_table_get_first_active_connection(ct);
-    while (c) {
+    while (c && count < ct->size) {
         next = connection_table_get_next_active_connection(ct, c);
         if (c->c_state == CONN_STATE_FREE) {
             connection_table_move_connection_out_of_active_list(ct, c);


Firstyear:
Note this does NOT fix #4506.

Firstyear:
@mreynolds I think this should be backported all the way back to 1.4.2 if you agree.

progier389 (changes requested):
Something looks wrong here: 
   'count'  should never be greater than the ct->size:
   The while loop fills up the poll descriptor array (whose size is ct->size) 
      so count is the number of file descriptor to poll
   As the connections to poll are already in the open connection array 
      (whose size is also ct->size)
   having c > ct->size should never happens (and count == ct->size means that all open connections are polled ( and in this case c would be NULL since all connections have been looked up) 

so if you really hit this condition in your tests that means that something was already 
 seriously broken  (loop in the active connection list or open connection array overflow)  
 ==> IMHO if you really want to add a test in the while loop 
    then a PR_ASSERT(count < ct->size) is a better idea

Firstyear:
> Something looks wrong here:
> 'count' should never be greater than the ct->size:

The ct->fd array is allocated by ct->size comes from table_size and is allocated as such. But we *also* populate the first few fd's with the listening/accept sockets fds. This means it's possible to have ct->size of fds + number of listeners fds. 

So yes I completely agree it would be so impossibly hard to trigger a scenario where all "thousands" of active conns are ready to poll *and* all the listening fd's are ready as well. But it is *theoretically* possible that it could happen, which is why I added this check.  :)

progier389 (approved):

progier389:
> > Something looks wrong here:
> > 'count' should never be greater than the ct->size:
> 
> The ct->fd array is allocated by ct->size comes from table_size and is allocated as such. But we _also_ populate the first few fd's with the listening/accept sockets fds. This means it's possible to have ct->size of fds + number of listeners fds.
> 
> So yes I completely agree it would be so impossibly hard to trigger a scenario where all "thousands" of active conns are ready to poll _and_ all the listening fd's are ready as well. But it is _theoretically_ possible that it could happen, which is why I added this check. :)

Ok. Thank you for the explanation, I thought that the listening sockets also were in the connection array but I now realized that they are handled differently.  So the change is needed.

Firstyear:
```
   f221b9bf0..1485f60b5  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
   79e924b4f..7c4b4f132  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
   eee0e3308..6e01872d5  389-ds-base-1.4.2 -> 389-ds-base-1.4.2
```

PR #4522, opened 2021-01-08
Issue 4219 - Log internal unindexed searches (notes=A)

vashirov:
Description:
Add a test case.

Relates: https://github.com/389ds/389-ds-base/issues/4219

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/indexes/regression_test.py
+++ b/dirsrvtests/tests/suites/indexes/regression_test.py
@@ -12,13 +12,74 @@ import pytest
 import ldap
 from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
 from lib389.index import Indexes
-from lib389.backend import Backends
+from lib389.backend import Backends, DatabaseConfig
 from lib389.idm.user import UserAccounts
+from lib389.idm.group import Groups, Group
 from lib389.topologies import topology_st as topo
 from lib389.utils import ds_is_older
+from lib389.plugins import MemberOfPlugin
 
 pytestmark = pytest.mark.tier1
 
+
+@pytest.fixture(scope="function")
+def add_a_group_with_users(request, topo):
+    """
+    Add a group and users, which are members of this group.
+    """
+    groups = Groups(topo.standalone, DEFAULT_SUFFIX, rdn=None)
+    group = groups.create(properties={'cn': 'test_group'})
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
LGTM

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -12,14 +12,74 @@
...
+    topo.standalone.restart()
+
+    def fin():
+        """
+        Set nsslapd-idlistscanlimit back to the default value
+        """
+        # If the server crashed, start it again to do the cleanup
+        if not topo.standalone.status():
+            topo.standalone.start()
+        db_cfg.set([('nsslapd-idlistscanlimit', '4000')])
```
droideck: Very minor thing, I think it'll be, probably, better not to hard coded it... 
We can fetch the value at the beginning of the fixture and set it back here at the 'fin()'.

droideck (approved):
Besides that - looks good to me!

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -12,14 +12,74 @@
...
+    request.addfinalizer(fin)
+
+
+@pytest.fixture(scope="function")
+def set_small_idlistscanlimit(request, topo):
+    """
+    Set nsslapd-idlistscanlimit to a smaller value to accelerate the reproducer
+    """
+    db_cfg = DatabaseConfig(topo.standalone)
+    db_cfg.set([('nsslapd-idlistscanlimit', '50')])
```
tmihinto: According to the [documentation](https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/configuration_command_and_file_reference/database_plug_in_attributes#nsslapd_idlistscanlimit), the min value of 'nsslapd-idlistscanlimit' is 100.

dirsrvtests/tests/suites/indexes/regression_test.py
vashirov: Good point, fixed.

dirsrvtests/tests/suites/indexes/regression_test.py
vashirov: Nice catch! Fixed.

PR #4523, opened 2021-01-08
Client keytab fixes

frozencemetery:
The impetus for these changes is an unintended behavior being used in `set_krb5_creds()`.  Basically, the function creates a principal with realm "" (i.e., an empty string), and expects this to function as a kind of wildcard.  However, this was not the intent of krb5, and depending on the DNS canonicalization settings, it may or may not work.

I present two approaches to fixing the problem.  My preferred approach is to eliminate `set_krb5_creds()` entirely in favor of krb5's [client keytab](https://web.mit.edu/kerberos/krb5-latest/doc/basic/keytab_def.html#default-client-keytab) support.  However, if this is undesirable, I've also included a commit that preserves the current behavior while just fixing the bug.

Finally, regarding Heimdal, I want to emphasize that 389ds's Kerberos manipulations haven't been built for against Heimdal in almost two years, so nothing is being lost here; it's just code simplification.  Heimdal lacks `krb5_get_init_creds_opt_set_out_ccache()` which makes the behavior-preserving fix impossible; however, if/when Heimdal has a new feature release, Heimdal will support client keytabs with similar semantics to MIT krb5 (which is part of why I prefer that approach).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -839,8 +839,6 @@ if test "$krb5_vendor" = "MIT"; then
     LIBS="$KERBEROS_LIBS"
     AC_CHECK_FUNCS([krb5_cc_new_unique])
     LIBS="$save_LIBS"
-elif test "$krb5_vendor" = "Heimdal"; then
-    AC_DEFINE(HAVE_HEIMDAL_KERBEROS, 1, [Define if you have Heimdal Kerberos])
 fi
 
 if $PKG_CONFIG --exists pcre; then
--- a/ldap/servers/slapd/ldaputil.c
+++ b/ldap/servers/slapd/ldaputil.c
@@ -63,14 +63,6 @@
 #include "prinrval.h"
 #include "snmp_collator.h"
 
-#ifdef HAVE_HEIMDAL_KERBEROS
-#include <com_err.h>
-#endif
-
-#ifndef MAX_KEYTAB_NAME_LEN
-#define MAX_KEYTAB_NAME_LEN 1100
-#endif
-
 /* need mutex around ldap_initialize - see https://fedorahosted.org/389/ticket/348 */
 static PRCallOnceType ol_init_callOnce = {0, 0, 0};
 static PRLock *ol_init_lock = NULL;
@@ -1308,15 +1300,6 @@ typedef struct
     char *realm;
 } ldapSaslInteractVals;
 
-#ifdef HAVE_KRB5
-static void set_krb5_creds(
-    const char *authid,
-    const char *username,
-    const char *passwd,
-    const char *realm,
-    ldapSaslInteractVals *vals);
-#endif
-
 static void *
 ldap_sasl_set_interact_vals(LDAP *ld, const char *mech, const char *authid, const char *username, const char *passwd, const char *realm)
 {
@@ -1373,12 +1356,6 @@ ldap_sasl_set_interact_vals(LDAP *ld, const char *mech, const char *authid, cons
             vals->realm = slapi_ch_strdup("");
         }
     }
-
-#ifdef HAVE_KRB5
-    if (mech && !strcmp(mech, "GSSAPI")) {
-        set_krb5_creds(authid, username, passwd, realm, vals);
-    }
-#endif /* HAVE_KRB5 */
 
     return vals;
 }
@@ -1548,11 +1525,7 @@ show_one_credential(int authtracelevel,
     char *logname = "show_one_credential";
     krb5_error_code rc;
     char *name = NULL, *sname = NULL;
-#ifdef HAVE_HEIMDAL_KERBEROS
-    krb5_timestamp startts, endts, renewts;
-#else
     char startts[BUFSIZ], endts[BUFSIZ], renewts[BUFSIZ];
-#endif
 
     if ((rc = krb5_unparse_name(ctx, cred->client, &name))) {
         slapi_log_err(SLAPI_LOG_ERR, logname,
@@ -1569,14 +1542,6 @@ show_one_credential(int authtracelevel,
     if (!cred->times.starttime) {
         cred->times.starttime = cred->times.authtime;
     }
-#ifdef HAVE_HEIMDAL_KERBEROS
-    slapi_log_error(authtracelevel, logname,
-                    "\tKerberos credential: client [%s] server [%s] "
-                    "start time [%s] end time [%s] renew time [%s] "
-                    "flags [0x%x]\n",
-                    name, sname, ctime(&startts), ctime(&endts),
-                    ctime(&renewts), (uint32_t)cred->flags.i);
-#else
     krb5_timestamp_to_sfstring((krb5_timestamp)cred->times.starttime,
                                startts, sizeof(startts), NULL);
     krb5_timestamp_to_sfstring((krb5_timestamp)cred->times.endtime,
@@ -1590,7 +1555,6 @@ show_one_credential(int authtracelevel,
                   "flags [0x%x]\n",
                   name, sname, startts, endts,
                   renewts, (uint32_t)cred->ticket_flags);
-#endif
 
 cleanup:
     krb5_free_unparsed_name(ctx, name);
@@ -1693,13 +1657,8 @@ credentials_are_valid(
        order to set mcreds.server required in order
        to use krb5_cc_retrieve_creds() */
 /* get default realm first */
-#ifdef HAVE_HEIMDAL_KERBEROS
-    realm_str = krb5_principal_get_realm(ctx, princ);
-    realm_len = krb5_realm_length(realm_str);
-#else
     realm_len = krb5_princ_realm(ctx, princ)->length;
     realm_str = krb5_princ_realm(ctx, princ)->data;
-#endif
     tgs_princ_name = slapi_ch_smprintf("%s/%*s@%*s", KRB5_TGS_NAME,
                                        realm_len, realm_str,
                                        realm_len, realm_str);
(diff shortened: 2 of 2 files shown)


frozencemetery:
(I have tested this with freeipa; it needs the change in the linked PR.)

frozencemetery:
Downstream, this is related to [Fedora bug 1868482](https://bugzilla.redhat.com/show_bug.cgi?id=1868482)

Firstyear:
I don't see anything wrong with this change. When you say you tested with freeipa, did you also test replication worked?

frozencemetery:
Yes.  My specific test was Fedora 32 (with these patches applied to 389ds).  I set up a freeipa instance and a replica (ipa-replica-install), similar to what openqa does.

mreynolds389:
I'm okay with this as well, but I need to ask you to reformat the commit message to adhere to our policy:

Thanks Robbie!

droideck (approved):
Looks good to me too!
Seems like the long debugging issue is finally coming to an end! Congrats!

frozencemetery:
Apologies, I was unaware of that policy.  I've squashed and opened https://github.com/389ds/389-ds-base/issues/4537

mreynolds389:
> Apologies, I was unaware of that policy. I've squashed and opened #4537

No worries, every team does it a little different.  We use our commit messages (mainly the first line) to structure our release notes and specfile changelogs.  Thanks again!

Firstyear:
Merging as there are now three acks and the commit was cleaned up :)

Firstyear:
Thanks @frozencemetery for this set of fixes.

tbordaz:
Hi @frozencemetery, I am very novice with krb5 so sorry in advance for my dummy question. My understanding is that set_krb5_creds was somehow overwriting the krb5 client keytab. I understand the benefit for ipa that defines a client keytab to remove that call. 
For others deployment where client keytab are not defined, is there a risk that without the call to set_krb5_creds the authentication will fail ?

mreynolds389:
Also this introduced comiler wanrings:

```
../389-ds-base/ldap/servers/slapd/ldaputil.c:1631:1: warning: ‘credentials_are_valid’ defined but not used [-Wunused-function]
 1631 | credentials_are_valid(
      | ^~~~~~~~~~~~~~~~~~~~~
../389-ds-base/ldap/servers/slapd/ldaputil.c:1622:1: warning: ‘looks_like_a_princ_name’ defined but not used [-Wunused-function]
 1622 | looks_like_a_princ_name(const char *name)
      | ^~~~~~~~~~~~~~~~~~~~~~~
../389-ds-base/ldap/servers/slapd/ldaputil.c:1570:1: warning: ‘show_cached_credentials’ defined but not used [-Wunused-function]
 1570 | show_cached_credentials(int authtracelevel,
      | ^~~~~~~~~~~~~~~~~~~~~~~
```

Looks like these functions are no longer needed.  If so, they should be removed...

frozencemetery:
> Hi @frozencemetery, I am very novice with krb5 so sorry in advance for my dummy question.

Hi @tbordaz, not a problem - it's my job to explain :)

> My understanding is that set_krb5_creds was somehow overwriting the krb5 client keytab.

Kerberos/GSSAPI have the concept of an initiator and an acceptor (which we will sometimes call client and server, though this isn't strictly true).  The credentials acquired will differ between the two - that is, there are both initiator credentials and acceptor credentials, and they're not the same.

A keytab is a blob (usually in a file somewhere) that can be used to acquire credentials.  So krb5 has two environment variables - $KRB5_KTNAME and $KRB5_CLIENT_KTNAME - for the different kinds of credentials.  For historical reasons, $KRB5_CLIENT_KTNAME is much newer: before krb5-1.11 it wasn't possible to acquire initiator credentials using a keytab via environment variable.  This is probably the main reason `set_krb5_creds()` was created, though I don't actually know.

So: `set_krb5_creds()` attempted to populate a ccache with both kinds of credentials using the $KRB5_KTNAME environment variable.  Other than the bug mentioned with empty realm handling, this is actually a weird thing to do: in particular, it means that a separate client keytab can't be specified because it would be overridden.  (krb5 will of course prefer the contents of the credential cache.)

> I understand the benefit for ipa that defines a client keytab to remove that call.

There's actually no benefit to IPA for configuring them separately.  The benefit comes from fixing the empty realm issue.

> For others deployment where client keytab are not defined, is there a risk that without the call to set_krb5_creds the authentication will fail ?

A deployment which needed initiator creds (e.g., for replication) and depended on the behavior of `set_krb5_creds()` will need to add another line of configuration for $KRB5_CLIENT_KEYTAB.  The IPA PR is an example of what this looks like.

In case this was not something 389ds was okay with, I did submit another version - prior to squashing, that was https://github.com/frozencemetery/389-ds-base/commit/f092f9cb687ff13be29e5ce58a7af69ae87a6a81 - which would have preserved the existing behavior.

frozencemetery:
> Also this introduced comiler wanrings:

Should be fixed by https://github.com/389ds/389-ds-base/pull/4545

tbordaz:
@frozencemetery thank you sooo much for your precise explanations and patience !
What are the impact of the fix on a deployment that does not set KRB5_CLIENT_KEYTAB ?

So a former 389-ds deployment did not need to define KRB5_CLIENT_KEYTAB and was using a "weird" empty realm to get the initiator creds. Correct ?
After upgrading to that fix, the former deployment will need to set KRB5_CLIENT_KEYTAB. Else it will have no initiator creds and auth will fail. Correct ?

frozencemetery:
> What are the impact of the fix on a deployment that does not set KRB5_CLIENT_KEYTAB ?

I believe this will cause their replication to fail during authentication.  I'm not a 389ds expert, but it would affect any scenario in which 389ds is the client/initiator.

> So a former 389-ds deployment did not need to define KRB5_CLIENT_KEYTAB and was using a "weird" empty realm to get the initiator creds. Correct ?

Right.  With the old code, most of the time this would work, but sometimes it would not.

>  After upgrading to that fix, the former deployment will need to set KRB5_CLIENT_KEYTAB. Else it will have no initiator creds and auth will fail. Correct ?

Auth will fail where 389ds is the client/initiator, yeah.

mreynolds389:
Sounds like we need to make this configurable so we don't break anybody or mixed deployments.

Firstyear:
> Sounds like we need to make this configurable so we don't break anybody or mixed deployments.

Alternately, if we detect KRB5_KEYTAB but *not* KRB5_CLIENT_KEYTAB could we log errors about the fact KRB will fail and resolution steps? I think the krb5 cleanup is valid, and we should direct people into how to resolve the setup.

Another option is a healthcheck command or similar?

tbordaz:
> > Sounds like we need to make this configurable so we don't break anybody or mixed deployments.
> 
> Alternately, if we detect KRB5_KEYTAB but _not_ KRB5_CLIENT_KEYTAB could we log errors about the fact KRB will fail and resolution steps? I think the krb5 cleanup is valid, and we should direct people into how to resolve the setup.
> 
> Another option is a healthcheck command or similar?

Would it be possible to detect missing KRB5_CLIENT_KEYTAB and in that case fall-back to the "weird" empty realm ? Would it work ? @frozencemetery what do you think ?

frozencemetery:
My feeling is that it's going to end up being weird somehow - it's just a question of how you want it to be weird :)

- If we do nothing more, 389ds's Kerberos behaves like that of other applications, but this is different than it was before, requiring configuration change.
- If we add a behavior configuration handle, we basically keep the old behavior, which makes 389ds different than ever other Kerberos application.
- If we conditionally set $KRB5_CLIENT_KEYTAB when it's not set but $KRB5_KEYTAB is, we prevent admins from using a client keytab at the [default location](https://web.mit.edu/kerberos/krb5-latest/doc/basic/keytab_def.html#default-client-keytab) when configuring $KRB5_KEYTAB.

Up to you, of course.  None of those three are clearly wrong to my mind.

Firstyear:
> My feeling is that it's going to end up being weird somehow - it's just a question of how you want it to be weird :)
> 
> * If we do nothing more, 389ds's Kerberos behaves like that of other applications, but this is different than it was before, requiring configuration change.

IMO I'd rather the configuration change to make 389ds in line with other applications. That's fine, but the issue is we tend to have strict requirements about how much churn we can have within a release, and especially with potentially confusing and difficult topics like KRB.

So I think the question becomes for @mreynolds389 and others about when and how we want to do this.

We may need to do something like adding the configuration handle for the old behaviour as a shim until the end of a release cycle, or we need to ensure that we only do this in 2.0.0 and don't backport to 1.4.x which is already released.

tbordaz:
> My feeling is that it's going to end up being weird somehow - it's just a question of how you want it to be weird :)
> 
>     * If we do nothing more, 389ds's Kerberos behaves like that of other applications, but this is different than it was before, requiring configuration change.
> 
>     * If we add a behavior configuration handle, we basically keep the old behavior, which makes 389ds different than ever other Kerberos application.
> 
>     * If we conditionally set $KRB5_CLIENT_KEYTAB when it's not set but $KRB5_KEYTAB is, we prevent admins from using a client keytab at the [default location](https://web.mit.edu/kerberos/krb5-latest/doc/basic/keytab_def.html#default-client-keytab) when configuring $KRB5_KEYTAB.
> 
> 
Thanks @frozencemetery for making those 3 options clear.
I would like to understand if 389-ds is able to use KRB5_CLIENT_KEYTAB (your new code) when it is available and fallback to the previous behaviour (bad I agree) if KRB5_CLIENT_KEYTAB is not available.
My understanding is that KRB5_CLIENT_KEYTAB can be defined as environment variable *or* at a default location (use of KRB5_KEYTAB). 389-ds can detect the environment variable. Is there a programmatic interface that 389-ds could use to detect if the client keytab will be available from the [default location](https://web.mit.edu/kerberos/krb5-latest/doc/basic/keytab_def.html#default-client-keytab) ?

frozencemetery:
>  My understanding is that KRB5_CLIENT_KEYTAB can be defined as environment variable or at a default location (use of KRB5_KEYTAB). 389-ds can detect the environment variable. Is there a programmatic interface that 389-ds could use to detect if the client keytab will be available from the default location ?

It's possible to do - one could use the profile interface (profile.h) to figure out what it's configured to be, and then try to acquire credentials from that location (optionally checking whether it's a file/exists first).  This risks a false positive for anyone expecting the old behavior, but since dirsrv runs as its own user (I think?) the risk there should be small.  Is it worth the complexity?  I do not know.

mreynolds389:
Turns out this fix is breaking IPA replication.  There are tons of GSSAPI failures in replication tests.  See https://bugzilla.redhat.com/show_bug.cgi?id=1918075

We need to revert this until it is properly tested and is backwards compatible.

mreynolds389:
Ok, this should be fixed in freeipa via:   https://github.com/freeipa/freeipa/pull/5409

So no need to revert :-D

We still need to make all of this backwards compatible, and it should be done in our next sprint.  I'll open a JIRA ticket for that

PR #4524, opened 2021-01-11
Issue 4315 - performance search rate: nagle triggers high rate of set…

bsimonova:
…socketopt

Description:
The config value of nsslapd-nagle is now set to 'off' by default.
Added a test case, that checks the value.

Relates: https://github.com/389ds/389-ds-base/issues/4315

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/config/config_test.py
+++ b/dirsrvtests/tests/suites/config/config_test.py
@@ -41,6 +41,26 @@ def big_file():
     return TEMP_BIG_FILE
 
 
+@pytest.mark.bz1897248
+@pytest.mark.ds4315
+@pytest.mark.skipif(ds_is_older('1.4.3.16'), reason="This config setting exists in 1.4.3.16 and higher")
+def test_nagle_default_value(topo):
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -551,6 +551,26 @@ def test_require_internal_index(topo):
         user.delete()
 
 
+@pytest.mark.bz1897248
+@pytest.mark.ds4315
+@pytest.mark.skipif(ds_is_older('1.4.3.16'), reason="This config setting exists in 1.4.3.16 and higher")
+def test_nagle_default_value(topo):
+    """Test that nsslapd-nagle attribute is off by default
```
droideck: Probably, it makes more sense to put it at the beginning of the test suite as it checks for a default value.

droideck (approved):
Besides the small comment about where to put the test case - you have my ack!

bsimonova:
Good point, thanks!
Moved the test to the beginning and will merge.

Issue #4521, opened 2021-01-08
DS crash in deref plugin while dereferencing an entry that exists but that is not returned by internal search


tbordaz:
**Issue Description**
Deref control dereference a set of attributes that are present in the returned set of entries. Each attribute value is assume to be a DN of an entry. if for some reason the deref plugin can not retrieve (internal search) the entry, the server crash.

**Package Version and Platform:**
all versions

**Steps to Reproduce**
Testcase to be provided

**Expected results**
The deref control should be handled without a crash even if the returned entries a corrupted

tbordaz:
89367c674..acaf2235c  master
62ba2a273..f221b9bf0  389-ds-base-1.4.4
0acda489f..79e924b4f  389-ds-base-1.4.3
f917145e1..d8963a0af  389-ds-base-1.3.10

PR #4525, opened 2021-01-11
Issue 4521 - DS crash in deref plugin if dereferenced entry exists bu…

tbordaz:
…t is not returned by internal search

Bug description:
	For each returned entry, deref plugin dereferences some attribute values that refer to entries.
	To do this it does an internal search (scope base) with each attribute values.
	Deref plugin assumes that if internal search succeeds, a single entry is returned.
	It exists cases (not identified) where internal search succeeds but returns no entry.
	In such case (search succeeds but no entry returned) the server crash.
	Note: wonder if DB deadlock could lead to such situation.

Fix description:
	Make a hardening fix that logs warning in such case

relates: https://github.com/389ds/389-ds-base/issues/4521

Reviewed by: ?

Platforms tested: F31


mreynolds389 (approved):
LGTM

tbordaz:
89367c674..acaf2235c  master
62ba2a273..f221b9bf0  389-ds-base-1.4.4
0acda489f..79e924b4f  389-ds-base-1.4.3
f917145e1..d8963a0af  389-ds-base-1.3.10

PR #4538, opened 2021-01-12
Issue 4535 - lib389 - healthcheck throws exception if backend is not …

### LINKED ISSUE #4535: lib389 - clients of lib389 healthcheck raise exceptions when a backend is not replicated
**Issue Description**
Freeipa's healthcheck tool throws errors when processing non-replicated backends.  In dsctl these are caught as part of the CLI and are handled correctly, but not when a client using our API calls these functions directly.

See:  https://github.com/freeipa/freeipa-healthcheck/issues/163

We need to catch an exception when attempting to get the replica:

```
def _lint_cl_trimming(self):
        """Check that cl trimming is at least defined to prevent unbounded growth"""
        suffix = self.get_attr_val_utf8('nsslapd-suffix')
        replicas = Replicas(self._instance)
        replica = replicas.get(suffix)   ---->  Need to catch exception here
        bename = self.lint_uid()
        if replica is not None:
```

mreynolds389:
…replicated

Bug Description:

If a backend is not replicated then healthcheck backend cl_trimming check will
throw an exception.  Now dsctl catches this error and moves on, but ipa healthcheck
complains because it is directly using the API.

Fix Description:

Catch the exception is perpetration is not enabled, and just move to the next check.

Fixes: https://github.com/389ds/389-ds-base/issues/4535

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -511,22 +511,24 @@ class Backend(DSLdapObject):
 
     def _lint_cl_trimming(self):
         """Check that cl trimming is at least defined to prevent unbounded growth"""
+        bename = self.lint_uid()
         suffix = self.get_attr_val_utf8('nsslapd-suffix')
         replicas = Replicas(self._instance)
-        replica = replicas.get(suffix)
-        bename = self.lint_uid()
-        if replica is not None:
+        try:
+            # Check if replication is enabled
+            replicas.get(suffix)
+            # Check the changelog
             cl = Changelog(self._instance, suffix=suffix)
-            try:
-                if cl.get_attr_val_utf8('nsslapd-changelogmaxentries') is None and \
-                   cl.get_attr_val_utf8('nsslapd-changelogmaxage') is None:
-                    report = copy.deepcopy(DSCLLE0001)
-                    report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
-                    report['check'] = f'backends:{bename}::cl_trimming'
-                    yield report
-            except:
-                # No changelog
-                pass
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/backend.py
```diff
@@ -511,22 +511,23 @@ def _lint_mappingtree(self):
...
-                # No changelog
-                pass
+            if cl.get_attr_val_utf8('nsslapd-changelogmaxentries') is None and \
+               cl.get_attr_val_utf8('nsslapd-changelogmaxage') is None:
+                report = copy.deepcopy(DSCLLE0001)
+                report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
+                report['check'] = f'backends:{bename}::cl_trimming'
+                yield report
+        except:
+            # Suffix is not replicated
```
Firstyear, line 529: Might just be worth a log.debug to say "skipping bename as no replication enabled"  so that if we have a CU issue and we get debug logs we can see there may be a replication agreement inbalance?

Firstyear (approved):
Besides one possible improvement, I think this looks good to me. Ack!

droideck:
The test suite fails for me if it's run as a whole.
Probably, `test_healthcheck_replication_out_of_sync_broken` is to blame as it redefines `topology_m2` with `topology_m3` rewriting the whole thing.

droideck (changes requested):
The test suite fails for me if it's run as a whole.
Probably, test_healthcheck_replication_out_of_sync_broken is to blame as it redefines topology_m2 with topology_m3 rewriting the whole thing.

src/lib389/lib389/backend.py
mreynolds389: Done

PR #4541, opened 2021-01-13
Issue #4534 - libasan read buffer overflow in filtercmp

progier389:
Bug Description:
    slapi_filter_compare function calls memcmp using first key length which may be longer than second key length.
    
 Bug Fix:
     Use slapi_berval_cmp that check both value sizes instead of memcmp

Issue: [#4534](https://github.com/389ds/389-ds-base/issues/4534) 
     
Reviewed by: [droideck](https://github.com/droideck)


droideck (approved):
Looks good and clean to me!
And tests pass. Ack

Issue #4542, opened 2021-01-13
Search results are different between RHDS10 and RHDS11


mreynolds389:
**Issue Description**

The search results are different between older version of RHDS10 and RHDS11 when non-existent entry is specified as BaseDN like

  (RHDS10)
  $ ldapsearch -LLL-x -D "uid=testuser1,ou=People,dc=ad,dc=example,dc=com" -W -b "ou=dummy,dc=ad,dc=example,dc=com"
  $ echo $?
  32

  (RHDS11)
  $ ldapsearch -LLL-x -D "uid=testuser1,ou=People,dc=ad,dc=example,dc=com" -W -b "ou=dummy,dc=ad,dc=example,dc=com"
  $ echo $?
  0

We should be returning error 32

Cloned from:  https://bugzilla.redhat.com/show_bug.cgi?id=1909675

mreynolds389:
16e400869..d641e8528  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
1f859aa52..6c8755687  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4543, opened 2021-01-13
Issue 5442 - Search results are different between RHDS10 and RHDS11

mreynolds389:
Bug Description:  

In 1.4.x we introduced a change that was overly strict about how a search on a non-existent subtree returned its error code.  It was changed from returning an error 32 to an error 0 with  zero entries returned.

Fix Description:  

Change the server behavior back to returning error 32 in this case.

relates: https://github.com/389ds/389-ds-base/issues/4542

### DIFF
--- a/dirsrvtests/tests/suites/acl/misc_test.py
+++ b/dirsrvtests/tests/suites/acl/misc_test.py
@@ -12,7 +12,7 @@ import ldap
 import os
 import pytest
 
-from lib389._constants import DEFAULT_SUFFIX, PW_DM
+from lib389._constants import DEFAULT_SUFFIX, PW_DM, DN_DM
 from lib389.idm.user import UserAccount, UserAccounts
 from lib389._mapped_object import DSLdapObject
 from lib389.idm.account import Accounts, Anonymous
@@ -408,14 +408,112 @@ def test_do_bind_as_201_distinct_users(topo, clean, aci_of_user):
         user = uas.create_test_user(uid=i, gid=i)
         user.set('userPassword', PW_DM)
 
-    for i in range(len(uas.list())):
-        uas.list()[i].bind(PW_DM)
+    users = uas.list()
+    for user in users:
+        user.bind(PW_DM)
 
     ACLPlugin(topo.standalone).replace("nsslapd-aclpb-max-selected-acls", '220')
     topo.standalone.restart()
 
-    for i in range(len(uas.list())):
-        uas.list()[i].bind(PW_DM)
+    users = uas.list()
+    for user in users:
+        user.bind(PW_DM)
+
+
+def test_info_disclosure(request, topo):
+    """Test that a search returns 32 when base entry does not exist
+
+    :id: f6dec4c2-65a3-41e4-a4c0-146196863333
+    :setup: Standalone Instance
+    :steps:
+        1. Add aci
+        2. Add test user
+        3. Bind as user and search for non-existent entry
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Error 32 is returned
+    """
+
+    ACI_TARGET = "(targetattr = \"*\")(target = \"ldap:///%s\")" % (DEFAULT_SUFFIX)
+    ACI_ALLOW = "(version 3.0; acl \"Read/Search permission for all users\"; allow (read,search)"
+    ACI_SUBJECT = "(userdn=\"ldap:///all\");)"
+    ACI = ACI_TARGET + ACI_ALLOW + ACI_SUBJECT
+
+    # Get current ACi's so we can restore them when we are done
+    suffix = Domain(topo.standalone, DEFAULT_SUFFIX)
+    preserved_acis = suffix.get_attr_vals_utf8('aci')
+
+    def finofaci():
+        domain = Domain(topo.standalone, DEFAULT_SUFFIX)
+        try:
+            domain.remove_all('aci')
+            domain.replace_values('aci', preserved_acis)
+        except:
+            pass
+    request.addfinalizer(finofaci)
+
+    # Remove aci's
+    suffix.remove_all('aci')
+
+    # Add test user
+    USER_DN = "uid=test,ou=people," + DEFAULT_SUFFIX
+    users = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    users.create(properties={
+        'uid': 'test',
+        'cn': 'test',
+        'sn': 'test',
+        'uidNumber': '1000',
+        'gidNumber': '2000',
+        'homeDirectory': '/home/test',
+        'userPassword': PW_DM
+    })
+
+    # bind as user
+    conn = UserAccount(topo.standalone, USER_DN).bind(PW_DM)
+
+    # Search fo existing base DN
+    test = Domain(conn, DEFAULT_SUFFIX)
+    try:
+        test.get_attr_vals_utf8_l('dc')
+        assert False
+    except IndexError:
+        pass
+
+    # Search for a non existent bases
+    subtree = Domain(conn, "ou=does_not_exist," + DEFAULT_SUFFIX)
+    try:
+        subtree.get_attr_vals_utf8_l('objectclass')
+    except IndexError:
+        pass
+    subtree = Domain(conn, "ou=also does not exist,ou=does_not_exist," + DEFAULT_SUFFIX)
+    try:
+        subtree.get_attr_vals_utf8_l('objectclass')
+    except IndexError:
+        pass
+    # Try ONE level search instead of BASE
+    try:
+        Accounts(conn, "ou=does_not_exist," + DEFAULT_SUFFIX).filter("(objectclass=top)", ldap.SCOPE_ONELEVEL)
+    except IndexError:
+        pass
+
+    # add aci
+    suffix.add('aci', ACI)
+
+    # Search for a non existent entry which should raise an exception
+    with pytest.raises(ldap.NO_SUCH_OBJECT):
+        conn = UserAccount(topo.standalone, USER_DN).bind(PW_DM)
+        subtree = Domain(conn, "ou=does_not_exist," + DEFAULT_SUFFIX)
+        subtree.get_attr_vals_utf8_l('objectclass')
+    with pytest.raises(ldap.NO_SUCH_OBJECT):
+        conn = UserAccount(topo.standalone, USER_DN).bind(PW_DM)
+        subtree = Domain(conn, "ou=also does not exist,ou=does_not_exist," + DEFAULT_SUFFIX)
+        subtree.get_attr_vals_utf8_l('objectclass')
+    with pytest.raises(ldap.NO_SUCH_OBJECT):
+        conn = UserAccount(topo.standalone, USER_DN).bind(PW_DM)
+        DN = "ou=also does not exist,ou=does_not_exist," + DEFAULT_SUFFIX
+        Accounts(conn, DN).filter("(objectclass=top)", ldap.SCOPE_ONELEVEL, strict=True)
+
 
 if __name__ == "__main__":
     CURRENT_FILE = os.path.realpath(__file__)
--- a/ldap/servers/plugins/acl/acl.c
+++ b/ldap/servers/plugins/acl/acl.c
@@ -2111,10 +2111,11 @@ acl__resource_match_aci(Acl_PBlock *aclpb, aci_t *aci, int skip_attrEval, int *a
     aci_right = aci->aci_access;
     res_right = aclpb->aclpb_access;
     if (!(aci_right & res_right)) {
-        /* If we are looking for read/search and the acl has read/search
-        ** then go further because if targets match we may keep that
-        ** acl in  the entry cache list.
-        */
+        /*
+         * If we are looking for read/search and the acl has read/search
+         * then go further because if targets match we may keep that
+         * acl in the entry cache list.
+         */
         if (!((res_right & (SLAPI_ACL_SEARCH | SLAPI_ACL_READ)) &&
               (aci_right & (SLAPI_ACL_SEARCH | SLAPI_ACL_READ)))) {
             matches = ACL_FALSE;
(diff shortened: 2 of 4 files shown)


Firstyear:
Interesting find! Is this related to the behaviour thierry saw in deref recently? Ack from me at least, but I think it would be good also for @tbordaz to look too :)

elkris:
I don't know about tis specific scenario, but we had a few security issues regarding return codes: the client should not be able to gain inormation by the return code it otherwise would have no access. 
In this case the rc=32 vs rc=0 could be used to determine if an entry is there, just use it as search base and check rc. 
I think the correct behaviour would be to return rc=32 if the client would have access rights to search/read the entry and rc=0 otherwise.

tbordaz:
I agree with @elkris comment, if the bound entry has rights to read the entry we should return 32 else 0.
@mreynolds389 the fix looks good but I have a doubt. Did you try the same search with a bound user *not* having read rights ? it should still return LDAP_SUCCESS.

mreynolds389:
> I agree with @elkris comment, if the bound entry has rights to read the entry we should return 32 else 0.
> @mreynolds389 the fix looks good but I have a doubt. Did you try the same search with a bound user _not_ having read rights ? it should still return LDAP_SUCCESS.

For a non-existent DN the ACL plugin is returning error 50 regardless if there is an aci allowing read access to the suffix.  Need to dig deeper, but this might require additional changes to the ACL code...

ldap/servers/plugins/acl/acl.c
```diff
@@ -2773,64 +2743,35 @@ acl__resource_match_aci(Acl_PBlock *aclpb, aci_t *aci, int skip_attrEval, int *a
...
-            ** consider this acl.
-            ** the opposite is true if it is a deny rule, only a deny without
-            ** any targetattr should deny modrdn
-            ** default: matches = ACL_TRUE;
-            */
-                ;
-            } else {
-                matches = ACL_FALSE;
-            }
         }
```
mreynolds389: This code removal above is the "fix"...

mreynolds389: The removal of this code does not cause any regressions in the ACI test suite, but I am not sure it is 100% valid either because the purpose of this code block is not entirely clear.   Please review...

Firstyear: Doing some git investigation (and having to checkout from before the great code format clean up of 2017), I can see the following in git blame:

```
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2790)    } else if ( (c_attrEval) ||
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2791)                (aci->aci_type & ACI_TARGET_ATTR)) {
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2792)            if ((aci_right & ACL_RIGHTS_TARGETATTR_NOT_NEEDED) &&
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2793)                (aclpb->aclpb_access & ACL_RIGHTS_TARGETATTR_NOT_NEEDED)) {
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2794)                    /*
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2795)                    ** Targetattr rule doesn't  make any sense
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2796)                    ** in this case. So select this rule
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2797)                    ** default: matches = ACL_TRUE;
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2798)                    */
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2799)                    ;
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2800)            } else if (aci_right & SLAPI_ACL_WRITE &&
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2801)                      (aci->aci_type & ACI_TARGET_ATTR) &&
fe0491c49f (Ludwig Krispenz          2013-07-09 16:11:12 +0200 2802)                      !(c_attrEval) &&
fe0491c49f (Ludwig Krispenz          2013-07-09 16:11:12 +0200 2803)                      (aci->aci_type & ACI_HAS_ALLOW_RULE)) {
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2804)                    /* We need to handle modrdn operation.  Modrdn doesn't
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2805)                    ** change any attrs but changes the RDN and so (attr=NULL).
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2806)                    ** Here we found an acl which has a targetattr but
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2807)                    ** the resource doesn't need one. In that case, we should
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2808)                    ** consider this acl.
fe0491c49f (Ludwig Krispenz          2013-07-09 16:11:12 +0200 2809)                    ** the opposite is true if it is a deny rule, only a deny without
fe0491c49f (Ludwig Krispenz          2013-07-09 16:11:12 +0200 2810)                    ** any targetattr should deny modrdn
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2811)                    ** default: matches = ACL_TRUE;
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2812)                    */
^b2093e301 (cvsadm                   2005-01-21 00:44:34 +0000 2813)                    ;

```

It looks like the majority of this code is "quite old" to say the least, but what is interesting is the condition change from ludwig in 2013. According to git its:

```
commit fe0491c49fdb180f1bd291ba78148da7c58b5e90
Author: Ludwig Krispenz <lkrispen@redhat.com>
Date:   Tue Jul 9 16:11:12 2013 +0200

    Ticket 47399 - RHDS denies MODRDN access if ACI list contains any DENY rule

    Bug Description:  if there is a deny rule targeting a specific attribute it also
                      denies the modrdn operation

    Fix Description:   only apply deny rules to modrdn if no target attr exists

    https://fedorahosted.org/389/ticket/47339

    Reviewed by: Rich, thanks
```

So I'm not sure the delete of this block is the right change here, especialy given how ... fun the aci code is at the best of times. I think we need some more research?

mreynolds389:
Okay, I "better" fix.  Please review...

 The problem that I have is that I don't know why we have that code block in the first place.  I don't understand why we want to reject access if the calling function did not provide an attr(c_attrEval) and there is a targetattr set in the aci.   The code talks about only MODRDN operatrions set c_attrEval to NULL.  But it also happens when the original search base does not exist (maybe other scenarios?)

The new fix is say the ACI matches if it's a search/read ACI, there is a "SEARCH" access request, and the c_attrEval is NULL.

Firstyear:
Yep, this is a much better approach IMO. Ack from me.

tbordaz:
This fix looks good but I have some difficulty to anticipate its side effect.

I think the change of behavior is due to https://github.com/389ds/389-ds-base/commit/0b932d4b926d46ac5060f02617330dc444e06da1#diff-202441b945c6ac784560cba25039e1c0fa3f9b9b775491cbeffe5026a3ecd9a6R214
If the base search does not exist it returns a "neutral" success to not discloase any information. The exception is if if the requestor has rights (search/read) on the parent entry, then it returns NO_SUCH_ENTRY. I think the problem come from the way it calls plugin_call_acl_plugin. It is missing an attribute list so it returns invalid access to the parent entry. I stole some code from vlv handling (ldbm_back_search) and it looks it is working.
The following change is another option:
<pre>
diff --git a/ldap/servers/slapd/back-ldbm/findentry.c b/ldap/servers/slapd/back-ldbm/findentry.c
index 6e53a0aea..59122fbb3 100644
--- a/ldap/servers/slapd/back-ldbm/findentry.c
+++ b/ldap/servers/slapd/back-ldbm/findentry.c
@@ -207,7 +207,11 @@ find_entry_internal_dn(
                     break;
                 }
                 if (acl_type > 0) {
-                    err = plugin_call_acl_plugin(pb, me->ep_entry, NULL, NULL, acl_type,
+                    char *dummyAttr = "dummy#attr";
+                    char *dummyAttrs[2] = {NULL, NULL};
+
+                    dummyAttrs[0] = dummyAttr;
+                    err = plugin_call_acl_plugin(pb, me->ep_entry, dummyAttr, NULL, acl_type,
                                                  ACLPLUGIN_ACCESS_DEFAULT, &errbuf);
                 }
                 if (((acl_type > 0) && err) || (op_type == SLAPI_OPERATION_BIND)) {

</pre>

mreynolds389:
> This fix looks good but I have some difficulty to anticipate its side effect.

I know, it worries me as well...

> 
> I think the change of behavior is due to [0b932d4#diff-202441b945c6ac784560cba25039e1c0fa3f9b9b775491cbeffe5026a3ecd9a6R214](https://github.com/389ds/389-ds-base/commit/0b932d4b926d46ac5060f02617330dc444e06da1#diff-202441b945c6ac784560cba25039e1c0fa3f9b9b775491cbeffe5026a3ecd9a6R214)
> If the base search does not exist it returns a "neutral" success to not discloase any information. The exception is if if the requestor has rights (search/read) on the parent entry, then it returns NO_SUCH_ENTRY. I think the problem come from the way it calls plugin_call_acl_plugin. It is missing an attribute list so it returns invalid access to the parent entry. I stole some code from vlv handling (ldbm_back_search) and it looks it is working.
> The following change is another option:
> 
> diff --git a/ldap/servers/slapd/back-ldbm/findentry.c b/ldap/servers/slapd/back-ldbm/findentry.c
> index 6e53a0aea..59122fbb3 100644
> --- a/ldap/servers/slapd/back-ldbm/findentry.c
> +++ b/ldap/servers/slapd/back-ldbm/findentry.c
> @@ -207,7 +207,11 @@ find_entry_internal_dn(
>                      break;
>                  }
>                  if (acl_type > 0) {
> -                    err = plugin_call_acl_plugin(pb, me->ep_entry, NULL, NULL, acl_type,
> +                    char *dummyAttr = "dummy#attr";
> +                    char *dummyAttrs[2] = {NULL, NULL};
> +
> +                    dummyAttrs[0] = dummyAttr;

Doesn't look like you are using dummyAttrs at all, just dummyAttr.  Is that correct?

> +                    err = plugin_call_acl_plugin(pb, me->ep_entry, dummyAttr, NULL, acl_type,
>                                                   ACLPLUGIN_ACCESS_DEFAULT, &errbuf);
>                  }
>                  if (((acl_type > 0) && err) || (op_type == SLAPI_OPERATION_BIND)) {

Just from looking at this it definitely makes sense, like I had said before, since we were not passing in an attribute it was going down a weird code path.  I didn't think you could just create a random dummy attribute without other weird side effects, but if VLV does it then it makes sense and it passes your test, so that works for me  :-)

tbordaz:
@mreynolds389 you are right I cut/past a broken patch. plugin_call_acl_plugin should be called with dummyAttrs array.
The fix is

<pre>
diff --git a/ldap/servers/slapd/back-ldbm/findentry.c b/ldap/servers/slapd/back-ldbm/findentry.c
index 6e53a0aea..3dcb13185 100644
--- a/ldap/servers/slapd/back-ldbm/findentry.c
+++ b/ldap/servers/slapd/back-ldbm/findentry.c
@@ -207,7 +207,11 @@ find_entry_internal_dn(
                     break;
                 }
                 if (acl_type > 0) {
-                    err = plugin_call_acl_plugin(pb, me->ep_entry, NULL, NULL, acl_type,
+                    char *dummyAttr = "dummy#attr";
+                    char *dummyAttrs[2] = {NULL, NULL};
+
+                    dummyAttrs[0] = dummyAttr;
+                    err = plugin_call_acl_plugin(pb, me->ep_entry, dummyAttrs, NULL, acl_type,
                                                  ACLPLUGIN_ACCESS_DEFAULT, &errbuf);
                 }
                 if (((acl_type > 0) && err) || (op_type == SLAPI_OPERATION_BIND)) {
</pre>

It looks a dummy attribute is used in several plugins to check if the bound user is granted use of specific controls (account, vlvl, sync_repl). Those plugins are calling slapi_access_allowed (with a dummy attribute). So the patch should be changed to use the same approach.

Firstyear:
Sorry for the delay, I'll review this in two days time.

mreynolds389:
Okay, applied Thierry's changes in a new commit.  Tests look good, please review...

ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -207,7 +207,11 @@ find_entry_internal_dn(
                     break;
                 }
                 if (acl_type > 0) {
-                    err = plugin_call_acl_plugin(pb, me->ep_entry, NULL, NULL, acl_type,
+                    char *dummyAttr = "dummy#attr";
+                    char *dummyAttrs[2] = {NULL, NULL};
+
+                    dummyAttrs[0] = dummyAttr;
+                    err = plugin_call_acl_plugin(pb, me->ep_entry, dummyAttrs, NULL, acl_type,
```
tbordaz: Instead of calling directly the acl plugin you could also use slapi_access_allowed (like https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/sync/sync_refresh.c#L40)

dirsrvtests/tests/suites/acl/misc_test.py
```diff
@@ -408,14 +408,94 @@ def test_do_bind_as_201_distinct_users(topo, clean, aci_of_user):
...
+
+    # add aci
+    suffix.add('aci', ACI)
+
+    # Search for a non existent entry which should raise an exception
+    with pytest.raises(ldap.NO_SUCH_OBJECT):
+        conn = UserAccount(topo.standalone, USER_DN).bind(PW_DM)
+        subtree = Domain(conn, "ou=does_not_exist," + DEFAULT_SUFFIX)
+        subtree.get_attr_vals_utf8_l('objectclass')
+
```
tbordaz, line 516: It should also return NO_SUCH_OBJECT for base search and one-level search.
Also is it possible to test something like: 

subtree = Domain(conn, "ou=even_more,ou=does_not_exist," + DEFAULT_SUFFIX)

tbordaz (commented):
Except those minor comments the fix looks good to me

mreynolds389:
@tbordaz - applied the requested changes, please review...

tbordaz:
Thanks @mreynolds389 for your patience. You have my ACK

Issue #4526, opened 2021-01-11
sync_repl: when completing an operation in the pending list, it can select the wrong operation 


tbordaz:
**Issue Description**
With nested operations, a primary operation can trigger multiple nested operation. The pending list manages the list of nested operation. Each operation being identified in the pending list with the operation data structure. A problem with postop operation is that the same operation data structure can exist in the pending list. It can select the wrong one. In such case the pending list is never completed and flush to the sync_repl thread.

**Package Version and Platform:**
all versions

**Steps to Reproduce**
Difficult to reproduce. Sometime (1 out of 10) reproduced with IPA DNSSEC. 

**Expected results**
A pending list should complete and DNSSEC tests should not fail

tbordaz:
abb93243b..ba0e91b4f  master
96edca86a..9f9eb91e2  389-ds-base-1.4.4
fc16302d3..263a87bc2  389-ds-base-1.4.3

PR #4553, opened 2021-01-19
Issue 4526 - sync_repl: when completing an operation in the pending l…

tbordaz:
…ist, it can select the wrong operation

Bug description:
	When an operation complete, it was retrieved in the pending list with
	the address of the Operation structure. In case of POST OP nested operations
	the same address can be reused. So when completing an operation there could be
	a confusion which operation actually completed.

Fix description:
	The fix defines a new operation extension (sync_persist_extension_type).
	This operation extension contains an index (idx_pl) of the op_pl in the
	the pending list.
	And additional safety fix is to dump the pending list in case it becomes large (>10).
	The pending list is dumped with SLAPI_LOG_PLUGIN.

relates: https://github.com/389ds/389-ds-base/issues/4526

Reviewed by:

Platforms tested: F31

### DIFF
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -44,6 +44,9 @@ static int sync_release_connection(Slapi_PBlock *pb, Slapi_Connection *conn, Sla
  * per thread pending list of nested operation..
  * being a betxn_preop the pending list has the same order
  * that the server received the operation
+ *
+ * In case of DB_RETRY, this callback can be called several times
+ * The detection of the DB_RETRY is done via the operation extension
  */
 int
 sync_update_persist_betxn_pre_op(Slapi_PBlock *pb)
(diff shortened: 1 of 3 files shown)


ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -180,46 +222,50 @@ sync_update_persist_op(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *eprev, ber
...
     }
     
-#if DEBUG
-    /* dump the pending queue */
-    for (curr_op = prim_op; curr_op; curr_op = curr_op->next) {
-        char *flags_str;
-        char * entry_str;
+    /* for diagnostic of the pending list, dump its content if it is too long */
+    for (count = 0, curr_op = prim_op; curr_op; count++, curr_op = curr_op->next);
+    if (count > 10) {
```
Firstyear: Greater then 10? But then it goes to a loop with slapi_log_plugin? Shouldn't we check if the log level is set as well before we do the dump? or otherwise just log a plain warn?

Firstyear (changes requested):

Firstyear:
I have reviewed this again and I think it's okay. If you don't mind I'd like to give my mind a break and I'll review a second time if thats okay? That way I'm sure i understand it :)

tbordaz:
> I have reviewed this again and I think it's okay. If you don't mind I'd like to give my mind a break and I'll review a second time if thats okay? That way I'm sure i understand it :)

Thank you soooo much @Firstyear for your reviews. I know that I have been so boring with 'pending list' stabilization :)
No problem to wait for a second review.
Just a note regarding the changes. The fix contains basically two aspects:

- use of operation extension to uniquely identify (in the pending list) an operation, because a same operation block is prone to be reused in nested operation (postop operations)
- handling of db_retry that was the most recent issue with pending list (sometime detected with freeipa/dnssec tests). If an update is retried N times, then sync_update_persist_betxn_pre_op is called N times. The main job of sync_update_persist_betxn_pre_op is to add the operation in the pending list and to register its index in the operation extension. The fix uses the operation extension to detect that it is an update retry (op_ext != NULL).

Firstyear (approved):
Yep, rereading this I'm happy with it :) Thanks @tbordaz! Ack,.

PR #4556, opened 2021-01-20
Issue 4153 - Added a CI test

aadhikar:
Bug Description: The numSubordinates value doesn't always match the number of direct subordinate(s)

Relates: https://github.com/389ds/389-ds-base/issues/4153

Reviewed by: ??

### DIFF
--- a/dirsrvtests/tests/suites/monitor/monitor_test.py
+++ b/dirsrvtests/tests/suites/monitor/monitor_test.py
@@ -5,6 +5,9 @@ from lib389.monitor import *
 from lib389.backend import Backends, DatabaseConfig
 from lib389._constants import *
 from lib389.topologies import topology_st as topo
+from lib389._mapped_object import DSLdapObjects
+
+pytestmark = pytest.mark.tier1
 
 DEBUGGING = os.getenv("DEBUGGING", default=False)
 if DEBUGGING:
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/monitor/monitor_test.py
```diff
@@ -145,6 +146,35 @@ def test_monitor_backend(topo):
...
+        "(|(objectclass=*)(objectclass=ldapsubentry))" and scope as one.
+        4. Compare numSubordinates value with the number of sub-entries.
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Should be same
+    """
+
+    raw_objects = DSLdapObjects(topo.standalone, basedn='cn=monitor')
```
droideck, line 173: I think we have 'Monitor()' class for that case (lib389.monitor)...

droideck (approved):
Besides the small issue - LGTM!

dirsrvtests/tests/suites/monitor/monitor_test.py
aadhikar: I can extract the numSubordinates value while using Monitor class but  I am not sure I can use filters with it.

droideck: You should be able, I think

droideck (commented):
Okay, discard my previous comment. It's impossible right now.
I've opened an issue for improvement though - #4559

PR #4561, opened 2021-01-21
Issue 4560 - Enable additional python_files '*_test_*' discovery

### LINKED ISSUE #4560: Enable additional python_files '*_test_*' discovery
We have tests that currently are not discovered with 'pytest' - dbgen_test_usan.py
Also, it will be nice to have an option to name the test suites accordingly to the topology they use.
(i.e. replication/regression_test_m2.py, replication/regression_test_m3.py, etc.)

 Add a new rule for py.test discovery - '*_test_*'.

droideck:
Bug Description: We have tests that currently are not discovered
with 'pytest' - dbgen_test_usan.py
Also, it will be nice to have an option to name the test suites
accordingly to the topology they use.
(i.e. repl/regression_test_m2.py, repl/regression_test_m3.py, etc.)

Fix Description: Add a new rule for py.test discovery - 'test'.

Fixes: #4560

Reviewed by: ?

### DIFF
--- a/dirsrvtests/pytest.ini
+++ b/dirsrvtests/pytest.ini
@@ -4,3 +4,4 @@ markers =
     tier1: mark a test as part of tier1
     tier2: mark a test as part of tier2
     tier3: mark a test as part of tier3
+python_files = test_*.py *_test.py *_test_*.py


sgouvern (commented):
It seems a nice change to me. 
Ack for me, but it would be good to have Viktor's review as well.

vashirov:
I'd prefer to follow the existing naming convention and rename `dbgen_test_usan.py` to `dbgen_usan_test.py`. I also don't see anything wrong with `repl/regression_m2_test.py`. But as long as the automation works, I'm fine with it. It's all bikeshedding, really.

droideck:
> I'd prefer to follow the existing naming convention and rename `dbgen_test_usan.py` to `dbgen_usan_test.py`. I also don't see anything wrong with `repl/regression_m2_test.py`. But as long as the automation works, I'm fine with it. It's all bikeshedding, really.

There are two points I like:

1. It's more structured in regards - suite_name_test[_optional suffix which specifies difference from other files].py.
    So it's always easy to see and 'parse' the meaning.
2. It doesn't break existing logic but improves it. And it just feels natural to be able to have '_test_' keyword at any place.

I understand that this is an opinionated approach so I am asking our QE team and I won't push if I don't have both acks from you, @vashirov @sgouvern :)

vashirov (changes requested):
I've looked at our automation and this change breaks it. We use betelgeuse for test case parsing and import into our TCMS, it requires that the test module starts with or ends with `test`: https://github.com/SatelliteQE/betelgeuse/blob/b36cc1bbc99441427ea3893686f507ecdcf08c87/betelgeuse/collector.py#L134
So if the test case has `_test_` in the middle, it won't be collected. 

I also don't feel confident about `dbgen_test_usan.py`. I think the original test case should be rewritten to support parametrized arguments for CLI (ints or strings) instead of generating whole new test module just for this.

droideck:
> I've looked at our automation and this change breaks it. We use betelgeuse for test case parsing and import into our TCMS, it requires that the test module starts with or ends with `test`: https://github.com/SatelliteQE/betelgeuse/blob/b36cc1bbc99441427ea3893686f507ecdcf08c87/betelgeuse/collector.py#L134
> So if the test case has `_test_` in the middle, it won't be collected.

Okay, then, for sure, we can't use `_test_`. I will drop the PR then...
 
> I also don't feel confident about `dbgen_test_usan.py`. I think the original test case should be rewritten to support parametrized arguments for CLI (ints or strings) instead of generating whole new test module just for this.

I'll open an issue for that so QE can proceed. (or someone from Dev will check - we'll define that in the following plannings).

PR #4564, opened 2021-01-22
Update dscontainer

Jackbennett:
When bringing a container up I don't see why this can't be configured by the env like everything else.

With this change we can bring a directory up via kubernetes and an init job.

If you don't want to do this can we split the start function `-r` from the create config step perhaps with `-c`

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -271,16 +271,20 @@ def begin_magic():
 
         # Create the marker to say we exist. This is also a good writable permissions
         # test for the volume.
-        with open(DSRC_CONTAINER, 'w') as f:
-            f.write("""
+        basedn = '# basedn = dc=example,dc=com'
+        if suffix := os.getenv("SUFFIX_NAME"):
+            basedn = f'basedn = {suffix}'
+        config_file = """
 [localhost]
 # Note that '/' is replaced to '%%2f' for ldapi url format.
 # So this is pointing to /data/run/slapd-localhost.socket
 uri = ldapi://%%2fdata%%2frun%%2fslapd-localhost.socket
 binddn = cn=Directory Manager
 # Set your basedn here
-# basedn = dc=example,dc=com
-            """)
+{0}
+""".format(basedn)
+        with open(DSRC_CONTAINER, 'w') as f:
+            f.write(config_file)
         os.chmod(DSRC_CONTAINER, 0o755)
 
     # Setup TLS from PEM files as required.


Firstyear:
Hey @Jackbennett. I think this is a good change in this context. If you set a suffix, we also could extend this to automatically create the first database as well. What do you think?

mreynolds389:
> Hey @Jackbennett. I think this is a good change in this context. If you set a suffix, we also could extend this to automatically create the first database as well. What do you think?

Wasn't there a plan to provide some type of config file for other settings (including replication, etc)?

Jackbennett:
>If you set a suffix, we also could extend this to automatically create the first database as well. What do you think?

That'd be helpful. We bring up a deployment which also runs [this k8s job](https://github.com/gabibbo97/k8s-misc/blob/master/389ds/init-job.yaml) with an added `dsidm -D "$DS_BIND_DN" -w "$(DS_BIND_PW)" $LDAP_SEVER initialise` roughly line:36 to get things ready. It's fairly ugly however expecting a few jobs to fail at first run until it connects once 389 is ready.

If all that is needed and can be rolled into the startup. Perhaps we could also populate `InstallLdifFile` as [legacy documented (See End)](https://directory.fedoraproject.org/docs/389ds/legacy/install-guide.html) with some kind of pre-load might be nice. But perhaps you'd just replicate in from some pre-config environment?

This change as-is logs a rather ugly "Using ConfigParser to read a file without section name" python err, although the file was  written correctly. Struggling to trigger that exact combination again. Might just want to check the container startup output when running. I don't know why, I can't see the code using `ConfigParser` at all.

vashirov:
@Firstyear, I have a very similar use case: create an instance, add backend and suffix, change default DM password, configure plugins (memberOf), disable anonymous access and replace default certs. All of this has to be done at the build stage, because I can't use init container (I'm not running this in kubernetes). I was thinking about extending `_begin_environment_config()`, but ended up adding another parameter to the `dscontainer` that executes an init script with custom `dsconf` commands, right after `ns-slapd` is up and running, and terminates `ns-slapd` once the script is done. What do you think about this approach?

Firstyear:
I do like the sentiment to this, but I'm not sure about scripting it like this. dsconf is very "task oriented" not "state oriented" so you would need to account for "what if the plugin is already enabled".

I think to make this as "supportable" as possible, the best approach would be a toml/ini/yaml file that contains "desired state" that we want the instance in the container to arrive at. That way it can be configuration managed, deployed to 1 or many containers, and then it's stateful so we can apply it and ensure the instance is at that state.

So I think longer term that would be the better approach, and we could probably have it expose to dsconf <instance> apply-state or similar so that non-container instances can benefit. What do you think?

vashirov:
> I do like the sentiment to this, but I'm not sure about scripting it like this. dsconf is very "task oriented" not "state oriented" so you would need to account for "what if the plugin is already enabled".

I like that `dscontainer` is very minimal, it doesn't assume anything about the environment. This is good for container orchestration where you can put the logic not in the DS container itself, but in some other patterns instead such as [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) or [operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/).

The problem I'm facing doesn't involve container orchestration at all, so these patterns can't be applied. I'm working with internal teams to enable testing against RHDS for their products. All they want is a generic LDAP server in a container that is preconfigured with a suffix, test users, etc. When they run their tests, they spin up a throwaway container with LDAP server and run their tests against it. Once the tests are done, container is deleted. So during the lifecycle of such container they don't interact with it using `dsconf` at runtime at all. The only time `dsconf` is used is at the container build time, where state is known and `dsconf` is very forgiving:
```
[root@server-f33 ds]# dsconf server-f33 plugin memberof enable 
Enabled plugin 'MemberOf Plugin'
[root@server-f33 ds]# dsconf server-f33 plugin memberof enable 
Plugin 'MemberOf Plugin' already enabled
[root@server-f33 ds]# echo $?
0
```

So with that scenario in mind it would be very helpful to run `dscontainer` at the build time, do some init stuff, shutdown the server and seal the container. 

> I think to make this as "supportable" as possible, the best approach would be a toml/ini/yaml file that contains "desired state" that we want the instance in the container to arrive at. That way it can be configuration managed, deployed to 1 or many containers, and then it's stateful so we can apply it and ensure the instance is at that state.
> 
> So I think longer term that would be the better approach, and we could probably have it expose to dsconf apply-state or similar so that non-container instances can benefit. What do you think?

I think this is a perfect fit for Kubernetes operator. You specify a desired state and its reconciliation loop will drive the actual state to the desired state. And it can use lib389 or ds* tools inside the container as an API to achieve that. 
I've started to work on [such operator ](https://github.com/vashirov/ds-operator) and the plan is to have that as a "supportable" solution to use RHDS in Kubernetes/OpenShift.

And if we're speaking of a scenario outside of container orchestration (think traditional VMs, bare metal, etc.), Ansible (or any other config management system) would be a better fit. It can use lib389 or ds* tools as primitives. From our side we can ensure that lib389 provides machine readable output to describe the state, so that Ansible/operator can make right decisions.

Jackbennett:
> All they want is a generic LDAP server in a container that is preconfigured with a suffix, test users, etc. When they run their tests, they spin up a throwaway container with LDAP server and run their tests against it.

This is the current motivation for the patch. 

> It would be very helpful to run dscontainer at the build time

To split the setup operations from the init function starting the container would go a long way to having this work before implementing a whole operator.

vashirov:
@Firstyear, any objections for this the PR?

@Jackbennett, here's a PoC of oneshot script with `dscontainer`: https://github.com/vashirov/ds-container/tree/dscontainer_oneshot/fedora-minimal

Firstyear:
Sorry, I thought it was an issue not a PR so I had forgotten to review. I'm happy with it, so have merged it with a fixed commit string in the squash process. Thanks @Jackbennett. I think we should open an issue about better container orchestration though.

mreynolds389:
> Sorry, I thought it was an issue not a PR so I had forgotten to review. I'm happy with it, so have merged it with a fixed commit string in the squash process. Thanks @Jackbennett. I think we should open an issue about better container orchestration though.

FYI, the git commit message did not comply with our format:

mreynolds389:
Added commit to 1.4.4 branch:

b8c4510b1..ffa4563dd  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

Firstyear:
> > Sorry, I thought it was an issue not a PR so I had forgotten to review. I'm happy with it, so have merged it with a fixed commit string in the squash process. Thanks @Jackbennett. I think we should open an issue about better container orchestration though.
> 
> FYI, the git commit message did not comply with our format:
> 
> https://www.port389.org/docs/389ds/contributing.html#getting-the-patch-ready

During the squash/merge I edited the commit to be compliant? Did it not "stick"?

Issue #4324, opened 2020-09-14
Performance search rate: change entry cache monitor to recursive pthread mutex


tbordaz:
Entry cache (entry and dn) are protected by a PRMonitor. Changing this Monitor into a recursive pthread mutex (cache aligned) increases throughput by 8%

Firstyear:
I think I've looked at this in the past, that PRLock and PRMonitor are both slower than pthread mutex and recursive pthread mutexs. So it's probably not the only place in the code base that this would create a benefit.

tbordaz:
fe0f6152f..abb93243b  master
d641e8528..14f67ebeb  389-ds-base-1.4.4
6c8755687..fc16302d3  389-ds-base-1.4.3

mreynolds389:
8d40f2784..d4c2ab388  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
7f20bccb0..9a93ec7c1  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

Fixes crash on aarch64

droideck:
de65f12a0..6d5537900  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
   d4c4ff1b9..8564399c5  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
   9a47b3d22..7ba233bf7  389-ds-base-2.0 -> 389-ds-base-2.0
   e28c36c84..6c95604f6  389-ds-base-2.1 -> 389-ds-base-2.1
   3043f85e9..21314ec9e  389-ds-base-2.2 -> 389-ds-base-2.2

PR #4569, opened 2021-01-25
Issue 4324 - Performance search rate: change entry cache monitor to r…

tbordaz:
…ecursive pthread mutex

Bug description:
	The entry cache is protected with recursive mutex. Currently it is
	implemented using PR_Monitor (NSPR). When the entry cache mutex
	becomes the bottleneck (for example base search searchrate on
	the same entry), using pthread recursive mutex gives 8% benefit.

Fix description:
	Changing the c_mutex from PR_Monitor to pthread recursive mutex

relates: https://github.com/389ds/389-ds-base/issues/4324

Reviewed by:

Platforms tested:

### DIFF
--- a/ldap/servers/slapd/ch_malloc.c
+++ b/ldap/servers/slapd/ch_malloc.c
@@ -349,3 +349,57 @@ slapi_ct_memcmp(const void *p1, const void *p2, size_t n)
     }
     return result;
 }
+
+#define SYS_CACHELINE_SIZE "/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size"
+static
+size_t cache_line_size() 
+{
+    PRFileDesc *prfd;
+    char buf[10] = {0};
+    size_t i = 0;
+
+    prfd = PR_Open(SYS_CACHELINE_SIZE, PR_RDONLY, SLAPD_DEFAULT_FILE_MODE);
+    if (PR_Read(prfd, buf, sizeof (buf)) < 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "cache_line_size", "Can not read %s\n", SYS_CACHELINE_SIZE);
+        PR_Close(prfd);
+        goto done;
+    }
+    PR_Close(prfd);
+    i = atoi(buf);
+
+done:
+    if ((i == 0) || (i > 64)) {
+        slapi_log_err(SLAPI_LOG_INFO, "cache_line_size", "unexpected cpu cache line size => assign 64\n");
+        i = 64;
+    }
+    return i;
+}
+
+pthread_mutex_t*
+slapi_pthread_mutex_alloc(int type)
+{
+    int32_t cacheLineSiz = cache_line_size();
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/ch_malloc.c
```diff
@@ -349,3 +349,57 @@ slapi_ct_memcmp(const void *p1, const void *p2, size_t n)
...
+    pthread_mutexattr_init(&monitor_attr);
+    pthread_mutexattr_settype(&monitor_attr, type);
+    if (pthread_mutex_init(&orig_mutex, &monitor_attr) != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "slapi_pthread_mutex_alloc", "pthread_mutex_init failed\n");
+        return NULL;
+    }
+    /* now align it to buffer aligned to the cpu cache line */
+    aligned_mutex = (pthread_mutex_t *) slapi_ch_memalign(aligned_size, cacheLineSiz);
+    memmove(aligned_mutex, (const void *) &orig_mutex, sizeof(pthread_mutex_t));
+    return(aligned_mutex);
```
mreynolds389, line 397: I don't know much about this aligning, could you elaborate on this?

Regardless this all looks fine, I just want to learn more about what you are doing with the CPU alignment.

droideck: Looks like it's an alloc that works best for some of cpu operations.
It was used by nunc_stans too, apparently:)
https://github.com/389ds/389-ds-base/blob/master/ldap/servers/slapd/slapi-plugin.h#L5792

droideck (approved):
I've run a few tests (with ASAN) - and it looks good:)

And, as far as I can understand the code - it looks good!

ldap/servers/slapd/ch_malloc.c
tbordaz: CPU cache loads memory per "lines". For example on i7 a cpu cache line is 64bytes.
If a sensitive data (like a mutex) is split over two lines it requires two loads. Another risk  is that several locks are located in the same cache line, so there is a risk of "false" contention if those locks are accessed at the same time.

ldap/servers/slapd/ch_malloc.c
```diff
@@ -349,3 +349,57 @@ slapi_ct_memcmp(const void *p1, const void *p2, size_t n)
...
+        slapi_log_err(SLAPI_LOG_INFO, "cache_line_size", "unexpected cpu cache line size => assign 64\n");
+        i = 64;
+    }
+    return i;
+}
+
+pthread_mutex_t*
+slapi_pthread_mutex_alloc(int type)
+{
+    int32_t cacheLineSiz = cache_line_size();
```
Firstyear, line 381: We shouldn't call this every alloc because the value is effectively static. Is there a way we could make this a global and just call detect_cache_line_size once in main()?

ldap/servers/slapd/ch_malloc.c
mreynolds389: thanks for the info guys

ldap/servers/slapd/ch_malloc.c
mreynolds389: +100

PR #4570, opened 2021-01-26
Issue 4513 - Add DS version check to SSL version test

droideck:
Description: Starting from Fedora 33, cryptographic protocols
(TLS 1.0 and TLS 1.1) were moved to LEGACY
Add a 389-ds-base version check so we don't check for the policies
if DS is newer than 1.4.3.
https://fedoraproject.org/wiki/Changes/StrongCryptoSettings2

Relates: #4513

Reviewed by: ?


mreynolds389:
LGTM

Issue #4563, opened 2021-01-22
Failure on s390x: 'Fails to split RDN "o=pki-tomcat-CA" into components'


tjaalton:
**Issue Description**

pkispawn fails on s390x, and this is from the 389 log:

[22/Jan/2021:09:04:35.687297399 +0000] - ERR - attrcrypt_cipher_init - No symmetric key found for cipher AES in backend pki-tomcat-CA, attempting to create one...
[22/Jan/2021:09:04:35.692412357 +0000] - INFO - attrcrypt_cipher_init - Key for cipher AES successfully generated and stored
[22/Jan/2021:09:04:35.692791984 +0000] - ERR - attrcrypt_cipher_init - No symmetric key found for cipher 3DES in backend pki-tomcat-CA, attempting to create one...
[22/Jan/2021:09:04:35.697798362 +0000] - INFO - attrcrypt_cipher_init - Key for cipher 3DES successfully generated and stored
[22/Jan/2021:09:04:35.887915222 +0000] - ERR - entry_get_rdn_mods - Fails to split RDN "o=pki-tomcat-CA" into components

This started after 1.4.4.4, and the error message would suggest that something is wrong with commit 2ccd0bed4e60e.

 s390x is big-endian, and other architectures don't have this issue, so endianness might have something to do with it?

**Package Version and Platform:**
 - Platform: Debian/Ubuntu
 - Package and version: 389-ds-base-1.4.4.10-1

tjaalton:
reverting that commit allowed pkispawn to work again

slyon:
I can confirm the test works again after the above mentioned commit is reverted.

Here are the logs of the pkispawn test:
https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-hirsute-slyon-testing/hirsute/s390x/d/dogtag-pki/20210122_114408_9ecb1@/log.gz

mreynolds389:
To help reproduce this internally (outside of pkispawn), what LDAP operation is failing?  And what is that operation trying to do?

tjaalton:
'INFO: Adding aci into o=pki-tomcat-CA'

and then it fails. Trying to look deeper, but a quick grep gives no hits in dogtag..

mreynolds389:
> 'INFO: Adding aci into o=pki-tomcat-CA'
> 
[...]

PR #4573, opened 2021-01-27
Issue 4563 - Failure on s390x: 'Fails to split RDN "o=pki-tomcat-CA" …

tbordaz:
…into components'

Bug description:
	SLAPI_OPERATION_TYPE is a stored/read as an int (slapi_pblock_get/set).
	This although the storage field is an unsigned long.
	Calling slapi_pblock_get with an long (8 btyes) destination creates
	a problem on big-endian (s390x).

Fix description:
	Define destination op_type as an int (4 bytes)

relates: https://github.com/389ds/389-ds-base/issues/4563

Reviewed by:

Platforms tested: F31 (little endian)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_modify.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_modify.c
@@ -216,7 +216,7 @@ error:
 int32_t
 entry_get_rdn_mods(Slapi_PBlock *pb, Slapi_Entry *entry, CSN *csn, int repl_op, Slapi_Mods **smods_ret)
 {
-    unsigned long op_type = SLAPI_OPERATION_NONE;
+    int op_type = SLAPI_OPERATION_NONE;
     char *new_rdn = NULL;
     char **dns = NULL;
     char **rdns = NULL;


slyon:
Applying this patch to our 389-ds-base package and re-running the (previously) regressed dogtag-pki test suite makes it PASS:
https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-hirsute-slyon-testing/hirsute/s390x/d/dogtag-pki/20210127_140849_b37ed@/log.gz

mreynolds389:
I know this fixes it, but should we be using strong int types here?  int32_t?

tbordaz:
@slyon thanks for precious help.
@mreynolds389 there are many 'int' set/get from pblock. I do not think it worth changing it (and all of them)

mreynolds389:
> @slyon thanks for precious help.
> @mreynolds389 there are many 'int' set/get from pblock. I do not think it worth changing it (and all of them)

I was not suggesting changing all of them, just where it might be critical, but the problem here was simply the wrong type being used.  Anyway, ACK from me

Firstyear:
Ack from me too.

Issue #4577, opened 2021-01-27
Add GitHub actions


vashirov:
**Issue Description**
Using GitHub actions we should be able to run tests, static analysis, compile using gcc/clang, etc. on each PR or push to the repository. 

- [x] compilation using gcc and clang
- [x] covscan
- [x] cppcheck
- [x] gcc static analyzer
- [x] clang static analyzer
- [ ] lib389 tests
- [x] dirsrv tests

Firstyear:
Is something going weird here? Had a ton of notifications about this ....

vashirov:
Sorry for the spam, I was updating my PR, didn't realize it generated so much noise :/

droideck:
@vashirov I think you've made more progress than noted (in the checkboxes in the description). Or are we done completely?

vashirov:
We're not done completely. I have patches for the remaining items, but they need some love.
I was hesitant to add more checks because the existing ones generated a lot of warnings, but no one was looking at them. GitHub also has a limit of 50 visible annotations per run, so adding more checks would generate warnings that are simply not visible.

But now I see that we're down to 32 in the latest run: https://github.com/389ds/389-ds-base/actions/runs/1855623293
So I will revive my patches and submit PRs.

For lib389 tests we have https://github.com/389ds/389-ds-base/issues/5005 that needs to be fixed first, otherwise the check would always fail. I don't have capacity to work on this during Q1, so if someone else has free cycles to work on this, it would be great.

PR #4578, opened 2021-01-27

vashirov:
Add first set of actions to compile project using gcc and clang.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,0 +1,67 @@
+name: Compile
+on:
+  - pull_request
+  - push
+
+jobs:
+  compile:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        name:
+          - Fedora 33 GCC
+          - Fedora 33 GCC strict
+          - Fedora 33 Clang
+          - Fedora 33 Clang strict
+          - Fedora 33 Clang -Weverything
+
+        include:
+          - name: Fedora 33 GCC
+            image: fedora:33
+            compiler: gcc
+            cpp-compiler: g++
+
+          - name: Fedora 33 GCC strict
+            image: fedora:33
+            compiler: gcc
+            cpp-compiler: g++
+            cflags:  "-O2 -g -Wall -Wextra -Wundef -Wpointer-arith -Wfloat-equal \
+            -Wstrict-prototypes -Wstrict-overflow=5 -Wwrite-strings -Winit-self \
+            -Wuninitialized -Wno-sign-compare -Wshadow -Wformat-security"
+
+          - name: Fedora 33 Clang
+            image: fedora:33
+            compiler: clang
+            cpp-compiler: clang++
+            cflags: "-Wno-error=unused-command-line-argument"
+
+          - name: Fedora 33 Clang strict
+            image: fedora:33
+            compiler: clang
+            cpp-compiler: clang++
+            cflags: "-Wall -Wextra -Wno-error=unused-command-line-argument"
+
+          - name: Fedora 33 Clang -Weverything
+            image: fedora:33
+            compiler: clang
+            cpp-compiler: clang++
+            cflags: "-Weverything -Wno-error=unused-command-line-argument"
+
+    container:
+      image: ${{ matrix.image }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build dependencies
+        run: ${GITHUB_WORKSPACE}/.github/scripts/install_deps.sh
+
+      - uses: ammaraskar/gcc-problem-matcher@master
+      - name: Build using ${{ matrix.compiler }}
+        run: cd $GITHUB_WORKSPACE && autoreconf -fvi && ./configure && make V=0
+        env:
+          CC: ${{ matrix.compiler }}
+          CXX: ${{ matrix.cpp-compiler }}
+          CFLAGS: ${{ matrix.cflags || env.CFLAGS }}
+          CXXFLAGS: ${{ matrix.cxxflags || env.CXXFLAGS }}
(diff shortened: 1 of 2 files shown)


vashirov:
Example job run: https://github.com/vashirov/389-ds-base/actions/runs/515832978

vashirov:
Example job run with an error/warning introduced in the PR: https://github.com/vashirov/389-ds-base/actions/runs/515910523

droideck:
Just 1 day ago I've started to put together the patch like that one:)

Also, it'll be nice to add compiler warning checks (as per meeting discussion)

But anyway, gonna check that tomorrow:)
(from a first glance, it looks great already)

vashirov:
Yes, checking for compiler warnings is enabled via different set of cflags, these warnings are listed on the run page (see my second comment).
I also have few other checks coming: coverity scan, cppcheck, clang-format and py.test (lib389 for now, dirsrvtests a bit later).

Firstyear:
I think this looks great, ack from me.

Firstyear (approved):

.github/workflows/compile.yml
```diff
@@ -0,0 +1,77 @@
...
+        name:
+          - Fedora 33 GCC
+          - Fedora 33 GCC strict
+          - Fedora 33 Clang
+          - Fedora 33 Clang strict
+          - Fedora 33 Clang -Weverything
+
+        include:
+          - name: Fedora 33 GCC
+            image: fedora:33
```
droideck, line 21: I haven't tried yet but can we use Fedora 34 here?
There is image fedora:34 but I haven't checked if there will be any other issues.

In general, I think, it's better to test 389-ds 'master' on the distro for which we build RPMs - master==fedora-rawhide.

.github/workflows/compile.yml
```diff
@@ -0,0 +1,77 @@
...
+bind-utils bzip2 bzip2-devel cargo clang coreutils-common \
+cracklib-devel cracklib-dicts dnf-utils doxygen gcc gcc-c++ \
+git hostname icu krb5-devel libasan libcmocka-devel libdb-devel \
+libevent-devel libicu-devel libtalloc-devel libtevent-devel \
+libtool libubsan lm_sensors-devel make net-snmp-devel net-tools \
+nspr-devel nss-devel openldap-devel openssl openssl-devel pam-devel \
+pcre-devel procps-ng psmisc python3-argcomplete python3-argparse-manpage \
+python3-devel python3-ldap python3-libselinux python3-pip \
+python3-policycoreutils python3-pyasn1 python3-pyasn1-modules rpm-build \
+rsync rust systemd-devel uuid wget zlib-devel"
```
droideck: I think it will look cleaner line by line because it makes it easier to see the list of dependencies and it's a cleaner git history if we edit any of the dependencies.

I think YAML allows the syntax like this:

```
run: >
  dnf -y install
  389-ds-base
  389-ds-base-snmp
  autoconf
  automake
  ...
```

droideck (changes requested):

.github/workflows/compile.yml
```diff
@@ -0,0 +1,77 @@
+name: Build policies
```
droideck: Is it a typo? I am not sure I follow which 'policies' are referred here...

.github/workflows/compile.yml
vashirov: > I haven't tried yet but can we use Fedora 34 here?
There is image fedora:34 but I haven't checked if there will be any other issues.

We can, but it's not branched yet, it's still rawhide. We expect some unrelated failures due to instability of rawhide. I'd prefer to add F34 once it's branched.

> In general, I think, it's better to test 389-ds 'master' on the distro for which we build RPMs - master==fedora-rawhide.

The point of this test is to use stable tools (gcc, clang) to test unstable code (master). If tools are from an unstable version, we can get unpredictable results. I can add another variant to the matrix, but I'm not sure how useful it would be.

.github/workflows/compile.yml
vashirov: Perhaps it would be better if I put this command in a script? Then it can be reused for other jobs.

.github/workflows/compile.yml
vashirov: It refers to build policies (on push, on pull request), but I think I lost another `-name` somewhere to name the actual test.

.github/workflows/compile.yml
droideck: Okay, sounds good and I think you're right.

The only issues I'm worried about are the Fedora changes related issues (like crypto policies may be different on F33 and rawhide, etc.).
But it's really rare and we can catch it separately.

.github/workflows/compile.yml
droideck: Now I'm thinking, it'll be even better to use 'one source of truth' then...

We have dependencies in 'spec' file, so may be we can get them from there? Like we do in https://www.port389.org/docs/389ds/development/building.html

```
# git clone https://github.com/389ds/389-ds-base.git
# cd 389-ds-base
# sudo dnf install `grep "^BuildRequires" rpm/389-ds-base.spec.in | awk '{print $2}' | sed -e "s/%{python3_pkgversion}/3/"`
```

vashirov: I can't use this on Debian or SUSE. It's also out of sync from Fedora spec file (doesn't have `make` for example), it also pulls in `npm` and `nodejs` that are not needed for this test. So I'd prefer to maintain different lists of dependencies for each distro.

droideck (approved):
Okay, thank you for resolving my annoying queries :D
You have my ack. Great job!

vashirov:
@droideck, those were good questions! I will address inconsistent dependencies in the following PRs and add more flexibility, but for now I want to have MVP that can help catch issues today. So thank you for your review, appreciate that :)

PR #4580, opened 2021-01-28
Issue 4348 - Add tests for dsidm

bsimonova:
Description:
Created tests for dsidm user option and enhanced
the src/lib389/lib389/cli_idm/__init__.py and src/lib389/lib389/cli_base/__init__.py
so the output gets caught to topology LogCapture to compare the results.

Relates: https://github.com/389ds/389-ds-base/issues/4348

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsidm_user_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_user_test.py
@@ -1,0 +1,426 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import time
+import subprocess
+import pytest
+import logging
+import os
+
+from lib389 import DEFAULT_SUFFIX
+from lib389.cli_idm.user import list, get, get_dn, create, delete, modify, rename
+from lib389.topologies import topology_st
+from lib389.cli_base import FakeArgs
+from lib389.utils import ds_is_older, ensure_str
+from lib389.idm.user import nsUserAccounts
+from . import check_value_in_log_and_reset
+
+pytestmark = pytest.mark.tier0
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="function")
+def create_test_user(topology_st, request):
+    user_name = 'test_user_1000'
+    users = nsUserAccounts(topology_st.standalone, DEFAULT_SUFFIX)
+
+    log.info('Create test user')
+    if users.exists(user_name):
+        test_user = users.get(user_name)
+        test_user.delete()
+    else:
+        test_user = users.create_test_user()
+
+    def fin():
+        log.info('Delete test user')
+        if test_user.exists():
+            test_user.delete()
+
+    request.addfinalizer(fin)
+
+
+@pytest.mark.skipif(ds_is_older("1.4.2"), reason="Not implemented")
+def test_dsidm_user_list(topology_st, create_test_user):
+    """ Test dsidm user list option
+
+    :id: a7400ac2-b629-4507-bc05-c6402a5b437b
+    :setup: Standalone instance
+    :steps:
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/clu/dsidm_user_test.py
```diff
@@ -0,0 +1,426 @@
...
+    topology_st.logcap.flush()
+
+    log.info('Test dsidm user list without json')
+    list(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+    check_value_in_log_and_reset(topology_st, check_value=user_value)
+
+    log.info('Test dsidm user list with json')
+    args.json = True
+    list(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+    check_value_in_log_and_reset(topology_st, content_list=json_list, check_value=user_value)
```
droideck, line 92: I have an assertion error at that point:

```
INFO:dirsrvtests.tests.suites.clu.dsidm_user_test:Test dsidm user list with json
{
    "type": "list",
    "items": [
        "demo_user",
        "test_user_1000"
    ]
}
INFO:dirsrvtests.tests.suites.clu:Check if content is present in output
FAILED
```

Here:

```
    def check_value_in_log_and_reset(topology, content_list=None, content_list2=None, check_value=None,
                                     check_value_not=None):
        if content_list2 is not None:
            log.info('Check if content is present in output')
            for item in content_list + content_list2:
                assert topology.logcap.contains(item)

        if content_list is not None:
            log.info('Check if content is present in output')
            for item in content_list:
>               assert topology.logcap.contains(item)
E               AssertionError
```

droideck (changes requested):

dirsrvtests/tests/suites/clu/dsidm_user_test.py
bsimonova: Weird, it passes for me on Fedora, Rhel8.
On what version do you test? 
It is dependent on changes to lib389/cli_idm/__init__.py where I added printing the values to log.

droideck: Sorry for the fuss, it was my env issue :)

dirsrvtests/tests/suites/clu/dsidm_user_test.py
```diff
@@ -0,0 +1,426 @@
...
+    check_value_in_log_and_reset(topology_st, content_list=user_content)
+
+    log.info('Test dsidm user get with json')
+    args.json = True
+    get(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+    check_value_in_log_and_reset(topology_st, content_list=json_content)
+
+
+@pytest.mark.bz1893667
+@pytest.mark.skip("Will fail because of bz1893667")
```
droideck: Probably, `@pytest.mark.xfail(reason="") is a better option here.

droideck (approved):
Besides the small issue - LGTM! Ack:)

vashirov:
This test fails with ERROR:
```
Error evaluating 'xfail' expression
    Will fail because of bz1893667
          ^
SyntaxError: invalid syntax
```
Looks like `pytest.mark.xfail` is missing `reason`.

src/lib389/lib389/cli_idm/__init__.py
```diff
@@ -90,6 +94,8 @@ def _generic_get(inst, basedn, log, manager_class, selector, args=None):
     if args and args.json:
         o = mc.get(selector, json=True)
         print(o)
```
mreynolds389, line 96: With changes that were made in the last year, I think we can drop the "print" statements completely, and just use log.info().  Same for the changes below this...

bsimonova: Thanks for review, I am fixing it here: https://github.com/389ds/389-ds-base/pull/4594

bsimonova:
@vashirov 

> This test fails with ERROR:
> 
> ```
> Error evaluating 'xfail' expression
>     Will fail because of bz1893667
>           ^
> SyntaxError: invalid syntax
> ```
> 
> Looks like `pytest.mark.xfail` is missing `reason`.

Sorry about that, I did not get a notification.
Fixing the issue here: https://github.com/389ds/389-ds-base/pull/4594

PR #4599, opened 2021-02-06

vashirov:
Description:

* Update compilation tests to use prebuilt container images
* Add pytest workflow for dirsrvtests

Test suite matrix is generated automatically based
on the contents of the tests suites directory.
Replication test suite is split up futher to speed up test
execution.

Reviewed by: ??


vashirov:
A note about container images. They are currently at https://github.com/vashirov/ci-images and built nightly or on git push. The resulting containers are then pushed to https://quay.io/repository/389ds/ci-images?tab=tags. I will move repo under 389ds org once this is merged.

vashirov:
Ugh, looks like Publish test results action doesn't work if the PR is not from the same org/repo, but from a fork. Which makes it useless, since it can't comment on PRs. So we won't get comments like this: https://github.com/vashirov/389-ds-base/pull/5#issuecomment-774426098
I will remove it then.

droideck (approved):
Looks good to me!

These `DeprecationWarning` are really annoying though... We should fix them sooner then later, I think, as it's hard to scroll through the results.

And yep, we have only 4 failed tests! Pretty soon we'll have 100% CI PASS rate with PR which is cool :)

vashirov:
> These `DeprecationWarning` are really annoying though... We should fix them sooner then later, I think, as it's hard to scroll through the results.

`py.test` prints at the end `short test summary info` which has all the failing tests. I'm also investigating how to add tests summary as a comment to the PR itself (there is an action, but it works only for original repos, not when a PR is done from a fork).
And I agree, we should fix these warnings. 

> And yep, we have only 4 failed tests! Pretty soon we'll have 100% CI PASS rate with PR which is cool :)

4 failing test suites, there are more individual test failures :/ But yeah, let's fix them :)

Issue #4600, opened 2021-02-08
performance modify rate: reduce lock contention on the object extension factory


tbordaz:
**Issue Description**
/opt/ldapbench/bin/modrate --hostname instance --port 389 --bindDN 'cn=Directory Manager' --bindPassword xxx --entryDN id=user.[1-10000],ou=People,dc=example,dc=com" --attribute street --attribute description --valueLength 12 --numThreads 5

During these modifies, target entry is allocated/duplicated/free. entry structure contains a object (entry) extension that is allocated via a factory slapi_register_object_extension.
ATM the purpose of this factory is to store  "unhashed#user#password" in case of update of the password (pw_exp_init).

The factory is protected via a NSPR simple mutex and is the second hottest lock in contention
<pre>
#        ----------- Cacheline ----------    Total      Tot  ----- LLC Load Hitm -----  ---- Store Reference ----  --- Load Dram ----      LLC    Total
# Index             Address  Node  PA cnt  records     Hitm    Total      Lcl      Rmt    Total    L1Hit   L1Miss       Lcl       Rmt  Ld Miss    Loads
# .....  ..................  ....  ......  .......  .......  .......  .......  .......  .......  .......  .......  ........  ........  .......  .......
#
      0      0x7f84e9c28500     0     158     1877   16.32%       31       31        0      838      838        0         0         0        0     1039
      1      0x7f84e9c2a980     0      71      109    5.79%       11       11        0       29       29        0         0         0        0       80
      2      0x7f84e9c28340     0      49       77    4.21%        8        8        0       14       14        0         0         0        0       63
      3      0x7f84d74e1d00     0     223      308    3.68%        7        7        0      210      210        0         0         0        0       98
      4      0x7f84e4c18600     0      70       99    3.16%        6        6        0       34       34        0         0         0        0       65

#        ----- HITM -----  -- Store Refs --  --------- Data address ---------                      ---------- cycles ----------    Total
#   Num      Rmt      Lcl   L1 Hit  L1 Miss              Offset  Node  PA cnt        Code address  rmt hitm  lcl hitm      load  records

  -------------------------------------------------------------
      1        0       11       29        0      0x7f84e9c2a980
  -------------------------------------------------------------
           0.00%    0.00%   55.17%    0.00%                 0x0     0       1      0x7f84ee04ea86         0         0       189       28
            |
            ---__GI___pthread_mutex_lock (inlined)
               PR_Lock
               |
               |--30.92%--factory_destroy_extension
[...]

PR #4601, opened 2021-02-08
factory NSPR lock to pthread lock

tbordaz:
Issue 4600 - performance modify rate: reduce lock contention on the object extension factory

Bug description:
	object extension factory uses a simple mutex to protect allocation/destroy object.
	This mutex is a NSPR mutex. Using modrate load (entry object), the mutext is the
	second hottest contention. Moving it to pthread mutex moves it down to the fifth
	hotest.
	giving a small throughput benefit (1%)

Fix description:
	Use pthread normal mutex

relates: https://github.com/389ds/389-ds-base/issues/4600

Reviewed by:  ?

Platforms tested: RHEL 8.3 and F31

### DIFF
--- a/ldap/servers/slapd/factory.c
+++ b/ldap/servers/slapd/factory.c
@@ -121,7 +121,7 @@ struct factory_type
 {
     char *name;                                           /* The name of the object that can be extended */
     int extension_count;                                  /* The number of extensions registered for this object */
-    PRLock *extension_lock;                               /* Protect the array of extensions */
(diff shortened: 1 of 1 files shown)


droideck (commented):
Please, fix me if I'm wrong. But should we also change the type of `extension_lock` (make it `pthread_mutex_t`)?

tbordaz:
oh !! absolutely right . Thanks for that catch. I will fix that.

droideck (approved):
LGTM!

Firstyear:
Ack from me too

PR #4602, opened 2021-02-09
Issue 4595 - Paged search lookthroughlimit bug

### LINKED ISSUE #4595: Paged search lookthroughlimit counter doesnt take read ahead into account
**Issue Description**
During a paged search with lookthroughlimit enabled, a lookthroughlimit counter is used to keep track of how many entries are examined. A paged search reads ahead one entry to catch the end of the search so it doesn't show the prompt when there are no more entries. The lookthroughlimit counter doesn't take read ahead into account when tracking how many entries have been examined.

 - Platform: 5.6.6-300.fc32.x86_64
 - Package and version: 389-ds-base-2.0.2.20210204gitf38b124f0

**Steps to Reproduce**
1. Populate DB with 101 entries.
2. Configure lookthroughlimit attribute to a value of 122. (nsslapd-lookthroughlimit: 122)
3. Run a paged search operation with an unindexed search filter. (... "(uid=test0*)" -E pr=10/noprompt))
4. Paged search fails with admin limit error.  
```
# search result
search: 11
result: 11 Administrative limit exceeded

# numResponses: 108
# numEntries: 98
```

**Expected results**
The paged search returns 101 entries with no error.
```
# search result
search: 12
result: 0 Success
control: 1.2.840.113556.1.4.319 false MAUCAQAEAA==
pagedresults: cookie=

# numResponses: 112
# numEntries: 101
```

**Additional context**
The CI tests highlighted this issue, test parameters are used here to reproduce the issue.

jchapma:
Bug Description: During a paged search with lookthroughlimit enabled, a
lookthroughlimit counter is used to keep track of how many entries are
examined. A paged search reads ahead one entry to catch the end of the
search so it doesn't show the prompt when there are no more entries. The
lookthroughlimit counter doesn't take read ahead into account when
tracking how many entries have been examined.

Fix Description: Instead of using a lookthroughlimit counter to track the
entries examined use the search result sr_current index contained in the
pblock.

Fixes: https://github.com/389ds/389-ds-base/issues/4595

Relates: https://github.com/389ds/389-ds-base/issues/4513

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -9,7 +9,7 @@
 
 from subprocess import check_output, PIPE, run
 from lib389 import DirSrv
-from lib389.idm.user import UserAccounts
+from lib389.idm.user import UserAccount, UserAccounts
 import pytest
 from lib389.tasks import *
 from lib389.utils import *
@@ -93,6 +93,24 @@ def rootdse_attr(topology_st, request):
     request.addfinalizer(fin)
 
     return rootdse_attr_name
+
+
+def change_conf_attr(topology_st, suffix, attr_name, attr_value):
+    """Change configuration attribute in the given suffix.
+
+    Returns previous attribute value.
+    """
+
+    entry = DSLdapObject(topology_st.standalone, suffix)
+
+    attr_value_bck = entry.get_attr_val_bytes(attr_name)
+    log.info('Set %s to %s. Previous value - %s. Modified suffix - %s.' % (
+        attr_name, attr_value, attr_value_bck, suffix))
+    if attr_value is None:
+        entry.remove_all(attr_name)
+    else:
+        entry.replace(attr_name, attr_value)
+    return attr_value_bck
 
 
 def test_basic_ops(topology_st, import_example_ldif):
@@ -605,6 +623,73 @@ def test_basic_searches(topology_st, import_example_ldif):
             assert False
 
     log.info('test_basic_searches: PASSED')
+
+
+@pytest.mark.parametrize('limit,resp',
+                         ((('200'), 'PASS'),
+                         (('50'), ldap.ADMINLIMIT_EXCEEDED)))
+def test_basic_search_lookthroughlimit(topology_st, limit, resp, import_example_ldif):
+    """
+    Tests normal search with lookthroughlimit set high and low.
+
+    :id: b5119970-6c9f-41b7-9649-de9233226fec
+
+    :setup: Standalone instance, add example.ldif to the database, search filter (uid=*).
+
+    :steps:
+         1. Import ldif user file.
+         2. Change lookthroughlimit to 200.
+         3. Bind to server as low priv user
+         4. Run search 1 with "high" lookthroughlimit.
+         5. Change lookthroughlimit to 50.
+         6. Run search 2 with "low" lookthroughlimit.
+         8. Delete user from DB.
+         9. Reset lookthroughlimit to original.
+
+    :expectedresults:
+         1. First search should complete with no error.
+         2. Second search should return ldap.ADMINLIMIT_EXCEEDED error.
+    """
+
+    log.info('Running test_basic_search_lookthroughlimit...')
+
+    search_filter = "(uid=*)"
+
+    ltl_orig = change_conf_attr(topology_st, 'cn=config,cn=ldbm database,cn=plugins,cn=config', 'nsslapd-lookthroughlimit', limit)
+
+    try:
+        users = UserAccounts(topology_st.standalone, DEFAULT_SUFFIX, rdn=None)
+        user = users.create_test_user()
+        user.replace('userPassword', PASSWORD)
+    except ldap.LDAPError as e:
+        log.fatal('Failed to create test user: error ' + e.args[0]['desc'])
+        assert False
+
+    try:
+        conn = UserAccount(topology_st.standalone, user.dn).bind(PASSWORD)
+    except ldap.LDAPError as e:
+        log.fatal('Failed to bind test user: error ' + e.args[0]['desc'])
+        assert False
+
+    try:
+        if resp == ldap.ADMINLIMIT_EXCEEDED:
+            with pytest.raises(ldap.ADMINLIMIT_EXCEEDED):
+                searchid = conn.search(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE, search_filter)
+                rtype, rdata = conn.result(searchid)
+        else:
+            searchid = conn.search(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE, search_filter)
+            rtype, rdata = conn.result(searchid)
+            assert(len(rdata) == 151) #151 entries in the imported ldif file using "(uid=*)"
+    except ldap.LDAPError as e:
+        log.fatal('Failed to perform search: error ' + e.args[0]['desc'])
+        assert False
+
+    finally:
+        #Cleanup
+        change_conf_attr(topology_st, 'cn=config,cn=ldbm database,cn=plugins,cn=config', 'nsslapd-lookthroughlimit', ltl_orig)
+        user.delete()
+
+    log.info('test_basic_search_lookthroughlimit: PASSED')
 
 
 @pytest.fixture(scope="module")
--- a/ldap/servers/slapd/back-ldbm/ldbm_search.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_search.c
@@ -1823,6 +1823,7 @@ ldbm_back_prev_search_results(Slapi_PBlock *pb)
             sr->sr_entry = NULL;
         }
         idl_iterator_decrement(&(sr->sr_current));
+        --sr->sr_lookthroughcount;
     }
     return;
 }


droideck (approved):
Looks good to me!
But I'll leave @mreynolds389 and @Firstyear to check too as they've contributed to this code more.

mreynolds389:
I'd like to see you add a CI test that verifies nsslapd-lookthroughlimit is not broken for non-paged searches.  I noticed that we don't have any tests that actually test the lookthroughlimit for normal searches.  So a simple way to write the test would be to create and ldif file will 100 entries, set the lookthroughlimit to 50, bind as a regular user, because the Directory Manager account will bypass the lookthroughlimit, then do a full search of the database.  That search should fail with an ldap.ADMINLIMIT_EXCEEDED exception.

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1538,7 +1538,7 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
             goto bail;
         }
         /* check lookthrough limit */
-        if (llimit != -1 && sr->sr_lookthroughcount >= llimit) {
+        if (llimit != -1 && sr->sr_current >= llimit) {
```
Firstyear: We don't need to test this in the while loop, since sr_current is the number of ids in the partial candidate set (ie it's never going to change within the loop). So we actually should move this up to line ~1514.

Firstyear (changes requested):

Firstyear:
As the comment notes, what do you do in ALLIDS in this case?

droideck:
> I'd like to see you add a CI test that verifies nsslapd-lookthroughlimit is not broken for non-paged searches. I noticed that we don't have any tests that actually test the lookthroughlimit for normal searches. So a simple way to write the test would be to create and ldif file will 100 entries, set the lookthroughlimit to 50, bind as a regular user, because the Directory Manager account will bypass the lookthroughlimit, then do a full search of the database. That search should fail with an ldap.ADMINLIMIT_EXCEEDED exception.

Oops, I thought we have tests for lookthroughlimit... My bad.

ldap/servers/slapd/back-ldbm/ldbm_search.c
jchapma: Hi @Firstyear 

As we iterate through the candidate list on +1573,  sr_current is incremented as we get the id for each entry. So we need to keep the check for lookthroughlimit in the while loop.

jchapma:
Hi @mreynolds389 

I have tested this with a normal search, but its a good idea to add a CI test for a normal search with lookthroughlimit enabled.

ldap/servers/slapd/back-ldbm/ldbm_search.c
Firstyear: Ahh yes, you're correct. I had to re-read the code a few times to be sure. You could actually use:

```
if (llimit != 1 && sr->sr_candidates->b_nids >= llimit) {
}
```

But again, outside the loop. I'd want @progier389 to sanity check though too since this changes the approach a bit.

progier389 (changes requested):
I am aftraid that I have a some concerns about this change:

it changes the behavior of ALLIDs with holes handling:

in the following scenario:
  On standalone instance (i.e: without replication)
  Set Allids and lookthoughlimit to 100 and nsslapd-require-index to 0
  generate a suffix with 150 entries
  then remove 1 entries per tow (for example all entries with even numbers).
Without the fix, I expect to have an unindexed search returning 75 entries
With thee fix we will hit the lookthrough limit (because the last ID is greater than lookthrough limits)

Another remark: as we have a problem with the paged search, I would have expected to see a change in the paged result specific tests rather than a generic change:
We can argue whether this issue described a bug . 
  either is look ahead entry is considered as looked at (in this case there is no bug)
  or not (in which case we just need to decrement sr_lookthroughcount when decrementing the iterator in ldbm_back_prev_search_results. )
I feel that the answer depends whether we just look at the candidate list or we filtered the entry ( i.e subentry test + acl evaluation +  search filter evaluation)

jchapma:
> I am aftraid that I have a some concerns about this change:
> 
> it changes the behavior of ALLIDs with holes handling:
> 
> in the following scenario:
> On standalone instance (i.e: without replication)
> Set Allids and lookthoughlimit to 100 and nsslapd-require-index to 0
> generate a suffix with 150 entries
> then remove 1 entries per tow (for example all entries with even numbers).
> Without the fix, I expect to have an unindexed search returning 75 entries
> With thee fix we will hit the lookthrough limit (because the last ID is greater than lookthrough limits)
> 
> Another remark: as we have a problem with the paged search, I would have expected to see a change in the paged result specific tests rather than a generic change:
> We can argue whether this issue described a bug .
> either is look ahead entry is considered as looked at (in this case there is no bug)
> or not (in which case we just need to decrement sr_lookthroughcount when decrementing the iterator in ldbm_back_prev_search_results. )
> I feel that the answer depends whether we just look at the candidate list or we filtered the entry ( i.e subentry test + acl evaluation + search filter evaluation)





> I am aftraid that I have a some concerns about this change:
> 
> it changes the behavior of ALLIDs with holes handling:
> 
> in the following scenario:
> On standalone instance (i.e: without replication)
> Set Allids and lookthoughlimit to 100 and nsslapd-require-index to 0
> generate a suffix with 150 entries
> then remove 1 entries per tow (for example all entries with even numbers).
> Without the fix, I expect to have an unindexed search returning 75 entries
> With thee fix we will hit the lookthrough limit (because the last ID is greater than lookthrough limits)
> 
> Another remark: as we have a problem with the paged search, I would have expected to see a change in the paged result specific tests rather than a generic change:
> We can argue whether this issue described a bug .
> either is look ahead entry is considered as looked at (in this case there is no bug)
> or not (in which case we just need to decrement sr_lookthroughcount when decrementing the iterator in ldbm_back_prev_search_results. )
> I feel that the answer depends whether we just look at the candidate list or we filtered the entry ( i.e subentry test + acl evaluation + search filter evaluation)

That is a very good point about the impact this change will have on an ALLIDS search, I did not think about that use case. 
So how about i revert these changes, and keep sr_lookthroughcount in sync with read ahead, as Pierre suggested ?

Firstyear:
> 
> That is a very good point about the impact this change will have on an ALLIDS search, I did not think about that use case.
> So how about i revert these changes, and keep sr_lookthroughcount in sync with read ahead, as Pierre suggested ?

I think that my previous comment could work. We can check `if (llimit != 1 && !ALLIDS(sr->sr_candidates) && sr->sr_candidates->b_nids >= llimit)`  and for ALLIDS I think we should use @progier389's suggestion? Unless there is a way to count the content of ALLIDS without actually having to apply the filter tests.

jchapma:
> > That is a very good point about the impact this change will have on an ALLIDS search, I did not think about that use case.
> > So how about i revert these changes, and keep sr_lookthroughcount in sync with read ahead, as Pierre suggested ?
> 
> I think that my previous comment could work. We can check `if (llimit != 1 && !ALLIDS(sr->sr_candidates) && sr->sr_candidates->b_nids >= llimit)` and for ALLIDS I think we should use @progier389's suggestion? Unless there is a way to count the content of ALLIDS without actually having to apply the filter tests.

Hi @Firstyear 
I do like the idea of moving this check outside the loop, I'm currently playing with this idea. Thanks.

progier389:
I am not sure we should move the test out of the loop:
   That would means that lookthrough limit will now be tested before the size limit and you will get a different ldap operation result code for some operations ... 
   Furthermore  the time spend by a simple integer test is negligible even if the search returns thousands of entries so I am not sure it is worth to change that ...

Firstyear:
> I am not sure we should move the test out of the loop:
> That would means that lookthrough limit will now be tested before the size limit and you will get a different ldap operation result code for some operations ...
> Furthermore the time spend by a simple integer test is negligible even if the search returns thousands of entries so I am not sure it is worth to change that ...

It's not the time to do the int ops I'm worried about - It's that lookthroughlimit is to "prevent" the server doing excess work applying the tests. So doing this "sooner" means that we would reject the op sooner. However that is a behavioural change, as currently  Ithink we would sent entries up to the look through limit before stopping.

jchapma:
Hi Folks, I propose creating a seperate issue for investigating search behaviour and how it can be improved.

A CI test highlighted the issue of lookthroughcount not being synced with paged search read ahead and this PR is focused on that. I have included a CI test that performs a normal search with lookthrouglimit set, just to make sure we dont break anything with this one line change.

Firstyear:
> Hi Folks, I propose creating a seperate issue for investigating search behaviour and how it can be improved.

Yep, I fully agree with this :) 

I think the change seems reasonable to me now, I'll let @progier389 double check in case I have missed something :)

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -607,6 +625,85 @@ def test_basic_searches(topology_st, import_example_ldif):
...
+
+    log.info('Running test_basic_search_lookthroughlimit...')
+
+    search_filter = r'(uid=*)'
+
+    ltl_orig = change_conf_attr(topology_st, 'cn=config,cn=ldbm database,cn=plugins,cn=config', 'nsslapd-lookthroughlimit', limit)
+
+    try:
+        topology_st.standalone.add_s(Entry((USER1_DN,
+                                            {'objectclass': "top extensibleObject".split(),
```
mreynolds389: Sorry I didn't see this earlier.  We are not using any of the "_s" functions from python-ldap in our test cases.  Everything should be a DSLdapObject.  So no "add_s()", "search_s", "modify_s", etc.  You should be able to use UserAccount/UserAccounts for most of this.   If you need help, Simon or myself can help you converting this over to DSLdapObject.

jchapma: Done

progier389 (approved):
LGTM

jchapma:
4cf9be7edd6ccb274d3316b367536bb2fde958cd    389-ds-base-1.4.3
1ea19483e1501ac3bb20d28a6e980d57b8e71d8e 389-ds-base-1.4.4

PR #4613, opened 2021-02-10
issue 4612 - Fix pytest fourwaymmr_test for non root user

### LINKED ISSUE #4612: Fix pytest fourwaymmr_test for non root user.
If pytest is run by a non root user,
fourwaymmr_test fails because the dse.ldif chown fails with PermissionDenied exception 

Problem reproduction:
On fedora using master branch:
Generate a local build (according to https://www.port389.org/docs/389ds/development/building.html#building-into-a-prefix-location-non-standardnot-root instructions)
 Run pytest  .../fourwaymmr_test.py
 
 Test fails because of PermissionDenied exception while perming a chown dirsrv on dse.ldif 
 
 The exception should be ignored (because the chown is only needed when running as root)

progier389:
Bug Description: Fix pytest fourwaymmr_test for non root user

Fix Description: Ignore PermissionDenied exception while doing the chown dirsrv on the dse.ldif file

fixes: #4612

Review by: @droideck

### DIFF
--- a/dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py
+++ b/dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py
@@ -14,6 +14,7 @@ from lib389.topologies import topology_m4 as topo_m4
 from lib389.replica import *
 from lib389.idm.user import UserAccounts
 from lib389.agreement import *
+from contextlib import suppress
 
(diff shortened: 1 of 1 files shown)


droideck (approved):
Nice move with the `suppress`!
LGTM!

mreynolds389:
Also LGTM

PR #4616, opened 2021-02-12
Issue 4324 - Some architectures the cache line size file does not exist

mreynolds389:
Bug Description:  

When optimizing our mutexes we check for a system called coherency_line_size that contains the size value, but if the file did not exist the server would crash in PR_Read (NULL pointer for fd).

Fix Description:  Check PR_Open() was successfully before calling PR_Read().

Relates: https://github.com/389ds/389-ds-base/issues/4324

Reviewed by: tbordaz(Thanks!)


tbordaz:
The patch looks good to me. Thanks for fixing this. ACK

PR #4621, opened 2021-02-15
Issue 4620 - Add oneshot mode to `dscontainer`

### LINKED ISSUE #4620: Add oneshot mode to `dscontainer`
I'd like to have a DS container created and configured at *build* time. This would allow me to have a throwaway container that I can spin up and destroy when it's no longer needed, for example to be used in testing with other software that needs an LDAP server.

I would like to have a "oneshot" mode. When it is used, `dscontainer` will run once, execute an optional script that can be used to set additional configuration parameters, and shutdown the server. 

Init container and operator are alternatives in kubernetes environment. But it creates an additional complexity and cannot be used outside of this environment.

**Additional context**
See discussion in this PR: https://github.com/389ds/389-ds-base/pull/4564

vashirov:
Description:

* Add oneshot mode to `dscontainer` via -o/--oneshot option
When this parameter is used without value, `dscontainer` will shutdown right after the instance is initialized.
If an optional parameter is passed, pointing to a valid script, this
script will be executed before the shutdown.

* Update help message and comments
* Fix SyntaxWarning: "is" with a literal. Did you mean "=="?

Fixes: https://github.com/389ds/389-ds-base/issues/4620

Reviewed by: ???

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -169,8 +169,11 @@ def begin_magic():
     # EXPOSE 3389 3636
     # RUN mkdir -p /data/config && \
     #     mkdir -p /data/ssca && \
+    #     mkdir -p /data/run && \
+    #     mkdir -p /run/dirsrv && \
     #     ln -s /data/config /etc/dirsrv/slapd-localhost && \
     #     ln -s /data/ssca /etc/dirsrv/ssca && \
+    #     ln -s /data/run /run/dirsrv
     # # Temporal volumes for each instance
     # VOLUME /data
     #
@@ -352,6 +355,19 @@ binddn = cn=Directory Manager
 
     log.info("389-ds-container started.")
 
+    # If oneshot parameter is passed, run an optional script and
+    # shutdown the server.
+    if args.oneshot != False:
+        if args.oneshot != None:
+            if os.path.exists(args.oneshot):
+                log.info("Running oneshot script...")
+                oneshot_p = subprocess.Popen(args.oneshot, stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE, env=os.environ.copy())
+                oneshot_p.wait()
+            else:
+                log.info("Oneshot script is not found!")
+        kill_ds()
+
     # Now block until we get shutdown! If we are signaled to exit, this
     # will trigger the atexit handler from above.
     try:
@@ -401,8 +417,11 @@ have certain settings to work correctly.
 \tEXPOSE 3389 3636
 \tRUN mkdir -p /data/config && \\
 \t    mkdir -p /data/ssca && \\
+\t    mkdir -p /data/run && \\
+\t    mkdir -p /run/dirsrv && \\
 \t    ln -s /data/config /etc/dirsrv/slapd-localhost && \\
 \t    ln -s /data/ssca /etc/dirsrv/ssca && \\
+\t    ln -s /data/run /run/dirsrv
 \tVOLUME /data
 
 This is an example of the minimal required configuration. The 389
@@ -420,6 +439,9 @@ container host.
     parser.add_argument('-r', '--runit',
                         help="Actually run the instance! You understand what that means ...",
                         action='store_true', default=False, dest='runit')
+    parser.add_argument('-o', '--oneshot',
+                        help="Run the instance once and shutdown. Optionally run a config script.",
+                        action='store', default=False, nargs='?', type=str, dest='oneshot')
     parser.add_argument('-H', '--healthcheck',
                         help="Start a healthcheck inside of the container for an instance. You should understand what this means ...",
                         action='store_true', default=False, dest='healthcheck')
@@ -431,9 +453,7 @@ container host.
     if args.runit:
         begin_magic()
     elif args.healthcheck:
-        if begin_healthcheck(None) is (False, True):
+        if begin_healthcheck(None) == (False, True):
             sys.exit(0)
         else:
             sys.exit(1)
-
-
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -361,15 +361,19 @@ binddn = cn=Directory Manager
         if args.oneshot != None:
             if os.path.exists(args.oneshot):
                 log.info("Running oneshot script...")
-                oneshot_p = subprocess.Popen(args.oneshot, stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE, env=os.environ.copy(), universal_newlines=True)
+                oneshot_p = subprocess.Popen(
+                    args.oneshot,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    env=os.environ.copy(),
+                    universal_newlines=True)
                 rc = oneshot_p.wait()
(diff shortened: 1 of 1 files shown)


src/lib389/cli/dscontainer
```diff
@@ -352,6 +355,19 @@ binddn = cn=Directory Manager
...
 
+    # If oneshot parameter is passed, run an optional script and
+    # shutdown the server.
+    if args.oneshot != False:
+        if args.oneshot != None:
+            if os.path.exists(args.oneshot):
+                log.info("Running oneshot script...")
+                oneshot_p = subprocess.Popen(args.oneshot, stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE, env=os.environ.copy())
+                oneshot_p.wait()
```
Firstyear: Is it worth displaying the return code from the oneshot script here?

src/lib389/cli/dscontainer
```diff
@@ -352,6 +355,19 @@ binddn = cn=Directory Manager
...
+            else:
+                log.info("Oneshot script is not found!")
+        kill_ds()
```
Firstyear, line 377: Why immediately shutdown rather than leaving the instance running?

Firstyear (changes requested):
Just some minor-ish questions :)

src/lib389/cli/dscontainer
vashirov: Yes, good idea, I will add it.

src/lib389/cli/dscontainer
vashirov: Because I want to run `dscontainer` in oneshot mode at the container *build* time, rather than as entry point. So that the resulting container will have a preconfigured instance. 
Here's an example: https://github.com/vashirov/ds-container/blob/dscontainer_oneshot/fedora-minimal/Containerfile.f33#L21

src/lib389/cli/dscontainer
vashirov: In addition to exit code I've added script output for debugging purposes, please review.

src/lib389/cli/dscontainer
```diff
@@ -352,6 +355,23 @@ binddn = cn=Directory Manager
...
     log.info("389-ds-container started.")
 
+    # If oneshot parameter is passed, run an optional script and
+    # shutdown the server.
+    if args.oneshot != False:
+        if args.oneshot != None:
+            if os.path.exists(args.oneshot):
+                log.info("Running oneshot script...")
+                oneshot_p = subprocess.Popen(args.oneshot, stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE, env=os.environ.copy(), universal_newlines=True)
```
droideck: Indentation is a bit odd here but it's really a minor nitpick.

droideck (approved):
The rest LGTM!

src/lib389/cli/dscontainer
Firstyear: The problem is that, any content you change in the volume (or /data which becomes the volume) will not always be overlaid into the volume that is consumed by the user. So you can't configure at build time, you can only rely on configuration to be applied at first run. That's why it would be better to have a stateful expression because this could be run multiple times and you can't guarantee that the changes will be present after a build. 

So while it may work for you to have this run at build time, it won't work for *other* people. Which is a bit why I'm worried about making a feature like this which could have a lot of sharp edges ...

src/lib389/cli/dscontainer
```diff
@@ -352,6 +355,23 @@ binddn = cn=Directory Manager
...
+                rc = oneshot_p.wait()
+                log.debug(f"Oneshot script output:\n{oneshot_p.communicate()[0]}")
+                if rc != 0:
+                    log.error(f"Oneshot script failed with exit code {rc}")
+                    sys.exit(rc)
+            else:
+                log.info("Oneshot script is not found!")
```
Firstyear: Should be an error, if it was an argument but not found, this is a failure.

src/lib389/cli/dscontainer
vashirov: > The problem is that, any content you change in the volume (or /data which becomes the volume) will not always be overlaid into the volume that is consumed by the user. 

Can you please elaborate and provide an example? Is this still a problem in a throwaway-ldap-container-for-testing use case where there are no external volumes?

> So you can't configure at build time, you can only rely on configuration to be applied at first run. 

Are you talking about `dscontainer` in particular or about containers in general?

> That's why it would be better to have a stateful expression because this could be run multiple times and you can't guarantee that the changes will be present after a build.

Could you please help me understand what change do you want me to do? Or provide a concrete practical example why the current approach can go wrong?

> So while it may work for you to have this run at build time, it won't work for _other_ people. Which is a bit why I'm worried about making a feature like this which could have a lot of sharp edges ...

From what I saw, other people are doing the same thing with our DS containers, but with extra steps (`dscreate` and fake `systemctl` scripts). So why not make their life easier?

vashirov: > Because I want to run `dscontainer` in oneshot mode at the container _build_ time, rather than as entry point. So that the resulting container will have a preconfigured instance.
> Here's an example: https://github.com/vashirov/ds-container/blob/dscontainer_oneshot/fedora-minimal/Containerfile.f33#L21

To clarify: I run `dscontainer` in oneshot mode during the build, and then it is set as a CMD in a run mode.

Firstyear: > > The problem is that, any content you change in the volume (or /data which becomes the volume) will not always be overlaid into the volume that is consumed by the user.
> 
> Can you please elaborate and provide an example? Is this still a problem in a throwaway-ldap-container-for-testing use case where there are no external volumes?

It's not a problem for testing or throw-away. It is a problem for release builds though, or people building and layering things together. 

> 
> > So you can't configure at build time, you can only rely on configuration to be applied at first run.
> 
> Are you talking about `dscontainer` in particular or about containers in general?

Containers in general. 

> 
> > That's why it would be better to have a stateful expression because this could be run multiple times and you can't guarantee that the changes will be present after a build.
> 
> Could you please help me understand what change do you want me to do? Or provide a concrete practical example why the current approach can go wrong?

My concern is that while the one shot is a quick solution to your problem that you want to solve, it will be used by other people in the wild. Containers should be driven by stateful configuration management where you express a desired state, rather than "please run these things to get to this state". I think that the oneshot mode has a lot of ways it "could go wrong" in production to users of the containers 

> 
> > So while it may work for you to have this run at build time, it won't work for _other_ people. Which is a bit why I'm worried about making a feature like this which could have a lot of sharp edges ...
> 
> From what I saw, other people are doing the same thing with our DS containers, but with extra steps (`dscreate` and fake `systemctl` scripts). So why not make their life easier?

I would assume they are doing this because we don't have a good way to provide stateful configuration into the container yet, so that's why I think that working on a way to do this that solves both cases would be really good - but I accept it would certainly be "more work" than what you have proposed here.

vashirov: Okay, I see your point. I agree that stateful configuration would be a way to go. And yes, it would be a non-trivial amount of work to get it done. I will open a new ticket to track it. I will also open a new ticket to fix a syntax warning and usage information that were fixed in this PR.

Thanks!

Issue #4596, opened 2021-02-05
Build with clang/lld fails with undefined reference error when LTO is enabled


vashirov:
**Issue Description**
In Fedora 33 LTO is enabled by default: https://fedoraproject.org/wiki/LTOByDefault
Build with clang/lld fails with undefined reference error.

**Steps to Reproduce**
On Fedora 33 with 
```
lld-11.0.0-1.fc33.x86_64
clang-11.0.0-2.fc33.x86_64
llvm-libs-11.0.0-1.fc33.x86_64
```

```
$ sed -i -e 's/COCKPIT_ON = 1/COCKPIT_ON = 0/g' -e 's/CLANG_ON = 0/CLANG_ON = 1/g' rpm.mk
$ make -f rpm.mk rpms
```

```
/bin/sh ./libtool  --tag=CC   --mode=link clang -g -O2 -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1     -O2 -flto -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS --config /usr/lib/rpm/redhat/redhat-hardened-clang.cfg -fstack-protector-strong   -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -avoid-version -Wl,-z,relro -Wl,--as-needed  -Wl,-z,now  -flto -o libattr-unique-plugin.la -rpath /usr/lib64/dirsrv/plugins ldap/servers/plugins/uiduniq/libattr_unique_plugin_la-7bit.lo ldap/servers/plugins/uiduniq/libattr_unique_plugin_la-uid.lo ldap/servers/plugins/uiduniq/libattr_unique_plugin_la-utils.lo libslapd.la -lplds4 -lplc4 -lnspr4 -lpthread -ldl  
ld.lld: error: ./.libs/libslapd.so: undefined reference to __rust_probestack [--no-allow-shlib-undefined]
ld.lld: error: ./.libs/libslapd.so: undefined reference to __muloti4 [--no-allow-shlib-undefined]
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```

If I disable LTO in the spec file, build succeeds:
```
%global _lto_cflags %nil
```

Firstyear:
[...]

PR #4625, opened 2021-02-16

vashirov:
Description:

* Add ASAN variant for tests
* Add HTML report for pytest, that contains ASAN/LSAN reports

Reviewed by:

### DIFF
--- a/.github/scripts/generate_matrix.py
+++ b/.github/scripts/generate_matrix.py
@@ -1,6 +1,7 @@
 import os
 import glob
 import json
+from itertools import product
 
 suites = next(os.walk('dirsrvtests/tests/suites/'))[1]
 
@@ -13,7 +14,15 @@ repl_tests = glob.glob('dirsrvtests/tests/suites/replication/*_test.py')
 suites += [repl_test.replace('dirsrvtests/tests/suites/', '') for repl_test in repl_tests]
 suites.sort()
 
-suites_list = [{ "suite": suite} for suite in suites]
-matrix = {"include": suites_list}
+variants = ['gcc', 'gcc-asan']
+# Disable clang matrix for now, otherwise we generate more than 256 jobs...
+# variants += ['clang', 'clang-asan']
+
+
+jobs = []
+for suite, variant in product(suites, variants):
+    jobs.append({"suite": suite, "variant": variant})
+
+matrix = {"include": jobs}
 
 print(json.dumps(matrix))
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,16 +4,69 @@ on:
   push:
   pull_request:
   schedule:
-    - cron:  '0 0 * * *'
+    - cron: "0 0 * * *"
 
 jobs:
-  build:
-    name: Build
+  build_rpms:
+    name: Build RPMs
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        #variant: [gcc, gcc-asan, clang, clang-asan]
+        variant: [gcc, gcc-asan]
+
+        include:
+          - variant: gcc
+            asan: 0
+            clang: 0
+          - variant: gcc-asan
+            asan: 1
+            clang: 0
+          # - variant: clang
+          #   asan: 0
+          #   clang: 1
+          # - variant: clang-asan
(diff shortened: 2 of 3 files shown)


vashirov:
It doesn't show all checks because I split the workflows. And for some (security?) reasons GitHub doesn't execute additional workflows that are not yet merged. Only modified workflows. I will post a link with a full run from my fork.

vashirov:
https://github.com/vashirov/389-ds-base/actions/runs/571115253

vashirov:
Hmm, switching to F32 didn't help, test artifacts are still not useful. I will set this to WIP until I fix ASAN reports.

vashirov:
Ok, now there are memory leaks reports. But I have another problem now and need your opinion.
With `fast_unwind_on_malloc=1` the traces are pretty much useless. They just indicate that there is a leak, but not print a proper stack trace.
With `fast_unwind_on_malloc=0` tests run 3-4 times longer and sometimes hang. So for PR checks it's not ideal. 

I propose to run with `fast_unwind_on_malloc=1` on PRs and have a nightly run (which is not time sensitive) with `0`. 
WDYT?

vashirov:
Forgot the link: https://github.com/vashirov/389-ds-base/actions/runs/574633583
Check artifacts at the bottom. Archives with `-asan` have html page which has links to ASAN/LSAN reports for exact test that generated them.

Firstyear:
I think we can have fast_unwind_on_malloc=1, because knowing that *any* test has failed indicates to us to investigate further or to look at the nightly which does have the ability to generate the exact report. So I'm okay with this :)

droideck (approved):
Looks good to me!

.github/scripts/generate_matrix.py
```diff
@@ -13,7 +14,15 @@
 suites += [repl_test.replace('dirsrvtests/tests/suites/', '') for repl_test in repl_tests]
 suites.sort()
 
-suites_list = [{ "suite": suite} for suite in suites]
-matrix = {"include": suites_list}
+variants = ['gcc', 'gcc-asan']
```
Firstyear, line 17: gcc-asan doesn't work with rust, so we can't use this. We need clang-asan to be consistent, especially so to match the correct llvm versions.

Firstyear (changes requested):

.github/scripts/generate_matrix.py
vashirov: Ok, I will replace `gcc-asan` with `clang-asan`.

vashirov: @Firstyear, clang with `-fsanitize=address` fails to compile: https://github.com/vashirov/389-ds-base/runs/3554323865#step:6:3541
Any ideas?

vashirov: I found that I filed an issue for this some time ago https://github.com/389ds/389-ds-base/issues/4596
I feel that `clang-asan` would require more efforts than I expected, so I'd like to to keep `gcc-asan` for now and add `clang-asan` in a follow up PR.

Firstyear: Didn't we change how lto was managed here to make it work? I'm building and using clang-asan just fine? 

The issue with gcc-asan is that I've noticed more and more that gcc asan and rust's llvm backed asan aren't compatible and can not be mixed, which has led to failures to build. That's why I've been using clang now so that they both have the same llvm backend and asan versions.

vashirov: No, we didn't :( I couldn't make LTO builds with clang and then I switched to other things, forgetting about the issue.
I need some help from you regarding Rust + LTO, I will put the details in the related issue.

I haven't encountered any failures myself. Ideally I'd like to have the full matrix (`gcc`, `gcc-asan`, `clang`, `clang-asan`), so that we can catch these build issues too. But I'm hitting GH actions limit of 256 jobs in the matrix. I have some ideas how to circumvent this with composable actions, will try it next week.

Firstyear: Okay, I'll look at the LTO issue later, but I think we can't do gcc-asan + rust at all. I've had so many conflicts between the two ASAN apis as they exist, and a bunch of false positives at this point, that I don't think it should be considered reliable.

PR #4631, opened 2021-02-17
Issue 4169 - UI - port charts to PF4

mreynolds389:
Description:  

Ported the charts under the monitor tab to use PF4 sparkline charts and provide realtime stats on the caches.

Relates: https://github.com/389ds/389-ds-base/issues/4169

### DIFF
--- a/src/cockpit/389-console/src/lib/tools.jsx
+++ b/src/cockpit/389-console/src/lib/tools.jsx
@@ -75,7 +75,7 @@ export function get_date_string(timestamp) {
         parseInt(minute),
         parseInt(sec)
     );
-    return date.toLocaleString();
+    return date.toLocaleString('en-ZA');
 }
 
 // Take two directory server tiemstamps and get the elapsed time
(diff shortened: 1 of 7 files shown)


src/cockpit/389-console/src/lib/tools.jsx
```diff
@@ -75,7 +75,7 @@ export function get_date_string(timestamp) {
         parseInt(minute),
         parseInt(sec)
     );
-    return date.toLocaleString();
+    return date.toLocaleString('en-ZA');
```
droideck, line 78: Hm, is there any particular reason for 'en-ZA' locale? (ZA part confuses me)

droideck (approved):
Besides this minor question, looks really good to me!

Charts are awesome!

src/cockpit/389-console/src/lib/tools.jsx
mreynolds389: Good question.  So in the monitor the server start date uses MM/DD/YYYY, but I want it to use a more universal format.  Locale "en-ZA" uses YYYY/MM/DD

I'm open to other suggestions as well.

droideck: Agree, it's a nice format. You have my full ack :)

PR #4636, opened 2021-02-18
Issue 4169 - UI - Migrate alerts to PF4

mreynolds389:
Description:  

Migrate the toast notifications to PF4 Alerts.

Also fixed a refresh problem on the Tuning page.

### DIFF
--- a/src/cockpit/389-console/src/css/ds.css
+++ b/src/cockpit/389-console/src/css/ds.css
@@ -703,3 +703,13 @@ h4 {
   margin-top: 50px;
   height: 135px;
 }
+
+pre {
+    white-space: pre-wrap;       /* Since CSS 2.1 */
+    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+    word-wrap: break-word;       /* Internet Explorer 5.5+ */
+}
+
+.pf-c-alert-group {
+    z-index: 20000 !important;
+}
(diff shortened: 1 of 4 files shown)


droideck (commented):
I wonder if it'd be possible to make it properly 'hover'...

Because now the alerts are always displayed on the same level. Because of this, sometimes you can't read alert (check my back up screenshot - the 'successful alert' is displayed in the gray area behind), and also it's impossible to read 'instance menu' if there are some alerts (but I think this part was like that even before the PR)

droideck:
Screenshot from 2021-02-22 15-11-28
Screenshot from 2021-02-22 15-12-28

mreynolds389:
> I wonder if it'd be possible to make it properly 'hover'...

Fixed

> also it's impossible to read 'instance menu' if there are some alerts (but I think this part was like that even before the PR)

Yeah its always been that way, not sure what you would prefer instead.  The alerts should timeout, or can be closed.

droideck:
> > I wonder if it'd be possible to make it properly 'hover'...
> 
> Fixed
> 
> > also it's impossible to read 'instance menu' if there are some alerts (but I think this part was like that even before the PR)
> 
> Yeah its always been that way, not sure what you would prefer instead. The alerts should timeout, or can be closed.

I imagine that ideally when this sequence of events happens:
1. Alert is displayed
2. I click on the 'Instance' menu button (where Restart, Stop, etc.)

Then the 'Instance' menu is displayed above the alert as it has happened later.

droideck (approved):
You have my ack as the issue I mentioned is really minor and it probably doesn't worth the fuss.

PR #4645, opened 2021-02-19
Issue 4643 - Add a tool that generates Rust dependencies for a specfile

### LINKED ISSUE #4643: Add a tool that helps generate Rust dependencies in specfile
The Fedora builds of 389-DS uses the vendored crates to build the official packages for Rawhide. Vendoring and bundling dependencies is in violation of Fedora policies. As an upstream project we are free to ship vendored code. But as a downstream Fedora project we must not use the vendored code.

Add a tool that will help to generate 'Provides: bundled(crate(foo)) = version' for Cargo.lock file content.
Also, we need to comulate the Licenses for each of the crate.

Add a macro to 389-ds-base RPM profile which we can use in our specfiles.

**Bugzilla**
 https://bugzilla.redhat.com/show_bug.cgi?id=1920959

droideck:
Issue 4643 - Add a tool that generates Rust dependencies for a specfile

Description: The Fedora builds of 389-DS uses the vendored crates
to build the official packages for Rawhide. Vendoring and bundling
dependencies is in violation of Fedora policies. As an upstream project
we are free to ship vendored code. But as a downstream Fedora project
we must not use the vendored code.

Add a tool that will help to generate 'Provides: bundled(crate(foo)) = version'
for Cargo.lock file content.
Replace License field which should contain all of the package licenses
we bundle in the specfile.

Fixes: #4643

Reviewed by: @Firstyear, @decathorpe, @mreynolds389 (Thanks!)

### DIFF
--- a/rpm.mk
+++ b/rpm.mk
@@ -44,6 +44,9 @@ download-cargo-dependencies:
 	cargo fetch --manifest-path=./src/Cargo.toml
 	tar -czf vendor.tar.gz vendor
 
+bundle-rust-on-fedora:
+	python3 rpm/bundle-rust-downstream.py ./src/Cargo.lock $(FEDORA_SPECFILE) ./vendor
+
 install-node-modules:
 ifeq ($(COCKPIT_ON), 1)
 	cd src/cockpit/389-console; make -f node_modules.mk install
--- a/rpm/bundle-rust-downstream.py
+++ b/rpm/bundle-rust-downstream.py
@@ -1,0 +1,199 @@
+#!/usr/bin/python3
+
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+# PYTHON_ARGCOMPLETE_OK
+
+import os
+import sys
+import time
+import signal
+import argparse
+import argcomplete
+import shutil
+import toml
+from lib389.cli_base import setup_script_logger
+from rust2rpm import licensing
+
+SPECFILE_COMMENT_LINE = 'Bundled cargo crates list'
+START_LINE = f"##### {SPECFILE_COMMENT_LINE} - START #####\n"
+END_LINE = f"##### {SPECFILE_COMMENT_LINE} - END #####\n"
+
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description="""Add 'Provides: bundled(crate(foo)) = version' to a Fedora based specfile. 
+Additionally, add a helper comment with a comulated License metainfo which is based on Cargo.lock file content.""")
+
+parser.add_argument('-v', '--verbose',
+                    help="Display verbose operation tracing during command execution",
+                    action='store_true', default=False)
+
+parser.add_argument('cargo_lock_file',
+                    help="The path to Cargo.lock file.")
+parser.add_argument('spec_file',
+                    help="The path to spec file that will be modified.")
+parser.add_argument('vendor_dir',
+                    help="The path to the vendor directory file that will be modified.")
+parser.add_argument('--backup-specfile',
+                    help="Make a backup of the downstream specfile.",
+                    action='store_true', default=False)
+
+
+# handle a control-c gracefully
+def signal_handler(signal, frame):
+    print('\n\nExiting...')
+    sys.exit(0)
+
+
+def get_license_list(vendor_dir):
+    license_list = list()
+    for root, _, files in os.walk(vendor_dir):
+        for file in files:
+            name = os.path.join(root, file)
+            if os.path.isfile(name) and "Cargo.toml" in name:
+                with open(name, "r") as file:
+                    contents = file.read()
+                data = toml.loads(contents)
+                license, warning = licensing.translate_license_fedora(data["package"]["license"])
+
+                # Normalise
+                license = license.replace("/", " or ").replace(" / ", " or ")
+                license = license.replace("Apache-2.0", "ASL 2.0")
+                license = license.replace("WITH LLVM-exception", "with exceptions")
+                if "or" in license or "and" in license:
+                    license = f"({license})"
+                if license == "(MIT or ASL 2.0)":
+                    license = "(ASL 2.0 or MIT)"
+
+                if license not in license_list:
+                    if warning is not None:
+                        # Ignore known warnings
+                        if not warning.endswith("LLVM-exception!") and \
+                        not warning.endswith("MIT/Apache-2.0!"):
+                            print(f"{license}: {warning}")
+                    license_list.append(license)
+    return " and ".join(license_list)
+
+
+def backup_specfile(spec_file):
+    time_now = time.strftime("%Y%m%d_%H%M%S")
+    log.info(f"Backing up file {spec_file} to {spec_file}.{time_now}")
+    shutil.copy2(spec_file, f"{spec_file}.{time_now}")
+
+
+def replace_license(spec_file, license_string):
+    result = []
+    with open(spec_file, "r") as file:
+        contents = file.readlines()
+        for line in contents:
+            if line.startswith("License: "):
+                result.append("# IMPORTANT - Check if it looks right. Additionally, "
+                              "compare with the original line. Then, remove this comment and # FIX ME - part.\n")
+                result.append(f"# FIX ME - License:          GPLv3+ and {license_string}\n")
+            else:
+                result.append(line)
+    with open(spec_file, "w") as file:
+        file.writelines(result)
+    log.info(f"Licenses are successfully updated - {spec_file}")
+
+
+def clean_specfile(spec_file):
+    result = []
+    remove_lines = False
+    cleaned = False
+    with open(spec_file, "r") as file:
+        contents = file.readlines()
+
+    log.info(f"Remove '{SPECFILE_COMMENT_LINE}' content from {spec_file}")
+    for line in contents:
+        if line == START_LINE:
+            remove_lines = True
+            log.debug(f"Remove '{START_LINE}' from {spec_file}")
+        elif line == END_LINE:
+            remove_lines = False
+            cleaned = True
+            log.debug(f"Remove '{END_LINE}' from {spec_file}")
+        elif not remove_lines:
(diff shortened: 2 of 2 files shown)


droideck:
I've discussed the issue with Fabio and he said he's okay with bundling approach.

We still need to take care of Licenses.
I'll copy-paste Fabio's answer here, because, @Firstyear, you said you have your own way for metadata processing.
Please, check if you're okay with this idea and I'll write down one or another apporach.

```
However, something that's going to be manual work regardless, will be
the determination of the effective License. First, you have the same
problem as with listing vendored dependencies (not all of them end up
in the built binary), and second, cargo metadata only contains license
information in SPDX format, which can't be used in Fedora (yet). The
"easiest" solution for this would probably be to grep all Cargo.toml
files for their "license" field and do a | sort | uniq pass over them.
This would reduce the "manual" steps to 1) eliminating duplicates due
to formatting differences (e.g. / vs. OR, or reversed order for OR
arguments), and 2) then convert the remaining SPDX tags into the
corresponding Fedora License tags, and 3) simplify the result.

The rust2rpm module we use for Rust packaging in Fedora has a helper
module for converting SPDX to Fedora License tags as part of the .spec
generation process for a *single* package:
https://pagure.io/fedora-rust/rust2rpm/blob/master/f/rust2rpm/licensing.py
It will not help you with 1) gathering all specified Licenses or 3)
simplifying the list at the end, but at least this should help with 2)
(the SPDX->Fedora conversion process for *single* License metadata
entries).
```

Also, I was thinking if we can use this script to generate the list directly in specfile by writing our own macro... It may save us some time but we still need to do some part of manual work for licenses.

Also, Fabio, if you'll have time, please, check if you're okay with the tool I developed here.

decathorpe:
That looks way too complicated, but if it works 🤷🏻‍♂️ I thought you'd just run something like my script and copy-paste the output into the .spec file and replace the right things.

droideck:
> That looks way too complicated, but if it works 🤷🏻‍♂️ I thought you'd just run something like my script and copy-paste the output into the .spec file and replace the right things.

I'd rather automate as much as possible because we do a lot of releases...
And when we create a tarball we generate a new Cargo.lock file so it may have new versions which, according to guidelines, must be up to date with each 389-ds-base version.

rpm.mk
```diff
@@ -44,6 +44,9 @@ download-cargo-dependencies:
 	cargo fetch --manifest-path=./src/Cargo.toml
 	tar -czf vendor.tar.gz vendor
 
+bundle-rust-on-fedora:
```
Firstyear, line 47: did you also want this to call the vendor function too so that the Cargo.lock is up to date?

rpm.mk
```diff
@@ -44,6 +44,9 @@ download-cargo-dependencies:
...
+	rpm/bundle-rust-downstream ./src/Cargo.lock $(FEDORA_SPECFILE) --backup-specfile
```
Firstyear: probably shouldn't rely on the shebang here if this is not executeable, better to run as python3 rpm/bundle....

rpm/bundle-rust-downstream
```diff
@@ -0,0 +1,142 @@
...
+import sys
+import signal
+import argparse
+import argcomplete
+import shutil
+import toml
+from lib389.cli_base import setup_script_logger
+
+SPECFILE_COMMENT_LINE = 'Bundled cargo crates list'
+START_LINE = f"# {SPECFILE_COMMENT_LINE} - START\n"
```
Firstyear: minor, but I'd make this "bigger". Like "##### {SPEC..} - START #####" just so the section is really clearly highlighted.

rpm/bundle-rust-downstream
```diff
@@ -0,0 +1,142 @@
...
+
+# handle a control-c gracefully
+def signal_handler(signal, frame):
+    print('\n\nExiting...')
+    sys.exit(0)
+
+
+def backup_specfile(spec_file):
+    log.info(f"Backing up file {spec_file} to {spec_file}.backup")
+    shutil.copy2(spec_file, f"{spec_file}.backup")
```
Firstyear: should we date these files rather than overwriting? If you did this operation twice in a row, you would delete the original backup ...

rpm/bundle-rust-downstream
```diff
@@ -0,0 +1,142 @@
...
+
+
+def write_provides_bundled_crate(cargo_lock_file, spec_file, cleaned):
+    # Generate 'Provides' out of cargo_lock_file
+    with open(cargo_lock_file, "r") as file:
+        contents = file.read()
+    data = toml.loads(contents)
+    provides_lines = []
+    for package in data["package"]:
+        provides_lines.append(f"Provides:  bundled(crate({package['name']})) = {package['version']}\n")
```
Firstyear: subtle issue, but some crate dependencies have multiple hyphens in the name which can cause spec file errors as the parser thinks the version is incorrect and has multiple hyphens. My tool has a work around here:

https://github.com/Firstyear/cargo-lock2rpmprovides/blob/main/src/main.rs#L176

Firstyear (changes requested):
Yeah, looks okay, I added some comments. Something I would ask is how you plan to manage the license string. Rust cargo.toml can contain a SPDX license string, which I parse here: https://github.com/Firstyear/cargo-lock2rpmprovides/blob/main/src/main.rs#L47

But also some packages can have a license file or none at all. So my tool also prints a warning in that case. IE:

```
Unable to check license from "/Users/william/development/kanidm/kanidm/vendor/nss_kanidm/Cargo.toml". You may need to check this manually
Unable to check license from "/Users/william/development/kanidm/kanidm/vendor/pam_kanidm/Cargo.toml". You may need to check this manually
Unable to find license in "/Users/william/development/kanidm/kanidm/vendor/ring/Cargo.toml". You may need to check "/Users/william/development/kanidm/kanidm/vendor/ring/LICENSE" for details.
Unable to find license in "/Users/william/development/kanidm/kanidm/vendor/sshkeys/Cargo.toml". You may need to check "/Users/william/development/kanidm/kanidm/vendor/sshkeys/LICENSE" for details.
Unable to find license in "/Users/william/development/kanidm/kanidm/vendor/webpki/Cargo.toml". You may need to check "/Users/william/development/kanidm/kanidm/vendor/webpki/LICENSE" for details.
License: ( Apache-2.0 OR BSL-1.0 ) AND ( Apache-2.0 OR ISC OR MIT ) AND ( Apache-2.0 OR MIT ) AND ( Apache-2.0 WITH LLVM-exception OR Apache-2.0 OR MIT ) AND ( CC0-1.0 OR Apache-2.0 ) AND ( MIT OR Apache-2.0 OR Zlib ) AND ( Unlicense OR MIT ) AND ( Zlib OR Apache-2.0 OR MIT ) AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND CC0-1.0 AND ISC AND MIT AND MPL-2.0 AND MPL-2.0+ AND 
```

(NOTE this license string is for a different project which has a lot of dependencies so it's a good test of the lock2rpmprovides tool.)

EDIT: If you wanted, I could make my tool have a --json flag so that you can just call it then parse the dict into your python here to automaticaly patch the rpm spec? Or we could add the spec file update function into mytool? Or you could recreate the license checker in python.

Up to you, open to all of these suggestions.

rpm/bundle-rust-downstream
decathorpe: rust2rpm has normalization / mapping functions from cargo/semver into RPM-compatible version strings:
https://pagure.io/fedora-rust/rust2rpm/blob/master/f/rust2rpm/metadata.py

Using those would make the bundled Provides use the same version strings as the crates packaged as "native" RPM pacakges.

Firstyear: Hmm I'm not clear on how this code would handle the version string for 'wasi'. It has a version string of '0.9.0+wasi-snapshot-preview1' which is a dependency of getrandom -> ahash -> concread. The version string here is valid for rust, but rpm was having issues at it contains multiple hypens in it's current form. How would you normalise or parse this?

decathorpe: You're right, it appears that it doesn't handle this case. We don't have wasi packaged yet, which is probably why this bug went unnoticed. However, for other packages with "+" in the version string (like cargo-c, zstd-sys, libgit2-sys, etc) we normalize these downstream with patches (since the +suffix is not strictly version information, but "context" information).

Anyway, reported for rust2rpm: https://pagure.io/fedora-rust/rust2rpm/issue/124

Firstyear: Commented on the issue/subscribed. It would be good to know what you plan to do to normalise here so I can keep cargo-lock2rpmprovides consistent.

At the very least, for now I think we can just normalise it by hand as we generate the provides line until we have this sorted/fixed.

rpm/bundle-rust-downstream
droideck: I also thought about that but I was hesitant as it may produce a lot of files that the admin will clean eventually.
But thinking more... I think it's better than losing some info if you run it accidentally twice with a wrong result...

Firstyear: You could make a sub directory or a dotfolder for the backups to land in instead. Or you could put them into /tmp? That would keep them available but also makes it a bit neater. 

I guess the other thing, is do we need the backup? We have git for version control ....

rpm.mk
droideck: Looking at https://directory.fedoraproject.org/docs/389ds/howto/howto-fedora-release-process.html (for which I develop the tool so it's easier to do builds) - it makes more sense to me to divide the process.

1. We create the tarball with all of the tags (and Cargo.lock is updated)
2. We go to Fedora RPM repo, check out the right branch, and then update the specfile.

rpm/bundle-rust-downstream
droideck: I agree but maybe it'll be useful for some workflows...
It's just a couple of lines and it's optional. I'll ask Mark what he would like for rpm.mk :)

Firstyear: Yeah, it's not the "biggest concern" here, but I get your point that it's useful in some work flows. Given this affects you and @mreynolds389 the most, i'll leave it to your judgement here mate.

rpm.mk
Firstyear: Sure, just was curious about this step is all. Again, it's the process for you and @mreynolds389 so it's up to you.

droideck:
Changes are made. Please, check :)

Firstyear (approved):

Issue #4644, opened 2021-02-19
Large updates can reset the CLcache to the beginning of the changelog


tbordaz:
**Issue Description**
The replication agreements are using bulk load to load updates. For bulk load it uses a cursor with DB_MULTIPLE_KEY and DB_NEXT. Before it initializes the cursor with DB_SET.

If during the cursor/DB_SET the CSN refers to an update that is larger than the size of the provided buffer, then the cursor remains not initialized and c_get returns DB_BUFFER_SMALL.
The consequence is that the next c_get(DB_MULTIPLE_KEY and DB_NEXT) will return the first record in the changelog DB.

**Package Version and Platform:**
 All version

**Steps to Reproduce**
See BZ

**Expected results**
When doing a bulk load from a specific update,
the changelog cache should successfully fetch the update and not reset the CLcache to the head of the changelog.

tbordaz:
9bdd9ec3f..f581979f9  master
58725517a..3687e1cae  389-ds-base-1.4.4
34199dcb9..1c55fd2fa  389-ds-base-1.4.3

PR #4647, opened 2021-02-22
Issue 4644 - Large updates can reset the CLcache to the beginning of …

### LINKED ISSUE #4492: Changelog cache can upload updates from a wrong starting point (CSN)

tbordaz:
…the changelog

Bug description:
	The replication agreements are using bulk load to load updates.
	For bulk load it uses a cursor with DB_MULTIPLE_KEY and DB_NEXT.
	Before using the cursor, it must be initialized with DB_SET.

	If during the cursor/DB_SET the CSN refers to an update that is larger than
	the size of the provided buffer, then the cursor remains not initialized and
	c_get returns DB_BUFFER_SMALL.

	The consequence is that the next c_get(DB_MULTIPLE_KEY and DB_NEXT) will return the
	first record in the changelog DB. This break CLcache.

Fix description:
	The fix is to harden cursor initialization so that if DB_SET fails
	because of DB_BUFFER_SMALL. It reallocates buf_data and retries a DB_SET.
	If DB_SET can not be initialized it logs a warning.

	The patch also changes the behaviour of the fix #4492.
	#4492 detected a massive (1day) jump prior the starting csn and ended the
	replication session. If the jump was systematic, for example
	if the CLcache got broken because of a too large updates, then
	replication was systematically stopped.
	This patch suppress the systematically stop, letting RA doing a big jump.
	From #4492 only remains the warning.

relates: https://github.com/389ds/389-ds-base/issues/4644

Reviewed by:

Platforms tested: F31

### DIFF
--- a/ldap/servers/plugins/replication/cl5_clcache.c
+++ b/ldap/servers/plugins/replication/cl5_clcache.c
@@ -365,9 +365,7 @@ clcache_load_buffer(CLC_Buffer *buf, CSN **anchorCSN, int *continue_on_miss, cha
             }
             csn_as_string(buf->buf_current_csn, 0, curr);
             slapi_log_err(loglevel, buf->buf_agmt_name,
-                      "clcache_load_buffer - bulk load cursor (%s) is lower than starting csn %s. Ending session.\n", curr, initial_starting_csn);
-            /* it just end the session with UPDATE_NO_MORE_UPDATES */
-            rc = CLC_STATE_DONE;
+                      "clcache_load_buffer - bulk load cursor (%s) is lower than starting csn %s.\n", curr, initial_starting_csn);
         }
     }
 
@@ -408,10 +406,7 @@ clcache_load_buffer(CLC_Buffer *buf, CSN **anchorCSN, int *continue_on_miss, cha
                     }
                     csn_as_string(buf->buf_current_csn, 0, curr);
                     slapi_log_err(loglevel, buf->buf_agmt_name,
-                            "clcache_load_buffer - (DB_SET_RANGE) bulk load cursor (%s) is lower than starting csn %s. Ending session.\n", curr, initial_starting_csn);
-                    rc = DB_NOTFOUND;
-
-                    return rc;
+                            "clcache_load_buffer - (DB_SET_RANGE) bulk load cursor (%s) is lower than starting csn %s.\n", curr, initial_starting_csn);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/replication/cl5_clcache.c
```diff
@@ -467,17 +498,24 @@ clcache_load_buffer_bulk(CLC_Buffer *buf, int flag)
...
-                               &buf->buf_data,
-                               DB_SET);
+            rc = clcache_cursor_set(cursor, buf);
         }
 
-        /*
-         * Continue if the error is no-mem since we don't need to
-         * load in the key record anyway with DB_SET.
-         */
         if (0 == rc || DB_BUFFER_SMALL == rc) {
```
progier389, line 504: Should rather be: if (00 == rc) {
  and let the function fails if rc == DB_BUFFER_SMALL (that will stop the replication with an error message about the fact that we failed to get next csn) 
Note: that is IMHO a minor issue because I do not think that c_get could return DB_BUFFER_SMALL after the realloc (because changelog records size does never change)

ldap/servers/plugins/replication/cl5_clcache.c
```diff
@@ -467,17 +498,24 @@ clcache_load_buffer_bulk(CLC_Buffer *buf, int flag)
...
+            * that the cursor will be reset to the beginning of the changelog.
+            * Returning an error at this point will stop replication that is
+            * a risk. So just accept the risk of a reset to the beginning of the CL
+            * and log an alarming message.
+            */
+           if (rc == DB_BUFFER_SMALL) {
+               slapi_log_err(SLAPI_LOG_WARNING, buf->buf_agmt_name,
+                             "clcache_load_buffer_bulk - Fail to position on csn=%s from the changelog (too large update ?). Risk of full CL evaluation.\n",
+                             (char *)buf->buf_key.data);
+           }
```
progier389, line 518: should be removed if  above test is 0 == rc

progier389 (approved):
I would have tested the double DB_BUFFER_SMALL error differently (aborting the replication rather that looping back on start of changelog)  (as described in my comments) 
but as i doubt this case could really happen, the current fix is ok for me.

ldap/servers/plugins/replication/cl5_clcache.c
tbordaz: I made a previous hotfix #4492 that stop replication session when detecting a big jump backward. The hotfix was failing because replication got broken with constant big jumps session after session. Without that hotfix replication kept going on even if restarted from the beginning of the CL.

So I prefer to get a permissive code that let replication going on even at a cost of a big jump.
In case of DB_BUFFER_SMALL during DB_SET (that I hope will no long happen) we just log a warning to help diagnose

progier389: Fine. Anyway as I do not think that we could hit the double DB_BUFFER_SMALL, I have no strong feeling about the way we handle it ! -;)

Issue #4646, opened 2021-02-20
UI/CLI - DNA plugin shared config child entries are not being managed correctly


mreynolds389:
**Issue Description**

The new CLI/UI assume that the DNA shared config child entries are managed by the Admin, but in fact they are managed by the server automatically.  The only thing an Admin can/should change in these entries is the remote Bind Method and remote Conn Protocol for retrieving new range sets.  The admin should not adjust any other attributes in the shared config sub entries.

This is will impact Docs, UI, and CLI.

mreynolds389:
3687e1cae..e86389b4c  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
1c55fd2fa..de09b8f99  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
74ac1f5b1..7e605d48a  389-ds-base-1.4.2 -> 389-ds-base-1.4.2

PR #4650, opened 2021-02-22
Issue 4646 - CLI/UI - revise DNA plugin management

mreynolds389:
Bug Description:

There was a false assumption that you have to create the shared DNA
server configuration entry, but in fact the server creates and manages
this entry.  The only thing you should edit in this entry are the
remote Bind Method and Connection Protocol.

Fix Description:

Remove the options to create the shared config entry, and edit the
core/reserved attributes.

Also fixed some issues where we were not showing CLI plugin output in
proper JSON.  This required some changes to the UI as well.

Relates: https://github.com/389ds/389-ds-base/issues/4646

### DIFF
--- a/ldap/servers/plugins/dna/dna.c
+++ b/ldap/servers/plugins/dna/dna.c
@@ -1884,16 +1884,23 @@ dna_get_shared_servers(struct configEntry *config_entry, PRCList **servers, int 
                     continue;
                 }
                 /* see if we defined a server manually */
-                if (server->remote_bind_method) {
+                if (server->remote_bind_method || server->remote_conn_prot) {
                     char *reason = NULL;
                     int err = 0;
 
+                    if (server->remote_bind_method == NULL || server->remote_conn_prot == NULL) {
+                        reason = "You must set both a bind method, and a connection protocol";
+                        err = 1;
+                        goto done;
+                    }
                     if (strcasecmp(server->remote_bind_method, DNA_METHOD_DIGESTMD5) == 0 ||
-                        strcasecmp(server->remote_bind_method, DNA_METHOD_SIMPLE) == 0) {
+                        strcasecmp(server->remote_bind_method, DNA_METHOD_SIMPLE) == 0)
+                    {
                         /* requires a DN and password */
                         if (!server->remote_binddn || !server->remote_bindpw) {
                             reason = "missing bind DN and/or password.";
                             err = 1;
+                            goto done;
                         }
                     }
                     if (strcasecmp(server->remote_bind_method, DNA_METHOD_SSL) == 0) {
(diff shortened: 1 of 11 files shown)


droideck:
I probably miss something... But the docs say that admin should add dnaSharedConfig (it was like that for RHDS 10)
https://access.redhat.com/documentation/en-us/red_hat_directory_server/10/html-single/administration_guide/index#dna-syntax

and I also see the same pattern across our tests.

Probably, the fact has changed at some point... If you can easily look it up, could you please share the version/the point where it has changed?

mreynolds389:
> I probably miss something... But the docs say that admin should add dnaSharedConfig (it was like that for RHDS 10)
> https://access.redhat.com/documentation/en-us/red_hat_directory_server/10/html-single/administration_guide/index#dna-syntax
> 
> and I also see the same pattern across our tests.
> 
> Probably, the fact has changed at some point... If you can easily look it up, could you please share the version/the point where it has changed?

I think the docs have been wrong for a very long time (I just filed a doc bug: https://bugzilla.redhat.com/show_bug.cgi?id=1931582).  It's a just confusing setup.  The server definitely creates the shared config subentries automatically.   And it maintains the remaining values, etc.  So we should not have the ability to alter it except for the new remote settings.

I also checked FreeIPA, they also do not create those subentries but they use them.

mreynolds389:
This reminds me, there are other problems with the DNA CLI.  Listing shared config entries only shows the first RDN, not the port.  So you can not use these "list" values for "showing" the entry.  Basically the list is useless.  I think maybe the CLI should be changed to use a single positional arg "HOSTNAME:PORT" instead of two separate positional arguments "HOSTNAME" and "PORT").  Either way I need to fix the "listing" to include the port, but what are your thoughts on changing the positional args?

droideck:
> > I probably miss something... But the docs say that admin should add dnaSharedConfig (it was like that for RHDS 10)
> > https://access.redhat.com/documentation/en-us/red_hat_directory_server/10/html-single/administration_guide/index#dna-syntax
> > and I also see the same pattern across our tests.
> > Probably, the fact has changed at some point... If you can easily look it up, could you please share the version/the point where it has changed?
> 
> I think the docs have been wrong for a very long time (I just filed a doc bug: https://bugzilla.redhat.com/show_bug.cgi?id=1931582). It's just a confusing setup. The server definitely creates the shared config subentries automatically. And it maintains the remaining values, etc. So we should not have the ability to alter it except for the new remote settings.
> 
> I also checked FreeIPA, they also do not create those subentries but they use them.

Okay, got it!

> This reminds me, there are other problems with the DNA CLI. Listing shared config entries only shows the first RDN, not the port. So you can not use these "list" values for "showing" the entry. Basically the list is useless. I think maybe the CLI should be changed to use a single positional arg "HOSTNAME:PORT" instead of two separate positional arguments "HOSTNAME" and "PORT"). Either way I need to fix the "listing" to include the port, but what are your thoughts on changing the positional args?

I wonder if we can (and somebody does) specify the entry like this with dnaSecurePortNum:

    dnaHostname=host_name+dnaSecurePortNum=port_number,ou=ranges,dc=example,dc=com
    
If it's not a practice, I think 'hostname:port' looks good.

mreynolds389:
> > This reminds me, there are other problems with the DNA CLI. Listing shared config entries only shows the first RDN, not the port. So you can not use these "list" values for "showing" the entry. Basically the list is useless. I think maybe the CLI should be changed to use a single positional arg "HOSTNAME:PORT" instead of two separate positional arguments "HOSTNAME" and "PORT"). Either way I need to fix the "listing" to include the port, but what are your thoughts on changing the positional args?
> 
> I wonder if we can (and somebody does) specify the entry like this with dnaSecurePortNum:
> 
> ```
> dnaHostname=host_name+dnaSecurePortNum=port_number,ou=ranges,dc=example,dc=com
> ```

Checking the DNA source code, we only use the dnaPortNum for constructing the DN of the sub-entries it creates.  But if someone created their own entry using dnaSecurePortNum it would still work.  But I think only FreeIPA really uses DNA so the risk is minimal, and it would only impact CLI/UI - not the actual plugin functionality.

mreynolds389:
Made a few more changes, and even fixed a crash in dna when you misconfigure the remote bind & conn protocol.  Please review..

droideck (approved):
Looks good to me!

PR #4651, opened 2021-02-22
Issue 4648 - Fix some issues and improvement around CI tests

progier389:
Description: 
  Some pytests fails:
      - because database file suffix moved from .db to .db4 some time ago.
      - because semanage fails if ci test are run by non root user
  And it would be useful to keep the ns-slapd stderr for debug build 

Solution:
     - use wildcard when checking the path
     - fix dbscan to recognize changelog whatever is the db version.
     - when detaching ns-slapd process, redirect stderr into a tmp file rather than into /dev/null
     - to not try to relabel the ports if user is not root     
     
  Reviewers: @mreynolds389 @vashirov @droideck

### DIFF
--- a/dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
+++ b/dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
@@ -140,7 +140,7 @@ def local_policy(topology_st, add_user):
 
     log.info("Setting fine grained policy for user ({})".format(USER_DN))
 
-    subprocess.call(['%s/dsconf' % topology_st.standalone.get_sbin_dir(),
+    subprocess.call(['%s/dsconf' % topology_st.standalone.get_bin_dir(),
                      'slapd-standalone1',
                      'localpwp',
                      'adduser',
--- a/dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
+++ b/dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
@@ -140,7 +140,7 @@ def local_policy(topology_st, add_user):
 
     log.info("Setting fine grained policy for user ({})".format(USER_DN))
 
-    subprocess.call(['%s/dsconf' % topology_st.standalone.get_bin_dir(),
+    subprocess.call(['%s/dsconf' % topology_st.standalone.get_sbin_dir(),
                      'slapd-standalone1',
                      'localpwp',
                      'adduser',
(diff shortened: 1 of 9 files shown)


dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
```diff
@@ -140,7 +140,7 @@ def local_policy(topology_st, add_user):
 
     log.info("Setting fine grained policy for user ({})".format(USER_DN))
 
-    subprocess.call(['%s/dsconf' % topology_st.standalone.get_sbin_dir(),
+    subprocess.call(['%s/dsconf' % topology_st.standalone.get_bin_dir(),
```
mreynolds389: But dsconf is in /usr/sbin by default.  Maybe this needs to handle both locations?

progier389: find $DSDIR | grep dsconf
/home/progier/sb/i4648/tst/ci-install/bin/dsconf
 I will double check why it it there in my build ...

ldap/servers/slapd/detach.c
```diff
@@ -163,9 +164,22 @@ detach(int slapd_exemode, int importexport_encrypt, int s_port, daemon_ports_t *
             return 1;
         }
         (void)dup2(sd, 0);
+        /* Lets ignore libaccess printf */
         (void)dup2(sd, 1);
         (void)dup2(sd, 2);
         close(sd);
+#ifdef DEBUG
+        /* But preserve other errors like loader undefined symbols */
+		sprintf(buf, "/tmp/ns-slapd-%d-XXXXXX.stderr", getpid());
```
vashirov: We have `PrivateTmp` enabled by default for our dirsrv service. Perhaps this should be logged in some other place, such as `/run/dirsrv`? Then it would be possible to pick up by our tests.

dirsrvtests/tests/suites/password/pwdPolicy_warning_test.py
vashirov: Yeah, in rpm builds it goes in `/usr/sbin/`: https://github.com/389ds/389-ds-base/blob/master/rpm/389-ds-base.spec.in#L654

progier389: Ok it is a problem in my environment - I must revert that change

ldap/servers/slapd/detach.c
progier389: Good point. 
But /run/dirsrv is not writable for non root install and I do not want to have to read the dse.ldif so early to get the install path. 
So I propose to first try in /run/dirsrv then in /tmp in case of failure.
An alternative would be to use /var/tmp

droideck (approved):
The changes LGTM!
Nice clean-up! Thanks!

Issue #4649, opened 2021-02-22
crash in sync_repl when a MODRDN create a cenotaph


tbordaz:
**Issue Description**
When an operation is flagged OP_FLAG_NOOP, it skips BETXN plugins but calls POST plugins.
As a consequence, it breaks a sync_repl assumption that operation index (in the pending list) stored in BETXN preop are later reused in POST op. It leads to dereferencing a NULL pointer

**Package Version and Platform:**
 Only master (#4526)

**Steps to Reproduce**
Automatic TC to be provided

**Expected results**
Should not crash

tbordaz:
1c6adb372..9bdd9ec3f  master
4e3b2d316..58725517a  389-ds-base-1.4.4
1c9180bac..34199dcb9  389-ds-base-1.4.3

tbordaz:
In 1.4.3, testcase was missing a part
40ef2e923..942cc16d2  389-ds-base-1.4.3

PR #4652, opened 2021-02-22
Issue 4649 - crash in sync_repl when a MODRDN create a cenotaph

tbordaz:
Bug description:
	When an operation is flagged OP_FLAG_NOOP, it skips BETXN plugins but calls POST plugins.
	For sync_repl, betxn (sync_update_persist_betxn_pre_op) creates an operation extension to be
	consumed by the post (sync_update_persist_op). In case of OP_FLAG_NOOP, there is no
	operation extension.

Fix description:
	Test that the operation is OP_FLAG_NOOP if the operation extension is missing

relates: https://github.com/389ds/389-ds-base/issues/4649

Reviewed by:

Platforms tested: F31


Firstyear (approved):
Ack, looks good.

Issue #4656, opened 2021-03-05
Remove problematic language from CLI tools and UI


mreynolds389:
**Issue Description**

Remove terms like "master" and "slave" from UI/CLI tools and replace it with "Supplier/Consumer"

mreynolds389:
e8a4c61b7..f32cd6d1f  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
9e5fe7aec..444ffba00  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

Next, need to make 1.4.4 and 1.4.3 backwards compatible with replication CLI

mreynolds389:
> [e8a4c61](https://github.com/389ds/389-ds-base/commit/e8a4c61b7f0952846b9ff02fc9c1a373249e4e65)..[f32cd6d](https://github.com/389ds/389-ds-base/commit/f32cd6d1fcb4a73825c3554d965115c227b15d0b) 389-ds-base-1.4.4 -> 389-ds-base-1.4.4
> [9e5fe7a](https://github.com/389ds/389-ds-base/commit/9e5fe7aec3bf43ad59f348eb9934cf9275123b4b)..[444ffba](https://github.com/389ds/389-ds-base/commit/444ffba0096a64e4bd5f7c2e7c9b54a733eb81ef) 389-ds-base-1.4.3 -> 389-ds-base-1.4.3
> 
> Next, need to make 1.4.4 and 1.4.3 backwards compatible with replication CLI


444ffba00..6f43954b1  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
f32cd6d1f..6fddab6d6  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

Lastly, ds-replcheck was missed during the original change, need to fix that in master branch...

mreynolds389:
> Lastly, ds-replcheck was missed during the original change, need to fix that in master branch...

f3c131296..d5fdea905  master -> master

minexew:
Is the main Git branch also going to be renamed to... main?

mreynolds389:
> Is the main Git branch also going to be renamed to... main?

No, this is just in the software and documentation that Red Hat ships.

Firstyear:
@mreynolds389 IIRC github have released some tooling to make master -> main change easier.

mreynolds389:
> @mreynolds389 IIRC github have released some tooling to make master -> main change easier.

[...]

PR #4657, opened 2021-03-05
Issue 4656 - Remove problematic language from UI/CLI/lib389

mreynolds389:
Description:  

Replace "master" and "slave" with more appropriate names


elkris:
Hi Mark,
I think it is more than time to make these language changes and get rid of the master-slave pattern (which never was a correct description of the roles, no "master" would have looked at a "slave" as a identical copy of himself). But I am not sure if the new terminology is optimal. There are two different points of view, one is temporary in a replication session where one supplier updates a consumer, and in the next session the roles can be reversed. And then there is the distinction of "read only" and "updatable" replicas. master would have to be replaced by a term describing a replica which accepts updates - and I can't find a proper term, not in english, but also not in german.  If we can't find anything better the overall use of "supplier" should be fine, but it is not optimal.
And I have not checked all changes, but in full consequence even terms like MMR would have to be changed

mreynolds389:
> Hi Mark,
> I think it is more than time to make these language changes and get rid of the master-slave pattern (which never was a correct description of the roles, no "master" would have looked at a "slave" as a identical copy of himself). But I am not sure if the new terminology is optimal. There are two different points of view, one is temporary in a replication session where one supplier updates a consumer, and in the next session the roles can be reversed. And then there is the distinction of "read only" and "updatable" replicas. master would have to be replaced by a term describing a replica which accepts updates - and I can't find a proper term, not in english, but also not in german. If we can't find anything better the overall use of "supplier" should be fine, but it is not optimal.
> And I have not checked all changes, but in full consequence even terms like MMR would have to be changed

Oh there are going to be a lot of changes related to this.  This is just phase one.  We still have the config, plugin entry, logging, source code, etc.  The current plan we came up with was from the Admin perspective, it's Supplier/Hub/Consumer, but internally it will be:  sender/receiver  - so this will be used in logging and other naming in the replication code.  We debated many many terms (readwrite/readwrite-hub/readonly,... etc, etc, etc), but we settled on Supplier/Hub/Consumer (exposed terminology) and  sender/receiver (internal terminology).  Sorry, I guess we should have made the discussion public on 389-devel list.

Firstyear:
> discussion public on 389-devel list.

Yeeeeaaaaaaahhhhhh probably should have :) 

I still agree it's a change we need to make though - even on mailing lists I've been trying to avoid the M/S terminology.

Firstyear:
Had a quick look and it's okay from me. Only thing to be careful of is that it may break some peoples CLI scripts but IMO our cli isn't api stable so it's not a problem to me.

mreynolds389:
> Had a quick look and it's okay from me. Only thing to be careful of is that it may break some peoples CLI scripts but IMO our cli isn't api stable so it's not a problem to me.

Actually in 1.4.x I'm going to make it CLI backwards compatible, but in 2.x it needs to be removed.

mreynolds389:
> > discussion public on 389-devel list.
> 
> Yeeeeaaaaaaahhhhhh probably should have :)
> 
> I still agree it's a change we need to make though - even on mailing lists I've been trying to avoid the M/S terminology.

To be honest, it was only a RHEL requirement, and I thought it was going to be straightforward without much debate.  Simple change...  then it did turn into a debate internally, and now externally :-)

Firstyear:
> To be honest, it was only a RHEL requirement, and I thought it was going to be straightforward without much debate. Simple change... then it did turn into a debate internally, and now externally :-)

Hey, it's not a debate :) I think the change is good and should be done. No debate here at all, and thanks for putting in the hard work to achieve it!

I just would have liked upstream to be included from the start in the conversation. That's all. Especially as we have some more contributors who aren't inside RH it's good to have us all involved. 

But again. It looks like a good change to me, so ack here. Like anything I'm sure we'll hit a rough edge but we should do this.

Issue #4663, opened 2021-03-09
dsconf fails with 'ERROR: Error: values has to be a tuple, was None' when adding a new objectClass


vashirov:
**Issue Description**
`dsconf` fails with `ERROR: Error: values has to be a tuple, was None` when adding a new objectClass

**Package Version and Platform:**
1.4.4.3+

**Steps to Reproduce**
```
[root@server-f33 ds]# dsconf -v server-f33 schema objectclasses add testOC 
DEBUG: The 389 Directory Server Configuration Tool
DEBUG: Inspired by works of: ITS, The University of Adelaide
DEBUG: dsrc path: /root/.dsrc
DEBUG: dsrc container path: /data/config/container.inf
DEBUG: dsrc instances: []
DEBUG: dsrc no such section: slapd-server-f33
DEBUG: Called with: Namespace(instance='server-f33', verbose=True, binddn=None, bindpw=None, prompt=False, pwdfile=None, basedn=None, starttls=False, json=False, name='testOC', oid=None, desc=None, x_origin=None, must=None, may=None, kind=None, sup=None, func=<function add_objectclass at 0x7f5917562dc0>)
DEBUG: Instance details: {'uri': 'server-f33', 'basedn': None, 'binddn': None, 'bindpw': None, 'saslmech': None, 'tls_cacertdir': None, 'tls_cert': None, 'tls_key': None, 'tls_reqcert': None, 'starttls': False, 'prompt': False, 'pwdfile': None, 'args': {'ldapurl': 'server-f33', 'root-dn': None}}
DEBUG: Allocate <class 'lib389.DirSrv'> with ldapi://%2fvar%2frun%2fslapd-server-f33.socket
DEBUG: Allocate <class 'lib389.DirSrv'> with %2fvar%2frun%2fslapd-server-f33.socket
DEBUG: Allocate <class 'lib389.DirSrv'> with server-f33.example.com:389
DEBUG: Allocate <class 'lib389.DirSrv'> with server-f33.example.com:389
DEBUG: Allocate <class 'lib389.DirSrv'> with ldapi://%2fvar%2frun%2fslapd-server-f33.socket
DEBUG: Allocate <class 'lib389.DirSrv'> with %2fvar%2frun%2fslapd-server-f33.socket
DEBUG: Allocate <class 'lib389.DirSrv'> with server-f33.example.com:389
DEBUG: Allocate <class 'lib389.DirSrv'> with server-f33.example.com:389
[...]

PR #4664, opened 2021-03-09
Issue 4663 - CLI - unable to add objectclass/attribute without x-origin

mreynolds389:
Bug Description:

A regression was introduced that cuased attribute/objectclass adds
to fail becuase x-origin was not set.

Fix Description:

Make sure x_origin is initialized a tuple, which is how python-ldap
expects it to be, and make sure we don't set None in the schema
object either.

relates: https://github.com/389ds/389-ds-base/issues/4663


progier389 (approved):
LGTM

Issue #4459, opened 2020-11-24
dsctl ignores custom path and only uses paths in defaults.inf


mreynolds389:
**Issue Description**

If you set a custom backup directory during installation:

    [general]
    config_version = 2
    full_machine_name = localhost.localdomain
    start = True

    [slapd]
    port = 389
    backup_dir = /var/opt/dirsrv/slapd-{instance_name}/bak

The directory is correctly created, but when you run the CLI command to list the backups it fails:

    # dsctl -j localhost backups
    {
        "desc": "[Errno 2] No such file or directory: '/var/lib/dirsrv/slapd-localhost/bak'"
    }

Notice it looks in "/var/lib/dirsrv", and not "/var/opt/dirsrv".  This is because dsctl accesses the instance in an offline mode, so it only get the server's paths from defaults.inf.  We should attempt to get the server's paths from DSEldif class, only if that fails should we fallback to defaults.inf.

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=1901020

Firstyear:
Some days I regret supporting all the features of the original setup-ds.pl, like custom paths because it leads to things like this, but also, this would break SELinux policy too ....

But yes, I think in this case we'd need to read the dse.ldif in offline mode too.

mreynolds389:
> Some days I regret supporting all the features of the original setup-ds.pl, like custom paths because it leads to things like this, but also, this would break SELinux policy too ....

Actually the installer applies the selinux labels and it is correct after install.

mreynolds389:
> > Some days I regret supporting all the features of the original setup-ds.pl, like custom paths because it leads to things like this, but also, this would break SELinux policy too ....
> 
> Actually the installer applies the selinux labels and it is correct after install.

Actually it looks like it does not apply the correct label, and it looks like the uninstall is not removing the custom path :-/

Firstyear:
During uninstall we probably read from defaults, not dse.ldif as well. 

And during install we don't label any files because we assume the system policy exists and will label them correctly for us. 

IMO the issue is the user needlessly using /var/opt instead of /var/lib - unless they have a really really really compelling, amazing reason, I think they should not be doing this (and it's why there was a recent change to "hide" the file path options in the .inf template, so that people can't see that these changes are possible).

vashirov:
[...]

PR #4668, opened 2021-03-10
Issue 4459 - lib389 - Default paths should use dse.ldif if the server…

mreynolds389:
… is down

Bug Description:  

If a custom path is used for something like the backup directory, dsctl will still use the default path from defaults.inf.

Fix Description:  

When initializing the default Paths consult dse.ldif for some of the paths.

relates: https://github.com/389ds/389-ds-base/issues/4459


Firstyear (approved):
Ack, good work mate,

Issue #4666, opened 2021-03-09
Database chaining doesn't fail to other servers in farm when bind fails and disallowing anonymous bind


glbyers:
In our environment, we'd like to use a chaining backend to push BIND operations up to masters by way of the consumer (rather than client referral). We'd like to do this to ensure password lockout attributes are propagated to all consumers equally via our standard replication agreements. This is described here - https://directory.fedoraproject.org/docs/389ds/howto/howto-chainonupdate.html.

NOTE, we do not have hubs in our topology. Just masters and consumers, so no intermediate chaining.

We tested this process in our environment and it worked beautifully until we took it to production. Currently, we have just 2 masters and they are both sitting on some over-subscribed hardware that suffers from I/O starvation at certain times of the day. The plan is to scale out our masters eventually, but we're a little hamstrung with other projects and priorities. It worked extremely well until that time of day when masters suffered from I/O starvation, and hence, very long I/O wait times. This is generally short lived and happens at alternate times of the day for each of the masters. However, it seems that once both nsfarmservers have "failed", there is never any attempt by the consumer to retry them. This leads to bind errors as follows;

```
ldapwhoami -x -D "<binddn>" -W
Enter LDAP Password:
ldap_bind: Operations error (1)
        additional info: FARM SERVER TEMPORARY UNAVAILABLE
```

Except it is not temporary. It never recovers, even though all members of nsfarmservers are now healthy again.

 I tested various combinations of the chaining tuning params without success and after further debugging, confirmed that it always starts after a bind operation timeout. Looking into the chaining plugin code, I see that on operation timeout results in a call to cb_ping_farm to see if we can find another server in the pool that is available. However, it performs this operation (the comment is telling);
```

    /* NOTE: This will fail if we implement the ability to disable
       anonymous bind */
    rc = ldap_search_ext_s(ld, NULL, LDAP_SCOPE_BASE, "objectclass=*", attrs, 1, NULL,
                           NULL, &timeout, 1, &result);
    if (LDAP_SUCCESS != rc) {
        slapi_ldap_unbind(ld);
        cb_update_failed_conn_cpt(cb);
        return LDAP_SERVER_DOWN;
    }
```

So basically, because we've disallowed anonymous bind for anything but rootdse, it will always fail to find another available server. I have confirmed this by allowing anonymous bind on our masters while the issue was present, then subsequent binds on the consumers start working again.

[...]

PR #4669, opened 2021-03-12
Chaining DB with anonymous disallowed on remote server.

Firstyear:
This is an initial test of this case.

### DIFF
--- a/ldap/servers/plugins/chainingdb/cb_conn_stateless.c
+++ b/ldap/servers/plugins/chainingdb/cb_conn_stateless.c
@@ -846,23 +846,45 @@ cb_stale_all_connections(cb_backend_instance *cb)
 int
 cb_ping_farm(cb_backend_instance *cb, cb_outgoing_conn *cnx, time_t end_time)
 {
-
+    int version = LDAP_VERSION3;
     char *attrs[] = {"1.1", NULL};
     int rc;
+    int ret;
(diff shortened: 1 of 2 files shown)


droideck (approved):
The change and the test look good to me (to my understanding of #4666).

Still, I think @mreynolds389 should take a look too as noted in the issue as it's a change in behaviour and we should assess it correctly (versions where it may land, etc.)

ldap/servers/plugins/chainingdb/cb_conn_stateless.c
```diff
@@ -846,40 +846,84 @@ cb_stale_all_connections(cb_backend_instance *cb)
...
+        return LDAP_INVALID_CREDENTIALS;
+    }
 
     secure = cb->pool->secure;
     if (cb->pool->starttls) {
         secure = 2;
     }
     ld = slapi_ldap_init(cb->pool->hostname, cb->pool->port, secure, 0);
     if (NULL == ld) {
+        if (ret == 0) {
```
mreynolds389, line 895: Everything looks good, but you probably don't technically need this "ret" check, same with the one check down below.  But it's also technically not any less performat, since slapi_ch_free_string does a NULL check :-)  All that being said, ack.

progier389 (approved):
LGTM

PR #4675, opened 2021-03-17
Issue 4169 - UI - Add PF4 charts for server stats

mreynolds389:
Description:  Added charts for current connections (that does NOT use cn=monitor),
              server memory size and CPU usage.


droideck:
Looks really good!
I have only one small concern both code and visual-wise.

When a user first opens the charts it shows 10 empty values which, I think, looks odd because then it starts back from 1 tick and continues above 10 - so it's a bit confusing.
Can we maybe start from 1 tick to begin with? The chart will optimize it (look at CPU chart I modified). It looks a bit empty but I think that's the point of 'beginning of the monitoring' action:
charts1

An that's how many ticks look like (so everything is okay if we start from 1 and not from 10):
charts2

Also, code-wise, I think these lists can be initialized with something like this:
console.log(Array.from(Array(10).keys(), x => ({"name": "", "x": x + 1, "y": 0})));

What do you think?

mreynolds389:
Okay I think I fixed everything you had a concern about.  Note - I could not use the Array() technique because it basically created an array of Objects (and function), so I had to use a for loop to do optimization of the initialization.  Please review...

droideck:
One small issue and it's probably not this PR's fault but:
cache-chart
For the smaller chart (cache in suffix monitoring), it probably should start from 6 instead of 11.

droideck (approved):
Besides the minor issue - LGTM!

src/cockpit/389-console/src/lib/monitor/serverMonitor.jsx
```diff
@@ -1,201 +1,550 @@
...
+            connChart: [...this.state.initChart],
+        });
+    }
+
+    refreshCharts() {
+        let cmd = [
+            "ps", "-ef", "|", "grep", "slapd-" + this.props.serverId,
+            "|", "grep", "dirsrv"
+        ];
+        cmd = "ps -ef | grep slapd-" + this.props.serverId + " | grep dirsrv";
```
tmihinto: Second assignment to the variable "cmd".
You might want to remove the first one.

PR #4704, opened 2021-03-29
Issue 3585 - LDAP server returning controltype in different sequence

bsimonova:
Description:
Added a test to check sequence of ldap controlType returned
when there are remaining or exhausted grace login.
Automation was not possible until now because of bug 1757699 in python-ldap
where no controls were returned in the error message after exception was raised
with exhausted grace login. The bug is fixed now.

Relates: https://github.com/389ds/389-ds-base/issues/3585

Reviewed by: ???


droideck (approved):
LGTM!

PR #4705, opened 2021-03-29
Issue 4169 - UI - migrate monitor tables to PF4

mreynolds389:
Description:  

Migrate from PF3 tables to PF4 tables.  This patch mostly handles the tables under the monitor tab, but there are many more tables that need migrating...


droideck (approved):
LGTM!
One small thing I wonder about (if it makes sense and it's easy to implement)

Now, a user has to click on '>' symbol. And if the user clicks on any other place in the row - the row is not expanded.
Should we make it so the field is expanded if the user clicks on any place within the row he'd like to expand? It feels a bit more natural but then we'll have a bit of a problem if the user would like to copy-paste any row's content... 
What do you think?

mreynolds389:
> LGTM!
> One small thing I wonder about (if it makes sense and it's easy to implement)
> 
> Now, a user has to click on '>' symbol. And if the user clicks on any other place in the row - the row is not expanded.
> Should we make it so the field is expanded if the user clicks on any place within the row he'd like to expand? It feels a bit more natural but then we'll have a bit of a problem if the user would like to copy-paste any row's content...
> What do you think?

Not sure it would work that way.  Maybe if I add "cellFormatters: [expandable]" to each column that might work.  Otherwise it would take a lot of overriding of the PF component to get that kind of behavior.  Let me try that tomorrow with the "expandable"  callFormatter.  If it doesn't work I'd prefer not add JS to achieve this behavior - just trying to stick with as much standard PF4 behavior as possible.

mreynolds389:
> > LGTM!
> > One small thing I wonder about (if it makes sense and it's easy to implement)
> > Now, a user has to click on '>' symbol. And if the user clicks on any other place in the row - the row is not expanded.
> > Should we make it so the field is expanded if the user clicks on any place within the row he'd like to expand? It feels a bit more natural but then we'll have a bit of a problem if the user would like to copy-paste any row's content...
> > What do you think?
> 
> Not sure it would work that way. Maybe if I add "cellFormatters: [expandable]" to each column that might work. Otherwise it would take a lot of overriding of the PF component to get that kind of behavior. Let me try that tomorrow with the "expandable" callFormatter. If it doesn't work I'd prefer not add JS to achieve this behavior - just trying to stick with as much standard PF4 behavior as possible.

Sorry that did not work.  While it would be nice, I just want to avoid tech debt on this (like when we move to PF5), and just keep to the standard PF behavior.  I'm going to merge this, thanks for the review!

Issue #4700, opened 2021-03-25
Regression in winsync replication agreement


flo-renaud:
**Issue Description**
In IPA environment, the configuration of a winsync replication agreement fails in the step adding `cn=replication manager,cn=config`. The regression was introduced with 389-ds-base-1.4.4.12-1.fc33.x86_64 (no issue with 389-ds-base-1.4.4.11-1.fc33.x86_64).
The command `ipa-replica-manage connect --winsync --binddn cn=Administrator,cn=users,dc=ad,dc=test --bindpw Secret123 --password Secret.123 --cacert /tmp/tmp.OCf3xgAY5f --passsync dummy ad-root.ad.test -v` fails with ldap.OPERATIONS_ERROR: Invalid DN.

**Package Version and Platform:**
 389-ds-base-1.4.4.12-1.fc33.x86_64

**Steps to Reproduce**
IPA nightly test test_integration/test_winsyncmigrate.py

More information available in https://pagure.io/freeipa/issue/8691

mreynolds389:
The only fixes in 1.4.4.12 are:

- Issue 4579 - libasan detects heap-use-after-free in URP test (#4584)
- Issue 4563 - Failure on s390x: ‘Fails to split RDN “o=pki-tomcat-CA” into components’ (#4573)
- Issue 4526 - sync_repl: when completing an operation in the pending list, it can select the wrong operation (#4553)
- Issue 4396 - Minor memory leak in backend (#4558) (#4572)
- Issue 4324 - Performance search rate: change entry cache monitor to recursive pthread mutex (#4569)
- Issue 5442 - Search results are different between RHDS10 and RHDS11

None of them directly impact Add operations or winsync/replication. The logs from the ipa ticket show:

    [03/Feb/2021:19:46:54.462782376 +0000] conn=6 op=5 ADD dn="cn=replication manager,cn=config"
    [03/Feb/2021:19:46:54.524714900 +0000] conn=6 op=5 RESULT err=1 tag=105 nentries=0 wtime=0.000111775 optime=0.061935050 etime=0.062040261

and the error log contains:

    [03/Feb/2021:19:46:54.527606161 +0000] - ERR - ipa_sidgen_add_post_op - [file ipa_sidgen.c, line 128]: Missing target entry.

ipa_sidgen_add_post_op():

    ...
    ret = slapi_pblock_get(pb, SLAPI_TARGET_DN, &dn_str); 
    if (ret != 0) { 
[...]

Issue #4396, opened 2020-10-22
Memory leak in BE (minor)


Firstyear:
```
=================================================================
==53701==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 40 byte(s) in 1 object(s) allocated from:
   #0 0x7fe2652ec27f in __interceptor_malloc (/usr/lib64/libasan.so.6+0xae27f)
   #1 0x7fe264e96b34 in slapi_ch_malloc /home/william/development/389ds/ds/ldap/servers/slapd/ch_malloc.c:95
   #2 0x7fe264eb41c9 in slapi_sdn_new /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:1888
   #3 0x7fe264eb4a4b in slapi_sdn_new_normdn_byval /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:2087
   #4 0x7fe264eb6871 in slapi_sdn_dup /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:2474
   #5 0x7fe264e8eb25 in be_addsuffix /home/william/development/389ds/ds/ldap/servers/slapd/backend.c:176
   #6 0x444bae in setup_internal_backends /home/william/development/389ds/ds/ldap/servers/slapd/fedse.c:2887
   #7 0x448d11 in main /home/william/development/389ds/ds/ldap/servers/slapd/main.c:718
   #8 0x7fe264655e09 in __libc_start_main (/lib64/libc.so.6+0x27e09)

Direct leak of 40 byte(s) in 1 object(s) allocated from:
   #0 0x7fe2652ec27f in __interceptor_malloc (/usr/lib64/libasan.so.6+0xae27f)
   #1 0x7fe264e96b34 in slapi_ch_malloc /home/william/development/389ds/ds/ldap/servers/slapd/ch_malloc.c:95
   #2 0x7fe264eb41c9 in slapi_sdn_new /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:1888
   #3 0x7fe264eb4a4b in slapi_sdn_new_normdn_byval /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:2087
   #4 0x7fe264eb6871 in slapi_sdn_dup /home/william/development/389ds/ds/ldap/servers/slapd/dn.c:2474
   #5 0x7fe264e8eb25 in be_addsuffix /home/william/development/389ds/ds/ldap/servers/slapd/backend.c:176
[...]

PR #4364, opened 2020-10-08
Issue 4265 - UI - Make the secondary plugins read-only

droideck:
Description: As some of the changes may break the server.
We should make all the plugins in the UI Plugins table read-only.
Only the ones in the left column should editable.
The change is only for UI.

Fixes: #4265

Reviewed by: ?

PR #4712, opened 2021-04-01
Issue 4700 - Regression in winsync replication agreement

tbordaz:
Bug description:
	#4396 fixes a memory leak but did not set 'cn=config' as
	DSE backend.
	It had no signicant impact unless with sidgen IPA plugin

Fix description:
	revert the portion of the #4364 patch that set be_suffix

relates: https://github.com/389ds/389-ds-base/issues/4700

Reviewed by:

Platforms tested: F33


progier389 (approved):
LGTM

Issue #4680, opened 2021-03-22
389ds coredump (@389ds/389-ds-base-nightly) in replica install with CA


flo-renaud:
**Issue Description**
The nightly tests using the copr repository @389ds/389-ds-base-nightly fail in `ipa-replica-install` if `--setup-ca` is added. See for instance PR #[790](https://github.com/freeipa-pr-ci2/freeipa/pull/790) with `389ds-fedora/simple_replication` ([logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/787260f2-89b9-11eb-9794-fa163eef4f62/)):
```
ipa: ERROR: stderr: Lookup failed: Preferred host replica0.ipa.test does not provide DNS.
Reverse DNS resolution of address 192.168.122.155 (master.ipa.test) failed. Clients may not function properly. Please check your DNS setup. (Note that this check queries IPA DNS directly and ignores /etc/hosts.)
Replica DNS records could not be added on master: Insufficient access: Insufficient 'add' privilege to add the entry 'idnsname=replica0,idnsname=ipa.test.,cn=dns,dc=ipa,dc=test'.
Custodia uses 'master.ipa.test' as master peer.
Failed to configure CA instance
See the installation logs and the following files/directories for more information:
  /var/log/pki/pki-tomcat
CA configuration failed.
The ipa-replica-install command failed. See /var/log/ipareplica-install.log for more information
```

The failure happens with a ns-slapd crash.
[ipa-replica-install.log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/787260f2-89b9-11eb-9794-fa163eef4f62/test_integration-test_simple_replication.py-TestSimpleReplication-install/replica0.ipa.test/var/log/ipareplica-install.log.gz):
```
2021-03-20T20:37:52Z DEBUG Starting external process
2021-03-20T20:37:52Z DEBUG args=['/usr/sbin/pkispawn', '-s', 'CA', '-f', '/tmp/tmpcxef7gno', '--debug']
2021-03-20T20:40:21Z DEBUG Process finished, return code=1
```

In the same timeframe, [journal](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/787260f2-89b9-11eb-9794-fa163eef4f62/test_integration-test_simple_replication.py-TestSimpleReplication-install/replica0.ipa.test/journal.gz) shows a crash of ns-slapd:
```
[...]

PR #4715, opened 2021-04-02
Issue 4680 - 389ds coredump (@389ds/389-ds-base-nightly) in replica install with CA

progier389:
Symptoms: 389ds coredump while running ipa-ca-install (during IPA nightly test) 

Cause: The changelog cache context (a CLC_Buffer named buf) get overwritten while reading the next block of changes after a large update (Because the read buffer was reset to wrong pointer (i.e the CLC_Buffer pointer rather than to the data buffer contained in that CLC_Buffer)

Fix:  reset the bulk data buffer to the right pointer.

Reviewer: tbordaz

relates to: [#4680](https://github.com/389ds/389-ds-base/issues/4680)

Tested on: F32

### DIFF
--- a/ldap/servers/plugins/replication/cl5_clcache.c
+++ b/ldap/servers/plugins/replication/cl5_clcache.c
@@ -1103,7 +1103,7 @@ clcache_cursor_get(dbi_cursor_t *cursor, CLC_Buffer *buf, dbi_op_t dbop)
          * if not sufficient it will be increased again
          */
         slapi_ch_free(&bulkdata->data);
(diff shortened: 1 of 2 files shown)


tbordaz:
LGTM

do you think it worth adding this additional [check](https://github.com/389ds/389-ds-base/issues/4680#issuecomment-803958963) ?

tbordaz (approved):

progier389:
> LGTM
> 
> do you think it worth adding this additional [check](https://github.com/389ds/389-ds-base/issues/4680#issuecomment-803958963) ?

I do not think it is absolutely needed with current code but the check make the API more robust and if we compare the cost of adding the check versus the cost of bug and diagnostic there is no doubt that we should add it. I will do that.

PR #4716, opened 2021-04-02
Issue 4699 - backend redesign phase 4 - db-mdb plugin implementation

### LINKED ISSUE #4699: backend redesign phase 4: mdb plugin implemenation
This is part of the ["Berkeley database" replacement by "Lightning Memory-Mapped Database Manager"](https://directory.fedoraproject.org/docs/389ds/design/design.html#backend-replacement) feature

This part is about implementing the plugin that interface dbimpl API  and the lmdb library, 
 The plugin is quite similar to the db-bdb one and its specific points are described [here](https://directory.fedoraproject.org/docs/389ds/design/backend-redesign-phase4.html)

progier389:
That is the bakend redesign phase 4  the db-mdb "plugin" implementation.
The design document is in [https://directory.fedoraproject.org/docs/389ds/design/backend-redesign-phase4.html](https://directory.fedoraproject.org/docs/389ds/design/backend-redesign-phase4.html)

Commits Description:
1. (automatic) : Issue 4699 - Backend redesign phase 4 - Rename dblayer and unprefixed… …
2. (manual) - Issue 4699 - Backend redesign phase 4 - Move some common functions fr… …
3. (automatic) - Issue 4699 - Backend redesign phase 4 - Create mdb plugin skeleton …
To be continued ... 

(Note for reviewers:  (automatic) means that changes are generated via scripts and IMHO do not need to be reviewed.)
 
Relates to #4699 

Reviewed by:

Tested on: F32


Makefile.am
```diff
@@ -1270,12 +1270,26 @@ libback_ldbm_la_SOURCES = ldap/servers/slapd/back-ldbm/ancestorid.c \
...
+	ldap/servers/slapd/back-ldbm/db-mdb/mdb_ldif2db.c \
+	ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c \
+	ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+
 
 
 libback_ldbm_la_CPPFLAGS = $(AM_CPPFLAGS) $(DSPLUGIN_CPPFLAGS) @db_inc@
 libback_ldbm_la_DEPENDENCIES = libslapd.la
-libback_ldbm_la_LIBADD = libslapd.la $(DB_LINK) $(LDAPSDK_LINK) $(NSPR_LINK)
+libback_ldbm_la_LIBADD = libslapd.la $(DB_LINK) $(LDAPSDK_LINK) $(NSPR_LINK) -llmdb
```
Firstyear: Shouldn't this -llmdb be part of db_link?

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
```diff
@@ -0,0 +1,1375 @@
...
+                " Decreasing the value from %lud to %lud\n", val, curval);
+       }
+        val = curval;
+    }
+
+    if (apply) {
+        conf->dsecfg.max_size = val;
+        if (CONFIG_PHASE_RUNNING == phase) {
+            slapi_log_err(SLAPI_LOG_NOTICE, "dbmdb_ctx_t_db_max_size_set",
+                          "New nsslapd-mdb-max-size will not take affect until the server is restarted\n");
```
Firstyear, line 299: Max size can only be set once at db creation time. Changes to this value will NOT take effect later, the db needs recreation.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
```diff
@@ -0,0 +1,1375 @@
...
+static void *
+dbmdb_ctx_t_db_durable_transactions_get(void *arg)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+
+    return (void *)((uintptr_t)(MDB_CONFIG(li)->dsecfg.durable_transactions));
+}
+
+static int
+dbmdb_ctx_t_db_durable_transactions_set(void *arg, void *value, char *errorbuf __attribute__((unused)), int phase __attribute__((unused)), int apply)
```
Firstyear, line 420: IMO we shouldn't enable or allow unsafe options like this ....

Firstyear (changes requested):
Wow mate, amazing work. So much to review, but a quick skim, it's a great start. I'll have a more detailed looked later this week, and I'm sure @mreynolds389 and @tbordaz will too.

Makefile.am
progier389: it is an option, I rather think about adding a second set of variable for lmdb and keep @db_inc@ and DB_LINK for bdb 

 using DB_LINK and replacing @db_inc@ by DB_INC should do the trick too ...

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
progier389: I think you are wrong as mdb_env_set_mapsize document explictly says that size could be increased by another process
 and there is an error code to handle that. 
That said size could not be decreased without rebuilding the db.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
progier389: If I remember rightly this one is about insuring that changes are really flushed on disk - and protect only against system crash and power failure. (IMHO the real risk is very limited (especially if power is secured and in multi supplier topology) and some people may want to assume it)
That said there are a lot of other options which are really unsafe, and like you I feel that we should not allow their use,

Makefile.am
progier389: Ok, I changed that one in last commit: replaced @db_inc@ by $(DB_INC) and moved -lldbm in DB_LINK

progier389:
Pushed a new commit with all changes to load ldif example file with ldapmodify.
Looks like there is an issue as ldapsearch is not working yet as it should

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
Firstyear: We shouldn't allow a setting where people can turn this off is my point :) I think even with multiple powersources, redudant hardware, you have to consider memory corruption, or kernel panic, or a crash. This could cause dataloss and issues, so it's better to the correct thing over the fast thing. I don't really like handing dangerous weapons out in a project that prides itself on correctness :)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
Firstyear: okay, so we need a check to make sure this isn't reduced?

progier389: I do not think we should prevent that because:
   - mdb_open_env will silently increase the value if it is too small.
   - this will allow to reduce the size if the db get rebuild

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
progier389: Well I am of two mind about that one and will check what the other are thinking.
I understand very well your point of view and I agree with your arguments but on the other side 
 - the risks you mention exist with bdb since at least 20 years and as far as I know we never get bothered with those.
And if the feature got developed that was because it was needed by some customers. 
IMHO it has some interest:
  start the server with durable txn off to perform massive ldif imports 
  then restart the server after enabling durable txn on to be safe.

Firstyear: Well, if that's the feature then it should be exposed as "non_durable_txn_during_import" rather than relying on the human operator to correctly "undo" the setting.

Just because it was something that has existed for 20 years doesn't mean we need to bring it forward now. And just because we haven't been bothered, doesn't mean someone has not been affected by it - we just were not told they were.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
Firstyear: Okay then :)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
progier389: just realizing I never wrote the conclusion of that point that was discussed some time ago in a meeting:
We will keep the non_durable_txn_during_import" (at least for now)

progier389:
I would like to merge this big PR:
   the test result with bdb are clean ( The usual  failure 1 in config and two in passwords)
   there are still failure on mdb (instance down in some tests) - but
   my feeling since this PR is so big is to merge it and open new issue to fix any remarks or remaining issue in future PR.
  BTW William, do you know how to clear the change requested flag?

mreynolds389:
> I would like to merge this big PR:
> the test result with bdb are clean ( The usual failure 1 in config and two in passwords)
> there are still failure on mdb (instance down in some tests) - but
> my feeling since this PR is so big is to merge it and open new issue to fix any remarks or remaining issue in future PR.

Yes please get this in :)

> BTW William, do you know how to clear the change requested flag?

I think you can still merge this regardless of this flag.

PR #4722, opened 2021-04-14
Issue 4169 - UI - migrate replication tables to PF4

mreynolds389:
Description:  

Migrated replication tables to PF 4 and cleaned up replication monitoring.

### DIFF
--- a/src/cockpit/389-console/src/lib/monitor/monitorModals.jsx
+++ b/src/cockpit/389-console/src/lib/monitor/monitorModals.jsx
@@ -687,7 +687,10 @@ class FullReportContent extends React.Component {
                 </div>
             );
         }
-        let reportHeader = "";
+        let reportHeader =
+            <h5 className="ds-margin-top-xlg">
+                There is no report, you must first generate the report in the <b>Prepare Report</b> tab.
+            </h5>;
         if (reportData.length > 0) {
             reportHeader = (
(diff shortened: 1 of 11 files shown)


jchapma (commented):
Looks good to me

droideck (approved):
Looks great! Ack

P.S. yeah, we really should thing about replacing Repl Monitor tabs report... I'll try to think more about this UI/UX solution...

mreynolds389:
> Looks great! Ack
> 
> P.S. yeah, we really should thing about replacing Repl Monitor tabs report... I'll try to think more about this UI/UX solution...

Yeah, it's not terrible but could be better.  

In fact, if you noticed I took the repl and winsync agreements out of the monitor completely.   You can't get the status or lag time without the credentials so it's not really useful, so it's gone -> meaning we only have the "Sync Report" left.

droideck:
> > Looks great! Ack
> > P.S. yeah, we really should thing about replacing Repl Monitor tabs report... I'll try to think more about this UI/UX solution...
> 
> Yeah, it's not terrible but could be better.
> 
> In fact, if you noticed I took the repl and winsync agreements out of the monitor completely. You can't get the status or lag time without the credentials so it's not really useful, so it's gone -> meaning we only have the "Sync Report" left.

Sounds good to me:) No objections on my side.

Issue #4725, opened 2021-04-19
[RFE] DS - Update the password policy to support Temporary Password Rules


tbordaz:
**Is your feature request related to a problem? Please describe.**
Enhance password policy to support registration password (One Time Password). 
One Time Password design is [here](https://www.port389.org/docs/389ds/design/otp-password-policy.html)

**Describe the solution you'd like**
Such password, once set by the administrator, can be used, a fixed number of times and have a limit of validity period (start and stop).

tbordaz:
8006632e7..adc2c8f50  master

mreynolds389:
This introduced compiler warnings, please fix:

```
../389-ds-base/ldap/servers/slapd/pw.c: In function ‘slapi_check_tpr_limits’:
../389-ds-base/ldap/servers/slapd/pw.c:2641:77: warning: comparison between pointer and integer
 2641 | ntry_attr_hasvalue(bind_target_entry, "pwdTPRReset", "TRUE") == NULL) {
      |                                                              ^~

../389-ds-base/ldap/servers/slapd/pw.c:2649:15: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 2649 |         value = slapi_entry_attr_get_ref(bind_target_entry, "pwdTPRUseCount");
      |               ^
../389-ds-base/ldap/servers/slapd/pw.c:2654:13: warning: implicit declaration of function ‘update_tpr_pw_usecount’; did you mean ‘update_trp_pw_usecount’? [-Wimplicit-function-declaration]
 2654 |             update_tpr_pw_usecount(pb, bind_target_entry, (int32_t) use_count);
      |             ^~~~~~~~~~~~~~~~~~~~~~
      |             update_trp_pw_usecount
../389-ds-base/ldap/servers/slapd/pw.c:2684:15: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 2684 |         value = slapi_entry_attr_get_ref(bind_target_entry, "pwdTPRExpireAt");
      |               ^
../389-ds-base/ldap/servers/slapd/pw.c:2712:15: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 2712 |         value = slapi_entry_attr_get_ref(bind_target_entry, "pwdTPRValidFrom");
      |               ^
[...]

PR #4727, opened 2021-04-19
Issue 4725 - [RFE] DS - Update the password policy to support a One T…

tbordaz:
…ime Password with expiration

Bug description:
	Enhance password policy to support registration password (One Time Password).
	design is  https://www.port389.org/docs/389ds/design/otp-password-policy.html

Fix description:
	The fix introduces new password policy configuration attributes
        (passwordOTP*) and entry (user) operational attributes (pwdOTP*).
	It supports One Time Password (fixed use count) and validity
        window (valid since-until).
	During bind it checks if the OTP limits are violated.
	During password update it computes and set
	operational attributed (pwdOTP*).

relates: https://github.com/389ds/389-ds-base/issues/4725

Reviewed by:

Platforms tested: F33


ldap/servers/slapd/pw.c
```diff
@@ -2554,6 +2612,131 @@ check_pw_storagescheme_value(const char *attr_name __attribute__((unused)), char
...
+    }
+
+    dn = slapi_entry_get_ndn(bind_target_entry);
+    pwpolicy = new_passwdPolicy(pb, dn);
+    if (pwpolicy == NULL) {
+        /* OTP limits are part of password policy => no limit */
+        return 0;
+    }
+
+    if (slapi_entry_attr_hasvalue(bind_target_entry, "pwdOTPReset", "TRUE") == NULL) {
```
Firstyear: There seem to be some compile warns here?

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -300,6 +301,618 @@ def test_pwd_min_age(topology_st, create_user, password_policy):
...
+        7. Success
+    """
+
+    ExpireAt = 6
+    # Set password policy config, passwordMaxFailure being higher than
+    # passwordOTPMaxUse so that OTP is enforced first
+    topology_st.standalone.config.replace('passwordMustChange', 'on')
+    topology_st.standalone.config.replace('passwordOTPMaxUse', str(-1))
+    topology_st.standalone.config.replace('passwordOTPDelayValidFrom', str(-1))
+    topology_st.standalone.config.replace('passwordOTPDelayExpireAt', str(ExpireAt))
```
Firstyear: I'm not sure about the naming of DelayExpireAt and how it's working. Is this acting as a default offset of the expireAt value if one is not set?

Firstyear:
@tbordaz After our discussion last night, how does "Temporary Password Rules" sound instead of OTP? Alternately, Temporary Access Policy?

ldap/servers/slapd/pw.c
```diff
@@ -735,6 +752,47 @@ update_pw_info(Slapi_PBlock *pb, char *old_pw)
...
+                          "OTP password reset by an admin, pwdOTPReset=TRUE\n");
+            slapi_mods_add_string(&smods, LDAP_MOD_REPLACE, "pwdOTPReset", "TRUE");
+
+            /* If useCount is -1, this means this limit is not enforced */
+            if (pwpolicy->pw_otp_maxuse >= 0) {
+                use_count = "0";
+            } else {
+                use_count = "-1";
+            }
+            slapi_log_err(SLAPI_LOG_TRACE, "update_pw_info", "pwdOTPUseCount = %s\n", use_count);
```
Firstyear: One last comment - is it possible for someone to manually set these values? 

Consider I want to create a set of new accounts. I want them to be active from 9:00am local time, and the password expires at 5pm the same day. 

With the current "relative" generation of the expireat/validfrom it would be really hard to achieve this for bulk accounts.

So is it possible that I can set these values manually and it would be respected?

Firstyear (approved):

Firstyear:
Besides my comment about manual time definitions,  Ithink this looks good :)

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
tbordaz: 'passwordOTPDelayxxx' are offsets (in second) from timestamp when the password reset occurs (admin setting userpassword). Those offsets define a period of time when the userpassword is accepted for authentication.
For example, let a policy with passwordOTPDelayValidFrom=60, passwordOTPDelayExpireAt=3600 and userpassword being reset at 14:05. The end user can authenticate with the userpassword between 14:06 and 15:05.
The defaults of passwordOTPDelayxxx is '-1', it means that this limit of the window is not enforced.

ldap/servers/slapd/pw.c
tbordaz: 'pwdOTP*' attributes are operational and direct updates are rejected with unwilling to perform. So only the server will set those values according to the policy.
The current PR propose to configure delays that are used to compute the validity window. So it is not possible to define precise time (e.g. April 21rst 2021 9:00am to April 21rst 2021 5pm).
Creating a bulk of accounts the validity window of the 1rst account will differ from the last account.
An option to do that would be to add in the password policy an attribute like 'passwordOTPValidityRef' that could be a fixed date. If this attribute is defined then the server takes it as reference instead of current time (during userpassword reset).

Firstyear: @tbordaz I guarantee the number one request we will get from this feature, is exactly what I am requesting - the ability to set these values manually. As a past admin, it's exactly how I would approach it, and something I would need and take value from.

tbordaz: @Firstyear, I agree that the use case is nice (an admin enabling a set of accounts that each one should be activited within a fixed period of time).
I propose to update the design and describe a possible option ('passwordOTPValidityRef') and if you agree to open a separated RFE.

The current use case is rather on demand reset. A user forgot his password, ask for a new temporary one (sent by email for example), then the use of the new password has some limitations (#auth and period of time).

tbordaz:
> @tbordaz After our discussion last night, how does "Temporary Password Rules" sound instead of OTP? Alternately, Temporary Access Policy?

thanks @Firstyear for these proposal. I tend to prefer "Temporary Password Rules" because it is clear it is about password. I will update the design/patch

ldap/servers/slapd/pw.c
Firstyear: Well, I think adding more attributes isn't needed. Just make it so that these attributes can be set by an admin. That way:

* If the passwordOTPExpireAt and ValidFrom are NOT present, we generate them.
* If they are present, or they are set, we just use the new manually set values.

I think this is the simplest and clearest approach.

tbordaz: Initially I thought that a change in the schema would allow that. I tried to drop 'NO-USER-MODIFICATION' and it works for DM. But it also works for a user having the rights to update entry. And I would not like to add an aci to deny updates of  those attributes.

So a solution would be a change in op_shared_is_allowed_attr path to allow DM to update those attributes even if they have 'NO-USER-MODIFICATION'.
I will work on a new patch with that solution.

Firstyear: But if the user can modify the entry they can already affect nsAccountLock or the userPassword. So ACI is the correct way to manage this here .... I'm not really seeing the issue?

tbordaz: Sorry my concern was invalid. This is fine defining as modifiable those operational attributes. I changed the schema and create testcases to check pwdTPRValidFrom and pwdTPRExpireAt can be modified if an ACI grants the rights

Firstyear: That's perfect! Thank you!

ldap/schema/02common.ldif
```diff
@@ -78,6 +78,13 @@ attributeTypes: ( 2.16.840.1.113730.3.1.2350 NAME ( 'passwordDictPath' 'pwdDictP
 attributeTypes: ( 2.16.840.1.113730.3.1.2351 NAME ( 'passwordUserAttributes' 'pwdUserAttributes' ) DESC '389 Directory Server password policy attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2352 NAME ( 'passwordBadWords' 'pwdBadWords' ) DESC '389 Directory Server password policy attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2366 NAME 'pwdReset' DESC '389 Directory Server password policy attribute type' EQUALITY booleanMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.2378 NAME 'pwdTPRReset' DESC '389 Directory Server password policy attribute type' EQUALITY booleanMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation X-ORIGIN '389 Directory Server' )
```
Firstyear, line 81: Do these need to stay as NO_USER_MODIFICATION?

Firstyear:
One last comment but otherwise I'm really happy with this! Thank you @tbordaz :)

ldap/schema/02common.ldif
tbordaz: 'pwdTPRReset' is flag that trigger the logic of maxuse/period_validity. Changing it, even by an admin is a bad idea. I prefer to keep it not modifiable. 

'pwdTPRMaxUse' is not modifiable because it was not necessary to the use case (bulk reset of entries, valid with the same period). DO you think it should also be modifiable ?

Firstyear: I think there is some usage for an admin to change MaxUse seperately to the reset policy. But I agree with the reset flag being private. :)

tbordaz:
Thanks @Firstyear for your reviews. I updated the patch to make all pwdTPR* attribute modifiable (except pwdTPRReset). The patch is missing "valid' OIDs that I am struggling to obtain !

I will update the design to take into account the bulk reset you described.

Firstyear:
> Thanks @Firstyear for your reviews. I updated the patch to make all pwdTPR* attribute modifiable (except pwdTPRReset). The patch is missing "valid' OIDs that I am struggling to obtain !

Why struggling to obtain them? I was the owner of RHANA of OID's when I was there, shouldn't someone else in the team have taken over this when I "left"? 

Worst case, I'm the owner of OID's at SUSE now so I can allocate you the values required if needed. 

Alternately, we could just get an OID for 389-ds from IANA

> 
> I will update the design to take into account the bulk reset you described.

Thank you!!!!

tbordaz:
I finally managed to book the OIDs \o/ (Upgrading to F33 changed the crypto policy and my ssh key was failing)

Issue #4678, opened 2021-03-18
RFE automatique disable of virtual attribute checking


tbordaz:
**Is your feature request related to a problem? Please describe.**
During a SRCH, lookup for virtual attribute providers is done during evaluating a filter (against a candidate entry) and also for each returned attribute. The lookup of the virtual attribute hashtable is protected with a rwlock that creates contention, even if it is almost always free (in read). In addition the hashtable lookup is costly even if the search attribute is not present in the hashtable.

This was already detected to be an expensive component and a config attribute 'nsslapd-ignore-virtual-attrs' allows to skip virtual attribute lookup (see https://github.com/389ds/389-ds-base/issues/511).

This attribute is 'off' by default, so most of the deployments, even if they do not use virtual attributes, are paying the cost of virtual attribute

**Describe the solution you'd like**
A first possibility is to set 'nsslapd-ignore-virtual-attrs: on' when there is no virtual attributes.
389-ds server defines virtual attributes with 'cos' and 'role' plugins. If there is cos/role definitions and if there is no custom plugin, then set  'nsslapd-ignore-virtual-attrs: on'.
This should be detected at startup and also requires cos/role plugin to enable the flag if a new definition comes in.

A second possibility is too have healthcheck, detecting the same (no cos/role definitions and no custom plugin)

A third possibility is to set 'nsslapd-ignore-virtual-attrs: on' by default and turn it on when a virtual attribute is inserted (other than 'dummyAttr').

**Describe alternatives you've considered**

**Additional context**

Firstyear:
I think a better option here is that we should set `nsslapd-ignore-virtual-attrs: on` by default, and then at plugin init from cos/role, we enable it internally by writing to the config.

tbordaz:
I am not sure to understand your proposal. ATM cos/roles are enabled by default. Do you suggest to disable them by default ?
Note that even if cos/roles are enabled, it does not mean it exists definitions for virtual attributes.

Firstyear:
Sorry I think my proposal is that roles/cos are enabled by default. By default we set nsslapd-ignore-virtual-attrs: off.

Then if roles / cos have any cos template or roles defined, they internally switch nsslapd-ignore-virtual-attrs: on. 

So that way roles/cos can be "enabled" but until any role or template is defined we don't enable virtual attrs. This way, an administrator doesn't have to change "another setting" to make roles/templates work.

tbordaz:
[...]

Issue #4667, opened 2021-03-10
incorrect accounting of readers in vattr rwlock


tbordaz:
**Issue Description**
Fedora 33 389-ds 1.4.4.13-2, a [prci test](https://github.com/freeipa/freeipa-pr-ci/issues/422#issuecomment-793504398) is showing a DS hang.

The hangs occurs when a thread (cos cache rebuild) tries to acquire vattr rwlock (the_map->lock) in write. As it remains readers, this thread is stopped but because of priority of writers it blocks others SRCH threads that try to acquire the rwlock in read.

The hang should finished when the readers threads that acquired the lock before writer thread (cos cache) release the lock.
The problem is that there is no others readers threads. The backtrace is only showing readers that are waiting for the writers.

The backtrace is showing 5 readers and 1 writers but the lock is showing 14 readers. So some readers, that complete their task, has not released the lock

<pre>
(gdb) print *the_map->lock
$41 = {__data = {__readers = 14, __writers = 0, __wrphase_futex = 2, __writers_futex = 1, __pad3 = 0, __pad4 = 0, __cur_writer = 0, __shared = 0, __rwelision = 0 '\000',
    __pad1 = "\000\000\000\000\000\000", __pad2 = 0, __flags = 2},
  __size = "\016\000\000\000\000\000\000\000\002\000\000\000\001", '\000' , "\002\000\000\000\000\000\000", __align = 14}

</pre>

Related ticket:
#[51068](https://pagure.io/389-ds-base/issue/51068) That gives priority to writers and hang later readers
#[49873](https://pagure.io/389-ds-base/issue/49873) That acquires the map rwlock at the operation level using a per cpu variable

**Package Version and Platform:**
 I suspect it happens since #49873 (RHEL8)

**Steps to Reproduce**
The hang is **not** systematic. It happens from time to time with freeipa  [prci test](https://github.com/freeipa/freeipa-pr-ci/issues/422#issuecomment-793504398)

**Expected results**
DS should not hang

tbordaz:
The PRCI tests are not failing systematically. A first estimate is that it occurs 1 time out of 10 runs. (TBC)

tbordaz:
This bug seems to be reproducible 1% time via a simpler testcase log in prci

tbordaz:
e501b83aa..3250a3e43  master
58dbf084a..6446fe27a  389-ds-base-1.4.4
2e5b52601..8fcf5dd37  389-ds-base-1.4.3

PR #4732, opened 2021-04-21
Issue 4667 - incorrect accounting of readers in vattr rwlock

### LINKED ISSUE #2932: Contention on virtual attribute lookup
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/49873
- Created at 2018-07-30 14:30:23 by [tbordaz](https://pagure.io/user/tbordaz) (@tbordaz)
- Closed at 2019-03-27 10:37:25 as fixed
- Assigned to [tbordaz](https://pagure.io/user/tbordaz) (@tbordaz)
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=1627846

---

#### Issue Description
virtual attribute lookup can be slow because of slowness of RW lock acquisition when there are many threads competing for the lock.
This was reported in https://pagure.io/389-ds-base/issue/511.  The fix was to disable virtual attribute checking when evaluating a filter.

The RC of the slowness is https://bugzilla.redhat.com/show_bug.cgi?id=200247, but this bug is not a priority.

Rather than disabling vattr lookup, this ticket is to evaluate the possibility to acquire the RW lock once during the operation. vattr are rarely updated so an option would be to acquire the lock in SRCH op (from pre-to-post op) and set a flag in percpu.

#### Package Version and Platform
All version

#### Steps to reproduce

Using IPA deployement, many groups, users/hosts and do a sssd fetch (no cache)

#### Actual results
Many threads contention

#### Expected results

No threads in contention

tbordaz:
Bug description:
	The fix #2932 (Contention on virtual attribute lookup) reduced
	contention on vattr acquiring vattr lock at the operation
	level rather than at the attribute level (filter and
        returned attr).
        The fix #2932 is invalid. it can lead to deadlock scenario
	(3 threads). A vattr writer (new cos/schema) blocks
        an update thread that hold DB pages and later needs vattr.
	Then if a reader (holding vattr) blocks vattr writer and later
        needs the same DB pages, there is a deadlock.
	The decisions are:
		- revert #2932 (this issue)
		- Skip contention if deployement has no vattr #4678
		- reduce contention with new approaches
                  (COW and/or cache vattr struct in each thread)
		  no issue opened

Fix description:
	The fix reverts #2932

relates: https://github.com/389ds/389-ds-base/issues/4667

Reviewed by:

Platforms tested:  F33

### DIFF
--- a/ldap/servers/slapd/detach.c
+++ b/ldap/servers/slapd/detach.c
@@ -145,10 +145,6 @@ detach(int slapd_exemode, int importexport_encrypt, int s_port, daemon_ports_t *
             }
             break;
         }
-        /* The thread private counter needs to be allocated after the fork
-         * it is not inherited from parent process
-         */
-        vattr_global_lock_create();
 
         /* call this right after the fork, but before closing stdin */
         if (slapd_do_all_nss_ssl_init(slapd_exemode, importexport_encrypt, s_port, ports_info)) {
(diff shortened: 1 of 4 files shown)


Firstyear:
Revert seems okay to me.

droideck (approved):
LGTM too!

Issue #4711, opened 2021-04-01
SIGSEV with sync_repl


tbordaz:
**Issue Description**
When running a sync_repl client, it crashes current master
<pre>
=================================================================
==2268384==ERROR: AddressSanitizer: SEGV on unknown address 0x000041b58abb (pc 0x7f2c75c6c68d bp 0x7f2be7fc3d60 sp 0x7f2be7fc3d40 T12)
==2268384==The signal is caused by a READ memory access.
    #0 0x7f2c75c6c68c in slapi_value_get_string ldap/servers/slapd/value.c:371
    #1 0x7f2c71534436 in sync_create_state_control ldap/servers/plugins/sync/sync_util.c:194
    #2 0x7f2c7153a89f in sync_srch_refresh_pre_entry ldap/servers/plugins/sync/sync_refresh.c:275
    #3 0x7f2c75bb7365 in plugin_call_func ldap/servers/slapd/plugin.c:2002
    #4 0x7f2c75bb6fb6 in plugin_call_list ldap/servers/slapd/plugin.c:1945
    #5 0x7f2c75baf71c in plugin_call_plugins ldap/servers/slapd/plugin.c:414
    #6 0x7f2c75bf9a6a in send_ldap_search_entry_ext ldap/servers/slapd/result.c:1488
    #7 0x7f2c75bf75e6 in send_ldap_search_entry ldap/servers/slapd/result.c:1050
    #8 0x7f2c75b82664 in send_entry ldap/servers/slapd/opshared.c:1140
    #9 0x7f2c75b83687 in iterate ldap/servers/slapd/opshared.c:1326
    #10 0x7f2c75b8493b in send_results_ext ldap/servers/slapd/opshared.c:1543
    #11 0x7f2c75b7fffe in op_shared_search ldap/servers/slapd/opshared.c:882
    #12 0x4700b1 in do_search ldap/servers/slapd/search.c:388
    #13 0x4243c7 in connection_dispatch_operation ldap/servers/slapd/connection.c:659
    #14 0x42a5f5 in connection_threadmain ldap/servers/slapd/connection.c:1777
    #15 0x7f2c75549b33  (/lib64/libnspr4.so+0x2bb33)
    #16 0x7f2c754de4e1 in start_thread (/lib64/libpthread.so.0+0x94e1)
    #17 0x7f2c753966a2 in clone (/lib64/libc.so.6+0x1016a2)

AddressSanitizer can not provide additional info.
[...]

PR #4738, opened 2021-04-26
Issue 4711 - SIGSEV with sync_repl

tbordaz:
Bug description:
	sync_repl sends back entries identified with a unique
	identifier that is 'nsuniqueid'. If 'nsuniqueid' is
	missing, then it may crash

Fix description:
	Check a nsuniqueid is available else returns OP_ERR

relates: https://github.com/389ds/389-ds-base/issues/4711

Reviewed by: Pierre Rogier, James Chapman

Platforms tested:  F33

### DIFF
--- a/ldap/servers/plugins/sync/sync_util.c
+++ b/ldap/servers/plugins/sync/sync_util.c
@@ -165,8 +165,8 @@ sync_create_state_control(Slapi_Entry *e, LDAPControl **ctrlp, int type, Sync_Co
     BerElement *ber;
     struct berval *bvp;
     char *uuid;
-    Slapi_Attr *attr;
-    Slapi_Value *val;
+    Slapi_Attr *attr = NULL;
+    Slapi_Value *val = NULL;
 
     if (type == LDAP_SYNC_NONE || ctrlp == NULL || (ber = der_alloc()) == NULL) {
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/sync/sync_util.c
```diff
@@ -191,6 +191,14 @@ sync_create_state_control(Slapi_Entry *e, LDAPControl **ctrlp, int type, Sync_Co
     } else {
         slapi_entry_attr_find(e, SLAPI_ATTR_UNIQUEID, &attr);
         slapi_attr_first_value(attr, &val);
+        if ((attr == NULL) || (val == NULL)) {
```
Firstyear, line 194: If attr is some here do we need to free it? Or is it owned by the entry?

Firstyear (approved):
Looks good mate, just want to check about a free case but otherwise ack.

ldap/servers/plugins/sync/sync_util.c
tbordaz: Thanks @Firstyear for your review. The attribute 'attr' or value 'val' are references into the entry 'e' they should not be freed here, later the operation cleanup will free them.

Issue #4742, opened 2021-04-29
UI should always use LDAPI path when calling CLI 


mreynolds389:
**Issue Description**

In some places in the UI code we call dsconf like:    

    dsconf -j slapd-instance ...

Instead of:

    dsconf -j "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket"

The problem is that if you setup the ".dsrc" file to use something other than LDAPI then the UI hangs.  We need to always call the CLI using the LDAP socket.

mreynolds389:
6446fe27a..4ba0f07d4  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
8fcf5dd37..34dd83e89  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4743, opened 2021-04-29
Issue 4742 - UI - should always use LDAPI path when calling dsconf

mreynolds389:
Bug Description:

In some places in the UI code we call dsconf like:

dsconf -j slapd-instance ...

Instead of:

dsconf -j "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket"

The problem is that if you setup the ".dsrc" file to use something other than LDAPI then the UI hangs.

Fix Description:

We need to always call dsconf using the LDAPI socket.

Relates: https://github.com/389ds/389-ds-base/issues/4742


droideck (approved):
LGTM!

Issue #4747, opened 2021-04-30
Remove unstable/unstatus tests from PRCI


tbordaz:
**Issue Description**
PRCI runs several suites. 9 of them are failing because of ~20 testcases failing.
So PRCI fails even if a **new** regression is not detected.
It is time consuming to make a clear status of those testcases (bug in the testcase or in DS core) and until all status are done the added-value of PRCI is reduced.

**Package Version and Platform:**
 master

**Steps to Reproduce**
Do a dummy fix, create a PR, monitor PRCI result

**Expected results**
If there is no regression, PRCI should be 100%

tbordaz:
ad3e77d71..8006632e7  master

mreynolds389:
49a5f1c17..66bbfee82  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
907ced2b7..45b370d5c  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

tbordaz:
Missing "from lib389.utils import ds_is_older" in 1.4.3
8aa5ef7dd..c615aef66  389-ds-base-1.4.3

tbordaz:
f542f8902..ff977cc87  master
a16a65325..e4612cd47  389-ds-base-1.4.4
d5896fb43..f8f616b6a  389-ds-base-1.4.3

PR #4748, opened 2021-04-30
Issue 4747 - Remove unstable/unstatus tests from PRCI

tbordaz:
Bug description:
	Some tests are failing although there is no regression.
	Those tests may hide a real regression and reduce
	the added value of PRCI

Fix description:
	Flag failing tests with @pytest.mark.xfail(raises=RuntimeError)

relates: https://github.com/389ds/389-ds-base/issues/4747

Reviewed by:

Platforms tested: F33

### DIFF
--- a/dirsrvtests/tests/suites/acl/keywords_test.py
+++ b/dirsrvtests/tests/suites/acl/keywords_test.py
@@ -223,7 +223,8 @@ def test_user_binds_without_any_password_and_cannot_access_the_data(topo, add_us
     with pytest.raises(ldap.INSUFFICIENT_ACCESS):
         org.replace("seeAlso", "cn=1")
 
-
+#unstable or unstatus tests, skipped for now
+@pytest.mark.flaky(max_runs=2, min_passes=1)
 def test_user_can_access_the_data_when_connecting_from_any_machine(
         topo, add_user, aci_of_user
 ):
@@ -253,6 +254,8 @@ def test_user_can_access_the_data_when_connecting_from_any_machine(
     OrganizationalUnit(conn, DNS_OU_KEY).replace("seeAlso", "cn=1")
 
 
+#unstable or unstatus tests, skipped for now
+@pytest.mark.flaky(max_runs=2, min_passes=1)
 def test_user_can_access_the_data_when_connecting_from_internal_ds_network_only(
         topo, add_user, aci_of_user
 ):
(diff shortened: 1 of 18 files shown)


droideck (changes requested):
There are 3 more failures for some reason... Could you please check it?

dirsrvtests/tests/suites/acl/keywords_test.py
```diff
@@ -223,7 +223,8 @@ def test_user_binds_without_any_password_and_cannot_access_the_data(topo, add_us
     with pytest.raises(ldap.INSUFFICIENT_ACCESS):
         org.replace("seeAlso", "cn=1")
 
-
+#unstable or unstatus tests, skipped for now
+@pytest.mark.xfail(raises=ldap.INSUFFICIENT_ACCESS)
```
vashirov: When a test marked with `xfail` passes, it is marked as `xpass` (unexpectedly passing). `xpass` is interpreted by some systems as a failure (because it's unexpected and needs investigation).
Instead of `xfail` I suggest to mark these tests as `@pytest.mark.flaky`. This way they can be skipped in PR CI by excluding all tests with this mark (`pytest -m "not flaky"`). 
Additionally we can take advantage of [pytest flaky plugin](https://github.com/box/flaky), that can automatically rerun flaky tests on failure.

vashirov: PR CI pytest command can be extended with the mark selector in https://github.com/389ds/389-ds-base/blob/4559a89c0877e45143fa3c9cc34393d4d0cd2696/.github/workflows/pytest.yml#L68

vashirov: Also, looking at the failing tests, they are mostly related to IPv6 connectivity. I will investigate what configuration changes we need to do in GH Actions.

dirsrvtests/tests/suites/acl/keywords_test.py
```diff
@@ -223,7 +224,9 @@ def test_user_binds_without_any_password_and_cannot_access_the_data(topo, add_us
     with pytest.raises(ldap.INSUFFICIENT_ACCESS):
         org.replace("seeAlso", "cn=1")
 
-
+#unstable or unstatus tests, skipped for now
+@pytest.mark.flaky
```
vashirov: We should use only one flaky decorator here. 
`@flaky` doesn't add the mark and requires an explicit import of the `flaky` module.
`@pytest.mark.flaky` adds the mark (so it can be skipped), it also can accept parameters as `@flaky`, so when a flaky module is present, they will be used. 
If the flaky module is not present, they will be ignored.

vashirov (commented):
Please also update the commit message that we're using `flaky` instead of `xfail`.

tbordaz:
@vashirov , @droideck , this PR is now running quite stable 100% success. Is it okay to push it upstream ?

dirsrvtests/tests/suites/acl/keywords_test.py
tbordaz: Changes done

droideck (approved):
Looks good to me!
Ack if @vashirov is okay with it.

vashirov (approved):
LGTM, thank you!

PR #4754, opened 2021-05-05
Issue 4753 - Adjust our tests to 389-ds-base-snmp missing in RHEL 9 A…

### LINKED ISSUE #4753: Adjust our tests to 389-ds-base-snmp missing in RHEL 9 Appstream
Description:
With RHEL 9, 389-ds-base-snmp is no longer delivered in AppStream.
We need to adapt our tests which rely on 389-ds-base-snmp, so that they are skipped if it is missing.

Fixes: 

Reviewed by: ???

aadhikar:
…ppstream

Description: With RHEL 9, 389-ds-base-snmp is no longer delivered in AppStream.
We need to adapt our tests which rely on 389-ds-base-snmp, so that they are skipped if it is missing.

Fix Description: Added skipif to tests which rely on 389-ds-base-snmp

Fixes: https://github.com/389ds/389-ds-base/issues/4753

Reviewed by: ???

### DIFF
--- a/src/lib389/lib389/utils.py
+++ b/src/lib389/lib389/utils.py
@@ -1431,12 +1431,3 @@ def is_valid_hostname(hostname):
         hostname = hostname[:-1] # strip exactly one dot from the right, if present
     allowed = re.compile("(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
     return all(allowed.match(x) for x in hostname.split("."))
-
-
-def snmp_present():
-    rpm_handle = os.popen('rpm -qa --qf "%{NAME}\n"')
-    rpm_list = rpm_handle.read().splitlines()
-    if '389-ds-base-snmp' in rpm_list:
-        return True
-    else:
-        return False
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/utils.py
```diff
@@ -1431,3 +1431,12 @@ def is_valid_hostname(hostname):
         hostname = hostname[:-1] # strip exactly one dot from the right, if present
     allowed = re.compile("(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
     return all(allowed.match(x) for x in hostname.split("."))
+
+
+def snmp_present():
```
vashirov: I don't think this should be in lib389.
We have `get_rpm_version()` in `conftest.py`: https://github.com/389ds/389-ds-base/blob/a79433498709319814b7dae93c3dec9a0ed32c5c/dirsrvtests/conftest.py#L27-L34
It also won't fail on the systems where `rpm` is not present.

sgouvern (approved):
Minor remark : the reason why the test is skipped ("not present") does not appear in the test output.
Apart from that, ok for me

sgouvern (commented):
My bad, indeed the reason is displayed : (Not present). 
But could you please elaborate a bit more so that the reason for skipping appears clearly, like '389-ds-base-snmp package is not present' ?

aadhikar:
@sgouvern sure that could be added, Thanks :)

sgouvern (commented):
It's nitpicking, but the reason should be "389-ds-base-snmp package is ***not*** present" :)

aadhikar:
@sgouvern my bad It's corrected now :)

sgouvern (approved):
Thanks :)
LGTM !

PR #4755, opened 2021-05-05
Issue 4575 Update test docstrings metadata

aadhikar:
Description: Mapping all the test cases to the requirements.

Fix Description: Adding __init__.py file and docstrings to all the test suites

Fixes: https://github.com/389ds/389-ds-base/pull/4754

Reviewed by: @vashirov

### DIFF
--- a/dirsrvtests/tests/suites/state/__init__.py
+++ b/dirsrvtests/tests/suites/state/__init__.py
@@ -1,0 +1,3 @@
+"""
+   :Requirement: 389-ds-base: 389-ds-base: Operational Attributes
+"""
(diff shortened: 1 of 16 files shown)


dirsrvtests/tests/suites/state/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+   :Requirement: 389-ds-base: Replication
```
aadhikar: Adding this under replication as not sure should I create a new one or not

sgouvern: I think the requirement here could be 'operational attributes'

sgouvern (commented):
It seems ok to me
LGTM !

sgouvern (approved):

Issue #4759, opened 2021-05-06
Fix coverity issue


tbordaz:
**Issue Description**
The support of wtime/optime in the access logs, record the hrtime in the operation structure. #/4218 assign the time without checking that operation exists. I triggers failure in coverity scan

**Package Version and Platform:**
since 1.4.3

**Steps to Reproduce**
Run covcan

**Expected results**
No error in opshared.c and extended.c regarding call to slapi_operation_set_time_started

droideck:
@tbordaz can we close it as the PR was merged? Or do you want to leave the issue for future use?

tbordaz:
Opps thanks for the heads up @droideck .

I forgot to add the commit hashes. This bug being fixed in all targeted releases close
9b62aede5..ad3e77d71 master
9b62aede5..ad3e77d71 389-ds-base-2.0
e0cf8a9cb..70c3e9e71 389-ds-base-1.4.4
d0e0ab37a..099c7c27f 389-ds-base-1.4.3
ca3776e63..8b236bc2c 389-ds-base-1.3.10

PR #4760, opened 2021-05-06
Issue 4759 - Fix coverity issue

tbordaz:
Bug description:
	with #4218 (wtime, optime in access log), hrtime is set in the
	operation. But it is done before checking if the operation is
	set. covscan fails

Fix description:
	move the setting after verification that operation != NULL

relates: https://github.com/389ds/389-ds-base/issues/4759

Reviewed by:

Platforms tested:


droideck (approved):
LGTM!

tbordaz:
9b62aede5..ad3e77d71  master

Need backport all the way down to 1.3.10

PR #4762, opened 2021-05-07
Issue 4623 - RFE - Monitor the current DB locks

### LINKED ISSUE #4623: RFE - Monitor the current DB locks ( nsslapd-db-current-locks )
db lock gets exhausted because of unindexed internal searches (under a transaction). Indexing those searches is the way to prevent exhaustion.

To prevent db lock exhaustion and help admin task a possible solutions would be:

- If db lock get exhausted during a txn, it leads to db panic and the later recovery can possibly fail. That leads to a full reinit of the instance where the db locks got exhausted. The server should monitor the db lock and trigger server shutdown (similar to disk full) if the db lock is close to be exhausted. Because of the performance impact, the monitoring should be limited to unindexed (allid(candidate)) internal searches (under a txn) and periodically (after each 1000 evaluated candidate). unindexed should be flagged in ldbm_back_search. transaction can be tested with pblock(SLAPI_TXN), monitoring should be done in iterate, internal op is an operation flag OP_FLAG_INTERNAL.

- To help indexing the appropriate attributes, unindexed internal search (under txn) should log a warning with the search filter.

- a config parameter should toggle monitoring/shutdown. By default it should be enabled. 

- Monitoring returns value that may be not exact. The threshold to trigger the shutdown should take into account that the value is not perfect.

droideck:
Description: DB lock gets exhausted because of unindexed internal searches
(under a transaction). Indexing those searches is the way to prevent exhaustion.
If db lock get exhausted during a txn, it leads to db panic and the later recovery
can possibly fail. That leads to a full reinit of the instance where the db locks
got exhausted.

Add two attributes to global BDB config: "nsslapd-db-locks-monitoring" and
"nsslapd-db-locks-monitoring-threshold". By default, nsslapd-db-locks-monitoring
is turned on and nsslapd-db-locks-monitoring-threshold is set to 95.

When current locks are close to maximum locks value on 95% - returning
the next candidate will fail until the maximum of locks won't be
increased or current locks are released.

Add the setting to UI and CLI tools.

Fixes: https://github.com/389ds/389-ds-base/issues/4623

Reviewed by: ?


droideck:
The design is ready for review here:
https://github.com/mreynolds389/389wiki/pull/56

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1515,6 +1517,17 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
 
     /* Find the next candidate entry and return it. */
     while (1) {
+        if (li->li_dblock_monitoring && li->li_dblock_threshold_reached) {
```
Firstyear: This isn't thread safe - these are int values and because they are traversing thread boundaries, search threads have no guarantee that they will see these changes from the lock monitor thread. These should be changed to be atomics (ordering relaxed) to ensure the values are syncing.

Firstyear (changes requested):

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -573,12 +573,20 @@ struct ldbminfo
     char *li_backend_implement;          /* low layer backend implementation */
     int li_noparentcheck;                /* check if parent exists on add */
 
-    /* the next 3 fields are for the params that don't get changed until
+    /* db lock monitoring */
+    int li_dblock_monitoring;                 /* enables db locks monitoring thread - requires restart  */
```
tbordaz: Why not int32_t ?

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -573,12 +573,20 @@ struct ldbminfo
...
+    uint32_t li_dblock_threshold;  /* specified in percentage - when reached, aborts the search in ldbm_back_next_search_entry */
+    uint32_t li_dblock_threshold_reached;
+
+    /* the next 4 fields are for the params that don't get changed until
      * the server is restarted (used by the admin console)
      */
     char *li_new_directory;
     uint64_t li_new_dbcachesize;
     int li_new_dblock;
+    int li_new_dblock_monitoring;
```
tbordaz: int32_t ?

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -573,12 +573,20 @@ struct ldbminfo
     char *li_backend_implement;          /* low layer backend implementation */
     int li_noparentcheck;                /* check if parent exists on add */
```
tbordaz, line 568: My understanding is that dblock monitoring will remain specific to bdb. Would it make sense to define those fields in bdb_config rather than in generic ldbminfo struct ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
```diff
@@ -228,6 +228,103 @@ bdb_config_db_lock_set(void *arg, void *value, char *errorbuf, int phase, int ap
...
+
+static void *
+bdb_config_db_lock_pause_get(void *arg)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+
+    return (void *)((uintptr_t)(slapi_atomic_load_32(&(li->li_dblock_monitoring_pause), __ATOMIC_RELAXED)));
+}
+
+#define DBLOCK_MONITORING_PAUSE_DEFAULT 100
```
tbordaz: Just a comment saying that 100 = 1sec

ldap/servers/slapd/back-ldbm/ldbm_config.h
```diff
@@ -104,6 +104,9 @@ struct config_info
 #define CONFIG_DB_VERBOSE "nsslapd-db-verbose"
 #define CONFIG_DB_DEBUG "nsslapd-db-debug"
 #define CONFIG_DB_LOCK "nsslapd-db-locks"
+#define CONFIG_DB_LOCKS_MONITORING "nsslapd-db-locks-monitoring"
```
tbordaz: Being a on/off toggle it could also be nsslapd-db-locks-monitoring-enable ?

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1515,6 +1517,18 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
...
+        if (li->li_dblock_monitoring &&
+            slapi_atomic_load_32(&(li->li_dblock_threshold_reached), __ATOMIC_RELAXED)) {
+            errmsg = PR_smprintf("DB locks threshold is reached (nsslapd-db-locks-monitoring-threshold "
+                                 "under cn=bdb,cn=config,cn=ldbm database,cn=plugins,cn=config). "
+                                 "Please, increase nsslapd-db-locks according to your needs.");
+            slapi_log_err(SLAPI_LOG_CRIT, "ldbm_back_next_search_entry", "%s\n", errmsg);
+            slapi_pblock_set(pb, SLAPI_SEARCH_RESULT_ENTRY, NULL);
+            delete_search_result_set(pb, &sr);
+            rc = SLAPI_FAIL_GENERAL;
+            slapi_send_ldap_result(pb, LDAP_UNWILLING_TO_PERFORM, NULL, errmsg, 0, NULL);
```
tbordaz: I would not send this detailed error message to the ldapclient. rather something like "Transient DB ressource critical level".

tbordaz (commented):
Some (late) minor comments/remarks

ldap/servers/slapd/back-ldbm/back-ldbm.h
droideck: Maybe I miss something but I need to check these variables in the ldbm_search.c - **li_dblock_monitoring, li_dblock_monitoring_pause, li_dblock_threshold_reached.** 

All is left is **li_dblock_threshold** so I've just left it here for consistency.

I see that we do the same for other bdb settings like **li_dblock** and others.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
droideck: Hm, but I thought **PR_MillisecondsToInterval** accepts milliseconds (0.001 of a second) and converts it to Interval Time.
Am I confused? And do I miss something obvious?..
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Reference/NSPR_functions#interval_time

ldap/servers/slapd/back-ldbm/ldbm_config.h
droideck: I've created it this way because we already have **nsslapd-disk-monitoring** with the same functionality so I've decided to keep it consistent.
I don't have a strong opinion on this so if you think it's better as **nsslapd-db-locks-monitoring-enable**, - I can change it.

ldap/servers/slapd/back-ldbm/back-ldbm.h
tbordaz: That is right, some fields that are specific to bdb_config are still in ldbminfo. The move could be easy for some of them (like li_dblock or li_dbcachesize I guess) but it may be more complex for others like li_directory).
Now as we are introducing new fields we can put them directly in bdb_config

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
tbordaz: Sorry I was first wrong (assuming 1sec) and also not clear. if '100' stands for 100ms then adding a comment helps to know what this value is for: /* by default pausing 100ms between each db_stat */.

Is 100ms not to intensive to gather statistics by default ?

ldap/servers/slapd/back-ldbm/ldbm_config.h
tbordaz: I tend to prefer nsslapd-db-locks-monitoring-enable, but it also depends the amount of work (code, test, design) that costs. If it is painful then ignore my (late) comment

src/lib389/lib389/cli_conf/backend.py
```diff
@@ -993,6 +996,13 @@ def create_parser(subparsers):
...
     set_db_config_parser.add_argument('--logbufsize', help='Specifies the transaction log information buffer size')
     set_db_config_parser.add_argument('--locks', help='Sets the maximum number of database locks')
+    set_db_config_parser.add_argument('--locks-monitoring', help='Set to "on" or "off" to monitor DB locks. When it crosses the percentage value '
+                                                                 'set with "--locks-monitoring-threshold" ("on" by default)')
+    set_db_config_parser.add_argument('--locks-monitoring-threshold', help='Sets the DB lock exhaustion value in percentage. If too many locks are '
+                                                                           'acquired, the server will abort the searches while the number of locks '
+                                                                           'are not decreased. It helps to avoid DB corruption and long recovery.')
+    set_db_config_parser.add_argument('--locks-monitoring-pause', help='Sets the DB lock monitoring value for the amount of time that the monitoring'
```
jchapma: For the sake of clarity it might it be useful to add the time unit here.

ldap/servers/slapd/back-ldbm/ldbm_search.c
jchapma: In my opinion a concise error message would work, something like.  "DB locks threshold is reached (nsslapd-db-locks-monitoring-threshold)" Then the sysadmin knows what has happened and can find out how to fix it.

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1515,6 +1517,18 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
 
     /* Find the next candidate entry and return it. */
     while (1) {
+        if (li->li_dblock_monitoring &&
+            slapi_atomic_load_32((int32_t *)&(li->li_dblock_threshold_reached), __ATOMIC_RELAXED)) {
+            errmsg = PR_smprintf("DB locks threshold is reached (nsslapd-db-locks-monitoring-threshold "
```
jchapma: From what I understand PR_smprintf might be an expensive function for this part of the code, allocating memory, formatting stuff. Could we use something like slapi_send_ldap_result(.. LDAP_DBLOCKS_EXCEEDED...)  instead.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -2988,6 +2997,82 @@ perf_threadmain(void *param)
...
 
+
+/*
+ * create a thread for locks_monitoring_threadmain
+ */
+static int
+bdb_start_locks_monitoring_thread(struct ldbminfo *li)
+{
+    int return_value = 0;
+    if ((li->li_dblock_monitoring) && (li->li_dblock_threshold)) {
```
jchapma: Could it happen that there is no dblock threshold set but dblock monitoring is enabled, if it did then the dblock monitoring thread wouldn't start.

jchapma (commented):
This is a really nice feature to have, just some minor comments from me.

ldap/servers/slapd/back-ldbm/back-ldbm.h
droideck: As discussed on IRC and agreed, we won't move the fields to bdb_config as it may make code more obscure and fragile (as bdb_config is a thing in itself and it requires deeper integration). As the result, we'll need much more testing and it will delay the quality.
I've mentioned the additional information in the comments in the code as discussed on IRC.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
droideck: I agree, I've changed the default to 1000ms.

ldap/servers/slapd/back-ldbm/ldbm_config.h
droideck: Sure, sounds good!

ldap/servers/slapd/back-ldbm/ldbm_search.c
droideck: Good idea, I'll add it with my next commit.

ldap/servers/slapd/back-ldbm/ldbm_search.c
droideck: Good catch! Thanks!

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
droideck: It is intentional. If the threshold is empty or 0, it won't start as we need both attributes for the monitoring to function.

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -226,6 +254,41 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                <Form isHorizontal>
+                    <FormGroup
+                        label="DB Locks Threshold Percentage"
+                        fieldId="monitoringthreshold"
+                        title="Sets the DB lock exhaustion value in percentage. If too many locks are acquired, the server will abort the searches while the number of locks are not decreased. It helps to avoid DB corruption and long recovery. (nsslapd-db-locks-monitoring-threshold)"
+                    >
+                        <TextInput
+                            id="dblocksMonitoringThreshold"
+                            name="dblocksMonitoringThreshold"
+                            type="text"
```
mreynolds389: Should be type "number"

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -226,6 +254,41 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                    </FormGroup>
+                    <FormGroup
+                        label="DB Locks Pause Time"
+                        fieldId="monitoringpause"
+                        title="Sets the DB lock monitoring value for the amount of time that the monitoring thread spends waiting between checks. (nsslapd-db-locks-monitoring-pause)"
+                    >
+                        <TextInput
+                            id="dblocksMonitoringPause"
+                            name="dblocksMonitoringPause"
+                            type="text"
```
mreynolds389: Should also be type "number".  Might want to add min and max values too...

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -528,7 +577,40 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                                    id="dblocks"
+                                    aria-describedby="dblocks"
+                                    name="dblocks"
+                                    onChange={this.handleChange}
+                                />
+                            </FormGroup>
                         </Form>
+                        <Grid className="ds-margin-top-xlg">
+                            <GridItem span={12}>
+                                <h4 className="ds-sub-header">DB Locks Monitoring</h4>
```
mreynolds389: This is too large.  Maybe h5?

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -528,7 +577,40 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                                    isChecked={this.state.dblocksMonitoring}
+                                    onChange={this.select_db_locks_monitoring}
+                                    aria-label="uncontrolled checkbox example"
+                                />
+                            </GridItem>
+                            <GridItem span={5}>
+                                {dblocksMonitor}
+                            </GridItem>
+                        </Grid>
+
```
mreynolds389: I think these settings should be under the check box, and not to the "right" of it.

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -528,7 +577,40 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                                    value={this.state.dblocks}
+                                    type="text"
+                                    id="dblocks"
+                                    aria-describedby="dblocks"
+                                    name="dblocks"
+                                    onChange={this.handleChange}
+                                />
+                            </FormGroup>
                         </Form>
+                        <Grid className="ds-margin-top-xlg">
```
mreynolds389: This should be indented like the rest of the advanced settings.  It's too far to the left so it is not aligned with the rest of the settings.

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
mreynolds389: and it should be indented under that checkbox...

Firstyear:
I'm happy with the atomic fixes :)

jchapma:
LGTM

jchapma (commented):
LGTM

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
```diff
@@ -229,6 +229,102 @@ bdb_config_db_lock_set(void *arg, void *value, char *errorbuf, int phase, int ap
...
+}
+
+static int
+bdb_config_db_lock_threshold_set(void *arg, void *value, char *errorbuf, int phase __attribute__((unused)), int apply)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+    int retval = LDAP_SUCCESS;
+    u_int32_t val = (u_int32_t)((uintptr_t)value);
+
+    if (val >= 100) {
```
tbordaz: Shouldn't we also check a minimal rate ?
For example if it is set to 10%, does it make sense ?. IMHO it could be [70%, 99%].
The impact of the threshold is significant (DS will very regularly interrupt some searches, then allow some others. ). So it should not be triggered too frequently.
For example, the risk does not exist to exhaust the locks with 20%.

tbordaz (commented):
A minor comment, else the fix LGTM

mreynolds389:
You patch does not apply on master branch, can you rebase?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
droideck: Done, please, check.

droideck:
> You patch does not apply on master branch, can you rebase?

I've rebased everything into one commit and onto master. Please, check.

mreynolds389:
Doesn't look like you changed anything in the UI in regards to my suggestions...

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -226,6 +255,45 @@ export class GlobalDatabaseConfig extends React.Component {
...
+                    >
+                        <TextInput
+                            id="dblocksMonitoringPause"
+                            name="dblocksMonitoringPause"
+                            type="number"
+                            aria-describedby="dblocksMonitoringPause"
+                            value={dblocksPause}
+                            onChange={this.handleChange}
+                            validated={parseInt(dblocksPause) < 1 ? ValidatedOptions.error : ValidatedOptions.default}
+                        />
```
droideck: @mreynolds389 I've 

- made both fields **type="number"** and
- added the validation - **validated={parseInt(dblocksThreshold) < 70 || parseInt(dblocksThreshold) > 95 ? ValidatedOptions.error : ValidatedOptions.default}**.

Did I miss your point somehow?..

mreynolds389:
These were comments 

Screenshot from 2021-05-19 10-26-48

droideck:
> These were my comments

Ahh, okay. It's a funny situation because... You haven't sent them.
If it's written as 'Pending' - it is not sent. You need to click the green button 'Review changes' and then 'Submit review'.

GitHub is a bit confusing here and, IMHO, it's not a very good design choice they've made. It's really not intuitive.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
tbordaz: Thanks @droideck . The RFE is so nice, customers will love it.

tbordaz (approved):
server code is looking good to me. Ack on this part

mreynolds389:
If you apply this diff I'll be happy :-) It improves the layout and it fixes a browser warning about nested Forms.

```
diff --git a/src/cockpit/389-console/src/lib/database/databaseConfig.jsx b/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
index c8c118587..042786442 100644
--- a/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
+++ b/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
@@ -260,8 +260,8 @@ export class GlobalDatabaseConfig extends React.Component {
         let dblocksPause = this.state.dblocksMonitoringPause;
 
         if (this.state.dblocksMonitoring) {
-            dblocksMonitor = <div className="ds-margin-left">
-                <Form isHorizontal>
+            dblocksMonitor =
+                <div className="ds-margin-left ds-margin-top">
                     <FormGroup
                         label="DB Locks Threshold Percentage"
                         fieldId="monitoringthreshold"
@@ -292,8 +292,7 @@ export class GlobalDatabaseConfig extends React.Component {
                             validated={parseInt(dblocksPause) < 1 ? ValidatedOptions.error : ValidatedOptions.default}
                         />
                     </FormGroup>
-                </Form>
-            </div>;
+                </div>;
         }
```

droideck:
> If you apply this diff I'll be happy :-) It improves the layout and it fixes a browser warning about nested Forms.

Good catch! Thanks!
Done.

PR #4766, opened 2021-05-12
Issue 4765 - database suffix unexpectdly changed from .db to .db4

### LINKED ISSUE #4765: database suffix unexpectdly changed from .db to .db4
This is a clone of redHat bugzilla 1947762 - IPA server does not start

The regression was caused by with Issue 4469 - Backend redesing phase 3a:
it is a side effect of moving the bdb dependencies out of the backend:
 db.h is no more included in back-ldbm.h 
so DB_VERSION_MAJOR is no more defined and 
#if 1000 * DB_VERSION_MAJOR + 100 * DB_VERSION_MINOR >= 5000 became false 
Should move that check in db-bdb plugin and define a callback to get the filename-suffix from the plugin

**Issue Description**
Berkeley db database suffix is .db4 instead of .db as it should.

 - Platform: Fedora
 - Package and version: [e.g. 389-ds-base-1.4.4.4-20200721git5d41dc5a4.fc32.x86_64]

**Steps to Reproduce**
1. Setup a directory instance
2. ls /var/lib/dirsrv/slapd-*/db/userroot/id2entry.db*
3. If suffix is .db4 instead of .db then the bug is not fixed.

progier389:
Description: when removing the bdb dependency from the backend
 include <db.h> was removed from back-ldbm.h 
 because of that the #if used to check bdb version and define the suffix become false and the db suffix reverted to old values
  so we got .db4 suffix instead of .db

The solution is:
   to move the bdb version test inside db-bdb plugin (where bdb version defines are available)
   to add a callback to get the db suffix from db-pdb plugin

Fixes: #4765

Reviewed by: @tbordaz @jchapma

### DIFF
--- a/ldap/servers/slapd/back-ldbm/vlv_srch.c
+++ b/ldap/servers/slapd/back-ldbm/vlv_srch.c
@@ -514,7 +514,6 @@ vlvIndex_init(struct vlvIndex *p, backend *be, struct vlvSearch *pSearch, const 
 {
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
     const char *file_suffix = dblayer_get_db_suffix(be);
-    dblayer_private *prv = NULL;
     char *filename = NULL;
 
     if (NULL == p)
--- a/ldap/servers/slapd/back-ldbm/vlv_srch.c
+++ b/ldap/servers/slapd/back-ldbm/vlv_srch.c
@@ -30,7 +30,6 @@ char *const type_vlvEnabled = "vlvEnabled";
 char *const type_vlvUses = "vlvUses";
 
 static const char *file_prefix = "vlv#"; /* '#' used to avoid collision with real attributes */
-static const char *file_suffix = LDBM_FILENAME_SUFFIX;
 
 static int vlvIndex_createfilename(struct vlvIndex *pIndex, char **ppc);
 
@@ -514,6 +513,8 @@ void
 vlvIndex_init(struct vlvIndex *p, backend *be, struct vlvSearch *pSearch, const Slapi_Entry *e)
 {
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    const char *file_suffix = dblayer_get_db_suffix(be);
+    dblayer_private *prv = NULL;
     char *filename = NULL;
 
     if (NULL == p)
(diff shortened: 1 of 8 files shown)


ldap/servers/slapd/back-ldbm/vlv_srch.c
```diff
@@ -514,6 +513,8 @@ void
 vlvIndex_init(struct vlvIndex *p, backend *be, struct vlvSearch *pSearch, const Slapi_Entry *e)
 {
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    const char *file_suffix = dblayer_get_db_suffix(be);
+    dblayer_private *prv = NULL;
```
jchapma: This variable is unused.

ldap/servers/slapd/back-ldbm/vlv_srch.c
```diff
@@ -514,6 +513,8 @@ void
 vlvIndex_init(struct vlvIndex *p, backend *be, struct vlvSearch *pSearch, const Slapi_Entry *e)
 {
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    const char *file_suffix = dblayer_get_db_suffix(be);
```
jchapma, line 516: Compiler complains  warning: initialization of ‘const char *’ from ‘int’ makes pointer from integer without a cast. I guess its in the case of the ternary operator returning NULL.

ldap/servers/slapd/back-ldbm/vlv_srch.c
progier389: Ok I fixed that one.

ldap/servers/slapd/back-ldbm/vlv_srch.c
progier389: Thanks James for having noticed the warning that I overlooked !
I am afraid that your guess is wrong: function prototype was missing because I forgot to commit proto-back-ldbm.h so function was assumed to return int ! -;)
  I just pushed a new commit to solve both issues.

tbordaz: @progier389 , sorry I can not see the new commit ? and where file_suffix != NULL is tested

tbordaz (approved):
LGTM

progier389:
Sorry Thierry, I have not seen your second comment. 
About the second commit,: in my browser the PR shows 2 commits, so I suspect that you hit a refresh issue.
About the fact that NULL is not tested: it is not a real issue because file_suffix is only used when trying to access a vlv index 
 which means that "be" is a regular db backend (and dblayer_get_db_suffix cannot return NULL in that case)

Issue #4379, opened 2020-10-15
allow more than 1 empty AttributeDescription for ldapsearch, without the risk of denial of service


tbordaz:
**Issue Description**
https://github.com/389ds/389-ds-base/issues/3028 enforce a strict limit of empty attributeDescription.
This is good but the limit is so low that some application may start failing
This bug is to extend a bit that limit so that application wont break

**Package Version and Platform:**
 - Platform:all 
 - Package and version: since https://github.com/389ds/389-ds-base/issues/3028  (1.4.1)

**Steps to Reproduce**

ldapsearch -LLL.... -b 'dc=example,dc=com' "" "" cn
Protocol error (2)

<pre>
[15/Oct/2020:15:45:00.555817663 +0200] conn=7 op=1 SRCH base="dc=example,dc=com" scope=2 filter="(objectClass=*)", invalid attribute request
</pre>

**Expected results**
It should succeed up to 10 empty attributes

tbordaz:
43c691569..b8b169142  master
311dab59a..4508d3c4d  389-ds-base-1.4.3
f39758817..ab00f9753  389-ds-base-1.4.2
22d67399b..9faf2d77a  389-ds-base-1.3.10

tbordaz:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1870624

tbordaz:
@aadhikar would you mind to have a look at the recent PR that possibly creates a regression

268d1c7ed Issue 4379 - Allow more than 1 empty AttributeDescription for ldapsearch, without the risk of denial of service

Test: tests/suites/acl/misc_test.py::test_info_disclosure

aadhikar:
@tbordaz by just looking at it I think it's failing because we changed the filter function to have one more argument that is `attrlist` which is the second argument so we need to explicitly specify `scope=`

tbordaz:
@aadhikar your eyes are much better than mine. Thanks for spotting this. I will test a fix.

tbordaz:
ff830604e..f53b28448  master 
268d1c7ed..ff830604e master
53f372cea..268d1c7ed master

PR #4767, opened 2021-05-12
Issue 4379 - Allow more than 1 empty AttributeDescription for ldapsea…

aadhikar:
…rch, without the risk of denial of service

Desciption: Added a test case to verify up to 10 empty values and a negetive
case to check max limit.

Relates: https://github.com/389ds/389-ds-base/issues/4379

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -1314,6 +1314,33 @@ def test_critical_msg_on_empty_range_idl(topology_st):
     # Step 5
     assert not topology_st.standalone.searchErrorsLog('CRIT - list_candidates - NULL idl was recieved from filter_candidates_ext.')
 
+@pytest.mark.bz1870624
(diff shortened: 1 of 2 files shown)


bsimonova (approved):
I see only typo in Description in commit message and "Should succeed" in expected results but otherwise LGTM!

bsimonova:
Ok, I see there is an issue in checks and I can confirm the test fails for me with ldap.SERVER_DOWN when it is ran together with other tests. It works individually. This should be investigated further.

bsimonova (changes requested):
Please investigate why the test fails when run together with other tests, maybe it needs to be in separate file.

aadhikar:
@bsimonova Thanks for your investigation, test_ldbm_modification_audit_log is making the test fail.

bsimonova (approved):
LGTM!

droideck (approved):
LGTM too!

Issue #4758, opened 2021-05-06
Add tests for WebUI


bsimonova:
I am creating this issue to keep track of commits for WebUI tests I am developing using Playwright for Python tool.

PR #4771, opened 2021-05-19
Issue 4758 - Add tests for WebUI

bsimonova:
Description:
Added some basic tests for WebUI (login, create instance etc.). These tests need
specific environment to be run so they will be skipped in normal conditions.
Playwright for Python was used to develop the tests and they pass in latest
RHDS version with Firefox and Chromium.

Relates: https://github.com/389ds/389-ds-base/issues/4758

Reviewed by: ???

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -66,7 +66,12 @@ jobs:
         set -x
         CID=$(sudo docker run -d -h server.example.com --privileged --rm --shm-size=4gb -v /sys/fs/cgroup:/sys/fs/cgroup:rw,rslave -v ${PWD}:/workspace quay.io/389ds/ci-images:test)
         sudo docker exec $CID sh -c "dnf install -y -v dist/rpms/*rpm"
-        sudo docker exec $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml --html=pytest.html -v dirsrvtests/tests/suites/${{ matrix.suite }}
+        export PASSWD=$(openssl rand -base64 32)
+        sudo docker exec $CID sh -c "echo \"${PASSWD}\" | passwd --stdin root"
+        sudo docker exec $CID sh -c "systemctl start dbus.service"
+        sudo docker exec $CID sh -c "systemctl enable --now cockpit.socket"
+        sudo docker exec -e WEBUI=1 -e DEBUG=pw:api -e PASSWD="${PASSWD}" $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml --html=pytest.html --browser=firefox --browser=chromium -v dirsrvtests/tests/suites/${{ matrix.suite }}
+
     - name: Make the results file readable by all
       if: always()
       run: |
--- a/dirsrvtests/tests/suites/webui/__init__.py
+++ b/dirsrvtests/tests/suites/webui/__init__.py
@@ -1,0 +1,110 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+
+import pytest
+import distro
+import os
+
+from lib389.utils import *
+from lib389.topologies import topology_st
+
+pytest.importorskip('playwright')
+
+RHEL = 'Red Hat Enterprise Linux'
+
+
+# in cockpit-250 some selectors got renamed so this function returns True if cockpit version is 250 or higher
+def check_cockpit_version_is_higher():
+    f = os.popen('rpm -q cockpit')
+    arq = f.readline()
+    version = arq.split("-")[1]
+
+    return int(version) >= 250
+
+
+# the iframe selection differs for chromium and firefox browser
+def determine_frame_selection(page, browser_name):
+    if browser_name == 'firefox':
+        frame = page.query_selector('iframe[name=\"cockpit1:localhost/389-console\"]').content_frame()
+    else:
+        frame = page.frame('cockpit1:localhost/389-console')
+
+    return frame
+
+
+# sometimes on a slow machine the iframe is not loaded yet, so we check for it until
+# it is available or the timeout is exhausted
+def check_frame_assignment(page, browser_name):
+    timeout = 80
+    count = 0
+    frame = determine_frame_selection(page, browser_name)
+
+    while (frame is None) and (count != timeout):
+        log.info('Waiting 0.5 seconds for iframe availability')
+        time.sleep(0.5)
+        count += 0.5
+        frame = determine_frame_selection(page, browser_name)
+
+    return frame
+
+
+def remove_instance_through_lib(topology):
+    log.info('Check and remove instance before starting tests')
+    if topology.standalone.exists():
+        topology.standalone.delete()
+
+
+def remove_instance_through_webui(topology, page):
+    frame = page.frame('cockpit1:localhost/389-console')
+
+    log.info('Check if instance exist')
+    if topology.standalone.exists():
+        log.info('Delete instance')
+        frame.wait_for_selector('#ds-action')
+        frame.click('#ds-action')
+        frame.click('#remove-ds')
+        frame.check('#modalChecked')
+        frame.click('//button[normalize-space(.)=\'Remove Instance\']')
+        frame.wait_for_selector("#no-inst-create-btn")
+        log.info('Instance deleted')
+
+
+def setup_login(page):
+    password = ensure_str(os.getenv('PASSWD'))
+    page.set_viewport_size({"width": 1920, "height": 1080})
+
+    # increase default timeout to wait enough time on a slow machine for selector availability
+    # (it will wait just enough time for the selector to be available,
+    # it won't stop for 80 000 miliseconds each time it is called)
+    page.set_default_timeout(65000)
+
+    page.goto("http://localhost:9090/")
+
+    # We are at login page
+    page.fill('#login-user-input', 'root')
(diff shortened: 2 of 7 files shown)


bsimonova:
I see a failure in checks because of missing playwright. Will fix this.

vashirov:
> I see a failure in checks because of missing playwright. Will fix this.

Right, test container image was missing playwright. I've just added it.
But please add a check to the test for this module and skip its execution if the module is not found.
Thanks.

vashirov:
@bsimonova, please also add `WEBUI=1` env variable for py.test here: https://github.com/389ds/389-ds-base/blob/f5b2cfb35a1686efa4ec03d61aaa1a22c2056ca4/.github/workflows/pytest.yml#L68
Thanks.

bsimonova:
I added importorskip for playwright and modified the pytest.yml.

bsimonova:
@vashirov Can you please advise me how to put WEBUI=1 in pytest.yml correctly there? It seems to break the checks as it is now.

vashirov:
Try to pass ` -e WEBUI=1` to `docker exec`, not to `py.test`, see https://docs.docker.com/engine/reference/commandline/exec/#run-docker-exec-on-a-running-container for the examples.

dirsrvtests/tests/suites/webui/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+def setup_login(page):
+    page.set_viewport_size({"width": 1920, "height": 1080})
+    page.goto("http://localhost:9090/")
+
+    # We are at login page
+    page.fill('#login-user-input', 'root')
+    page.fill('#login-password-input', 'redhat')
+    page.click("#login-button")
+    page.wait_for_selector('text=Red Hat Directory Server')
```
vashirov: I've added missing dependencies. Test now fails here. On Fedora it's "389 Directory Server".
It would be also nice to capture screenshot on failure, so it would be easier to investigate such failures. See example here https://github.com/microsoft/playwright-pytest#screenshot-on-test-failure

vashirov: And `DEBUG=pw:api` would be useful too for stack traces from playwright.

dirsrvtests/tests/suites/webui/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    #assert frame.is_visible("#no-inst-create-btn")
+
+
+def setup_login(page):
+    page.set_viewport_size({"width": 1920, "height": 1080})
+    page.goto("http://localhost:9090/")
+
+    # We are at login page
+    page.fill('#login-user-input', 'root')
+    page.fill('#login-password-input', 'redhat')
```
vashirov: Found another issue with the CI container image. We don't have password set there. I don't like the idea of hardcoding the password, I'd rather generate it on runtime, change it before the start of the test and pass the value to the test, so it can use it. I'll send you the proposed changes.

dirsrvtests/tests/suites/webui/__init__.py
bsimonova: Ahh, I knew I wanted to include screenshots :D thanks for reminder.
I had to install slugify and unicode-slugify to make the paths work. Would this dependency be ok with us or do we want to modify it without slugify?
You mean to add DEBUG=pw:api to pytest.yml?

dirsrvtests/tests/suites/webui/__init__.py
bsimonova: I also don't like the hardcoded value so I am glad there is another suggestion. Thanks, will check the changes.

dirsrvtests/tests/suites/webui/__init__.py
vashirov: > I had to install slugify and unicode-slugify to make the paths work. Would this dependency be ok with us or do we want to modify it without slugify?

Are both of these needed or just `python-slugify`?

> You mean to add DEBUG=pw:api to pytest.yml?

Correct.

bsimonova: Just python-slugify should be enough.

dirsrvtests/tests/suites/webui/login/login_test.py
```diff
@@ -0,0 +1,113 @@
...
+         2. Success
+         3. Login successful
+         4. Page redirection successful
+         5. We are at login page
+    """
+
+    assert page.url == "http://localhost:9090/389-console"
+
+    log.info('Let us log out')
+    page.click('#content-user-name')
```
vashirov: In cockpit-250 on F34 it's a new selector "#navbar-dropdown".

dirsrvtests/tests/suites/webui/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # We are at login page
+    page.fill('#login-user-input', 'root')
+    page.fill('#login-password-input', password)
+    page.click("#login-button")
+    if RHEL in distro.linux_distribution():
+        page.wait_for_selector('text=Red Hat Directory Server')
+        page.click('text=Red Hat Directory Server')
+    else:
+        page.wait_for_selector('389 Directory Server')
```
vashirov: Missing `text=` in the selector

.github/workflows/pytest.yml
```diff
@@ -65,7 +65,9 @@ jobs:
         set -x
         CID=$(sudo docker run -d -h server.example.com --privileged --rm -v /sys/fs/cgroup:/sys/fs/cgroup:rw,rslave -v ${PWD}:/workspace quay.io/389ds/ci-images:test)
         sudo docker exec $CID sh -c "dnf install -y -v dist/rpms/*rpm"
-        sudo docker exec $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml -v dirsrvtests/tests/suites/${{ matrix.suite }}
+        export PASSWD=$(openssl rand -base64 32)
+        sudo docker exec $CID sh -c "echo \"${PASSWD}\" | passwd --stdin root"
+        sudo docker exec -e WEBUI=1 -e DEBUG=pw:api -e PASSWD="${PASSWD}" $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml -v dirsrvtests/tests/suites/${{ matrix.suite }}
```
vashirov: We should also start `cockpit.socket`, but maybe I should fix it in the image itself.

vashirov (changes requested):

dirsrvtests/tests/suites/webui/create/create_instance_test.py
```diff
@@ -0,0 +1,212 @@
...
+    :setup: Standalone instance
+    :steps:
+         1. Go to Red Hat Directory server side tab page
+         2. Check there is Create New Instance button when no instance exists
+    :expectedresults:
+         1. Success
+         2. Button is visible
+    """
+
+    frame = page.frame('cockpit1:localhost/389-console')
```
vashirov: All tests fail with firefox, because for some reason selection of the frame by name returns `None`. If I select a frame from `page.frames` list, then it works. But the index of the required frame can change. 
I ran codegen to see what it suggests, and after adding this snippet, frame selection works:
```
    # Click text=389 Directory Server
    # with page.expect_navigation(url="http://localhost:9090/389-console"):
    with page.expect_navigation():
        page.click("text=389 Directory Server")
    assert page.url == "http://localhost:9090/389-console"

```

bsimonova: I applied the changes and we should have the tests fixed for chromium now at least. But for firefox I still have issues accessing the frame. Your snippet works but it fails with frames when proceeding to create instance.
I ran codegen and it produced this:
```
         # Click text=389 Directory Server
         # with page.expect_navigation(url="https://localhost:9090/389-console"):
         with page.expect_navigation():
             page.click("text=389 Directory Server")
         # assert page.url == "http://localhost:9090/389-console"

         # Go to https://localhost:9090/389-console
         page.goto("http://localhost:9090/389-console")

         # Click text=Create New Instance
         page.frame(name="cockpit1:localhost/389-console").click("text=Create New Instance")
```

The last line works in codegen but fails in tests again with a NoneType object. I have to investigate more.

bsimonova: @vashirov
Also I see it fails in checks again with connection refused. All  tests pass on my machine for chromium (I am still fixing firefox). Could it be an issue with how I set the password for the docker?

vashirov: IIRC there was an issue with `dbus.service` and `cockpit.socket` not starting by default. 
Could you please add this before running the tests to see if it helps:
```
sudo docker exec $CID sh -c "systemctl start dbus.service"
sudo docker exec $CID sh -c "systemctl enable --now cockpit.socket"
```

vashirov: I've added cockpit to the test image and enabling `cockpit.socket` works now. Some tests have finally passed, but some are failing due to timeout. There should be screenshots generated on failure, but currently they are not captured as part of the artifact collection step, we need to extend it too. I'll look into this tomorrow.

bsimonova:
@vashirov 
I see that for firefox we are missing dependency "libdbus-glib-1.so.2". Can you please add package dbus-glib to the installation?
Thanks!

Also the changes I made:

- added --browser firefox and chromium option to pytest.yml, it shouldn't be affecting regular tests
- iframe selection is determined by browser type in __init__.py
- cockpit version is checked in __init__.py because of renamed selector
- increased the default timeout from 30 000 to 60 000 miliseconds as the default value exceeded easily on slower machines

I will investigate why the failing tests still have timeout at checkbox.

bsimonova:
@vashirov 
The tests seem to be working now as some selectors got renamed.
Please review. Thanks!

vashirov (approved):
Looks good, let's merge this :)

PR #4774, opened 2021-05-20
Issue 4169 - UI - Port plugin tables to PF4

mreynolds389:
Description:  

port the plugins tables to PF4.  This completes the entire table migration.


src/cockpit/389-console/src/lib/plugins/dna.jsx
```diff
@@ -1249,9 +1250,11 @@ class DNA extends React.Component {
                             <div className="ds-margin-top-xlg">
                                 <DNASharedTable
                                     rows={sharedConfigRows}
+                                    key={sharedConfigRows}
```
jchapma, line 1253: This is more of a general question, can we use an array as a key, my understanding is that sharedConfigRows can be modified by the user so it changes. Or am i not getting something ?

src/cockpit/389-console/src/lib/plugins/passthroughAuthentication.jsx
```diff
@@ -71,6 +80,120 @@ class PassthroughAuthentication extends React.Component {
             urlEntryModalShow: false
         };
 
+        this.onExcludeToggle = isExcludedOpen => {
+            this.setState({
+                isExcludedOpen
```
jchapma, line 88: This is just nitpicking so ignore it if you like. Sometimes we use key: value and sometimes we use the ES6 way.

jchapma (commented):
This looks great, big improvement. A few trivial comments from me.

src/cockpit/389-console/src/lib/plugins/dna.jsx
mreynolds389: So, the key can be used to tell the component that it needs to be re-rendered.  Without the key if sharedCofigRows changed it would not be reflected in the UI/component.  So when the key changes from the last render. then it tells the component to render from scratch again.

Arrays work as keys but objects do not.  So you will see places where I use a index counter as the key (tableKey) instead of an array or other variable object.

src/cockpit/389-console/src/lib/plugins/passthroughAuthentication.jsx
mreynolds389: This code above comes directly from patternfly 4 documentation.  We mix and match this styling up all over the code.  Do you think its something that needs to be fixed?

src/cockpit/389-console/src/lib/plugins/dna.jsx
jchapma: > So, the key can be used to tell the component that it needs to be re-rendered. Without the key if sharedCofigRows changed it would not be reflected in the UI/component. So when the key changes from the last render. then it tells the component to render from scratch again.
> 
> Arrays work as keys but objects do not. So you will see places where I use a index counter as the key (tableKey) instead of an array or other variable object.

Ahh I see, that's handy. Thanks for the explanation.

src/cockpit/389-console/src/lib/plugins/passthroughAuthentication.jsx
jchapma: > This code above comes directly from patternfly 4 documentation. We mix and match this styling up all over the code. Do you think its something that needs to be fixed?

Ye i noticed the different styling, no i think its fine as it is.

jchapma: LGTM

droideck:
1. I modify a Attribute Uniqueness table's entry;
2. I save the result;
3. The table entry is not updated (but only in the table. The edit modal window shows the right data)

droideck:
1. Try to edit a Pass Through Authentication table's entry
2. When I click 'Save' button - it still hangs and the whole Plugin table interface is grayed out

droideck (changes requested):
I'm not sure why you've committed the Cargo.lock change here...

The rest looks great :)
It's really a vast amount of work!

PR #4776, opened 2021-05-21
Issue 4775 - Add entryuuid CLI and Fixup

### LINKED ISSUE #4775: Add missing entryuuid cli
This was missed in the entryuuid feature addition, so we need this for dsconf.

Firstyear:
Bug Description: EntryUUID when added was missing it's CLI
and helpers for fixups.

Fix Description: Add the CLI elements.

fixes: https://github.com/389ds/389-ds-base/issues/4775

Review by: ???

### DIFF
--- a/src/lib389/lib389/cli_conf/plugin.py
+++ b/src/lib389/lib389/cli_conf/plugin.py
@@ -28,6 +28,7 @@ from lib389.cli_conf.plugins import retrochangelog as cli_retrochangelog
 from lib389.cli_conf.plugins import automember as cli_automember
 from lib389.cli_conf.plugins import posix_winsync as cli_posix_winsync
 from lib389.cli_conf.plugins import contentsync as cli_contentsync
+from lib389.cli_conf.plugins import entryuuid as cli_entryuuid
 
 SINGULAR = Plugin
 MANY = Plugins
@@ -115,6 +116,7 @@ def create_parser(subparsers):
     cli_retrochangelog.create_parser(subcommands)
     cli_posix_winsync.create_parser(subcommands)
     cli_contentsync.create_parser(subcommands)
+    cli_entryuuid.create_parser(subcommands)
 
     list_parser = subcommands.add_parser('list', help="List current configured (enabled and disabled) plugins")
     list_parser.set_defaults(func=plugin_list)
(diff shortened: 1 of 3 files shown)


mreynolds389:
Fixup fails to run for me:

# dsconf localhost plugin entryuuid fixup -f objectclass=* dc=example,dc=com
Attempting to add task entry...
EntryUUID fixup task for %s has failed. Please, check logs

So there is a bug  here reporting the error...

Logs don't provide much help:

[21/May/2021:09:22:10.881267236 -0400] - ERR - plugins/entryuuid/src/lib.rs:187 - task_handler -> fixup complete, failed -> Unknown
[21/May/2021:09:22:10.885831138 -0400] - ERR - plugins/entryuuid/src/lib.rs:20 - entryuuid_plugin_task_thread => task error -> GenericFailure


I am testing this against 1.4.3, not sure if something else is missing on my part, but the error handler should still be fixed.

mreynolds389:
I was not missing anything, turns out if you add a filter to the fixup task then things break...

Firstyear:
> I was not missing anything, turns out if you add a filter to the fixup task then things break...

I can't reproduce this mate,  what filter are you trying to add?

mreynolds389:
> > I was not missing anything, turns out if you add a filter to the fixup task then things break...
> 
> I can't reproduce this mate, what filter are you trying to add?

# dsconf localhost plugin entryuuid fixup -f objectclass=* dc=example,dc=com

I also tried objectclass=top

Firstyear:
I've just used both and it works. Can you try '(objectClass=*)' ?

Firstyear:
Yeah, it's the lack of '(' and ')' in the filter. This string is being passed directly into slapi_search_internal_set_pb_ext() which means likely it's just a 'feature' of the slapi api that it needs the outer ().

mreynolds389:
> Yeah, it's the lack of '(' and ')' in the filter. This string is being passed directly into slapi_search_internal_set_pb_ext() which means likely it's just a 'feature' of the slapi api that it needs the outer ().

Would it be too much to ask that the CLI or rust code adds the parenthesis if they are missing?  I don't think we require the parenthesis for other filters used by other DS CLI subcommands.  Just looking for consistency...

Firstyear:
I think if we are going to add this, it would be in the rust code. It's not too much to ask, I'll get it done today/tomorrow

Firstyear:
@mreynolds389 This fixes the filter-without-brace behaviour.

mreynolds389:
Ack

Issue #4706, opened 2021-03-29
negative wtime for compare operations


mreynolds389:
**Issue Description**

The wtime and optime values in the access log are invalid because we forget to set the start time for CMP operations

Steps to Reproduce:
1. ldapcompare -D uid=user,dc=example,dc=com -w password cn=demo_group,ou=groups,dc=example,dc=com gidNumber:9999

Actual results:
[03/Feb/2021:18:07:58.251465934 +0000] conn=12 op=0 BIND dn="uid=user,dc=example,dc=com" method=128 version=3
[03/Feb/2021:18:07:58.252098372 +0000] conn=12 op=0 RESULT err=0 tag=97 nentries=0 wtime=0.000182726 optime=0.000675624 etime=0.000854818 dn="uid=user,dc=example,dc=com"
[03/Feb/2021:18:07:58.252518220 +0000] conn=12 op=1 CMP dn="cn=demo_group,ou=groups,dc=example,dc=com" attr="gidNumber"
[03/Feb/2021:18:07:58.252746534 +0000] conn=12 op=1 RESULT err=5 tag=111 nentries=0 wtime=-161634.769201119 optime=161633.231282362 etime=0.000478713
[03/Feb/2021:18:07:58.502988956 +0000] conn=12 op=2 UNBIND
[03/Feb/2021:18:07:58.503016718 +0000] conn=12 op=2 fd=65 closed error - U1

mreynolds389:
0c51de739..ed4773408  master -> master
37dc95673..2cda79bce  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
237212dc8..6b78f4ffc  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=1924848

PR #4780, opened 2021-05-25
Issue 4706 - negative wtime for compare operations

aadhikar:
Bug Description: Compare operation generates negative wtime in the access log

Relates: https://github.com/389ds/389-ds-base/issues/4706

Reviewed by: @vashirov


droideck (approved):
LGTM!

PR #4783, opened 2021-05-26
Issue 4764 - replicated operation sometime checks ACI

### LINKED ISSUE #4764: replicated operation sometime checks ACI
+++ This bug was initially created as a clone of RedHat Bugzilla #1931820 +++

Description of problem:

From time to time (about once per month) a replicated operation failed due to LDAP error 50 
( LDAP_INSUFFICIENT_ACCESS ).

With the error log level 262144 ( Access control summary information ), the following message is logged:
[18/Feb/2021:17:56:22.531733035 +0100] - DEBUG - NSACLPlugin - print_access_control_summary - conn=594722 op=22 (main): Deny write on entry(companyid=123456789,cn=app,dc=example,dc=com).attr(companyID) to cn=replication manager,cn=config: no aci matched the subject by aci(174): aciname= "Admin ACI", acidn="dc=example,dc=com"

Problem was initially reported on RedHat 11.2 but the issue impact all versions on all platform.
 
Replicated operation are supposed to bypass the acl check. A debug build showed that the operation is not flagged as replicated while the connection is seen as a replication session connection. 
The operation is flagged as replicated when allocating an operation struct if its connection is marked as replication session.

Code review showed that there is a race condition and operation may be allocated before its LDAP packet is received.
  if nsslapd-enable-turbo-mode is enabled (which is the default)
  if connection is secured and some data get exchanged at the security protocol level 
 In that case it is possible that the check to set the replicated flags occurs while the replication start session is processed and
  before the connection get marked as replication session.
  
  Workaround:
    a simple work around is to add an aci allowing write to the replication manager 
   
  Proposed fix is to check again to set the replicated flag just before dispatching the operation 
   (i.e after the operation is read but before it get processed):
   
 ```
  diff --git a/ldap/servers/slapd/connection.c b/ldap/servers/slapd/connection.c
index 615aa3dbe..3be6456ab 100644
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -1771,6 +1771,13 @@ connection_threadmain()
             }
         }
 
+        /*
+         * Fix bz 1931820 issue (check to set OP_FLAG_REPLICATED may be done
+         * before replication session is properly set set.
+         */
+        if (replication_connection) {
+                       operation_set_flag(op, OP_FLAG_REPLICATED);
+        }
         /*
          * Call the do_<operation> function to process this request.
          */

```

### PR DESCRIPTION
**Pull request by progier389**
Bug Description: Replication operation sometime fails with err=50, 
   it occurs the failing operation is not flagged as replicated although its connection is flagged as a replication connection.
   The operation is flagged as  replicated by the replication factory constructor but this is done before reading the data so it is possible that op struct for first operation after setting up the replication session get built before receiving the operation
   (while replication session is not yet fully set up.
   
 Fix description:  Adding a second test (to set the operation as replicated  if done on a replicated connection) after having read the operation but before dispatching it does fix this risk of race condition                     

Relates: #4764

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -1772,6 +1772,14 @@ connection_threadmain()
         }
 
         /*
+         * Fix bz 1931820 issue (the check to set OP_FLAG_REPLICATED may be done
+         * before replication session is properly set).
+         */
+        if (replication_connection) {
+            operation_set_flag(op, OP_FLAG_REPLICATED);
+        }
+
+        /*
          * Call the do_<operation> function to process this request.
          */
         connection_dispatch_operation(conn, op, pb);

### CODE REVIEW

**Comment by mreynolds389** [comment]
LGTM, ack

**Comment by progier389** [comment]
change: 0cfdea7abcacfca6686a6cf84dbf7ae1167f3022

3111a1660..0cfdea7ab  master 
bba519cce..407a9ebeb  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
bba519cce..407a9ebeb  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

**Comment by aivanov389** [comment]
After this commit i have an instant crash after any change of the server data with enabled replica (389ds v1.4.4):
```
/usr/sbin/dsconf ldaps://ldap-model.example.com:636 -D 'cn=Directory Manager' -w 'mdp' replication create-manager --name 'cn=RepliX,cn=config' --passwd 'repman_pw'
/usr/sbin/dsconf ldaps://ldap-model.example.com:636 -D 'cn=Directory Manager' -w 'mdp' replication enable --suffix='dc=example,dc=com' --role='supplier' --replica-id=2 --bind-dn='cn=RepliX,cn=config'
```

Any modification on the server crashes it with the following in error log:
```
[04/Jun/2021:16:08:37.630371432 +0200] - ERR - NSMMReplicationPlugin - IsValidOperation - NULL operation CSN
[04/Jun/2021:16:08:37.631086979 +0200] - ERR - NSMMReplicationPlugin - write_changelog_and_ruv - Can't add a change for uid=... (uniqid: (null), optype: 8) to changelog csn 
```

Issue #4778, opened 2021-05-21
[RFE] Schedule execution of "compactdb" at specific date/time


mreynolds389:
Description of problem:
  Setting "nsslapd-changelogcompactdb-interval" schedules the compactdb run every X seconds - the default is 30 days. There seems to be now way to execute the "changelogcompactdb" run at a specific date/time.

Since the changelog was moved into the database backend, we only need one settings for a database and its replication changelog

https://bugzilla.redhat.com/show_bug.cgi?id=1748441

mreynolds389:
407a9ebeb..7b5834702  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
8168e9225..150ded627  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4784, opened 2021-05-26
Issue 4778 - RFE - Allow setting TOD for db compaction and add task

mreynolds389:
Description:  

Since database compaction can be costly it should be allowed to set a time to execute it during offpeak hours.  Once the compaction interval has been met, it will wait for the configured time of day to do the compaction.  The default is just before midnight: 23:59

A task was also created that can run compaction on demand, and can also just target the replication changelog.  This could be used in conjunction with a cronjob for more complex execution patterns.

ASAN tested and approved.

relates: https://github.com/389ds/389-ds-base/issues/4778

### DIFF
--- a/dirsrvtests/tests/suites/config/compact_test.py
+++ b/dirsrvtests/tests/suites/config/compact_test.py
@@ -1,0 +1,80 @@
+import logging
+import pytest
+import os
+import time
+from lib389.tasks import DBCompactTask
+from lib389.backend import DatabaseConfig
+from lib389.topologies import topology_m1 as topo
+
+log = logging.getLogger(__name__)
+
+
+def test_compact_db_task(topo):
+    """Specify a test case purpose or name here
+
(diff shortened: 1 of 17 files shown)


dirsrvtests/tests/suites/config/compact_test.py
```diff
@@ -0,0 +1,80 @@
+import logging
+import pytest
+import os
+import time
+from lib389.tasks import DBCompactTask
+from lib389.backend import DatabaseConfig
+from lib389.topologies import topology_m1 as topo
+
+log = logging.getLogger(__name__)
+
```
droideck, line 10: If you agree that this tests can be run as part of the PR CI, could you please add the following line here?

    pytestmark = pytest.mark.tier1

droideck (approved):
the rest, actually, looks very good and clean!
you have my ack.

mreynolds389:
> the rest, actually, looks very good and clean!
 Geez don't seem so surprised :-)

Issue #4789, opened 2021-06-03
Temporary password rules are not enforce with local password policy


tbordaz:
**Issue Description**
A local password policy may contain temporary password rules. However they are not taken into account and fallback to global ones.

**Package Version and Platform:**
 since 1.4.3 supporting temporary password rules (#4725 )

**Steps to Reproduce**
testcase to be provided

**Expected results**
local password policy should be taken into account

tbordaz:
f53b28448..f8e42061f  master
02ca55ddd..a16a65325  389-ds-base-1.4.4
05c07a61e..bfa85a2cc  389-ds-base-1.4.3

mreynolds389:
> [f53b284](https://github.com/389ds/389-ds-base/commit/f53b284480118e080ce034c1d51902d47b80686c)..[f8e4206](https://github.com/389ds/389-ds-base/commit/f8e42061f342d24495d159beae983aee145696cb) master
> [02ca55d](https://github.com/389ds/389-ds-base/commit/02ca55ddd0c52a51edd302fe5fc4938d2ffd2ea9)..[a16a653](https://github.com/389ds/389-ds-base/commit/a16a65325e6dfce5afd275600323ec06f9ff83e7) 389-ds-base-1.4.4
> [05c07a6](https://github.com/389ds/389-ds-base/commit/05c07a61e2530eb26112868494d3c459a128e30c)..[bfa85a2](https://github.com/389ds/389-ds-base/commit/bfa85a2cc302ea2373523652a791b1f3117f18bc) 389-ds-base-1.4.3

This was not committed to 1.4.3, so not sure about that commit message.  Here is the new commit:

ef2c21240..e5b2fb2d6  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4790, opened 2021-06-03
Issue 4789 - Temporary password rules are not enforce with local pass…

tbordaz:
…word policy

Bug description:
	When allocating a password policy structure (new_passwdPolicy)
        it is initialized with the local policy definition or
	the global one. If it exists a local policy entry, the TPR
        attributes (passwordTPRMaxUse, passwordTPRDelayValidFrom and
        passwordTPRDelayExpireAt) are not taken into account.

Fix description:
	Take into account TPR attributes to initialize the policy

relates: https://github.com/389ds/389-ds-base/issues/4789

Reviewed by:

Platforms tested: F34

### DIFF
--- a/ldap/servers/slapd/pw.c
+++ b/ldap/servers/slapd/pw.c
@@ -2356,6 +2356,18 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
                     if ((sval = attr_get_present_values(attr))) {
                         pwdpolicy->pw_dict_path = (char *)slapi_value_get_string(*sval);
                     }
+                } else if (!strcasecmp(attr_name, CONFIG_PW_TPR_MAXUSE)) {
+                    if ((sval = attr_get_present_values(attr))) {
+                        pwdpolicy->pw_tpr_maxuse = slapi_value_get_int(*sval);
+                    }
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/pw.c
```diff
@@ -2356,6 +2356,18 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
                     if ((sval = attr_get_present_values(attr))) {
                         pwdpolicy->pw_dict_path = (char *)slapi_value_get_string(*sval);
                     }
+                } else if (!strcasecmp(attr_name, CONFIG_PW_TPR_MAXUSE)) {
+                    if ((sval = attr_get_present_values(attr))) {
+                        pwdpolicy->pw_tpr_maxuse = (char *)slapi_value_get_int(*sval);
```
droideck: Is there a reason you cast int to 'char *'? Seems a bit odd...

tbordaz: Good catch.. it was a bad copy/paste :)
Updated the patch

droideck (approved):
LGTM!

Firstyear (approved):

tbordaz:
Thanks for the reviews

Issue #4788, opened 2021-06-01
CLI should support Temporary Password Rules attributes


tbordaz:
**Issue Description**
RFE [Temporary Password Rules](https://www.port389.org/docs/389ds/design/otp-password-policy.html
)  supports new password policies attributes passwordTPRMaxUse, passwordTPRValidFrom, passwordTPRExpireAt.
The CLI should allow to set those values

**Package Version and Platform:**
1.4.3 and after

**Steps to Reproduce**
At the moment only ldapmodify allows to set those attributes

**Expected results**
dsconf should get/set them

tbordaz:
Related to https://github.com/389ds/389-ds-base/issues/4725

tbordaz:
b251ffe74..59d889ad1  master
fa46922b6..8de81d540  389-ds-base-1.4.4
ddc28036c..3ec3e3bd5  389-ds-base-1.4.3

PR #4793, opened 2021-06-04
Ticket 4788

tbordaz:
Issue 4788 - CLI should support Temporary Password Rules attributes
    
Bug description:

     Since #4725, password policy support temporary password rules.
     CLI (dsconf) does not support this RFE and only direct ldap
     operation can configure global/local password policy

    
Fix description:

    Update dsconf to support this new RFE.
    To run successfully the testcase it relies on #4788
    
relates: #4788
    
Reviewed by: 
    
Platforms tested: F34

### DIFF
--- a/dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
+++ b/dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
@@ -34,7 +34,7 @@ log = logging.getLogger(__name__)
 
 
 @pytest.fixture(scope="module")
-def create_user(topology_st, request):
+def test_user(topology_st, request):
     """User for binding operation"""
     topology_st.standalone.config.set('nsslapd-auditlog-logging-enabled', 'on')
     log.info('Adding test user {}')
@@ -56,10 +56,11 @@ def create_user(topology_st, request):
         topology_st.standalone.simple_bind_s(DN_DM, PASSWORD)
 
     request.addfinalizer(fin)
+    return user
 
 
 @pytest.fixture(scope="module")
-def password_policy(topology_st, create_user):
+def password_policy(topology_st, test_user):
     """Set up password policy for subtree and user"""
 
     pwp = PwPolicyManager(topology_st.standalone)
@@ -72,7 +73,7 @@ def password_policy(topology_st, create_user):
 
 @pytest.mark.bz1845094
 @pytest.mark.skipif(ds_is_older('1.4.3.3'), reason="Not implemented")
-def test_pwdReset_by_user_DM(topology_st, create_user):
+def test_pwdReset_by_user_DM(topology_st, test_user):
     """Test new password policy attribute "pwdReset"
 
     :id: 232bc7dc-8cb6-11eb-9791-98fa9ba19b65
@@ -115,7 +116,7 @@ def test_pwdReset_by_user_DM(topology_st, create_user):
 
 
 @pytest.mark.skipif(ds_is_older('1.4.3.3'), reason="Not implemented")
-def test_pwd_reset(topology_st, create_user):
+def test_pwd_reset(topology_st, test_user):
     """Test new password policy attribute "pwdReset"
 
     :id: 03db357b-4800-411e-a36e-28a534293004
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -749,6 +749,139 @@ def fin():
...
+    # passwordTPRMaxUse so that TPR is enforced first
+    topology_st.standalone.config.replace('passwordMustChange', 'on')
+    topology_st.standalone.config.replace('passwordMaxFailure', str(global_tpr_maxuse + 20))
+    topology_st.standalone.config.replace('passwordTPRMaxUse', str(global_tpr_maxuse))
+    time.sleep(.5)
+
+    local_tpr_maxuse = global_tpr_maxuse + 5
+    # Reset user's password with a local password policy
+    # that has passwordTPRMaxUse higher than global
+    our_user = UserAccount(topology_st.standalone, TEST_USER_DN)
```
droideck: It's better to use the fixture directly.
For that, you need to add this line to '**def create_user()**'  function:
```
return user
```
And then, in the test here, you can just call it directly like:
```
subprocess.call(['%s/dsconf' % topology_st.standalone.get_sbin_dir(),
                     'slapd-standalone1',
                     'localpwp',
                     'adduser',
                     create_user.dn])
```
Hence, it makes sense to name the fixture '**test_user**' instead of '**create_user**'.

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -796,6 +929,12 @@ def test_global_tpr_delayValidFrom_1(topology_st, create_user, request):
     assert our_user.get_attr_val_utf8('pwdTPRReset') == 'TRUE'
     now = time.mktime(time.gmtime())
     log.info("compare pwdTPRValidFrom (%s) vs now (%s)" % (our_user.get_attr_val_utf8('pwdTPRValidFrom'), time.gmtime()))
+    time.sleep(.5)
+    log.info("compare pwdTPRValidFrom (%s) vs now (%s)" % (our_user.get_attr_val_utf8('pwdTPRValidFrom'), time.gmtime()))
+    time.sleep(1)
+    log.info("compare pwdTPRValidFrom (%s) vs now (%s)" % (our_user.get_attr_val_utf8('pwdTPRValidFrom'), time.gmtime()))
+    time.sleep(2)
+    log.info("compare pwdTPRValidFrom (%s) vs now (%s)" % (our_user.get_attr_val_utf8('pwdTPRValidFrom'), time.gmtime()))
     assert (gentime_to_posix_time(our_user.get_attr_val_utf8('pwdTPRValidFrom'))) >= (now + ValidFrom - 2)
```
droideck: Also, I see that the next test fails:
```
>       assert (gentime_to_posix_time(our_user.get_attr_val_utf8('pwdTPRValidFrom'))) >= (now + ValidFrom - 2)
...
        time_tuple = (int(gentime[:4]),
>                     int(gentime[4:6]),
                      int(gentime[6:8]),
                      int(gentime[8:10]),
                      int(gentime[10:12]),
                      int(gentime[12:14]),
                      0, 0, 0)
E       ValueError: invalid literal for int() with base 10: ''
```

And in the logs it's written that **pwdTPRValidFrom == -1**

droideck (changes requested):

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
tbordaz: Yes, those tests are failing because of local password policy leftover (from test_local_tpr_maxuse_5). I am testing a fix for those failure

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
tbordaz: Changed done. Would you please have a look again. Thanks

droideck (approved):
LGTM!

Issue #4797, opened 2021-06-09
ACL IP ADDRESS evaluation may corrupt  c_isreplication_session connection flags


progier389:
**Issue Description**
fix about 4764 issue revealed this issue

break is missing in slapi_pblock_set function between 
  case SLAPI_CONN_CLIENTNETADDR_ACLIP  and 
  case SLAPI_CONN_IS_REPLICATION_SESSION

Although usually undetected this issue may generates unexpected behavior (operation may be seen as replicated while it should not) because of that for some people, fix 4764 (which is correct) may lead to crash  
 
**Package Version and Platform:**
 - Platform: any
 - Package and version: latest [e.g. 389-ds-base-1.4.4.4-20200721git5d41dc5a4.fc32.x86_64]

**Steps to Reproduce**
1) Set up a master instance
2) Initialize database from ldif in
[tc.zip](https://github.com/389ds/389-ds-base/files/6626068/tc.zip)
3) Run perl SimpleBindSeveralOps.perl -h localhost -p <port>  -D uid=foo,ou=people,dc=example,dc=com -w foo -b dc=example,dc=com -s sub uid=demo_user
4) Run again the perl command  (it fails because server crashed after 3)

Note: for writing a nice test in pytest: 
   The perl just bind that perform a search then modify the found entry description
    (The search trigger the aci evaluation and mark the connection as replicated then the modify is seen as a replicated operation 
     and since the csn is missing it ended to untested error handling which leads to a crash)     
Note: I suspect we may hit the same issue without 4764 issue fix (but we should probably do several modify operation in the same connection )

tbordaz:
@aivanov389 , @progier389 (very) nice finding !!!
The fix for ticket #[3764](https://github.com/389ds/389-ds-base/issues/3764) was broken with a missing [break](https://github.com/389ds/389-ds-base/commit/463d6b754dbafd79bae5bd66ce80605dd5b128f5#diff-1d5e184f43261cad6e0c6a9020c6c455fe989d431034337397e8ffdbf6b038f8R2582) in a switch.

This current ticket applies to all the versions where #3764 was committed. Then milestone is 1.3.10.

tbordaz:
Fix verified. Had to install perl-LDAP and update SimpleBindSeveralOps (use ldap rather that ldaps (line66) and supplier port)

tbordaz:
@progier389, sorry for having "stolen" that ticket after you did all the hard work. I had to fix it rapidly for a release.

[...]

PR #4799, opened 2021-06-10
Issue 4797 - ACL IP ADDRESS evaluation may corrupt c_isreplication_se…

tbordaz:
…ssion connection flags

Bug description:
	The fix for ticket #3764 was broken with a missing break in a
	switch. The consequence is that while setting the client IP
	address in the pblock (SLAPI_CONN_CLIENTNETADDR_ACLIP), the
	connection is erroneously set as replication connection.
        This can lead to crash or failure of testcase
        test_access_from_certain_network_only_ip.
        This bug was quite hidden until the fix for #4764 is
        showing it more frequently

Fix description:
	Add the missing break

relates: https://github.com/389ds/389-ds-base/issues/4797

Reviewed by:

Platforms tested: F33

### DIFF
--- a/ldap/servers/slapd/pblock.c
+++ b/ldap/servers/slapd/pblock.c
@@ -2583,7 +2583,7 @@ slapi_pblock_set(Slapi_PBlock *pblock, int arg, void *value)
         pblock->pb_conn->c_authtype = slapi_ch_strdup((char *)value);
         pthread_mutex_unlock(&(pblock->pb_conn->c_mutex));
         break;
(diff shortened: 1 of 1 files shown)


mreynolds389:
Ack, this could have qualified the one-line-commit rule :-)

tbordaz:
Thanks Mark for the review. Was feeling a bit guilty on that regression, so I preferred another pair of eyes on that one line fix. Also I was able to run the testcase (just perfect).

PR #4801, opened 2021-06-11
Issue 4414 - disk monitoring - prevent division by zero crash

bsimonova:
Description:
Added a test to check DS will not crash when division by zero
occurs in disk monitoring. Also fixed a description in compact_test.py
because it was causing errors in Polarion.

Relates: https://github.com/389ds/389-ds-base/issues/4414

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
+++ b/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
@@ -1,0 +1,102 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import time
+import subprocess
+import pytest
+from lib389.tasks import *
+from lib389.utils import *
+from lib389.topologies import topology_st
+from lib389._mapped_object import DSLdapObjects
+
+pytestmark = pytest.mark.tier2
+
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
```diff
@@ -0,0 +1,102 @@
...
+        1. Success
+        2. Success
+        3. Success
+    """
+
+    standalone = topology_st.standalone
+
+    log.info('Check search works before changing the nsslapd-auditlog attribute')
+    try:
+        DSLdapObjects(topology_st.standalone, basedn='cn=disk space,cn=monitor').filter("(objectclass=*)", scope=0)
```
droideck, line 75: We have an object that can be used for that - https://github.com/389ds/389-ds-base/blob/master/src/lib389/lib389/monitor.py#L353 - MonitorDiskSpace.

Could you please try if it'll work with '**filter**' function?

droideck (changes requested):

dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
bsimonova: I tried the suggested object but it doesn't have filter attribute.

droideck: Ah, right, MonitorDiskSpace is a singular object (and only DSLdapObjects has the filter method).

droideck (approved):
I think it's okay to use DSLdapObjects directly here as it's really a testing case and there's no need to add something like Monitor**s**DiskSpace.

You have my ack! Thank you!

PR #4804, opened 2021-06-11
Issue 4656 - replication name change upgrade code causes crash with d…

mreynolds389:
…ynamic plugins

Bug Description:  

If dynamic plugins is enabled, the server will crash after restarting several plugins.  The global plugin list became corrupted, and an invalid plugin entry was read.

Fix Description:  

Always call the close function of a plugin even if its not started (this undoes a change from the previous patch that was not needed afterall).

Updated the replication plugin upgrade code logging to be more clear, and to be logged by default.

ASAN tested and approved

### DIFF
--- a/ldap/servers/slapd/plugin.c
+++ b/ldap/servers/slapd/plugin.c
@@ -3366,11 +3366,9 @@ plugin_remove_plugins(struct slapdplugin *plugin_entry, char *plugin_type __attr
                      */
                     return PLUGIN_BUSY;
                 }
-                if (plugin->plg_started) {
-                    Slapi_PBlock *pb = slapi_pblock_new();
-                    plugin_call_one(plugin, SLAPI_PLUGIN_CLOSE_FN, pb);
-                    slapi_pblock_destroy(pb);
-                }
+                Slapi_PBlock *pb = slapi_pblock_new();
+                plugin_call_one(plugin, SLAPI_PLUGIN_CLOSE_FN, pb);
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/plugin.c
```diff
@@ -3366,11 +3366,9 @@ plugin_remove_plugins(struct slapdplugin *plugin_entry, char *plugin_type __attr
...
                     return PLUGIN_BUSY;
                 }
-                if (plugin->plg_started) {
-                    Slapi_PBlock *pb = slapi_pblock_new();
-                    plugin_call_one(plugin, SLAPI_PLUGIN_CLOSE_FN, pb);
-                    slapi_pblock_destroy(pb);
-                }
+                Slapi_PBlock *pb = slapi_pblock_new();
+                plugin_call_one(plugin, SLAPI_PLUGIN_CLOSE_FN, pb);
+                slapi_pblock_destroy(pb);
```
tbordaz, line 3371: This change revert the function to the previous status, so it should be safe. Now I have a concern with badly written plugin where the call to the close function expects the call to the start function.

For example [acct_policy close function](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/acctpolicy/acct_init.c#L217), frees a static config entry that is only initialized during start. it may crash if globalcfg is not initialized to 0.

mreynolds389: Sounds like we need a new ticket to investigate this...

PR #4709, opened 2021-03-31
issue 4653: refactor ldbm backend to allow replacement of BDB - phase 3e - dbscan

progier389:
Here is the last item in phase3 : about removing bdb dependency from dbscan tool
Note: I am quite sure that the tools will need some new feature (as to be able to list the databases) but that will be part of phase 4
 
 What was done:
- Add a new optional parameter to dbscan tool to select the database implementation (bdb by default)
- remove direct dependency towards bdb from dbsacn tools and use dbi interface instead.
- a few dbimpl callbacks are added (to open the environment  and a database outside of the dblayer API)  

Reviewed by: @Firstyear, @tbordaz @mreynolds389 

Tested on F32 with the following script:

dbscan -f  $DBDIR/userroot/id2entry.db4
dbscan -f  $DBDIR/userroot/objectclass.db4
dbscan -r -f  $DBDIR/userroot/objectclass.db4
dbscan -k =nsorgperson -r -f  $DBDIR/userroot/objectclass.db4
dbscan -K 5  -r -f  $DBDIR/userroot/objectclass.db4
dbscan -K 5  -f  $DBDIR/userroot/id2entry.db4
dbscan -D bdb -K 5  -f  $DBDIR/userroot/id2entry.db4
dbscan -D mdb -K 5  -f  $DBDIR/userroot/id2entry.db4
dbscan -G 8  -r -f  $DBDIR/userroot/objectclass.db4
dbscan -s -f  $DBDIR/userroot/objectclass.db4

PR #4805, opened 2021-06-11
Issue 4709 - Fix double free in dbscan

mreynolds389:
Description:  Fix double free in dbscan - in main()

relates: https://github.com/389ds/389-ds-base/pull/4709

### DIFF
--- a/ldap/servers/slapd/tools/dbscan.c
+++ b/ldap/servers/slapd/tools/dbscan.c
@@ -1,5 +1,5 @@
 /** BEGIN COPYRIGHT BLOCK
- * Copyright (C) 2005 Red Hat, Inc.
+ * Copyright (C) 2021 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/tools/dbscan.c
```diff
@@ -1288,12 +1288,6 @@ main(int argc, char **argv)
     }
 
 done:
-    if (key.data) {
-        free(key.data);
-    }
-    if (data.data) {
-        free(data.data);
-    }
```
tbordaz, line 1296: The fix looks good, just to clarify my understanding is the problem that key.data and data.data were not set to NULL after 'free' ? So next dblayer_value_free triggered a double free ?

tbordaz (approved):

ldap/servers/slapd/tools/dbscan.c
mreynolds389: dblayer_value_free() is basically redundant of those frees I removed.  We don't need both...

droideck (approved):
LGTM too!

PR #4806, opened 2021-06-11
Issue 4093 - Fix MEP test case

mreynolds389:
Bug Description:  

Once some compiler warnings were fixed it accidentally fixed the modrdn behavior.  Previously the modrdn code accidentally ignored errors that the test case was taking for granted.  Once these checks were properly enforced the test case started to fail.

Fix Description:  Revise test case to "properly" check modrdn operations by creating the Managed Entry before assigning it to an entry, and then check for the revise managed entry DN after the modrdn takes place.

Also, improved CI debugging logging settings

relates: https://github.com/389ds/389-ds-base/issues/4093

### DIFF
--- a/dirsrvtests/tests/suites/plugins/managed_entry_test.py
+++ b/dirsrvtests/tests/suites/plugins/managed_entry_test.py
@@ -9,8 +9,8 @@
 import pytest
 import time
 from lib389.topologies import topology_st as topo
-from lib389.idm.user import UserAccount, UserAccounts
-from lib389.idm.account import Account, Accounts
+from lib389.idm.user import UserAccounts
+from lib389.idm.account import Account
 from lib389._constants import DEFAULT_SUFFIX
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/plugins/managed_entry_test.py
```diff
@@ -112,7 +111,7 @@ def __init__(self, instance, dn=None):
         self._rdn_attribute = 'uid'
         self._create_objectclasses = ['top', 'person', 'inetorgperson']
 
-#unstable or unstatus tests, skipped for now
+# unstable or unstatus tests, skipped for now
 @pytest.mark.flaky(max_runs=2, min_passes=1)
```
droideck: I think we can remove this line as the test is fixed.

droideck (approved):
The rest looks good to me!

PR #4807, opened 2021-06-15
Issue 4656 - Fix replication plugin rename dependency issues

mreynolds389:
Bug Description:  

If a plugin has a named dependency on the old Replication plugin name, and it is listed in the dse.ldif before the replication plugin then the "conversion" fails because the internal plugin dependency list was not properly updated

Fix Description:  

Update the plugin dependency list after we update a plugin's dependency.


droideck (approved):
Looks good to me!

PR #4808, opened 2021-06-15
Issue 4169 - UI - Migrate Typeaheads to PF4

jchapma:
Description: Migrate the current bootstrap typeaheads to
patternfly 4 select typeaheads

Relates: https://github.com/389ds/389-ds-base/issues/4169

Reviewed by:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,3 +235,4 @@ vendor
 vendor.tar.gz
 .history
 .vscode/launch.json
+.cargo/config
--- a/.gitignore
+++ b/.gitignore
@@ -231,6 +231,7 @@ venv
 .idea
 src/cockpit/389-console/cockpit_dist/
 src/cockpit/389-console/node_modules/
-ldap/servers/slapd/rust-slapi-private.h
 vendor
 vendor.tar.gz
+.history
+.vscode/launch.json
--- a/src/cockpit/389-console/src/lib/database/attrEncryption.jsx
+++ b/src/cockpit/389-console/src/lib/database/attrEncryption.jsx
@@ -10,10 +10,12 @@ import {
     Button,
     // Form,
     // FormGroup,
+    Select,
+    SelectVariant,
+    SelectOption,
     noop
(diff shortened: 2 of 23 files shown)


mreynolds389:
Looks good!  You used a slightly different coding style than we have been using for components, but I don't see any warnings about it so it is ok.  Please merge asap so I can adjust my current work on Forms.  Thanks!

droideck (changes requested):
Looks really good!

The only trouble I've encountered - I can't remove an item from typeahead field.

Browser: Chrome

Steps to reproduce:
1. Go to attribute uniqueness plugin settings, press three dots and then 'Edit Config'
2. In the 'Attribute Names' field try to press a small 'x' which is located right after the attribute name (.e. 'uid x').

Nothing happens and 'uid' is not removed from the list.

It should remove a separate item from the list.

P.S. the bit 'X' on the right side works perfectly though and it cleans the whole field

jchapma:
> Good catch @droideck I suspect @jchapma copied code that I had in place which might also be broken. So maybe he copied my bad code example? :-)

The problem was that I didn't copy your code example :(

mreynolds389:
There is a problem with "multiple" selection typeAheads.  You can not remove an element by clicking on the "x", or any other way for that matter.

jchapma:
> There is a problem with "multiple" selection typeAheads. You can not remove an element by clicking on the "x", or any other way for that matter.

Thats weird, I tested all the selects i modified and they worked fine. Let me run through them again.

mreynolds389:
> > There is a problem with "multiple" selection typeAheads. You can not remove an element by clicking on the "x", or any other way for that matter.
> 
> Thats weird, I tested all the selects i modified and they worked fine. Let me run through them again.

It was the global password policy "user attributes" typeahead.  I didn't try any others....

mreynolds389:
Seems to be working for me, ack

droideck (approved):
LGTM too!

PR #4810, opened 2021-06-17
Issue 4803 - Improve DB Locks Monitoring Feature Descriptions

### LINKED ISSUE #4803: Improve DB Locks Monitoring Feature Descriptions
**Issue Description**
The description of the field  "nsslapd-db-locks-monitoring-threshold" is unclear.

```
"Sets the DB lock exhaustion value in percentage (valid range is 70-95). 
If too many locks are acquired, the server will abort the searches while the number of locks are not decreased.
It helps to avoid DB corruption and long recovery. ()"
```

 - Package and version: 389-ds-base, master

**Additional context**
Probably, we can improve it like this:

`"When the threshold is reached, all searches are aborted until the number of active locks decreases below the configured threshold." `

Also, we should mention that the RHDS admin is expected to do something here.

Other parts of the feature docs can be improved.

droideck:
Description: The description of the field "nsslapd-db-locks-monitoring-threshold"
is unclear. Make the explanations more detailed and concise in both CLI
and Web UI.

Fixes: https://github.com/389ds/389-ds-base/issues/4803

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
+++ b/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
@@ -271,7 +271,7 @@ export class GlobalDatabaseConfig extends React.Component {
                     <FormGroup
                         label="DB Locks Threshold Percentage"
                         fieldId="monitoringthreshold"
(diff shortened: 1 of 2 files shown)


tbordaz:
Descriptions are looking good to me. Should they also mention the default value ?
ACK

EmmanuelKasper (changes requested):
Thanks Simon, very clear explanation, I am just concerned about the sentence:
`
It helps to avoid DB corruption and long recovery.
`
It is either saying too little or too much.

I would here rather write:
`This threshold is a safeguard against DB corruption which might be caused by locks exhaustion.`

or get rid of it, if the field description is too long.

Issue #4814, opened 2021-06-28
_cl5_get_tod_expiration may crash at startup


tbordaz:
**Issue Description**
The ticket #4778 allows to configure the scheduling of the database compaction. One of the configuration attribute is "nsslapd-changelogcompactdb-time" (that is "23:55" by default).

For the changelog compaction, the data structure that stores this attribute is initialized in cl5ConfigTrimming.

The spawn thread responsible of the changelog compaction, that uses this data structure, is started cl5Open.

The problem is that, during replication plugin init, cl5Open is called before cl5ConfigTrimming. There is a risk that the changelog compaction thread (_cl5TrimMain) uses s_cl5Desc.dbTrim.compactTime before it gets initialized.

**Package Version and Platform:**
branch 1.4.3 only (After #4778)

**Steps to Reproduce**
The crash is not systematic as it depends on thread scheduling

**Expected results**
Should not crash

tbordaz:
The crash backtrace is

    Thread 8 (Thread 0x7f7e278cd700 (LWP 15645)):
    #0  0x00007f7e17c9db51 in _cl5_get_tod_expiration (expire_time=0x0) at ldap/servers/plugins/replication/cl5_api.c:3107
    #1  0x00007f7e17c9db51 in _cl5TrimMain (param=<optimized out>) at ldap/servers/plugins/replication/cl5_api.c:3145
    #2  0x00007f7e2393f5d8 in _pt_root () at target:/lib64/libnspr4.so
    #3  0x00007f7e232da14a in start_thread () at target:/lib64/libpthread.so.0
    #4  0x00007f7e2564adc3 in clone () at target:/lib64/libc.so.6
 
    ...
 
    Thread 1 (Thread 0x7f7e278d2240 (LWP 15630)):
    #8  0x00007f7e17caf37a in add_new_agreement (e=e@entry=0x7f7e1a6b41c0) at ldap/servers/plugins/replication/repl5_agmtlist.c:157
    ...
    #18 0x00007f7e17cb077f in agmtlist_config_init () at ldap/servers/plugins/replication/repl5_agmtlist.c:758
    #19 0x00007f7e17cb9ae9 in multimaster_start (pb=0x7f7e1a617f00) at ldap/servers/plugins/replication/repl5_init.c:807
    #20 0x00007f7e17cb9ae9 in multimaster_start (pb=0x7f7e1a617f00) at ldap/servers/plugins/replication/repl5_init.c:757
[...]

PR #4816, opened 2021-06-29
Issue 4814 - _cl5_get_tod_expiration may crash at startup

tbordaz:
Bug description:
	This bug exist only in 1.4.3 branch
	In 1.4.3, CL open as a separated database so
        compaction mechanism is started along a CL
        mechanism (CL trimming).
        The problem is that the configuration of the CL
        compaction is done after the compaction mechanism
        (is started). Depending on thread scheduling it
        crashes

Fix description:
        Make sure configuration of compaction thread is
        taken into account (cl5ConfigSetCompaction) before
	the compaction thread starts (cl5open)

relates: https://github.com/389ds/389-ds-base/issues/4814

Reviewed by: Mark Reynolds

Platforms tested: 8.5

### DIFF
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -1016,6 +1016,20 @@ cl5GetState()
     return s_cl5Desc.dbState;
 }
 
+void
+cl5ConfigSetCompaction(int compactInterval, char *compactTime)
+{
+
+    if (compactInterval != CL5_NUM_IGNORE) {
+        s_cl5Desc.dbTrim.compactInterval = compactInterval;
+    }
+
+    if (strcmp(compactTime, CL5_STR_IGNORE) != 0) {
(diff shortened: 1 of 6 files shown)


ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -1016,6 +1016,20 @@ cl5GetState()
     return s_cl5Desc.dbState;
 }
 
+void
+cl5ConfigSetCompaction(int compactInterval, char *compactTime)
```
droideck, line 1020: Should we also change 'cl5ConfigTrimming' function in cl5_api.c file accordingly?

tbordaz: absolutely right. Sorry for that. Fixiing it

tbordaz: New patch pushed

PR #4818, opened 2021-06-29
Issue 4563 - Failure on s390x: 'Fails to split RDN  "o=pki-tomcat-CA"…

### LINKED ISSUE #4563: Failure on s390x: 'Fails to split RDN "o=pki-tomcat-CA" into components'

vashirov:
… into components'

Bug Description:
SLAPI_OPERATION_TYPE is defined as unsigned long. In various places we
use int instead. This works fine on little-endian platforms, but fails
on big-endian, s390x in particular.

Fix Description:
Update all places where we use slapi_pblock_get with
SLAPI_OPERATION_TYPE to use unsigned long.

Fixes: https://github.com/389ds/389-ds-base/issues/4563

Reviewed by: ???

### DIFF
--- a/ldap/servers/plugins/acl/acl.h
+++ b/ldap/servers/plugins/acl/acl.h
@@ -472,7 +472,7 @@ struct acl_pblock
 
     Slapi_Entry *aclpb_client_entry; /* A copy of client's entry */
     Slapi_PBlock *aclpb_pblock;      /* back to LDAP PBlock */
-    int aclpb_optype;                /* current optype from pb */
+    unsigned long aclpb_optype;                /* current optype from pb */
 
     /* Current entry/dn/attr evaluation info */
     Slapi_Entry *aclpb_curr_entry; /* current Entry being processed */
--- a/ldap/servers/plugins/acl/acl_ext.c
+++ b/ldap/servers/plugins/acl/acl_ext.c
@@ -737,7 +737,7 @@ acl_init_aclpb(Slapi_PBlock *pb, Acl_PBlock *aclpb, const char *ndn, int copy_fr
     struct acl_cblock *aclcb = NULL;
     char *authType;
     void *conn;
-    int op_type;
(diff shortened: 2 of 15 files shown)


vashirov:
I reviewed slapi-nis and freeipa plugins code. The only other occurrence I found in freeipa topology plugin: https://github.com/freeipa/freeipa/blob/19d5b3b621dbdfe96b290ac2f7af63008d01aa80/daemons/ipa-slapi-plugins/topology/topology_pre.c#L28

I'd like to test this with ipa first to avoid any breaking changes.

mreynolds389:
Looks good to me.  Now I was looking at the build logs for x390x and I don't see any compiler warnings.  Perhaps the flag "-Wconversion" might expose this.  I'll have to test it...

PR #4819, opened 2021-06-30
Issue 4817 - BUG - locked crypt accounts on import may allow all passwords

### LINKED ISSUE #4817: CRYPT password hash with asterisk
**Issue Description**
If an entry contains an asterisk as the crypted password hash, binding is possible with any password for this entry
userPassword: {CRYPT}*

 - Platform: openSUSE Leap 15.2
 - Package and version: 389-ds-1.4.3.23~git0.f53d0132b-lp152.2.15.1.x86_64
 - Browser firefox

**Steps to Reproduce**
1. Create an entry (e.g. a posixAccount) with the userPassword set to "{CRYPT}*", e.g. by importing it from an ldif file
2. Try to bind with that entry using an arbitrary password (e.g. "llhh"
3. Check if the binding was successfull

**Expected results**
I would expect to fail the binding with any password because the asterisk is not a vaild character in a crypted password.

**Screenshots**

**Additional context**
Problem occured after importing entries from a NIS database. In NIS (and in /etc/shadow), the asterisk is often used for special users like "nobody".

Firstyear:
Bug Description: Due to mishanding of short dbpwd hashes, the
crypt_r algorithm was misused and was only comparing salts
in some cases, rather than checking the actual content
of the password.

Fix Description: Stricter checks on dbpwd lengths to ensure
that content passed to crypt_r has at least 2 salt bytes and
1 hash byte, as well as stricter checks on ct_memcmp to ensure
that compared values are the same length, rather than potentially
allowing overruns/short comparisons.

fixes: https://github.com/389ds/389-ds-base/issues/4817

Review by: ???

### DIFF
--- a/dirsrvtests/tests/suites/password/pwd_crypt_asterisk_test.py
+++ b/dirsrvtests/tests/suites/password/pwd_crypt_asterisk_test.py
@@ -1,0 +1,50 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 William Brown <william@blackhats.net.au>
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import ldap
+import pytest
+from lib389.topologies import topology_st
+from lib389.idm.user import UserAccounts
+from lib389._constants import (DEFAULT_SUFFIX, PASSWORD)
+
+pytestmark = pytest.mark.tier1
+
+def test_password_crypt_asterisk_is_rejected(topology_st):
+    """It was reported that {CRYPT}* was allowing all passwords to be
+    valid in the bind process. This checks that we should be rejecting
+    these as they should represent locked accounts. Similar, {CRYPT}!
+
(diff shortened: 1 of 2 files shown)


mreynolds389:
Does this also cover the other crypt schemes?  crypt-md5, etc..

Firstyear:
Kind of - the other schemes aren't affected by this because of their algo prefixing as detailed in man crypt_r here:

```
If salt is a character string starting with the characters "$id$" followed by a string terminated by "$":

$id$salt$encrypted
then instead of using the DES machine, id identifies the encryption method used and this then determines how the rest of the password string is interpreted. The following values of id are supported:
So $5$salt$encrypted is an SHA-256 encoded password and $6$salt$encrypted is an SHA-512 encoded one.
```

However, they would still potentially have had the issue with hash truncation due to the missing length check before ct_memcmp. I'm not sure how this could be actually attacked though for the other hash types.

mreynolds389:
LGTM

Firstyear:
```
   804b46e06..18271bad6  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
   5d956fcc7..c1926dfc6  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

PR #4821, opened 2021-07-01
Issue 4820 - RFE - control flow integrity

### LINKED ISSUE #4820: clang - control flow integrity
Clang supports control flow integrity and some other useful behaviours for production code as well as developer code, so we should put 389 in a position to enable these.

Firstyear:
Bug Description: Many attacks involved hijacking the
control flow of an executable to change it's behaviour.
While we can do many things to prevent this at development
time, we need to be ready for unexpected situations in
run time.

Fix Description: Enabling control flow integrity allows
enforcing that our projects logic flow only goes in certain,
known valid locations at compile time. This means a program
that violates these behaviours will be terminated to prevent
exploitation.

fixes: https://github.com/389ds/389-ds-base/issues/4820

Review by: ???

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,16 @@ MSAN_CFLAGS = @msan_cflags@
 TSAN_CFLAGS = @tsan_cflags@
 UBSAN_CFLAGS = @ubsan_cflags@
 
+if CFI_ENABLE
+# https://clang.llvm.org/docs/ControlFlowIntegrity.html#available-schemes
+# vcall is "forward edge" cfi which is what gives a lot of benefit security wise.
+CFI_CFLAGS = -flto=thin -fsanitize=cfi-cast-strict,cfi-vcall -fvisibility=hidden
+# Settings we could use in the future
+# -fsanitize=cfi-icall,cfi-nvcall,cfi-derived-cast,cfi-unrelated-cast,cfi-mfcall
+else
+CFI_CFLAGS =
+endif
+
 SYSTEMD_DEFINES = @systemd_defs@
 
 CMOCKA_INCLUDES = $(CMOCKA_CFLAGS)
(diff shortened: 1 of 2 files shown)


jchapma (commented):
This is a very interesting idea, you have my ack. 

As an aside, do you know how dependent CFI is on HW support, if I dont have HW support for CFI will it work on my platform ?

Firstyear:
In this case it's NOT hardware dependent. This is software CFI as part of llvm. CFI has two versions effectively - forward edge integrity, and backward edge integrity. The LLVM CFI here is forward edge integrity: "Are we jumping to a valid location?".

You are also correct that there are some hardware versions of this though! These tend to be backward edge integrity: "Are we returning to the place where we came from?". AMD shadow stack or ARM BTI both are examples of this. These can be done in software, but they incur a higher penalty on performance as they need software shadow stacks. AMD/ARM here have hardware enablement to improve the performance of these but that would be a different llvm feature ( https://releases.llvm.org/7.0.1/tools/clang/docs/ShadowCallStack.html ) which I think is disabled.

Hope that helps :)

jchapma:
> In this case it's NOT hardware dependent. This is software CFI as part of llvm. CFI has two versions effectively - forward edge integrity, and backward edge integrity. The LLVM CFI here is forward edge integrity: "Are we jumping to a valid location?".
> 
> You are also correct that there are some hardware versions of this though! These tend to be backward edge integrity: "Are we returning to the place where we came from?". AMD shadow stack or ARM BTI both are examples of this. These can be done in software, but they incur a higher penalty on performance as they need software shadow stacks. AMD/ARM here have hardware enablement to improve the performance of these but that would be a different llvm feature ( https://releases.llvm.org/7.0.1/tools/clang/docs/ShadowCallStack.html ) which I think is disabled.
> 
> Hope that helps :)

It does, thank you for your explanation.

PR #4827, opened 2021-07-02
Issue 4826 - Filter argparse-manpage from autogenerated requires

### LINKED ISSUE #4826: Filter argparse-manpage from autogenerated requires
python3-argparse-manpage is a buildtime only dependency. But because it's mentioned in setup.py and requirements.txt it gets picked up by RPM dependency generators:
```
rpm -q --requires python3-lib389 | grep argparse 
python3.9dist(argparse-manpage)
```
We should exclude it from runtime dependencies list.

vashirov:
Bug Description:
RPM dependency generators add argparse-manpage to the list of runtime
dependencies. But it's a buildtime only dependency.

Fix Description:
Use requires filter macro in the spec file.

Fixes: https://github.com/389ds/389-ds-base/issues/4826

Reviewed by: ???


mreynolds389:
LGTM

Issue #4603, opened 2021-02-09
 Indexing a single backend actually processes all configured backends


jchapma:
**Issue Description**
While trying to index a given backend, the command output shows that all configured backends are being processed.
This can be problematic if one wants to reindex a small database and the command takes time because it runs also on existing larger backends.

**Package Version and Platform:**
 - Platform: 5.6.6-300.fc32.x86_64
 - Package and version: 389-ds-base-2.0.2.20210204gitf38b124f0

**Steps to Reproduce**
1. Verify you have multiple backends:
````
sudo dsconf -D cn=dm  -b "dc=example,dc=com" -w password inst01 backend suffix list
dc=example,dc=com (userroot)
dc=tempstuff,dc=com (tempstuff)
dc=test01,dc=com (test01)
dc=test02,dc=com (test02)
dc=userdata,dc=com (userdata)

````
2. Stop the server
````
sudo dsctl inst01 stop
````
3. Index a single backend
````
sudo dsctl inst01 db2index userroot
````
4. Check the output of the db2index command or check the error logs
````
[09/Feb/2021:09:24:27.710654349 +0000] - INFO - bdb_import_main - reindex userroot: Reindexing complete.  Processed 14 entries in 1 seconds. (14.00 entries/sec)
[09/Feb/2021:09:24:28.845524048 +0000] - INFO - bdb_import_main - reindex userdata: Reindexing complete.  Processed 14 entries in 1 seconds. (14.00 entries/sec)
[09/Feb/2021:09:24:29.963887675 +0000] - INFO - bdb_import_main - reindex test02: Reindexing complete.  Processed 14 entries in 1 seconds. (14.00 entries/sec)
[09/Feb/2021:09:24:31.080888430 +0000] - INFO - bdb_import_main - reindex test01: Reindexing complete.  Processed 14 entries in 1 seconds. (14.00 entries/sec)
[09/Feb/2021:09:24:32.206386181 +0000] - INFO - bdb_import_main - reindex tempstuff: Reindexing complete.  Processed 14 entries in 1 seconds. (14.00 entries/sec)
````
**Actual results**
All backends are processed, when only one was specified.

**Expected results**
Only the specified backend should be processed.

**Additional context**
Issue reported by Têko Mihinto. 
https://bugzilla.redhat.com/show_bug.cgi?id=1916292

jchapma:
There is a memory leak that I couldnt find yet, I will create a separate issue for it.

PR #4831, opened 2021-07-06
Issue 4603 - Reindexing a single backend

jchapma:
Bug description:
	While trying to offline reindex a single backend, all backends
	are reindexed. This can introduce un-wanted latencies if one wants to reindex a
	single backend rather than reindexing all backends.

Fix description:
	DB checkpoint disabled for offline reindex
	CLI modified to provide extra reindex options
	DSELidf extended to get backend index attributes

Relates: https://github.com/389ds/389-ds-base/issues/4603

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -470,6 +470,8 @@ def test_basic_db2index(topology_st):
 
     assert indexNum+1 == numIndexes
 
+    topology_st.standalone.start()
+
 
 def test_basic_acl(topology_st, import_example_ldif):
     """Run some basic access control (ACL) tests
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -401,8 +401,7 @@ def test_basic_backup(topology_st, import_example_ldif):
 
     log.info('test_basic_backup: PASSED')
 
-
-def test_basic_db2index(topology_st, import_example_ldif):
+def test_basic_db2index(topology_st):
     """Assert db2index can operate correctly.
 
     :id: 191fc0fd-9722-46b5-a7c3-e8760effe119
@@ -410,16 +409,66 @@ def test_basic_db2index(topology_st, import_example_ldif):
     :setup: Standalone instance
 
     :steps:
-        1: call db2index
+        1: Call db2index with a single index attribute
+        2: Call db2index with multiple index attributes
+        3: Call db2index with no index attributes
 
     :expectedresults:
-        1: Index succeeds.
+        1: Index succeeds for single index attribute
+        2: Index succeeds for multiple index attributes
+        3: Index succeeds for all backend indexes which have been obtained from dseldif
 
     """
-    topology_st.standalone.stop()
-    topology_st.standalone.db2index()
-    topology_st.standalone.db2index(suffixes=[DEFAULT_SUFFIX], attrs=['uid'])
+
+    indexes = []
+
+    # Error log message to confirm a reindex
+    info_message = 'INFO - bdb_db2index - ' + DEFAULT_BENAME + ':' + ' Indexing attribute: '
+
+    log.info('Start the server')
     topology_st.standalone.start()
+
+    log.info('Offline reindex, stopping the server')
+    topology_st.standalone.stop()
+
+    log.info('Reindex with a single index attribute')
+    topology_st.standalone.db2index(bename=DEFAULT_BENAME, attrs=['uid'])
+    assert topology_st.standalone.searchErrorsLog(info_message + 'uid')
+
+    log.info('Restart the server to clear the logs')
+    topology_st.standalone.start()
+    topology_st.standalone.stop()
+
+    log.info('Reindex with multiple attributes')
+    topology_st.standalone.db2index(bename=DEFAULT_BENAME, attrs=['cn','aci','givenname'])
+    assert topology_st.standalone.searchErrorsLog(info_message + 'cn')
+    assert topology_st.standalone.searchErrorsLog(info_message + 'aci')
+    assert topology_st.standalone.searchErrorsLog(info_message + 'givenname')
+
+    log.info('Restart the server to clear the logs')
+    topology_st.standalone.start()
+    topology_st.standalone.stop()
+
+    log.info('Start the server and get all indexes for specified backend')
+    topology_st.standalone.start()
+    dse_ldif = DSEldif(topology_st.standalone)
+    indexes = dse_ldif.get_indexes(DEFAULT_BENAME)
+    numIndexes = len(indexes)
+    assert numIndexes > 0
+
+    log.info('Stop the server and reindex with all backend indexes')
+    topology_st.standalone.stop()
+    topology_st.standalone.db2index(bename=DEFAULT_BENAME, attrs=indexes)
+    log.info('Checking the server logs for %d backend indexes' % numIndexes)
+    for indexNum, index in enumerate(indexes):
+        if index in "entryrdn":
+            assert topology_st.standalone.searchErrorsLog(
+                'INFO - bdb_db2index - ' + DEFAULT_BENAME + ':' + ' Indexing ' + index)
+        else:
+            assert topology_st.standalone.searchErrorsLog(
+                'INFO - bdb_db2index - ' + DEFAULT_BENAME + ':' + ' Indexing attribute: ' + index)
+    
+    assert indexNum+1 == numIndexes
 
 
 def test_basic_acl(topology_st, import_example_ldif):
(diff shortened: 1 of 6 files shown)


mreynolds389:
I see you changed some of the default behavior in dseldif.py (not normalizing a dn).  Why was this changed?  Was it fixing something else?  It could impact how other operations are done using this class.  Just make sure this does not break any of the tests that use this Class:

Under dirsrvtests/tests/suites/

  backups/backup_test.py
  basic/basic_test.py
  config/regression_test.py
  ds_logs/regression_test.py:
  healthcheck/healthcheck_test.py
  plugins/acceptance_test.py
  replication/regression_m2c2_test.py
  replication/encryption_cl5_test.py
  replication/changelog_test.py


If all goes well, ack

jchapma:
> I see you changed some of the default behavior in dseldif.py (not normalizing a dn). Why was this changed? Was it fixing something else? It could impact how other operations are done using this class. Just make sure this does not break any of the tests that use this Class:

DSELdif was extended to get all the indexes for a backend, we need to keep the index attributes original case as we use them during the reindex. The index attr is plucked from the dse index string for the particular backend (dn: cn=givenName,cn=index,cn=test02,cn=ldbm database,cn=plugins,cn=config)

I ran all the tests that used this class and the change didnt break anything.

droideck:
> > I see you changed some of the default behavior in dseldif.py (not normalizing a dn). Why was this changed? Was it fixing something else? It could impact how other operations are done using this class. Just make sure this does not break any of the tests that use this Class:
> 
> DSELdif was extended to get all the indexes for a backend, we need to keep the index attributes original case as we use them during the reindex. The index attr is plucked from the dse index string for the particular backend (dn: cn=givenName,cn=index,cn=test02,cn=ldbm database,cn=plugins,cn=config)
> 
> I ran all the tests that used this class and the change didnt break anything.

Generally, the idea behind standardizing the DN - is about making it easier for a caller to not worry about the DN case.

DN by itself is case-insensitive so it makes sense to give this functionality to DSEldif by default too. Even if it doesn't break current DSEldif usage, I think, it's better to keep the original and "DN as intended" functionality as it may be very confusing for someone in the future and it may lead to bugs and more developer time.

AS for this PR, maybe we can add an additional option for DSELdif class instance? Something like 'dn_case_sensetive' or something like that. Then it can be used intentionally when someone need it (probably, we can add a docstring comment to get_indexes method too about that).
What do you think?

jchapma:
> Generally, the idea behind standardizing the DN - is about making it easier for a caller to not worry about the DN case.
> 
> DN by itself is case-insensitive so it makes sense to give this functionality to DSEldif by default too. Even if it doesn't break current DSEldif usage, I think, it's better to keep the original and "DN as intended" functionality as it may be very confusing for someone in the future and it may lead to bugs and more developer time.
> 
> AS for this PR, maybe we can add an additional option for DSELdif class instance? Something like 'dn_case_sensetive' or something like that. Then it can be used intentionally when someone need it (probably, we can add a docstring comment to get_indexes method too about that).
> What do you think

My understanding was that an attbiute is case sensitive, hence my change in DSEldif. Following your explanation and a look through the code I see my understanding was incorrect. :( 

I have reverted the dn case changes in DSEldif.

droideck (commented):
I haven't look deeper 'why' but some of the **src/lib389/lib389/tests/dseldif_test.py** tests fail for me...

```
>       entry_dn_i = self._contents.index("dn: {}\n".format(entry_dn.lower()))
E       ValueError: 'dn: cn=config\n' is not in list
```

Firstyear:
> My understanding was that an attbiute is case sensitive, hence my change in DSEldif. Following your explanation and a look through the code I see my understanding was incorrect. :(

The attribute may be case sensitive but once it becomes used as the RDN dn a DN, it's treated as insensitive. It just so happens that most of the attributes we use for RDN's themself are also case insensitive as well (uid, cn, ou, dc etc). 

Honestly, I'm sure some pretty haunted things would happen if you were to use a case sensitive attr as an RDN ... 🤔

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -401,25 +401,74 @@ def test_basic_backup(topology_st, import_example_ldif):
...
+
+    log.info('Offline reindex, stopping the server')
     topology_st.standalone.stop()
-    topology_st.standalone.db2index()
-    topology_st.standalone.db2index(suffixes=[DEFAULT_SUFFIX], attrs=['uid'])
+
+    log.info('Reindex with a single index attribute')
+    topology_st.standalone.db2index(bename=DEFAULT_BENAME, attrs=['uid'])
+    assert topology_st.standalone.searchErrorsLog(info_message + 'uid')
+
```
Firstyear, line 437: Is there a way to assert here that you DIDNT reinedx the other BE's?

src/lib389/lib389/cli_ctl/dbtasks.py
```diff
@@ -95,7 +113,8 @@ def dbtasks_verify(inst, log, args):
 def create_parser(subcommands):
     db2index_parser = subcommands.add_parser('db2index', help="Initialise a reindex of the server database. The server must be stopped for this to proceed.")
     # db2index_parser.add_argument('suffix', help="The suffix to reindex. IE dc=example,dc=com.")
-    db2index_parser.add_argument('backend', help="The backend to reindex. IE userRoot")
+    db2index_parser.add_argument('backend', nargs="?", help="The backend to reindex. IE userRoot", default=False)
+    db2index_parser.add_argument('--attr', nargs="*", help="The attribute's to reindex. IE --attr aci cn givenname", default=False)
```
Firstyear, line 117: minor, "the list of attributes to reindex".

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -2586,13 +2586,6 @@ bdb_rm_db_file(backend *be, struct attrinfo *a, PRBool use_lock, int no_force_ch
     if (NULL == pEnv) { /* db does not exist */
         return rc;
     }
-    /* Added for bug 600401. Somehow the checkpoint thread deadlocked on
-     index file with this function, index file couldn't be removed on win2k.
-     Force a checkpoint here to break deadlock.
-  */
-    if (0 == no_force_checkpoint) {
-        bdb_force_checkpoint(li);
```
Firstyear, line 2594: I'm a bit hazy but why do you need to remove the checkpoint here?

dirsrvtests/tests/suites/basic/basic_test.py
jchapma: We could check the logs for the non existence of the other BE's reindex messages.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
jchapma: Short story long...
Currently when a user requests an offline reindex the CLI maps this to upgradedb, which does a reindex of each backend by doing an export and import for each backend. This is a problem for some customers as they only requested the reindex of a single backend and wanted to avoid the overhead of reindexing all backends.

When requesting an offline reindex via the server, a libdb checkpoint error occurs:
/usr/sbin/ns-slapd db2index  -D /etc/dirsrv/slapd-inst01  -n test01  -t aci

... ERR - libdb - BDB1566 txn_checkpoint interface requires an environment configured for the transaction subsystem
... ERR - bdb_force_checkpoint - Checkpoint FAILED, error Unexpected dbimpl error code (22)

The last working version of a single backend reindex was in branch 389-ds-1.4.1, so I compared the behaviour of 1.4.1 to master to understand the libdb error.

During an offline reindex the existing index file is deleted and re-generated as they server processes the specified index attributes. During the deletion of the index file a checkpoint is requested, but only if transaction logging is disabled:
https://github.com/389ds/389-ds-base/blob/aeb90eb0c41fc48541d983f323c627b2e6c328c7/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c#L4232-L4251

In 1.4.1 a checkpoint is never performed because transaction logging was disabled, but in master a checkpoint is attempted because transaction logging is not disabled, although it was requested:
https://github.com/389ds/389-ds-base/blob/aeb90eb0c41fc48541d983f323c627b2e6c328c7/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c#L1409-L1412

It turns out that some attributes specific to the BDB backend are no longer in the ldbm config entry, CONFIG_DB_TRANSACTION_LOGGING is one of them.

So to disable transaction logging in libdb we should use:
https://github.com/389ds/389-ds-base/blob/aeb90eb0c41fc48541d983f323c627b2e6c328c7/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c#L2411-L2415
Instead of:
https://github.com/389ds/389-ds-base/blob/aeb90eb0c41fc48541d983f323c627b2e6c328c7/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c#L1411

So to answer your question, we dont need to remove the checkpoint, we just need to call the function that disabled transaction loggin in libdb. I have updated this PR to reflect this.

droideck: Nice investigation!
A small note, I think it makes sense to bring back the removed comment too.

jchapma: It does, my bad. Reverted the comment.

Firstyear: Great, thanks for the follow up investigation!

droideck (approved):
LGTM!

Issue #4443, opened 2020-11-16
Internal unindexed searches in syncrepl 


abbra:
On latest Fedora Rawhide test composes I see plenty of complains about partial unindexed searches from syncrepl plugin even though there are indexes defined for the attributes specified in the filter:

Logs for the compose test: https://openqa.fedoraproject.org/tests/722674#downloads

Typical errors in the log:
```
[15/Nov/2020:18:58:04.118737200 -0500] - ERR - log_result - Internal unindexed search: source (cn=server,cn=plugins,cn=config) search base="cn=changelog" filter="(&(changenumber>=72)(targetuniqueid=cb0ce98a-279a11eb-b8b0c9d4-8de4b180))" etime=0.002251050 nentries=1  notes=U details="Partially Unindexed Filter"
[15/Nov/2020:18:58:04.136015700 -0500] - ERR - log_result - Internal unindexed search: source (cn=server,cn=plugins,cn=config) search base="cn=changelog" filter="(&(changenumber>=73)(targetuniqueid=cb0ce98a-279a11eb-b8b0c9d4-8de4b180))" etime=0.002381400 nentries=1  notes=U details="Partially Unindexed Filter"
[15/Nov/2020:18:58:04.167502775 -0500] - ERR - log_result - Internal unindexed search: source (cn=server,cn=plugins,cn=config) search base="cn=changelog" filter="(&(changenumber>=74)(targetuniqueid=cb0ce98a-279a11eb-b8b0c9d4-8de4b180))" etime=0.000606675 nentries=1  notes=U details="Partially Unindexed Filter"
[15/Nov/2020:18:58:04.187813900 -0500] - ERR - log_result - Internal unindexed search: source (cn=server,cn=plugins,cn=config) search base="cn=changelog" filter="(&(changenumber>=74)(targetuniqueid=cb0ce98a-279a11eb-b8b0c9d4-8de4b180))" etime=0.000362425 nentries=1  notes=U details="Partially Unindexed Filter"
[15/Nov/2020:18:58:04.225106850 -0500] - ERR - log_result - Internal unindexed search: source (cn=server,cn=plugins,cn=config) search base="cn=changelog" filter="(&(changenumber>=75)(targetuniqueid=cb0ce986-279a11eb-b8b0c9d4-8de4b180))" etime=0.001161925 nentries=1  notes=U details="Partially Unindexed Filter"
[...]

PR #4834, opened 2021-07-13
Issue 4443 - Internal unindexed searches in syncrepl/retro changelog

mreynolds389:
Bug Description:

When a non-system index is added to a backend it is
disabled until the database is initialized or reindexed.
So in the case of the retro changelog the changenumber index
is always disabled by default since it is never initialized.
This leads to unexpected unindexed searches of the retro
changelog.

Fix Description:

If an index has "nsSystemIndex" set to "true" then enable it
immediately.

relates:  https://github.com/389ds/389-ds-base/issues/4443

### DIFF
--- a/dirsrvtests/tests/suites/retrocl/retrocl_indexing_test.py
+++ b/dirsrvtests/tests/suites/retrocl/retrocl_indexing_test.py
@@ -1,0 +1,68 @@
+import logging
+import pytest
+import os
+from lib389._constants import RETROCL_SUFFIX, DEFAULT_SUFFIX
+from lib389.topologies import topology_st as topo
+from lib389.plugins import RetroChangelogPlugin
+from lib389.idm.user import UserAccounts
+from lib389._mapped_object import DSLdapObjects
+log = logging.getLogger(__name__)
+
+
+def test_indexing_is_online(topo):
+    """Test that the changenmumber index is online right after enabling the plugin
+
+    :id: 16f4c001-9e0c-4448-a2b3-08ac1e85d40f
+    :setup: Standalone Instance
+    :steps:
+        1. Enable retro cl
+        2. Perform some updates
+        3. Search for "(changenumber>=-1)", and it is not partially unindexed
+        4. Search for "(&(changenumber>=-1)(targetuniqueid=*))", and it is not partially unindexed
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+    """
+
+    # Enable plugin
+    topo.standalone.config.set('nsslapd-accesslog-logbuffering',  'off')
+    plugin = RetroChangelogPlugin(topo.standalone)
+    plugin.enable()
+    topo.standalone.restart()
+
+    # Do a bunch of updates
+    users = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    user_entry = users.create(properties={
+        'sn': '1',
+        'cn': 'user 1',
+        'uid': 'user1',
+        'uidNumber': '11',
+        'gidNumber': '111',
+        'givenname': 'user1',
+        'homePhone': '0861234567',
+        'carLicense': '131D16674',
+        'mail': 'user1@whereever.com',
+        'homeDirectory': '/home'
+    })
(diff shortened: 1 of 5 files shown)


src/lib389/lib389/_mapped_object.py
```diff
@@ -148,6 +148,19 @@ def exists(self):
 
         return True
 
+    def search(self, scope="subtree", filter='objectclass=*'):
```
droideck, line 151: We already have this functionality in DSLDapObjects.filter(). I think it can be safely used here.

DSLdapObject (singular instance) is for single entry operations so it makes less sense to add 'search' function here.

droideck (changes requested):
A small note but otherwise - looks good!

droideck (approved):
LGTM!

dirsrvtests/tests/suites/retrocl/retrocl_indexing_test.py
```diff
@@ -0,0 +1,68 @@
...
+        'mail': 'user1@whereever.com',
+        'homeDirectory': '/home'
+    })
+    for count in range(0, 10):
+        user_entry.replace('mail', f'test{count}@test.com')
+
+    # Search the retro cl, and check for error messages
+    filter_simple = '(changenumber>=-1)'
+    filter_compound = '(&(changenumber>=-1)(targetuniqueid=*))'
+    retro_changelog_suffix = DSLdapObjects(topo.standalone, basedn=RETROCL_SUFFIX)
```
tbordaz, line 55: I think this test is fine but the original bug was an internal search triggered by sync_repl with a cookie.

ldap/servers/slapd/back-ldbm/ldbm_index_config.c
```diff
@@ -78,6 +78,14 @@ ldbm_index_parse_entry(ldbm_instance *inst, Slapi_Entry *e, const char *trace_st
         }
     }
 
+    if (0 == slapi_entry_attr_find(e, "nsSystemIndex", &attr)) {
```
tbordaz, line 82: I think you may '*is_system_index = PR_FALSE;' by default, just in case the caller does not initialize it.

ldap/servers/slapd/back-ldbm/ldbm_index_config.c
```diff
@@ -126,17 +135,18 @@ ldbm_instance_index_config_add_callback(Slapi_PBlock *pb __attribute__((unused))
...
 
     returntext[0] = '\0';
-    *returncode = ldbm_index_parse_entry(inst, e, "from DSE add", &index_name, returntext);
+    *returncode = ldbm_index_parse_entry(inst, e, "from DSE add", &index_name, &is_system_index, returntext);
     if (*returncode == LDAP_SUCCESS) {
         struct attrinfo *ai = NULL;
         /* if the index is a "system" index, we assume it's being added by
          * by the server, and it's okay for the index to go online immediately.
          * if not, we set the index "offline" so it won't actually be used
          * until someone runs db2index on it.
```
tbordaz, line 148: May be add a comment saying that if a caller thinks it is safe to enable the index immediately, he can achieve  that via nsSystemIndex

tbordaz (approved):
The patch looks good to me, minor comments. ACK

dirsrvtests/tests/suites/retrocl/retrocl_indexing_test.py
mreynolds389: I'm basically reproducing the same search from the bug report (using presence instead of eq for targetuniqueid).  If I run this test case without the code fix, it does reproduce the unindexed search.  So it should still be valid...

ldap/servers/slapd/back-ldbm/ldbm_index_config.c
mreynolds389: That is already done by the calling function, but I can doubly enforce it in this function.

ldap/servers/slapd/back-ldbm/ldbm_index_config.c
mreynolds389: Done

Issue #4837, opened 2021-07-20
persistent search returns entries even when an error is returned by content-sync-plugin


tbordaz:
**Issue Description**
When a ldap client sends a sync request control, the server response may contain a sync state control. If the server fails to create the control the search should fail. This is already implemented in master branch but is missing on maintenance branch 1.3.10 and 1.4.3.

**Package Version and Platform:**
on 1.3.9 and 1.4.3 only (already fixed in master)

**Steps to Reproduce**
to be provided

**Expected results**
[19/Jul/2021:11:45:30.174646214 -0400] - ERR - content-sync-plugin - sync_create_state_control - Entries are missing nsuniqueid. Unable to proceed.
[19/Jul/2021:11:45:30.179232382 -0400] - ERR - send_ldap_search_entry_ext - Error 1 returned by pre entry plugins

tbordaz:
e762f35bc..7b705e743  389-ds-base-1.4.3
a7b67cfa8..a2c73838d  389-ds-base-1.3.10

Already fixed in 1.4.4 and master

PR #4838, opened 2021-07-20
Issue 4837 - persistent search returns entries even when an error is …

tbordaz:
…returned by content-sync-plugin

Bug description:
	When a ldap client sends a sync request control, the server response may contain a sync state control.
        If the server fails to create the control the search should fail.

Fix description:
	In case the server fails to create the response control
        logs the failure

relates: https://github.com/389ds/389-ds-base/issues/4837

Reviewed by:

Platforms tested:


droideck (approved):
LGTM!

Issue #4696, opened 2021-03-23
upgrade password hash on bind also causes passwordExpirationtime to be updated


mreynolds389:
**Issue Description**

There is an unintended side effect of the "upgrade password on bind" feature.  It causes the password policy code to be engaged and it resets the passwordExpirationtime in the entry.  There could be other unintended password policy side effects as well.  We should look into a way to skip password policy updates when the hash is reset by this plugin.

PR #4840, opened 2021-07-22
Issue - 4696 - Password hash upgrade on bind

jchapma:
Description:
	There is an unintended side effect of the "upgrade password
	on bind" feature. It causes the password policy code to be
	engaged and it resets the passwordExpirationtime in the entry.

Fix description:
	Only allow an external password modify operation or an extended
	password modify operation update the password info.

Relates: https://github.com/389ds/389-ds-base/issues/4696

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/modify.c
+++ b/ldap/servers/slapd/modify.c
@@ -612,6 +612,7 @@ op_shared_modify(Slapi_PBlock *pb, int pw_change, char *old_pw)
     char *proxydn = NULL;
     int proxy_err = LDAP_SUCCESS;
     char *errtext = NULL;
+    int ext_op_passwd_oid = 0;
 
     slapi_pblock_get(pb, SLAPI_ORIGINAL_TARGET, &dn);
     slapi_pblock_get(pb, SLAPI_MODIFY_TARGET_SDN, &sdn);
@@ -1020,6 +1021,14 @@ op_shared_modify(Slapi_PBlock *pb, int pw_change, char *old_pw)
         set_db_default_result_handlers(pb);
         if (be->be_modify != NULL) {
             if ((rc = (*be->be_modify)(pb)) == 0) {
+                /* Is this an extended password op */
+                if(pb_conn) {
+                    if (slapi_op_get_type(pb_conn->c_ops) == SLAPI_OPERATION_EXTENDED) {
+                        if (strcmp(pb_conn->c_ops->o_params.p.p_extended.exop_oid, EXTOP_PASSWD_OID)== 0) {
+                            ext_op_passwd_oid = 1;
+                        }
+                    }
+                }
                 /* acl is not used for internal operations */
                 /* don't update aci store for remote acis  */
                 if ((!internal_op) &&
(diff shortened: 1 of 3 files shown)


Firstyear (approved):
Ack from me :)

tbordaz:
The patch looks good but to be honest I am not fully confident with it. I think there is a risk to skip update_pw_info for all internal update. A  plugin may update userpassword and expect the call to update_pw_info.

The internal update done during a bind, to rehash the userpassword (update_pw_encoding) , uses a flag OP_FLAG_ACTION_SKIP_PWDPOLICY. Would it be possible to skip passwordExpirationtime only if this flag is set ?
Similarly to what is done https://github.com/389ds/389-ds-base/blob/master/ldap/servers/slapd/pw.c#L666.

droideck (approved):
I am not very familiar with the password change part (but from the first look it's good). So I'll leave it to Thierry and William's judgements.

And the tests look good! Acking them from my side.

jchapma:
> The patch looks good but to be honest I am not fully confident with it. I think there is a risk to skip update_pw_info for all internal update. A plugin may update userpassword and expect the call to update_pw_info.
> 
> The internal update done during a bind, to rehash the userpassword (update_pw_encoding) , uses a flag OP_FLAG_ACTION_SKIP_PWDPOLICY. Would it be possible to skip passwordExpirationtime only if this flag is set ?
> Similarly to what is done https://github.com/389ds/389-ds-base/blob/master/ldap/servers/slapd/pw.c#L666.

@tbordaz 
Yes it is possible, it is also a way simpler solution. There was a typo when initializing the pblock for the internal operation that prevented the OP_FLAG_ACTION_SKIP_PWDPOLICY flag from being set. 

I will update the PR with the tested changes, thank you.

tbordaz:
I think this fix fixes an original [typo](https://github.com/389ds/389-ds-base/commit/b84669f83f6eb14e3e5f12f8ec741957b7ba6a50#diff-f863d95380a4959e5f8c886d71c7e3f619a9401b4b6c5817827788a7defa2669R3360) 
OP_FLAG_ACTION_SKIP_PWDPOLICY and OP_FLAG_SKIP_MODIFIED_ATTRS being digit values.
ACK for me

tbordaz:
ACK

ldap/servers/slapd/modify.c
```diff
@@ -1029,11 +1029,12 @@ op_shared_modify(Slapi_PBlock *pb, int pw_change, char *old_pw)
 
                 if (operation_is_flag_set(operation, OP_FLAG_ACTION_LOG_AUDIT))
                     write_audit_log_entry(pb); /* Record the operation in the audit log */
-
+
```
mreynolds389: please remove this white space

ldap/servers/slapd/modify.c
```diff
@@ -1434,7 +1435,7 @@ optimize_mods(Slapi_Mods *smods)
             (SLAPI_IS_MOD_ADD(prev_mod->mod_op) || SLAPI_IS_MOD_DELETE(prev_mod->mod_op)) &&
             (prev_mod->mod_op == mod->mod_op) &&
             (!strcasecmp(prev_mod->mod_type, mod->mod_type)))
-        {
+            {
```
mreynolds389: Per our coding style guidelines, this bracket should be moved back to the left

tbordaz (approved):
ACK

Issue #4836, opened 2021-07-17
Potential deadlock error due to the re-acquisition to the same lock


jenny-cheung:
Dear Developers:
                First, thank you for your checking. It seems that there is an error due to the two lock statement  `pthread_mutex_lock(&notify_lock);` in the function` check_replicas_are_done_cleaning(cleanruv_data *data)` . Should the second one be the unlock statement `pthread_mutex_unlock(&notify_lock);`.

https://github.com/389ds/389-ds-base/blob/4559a89c0877e45143fa3c9cc34393d4d0cd2696/ldap/servers/plugins/replication/repl5_replica_config.c#L2197

https://github.com/389ds/389-ds-base/blob/4559a89c0877e45143fa3c9cc34393d4d0cd2696/ldap/servers/plugins/replication/repl5_replica_config.c#L2199

```
static void
check_replicas_are_done_cleaning(cleanruv_data *data)
{
    ...;

        cleanruv_log(data->task, data->rid, CLEANALLRUV_ID, SLAPI_LOG_NOTICE,
                     "Not all replicas finished cleaning, retrying in %d seconds",
                     interval);
        if (!slapi_is_shutting_down()) {
            struct timespec current_time = {0};
            clock_gettime(CLOCK_MONOTONIC, &current_time);
            current_time.tv_sec += interval;
            pthread_mutex_lock(&notify_lock);  // first time 
            pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
            pthread_mutex_lock(&notify_lock);  //second time
        }

        interval *= 2;
        if (interval >= CLEANALLRUV_MAX_WAIT) {
            interval = CLEANALLRUV_MAX_WAIT;
        }
    }
    slapi_ch_free_string(&filter);
}
```

jenny-cheung:
Hi, developers @mreynolds389 @jchapma @Firstyear 

Any comments would be highly appreciated. Thank you.

Best Regards,

jenny-cheung:
Hi, developers @mreynolds389 @jchapma @Firstyear 

Any comments would be highly appreciated. Thank you.

Best Regards,

Firstyear:
[...]

PR #4847, opened 2021-07-26
Fix a poential deadlock bug due to re-acquisition

jenny-cheung:
Fix a potential deadlock bug due to the re-acquisition on the lock notify_lock. This patch is for issue #4836.

### DIFF
--- a/ldap/servers/plugins/replication/repl5_replica_config.c
+++ b/ldap/servers/plugins/replication/repl5_replica_config.c
@@ -2196,7 +2196,7 @@ check_replicas_are_done_cleaning(cleanruv_data *data)
             current_time.tv_sec += interval;
             pthread_mutex_lock(&notify_lock);
             pthread_cond_timedwait(&notify_cvar, &notify_lock, &current_time);
-            pthread_mutex_lock(&notify_lock);
+            pthread_mutex_unlock(&notify_lock);
         }
 
         interval *= 2;


Firstyear (approved):

Firstyear:
@mreynolds389 Did you want to double check this too?

mreynolds389:
I do have a concern, this code is used quite frequently from by FreeIPA, why have we not seen a problem with it until now?   Anyway the code is correct, but I'd like to see the replication CI tests run against it first.

droideck (approved):
Thank you @jenny-cheung for the good-find PR!  

One thing only, could you please modify your commit message according to this standard:

https://www.port389.org/docs/389ds/contributing.html#getting-the-patch-ready.

Thanks!

jenny-cheung:
@droideck Should I re-commit and re-create a PR? It seems the commit message can not be modified

Firstyear:
@jenny-cheung I can take it over and just fix/merge this.

Firstyear:
Closing in favour of #4936 with corrected commit message.

jenny-cheung:
@Firstyear @mreynolds389 Thanks! Could be an exploitable vulnerability?

Firstyear:
No it can not, as it would require administrative (root) access to trigger the clean.

jenny-cheung:
@Firstyear OK, thanks

tbordaz:
> I do have a concern, this code is used quite frequently from by FreeIPA, why have we not seen a problem with it until now? Anyway the code is correct, but I'd like to see the replication CI tests run against it first.

Yes that is surprising. According to pthread_mutex_lock documentation a relock of a default mutex leads to undefined behavior. So we were lucky to not deadlock there, also probably because this is called only once (startup). Finally later the thread, in the main loop, properly lock/unlock the lock and that is where the thread spends most of the time. Nice catch.

PR #4849, opened 2021-07-27
Issue 4848 - Force to require nss version greater or equal as the ver…

### LINKED ISSUE #4848: Force to require nss version greater or equal as the version available at the build time
**Issue Description**
Recently I've seen multiple issues that were caused by the fact that nss available at the build time was greater than installed on the system:
https://lists.fedorahosted.org/archives/list/freeipa-users@lists.fedorahosted.org/message/KYKZ4BEXDNCM2BES2Y2JDNZAZBHP7VO5/
https://github.com/freeipa-pr-ci2/freeipa/pull/1060
https://bugzilla.redhat.com/show_bug.cgi?id=1986327

In our spec file we require `nss >= 3.34`, but not the exact (or greater, as they are backward compatible) version available at the build time.

We should record nss version available at the build time and require it at the runtime.

vashirov:
…sion available at the build time

Description:
In our spec file we require nss >= 3.34, but not the exact (or greater,
as they are backward compatible) version available at the build time.

Fix Description:
We should record nss version available at the build time and require it
at the runtime.
Adapt a macro from samba spec file.

Fixes: https://github.com/389ds/389-ds-base/issues/4848

Reviewed by: ???


mreynolds389:
ack

Firstyear (approved):

droideck (approved):
LGTM

PR #4850, opened 2021-07-27
Issue 4460 - Fix isLocal and TLS paths discovery

droideck:
Description: Fix isLocal inconsistency in the 'allocate' code.
And make sure that the certdir (and other TLS paths) are accessible
(has read right) before setting ldap.OPT_X_TLS_*.

Relates: https://github.com/389ds/389-ds-base/issues/4460

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -940,6 +940,9 @@ class DirSrv(SimpleLDAPObject, object):
         else:
             super(DirSrv, self).__init__(uri, trace_level=TRACE_LEVEL)
 
+        # Set new TLS context only if we changed some of the options
+        new_tls_context = False
+
         if certdir is None and self.isLocal:
             certdir = self.get_cert_dir()
             # If we are trying to manage local instance and admin doesn't have access
@@ -955,27 +958,30 @@ class DirSrv(SimpleLDAPObject, object):
             # it can NOT change opts AT ALL.
             self.log.debug("Using external ca certificate %s", certdir)
             self.set_option(ldap.OPT_X_TLS_CACERTDIR, ensure_str(certdir))
+            new_tls_context = True
 
         if userkey is not None:
             # Note this sets LDAP.OPT not SELF. Because once self has opened
             # it can NOT change opts AT ALL.
             self.log.debug("Using user private key %s", userkey)
             self.set_option(ldap.OPT_X_TLS_KEYFILE, ensure_str(userkey))
+            new_tls_context = True
 
         if usercert is not None:
             # Note this sets LDAP.OPT not SELF. Because once self has opened
             # it can NOT change opts AT ALL.
             self.log.debug("Using user certificate %s", usercert)
             self.set_option(ldap.OPT_X_TLS_CERTFILE, ensure_str(usercert))
-
+            new_tls_context = True
 
         if certdir or starttls or uri.startswith('ldaps://'):
             try:
                 # Note this sets LDAP.OPT not SELF. Because once self has opened
                 # it can NOT change opts on reused (ie restart)
                 if reqcert is not None:
-                    self.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, reqcert)
                     self.log.debug("Using lib389 certificate policy %s", reqcert)
+                    self.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, reqcert)
+                    new_tls_context = True
                 else:
                     self.log.debug("Using /etc/openldap/ldap.conf certificate policy")
                 self.log.debug("ldap.OPT_X_TLS_REQUIRE_CERT = %s", self.get_option(ldap.OPT_X_TLS_REQUIRE_CERT))
@@ -984,7 +990,8 @@ class DirSrv(SimpleLDAPObject, object):
                 raise e
 
         # Tell python ldap to make a new TLS context with this information.
-        self.set_option(ldap.OPT_X_TLS_NEWCTX, 0)
+        if new_tls_context:
+            self.set_option(ldap.OPT_X_TLS_NEWCTX, 0)
 
         if starttls and not uri.startswith('ldaps'):
             self.start_tls_s(escapehatch='i am sure')
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -531,13 +531,12 @@ class DirSrv(SimpleLDAPObject, object):
         if SER_SERVERID_PROP not in args:
             self.log.debug('SER_SERVERID_PROP not provided, assuming non-local instance')
             # The lack of this value basically rules it out in most cases
+            self.isLocal = False
             self.ds_paths = Paths(instance=self, local=self.isLocal)
         else:
             self.ds_paths = Paths(serverid=args[SER_SERVERID_PROP], instance=self, local=self.isLocal)
             # Settings from args of server attributes
             self.serverid = args.get(SER_SERVERID_PROP, None)
-            # Probably local?
-            self.isLocal = True
 
         # Do we have ldapi settings?
         # Do we really need .strip() on this?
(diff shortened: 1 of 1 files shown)


droideck:
I've posted a bigger explanation and possible improvements to this comment: https://github.com/389ds/389-ds-base/issues/4460#issuecomment-887634353

So if it'll be decided, I'll improve this PR with a final fix (which may change the current behavior a bit).

Or I'll merge this PR first (as it fixes 1.4.3 TLS issue) and I'll add another PR right after that.

src/lib389/lib389/__init__.py
```diff
@@ -938,27 +937,30 @@ def open(self, uri=None, saslmethod=None, sasltoken=None, certdir=None, starttls
             self.log.debug("Using dirsrv ca certificate %s", certdir)
 
         if certdir is not None:
-            """
-            We have a certificate directory, so lets start up TLS negotiations
-            """
+            if not os.access(ensure_str(certdir), os.R_OK):
```
Firstyear, line 951: I'm not sure this is a good way to check this. It's pretty possible someone has a local instance and is trying to manage a remote one. 

Perhaps a better way is when we parse .dsrc, that sets a flag which is like "dsrc == true" and we can check that to know if the instance is coming from local or remote?

droideck: certdir is usually tied to the instance directory. So it's highly unlikely that someone has the instances named the same locally and remotely (I'd say it's a bit of an "unhealthy" practice even)

But I agree that it's still a supported option and we shouldn't break that usage...

> Perhaps a better way is when we parse .dsrc, that sets a flag which is like "dsrc == true" and we can check that to know if the instance is coming from local or remote?

I am not sure about that option as we still will encounter this issue:

1. Admin has ~/.dsrc and tries to connect to the remote instance. /etc/openldap/ldap.conf takes care of his certs.
2. We process ~/.dsrc and it doesn't have the cert info (as it's in ldap.conf) and still, we set 'server-id' which then affects the fact that the instance is considered local (hence lib389 looks for certdir instance directory which does not exist on that remote machine)

And it's only one scenario (there are much more things that can go wrong) and it's very cryptic to actually find out if the user is remote or local. So we need a certain and explicit way for specifying is the user wants to connect remotely or locally.

What do you think about the option I mentioned in the issue? 
https://github.com/389ds/389-ds-base/issues/4460#issuecomment-887634353
"When we check for 'isLocal' - check if 'uri' has 'localhost' (and its variations) in it.
If it's localhost or the 'uri' is not defined - we set 'isLocal = True'. And if it's True - then we can get the certdir from the local path (defaults.inf and dse.ldif)"

So I can remove the 'os.access()' part from current PR and I'll add the 'URI' processing so we always have a way to find out if we are local or remote. What do you think?

droideck:
I've removed the access right change and reworked how isLocal is set.
With the change, we decide if we are locally based on LDAP URI. 
Also, general clean-up was made.

Please, review.

mreynolds389:
This doesn't work for me:  

# dsconf ldaps://localhost.localdomain:636 -D cn=dm  config get
Enter password for cn=dm on ldaps://localhost.localdomain:636: 
Error: -1 - Can't contact LDAP server - [] - error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)

/etc/openldap/ldap.conf does contain:

    TLS_CACERTDIR /etc/dirsrv/slapd-localhost

But if I set up the .dsrc file:

    [localhost]
    uri = ldaps://localhost.localdomain:636
    tls_cacertdir = /etc/dirsrv/slapd-localhost

Then dsconf works:

# dsconf localhost config get
Enter password for cn=dm on ldaps://localhost.localdomain:636: 
dn: cn=config
...

This needs to work if you are not using a .dsrc file.  Sorry I see no change in behavior with this patch.

droideck:
> This doesn't work for me:
> 
> dsconf ldaps://localhost.localdomain:636 -D cn=dm config get
> ...
> Then dsconf works:
> dsconf localhost config get

The difference between these two is that in the first case you don't specify the instance name and you try to connect locally.

So it thinks that we are local but there is no certdir to find. So it tries to use /etc/openldap/ldap.conf data but the CA certificate is not trusted (you can run *trust anchor /etc/dirsrv/slapd-localhost/18a51399.0* command and then it will work).

The initial issues (both this and bugzillas) were about remote connections. And it is fixed (I think so. But still, please, check).

Another issue is that python-ldap rejects the connections with untrusted CA (even though openldap client tool ldapsearch allows it). But it was always like that and I am not sure if it's a real issue as we really shouldn't use untrusted CA system-wide.

And if you want to manage an instance locally, you can specify the instance name and it will pick up everything correctly (even untrusted CA will be accepted as it's specified explicitly).

mreynolds389:
The bugs and upstream tickets were all local connections, and it used to work.  That is the regression as I understand it.

mreynolds389:
> The bugs and upstream tickets were all local connections, and it used to work. That is the regression as I understand it. Why does it matter if it's local or remote - we should also use what is set in /etc/openldap/ldap.conf (if .dsrc certdir is not set). That's what ldapsearch does.

I clarify my point, if you don't supply an instance name, then we should use what is in /etc/openldap/ldap.conf

droideck:
> > The bugs and upstream tickets were all local connections, and it used to work. That is the regression as I understand it. Why does it matter if it's local or remote - we should also use what is set in /etc/openldap/ldap.conf (if .dsrc certdir is not set). That's what ldapsearch does.
> 
> I clarify my point, if you don't supply an instance name, then we should use what is in /etc/openldap/ldap.conf

I am really pretty sure that it's not a regression.
I've checked it on different machines (RHEL 8.2 with 1.4.2, RHEL 8.5, and Fedora 33 (with the current branch build) - and it was not possible to connect via localhost (without instance name) on either of them (from the oldest to the newest).

But I think I found the issue.
We've been creating a new TLS context even when we don't need one (we haven't set any TLS options via ldap.set_option).
I've added the condition and we'll see if it won't break tests.

@Firstyear what do you think? Did I miss something?

src/lib389/lib389/__init__.py
Firstyear: Even checking localhost isn't enough. I have idm.example.com as a localhost alias. And 127.0.0.0/8 maps to your machine. 

I think the only way to undo this mess is that we need to stop using allocate, and have dsctl call "local_simple_allocate" and dsconf/dsidm call "remote_simple_allocate". Then for our CI/tests we have a handler that can do any magic we need.

Firstyear:
IIRc if we don't setup the context before hand there was some other bug in ldap we would hit where you could never load the certs later I think ... TLS in ldap is a hot mess :(

src/lib389/lib389/__init__.py
droideck: > Even checking localhost isn't enough. I have idm.example.com as a localhost alias. And 127.0.0.0/8 maps to your machine.
> 
> I think the only way to undo this mess is that we need to stop using allocate, and have dsctl call "local_simple_allocate" and dsconf/dsidm call "remote_simple_allocate". Then for our CI/tests we have a handler that can do any magic we need.

I use 'isLocalhost' function, and it tries to resolve the hostname aliases - https://github.com/389ds/389-ds-base/blob/4559a89c0877e45143fa3c9cc34393d4d0cd2696/src/lib389/lib389/utils.py#L675

I was considering using remote_simple_allocate/local_simple_allocate but it really feels like too much risk for 1.4.3 (which is not working correctly now).
I agree that it's better to rewrite our CLI tools but this PR is about fixing the issue for as many possible situations as it can. And I think we should merge it because our users need it.

droideck:
> IIRc if we don't setup the context before hand there was some other bug in ldap we would hit where you could never load the certs later I think ... TLS in ldap is a hot mess :(

If I understand the python-ldap library correctly, it's intended to be used in a certain way:

1. Set ldap.OPT_X_TLS_* options (like certdir, etc.)
2. Create a new context with these values (it should be done after the first step) - setting OPT_X_TLS_NEWCTX option.
3. If no ldap.OPT_X_TLS_* set and there is no new TLS context - python-ldap will use system-wide context (/etc/openldap/ldap.conf)

I looked through our code and I don't see any places we set certificates after OPT_X_TLS_NEWCTX.
And if we'll do this in the future, I think, it'll be a misconfiguration. And if we change TLS options we should create a new context (maybe with a new TLS connection).

Also, all of the tests work now and everything passes.

I think we should still use OPT_X_TLS_NEWCTX as intended and we can improve our logic if we stumble on some issue where we need to load the certs into the existing connection (and then, I think, we should fix it by improving the code and update the context with new OPT_X_TLS_NEWCTX).

So I think we should merge it as it is now because it's what our users/customers need now (working dsconf tool).

Another option can be - I'll revert OPT_X_TLS_NEWCTX change and then we won't able to connect locally via ldaps://localhost:636 (without instance name) while using /etc/openldap/ldap.conf settings.

@mreynolds389 I'll give the final decision to you...

mreynolds389:
I like the way the patch is working now, ack from me.  This is definitely an improvement from how it was working before and it will solve many customer issues.

Firstyear:
I approve too :) Thanks @droideck !

PR #4852, opened 2021-08-02
Issue 4623 - RFE - Monitor the current DB locks ( nsslapd-db-current-…

bsimonova:
…locks )

Description:
Added additional tests for DB locks monitoring to check if invalid
values are correctly rejected for nsslapd-db-locks and
nsslapd-db-locks-monitoring-threshold.

Relates: https://github.com/389ds/389-ds-base/issues/4623

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/monitor/db_locks_monitor_test.py
+++ b/dirsrvtests/tests/suites/monitor/db_locks_monitor_test.py
@@ -27,6 +27,7 @@ from lib389.plugins import AttributeUniquenessPlugin
 from lib389.config import BDB_LDBMConfig
 from lib389.monitor import MonitorLDBM
 from lib389.topologies import create_topology, _remove_ssca_db
+from lib389.utils import ds_is_older
 
 pytestmark = pytest.mark.tier2
 db_locks_monitoring_ack = pytest.mark.skipif(not os.environ.get('DB_LOCKS_MONITORING_ACK', False),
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/monitor/db_locks_monitor_test.py
```diff
@@ -249,3 +250,62 @@ def test_exhaust_db_locks_big_pause(topology_st_fn, setup_attruniq_index_be_impo
...
+
+@pytest.mark.ds4623
+@pytest.mark.bz1812286
+@pytest.mark.skipif(ds_is_older("1.4.3.23"), reason="Not implemented")
+def test_invalid_threshold_range(topology_st_fn):
+    """Test that setting nsslapd-db-locks-monitoring-threshold to 60 % is rejected
+
+    :id: e4551de1-8582-4c13-b59d-3d5ec4701457
+    :setup: Standalone instance
+    :steps: 1. Set nsslapd-db-locks-monitoring-threshold to 60 %
```
droideck, line 269: I think, instead of hardcoding one value, we can parametrize the test with different invalid values from 1 to 69 range. As it may be that the feature may break at different places (hopefully not).

droideck (approved):
Besides the small suggestion (which is up to you) - LGTM!

dirsrvtests/tests/suites/monitor/db_locks_monitor_test.py
bsimonova: Good idea, thanks!

bsimonova: @droideck 
I added additional values to parametrize and fixed the docstrings.
Please review.

droideck (approved):
Great! You have my ack!
Thanks!

PR #4854, opened 2021-08-02
Issue 4778 - [RFE] Schedule execution of "compactdb" at specific date…

bsimonova:
…/time

Description:
Added additional tests to check invalid and no compaction value for
nsslapd-db-compactdb-interval.

Relates: https://github.com/389ds/389-ds-base/issues/4778

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/config/compact_test.py
+++ b/dirsrvtests/tests/suites/config/compact_test.py
@@ -5,7 +5,9 @@ import time
 from lib389.tasks import DBCompactTask
 from lib389.backend import DatabaseConfig
 from lib389.topologies import topology_m1 as topo
+from lib389.utils import ldap, ds_is_older
 
+pytestmark = pytest.mark.tier2
 log = logging.getLogger(__name__)
 
 
@@ -13,12 +15,13 @@ def test_compact_db_task(topo):
     """Test creation of dbcompact task is successful
 
     :id: 1b3222ef-a336-4259-be21-6a52f76e1859
+    :customerscenario: True
     :setup: Standalone Instance
     :steps:
         1. Create task
         2. Check task was successful
         3. Check errors log to show task was run
-        4. Create task just for replication
+        4. Create task just for changelog
     :expectedresults:
         1. Success
         2. Success
@@ -53,6 +56,7 @@ def test_compaction_interval_and_time(topo):
     """Test dbcompact is successful when nsslapd-db-compactdb-interval and nsslapd-db-compactdb-time is set
 
     :id: f361bee9-d7e7-4569-9255-d7b60dd9d92e
+    :customerscenario: True
     :setup: Supplier Instance
     :steps:
         1. Configure compact interval and time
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/config/compact_test.py
```diff
@@ -72,6 +74,62 @@ def test_compaction_interval_and_time(topo):
...
+    inst = topo.ms["supplier1"]
+    msg = 'value 2147483650 for attr nsslapd-db-compactdb-interval is greater than the maximum 2147483647'
+    config = DatabaseConfig(inst)
+
+    try:
+        config.set([('nsslapd-db-compactdb-interval', '2147483650'), ('nsslapd-db-compactdb-time', '00:01')])
+    except ldap.UNWILLING_TO_PERFORM as e:
+        log.info('Got expected error: {}'.format(str(e)))
+        assert msg in str(e)
+        time.sleep(6)
```
mreynolds389: Can you reduce this sleep time?  Maybe 3 seconds?  These "sleeps" add up when running the entire test suite.

bsimonova: I reduced the sleep time, please review. Thanks!

bsimonova:
@mreynolds389 
I updated the sleep time, please review. Thanks!

dirsrvtests/tests/suites/config/compact_test.py
```diff
@@ -72,6 +76,64 @@ def test_compaction_interval_and_time(topo):
     inst.deleteErrorLogs(restart=False)
 
 
+@pytest.mark.ds4778
+@pytest.mark.bz1748441
+@pytest.mark.skipif(ds_is_older("1.4.3.23"), reason="Not implemented")
+def test_no_compaction(topo):
```
tbordaz, line 82: Something not clear to me. compaction occurs on checkpointing (via checkpointing thread dynamic) and on compact task.
When sleeping 3 seconds are we waiting for the checkpointing to occur ? But if there is no update, is the checkpointing still trying to compact ?
I would rather expect to trigger the compact via the task, then check the message in the error log. But here I see no task.

bsimonova: @tbordaz 
I took the test_compaction_interval_and_time from Mark which sets nsslapd-db-compactdb-interval and nsslapd-db-compactdb-time. Then it waits and checks logs that compaction occurred. 
What I did in my test was that I set nsslapd-db-compactdb-interval to 0, which should not trigger any compaction and then I check the logs to confirm there was none. I just decreased the time.sleep as Mark suggested.
So I would say yes, we are waiting for the checkpointing if I understand correctly.

tbordaz: @bsimonova , sorry my fault. Checkpointing thread runs periodically (2.5sec) and if nsslapd-db-compactdb-interval is not null it calls a bdb_compact that logs "Compacting databases" and calls database compaction. So even if database compaction does nothing (no updates), the log is present.
Sorry for the noise

tbordaz (approved):

PR #4855, opened 2021-08-02
Issue 4169 - Migrate Replication & Schema tabs to PF4

mreynolds389:
Description:  Migrate the remaining components in the repl and schema tabs to PF4

### DIFF
--- a/src/cockpit/389-console/src/lib/replication/replChangelog.jsx
+++ b/src/cockpit/389-console/src/lib/replication/replChangelog.jsx
@@ -3,18 +3,17 @@ import React from "react";
 import PropTypes from "prop-types";
 import { log_cmd } from "../tools.jsx";
 import {
-    Col,
-    ControlLabel,
-    Form,
-    FormControl,
-    noop,
-    Row,
-    Spinner,
-} from "patternfly-react";
-import {
     Button,
     Checkbox,
-    Tooltip
+    Form,
+    FormSelect,
+    FormSelectOption,
+    Grid,
+    GridItem,
+    NumberInput,
+    Spinner,
+    Tooltip,
+    noop
 } from '@patternfly/react-core';
 import OutlinedQuestionCircleIcon from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
 
@@ -25,19 +24,41 @@ export class Changelog extends React.Component {
             loading: false,
             errObj: {},
             showConfirmDelete: false,
-            saveOK: false,
+            saveBtnDisabled: true,
             // Changelog settings
-            clMaxEntries: this.props.clMaxEntries,
-            clMaxAge: this.props.clMaxAge.slice(0, -1),
-            clMaxAgeUnit: this.props.clMaxAge.slice(-1).toLowerCase(),
-            clTrimInt: this.props.clTrimInt,
+            clMaxEntries: Number(this.props.clMaxEntries) == 0 ? -1 : Number(this.props.clMaxEntries),
+            clMaxAge: Number(this.props.clMaxAge.slice(0, -1)) == 0 ? -1 : Number(this.props.clMaxAge.slice(0, -1)),
+            clMaxAgeUnit: this.props.clMaxAge != "" ? this.props.clMaxAge.slice(-1).toLowerCase() : "s",
+            clTrimInt: Number(this.props.clTrimInt) == 0 ? -1 : Number(this.props.clTrimInt),
             clEncrypt: this.props.clEncrypt,
             // Preserve original settings
-            _clMaxEntries: this.props.clMaxEntries,
-            _clMaxAge: this.props.clMaxAge.slice(0, -1),
-            _clMaxAgeUnit: this.props.clMaxAge.slice(-1).toLowerCase(),
-            _clTrimInt: this.props.clTrimInt,
+            _clMaxEntries: Number(this.props.clMaxEntries) == 0 ? -1 : Number(this.props.clMaxEntries),
+            _clMaxAge: Number(this.props.clMaxAge.slice(0, -1)) == 0 ? -1 : Number(this.props.clMaxAge.slice(0, -1)),
+            _clMaxAgeUnit: this.props.clMaxAge != "" ? this.props.clMaxAge.slice(-1).toLowerCase() : "s",
+            _clTrimInt: Number(this.props.clTrimInt) == 0 ? -1 : Number(this.props.clTrimInt),
             _clEncrypt: this.props.clEncrypt,
+        };
+
+        this.minValue = -1;
+        this.maxValue = 20000000;
+
+        this.onMinus = (id) => {
+            this.setState({
+                [id]: Number(this.state[id]) - 1
+            }, () => { this.validateSaveBtn() });
+        };
+
+        this.onChange = (event, id) => {
+            const newValue = isNaN(event.target.value) ? 0 : Number(event.target.value);
+            this.setState({
+                [id]: newValue > this.maxValue ? this.maxValue : newValue < this.minValue ? this.minValue : newValue
+            }, () => { this.validateSaveBtn() });
+        };
+
+        this.onPlus = (id) => {
+            this.setState({
+                [id]: Number(this.state[id]) + 1
+            }, () => { this.validateSaveBtn() });
         };
 
         this.handleChange = this.handleChange.bind(this);
@@ -84,7 +105,7 @@ export class Changelog extends React.Component {
                         );
                         this.setState({
                             saving: false,
-                            saveOK: false,
+                            saveBtnDisabled: true,
                         });
                     })
                     .fail(err => {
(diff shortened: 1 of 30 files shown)


src/cockpit/389-console/src/lib/replication/replChangelog.jsx
```diff
@@ -25,21 +24,43 @@ export class Changelog extends React.Component {
...
             clEncrypt: this.props.clEncrypt,
             // Preserve original settings
-            _clMaxEntries: this.props.clMaxEntries,
-            _clMaxAge: this.props.clMaxAge.slice(0, -1),
-            _clMaxAgeUnit: this.props.clMaxAge.slice(-1).toLowerCase(),
-            _clTrimInt: this.props.clTrimInt,
+            _clMaxEntries: Number(this.props.clMaxEntries) == 0 ? -1 : Number(this.props.clMaxEntries),
+            _clMaxAge: Number(this.props.clMaxAge.slice(0, -1)) == 0 ? -1 : Number(this.props.clMaxAge.slice(0, -1)),
+            _clMaxAgeUnit: this.props.clMaxAge != "" ? this.props.clMaxAge.slice(-1).toLowerCase() : "s",
+            _clTrimInt: Number(this.props.clTrimInt) == 0 ? -1 : Number(this.props.clTrimInt),
```
droideck, line 38: changelog fields are empty if encryption is changed and saved.

1. Set 'Changelog Maximum Entries' to 1000;
2. Click on Encryption checkbox;
3. Save Settings

Everything disappears and then appears but the Age and Interval fields are empty (they were '-1')

src/cockpit/389-console/src/lib/replication/replTables.jsx
```diff
@@ -239,20 +239,17 @@ class ManagerTable extends React.Component {
...
-                    rows={this.state.rows}
-                    variant={TableVariant.compact}
-                    sortBy={this.state.sortBy}
-                    onSort={this.onSort}
-                >
-                    <TableHeader />
-                    <TableBody />
-                </Table>
-            </div>
+            <Table
```
droideck, line 244: Small thing.
If we have an error then the error has a different font and it's not an italic. I think it's too much different but it's up to you if you want to fix this.
image

src/cockpit/389-console/src/lib/replication/replChangelog.jsx
droideck: image

src/cockpit/389-console/src/lib/replication/replTasks.jsx
```diff
@@ -429,30 +420,30 @@ export class ReplRUV extends React.Component {
                 </div>
                 <hr />
                 <h4 className="ds-margin-top-xlg">Create Replica Initialization LDIF File</h4>
```
droideck: I wonder if should at least display the server LDIF directory path here...
Ideally, it will be list the LDIFs from the directory, IMO.
 What do you think?

src/cockpit/389-console/src/lib/replication/winsyncAgmts.jsx
```diff
@@ -1291,25 +1333,35 @@ export class WinsyncAgmts extends React.Component {
                     item={this.state.agmtName}
                     checked={this.state.modalChecked}
                     mTitle="Initialize Winsync Agreement"
-                    mMsg="Are you sure you want to initialize this winsync agreement"
+                    mMsg="Are you sure you want to initialize this winsync agreement?"
```
droideck, line 1339: When I click on Initialize Winsync Agreement button - nothing happens. I still can close the confirmation modal window but 'Initialize Agreement' button doesn't work.

P.S. I've created a dummy WinSync agreement for the testing purposes

droideck: When I close the modal window - it shows that it tries to initialize (Initializing is spinning). But it's spinning and spinning indefinitely.

src/cockpit/389-console/src/lib/schema/schemaModals.jsx
```diff
@@ -411,60 +395,62 @@ class AttributeTypeModal extends React.Component {
...
-                                </ControlLabel>
-                            </Col>
-                            <Col sm={8}>
-                                <FormControl
-                                    id="atName"
-                                    type="text"
+                    <Form isHorizontal autoComplete="off">
+                        <Grid title="Name of the attribute">
+                            <GridItem className="ds-label" span={3}>
+                                Attribute Name
```
droideck, line 400: I'm trying to add an attribute.
There are two mandatory fields 'Name' and 'Description'. I set both to 'test'.

After I click 'Save' I get an error (which is described in details in the Console log, but actual error notification is not descriptive).

I think we can make the error more verbose so the user knows what's wrong.

Also, I am not sure what is wrong with the attribute I've tried to create. Maybe it's an error in our tools...
Please, check the screenshot (error is yellow in logs)

image

droideck (changes requested):
Great job!
A few comments from me but mostly minor. :)

src/cockpit/389-console/src/lib/replication/replChangelog.jsx
mreynolds389: Fixed, on a side note, do you like these number inputs?   They are new to PF4, otherwise it would be as text input of type "number" - like what we use everywhere ("Server Tuning" page, etc).

src/cockpit/389-console/src/lib/replication/replTables.jsx
mreynolds389: Fixed, I also stopped the "refreshing" message that causes the page to constantly bounce during initialization

src/cockpit/389-console/src/lib/replication/replTasks.jsx
mreynolds389: I agree it can be improved.  I like the table idea.  Ok let me look into it...

src/cockpit/389-console/src/lib/replication/winsyncAgmts.jsx
mreynolds389: Fixed, but the endless spinning might be expected (and temporary).  It will probably timeout eventually.  For me it did not spin long but I used a invalid hostname "h" so it probably just failed right away.

src/cockpit/389-console/src/lib/schema/schemaModals.jsx
mreynolds389: Fixed, and fixed a bug, "top" is not an attribute

mreynolds389:
Ok changes made, and pushed.  For the "replica export', I added a table to the modal showing existing LDIF's.  Thoughts?

src/cockpit/389-console/src/lib/schema/schemaModals.jsx
droideck: If I first create an attribute (with only Name and Description), then when I try to edit and set Parent Attribute - the browser crashes with 'Uncaught TypeError: s.startsWith is not a function' error

droideck:
> Ok changes made, and pushed. For the "replica export', I added a table to the modal showing existing LDIF's. Thoughts?

Looks great!
Maybe one thing, - if a user sees the list of LDIFs he may want to check the LDIF locally. And it'll be nice to have the LDIF dir path written there. What do you think?

mreynolds389:
> > Ok changes made, and pushed. For the "replica export', I added a table to the modal showing existing LDIF's. Thoughts?
> 
> Looks great!
> Maybe one thing, - if a user sees the list of LDIFs he may want to check the LDIF locally. And it'll be nice to have the LDIF dir path written there. What do you think?

The path is stated as being the server's LDIF directory.  They can check the path location on the server settings page.  Perhaps the LDIF table is too much.  Should I just remove it?

droideck:
> > > Ok changes made, and pushed. For the "replica export', I added a table to the modal showing existing LDIF's. Thoughts?
> > 
> > 
> > Looks great!
> > Maybe one thing, - if a user sees the list of LDIFs he may want to check the LDIF locally. And it'll be nice to have the LDIF dir path written there. What do you think?
> 
> The path is stated as being the server's LDIF directory. They can check the path location on the server settings page. Perhaps the LDIF table is too much. Should I just remove it?

My main concern is that it feels like a dead end.
I'll explain... 

A user enters the Export LDIF modal window. He exports Replica LDIF. And... That's it... He doesn't have any information where he can find the LDIF directory and he needs to go manually there if he wants to manage it.

Web Application, IMO, should have a lot of cross-links and discoverability.
So the user doesn't need to go manually to Server Tab to check the path. Or go to Database tab and manage the LDIF table there.

So I see a few options here:

- Add the same table from Database tab to Replica Export modal window so the user can Import/Delete LDIFs right there;
- Add a hyperlink to other tabs (Database tab with LDIF table or Server tab so the user can find the path in one click);
- Leave it as it is not write down the LDIF path so the user doesn't need to do extra actions and trying to remember where to look and then find it.
- Remove the table completely and leave it as it was. But then we may get a user experience review in the future...

Personally, I prefer the first option. But I leave it up to you :)
Anyway, the UI looks really good and all of this - is just a minor polishing :)

mreynolds389:
Now I'm just leaning towards removing the "export replica" section entirely.  We don't need 3 different places where the database can be exported and imported.  There is an option in the database section for including the replication data in the export.

droideck:
> Now I'm just leaning towards removing the "export replica" section entirely. We don't need 3 different places where the database can be exported and imported. There is an option in the database section for including the replication data in the export.

Sounds good to me! There will be no place for user confusion if there'll be no place :D
Jokes aside, I agree that one place is enough.

jchapma:
**Browser: Firefox - 89.0.2**

**Replication tab**
I'm getting an Ooops when I load the replication tab, where it just hangs.
Screenshot from 2021-08-04 16-26-06


**Schema tab**
Objectclasses, Attributes, Matching Rules sub tabs.
Is it possible to allow the user scroll down the list of elements in each of the above sub tabs. Maybe its my setup but I cannot, also the add button is inaccessible. When I use the filter it does reduce the number of elements displayed, which does help.

Here is a screen shot of what I can see at 100%
Screenshot from 2021-08-04 16-31-30

Other than that it works for me.

mreynolds389:
@jchapma - looks like you are not testing the patch correctly.  The UI from your screenshots is using PF3, not PF4.  So something is definitely wrong with your setup.  Looks like you are testing this against 1.4.3 and not master branch...

jchapma:
> @jchapma - looks like you are not testing the patch correctly. The UI from your screenshots is using PF3, not PF4. So something is definitely wrong with your setup. Looks like you are testing this against 1.4.3 and not master branch...

@mreynolds389 - I cloned your PR and rebuilt the console, but forgot to restart cockpit to pick up the changes, apologies.

jchapma:
Had a play with the repl and schema tabs, looks and feels great to me. 

In schema -> Matching Rules, there are two matching rules with no description, I had a search for them and I understand they are 
language collations, so I'm not sure if they even need a description.

image

mreynolds389:
> Had a play with the repl and schema tabs, looks and feels great to me.
> 
> In schema -> Matching Rules, there are two matching rules with no description, I had a search for them and I understand they are
> language collations, so I'm not sure if they even need a description.
> 

Yeah I've also seen these non-description MR's.  I can probably improve that.  Let me see what I can come up with...

mreynolds389:
Changes made, please review...

jchapma:
**Schema tab**

I notice some behavior with the row collapse/expand functionality of the Objectclasses, Attributes and Matching Rules tables. I guess there is some mismatch with the row key/table being selected. 

For example:
When I click to expand any row on page 1 it works fine, but if i move to another page, say page 5, and click to expand row N, the row isn't expanded on page 5, instead row N on page 1 is expanded.

mreynolds389:
> **Schema tab**
> 
> I notice some behavior with the row collapse/expand functionality of the Objectclasses, Attributes and Matching Rules tables. I guess there is some mismatch with the row key/table being selected.
> 
> For example:
> When I click to expand any row on page 1 it works fine, but if i move to another page, say page 5, and click to expand row N, the row isn't expanded on page 5, instead row N on page 1 is expanded.

Oh boy, this means there is a problem with all the expanded lists in the UI :-(  Looking into it

mreynolds389:
> > **Schema tab**
> > I notice some behavior with the row collapse/expand functionality of the Objectclasses, Attributes and Matching Rules tables. I guess there is some mismatch with the row key/table being selected.
> > For example:
> > When I click to expand any row on page 1 it works fine, but if i move to another page, say page 5, and click to expand row N, the row isn't expanded on page 5, instead row N on page 1 is expanded.
> 
> Oh boy, this means there is a problem with all the expanded lists in the UI :-( Looking into it

Nice catch @jchapma !!!!!   Fixed!

Please review..

droideck (approved):
Looked through it one more time and it's great!
LGTM!

jchapma:
Look great to me, you have my ack.

Issue #4736, opened 2021-04-23
Errors from certutil are not propagated


vashirov:
**Issue Description**
When an error occurs with `certutil`, stderr output from it is not printed:
```
[root@fedora ~]# dsctl localhost tls generate-server-cert-csr -s "bad" 
Error: Command '['/usr/bin/certutil', '-R', '--keyUsage', 'digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment', '--nsCertType', 'sslClient,sslServer', '--extKeyUsage', 'clientAuth,serverAuth', '-s', 'bad', '-8', 'fedora', '-g', '4096', '-d', '/etc/dirsrv/slapd-localhost', '-z', '/etc/dirsrv/slapd-localhost/noise.txt', '-f', '/etc/dirsrv/slapd-localhost/pwdfile.txt', '-a', '-o', '/etc/dirsrv/slapd-localhost/Server-Cert.csr']' returned non-zero exit status 255.
``` 
It just states that the exit status is 255, but no `certutil` output, which in this case is 
```
certutil -s: improperly formatted name: "bad"
```

In `lib389/nss_ssl.py` we have many places with statements like this that show this behaviour:
```python
        result = ensure_str(check_output(cmd, stderr=subprocess.STDOUT))
        self.log.debug("nss output: %s", result)
```

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: 389-ds-base-2.0.3-3.fc34.x86_64

**Steps to Reproduce**
1. `dsctl localhost tls generate-server-cert-csr -s "bad" `

**Expected results**
Instead of a python traceback, a clear error message should be printed.

tbordaz:
Create a BZ RHDS

mreynolds389:
Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1974226

mreynolds389:
4bd1c9408..b99236af6  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
a68decb96..c0dd585d7  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

vashirov:
Here `certutil` is called twice:
https://github.com/389ds/389-ds-base/blob/33c815889ec3bc8eacc7df68d57370ce8cee32e0/src/lib389/lib389/nss_ssl.py#L198-L200

On the second call it asks for the password again, because it thinks it is a db upgrade.

mreynolds389:
[...]

PR #4856, opened 2021-08-02
Issue 4736 - CLI - Errors from certutil are not propagated

mreynolds389:
Description:  Errors from certutil are not returned to the client, and
only a generic failure code is returned.  The actual error text should be
returned to the client since it has meaning.  Just catch all th
expecptions and return the exception output as a ValueError.

relates: https://github.com/389ds/389-ds-base/issues/4736

Reviewed by: ?


Firstyear (approved):

mreynolds389:
Thanks @Firstyear !

Issue #4773, opened 2021-05-20
[RFE] Enable interval feature of DNA plugin to support  intervals sizes for subuids


tbordaz:
DNA plugin should support dnaInterval. This attribute is helpful for support [sub uids/gids](https://pagure.io/freeipa/issue/8361).
Actually DNA plugin already implement that attribute in stepping next granted values and management of ranges. The only restriction is that the setting was disabled (into ifdef) and so step was always dnaInterval=1

**Describe the solution you'd like**
suppress DNA_ENABLE_INTERVAL or define DNA_ENABLE_INTERVAL in dna.c

**Describe alternatives you've considered**
Long discussion in BZ. THe possibility to dereference the dnaInterval into the entries was abandoned as it will break range management (granting/acquisition)

mreynolds389:
3cbad9e89..58a1591b9  master -> master
e3104967d..a80d67704  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
3eeeb1c7f..4cd55311f  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

Still need CI tests

mreynolds389:
CI test

a08540a82..bfb3880cf  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
aa1bceb76..703b39aef  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4857, opened 2021-08-03
Issue 4773 - Extend CI tests for DNA interval assignment

aadhikar:
Description: Extend CI tests for DNA interval assignment

Relates: https://github.com/389ds/389-ds-base/issues/4773

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/plugins/dna_interval_test.py
+++ b/dirsrvtests/tests/suites/plugins/dna_interval_test.py
@@ -15,15 +15,50 @@ from lib389.plugins import DNAPlugin, DNAPluginConfigs
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccounts
 from lib389.topologies import topology_st
+from lib389.utils import *
 
 pytestmark = pytest.mark.tier1
 
 log = logging.getLogger(__name__)
 
-def test_dna_interval(topology_st):
+
+@pytest.fixture(scope="function")
+def dna_plugin(topology_st, request):
+    inst = topology_st.standalone
+    plugin = DNAPlugin(inst)
+    ous = OrganizationalUnits(inst, DEFAULT_SUFFIX)
+    ou_people = ous.get("People")
+
+    log.info("Add dna plugin config entry...")
+    configs = DNAPluginConfigs(inst, plugin.dn)
+    dna_config = configs.create(properties={'cn': 'dna config',
+                                            'dnaType': 'uidNumber',
+                                            'dnaMaxValue': '1000',
+                                            'dnaMagicRegen': '-1',
+                                            'dnaFilter': '(objectclass=top)',
+                                            'dnaScope': ou_people.dn,
+                                            'dnaNextValue': '10',
+                                            'dnaInterval': '10'})
+
+    log.info("Enable the DNA plugin and restart...")
+    plugin.enable()
+    inst.restart()
+
+    def fin():
+        inst.stop()
+        dse_ldif = DSEldif(inst)
+        dse_ldif.delete_dn(f'cn=dna config,{plugin.dn}')
+        inst.start()
+    request.addfinalizer(fin)
+
+    return dna_config
+
+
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/plugins/dna_interval_test.py
```diff
@@ -15,12 +15,44 @@
...
+                                            'dnaScope': ou_people.dn,
+                                            'dnaNextValue': '10',
+                                            'dnaInterval': '10'})
+
+    log.info("Enable the DNA plugin and restart...")
+    plugin.enable()
+    inst.restart()
+
+    def fin():
+        dna_config.delete()
```
aadhikar: Need to find a better way of removing the plugin's config as it is making the tests flaky. Sometimes it is making the server go under a shutdown state. A solution which I can think of: we should check if the server is up and if yes then shut it down remove the entries and restart it.

dirsrvtests/tests/suites/plugins/dna_interval_test.py
```diff
@@ -82,3 +94,98 @@ def test_dna_interval(topology_st):
...
+        'uidNumber': '-1',  # Magic regen value
+        'gidNumber': '222',
+        'givenname': 'interval2',
+        'homePhone': '0861234567',
+        'carLicense': '131D16674',
+        'mail': 'interval2@whereever.com',
+        'homeDirectory': '/home/interval2'})
+
+    # Verify DNA works
+    assert user.get_attr_val_utf8_l('uidNumber') == '90'
```
aadhikar, line 135: Additional checks to make sure that DNA is working correctly. It can be removed if not needed

dirsrvtests/tests/suites/plugins/dna_interval_test.py
```diff
@@ -82,3 +94,98 @@ def test_dna_interval(topology_st):
...
+            'uidNumber': '-1',  # Magic regen value
+            'gidNumber': '333',
+            'givenname': 'interval3',
+            'homePhone': '0861234567',
+            'carLicense': '131D16674',
+            'mail': 'interval3@whereever.com',
+            'homeDirectory': '/home/interval3'})
+
+        # Verify DNA works
+        assert user.get_attr_val_utf8_l('uidNumber') == '10'
```
aadhikar, line 187: Same as Above

dirsrvtests/tests/suites/plugins/dna_interval_test.py
```diff
@@ -82,3 +96,94 @@ def test_dna_interval(topology_st):
...
+    :expectedresults:
+        1. Operation should fail with SERVER_DOWN
+        2. Operation should fail with INVALID_SYNTAX
+        3. Success
+        4. Success
+        5. Operation should fail with OPERATIONS_ERROR
+    """
+    log.info("Make the config changes needed to test dnaInterval")
+    if attr_value == '0':
+        with pytest.raises(ldap.SERVER_DOWN):
```
tbordaz: Why server down ! does the test crash the server ?

tbordaz:
Regarding replication test, it should implement this  [scenario](https://directory.fedoraproject.org/docs/389ds/design/dna-plugin.html#interval-method)

In short if you have 3 suppliers with the same range [1-1000] and interval 3. But with a different initial next value: supplier1: 1, supplier2: 2, and supplier3: 3. 
Supplier1 will assign: 1, 4, 7,...
Supplier2 will assign: 2, 5, 8...
Supplier3 will assign: 3, 6, 9,...

So they should be able to share the same range without stepping on each other

dirsrvtests/tests/suites/plugins/dna_interval_test.py
aadhikar: Yes, it does when the value is 0 if this is not expected then it's a bug.

tbordaz: Congratulation you found a nice one

sgouvern:
Please add the :customerscenario: True to all the docstrings, as these are customer cases.
appart from that LGTM !

PR #4860, opened 2021-08-05
Issue 4859 - Don't version libns-dshttpd

### LINKED ISSUE #4859: Don't version libns-dshttpd
**Issue Description**
On every build `libns-dshttpd` has a version corresponding to the package version, e.g. `libns-dshttpd-2.0.7.so`.
It's unnecessary, as we are the only consumers of this library and we don't change its ABI on every build.
It also triggers rpmdiff test failures that have to be waived on each build.

vashirov:
Description:
On every build libns-dshttpd has a version corresponding
to the package version, e.g. libns-dshttpd-2.0.7.so.
It's unnecessary, as we are the only consumers of this library
and we don't change its ABI on every build.
It also triggers rpmdiff test failures that have to be waived on
each build.

Fixes: https://github.com/389ds/389-ds-base/issues/4859

Reviewed by: ???


mreynolds389:
LGTM

PR #4871, opened 2021-08-11
Issue 4763 - Attribute Uniqueness Plugin uses wrong subtree on ModRDN

### LINKED ISSUE #4763: Attribute Uniqueness Plugin uses wrong subtree on ModRDN
**Issue Description**
When using the Attribute uniqueness plugin, restricted to one subtree, moving an object with an already existing attribute to this subtree does not raise any exceptions. It appears that the originating subtree is searched instead.

 - Platform: Containerized Fedora (tested on 33 and 34)
 - Package and version: Tested with all versions >= 2.0.2

**Steps to Reproduce**

1. Create two container objects e.g `ou=c1,dc=example,dc=com` and `ou=c2,dc=example,dc=com`
2. Create an entry for the attr-uniq plugin restricted to the `c1` subtree, restricting (for example) the mail attribute.
3. Create two objects with the same value for `mail` in `c2` (this should be allowed).
4. Move one object to `c1`
5. Move the second object to `c1`. This should raise an exeption as another object with the same value for `mail` already exists in `c1` but works without issues.
6. Additionally, trying to move one of these objects back to `c2` raises an exception even though there is no uniqueness constraint on `c2`.

**Expected results**
The attr-uniq plugin should prevent invalid move operations.

**Additional context**
I've tried to get to the root cause of this issue, and debug logs suggest that https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/uiduniq/uid.c#L1395 is called with the old `parentDN`. However, as this is my first time digging through the codebase I was unable to figure out why this happens.

droideck:
Bug Description: When using the Attribute uniqueness plugin, restricted
to one subtree, moving an object with an already existing attribute
to this subtree does not raise any exceptions. It appears that the
originating subtree is searched instead.

Fix Description: Use parent DN of the new entry when searching
for attribute uniqueness.
Add test to plugins/attruniq_test.py suite.

Fixes: https://github.com/389ds/389-ds-base/issues/4763

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/plugins/attruniq_test.py
+++ b/dirsrvtests/tests/suites/plugins/attruniq_test.py
@@ -1,0 +1,83 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import pytest
+import ldap
+import logging
+from lib389.plugins import AttributeUniquenessPlugin
+from lib389.idm.user import UserAccounts
+from lib389.idm.group import Groups
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.topologies import topology_st
+
+pytestmark = pytest.mark.tier1
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


droideck:
Also,  these test suites which uses the Attribute Uniqueness plugin also PASS:

- /suites/betxns/* 
- /suites/plugins/acceptance_test.py 
- /suites/monitor/db_locks_monitor_test.py

tbordaz:
Nice catch. the fix LGTM, ACK

tbordaz (approved):
ACK

PR #4876, opened 2021-08-18
Issue 4872 - BUG - entryuuid enabled by default causes replication issues

### LINKED ISSUE #4872: entryuuid needs to be "off" by default
**Issue Description**

The entryuuid plugin and it's Rust implementation creates and uses a new attribute syntax OID and new schema.  But if you are using replication with a server that does not have the entryuuid plugin then replication breaks because that older replica can not learn the new schema because the its not aware of the syntax OID (from cn=entryuuid_syntax,cn=plugins,cn=config) used in the new schema.

So we need to disable the entryuuid plugin by default, and document that it only works with replication if all the servers in the topology have the new syntax plugin/entryuuid plugin.

Firstyear:
Bug Description: Due to older servers missing the syntax
plugin this breaks schema replication and causes cascading
errors.

Fix Description: This changes the syntax to be a case
insensitive string, while leaving the plugins in place
for other usage.

fixes: https://github.com/389ds/389-ds-base/issues/4872

Review by: ???

### DIFF
--- a/ldap/schema/03entryuuid.ldif
+++ b/ldap/schema/03entryuuid.ldif
@@ -13,4 +13,5 @@ dn: cn=schema
 #
 # attributes
 #
-attributeTypes: ( 1.3.6.1.1.16.4 NAME 'entryUUID' DESC 'UUID of the entry' EQUALITY UUIDMatch ORDERING UUIDOrderingMatch SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
+# attributeTypes: ( 1.3.6.1.1.16.4 NAME 'entryUUID' DESC 'UUID of the entry' EQUALITY UUIDMatch ORDERING UUIDOrderingMatch SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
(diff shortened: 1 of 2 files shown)


Firstyear:
So the reason this works is the schema enforces "SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation" which prevents user modification of this value. However, there is a potential flaw here that a user can *import* invalid values.

I think the solution here could be in fixup to validate the entryuuid syntaxes and regenerate them if they are invalid. That's the simplest solution, and covers the most bases I think with minimal invasive changes. Thoughts?

progier389:
Not sure we should really bother to validate that values are correct according to the entryuuid syntaxes
   in import files: we usually to no do this kind of check considering that if user modify an import file he is responsible of the data consistency (and in the other cases the values should match the syntax)
   ==> So IMHO You proposal is OK as it is.

progier389 (approved):
LGTM

mreynolds389:
Ack from me

Firstyear:
more that when we run the fixup task, should we look for invalid entryuuid's in that fixup process? Without that, there is no way the admin could actually fix a botched entryuuid without re-importing the entry. 

Anyway, if you are still happy feel free to squash+merge and reword the commit.

mreynolds389:
> more that when we run the fixup task, should we look for invalid entryuuid's in that fixup process? Without that, there is no way the admin could actually fix a botched entryuuid without re-importing the entry.
> 

We could look into this extended fixup in a new ticket.  Lets just get the fix in so I can do builds tomorrow.

PR #4881, opened 2021-08-24
Issue 4880: Revert removed_config_49298_test.py wrongly modified by issue 4699

progier389:
The error that I saw in the PR test was not normal and impacted the gating test this night
git diff showed that test was somehow unwillingly modified.
 I just reverted it out:
  git checkout HEAD~1 removed_config_49298_test.py
  git commit -m "Issue 4880: Revert removed_config_49298_test.py wrongly modified by issue 4699" 
  
 Note: that is a bit worrisome meaning that there was probably a merge issue I will check if there is some other 
  CI tests unexpectedly modified but if we have the same kind of issue on the slapd code it will be hell to find it (due to the amount of changes)


progier389:
The changes in the others CI tests are OK ( add new specifics tests - disable some bdb dependants tests when using mdb - make some test able to support both bdb and mdb )
So I will wait to see if the test result are ok then merge.

Issue #4875, opened 2021-08-16
dscreate should be more verbose


mreynolds389:
**Issue Description**

**dscreate** can sometimes stall, or hang.  It would be nice to know that it's making progress, and at what stage it is in.  Right now it more or less just says "starting" and "finished".   Adding just a few of the installer steps to the output would be nice.

PR #4883, opened 2021-08-25
Issue 4875 - CLI - Add some verbosity to installer

mreynolds389:
Description:  

Previously the installer would basically say "Starting" and "Finished".  If a step would run into a problem it is difficult to narrow down what is going wrong.  So add a little more output during the installation.

relates: https://github.com/389ds/389-ds-base/issues/4875

### DIFF
--- a/src/lib389/lib389/instance/setup.py
+++ b/src/lib389/lib389/instance/setup.py
@@ -542,7 +542,7 @@ class SetupDs(object):
         return True
 
     def _prepare_ds(self, general, slapd, backends):
-
+        self.log.info("Validate installation settings ...")
         assert_c(general['defaults'] is not None, "Configuration defaults in section [general] not found")
         self.log.debug("PASSED: using config settings %s" % general['defaults'])
         # Validate our arguments.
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/instance/setup.py
```diff
@@ -948,6 +951,7 @@ def _install_ds(self, general, slapd, backends):
         # Create the backends as listed
         # Load example data if needed.
         for backend in backends:
+            self.log.info("Create database backends ...")
```
Firstyear: Should be one line up (not in the loop)

Firstyear (approved):
Besides that nit, looks good.

src/lib389/lib389/instance/setup.py
mreynolds389: hehe, whoops, I'll get that fixed.

droideck (approved):
Looks good to me too!

src/lib389/lib389/instance/setup.py
mreynolds389: I decided to keep it in the loop, but display which backend is being created

Issue #4884, opened 2021-08-25
LDAP server crashes when dnaInterval attribute is set to 0


mreynolds389:
**Issue Description**

LDAP server crashes when dnaInterval attribute is set to 0

**Steps to Reproduce**

1.Add DNA plugin config entry

dsconf -D "cn=Directory Manager" -w password ldap://localhost plugin dna config "dna config" add --type uidNumber --filter "(objectclass=top)" --scope ou=People,dc=example,dc=com --next-value 10 --interval 10 --max-value 1000 --magic-regen -1

2. Enable DNA plugin and restart the server

dsconf -D "cn=Directory Manager" -w password ldap://localhost plugin dna enable

dsctl <server> restart

3. Set the dnaInterval to 0

ldapmodify -x -p 389 -h `hostname` -D "cn=Directory Manager" -w password << EOF
dn: cn=dna config,cn=Distributed Numeric Assignment Plugin,cn=plugins,cn=config
changetype: modify
replace: dnaInterval
dnaInterval: 0
EOF

modifying entry "cn=dna config,cn=Distributed Numeric Assignment Plugin,cn=plugins,cn=config"
ldap_result: Can't contact LDAP server (-1)

PR #4885, opened 2021-08-25
Issue 4884 - server crashes when dnaInterval attribute is set to zero

mreynolds389:
Bug Description:

A division by zero crash occurs if the dnaInterval is set to zero

Fix Description:

Validate the config value of dnaInterval and adjust it to the
default/safe value of "1" if needed.

CI test already exists in QE 

relates: https://github.com/389ds/389-ds-base/issues/4884

Reviewed by: ?


tbordaz (approved):
LGTM

PR #4891, opened 2021-08-30
Issue 4890 - Need cli to easely get simple performance statistics

### LINKED ISSUE #4880: CI Test suites/config/removed_config - test_removed_config fails in gating
**Issue Description**
Seen in gating tests on master and also in PR tests:

suites/config/removed_config_49298_test.py::test_removed_config FAILED   [ 81%]
suites/filter/basic_filter_test.py::test_search_attr ERROR               [ 82%]

>       topology = create_topology({ReplicaRole.STANDALONE: 1})
...
>           mkdir(name, mode)
E           FileExistsError: [Errno 17] File exists: '/etc/dirsrv/slapd-standalone1/schema'

Clearly a instance cleanup issue of previous failing test: test_removed_config
which failed because:

        #with pytest.raises(subprocess.CalledProcessError):
        with pytest.raises(ValueError):
>           topo.standalone.start()

>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['systemctl', 'start', 'dirsrv@standalone1']' returned non-zero exit status 1.

Looks like a test case issue (wrong exception checked) because an exit status of 1 seems ok
 (i.e: ns-slapd exit(1) opposed to return code > 128 if ns-slapd was killed by a signal)

*** Note I have seen the same test_removed_config failure in issue 4699 tests. 
I should double check if the test case has not been wrongly modified/merged by latest commit
and either roll back or fix the exception handling.
and see if the error case cleanup may be improved to avoid impacting the other tests.

*** Note gating shows other failures in acl tests (around IP addresses) that I did not see in the PR associated with issue 4699

progier389:
Description:  Add a dsrate CLI and a perftools.py module to 
      - generate a test instance and run a test scenario ( results are stored in a log file )
       - list test scenarios
       - parse the log file to display a summary in csv file format so that bdb and mdb result could easily be compared in a spreadsheet.  

==================

```
dsrate --help
usage: dsrate [-h] [--resultDir RESULTDIR] {csv,list,run} ...

positional arguments:
  {csv,list,run}        sub-command help
    csv                 Generate performance result summary in csv format.
    list                List available tests.
    run                 Generate a local test instance if it does not already exist and run tests.

optional arguments:
  -h, --help            show this help message and exit
  --resultDir RESULTDIR, -r RESULTDIR
                        results directory (default is /home/progier/sb/i4890/tst/ci-
                        install/var/log/dirsrv/perfdir

```
=================

```
dsrate run -h 
usage: dsrate run [-h] [--db {bdb,mdb}] [--users USERS] [--threads THREADS]
                  {search_uid,search_uid_in_dn,modify_sn,export/import,all}
                  [{search_uid,search_uid_in_dn,modify_sn,export/import,all} ...]

Generate a local test instance if it does not already exist and run tests.

positional arguments:
  {search_uid,search_uid_in_dn,modify_sn,export/import,all}
                        test(s) to run

optional arguments:
  -h, --help            show this help message and exit
  --db {bdb,mdb}, -d {bdb,mdb}
                        db library (default is mdb)
  --users USERS, -u USERS
                        number of users in test instance
  --threads THREADS, -t THREADS
                        number of threads in client tester

```
=================

```
dsrate csv -h 
usage: dsrate csv [-h] [--out OUT]

Generate performance result summary in csv format.

optional arguments:
  -h, --help         show this help message and exit
  --out OUT, -o OUT  csv file (default is stdout)
```
=================

```
dsrate list
search_uid:         Measure number of searches per seconds
                    using filter with random existing uid.

search_uid_in_dn:   Measure number of searches per seconds
                    using filter with random existing uid in
                    dn (i.e: (uid:dn:uid_value)).

modify_sn:          Measure number of modify per seconds
                    replacing sn by random value on random
                    entries.

export/import:      Measure export rate in entries per
                    seconds then measure import rate.

all:                Run all tests for 100,1K,10K,100K,1M
                    users with both db lib with 1, 4 and 8
                    threads.
```
==================

Fixes: #4880

Reviewed by:  ???

### DIFF
--- a/src/lib389/cli/dsrate
+++ b/src/lib389/cli/dsrate
@@ -1,0 +1,149 @@
+#!/usr/bin/python3
+
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+# PYTHON_ARGCOMPLETE_OK
+
+import argcomplete
+import argparse
+import pathlib
+import signal
+import sys
+import os
+from lib389.perftools import *
+
+parser_description="""
+Handle basic performance test
+"""
+
+parser_epilog="""
+Handle basic performance test.
+"""
+
+list_description="""
+List available tests.
+"""
+
+csv_description="""
+Generate performance result summary in csv format.
+"""
+
+run_description="""
+Generate a local test instance if it does not already exist and run tests.
+"""
+
+runall_description="""
+Run all tests for 100,1K,10K,100K,1M users with both db lib with 1, 4 and 8 threads.
+"""
+
+tests = PerformanceTools.listTests()
+
+def convArg(options, args, optkey, argkey):
+    # Convert one ParseArgs arg in PerformaceTools option
+    try:
+        options[optkey] = getattr(args,argkey)
+    except AttributeError:
+        pass
+
+def convArgs(args):
+    # Convert ParseArgs args in PerformaceTools options
+    options = {}
+    for k in ( 'resultDir', 'seed', 'suffix'):
+        convArg(options, args, k, k)
+    convArg(options, args, "db_lib", "db")
+    convArg(options, args, "nbUsers", "users")
+    convArg(options, args, "nb_threads", "threads")
+    return options
+
+
+def csvSubCmd(args):
+    perftools=PerformanceTools(convArgs(args))
+    PerformanceTools.log2Csv(perftools.getFilePath("log"), args.out)
+
+def justifyText(left, right, margin):
+    pos = 0;
+    l = len(right)
+    while (pos < l):
+        spacepos = -1
+        pos0 = pos
+
+        while (pos < l and right[pos] != '\n' and pos-pos0 <= margin):
+            if (right[pos] == ' '):
+                spacepos=pos
+            pos += 1
+        if (pos==l):
+            print (f"{left:<20}{right[pos0:pos]}")
+            break;
+        if pos > margin and spacepos > 0:
+            pos = spacepos
+        print (f"{left:<20}{right[pos0:pos]}")
+        spacepos = -1
+        pos += 1
+        left=""
+        
+
+def listSubCmd(args):
+    for name,test in tests.items():
+        justifyText(name+':', test.description(), 40)
+        print ("")
+
+def runSubCmd(args):
+    options = convArgs(args)
+    perftools = PerformanceTools.Tester.initTester(options)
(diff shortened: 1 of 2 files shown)


src/lib389/cli/dsrate
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/python3
+
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 William Brown <william@blackhats.net.au>
```
progier389: Wrong Copyright date and author

src/lib389/lib389/perftools.py
```diff
@@ -0,0 +1,571 @@
...
+        def run(self, perftools, args=None):
+            for users in ( 100, 1000, 10000, 100000, 1000000 ):
+                for db in ( 'bdb', 'mdb' ):
+                    perftools = PerformanceTools.Tester.initTester({'nbUsers': users, 'db_lib': db})
+                    for t in self._testers:
+                        if 'nb_threads' in t.argsused():
+                            for nbthreads in ( 1, 4, 8 ):
+                                t.run(perftools, { "nb_threads" : nbthreads })
+                        else:
+                            t.run(perftools)
```
progier389: Implementing the "dsrate run all" feature as a Tester make things a bit cumbersome (run  parameters are ignored and instance is generated twice for first test)  should may have implement a runall sub command backed on a runAllTests static method

tbordaz (approved):
Great tool. few remarks, adding a message in the help that you need to run the command as root. Also when running it displays 'Running ldclt ...', you may add a message saying how long it will run.

ACK

progier389:
> Great tool. few remarks, adding a message in the help that you need to run the command as root. Also when running it displays 'Running ldclt ...', you may add a message saying how long it will run.
> 
> ACK

Should find the right message because telling 'root' may be misleading as
 it depends how is configured $PREFIX/share/dirsrv/inf/defaults.inf and the user permissions
( FYI: In my environment I run the performance tools as my regular user ... )
I may add something like: 
"Note that this command creates new 389ds instances so it must be run as an user that is allowed to do so."

About ldclt, I can log "Running ldclt with a timeout of xxx seconds" because that the trusty information that I have
 ( the estimation of the expected time which is 30 seconds smaller is less reliable ...)

progier389:
The CI errors are interesting (as the PR just add 2 new files (one command and one lib389 new module) and should not impact the CI result) 
We find the usual random results: systemctl start failure - LDAP_SERVER_DOWN when deleting instance - 
and one a bit weird LDAP_SERVER_DOWN during BIND while the servers are seen UP in the exception catching fixture.
But anyway none of these errors can be related to this PR so I merge it.

Issue #4869, opened 2021-08-11
Retro Changelog is not trimming changes


mreynolds389:
**Issue Description**

Retro changelog is not trimming changes under some circumstances because we are mixing monotonic and utc times

[11/Aug/2021:11:01:49.792450548 +0200] - DEBUG - DSRetroclPlugin - cltrim: ldrc=0, first_time=1622032141, cur_time=3702600
[11/Aug/2021:11:01:49.842339467 +0200] - DEBUG - DSRetroclPlugin - retrocl_housekeeping - changelog does not need to be trimmed
[11/Aug/2021:11:06:49.953401407 +0200] - DEBUG - DSRetroclPlugin - cltrim: ldrc=0, first_time=1622032141, cur_time=3702900
[11/Aug/2021:11:06:50.012249703 +0200] - DEBUG - DSRetroclPlugin - retrocl_housekeeping - changelog does not need to be trimmed

Look at cur_time. That is the "uptime" of the system, not UTC.

[root@linge ~]# uptime
 11:21:42 up 42 days, 20:49,  2 users,  load average: 0.15, 0.14, 0.16 

In the retro changelog code (trim_changelog) we are using the same monotonic time for the interval and checking for the age of an entry.  We need to use the current time when checking the age and not the monotonic time.

keestux:
Maybe we need to look at the tests too. Is there a test that should have caught this? Whatever the answer (yes or no), we need to investigate.

Is it `nsslapd-changelog-trim-interval` or `nsslapd-changelogtrim-interval`?
```
$ git grep changelog-trim
dirsrvtests/tests/suites/password/regression_test.py:                                         (ldap.MOD_REPLACE, 'nsslapd-changelog-trim-interval', b"5s"),
dirsrvtests/tests/tickets/ticket47871_test.py:           (ldap.MOD_REPLACE, 'nsslapd-changelog-trim-interval', b"5s")]
ldap/servers/plugins/retrocl/retrocl.h:#define CONFIG_CHANGELOG_TRIM_INTERVAL "nsslapd-changelog-trim-interval"
```

mreynolds389:
@keestux - agreed our tests are not robust enough to pick up on this failure

vashirov:
> Is it `nsslapd-changelog-trim-interval` or `nsslapd-changelogtrim-interval`?
> 
`nsslapd-changelog-trim-interval` is for retro changelog and `nsslapd-changelogtrim-interval` is for changelog5: 
```
[...]

PR #4895, opened 2021-09-02
Issue 4869 - Fix retro cl trimming misuse of monotonic/realtime clocks

mreynolds389:
Bug Description:  

Monotonic clocks were used to check if an entry was old enough to be trimmed, but the real system time should be used.  So entries were never trimmed from the changelog.

Fix Description:  

Make sure monotonic clocks are only used for the trimming interval, and real time clocks are used for entry age.

relates: https://github.com/389ds/389-ds-base/issues/4869


Firstyear:
Ack from me :)

PR #4906, opened 2021-09-07
Issue 4460 - python3-lib389 ignore the configuration parameters from …

aadhikar:
…/etc/openldap/ldap.conf

Description: python3-lib389 v1.4.3 seems to ignore the configuration parameters from /etc/openldap/ldap.conf.

Relates: https://github.com/389ds/389-ds-base/issues/4460

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsconf_test.py
+++ b/dirsrvtests/tests/suites/clu/dsconf_test.py
@@ -1,0 +1,87 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+"""Test dsconf CLI with LDAPS"""
+
+import logging
+import pytest
+import subprocess
+import os
+from lib389._constants import DEFAULT_SUFFIX, DN_DM
+from lib389.topologies import topology_st
+
+pytestmark = pytest.mark.tier1
+
+log = logging.getLogger(__name__)
+
+@pytest.fixture(scope="function")
+def enable_config(request, topology_st, config_type):
+    if config_type == 'ldapfile':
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+    else:
+        data = ['[localhost]\n', 'tls_cacertdir = /etc/dirsrv/slapd-standalone1\n',
+                f'uri = {topology_st.standalone.get_ldaps_uri()}\n']
+        fd = open('/root/.dsrc', 'w')
+        for line in data:
+            fd.write(line)
+        fd.close()
+
+    def fin():
```
aadhikar, line 37: Not sure if this is the right way of handling this, need suggestions, Thanks!

Firstyear: yeah, this isn't a good way to do it. Remember, the test could be run as someone who is not-root (ie dirsrv user), and each time this test is run it's going to keep adding this. Imagine the test fails and fin isn't run, then it will remain for example? 

A better way is that the libldap tools read from the environment with each variable in ldap.conf prefixed with LDAP. 

For example  `LDAPTLS_CACERT=/path/to/cert dsconf ....`

that should allow an alternate way to set this value that is non destructive and should have the same effect. 

Hope that helps!

aadhikar: @Firstyear Hey thanks for your review, this bug is related to ```ldap.conf``` specifically so I think it's better to use ```LDAPCONF``` in this way we can be sure that settings from ```ldap.conf``` are respected correctly, let me know your thoughts on this. :)

sgouvern (approved):
Ack

sgouvern (approved):
Ack

sgouvern (commented):
Ack

sgouvern (commented):
Ack

PR #4907, opened 2021-09-07
Issue 4894 - IPA failure in ipa user-del --preserve

### LINKED ISSUE #4763: Attribute Uniqueness Plugin uses wrong subtree on ModRDN

droideck:
Bug Description: Starting with 389-ds 2.0.8 on rawhide,
any call to ipa user-del --preserve fails with
This entry already exists.

Fix Description: We should split 'dn' parameter in searchAllSubtrees
into parent and target. As one of them is used for excluding the
subtree checks and another one for searching.

Fixes: https://github.com/389ds/389-ds-base/issues/4763

Reviewed by: ?

### DIFF
--- a/ldap/servers/plugins/uiduniq/uid.c
+++ b/ldap/servers/plugins/uiduniq/uid.c
@@ -770,13 +770,13 @@ search_one_berval(Slapi_DN *baseDN, const char **attrNames, const struct berval 
  *
  * Return:
  *   LDAP_SUCCESS - no matches, or the attribute matches the
- *     destination (target) dn.
+ *     source (target) dn.
  *   LDAP_CONSTRAINT_VIOLATION - an entry was found that already
  *     contains the attribute value.
  *   LDAP_OPERATIONS_ERROR - a server failure.
  */
 static int
-searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char **attrNames, Slapi_Attr *attr, struct berval **values, const char *requiredObjectClass, Slapi_DN *sourceSDN, Slapi_DN *destinationSDN, PRBool unique_in_all_subtrees)
+searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char **attrNames, Slapi_Attr *attr, struct berval **values, const char *requiredObjectClass, Slapi_DN *destinationSDN, Slapi_DN *sourceSDN, PRBool unique_in_all_subtrees)
 {
     int result = LDAP_SUCCESS;
     int i;
@@ -788,12 +788,12 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
            * are unique in all the monitored subtrees
            */
 
-        /* First check the source entry is in one of
+        /* First check the destination entry is in one of
            * the monitored subtree, so adding 'values' would
            * violate constraint
            */
         for (i = 0; subtrees && subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(sourceSDN, subtrees[i])) {
+            if (slapi_sdn_issuffix(destinationSDN, subtrees[i])) {
                 in_a_subtree = PR_TRUE;
                 break;
             }
@@ -808,7 +808,7 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
     if (exclude_subtrees != NULL) {
         PRBool in_a_subtree = PR_FALSE;
         for (i = 0; exclude_subtrees && exclude_subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(sourceSDN, exclude_subtrees[i])) {
+            if (slapi_sdn_issuffix(destinationSDN, exclude_subtrees[i])) {
                 in_a_subtree = PR_TRUE;
                 break;
             }
@@ -828,8 +828,8 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
      * The DN should already be normalized, so we don't have to
      * worry about that here.
      */
-        if (unique_in_all_subtrees || slapi_sdn_issuffix(sourceSDN, sufdn)) {
-            result = search(sufdn, attrNames, attr, values, requiredObjectClass, destinationSDN, exclude_subtrees);
+        if (unique_in_all_subtrees || slapi_sdn_issuffix(destinationSDN, sufdn)) {
+            result = search(sufdn, attrNames, attr, values, requiredObjectClass, sourceSDN, exclude_subtrees);
             if (result)
                 break;
         }
@@ -903,20 +903,20 @@ getArguments(Slapi_PBlock *pb, char **attrName, char **markerObjectClass, char *
  *
  * Return:
  *   LDAP_SUCCESS - no matches, or the attribute matches the
- *     destination (target) dn.
+ *     source (target) dn.
  *   LDAP_CONSTRAINT_VIOLATION - an entry was found that already
  *     contains the attribute value.
  *   LDAP_OPERATIONS_ERROR - a server failure.
  */
 static int
-findSubtreeAndSearch(Slapi_DN *sourceSDN, const char **attrNames, Slapi_Attr *attr, struct berval **values, const char *requiredObjectClass, Slapi_DN *destinationSDN, const char *markerObjectClass, Slapi_DN **excludes)
+findSubtreeAndSearch(Slapi_DN *destinationSDN, const char **attrNames, Slapi_Attr *attr, struct berval **values, const char *requiredObjectClass, Slapi_DN *sourceSDN, const char *markerObjectClass, Slapi_DN **excludes)
 {
     int result = LDAP_SUCCESS;
     Slapi_PBlock *spb = NULL;
     Slapi_DN *curpar = slapi_sdn_new();
     Slapi_DN *newpar = NULL;
 
-    slapi_sdn_get_parent(sourceSDN, curpar);
+    slapi_sdn_get_parent(destinationSDN, curpar);
     while (slapi_sdn_get_dn(curpar) != NULL) {
         if ((spb = dnHasObjectClass(curpar, markerObjectClass))) {
             freePblock(spb);
@@ -925,7 +925,7 @@ findSubtreeAndSearch(Slapi_DN *sourceSDN, const char **attrNames, Slapi_Attr *at
            * to have the attribute already.
            */
             result = search(curpar, attrNames, attr, values, requiredObjectClass,
-                            destinationSDN, excludes);
+                            sourceSDN, excludes);
             break;
         }
         newpar = slapi_sdn_new();
@@ -1272,7 +1272,7 @@ preop_modrdn(Slapi_PBlock *pb)
     char *markerObjectClass = NULL;
     char *requiredObjectClass = NULL;
     Slapi_DN *sourceSDN = NULL;
-    Slapi_DN *superiorSDN;
+    Slapi_DN *destinationSDN;
     char *rdn;
     int deloldrdn = 0;
     int isupdatedn;
@@ -1318,7 +1318,7 @@ preop_modrdn(Slapi_PBlock *pb)
     }
 
     /* Get superior value - unimplemented in 3.0/4.0/5.0 DS */
-    err = slapi_pblock_get(pb, SLAPI_MODRDN_NEWSUPERIOR_SDN, &superiorSDN);
+    err = slapi_pblock_get(pb, SLAPI_MODRDN_NEWSUPERIOR_SDN, &destinationSDN);
     if (err) {
         result = uid_op_error(32);
         break;
@@ -1329,8 +1329,8 @@ preop_modrdn(Slapi_PBlock *pb)
      * its current level in the tree.  Use the source SDN for
      * determining which managed tree this belongs to
      */
-    if (!superiorSDN)
-        superiorSDN = sourceSDN;
+    if (!destinationSDN)
+        destinationSDN = sourceSDN;
 
     /* Get the new RDN - this has the attribute values */
     err = slapi_pblock_get(pb, SLAPI_MODRDN_NEWRDN, &rdn);
@@ -1364,7 +1364,7 @@ preop_modrdn(Slapi_PBlock *pb)
 
     /* Apply the rename operation to the dummy entry. */
     /* slapi_entry_rename does not expect rdn normalized */
-    err = slapi_entry_rename(e, rdn, deloldrdn, superiorSDN);
+    err = slapi_entry_rename(e, rdn, deloldrdn, destinationSDN);
     if (err != LDAP_SUCCESS) {
         result = uid_op_error(36);
         break;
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1219,7 +1219,7 @@ preop_modify(Slapi_PBlock *pb)
         } else {
             /* Subtrees listed on invocation line */
             result = searchAllSubtrees(config->subtrees, config->exclude_subtrees, attrNames, NULL,
-                                       mod->mod_bvalues, requiredObjectClass, sdn, config->unique_in_all_subtrees);
+                                       mod->mod_bvalues, requiredObjectClass, sdn, sdn, config->unique_in_all_subtrees);
```
Firstyear: It may be worth renaming sdn here to clarify what's going on to prevent future confusiong :)

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -793,7 +793,7 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
            * violate constraint
            */
         for (i = 0; subtrees && subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(dn, subtrees[i])) {
+            if (slapi_sdn_issuffix(parentDN, subtrees[i])) {
```
tbordaz: Is this change required for the fix ?
If I am not wrong I think that in the case that target DN (destination entry) is one of the subtrees, then testing the parent will result in_a_subtree=False, although it should be True.

It is a corner case but may be valuable if the moved entry is the head of a complete subtree.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -808,7 +808,7 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
     if (exclude_subtrees != NULL) {
         PRBool in_a_subtree = PR_FALSE;
         for (i = 0; exclude_subtrees && exclude_subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(dn, exclude_subtrees[i])) {
+            if (slapi_sdn_issuffix(parentDN, exclude_subtrees[i])) {
```
tbordaz: Same as above.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -828,8 +828,8 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
      * The DN should already be normalized, so we don't have to
      * worry about that here.
      */
-        if (unique_in_all_subtrees || slapi_sdn_issuffix(dn, sufdn)) {
-            result = search(sufdn, attrNames, attr, values, requiredObjectClass, dn, exclude_subtrees);
+        if (unique_in_all_subtrees || slapi_sdn_issuffix(parentDN, sufdn)) {
```
tbordaz: same as above.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1392,13 +1392,13 @@ preop_modrdn(Slapi_PBlock *pb)
              */
             if (NULL != markerObjectClass) {
                 /* Subtree defined by location of marker object class */
-                result = findSubtreeAndSearch(slapi_entry_get_sdn(e), attrNames, attr, NULL,
-                                              requiredObjectClass, superior,
+                result = findSubtreeAndSearch(superior, attrNames, attr, NULL,
```
tbordaz: Regarding this regression, is it the path that was failing ?

ldap/servers/plugins/uiduniq/uid.c
droideck: Okay, I think I should rename it all as it brings a lot of confusion. :)

It's actually sourceDN and destinationDN. So for modrdn we should use destinationDN for any plugin checks as it's where we should apply the policies.
And sourceDN should be used for ignoring in the 'search()' function.

ldap/servers/plugins/uiduniq/uid.c
droideck: Sure, thanks!

ldap/servers/plugins/uiduniq/uid.c
droideck: Nope, this path was okay but I've decided to clear the things up a bit (so it uses the same logic as searchAllSubtrees, and we have one call less in the execution)

Firstyear:
Looks a lot better mate :) ack from me, I'll leave @tbordaz to give his comments too.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -770,13 +770,13 @@ search_one_berval(Slapi_DN *baseDN, const char **attrNames, const struct berval
  *
  * Return:
  *   LDAP_SUCCESS - no matches, or the attribute matches the
- *     target dn.
+ *     destination (target) dn.
```
tbordaz: Sorry but I  am confused. With a MODRDN we have a  "sourceDN -> destinationDN". In my mind "target" is meaning SLAPI_<op>_TARGET_SDN, that is "sourceSDN". IIUC target is souce not destination.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -788,12 +788,12 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
            * are unique in all the monitored subtrees
            */
 
-        /* First check the target entry is in one of
+        /* First check the source entry is in one of
```
tbordaz: So it would be 'destination'

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -788,12 +788,12 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
...
            * the monitored subtree, so adding 'values' would
            * violate constraint
            */
         for (i = 0; subtrees && subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(dn, subtrees[i])) {
+            if (slapi_sdn_issuffix(sourceSDN, subtrees[i])) {
```
tbordaz: Here we should rather check if the destinationSDN lands into one of the monitored subtrees.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -808,7 +808,7 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
     if (exclude_subtrees != NULL) {
         PRBool in_a_subtree = PR_FALSE;
         for (i = 0; exclude_subtrees && exclude_subtrees[i]; i++) {
-            if (slapi_sdn_issuffix(dn, exclude_subtrees[i])) {
+            if (slapi_sdn_issuffix(sourceSDN, exclude_subtrees[i])) {
```
tbordaz: Here we should check if the destinationSDN lands into a not monitored subtree

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -820,16 +820,16 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
...
    */
     for (i = 0; subtrees && subtrees[i]; i++) {
         Slapi_DN *sufdn = subtrees[i];
         /*
      * The DN should already be normalized, so we don't have to
      * worry about that here.
      */
-        if (unique_in_all_subtrees || slapi_sdn_issuffix(dn, sufdn)) {
-            result = search(sufdn, attrNames, attr, values, requiredObjectClass, dn, exclude_subtrees);
+        if (unique_in_all_subtrees || slapi_sdn_issuffix(sourceSDN, sufdn)) {
```
tbordaz: No, we should check that subtree if the destinationSDN lands into that monitored sufdn

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -820,16 +820,16 @@ searchAllSubtrees(Slapi_DN **subtrees, Slapi_DN **exclude_subtrees, const char *
 
     /*
    * For each DN in the managed list, do uniqueness checking if
-   * the target DN is a subnode in the tree.
+   * the destination (target) DN is a subnode in the tree.
```
tbordaz, line 823: I agree it is destination but the word 'target' is IMHO confusing.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -903,20 +903,20 @@ getArguments(Slapi_PBlock *pb, char **attrName, char **markerObjectClass, char *
  *
  * Return:
  *   LDAP_SUCCESS - no matches, or the attribute matches the
- *     target dn.
+ *     destination (target) dn.
```
tbordaz: idem

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -925,7 +925,7 @@ findSubtreeAndSearch(Slapi_DN *parentDN, const char **attrNames, Slapi_Attr *att
            * to have the attribute already.
            */
             result = search(curpar, attrNames, attr, values, requiredObjectClass,
-                            target, excludes);
```
tbordaz, line 928: This is something that I do not understand. uiduniqness should check if the uid value already exist in the parent of the destinationSDN, not in the parent of the sourceSDN.
IMHO it should be slapi_sdn_get_parent(destinationSDN, curpar);

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1271,8 +1271,8 @@ preop_modrdn(Slapi_PBlock *pb)
     int err;
     char *markerObjectClass = NULL;
     char *requiredObjectClass = NULL;
-    Slapi_DN *sdn = NULL;
-    Slapi_DN *superior;
+    Slapi_DN *sourceSDN = NULL;
+    Slapi_DN *superiorSDN;
```
tbordaz: I prefer destinationSDN rather than superiorSDN

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1311,26 +1311,26 @@ preop_modrdn(Slapi_PBlock *pb)
...
     /*
      * No superior means the entry is just renamed at
-     * its current level in the tree.  Use the target DN for
+     * its current level in the tree.  Use the source SDN for
      * determining which managed tree this belongs to
      */
-    if (!superior)
-        superior = sdn;
+    if (!superiorSDN)
+        superiorSDN = sourceSDN;
```
tbordaz: superiorSDN being SLAPI_MODRDN_NEWSUPERIOR_SDN, if it is NULL shouldn't it be initialized with slapi_sdn_get_parent(sourceSDN, superiorSDN) ?

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1392,13 +1392,13 @@ preop_modrdn(Slapi_PBlock *pb)
              */
             if (NULL != markerObjectClass) {
                 /* Subtree defined by location of marker object class */
-                result = findSubtreeAndSearch(slapi_entry_get_sdn(e), attrNames, attr, NULL,
-                                              requiredObjectClass, superior,
+                result = findSubtreeAndSearch(superiorSDN, attrNames, attr, NULL,
+                                              requiredObjectClass, sourceSDN,
```
tbordaz, line 1396: Instead of sourceSDN, shouldn't it be <rdn>,<superiorSDN> ?

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1392,13 +1392,13 @@ preop_modrdn(Slapi_PBlock *pb)
...
                                               markerObjectClass, config->exclude_subtrees);
             } else {
                 /* Subtrees listed on invocation line */
                 result = searchAllSubtrees(config->subtrees, config->exclude_subtrees, attrNames, attr, NULL,
-                                           requiredObjectClass, superior, config->unique_in_all_subtrees);
+                                           requiredObjectClass, superiorSDN, sourceSDN, config->unique_in_all_subtrees);
```
tbordaz: Interface is searchAllSubtrees(.., source, destination,..), shouldn't it be searchAllSubtrees(..., sourceSDN, "<rdn>,<superiorSDN>",... )

tbordaz (commented):
@droideck thanks for the changes it is much clearer now.
Sorry to be pernickety but it remains things not clear to me and even wondering if there were not others hidden old bugs in that code

ldap/servers/plugins/uiduniq/uid.c
droideck: searchAllSubtrees is also used for modify and add, and it uses targetSDN parameter. So I left it here for people who will read this function after preop_modify or preop_add, so they can easily understand (and not being confused by "destination")

ldap/servers/plugins/uiduniq/uid.c
droideck: Later in the code, we use superiorSDN and construct the entry out of there. And slapi_sdn_get_parent is also used later.
We'll need to refactor a lot if we change it here...

ldap/servers/plugins/uiduniq/uid.c
droideck: Why? We need sourceSDN so we can ignore it in the results.

ldap/servers/plugins/uiduniq/uid.c
droideck: Same as above. We need sourceSDN so we can ignore it in the results.

ldap/servers/plugins/uiduniq/uid.c
tbordaz: Where later ?
Do you mean when it is used in slapi_entry_rename(e, rdn, deloldrdn, destinationSDN) ?

ldap/servers/plugins/uiduniq/uid.c
tbordaz: Your change looks valid to me, using destinationSDN and sourceSDN

ldap/servers/plugins/uiduniq/uid.c
tbordaz: Your change looks valid to me, using destinationSDN and sourceSDN

ldap/servers/plugins/uiduniq/uid.c
tbordaz: For me it is the last thing I do not understand. Unless I am mistaken, the forth param of slapi_entry_rename should be the parent of the destination entry and not the destination entry itself.

ldap/servers/plugins/uiduniq/uid.c
```diff
@@ -1364,7 +1364,7 @@ preop_modrdn(Slapi_PBlock *pb)
 
     /* Apply the rename operation to the dummy entry. */
     /* slapi_entry_rename does not expect rdn normalized */
-    err = slapi_entry_rename(e, rdn, deloldrdn, superior);
+    err = slapi_entry_rename(e, rdn, deloldrdn, destinationSDN);
```
tbordaz, line 1367: This is the last thing that I do not understand. 
Initially it was called with 'superior'. It made sens as the fourth param is expected to be the parent of the destination entry.
But now, if SLAPI_MODRDN_NEWSUPERIOR_SDN is NULL (case an entry is just renamed not moved), then destinationSDN takes the value of sourceSDN. It should take the value of the parent of sourceDN.

tbordaz (commented):
Thank you sooo much for all your changes. It makes the code much more simple to read IMHO.

droideck:
Okay, I see your point. And I think you found one more bug in the previously existent code.
Now it's fixed! Please, review :)

And thank you so much for the thorough review as usual!

tbordaz (approved):
Thanks @droideck for all that cleanup in addition to the orginal fix. The code is good to me. ACK

Issue #4912, opened 2021-09-08
dsidm command crashing when account policy plugin is enabled


mreynolds389:
**Issue Description**
 If the account policy plugin is enbaled but there is no config entry defined then dsism crashes like this:

    dsidm -v -b dc=example,dc=com localhost account entry-status uid=demo_user,ou=people,dc=example,dc=com

DEBUG: _gen_dn filter = (|(objectclass=nsAccount)(objectclass=nsPerson)(objectclass=simpleSecurityObject)(objectclass=organization)(objectclass=person)(objectclass=account)(objectclass=organizationalUnit)(objectclass=netscapeServer)(objectclass=domain)(objectclass=posixAccount)(objectclass=shadowAccount)(objectclass=posixGroup)(objectclass=mailRecipient))
DEBUG: _gen_dn dn = uid=demo_user,ou=people,dc=example,dc=com
DEBUG: cn=Account Policy Plugin,cn=plugins,cn=config getVal('nsslapd-pluginEnabled')
DEBUG: cn=Account Policy Plugin,cn=plugins,cn=config getVal('nsslapd-pluginarg0')
DEBUG: None get_attrs_vals_utf8(['stateattrname', 'altstateattrname', 'specattrname', 'limitattrname'])
DEBUG: search_ext() argument 1 must be str, not None
Traceback (most recent call last):
  File "/usr/sbin/dsidm", line 130, in <module>
    result = args.func(inst, dsrc_inst['basedn'], log, args)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/cli_idm/account.py", line 70, in entry_status
    status = acct.status()
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/idm/account.py", line 105, in status
    config_settings = config.get_attrs_vals_utf8(["stateattrname", "altstateattrname",
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/_mapped_object.py", line 639, in get_attrs_vals_utf8
    entry = self._instance.search_ext_s(self._dn, ldap.SCOPE_BASE, self._object_filter, attrlist=keys,
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/__init__.py", line 176, in inner
    return f(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 863, in search_ext_s
[...]

PR #4913, opened 2021-09-08
Issue 4912 - dsidm command crashing when account policy plugin is enabled

mreynolds389:
Bug Description:  

If the account policy plugin is enabled, but not configured then dsidm will crash when checking an entry's status.

Fix Description: 

Check if the config DN is present before trying to check its values.

relates: https://github.com/389ds/389-ds-base/issues/4912


Firstyear (approved):
Ack

Issue #4796, opened 2021-06-09
dsconf for nsslapd-state


falon:
**Is your feature request related to a problem? Please describe.**
With the old console on 389-ds 1.3.x I can manage the `nsslapd-state` for all suffixes.
In the new cockpit console I can't find an option to manage this attribute. It's always `backend`.
`dsconf ` doesn't manage this attribute too.

**Describe the solution you'd like**
At least, a new option in `dsconf `could be provided. For instance:

`dsconf instance backend suffix set --state`

Allowed values:

- `backend`: The back-end database processes all operations.
- `disabled`: The database is not available for processing operations. The server returns a No
such search object error in response to requests made by client applications.
- `referral`: Directory Server returns a referral URL for requests to this suffix.
- `referral on update`: The database is used for all operations. Only for update requests is a
referral sent.

**Describe alternatives you've considered**
Now I have to force an `ldapmodify `to set the value:

```
dn: cn=<suffix name>,cn=mapping tree,cn=config
changeype: modify
replace: nsslapd-state
nsslapd-state: referral on update
```

mreynolds389:
This can definitely be added to the CLI and UI.  We are currently working on porting the UI to Patternfly 4 so this is a good time to add this change...

progier389:
just a point about that one: by default  dsconf should not allow to change the state of a replicated backend. 
  (because replication sometimes overrides the state (and the referrals) and it will be confusing for user.
If we really want to offer the referral on update feature we also need to improve the ns-sate handling within the server 
  (for example by having 3 attributes:  ns-state (which is only readable and computed from nsstate;replica and ns-state;config)
   to insure there is no risk to revert in backend mode unexpectedly

mreynolds389:
4634ec6ab..2e4387db6  389-ds-base-2.0 -> 389-ds-base-2.0

This will eventually be backported to 389-ds-base-1.4.4 as part of a UI/lib389 rebase.

PR #4919, opened 2021-09-10
Issue 4796 - Add support for nsslapd-state to CLI & UI

mreynolds389:
Description:  

Add support for nsslapd-state to lib389 and UI.  Also added a check to prevent the changing of nsslapd-state for replicated suffixes.

Also did a little UI cleanup where a bottom margin was added  to the bottom of pages instead of using "hr" to create the gap.

relates: https://github.com/389ds/389-ds-base/issues/4796

### DIFF
--- a/src/cockpit/389-console/src/lib/database/suffix.jsx
+++ b/src/cockpit/389-console/src/lib/database/suffix.jsx
@@ -54,11 +54,13 @@ export class Suffix extends React.Component {
             dncachememsize: this.props.data.dncachememsize,
             readOnly: this.props.data.readOnly,
             requireIndex: this.props.data.requireIndex,
+            dbstate: this.props.data.dbstate,
             _cachememsize: this.props.data.cachememsize,
             _cachesize: this.props.data.cachesize,
             _dncachememsize: this.props.data.dncachememsize,
             _readOnly: this.props.data.readOnly,
             _requireIndex: this.props.data.requireIndex,
+            _dbstate: this.props.data.dbstate,
             savingConfig: false,
             saveBtnDisabled: true,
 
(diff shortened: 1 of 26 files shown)


mreynolds389:
@jchapma - nice catch!!  Fixed, please review...

src/cockpit/389-console/src/lib/database/suffix.jsx
```diff
@@ -773,6 +775,10 @@ export class Suffix extends React.Component {
             cmd.push("--dncache-memsize=" + this.state.dncachememsize);
             requireRestart = true;
         }
+        if (this.state._dbstate != this.state.dbstate) {
```
droideck, line 778: Currently, if the user changes 'Backend State' to 'Refferal' but no referrals exist, the user gets 'Unwilling to perform' message in the notification.

It is a bit cryptic.
Only in the Console Log we can see:

```
cockpit.js:1 {
    "msgtype": 103,
    "msgid": 9,
    "result": 53,
    "desc": "Server is unwilling to perform",
    "ctrls": [],
    "info": "need to set nsslapd-referral before moving to referral state\n"
}
```

I wonder if it's worth adding additional validation here (or print the meaningful message in the notification).

droideck (approved):
It's really minor (as we have a similar situation across other UI modules).
So you have my 'ack' as everything else is good!

src/cockpit/389-console/src/lib/database/suffix.jsx
mreynolds389: We can add a check for "info" in the message and print it..  I've done this in other parts of the code, but it was too big of change to do it across the board with my previous work.  Maybe we should open a new ticket for adding the "info" check when doing error notifications

mreynolds389: Ok, I fixed it, and I will open a ticket to make this approach consistent across the UI

PR #4924, opened 2021-09-15
Issue 4923 - issue about LMDB dbi versus txn handling

### LINKED ISSUE #4923: issue about LMDB dbi versus txn handling.
**Issue Description**
Some CI tests fails on lmdb and a build with DEBUG trace at db-mdb plugin level shows that some 
  operation fails with EINVAL error because the provided dbi is not valid within the provided txn. 
The issue (related to current dbi management design) is that:
  An add operation is performed 
    so a backend r/w transaction is open 
    an nsuniqueid is generated and to check its unicity a lookup is performed in nsuniqueid index
      to to that the nsuniqueid  dbi is open in a sub txn.
    sub txn is commited and database and db-mdb plugin context are updated.
    now the add operation fails  (which is expected in the test case scenario)
    so upper level backend txn is aborted 
       and this roll back the dbi creation within the db (but not within db-mdb plugin context)
    and we ends up with an inconsistent context that leads to EINVAL errors when rollbacked dbi are used. 

Solution:
   it is quite difficult to resynchronize the db-mdb plugin context (because the only trigger at plugin level 
       is that a r/w txn is aborted and for performance reason we do not really want to check all the dbis validity every time a txn is aborted.
   another approach is to pre-open all the dbis:
      - when starting an instance
      - when importing or reindexing  
   So the context get updated only once the txn is successfully committed.
   And never change until db env get closed or db instance get removed
   Furthermore in mdb case there is no resource usage issue with dbi: 
       ( Within lmdb, having an open dbi only consume a small amount of memory (no file descriptor are needed unlike in bdb))
   db instance get still open in the fly when adding a backend or an index but these config operation have no backend level txn
    that may be aborted after having successfully open the dbi.

 - Platform: Fedora
 - Package and version: current master

**Steps to Reproduce**
1. Run the auto_member or the clu pytests
2. See failure 
3. Check for EINVAL error in error log

progier389:


ldap/servers/slapd/back-ldbm/db-mdb/mdb_debug.c
```diff
@@ -494,9 +511,10 @@ dbg_import_elmt(const char *file, int lineno, const char *funcname, const char *
     dbgval2str(keystr, sizeof keystr, &key);
     dbgval2str(datastr, sizeof datastr, &data);
 
-    dbi_str = elmt2->slot->dbi.dbname;
+    if (elmt2->slot->dbi)
```
Firstyear: nit, add ` { } `

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -944,6 +953,8 @@ dbmdb_ancestorid_new_idl_create_index(backend *be, ImportJob *job, dbi_txn_t * t
     }
 
 out:
+    if (txn)
```
Firstyear: ` { } `

Firstyear (changes requested):
> another approach is to pre-open all the dbis:

I think this is actually required. Without this you could (in theory) have an index dbi change while you hold id2entry which would make them inconsistent relative to each other inside the write op. Similar with reads, you need to take every dbi simultaneously to ensure that they are all point-in-time consistent for that read operation covering the entire set of operations that is about to be performed.

progier389:
> > another approach is to pre-open all the dbis:
> 
> I think this is actually required. Without this you could (in theory) have an index dbi change while you hold id2entry which would make them inconsistent relative to each other inside the write op. Similar with reads, you need to take every dbi simultaneously to ensure that they are all point-in-time consistent for that read operation covering the entire set of operations that is about to be performed.

Thak you for your feedback William,
Although I do not think it is really an argument to preopen the dbis  because 
   [1] When using bdb the instances are open when needed
   [2] Transient inconsistency while removing an index does not matter ( as the index db will be soon destroyed anyway )
   [3] Adding an index on non empty database is inconsistent by nature ==> you must reindex after adding some new indexes  
         so inconsitency before the reindex does not really matter. 
the scenario of a config change while operation is in progress is interesting:
   I should double check but I suspect that current commit code may lead to an abort in that scenario:
       if attrinfo list is modified within config code,  ( Just need to double check whether attrinfo get modified after or before dbi get created  in config code) 
        a write operation still try to create the dbi if it is not open and this will abort in txn code because thread already have a txn
      (Anyway abort is not really needed, it make things easier to the early debug phase, but I could simply return an error
        instead )

ldap/servers/slapd/back-ldbm/db-mdb/mdb_debug.c
progier389: Ok, FYI: I fixed the 3 "if"

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
progier389: Fixed.

Firstyear:
> > > another approach is to pre-open all the dbis:
> > 
> > 
> > I think this is actually required. Without this you could (in theory) have an index dbi change while you hold id2entry which would make them inconsistent relative to each other inside the write op. Similar with reads, you need to take every dbi simultaneously to ensure that they are all point-in-time consistent for that read operation covering the entire set of operations that is about to be performed.
> 
> Thak you for your feedback William,
> Although I do not think it is really an argument to preopen the dbis because
> [1] When using bdb the instances are open when needed

I think we probably have subtle inconsistencies that can occur there as a result of the lack of proper transactional behaviours, but just that it may be too small a window or transient enough people don't notice.

> [2] Transient inconsistency while removing an index does not matter ( as the index db will be soon destroyed anyway )

That's fine. 

> [3] Adding an index on non empty database is inconsistent by nature ==> you must reindex after adding some new indexes

That's also fair. 

> so inconsitency before the reindex does not really matter.

Well it kinda does, because if the reindex fails, by wrapping it fully in transactions you can roll back to the previous state as a whole, rather that leaving the DB in a state with no indexes at all. 

> the scenario of a config change while operation is in progress is interesting:
> I should double check but I suspect that current commit code may lead to an abort in that scenario:
> if attrinfo list is modified within config code, ( Just need to double check whether attrinfo get modified after or before dbi get created in config code)
> a write operation still try to create the dbi if it is not open and this will abort in txn code because thread already have a txn
> (Anyway abort is not really needed, it make things easier to the early debug phase, but I could simply return an error
> instead )

I think there are probably lots of cases like this re-config change because 389 doesn't have a good history of transactional behaviours so I wouldn't be surprised to find more issues like that present in the code :)

progier389:
> Well it kinda does, because if the reindex fails, by wrapping it fully in transactions you can roll back to the previous state as a whole, rather that leaving the DB in a state with no indexes at all.
>
I disagree:
  if we reindex that usually means that previous index is inconsistent. so IMHO it is better to have no index (which is clearly visible) rather than inconsistent one (which is quite harder to diagnose)

 And anyway in practice there is not a single txn hat you can rollback:
  Keeping a single txn for the very long duration of a reindex task would have several disadvantages:
    - resource issue due to the number of operation associated with the txn.
    - lmdb programming guide warns against keeping write txn for a long time:
    because it blocks other write txn (in our case, operation on other backends)
  So while reindexing or importing a flow of  txn begin / txn commit is performed for each set of data to write 
  then to avoid using a partially build index the dbi is flagged as DIRTY when starting reindex/import task
    this flag prevents to open the dbi except for another reindex/import and is reset when import/reindex successfully complete
==> in error case we still ends up with no index. (FYI: a similar mechanism is used in bdb that is only the way to determine a dbi is dirty that is different)

> I think there are probably lots of cases like this re-config change because 389 doesn't have a good history of transactional behaviours so I wouldn't be surprised to find more issues like that present in the code :)

progier389:
Discovered an issue related to clu/dbgen_test.pl failure:
  the issue is that the serach returns 2000 entries on bdb while only 500 on mdb 
  So I compared the databases.
    Interresting point is that when I dump them with dbscan -D ?db -A -f ... most of the result are exactly identical.
    some of the differences are normal:
        empty database versus non existing database (as dbi are pre created they are empty when there is no entry containing 
          the attribute)
       difference in id2entry related to numsubordonates (likely due to broken ancestorid)
    some are not normal (in fact 1 index seems broken:) 
       **some ids are missing in ancestorid** and I suspect that is the root cause of some of my problems
    I  must now understand  why ancestorid get corrupted.

progier389:
I am afraid  I understand the issue: 
  I think that the hard lmdb limit of a page for duplicate values on DUPSORT dbi  has been reached after around 500 IDs
  and if is is the case that means that we should change the way ID lists are handled in the dbis
  back to the old index scheme ? (To bad, I did not implemented it) 
  use a new pattern (compressed bitmaps ?)
  implement a kind of overflow page for keys ?
  Use openldap  idlist ?

progier389:
Performed a basic test on lmdb and was able to create 5M values on a single key ==> so my suspicion i wrong: lmdb does support it as far as the key lenght + value lenght for a single key/value is smaller than 500 bytes 
==> lokking further to understand why the keys are not created (or not retrieved)

progier389:
The problem is related to the idl management and the bulk read operator 
I decided to avoid using that operation (which does not exists natively in lmdb) by implementing a dbmdb_idl_new_fetch directly with lmdb primitive (with the double advantage of having less overhead and working properly)
This fixed the ancestorid issue and  clu/dbgen_test.py tests are now passing.

progier389:
usn test failure was because a missing cmp function (entryusb index use string integer value (so key must use integer order rather than lexicographical order)

progier389:
Only 2 failing tests for both mdb and bdb and it seems harmless as they are tests that we see randomly fails like the import time )
So it is finally time to merge that one

tbordaz:
@progier389  Change of interface of slapi_log_error/slapi_log_error_ext (char* -> const char*) is an issue for existing plugins. Would it be possible to revert that part of the PR ?

Issue #4925, opened 2021-09-17
Performance ACI: targetfilter evaluation result can be reused


tbordaz:
**Issue Description**
An ACI may contain _targetfilter_. For a given operation (like search) a same targetfilter is evaluated many times. This evaluation is expensive (syntax/matching rule specific), especially in case of substring matching.

Some evaluation result can be cached and reused

**Package Version and Platform:**
all version

**Steps to Reproduce**
Provision a subtree of large number of entries with large number of attributes and a bind account that is allowed to read those entries/attributes.
Add ACIs with large dummy targetfilter granting access to entries. For example '...(&(description=*lrd)(description=*orld)...(description=*ello world))...'

a ldapsearch with the bind account and retrieve all the subtree entries with all attributes. The search should is slow.

**Expected results**
Search should not be extremely long

tbordaz:
22156fcdd..96dbbb214  master
ce00d6fb3..aaeff2d1c  389-ds-base-1.4.4
e6a06390c..53582d12f  389-ds-base-1.4.3

mreynolds389:
@tbordaz please cherry pick this to 389-ds-base-2.0 as well :-)

mreynolds389:
> @tbordaz please cherry pick this to 389-ds-base-2.0 as well :-)

done:

54d7cc78c..c661024b3  389-ds-base-2.0 -> 389-ds-base-2.0

tbordaz:
backport to 1.3.10
a2c73838d..d09cd92cc  389-ds-base-1.3.10

PR #4926, opened 2021-09-20
Issue 4925 - Performance ACI: targetfilter evaluation result can be r…

tbordaz:
…eused

Bug description:
	An ACI may contain targetfilter. For a given returned entry, of a
        SRCH request, the same targetfilter is evaluated for each of the
        returned attributes.
        Once the filter has been evaluated, it is useless to reevaluate
        it for a next attribute.

Fix description:
	The fix implements a very simple cache (linked list) that keeps
        the results of the previously evaluated 'targetfilter'.
        This cache is per-entry. For an operation, a aclpb is allocated
        that is used to evaluate ACIs against each successive entry.
        Each time a candidate entry is added in the aclpb
        (acl_access_allowed), the cache (aclpb_curr_entry_targetfilters)
        is freed. Then for each 'targetfilter', the original targetfilter
        is lookup from the cache. If this is the first evaluation of it
        then the result of the evaluation is stored into the cache using
        the original targetfilter as the key in the cache

relates: https://github.com/389ds/389-ds-base/issues/4925

Reviewed by:

Platforms tested: F34

### DIFF
--- a/ldap/servers/plugins/acl/acl.c
+++ b/ldap/servers/plugins/acl/acl.c
@@ -67,6 +67,9 @@ static void print_access_control_summary(char *source,
                                          const char *edn,
                                          aclResultReason_t *acl_reason);
 static int check_rdn_access(Slapi_PBlock *pb, Slapi_Entry *e, const char *newrdn, int access);
+static struct targetfilter_cached_result *targetfilter_cache_lookup(struct acl_pblock *aclpb, char *filter, PRBool filter_valid);
+static void targetfilter_cache_add(struct acl_pblock *aclpb, char *filter, int result, PRBool filter_valid);
+
 
 
 /*
@@ -176,6 +179,70 @@ check_rdn_access(Slapi_PBlock *pb, Slapi_Entry *e, const char *dn, int access)
     return (retCode);
 }
 
+/* Retrieves, in the targetfilter cache (list), this
+ * filter in case it was already evaluated
+ *
+ * filter: key to retrieve the evaluation in the cache
+ * filter_valid: PR_FALSE means that the filter key is truncated, PR_TRUE else
+ */
+static struct targetfilter_cached_result *
+targetfilter_cache_lookup(struct acl_pblock *aclpb, char *filter, PRBool filter_valid)
+{
+    struct targetfilter_cached_result *results;
+    if (! aclpb->targetfilter_cache_enabled) {
+        /* targetfilter cache is disabled */
+        return NULL;
+    }
+    if (filter == NULL) {
+        return NULL;
+    }
+    for(results = aclpb->aclpb_curr_entry_targetfilters; results; results = results->next) {
+        if (strcmp(results->filter, filter) == 0) {
+            return results;
+        }
+    }
+
+    return NULL;
+}
+
+/* Free all evaluations cached for this current entry */
+void
+targetfilter_cache_free(struct acl_pblock *aclpb)
+{
+    struct targetfilter_cached_result *results, *next;
+    if (aclpb == NULL) {
+        return;
+    }
+    for(results = aclpb->aclpb_curr_entry_targetfilters; results;) {
+        next = results->next;
+        slapi_ch_free_string(&results->filter);
+        slapi_ch_free((void **) &results);
+        results = next;
+    }
+    aclpb->aclpb_curr_entry_targetfilters = NULL;
+}
+
+/* add a new targetfilter evaluation into the cache (per entry)
+ * ATM just use a linked list of evaluation
+ *
+ * filter: key to retrieve the evaluation in the cache
+ * result: result of the evaluation
+ * filter_valid: PR_FALSE means that the filter key is truncated, PR_TRUE else
+ */
+static void
+targetfilter_cache_add(struct acl_pblock *aclpb, char *filter, int result, PRBool filter_valid)
+{
+    struct targetfilter_cached_result *results;
+    if (! filter_valid || ! aclpb->targetfilter_cache_enabled) {
+        /* targetfilter cache is disabled or filter is truncated */
+        return;
+    }
+    results = (struct targetfilter_cached_result *) slapi_ch_calloc(1, (sizeof(struct targetfilter_cached_result)));
+    results->filter = slapi_ch_strdup(filter);
+    results->next = aclpb->aclpb_curr_entry_targetfilters;
+    results->matching_result = result;
+    aclpb->aclpb_curr_entry_targetfilters = results;
+}
 /***************************************************************************
 *
 * acl_access_allowed
@@ -496,6 +563,7 @@ acl_access_allowed(
 
         /* Keep the ptr to the current entry */
         aclpb->aclpb_curr_entry = (Slapi_Entry *)e;
+        targetfilter_cache_free(aclpb);
 
         /* Get the attr info */
         deallocate_attrEval = acl__get_attrEval(aclpb, attr);
@@ -1927,7 +1995,7 @@ acl_modified(Slapi_PBlock *pb, int optype, Slapi_DN *e_sdn, void *change)
 *    None.
 *
 **************************************************************************/
-static int
+int
 acl__scan_for_acis(Acl_PBlock *aclpb, int *err)
 {
     aci_t *aci;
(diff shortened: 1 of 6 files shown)


ldap/servers/plugins/acl/acl.c
```diff
@@ -176,6 +179,60 @@ check_rdn_access(Slapi_PBlock *pb, Slapi_Entry *e, const char *dn, int access)
...
+}
+
+/* Free all evaluations cached for this current entry */
+void
+targetfilter_cache_free(struct targetfilter_cached_result **per_entry_results)
+{
+    struct targetfilter_cached_result *results, *next;
+    for(results = *per_entry_results; results;) {
+        next = results->next;
+        slapi_ch_free((void **) &results->filter);
```
mreynolds389: Minor nit pick, but to be compliant with our coding guidelines the char "filter" should be freed using "slapi_ch_free_string()"

mreynolds389:
Besides my minor comment, LGTM

ldap/servers/plugins/acl/acl.c
```diff
@@ -2408,10 +2466,53 @@ acl__resource_match_aci(Acl_PBlock *aclpb, aci_t *aci, int skip_attrEval, int *a
...
-                                        aci->targetFilter,
-                                        0 /*don't do access check*/) != 0) {
-                filter_matched = ACL_FALSE;
+            Slapi_DN *sdn;
+            char* attr_evaluated = "None";
+            char logbuf[1024];
+            char *filterstr; /* key to retrieve/add targetfilter value in the cache */
+            struct targetfilter_cached_result *previous_filter_test;
+            
+            /* only usefull for debug purpose */
```
Firstyear, line 2488: If its for debug could this be in a #DEBUG define?

ldap/servers/plugins/acl/acl.c
```diff
@@ -2408,10 +2466,53 @@ acl__resource_match_aci(Acl_PBlock *aclpb, aci_t *aci, int skip_attrEval, int *a
...
+            if (aclpb->aclpb_curr_attrEval && aclpb->aclpb_curr_attrEval->attrEval_name) {
+                attr_evaluated = aclpb->aclpb_curr_attrEval->attrEval_name;
+            }
+            sdn = slapi_entry_get_sdn(aclpb->aclpb_curr_entry);
+            
+            /* The key for the cache is the original filter */
+            filterstr = slapi_filter_to_string(aci->targetFilter, logbuf, sizeof (logbuf));
+            
+            previous_filter_test = targetfilter_cache_lookup(aclpb, filterstr);
+            if (previous_filter_test) {
```
Firstyear, line 2510: Is there some evidence of cache hit rates here during this kind of eval? It seems pretty situational maybe?

Firstyear (commented):
Would like to know more about the hitrates of this cache to show it's effectiveness :)

ldap/servers/plugins/acl/acl.c
tbordaz: @mreynolds389 Thanks for you review. I will fix that. In addition I think the config parameter 'nsslapd-targetfilter-cache' should be set once (for example when the aclpb is allocated) else even if it is atomic read it can create fake contention on it during parallel evaluation of targetfilters.

ldap/servers/plugins/acl/acl.c
tbordaz: @Firstyear many thanks also for having look at this PR.
ATM this PR is kind of preliminary and I would also appreciate any feedback on the use of a cache mechanism itself. ACI is a key security component and I am a bit concerned about caching anything there.
I think the cache is safe as it is at a per entry level. So when we evaluate a filter against an entry , the result of this evaluation is still valid for the same filter against the same entry again.

A concern is how do we know it is the same filter ? is strcmp valid for that ?

Regarding the hit rate, it depends on the ACI target definition and how much attributes are returned. In my tests, the SRCH returned 8 attributes and the ACI, containing the targetfilter, target all those attributes. So instead of evaluating 9 times the targetfilter (8 attributes + 1entry) it evaluates it only once. If the targetfilter is complex like 
`(&(objectclass=hosts)(fqdn=*.domain.us)(name=*foo)(description=*bla bla bla)....(title=*xxx))` and expensive to compute like substring matching rules, then extra evaluations have a significant impact. On a testcase retrieving 500 entries with 8 attributes and one targetfilter ACI it goes from 4500 evaluations without the cache to 500 with the cache. If there are several selected ACIs with targetfilters it makes things worse.
From a test etime dropped from 2.5s to 0.5s.

ldap/servers/plugins/acl/acl.c
tbordaz: For a given operation, if a selected ACI (with targetfilter) targets N requested attribute then the rate is N/N+1. so its minimal rate is 50% (one attribute requested).

ldap/servers/plugins/acl/acl.c
```diff
@@ -176,6 +179,63 @@ check_rdn_access(Slapi_PBlock *pb, Slapi_Entry *e, const char *dn, int access)
...
+void
+targetfilter_cache_free(struct acl_pblock *aclpb)
+{
+    struct targetfilter_cached_result *results, *next;
+    if (aclpb == NULL) {
+        return;
+    }
+    for(results = aclpb->aclpb_curr_entry_targetfilters; results;) {
+        next = results->next;
+        slapi_ch_free_string((void **) &results->filter);
```
mreynolds389: You can remove the void cast:

slapi_ch_free_string(&results->filter);

ldap/servers/plugins/acl/acl.c
```diff
@@ -2408,10 +2469,53 @@ acl__resource_match_aci(Acl_PBlock *aclpb, aci_t *aci, int skip_attrEval, int *a
                                                     ACL_EVAL_TARGET_FILTER);
             slapi_ch_free((void **)&lasinfo);
         } else {
-            if (slapi_vattr_filter_test(NULL, aclpb->aclpb_curr_entry,
-                                        aci->targetFilter,
-                                        0 /*don't do access check*/) != 0) {
-                filter_matched = ACL_FALSE;
+            Slapi_DN *sdn;
+            char* attr_evaluated = "None";
+            char logbuf[1024];
```
mreynolds389: Is this big enough?  What happens if the filter is longer than 1024 characters?

mreynolds389: I guess it's a corner case, since a target filter in an ACI would not typically be very big, but you never know.

tbordaz: I agree, I changed that so that if the filterstr gets truncated then it is ignored and we systematically reevaluate the targetfilter

ldap/servers/plugins/acl/acl.c
tbordaz: done

mreynolds389:
It might also be good to run this under under ASAN - just to be safe :-)

ldap/servers/plugins/acl/acl.c
Firstyear: "A concern is how do we know it is the same filter ? is strcmp valid for that ?" -- Yep, that's find in this case. 

And that's a good improvement there, I didn't know how it was evaluated multiple times like that :)

ldap/servers/plugins/acl/acl.c
Firstyear: Even minumum of 50% is good IMO.

ldap/servers/plugins/acl/acl.c
Firstyear: I guess the other factor there is if the filter is really large then those are going to be more expensive and that's really where we would make gains in performance?

tbordaz: I see your point: The larger is the targetfilter the more beneficial should be to cache its result.
On the other side I did not want to loop allocate buffer/convert filter/check filterstr is truncated because of performance impact. So using a static buffer was the most immediate solution as most of the targetfilters will fit into.
I see 2 options
-  loop allocate/convert/check
- use a static buffer and if the targetfilter is too large to fit into it then log an info that the cache is not used for this targetfilter.

tbordaz: Instead of looping allocating buffer/testing overwrite until the buffer is large enough, the modified fix keeps a static buffer and detect overwrite (redzone). In case of overwrite the filter is just ignore (does not go into the cache).

Also the buffer is not 2K.

tbordaz:
The ASAN tests are okay. only the usual leak (pwdhash)

mreynolds389:
> The ASAN tests are okay. only the usual leak (pwdhash)

I'm happy with this, ack again.

tbordaz:
@mreynolds389 , @Firstyear thanks for the reviews

PR #4933, opened 2021-09-22
Issue 4513 - fix CI tests and add debug logging

mreynolds389:
Description:  Harden some of the tests and add debug logging to the aci
tests that only fail in a container

Also udpated the doxy file and fixed some compiler warnings

relates: https://github.com/389ds/389-ds-base/issues/4513

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsctl_acceptance_test.py
+++ b/dirsrvtests/tests/suites/clu/dsctl_acceptance_test.py
@@ -9,7 +9,7 @@
 import logging
 import pytest
 import os
-from lib389._constants import *
+import time
 from lib389.topologies import topology_st as topo
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


droideck (commented):
Odd... GitHub CI reports 'test_connection_from_an_unauthorized_network' test failure.

But it works for me.

At the same time, if I run 'py.test -s -v dirsrvtests/tests/suites/acl/keywords_part2_test.py', I can 100% times reproduce a different failure in 'test_access_from_certain_network_only_ip':

```
        # After changing  the ip user cant access data
        with pytest.raises(ldap.INSUFFICIENT_ACCESS):
>           org.replace("seeAlso", "cn=1")
E           Failed: DID NOT RAISE <class 'ldap.INSUFFICIENT_ACCESS'>
```

mreynolds389:
@droideck actually I was able to fix all the ACI tests with ip/hostname rules.  So this is no longer a debug patch  Please review

dirsrvtests/tests/suites/clu/dsctl_acceptance_test.py
```diff
@@ -39,6 +39,7 @@ def test_custom_path(topo):
 
     # Set backup directory to LDIF directory
     topo.standalone.config.replace('nsslapd-bakdir', ldif_dir)
+    time.sleep(.5)
```
tbordaz, line 45: With that change the test is now pass. Do not you want to also remove the flaky mark ?

dirsrvtests/tests/suites/paged_results/paged_results_test.py
```diff
@@ -510,11 +508,10 @@ def test_search_with_timelimit(topology_st, create_user):
     finally:
         del_users(users_list)
 
-#unstable or unstatus tests, skipped for now
-@pytest.mark.flaky(max_runs=2, min_passes=1)
+
 @pytest.mark.parametrize('aci_subject',
                          ('dns = "{}"'.format(HOSTNAME),
-                          'ip = "{}"'.format(IP_ADDRESS)),
+                          'ip = "{}" or ip = "::1"'.format(IP_ADDRESS)),
```
tbordaz: This test is still failing for me. The test with IP is pass. The test with 'dns'  with localhost or with 'localhost or localhost.localdomain' is failing. When evaluating the ACI the server selects the aci but evaluation of the dns bind rule does not match. I do not know the reason. I guess it is reproducible and attaching to LASDnsEval would give us the culprit.
I agree with the changes in this PR, but on my laptop evaluation of dns constantly fails

mreynolds389: For me the paged results test (ip and dns) now passes on my laptop, but "dns" still fails in a container :-(   I'm working on it now (that's why there are all these commits to this PR)

tbordaz (approved):
All the tests are successful on my laptop. Thanks.
The CI tests of mapping_tree is raising a scary SERVER_DOWN but this is after 66min, so I think it can be a framework issue.
On my laptop I was not able to reproduce the server_down after several tries.
ACK

PR #4937, opened 2021-09-30
Issue 4925 - Performance ACI: targetfilter evaluation result can be r…

tbordaz:
…eused (#4926)

Bug description:
	An ACI may contain targetfilter. For a given returned entry, of a
        SRCH request, the same targetfilter is evaluated for each of the
        returned attributes.
        Once the filter has been evaluated, it is useless to reevaluate
        it for a next attribute.

Fix description:
	The fix implements a very simple cache (linked list) that keeps
        the results of the previously evaluated 'targetfilter'.
        This cache is per-entry. For an operation, a aclpb is allocated
        that is used to evaluate ACIs against each successive entry.
        Each time a candidate entry is added in the aclpb
        (acl_access_allowed), the cache (aclpb_curr_entry_targetfilters)
        is freed. Then for each 'targetfilter', the original targetfilter
        is lookup from the cache. If this is the first evaluation of it
        then the result of the evaluation is stored into the cache using
        the original targetfilter as the key in the cache

	The key to lookup/store the cache is the string representation
        of the targetfilter. The string contains a redzone to detect
        that the filter exceeds the maximum size (2K). If it exceeds
        then the key is invalid and the lookup/store is noop.

relates: #4925

Reviewed by: Mark Reynolds, William Brown (Thanks)

Platforms tested: F34

### DIFF
--- a/ldap/servers/plugins/acl/acl.c
+++ b/ldap/servers/plugins/acl/acl.c
@@ -66,6 +66,9 @@ static void print_access_control_summary(char *source,
                                          const char *edn,
                                          aclResultReason_t *acl_reason);
 static int check_rdn_access(Slapi_PBlock *pb, Slapi_Entry *e, const char *newrdn, int access);
+static struct targetfilter_cached_result *targetfilter_cache_lookup(struct acl_pblock *aclpb, char *filter, PRBool filter_valid);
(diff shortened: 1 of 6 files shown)


tbordaz:
Merge directly

PR #4940, opened 2021-10-06
Issue 4312 - performance search rate: contention on global monitoring…

tbordaz:
… counters

Bug description:
	The servers manages a set of counters in order to report metrics either with SRCH
	"cn=monitor" or with SNMP agent.
	The counters are global and so the threads updating the counters are all accessing
	the same counters and same memory addresses. The counters are accessed by workers and/or listener threads.
	All of them are in competition to access the counter memory addresses that creates contention.

Fix description:
	The fix spread the set of global counters into a per thread set
	of global counters
	https://www.port389.org/docs/389ds/design/global-counters-contention.html

relates: https://github.com/389ds/389-ds-base/issues/4312

Reviewed by:

Platforms tested:  F34

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -28,7 +28,7 @@
 #endif
 
 typedef Connection work_q_item;
-static void connection_threadmain(void);
+static void connection_threadmain(void *arg);
 static void connection_add_operation(Connection *conn, Operation *op);
 static void connection_free_private_buffer(Connection *conn);
 static void op_copy_identity(Connection *conn, Operation *op);
@@ -277,8 +277,8 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
     conn->c_connid = slapi_counter_increment(num_conns);
 
     if (!in_referral_mode) {
-        slapi_counter_increment(g_get_global_snmp_vars()->ops_tbl.dsConnectionSeq);
-        slapi_counter_increment(g_get_global_snmp_vars()->ops_tbl.dsConnections);
+        slapi_counter_increment(g_get_per_thread_snmp_vars()->ops_tbl.dsConnectionSeq);
+        slapi_counter_increment(g_get_per_thread_snmp_vars()->ops_tbl.dsConnections);
     }
 
     /*
@@ -429,6 +429,7 @@ init_op_threads()
     pthread_condattr_t condAttr;
     int32_t max_threads = config_get_threadnumber();
     int32_t rc;
+    int32_t *threads_indexes;
 
     /* Initialize the locks and cv */
     if ((rc = pthread_mutex_init(&work_q_lock, NULL)) != 0) {
@@ -457,12 +458,20 @@ init_op_threads()
 
     work_q_stack = PR_CreateStack("connection_work_q");
     op_stack = PR_CreateStack("connection_operation");
+    alloc_per_thread_snmp_vars(max_threads);
+    init_thread_private_snmp_vars();
+    
 
+    threads_indexes = (int32_t *) slapi_ch_calloc(max_threads, sizeof(int32_t));
+    for (size_t i = 0; i < max_threads; i++) {
+        threads_indexes[i] = i + 1; /* idx 0 is reserved for global snmp_vars */
+    }
+    
     /* start the operation threads */
     for (size_t i = 0; i < max_threads; i++) {
         PR_SetConcurrency(4);
         if (PR_CreateThread(PR_USER_THREAD,
-                            (VFP)(void *)connection_threadmain, NULL,
+                            (VFP)(void *)connection_threadmain, (void *) &threads_indexes[i],
                             PR_PRIORITY_NORMAL, PR_GLOBAL_THREAD,
                             PR_UNJOINABLE_THREAD,
                             SLAPD_DEFAULT_THREAD_STACKSIZE) == NULL) {
@@ -474,6 +483,10 @@ init_op_threads()
             g_incr_active_threadcnt();
         }
     }
+    /* Here we should free thread_indexes, but because of the dynamic of the new
+     * threads (connection_threadmain) we are not sure when it can be freed.
+     * Let's accept that unique initialization leak (typically 32 to 64 bytes)
+     */
 }
 
 static void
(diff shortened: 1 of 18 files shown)


ldap/servers/slapd/connection.c
```diff
@@ -457,12 +458,20 @@ init_op_threads()
 
     work_q_stack = PR_CreateStack("connection_work_q");
     op_stack = PR_CreateStack("connection_operation");
+    alloc_per_thread_snmp_vars(max_threads);
+    init_thread_private_snmp_vars();
+    
 
+    threads_indexes = (int32_t *) slapi_ch_calloc(max_threads, sizeof(int32_t));
```
Firstyear, line 465: Off by one error here, you account for a global index but not the max threads. Needs to be max_threads+1

ldap/servers/slapd/connection.c
```diff
@@ -457,12 +458,20 @@ init_op_threads()
...
+    for (size_t i = 0; i < max_threads; i++) {
+        threads_indexes[i] = i + 1; /* idx 0 is reserved for global snmp_vars */
+    }
+    
     /* start the operation threads */
     for (size_t i = 0; i < max_threads; i++) {
         PR_SetConcurrency(4);
         if (PR_CreateThread(PR_USER_THREAD,
-                            (VFP)(void *)connection_threadmain, NULL,
+                            (VFP)(void *)connection_threadmain, (void *) &threads_indexes[i],
```
Firstyear, line 474: Similar, off by one, you aren't reserving 0 for global since this iters from 0. You may want to make max_threads + 1 the global count instead.

Firstyear (changes requested):

ldap/servers/slapd/connection.c
tbordaz: Thanks @Firstyear for your review.
'threads_indexes' contains as slot as #workers. The purpose of 'threads_indexes' is to provide to each worker a unique and specific index (its own slot) in the array of global counters (per_thread_snmp_vars[]).

If there are 5 workers, threads_indexes contains 5 slots with threads_indexes[0]=1,.. threads_indexes[4]=5.

Then the first worker will use per_thread_snmp_vars[1], ... the fifth will use per_thread_snmp_vars[5].
The slot per_thread_snmp_vars[0] is a fallback one, when the calling thread has not register a value for 'thread_private_snmp_vars_idx'

ldap/servers/slapd/connection.c
tbordaz: connection_threadmain (workers) are using indexes 1..<#workers> in per_thread_snmp_vars

mreynolds389:
So are all these stats just used for SNMP?  I wonder if we should have a config option to enable these counters.  I don't think many people still use SNMP, and maybe we could disable it by default.  Would that provide more of a performance boost?   Just curious...

tbordaz:
> So are all these stats just used for SNMP? I wonder if we should have a config option to enable these counters. I don't think many people still use SNMP, and maybe we could disable it by default. Would that provide more of a performance boost? Just curious...

Good point. Indeed there are two mechanisms to retrieve stats: snmp_agt+shared_memory, search on 'cn=snmp,cn=monitor'.
The values on snmp_agt+shared_memory are periodically (each 10s) updated according to the global counters.
The values on 'cn=snmp,cn=monitor' are update on demand (SRCH).
So snmp_agt+shared_memory is costly even if there is no snmp_agt. By default it is enabled ('nsslapd-counters') and it could be good idea to disable it, as snmp agt is not frequently used. Also because of its locking mechanism I wonder if it can contribute to periodic slowdown. I will dig into this but I am not sure it is a good idea to disable it by default with this same ticket.

ldap/servers/slapd/daemon.c
```diff
@@ -971,6 +971,10 @@ slapd_daemon(daemon_ports_t *ports)
     }
 
     init_op_threads();
+            /* Start the SNMP collator if counters are enabled. */
+        if (config_get_slapi_counters()) {
+            snmp_collator_start();
+        }
```
mreynolds389, line 979: indentation here

mreynolds389:
> > So are all these stats just used for SNMP? I wonder if we should have a config option to enable these counters. I don't think many people still use SNMP, and maybe we could disable it by default. Would that provide more of a performance boost? Just curious...
> 
> Good point. Indeed there are two mechanisms to retrieve stats: snmp_agt+shared_memory, search on 'cn=snmp,cn=monitor'. The values on snmp_agt+shared_memory are periodically (each 10s) updated according to the global counters. The values on 'cn=snmp,cn=monitor' are update on demand (SRCH). So snmp_agt+shared_memory is costly even if there is no snmp_agt. By default it is enabled ('nsslapd-counters') and it could be good idea to disable it, as snmp agt is not frequently used. Also because of its locking mechanism I wonder if it can contribute to periodic slowdown. I will dig into this but I am not sure it is a good idea to disable it by default with this same ticket.

I agree, if we do disable it then we open a new ticket.  Thanks

Firstyear:
> So are all these stats just used for SNMP? I wonder if we should have a config option to enable these counters. I don't think many people still use SNMP, and maybe we could disable it by default. Would that provide more of a performance boost? Just curious...

This actually seems like a good idea yes.

ldap/servers/slapd/connection.c
tbordaz: @Firstyear, are my explanations good enough ?

Firstyear: Sorry about the delay @tbordaz! I see where my confusion was now, this is all good from me :)

ldap/servers/slapd/connection.c
Firstyear: Also same confusion as above, all good.

Firstyear (approved):
All good from me, I understand my mistake now, I think mark still has a comment about indentation though. :)

ldap/servers/slapd/daemon.c
tbordaz: This is fixed in the current patch

tbordaz:
> > > So are all these stats just used for SNMP? I wonder if we should have a config option to enable these counters. I don't think many people still use SNMP, and maybe we could disable it by default. Would that provide more of a performance boost? Just curious...
> > 
> > 
> > Good point. Indeed there are two mechanisms to retrieve stats: snmp_agt+shared_memory, search on 'cn=snmp,cn=monitor'. The values on snmp_agt+shared_memory are periodically (each 10s) updated according to the global counters. The values on 'cn=snmp,cn=monitor' are update on demand (SRCH). So snmp_agt+shared_memory is costly even if there is no snmp_agt. By default it is enabled ('nsslapd-counters') and it could be good idea to disable it, as snmp agt is not frequently used. Also because of its locking mechanism I wonder if it can contribute to periodic slowdown. I will dig into this but I am not sure it is a good idea to disable it by default with this same ticket.
> 
> I agree, if we do disable it then we open a new ticket. Thanks
> 
> The RFE ticket #4999 will disable SNMP monitoring

@mreynolds389 is this PR okay for you ?

mreynolds389:
LGTM

PR #4941, opened 2021-10-07
Issue 4938 - max_failure_count can be reached in dscontainer on slow …

### LINKED ISSUE #4938: max_failure_count can be reached in dscontainer on slow machine with missing debug exception trace
Hello,
On a very slow machine, the ldapi instance is very long to be available and the timeout can be reached in dscontainer.
The number of attempts is hardcoded in max_failure_count.
Is it possible to allow to modify the max_failure_count value, please?
The diagnostic is complicated by the fact that when max_failure_count is reached, the exception is hidden in logs:
 try:
    inst = _gen_instance()
    inst.open()
    if "dn: cn=Directory Manager" == inst.whoami_s():
        return (False, True)
    else:
        log.error("The instance may be misconfigured, unable to cn=Directory Manager autobind.")
        return (False, False)
**except:
    log.debug("Instance LDAPI not functional (yet?)")
    pass**

Perhaps when timeout is reached, this exception could be printed?
Best regards,
Loris

vashirov:
…machine with missing debug exception trace

Bug Description:
On a very slow machine max_failure_count can be reached to soon. For
troubleshooting and diagnostics this parameter should be configurable.

Fix Description:
Introduce a new env variable DS_STARTUP_TIMEOUT that accepts a number in
seconds. By default it is 60.
Log a traceback when we reach the timeout.

Fixes: https://github.com/389ds/389-ds-base/issues/4938

Reviewed by: ???

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -333,17 +333,23 @@ binddn = cn=Directory Manager
 
     # Wait on the health check to show we are ready for ldapi.
     healthy = False
-    max_failure_count = 20
+    startup_timeout = os.getenv("DS_STARTUP_TIMEOUT", 60)
+    max_failure_count = int(startup_timeout / 3)
     for i in range(0, max_failure_count):
         if ds_proc is None:
             log.warning("ns-slapd pid has disappeared ...")
             break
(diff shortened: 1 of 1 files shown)


Firstyear (approved):
Thanks! Ack from me :)

Issue #4944, opened 2021-10-08
Backport of accept thread to 389-ds-base-1.3.10 branch


jchapma:
**RFE backport**
Previously the server accepted connections and checked for new work in the same event loop. This could
cause connection table polling to delay accepts, and accepts to delay connection activity from being ready.

An accept thread was added to separate those functions allowing accept to occur in parallel.

**Performance comparison.**
To evaluate the benefit of back porting this RFE to the 1.3.10 branch, the server was loaded with 10K concurrent connections both without and with the back port. Performance metrics were obtained from logconv.pl and server access logs for each scenario and are approximate.

```
389-ds-base-1.3.10 without accept thread
LDAP Connections:           10000         (45.98/sec)  (2758.56/min)
```
```
389-ds-base-1.3.10  with accept thread
LDAP Connections:           10000         (66.74/sec)  (4004.52/min)
```
The accept thread improves the servers ability to handle multiple concurrent connections by ~45%. Another benefit is that the accept thread reduces the load on the server, in the 10K test the server maxed out at ~70%.

**Additional context**
Initial RFE - https://github.com/389ds/389-ds-base/commit/5f48459494a2ad19488c253e03e9839c1c218687

mreynolds389:
Too risky to backport, closing

PR #4945, opened 2021-10-08
Issue 4944 - Backport of accept thread to 1.3.10 branch

jchapma:
Description: Previously the server accepted connections and checked
for new work in the same event loop. This could cause connection table
polling to delay accepts, and accepts to delay connection activity from
being ready.

An accept thread was added to separate those functions allowing accept
to occur in parallel. This commit is a back port of the accept thread.

relates: https://github.com/389ds/389-ds-base/issues/4944

Reviewed by: ?


Firstyear:
Basicly looks like what I expected, thanks for this! Ack from me.

Firstyear (approved):

Issue #4943, opened 2021-10-08
Fix csn generator to limit time skew drift 


progier389:
**Issue Description**
Customers log analysis reported that the csn time skew drift to large values.
Code review showed that the way the local and remote offset are computed is a bit strange and probably wrong:

```
I notice that the csn generator algorithm has been quite modified since the iPlanet days,
and if I understand it correctly these changes are an attempt to reduce the local_offset
when the system clock moves ahead.
The idea is cool and should limit the time skew but the code seems wrong and that is probably why we still have trouble:

I have some doubt on the local offset computation (adjust_local_time):
 IMHO it should be:
       [A] local_offset = MAX_VAL(0, sample_time + local_offset - cur_time)
       because:
           new csn timestamp >= old csn timestamp ==>
           cur_time + new_local_offset + remote_offset >= sample_time + old_local_offset + remote_offset ==>
           new_local_offset >= sample_time + old_local_offset - cur_time
           and we also have new_local_offset >= 0 which leads to the [A} formula
 
and remote offset computation (csngen_adjust_time) looks clearly wrong:.
          we should have: 
              new csn timestamp >= remote csn timestamp ==>
              cur_time + local_offset + remote_offset >= remote csn timestamp
              remote_offset >= remote csn timestamp - cur_time - local_offset
              and remote_offset can only increase so:

        delta_remote_offset = remote_time - cur_time - gen->state.local_offset - gen->state.remote_offset;            
       if (delta_remote_offset > 0) {
            if (ignore_time_skew || (delta_remote_offset+gen->state.remote_offset <= CSN_MAX_TIME_ADJUST)) {
                gen->state.remote_offset += delta_remote_offset;
            } else /* remote_offset > CSN_MAX_TIME_ADJUST */
        }

Another point: 
  to avoid beat phenomena that could lead to 1s time skew drift per second
 if (delta_remote_offset > 0 ) 
  we can increase local_offset by 1 and decrease delta_remote_offset by 1

**Package Version and Platform:**
 - Platform: All
 - Package and version: [e.g. 389-ds-base-1.4.4.4-20200721git5d41dc5a4.fc32.x86_64]


**Steps to Reproduce**
Steps to reproduce the behavior:
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error

**Expected results**
A clear and concise description of what you expected to happen.

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Additional context**
Add any other context about the problem here.

[...]

PR #4946, opened 2021-10-08
Issue 4943 - Fix csn generator to limit time skew drift

progier389:
Fix some oddity in csngen code so that 
 - new csn is greater than old csn and remote csn
 - local_offset mainly increase when system clocks goes backward
    with two exceptions:
      increased 1 second if seqnum rolls over.
      set to 1 seconds if local_offset is zero  and remote offset is detected (in this case remote_offset is decreased by
         1 to compensate)
  - local_offset may decrease (until 0) when system clock increase    
  - remote offset increase when a csn generated locally would be lesser than the remote csn
  - remote offset never decrease

Test done: 
   all CI tests that pass on master are pass
   csngen_test is ok.
   csngen_multi_suppliers_test is OK (==> csn always increase) it also shows that a small noise (+- 1 sec) does not generate time skew drift but larger noise (+- 15 sec) generates drift. 
      
Issue: [4943](https://github.com/389ds/389-ds-base/issues/4943)

Reviewer: @tbordaz and @mreynolds389

### DIFF
--- a/ldap/servers/slapd/csngen.c
+++ b/ldap/servers/slapd/csngen.c
@@ -17,6 +17,8 @@
 #include <string.h>
 #include "prcountr.h"
 #include "slap.h"
+#include <math.h>
+
 
 #define CSN_MAX_SEQNUM 0xffff              /* largest sequence number */
 #define CSN_MAX_TIME_ADJUST 24 * 60 * 60   /* maximum allowed time adjustment (in seconds) = 1 day */
@@ -63,6 +65,7 @@ typedef struct csngen_state
 struct csngen
 {
     csngen_state state;      /* persistent state of the generator */
+    int32_t (*gettime)(struct timespec *tp); /* Get local time */
     callback_list callbacks; /* list of callbacks registered with the generator */
     Slapi_RWLock *lock;      /* concurrency control */
 };
@@ -78,7 +81,7 @@ static int _csngen_init_callbacks(CSNGen *gen);
 static void _csngen_call_callbacks(const CSNGen *gen, const CSN *csn, PRBool abort);
 static int _csngen_cmp_callbacks(const void *el1, const void *el2);
 static void _csngen_free_callbacks(CSNGen *gen);
-static int _csngen_adjust_local_time(CSNGen *gen, time_t cur_time);
+static int _csngen_adjust_local_time(CSNGen *gen);
 
 /*
  * **************************************************************************
@@ -121,6 +124,7 @@ csngen_new(ReplicaId rid, Slapi_Attr *state)
     _csngen_init_callbacks(gen);
 
     gen->state.rid = rid;
+    gen->gettime = slapi_clock_gettime;
 
     if (state) {
         rc = _csngen_parse_state(gen, state);
@@ -164,10 +168,7 @@ csngen_free(CSNGen **gen)
 int
 csngen_new_csn(CSNGen *gen, CSN **csn, PRBool notify)
 {
-    struct timespec now = {0};
     int rc = CSN_SUCCESS;
-    time_t cur_time;
-    int delta;
 
     if (gen == NULL || csn == NULL) {
         slapi_log_err(SLAPI_LOG_ERR, "csngen_new_csn", "Invalid argument\n");
@@ -180,39 +181,13 @@ csngen_new_csn(CSNGen *gen, CSN **csn, PRBool notify)
         return CSN_MEMORY_ERROR;
     }
 
-    if ((rc = slapi_clock_gettime(&now)) != 0) {
-        /* Failed to get system time, we must abort */
-        slapi_log_err(SLAPI_LOG_ERR, "csngen_new_csn",
-                "Failed to get system time (%s)\n",
-                slapd_system_strerror(rc));
-        return CSN_TIME_ERROR;
-    }
-    cur_time = now.tv_sec;
-
     slapi_rwlock_wrlock(gen->lock);
 
-    /* check if the time should be adjusted */
-    delta = cur_time - gen->state.sampled_time;
-    if (delta > _SEC_PER_DAY || delta < (-1 * _SEC_PER_DAY)) {
-        /* We had a jump larger than a day */
-        slapi_log_err(SLAPI_LOG_INFO, "csngen_new_csn",
-                "Detected large jump in CSN time.  Delta: %d (current time: %ld  vs  previous time: %ld)\n",
-                delta, cur_time, gen->state.sampled_time);
+    rc = _csngen_adjust_local_time(gen);
+    if (rc != CSN_SUCCESS) {
+        slapi_rwlock_unlock(gen->lock);
+        return rc;
     }
-    if (delta > 0) {
-        rc = _csngen_adjust_local_time(gen, cur_time);
-        if (rc != CSN_SUCCESS) {
-            slapi_rwlock_unlock(gen->lock);
-            return rc;
-        }
-    }
-    /* if (delta < 0) this means the local system time was set back
-     * the new csn will be generated based on sampled time, which is
-     * ahead of system time and previously generated csns.
-     * the time stamp of the csn will not change until system time
-     * catches up or is corrected by remote csns.
-     * But we need to ensure that the seq_num does not overflow.
-     */
 
     if (gen->state.seq_num == CSN_MAX_SEQNUM) {
         slapi_log_err(SLAPI_LOG_INFO, "csngen_new_csn", "Sequence rollover; "
@@ -307,8 +282,7 @@ csngen_adjust_time(CSNGen *gen, const CSN *csn)
     /* Get last local csn time */
     old_time = CSN_CALC_TSTAMP(gen);
     /* update local offset and sample_time */
-    cur_time = slapi_current_utc_time();
-    rc = _csngen_adjust_local_time(gen, cur_time);
+    rc = _csngen_adjust_local_time(gen);
 
     if (slapi_is_loglevel_set(SLAPI_LOG_REPL)) {
         cur_time = CSN_CALC_TSTAMP(gen);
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/csngen.c
```diff
@@ -313,79 +319,60 @@ csngen_adjust_time(CSNGen *gen, const CSN *csn)
...
-                              remote_offset, (long)CSN_MAX_TIME_ADJUST);
-                slapi_rwlock_unlock(gen->lock);
-                csngen_dump_state(gen, SLAPI_LOG_DEBUG);
-                return CSN_LIMIT_EXCEEDED;
-            }
-        } else if (remote_offset > 0) { /* still need to account for this */
-            gen->state.local_offset += remote_offset;
+    remote_offset = remote_time - CSN_CALC_TSTAMP(gen);
+    if (remote_offset > 0) {
+        if (!ignore_time_skew && (remote_offset > CSN_MAX_TIME_ADJUST)) {
```
progier389: should be gen->state.remote_offset+remote_offset

progier389:
Following the scrub meeting discussion, I modify master branch code to integrate the new csngen test and run it:
and I got interresting result:
   The test quickly (i.e: state 29 over 500) aborted because generated csn are not always increasing  although it should
     because in the test each created is "sent" to the other suppliers (i.e: adust_time(rid, csn) is called for all other rids) 
      before a new csn is generated.
     But test result shows clearly that rid 4 csn is smaller than previous csn generated on rid 3
     In other word. supplier 4  generated a csn that is smaller than a remote csn that  it received. 

==> Clearly a good reason to change the code    

[12/Oct/2021:17:13:44.154716384 +0200] - INFO - csngen_dump_state - CSN generator's state:
[12/Oct/2021:17:13:44.157724003 +0200] - INFO - csngen_dump_state - 	replica id: 3
[12/Oct/2021:17:13:44.160581090 +0200] - INFO - csngen_dump_state - 	sampled time: 16777266
[12/Oct/2021:17:13:44.163518593 +0200] - INFO - csngen_dump_state - 	local offset: 0
[12/Oct/2021:17:13:44.166321111 +0200] - INFO - csngen_dump_state - 	remote offset: 4
[12/Oct/2021:17:13:44.169456043 +0200] - INFO - csngen_dump_state - 	sequence number: 1
[12/Oct/2021:17:13:44.172412082 +0200] - INFO - csngen_dump_state - CSN generator's state:
[12/Oct/2021:17:13:44.175419500 +0200] - INFO - csngen_dump_state - 	replica id: 4
[12/Oct/2021:17:13:44.178284188 +0200] - INFO - csngen_dump_state - 	sampled time: 16777253
[12/Oct/2021:17:13:44.181162213 +0200] - INFO - csngen_dump_state - 	local offset: 12
[12/Oct/2021:17:13:44.184135142 +0200] - INFO - csngen_dump_state - 	remote offset: 3
[12/Oct/2021:17:13:44.187086116 +0200] - INFO - csngen_dump_state - 	sequence number: 267
[12/Oct/2021:17:13:44.190328434 +0200] - ERR - csngen_multi_suppliers_test - CSN generated in disorder state=29 rid=4

mreynolds389:
Nice improvement to this code, ACK from me.  There is no immediate rush for this, so lets have Thierry look at this too when he gets back.

ldap/servers/slapd/csngen.c
```diff
@@ -121,6 +123,7 @@ csngen_new(ReplicaId rid, Slapi_Attr *state)
     _csngen_init_callbacks(gen);
 
     gen->state.rid = rid;
+    gen->gettime = slapi_clock_gettime;
```
tbordaz: Few lines below, 'sampled_time' is set with slapi_current_utc_time.
slapi_clock_gettime and slapi_current_utc_time are both calling CLOCK_REALTIME.
It would be good if the name 'slapi_clock_gettime' would be something like 'slapi_clock_utc_gettime' so that we know call the same underlying function.

ldap/servers/slapd/csngen.c
```diff
@@ -313,79 +292,60 @@ csngen_adjust_time(CSNGen *gen, const CSN *csn)
...
-            } else {
-                gen->state.seq_num = remote_seqnum + 1;
-            }
+        gen->state.remote_offset += remote_offset;
+        /* To avoid beat phenomena between suppliers let put 1 second in local_offset
+         * it will be eaten at next clock tick rather than increasing remote offset
+         * If we do not do that we will have a time skew drift of 1 second per 2 seconds
+         * if suppliers are desynchronized by 0.5 second 
+         */
+        if (gen->state.local_offset == 0) {
```
tbordaz, line 320: It increase local_offset by 1 and decrease remote_offset by 1 only if local_offset == 0
Why not systematically ?

ldap/servers/slapd/csngen.c
```diff
@@ -597,94 +557,93 @@ _csngen_cmp_callbacks(const void *el1, const void *el2)
...
-                          "gen state back before %08lx%04x:%ld:%ld:%ld\n",
-                          new_time, gen->state.seq_num,
-                          gen->state.sampled_time,
-                          gen->state.local_offset,
-                          gen->state.remote_offset);
-        }
+    }
+    if (labs(time_diff) > _SEC_PER_DAY) {
+        /* We had a jump larger than a day */
+        slapi_log_err(SLAPI_LOG_INFO, "csngen_new_csn",
```
tbordaz, line 594: If we have a topology without any update for a day (direct or replicated). is _csngen_adjust_local_time called more frequently than a day ? I wonder if this log could occur on an update after a long quiet period. In such case there is not an alarming big jump but just a long period of quiet.

ldap/servers/slapd/csngen.c
```diff
@@ -597,94 +557,93 @@ _csngen_cmp_callbacks(const void *el1, const void *el2)
...
+                "Detected large jump in CSN time.  Delta: %ld (current time: %ld  vs  previous time: %ld)\n",
+                time_diff, cur_time, gen->state.sampled_time);
+    }
+    if (!ignore_time_skew && (gen->state.local_offset-time_diff > CSN_MAX_TIME_ADJUST)) {
```
tbordaz: please just add ' ' before/after '-'

ldap/servers/slapd/csngen.c
```diff
@@ -597,94 +557,93 @@ _csngen_cmp_callbacks(const void *el1, const void *el2)
...
+                      gen->state.sampled_time,
+                      gen->state.local_offset,
+                      gen->state.remote_offset);
+    }
 
-        gen->state.sampled_time = cur_time;
-        gen->state.local_offset = MAX_VAL(gen->state.local_offset, labs(time_diff));
-        gen->state.seq_num = 0;
+    gen->state.sampled_time = cur_time;
+    gen->state.local_offset = MAX_VAL(0, gen->state.local_offset - time_diff);
```
tbordaz, line 618: Do you mind to add a comment that local_offset is consumed as clock is regularly ticking. It looks obvious but when back to that code it is always surprising to local_offset decrease while we say csn are always going forward.

ldap/servers/slapd/csngen.c
```diff
@@ -853,3 +812,78 @@ _csngen_local_tester_main(void *data)
...
+    int vtime = _csngen_tester_state ;
+    tp->tv_sec = 0x1000000 + vtime + 2 * _csngen_tester_state_rid;
+    if (_csngen_tester_state_rid == 3) {
+        /* tp->tv_sec += _mynoise(vtime, 10, 1.5); */
+        tp->tv_sec += _mynoise(vtime, 30, 15);
+    }
+    return 0;
+}
+
+/* Mimic a fully meshed multi suplier topology */
```
tbordaz, line 843: Is it possibly to plug this test function from the csngen task (_csngen_gen_tester_main) ?

tbordaz (commented):
As always this patch is crystal clear :) !
I have just minor comments/questions

ldap/servers/slapd/csngen.c
mreynolds389: It will log the message, I had originally added this logging line, and it will report an INFO message when it sees large jumps, even idle jumps.  I had proposed we could bump the threshold to something larger than a day.  30 days or something like that just to keep the logs quieter.

ldap/servers/slapd/csngen.c
progier389: Hum not sure we should change the name of a function in the plugin API, that said we could always redefine a new slapi_clock_utc_gettime function ...

ldap/servers/slapd/csngen.c
progier389: The idea is to avoid a beat if the server are desynchronized by a fraction of second 
  ( increasing the remote_offset each time the clock tick ) 
  if there is already a local offset it would be decreased at next clock tick so there is no need to increase it further

ldap/servers/slapd/csngen.c
progier389: Added (and used) a define:
#define TIME_DIFF_WARNING_DELAY  (30*_SEC_PER_DAY)  /* log an info message when difference
                                                       between clock is greater than this delay */

ldap/servers/slapd/csngen.c
progier389: Done

ldap/servers/slapd/csngen.c
progier389: In fact this comment already exists (in csn_adjust_time header):
```
 * local offset: it is increased if
 *   - system time is going backward (compare sampled_time)
 *   - if 'csn' argument is ahead of csn that the csn generator would generate
 *     AND diff('csn', csngen.new_csn) < remote_offset
 *     then the diff "pour" into local_offset
 *  It is decreased as the clock is ticking, local offset is "consumed" as
 *  sampled_time progresses.
```

ldap/servers/slapd/csngen.c
progier389: I do not think it is pertinent: 
   as I understanding it _csngen_gen_tester_main is one of the thread handler for the csngen_test() framework 
   which run several threads doing several tasks on csn ( 
     testing mostly that the csn handling function does not return errors 
     and just logging the generated csn in error log (without testing the result)
   
 csngen_multi_suppliers_test is a different way of testing the csn (you could eventually call it from csngen_test() before starting the threads )
    it is sequential and chec
  but IMHO it would not make sense to call it while running other test threads: 
   1) it is not designed to be multi threaded (as it uses a global counter as clock)
       but as a sequential set of events (so it can checks that csn timestamps always increase) 
   2) It is easier to analyze the log output if it is not polluted by others test threads output

ldap/servers/slapd/csngen.c
progier389: I think I will go for a tradeoff:
 using a slapi_clock_utc_gettime define that point to slapi_clock_gettime to avoid to duplicate or redefine a pre-existing plugin API function

ldap/servers/slapd/csngen.c
tbordaz: okay. thanks

ldap/servers/slapd/csngen.c
tbordaz: perfect

ldap/servers/slapd/csngen.c
tbordaz: The idea of the csgen task is to be able to run unit tests on a system where csngen is behaving strangely. This without building a debug version.
It is fine to no run csngen_multi_suppliers_tests with the task. if it is needed we can plug it later with a different ticket.

tbordaz (approved):
The patch looks good to me. ACK

PR #4949, opened 2021-10-12
Issue 3584 - Fix PBKDF2_SHA256 hashing in FIPS mode

droideck:
Issue Description: Use PK11_Decrypt function to get hash data
because PK11_ExtractKeyValue function is forbidden in FIPS mode.
We can't extract keys while in FIPS mode. But we use PK11_ExtractKeyValue
for hashes, and it's not forbidden.

We can't use OpenSSL's PBKDF2-SHA256 implementation right now because
we need to support an upgrade procedure while in FIPS mode (update
hash on bind). For that, we should fix existing PBKDF2 usage, and we can
switch to OpenSSL's PBKDF2-SHA256 in the following versions.

Fix Description: Use PK11_Decrypt function to get the data.

Enable TLS on all CI test topologies while in FIPS because without
that we don't set up the NSS database correctly.

Add PBKDF2-SHA256 (OpenSSL) to ldif templates, so the password scheme is
discoverable by internal functions.

https://github.com/389ds/389-ds-base/issues/3584

Reviewed by: ?

### DIFF
--- a/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
+++ b/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
@@ -91,10 +91,11 @@ pbkdf2_sha256_extract(char *hash_in, SECItem *salt, uint32_t *iterations)
 SECStatus
 pbkdf2_sha256_hash(char *hash_out, size_t hash_out_len, SECItem *pwd, SECItem *salt, uint32_t iterations)
 {
-    SECItem *result = NULL;
     SECAlgorithmID *algid = NULL;
     PK11SlotInfo *slot = NULL;
     PK11SymKey *symkey = NULL;
+    SECItem *wrapKeyData = NULL;
+    SECStatus rv = SECFailure;
 
     /* We assume that NSS is already started. */
     algid = PK11_CreatePBEV2AlgorithmID(SEC_OID_PKCS5_PBKDF2, SEC_OID_HMAC_SHA256, SEC_OID_HMAC_SHA256, hash_out_len, iterations, salt);
@@ -104,7 +105,6 @@ pbkdf2_sha256_hash(char *hash_out, size_t hash_out_len, SECItem *pwd, SECItem *s
         slot = PK11_GetBestSlotMultiple(mechanism_array, 2, NULL);
         if (slot != NULL) {
             symkey = PK11_PBEKeyGen(slot, algid, pwd, PR_FALSE, NULL);
-            PK11_FreeSlot(slot);
             if (symkey == NULL) {
                 /* We try to get the Error here but NSS has two or more error interfaces, and sometimes it uses none of them. */
                 int32_t status = PORT_GetError();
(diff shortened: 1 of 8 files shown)


ldap/servers/slapd/main.c
```diff
@@ -2931,6 +2931,9 @@ slapd_do_all_nss_ssl_init(int slapd_exemode, int importexport_encrypt, int s_por
      * is enabled or not. We use NSS for random number generation and
      * other things even if we are not going to accept SSL connections.
      * We also need NSS for attribute encryption/decryption on import and export.
+     *
+     * It's important to remember that while in FIPS mode the administrator should always enable
```
progier389, line 2935: IMHO we should log an error if FIPS mode is enabled but not the security

src/lib389/lib389/utils.py
```diff
@@ -1434,3 +1434,14 @@ def is_valid_hostname(hostname):
...
+
+def is_fips():
+    if os.path.exists('/proc/sys/crypto/fips_enabled'):
+        with open('/proc/sys/crypto/fips_enabled', 'r') as f:
+            state = f.readline().strip()
+            if state == '1':
+                return True
+            else:
+                return False
+
```
progier389, line 1450: Should also return False if the path does not exists

progier389 (commented):
Seems OK except two remarks. 
A last question have you tested with ASAN that there are no new memory leaks ?

ldap/servers/slapd/main.c
mreynolds389: +1

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -123,18 +123,60 @@ pbkdf2_sha256_hash(char *hash_out, size_t hash_out_len, SECItem *pwd, SECItem *s
...
+    /*
+     * First, we need to generate a wrapped key for PK11_Decrypt call:
+     * slot is the same slot we used in PK11_PBEKeyGen()
+     * 256 bits / 8 bit per byte
+     */
+    PK11SymKey *wrapKey = PK11_KeyGen(slot, CKM_AES_ECB, NULL, 256/8, NULL);
+    PK11_FreeSlot(slot);
+    if (wrapKey == NULL) {
+        slapi_log_err(SLAPI_LOG_ERR, (char *)schemeName, "Unable to generate a wrapped key.\n");
+        return SECFailure;
```
mreynolds389, line 135: The logging line should list the function name:  pbkdf2_sha256_hash

mreynolds389: Same for the logging functions below...

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -123,18 +123,60 @@ pbkdf2_sha256_hash(char *hash_out, size_t hash_out_len, SECItem *pwd, SECItem *s
...
+     */
+    rv = PK11_Decrypt(wrapKey, CKM_AES_ECB, NULL, result, &result_len, hash_out_len, wrapKeyData->data, wrapKeyData->len);
+    PK11_FreeSymKey(symkey);
+    PK11_FreeSymKey(wrapKey);
+    SECITEM_FreeItem(wrapKeyData, PR_TRUE);
+
+    if (rv == SECSuccess) {
+        if (result != NULL && result_len <= hash_out_len) {
+            memcpy(hash_out, result, result_len);
+            slapi_ch_free_string((char **)&result);
```
mreynolds389: This is a case were slapi_ch_free() might be more appropriate :-)

mreynolds389:
This looks pretty good, but please do an ASAN run just to be safe.  Thanks!

progier389 (approved):
LGTM

Firstyear:
> Okay, the issue can be fixed by enabling Enforced FIPS mode for libgcrypt: https://www.gnupg.org/documentation/manuals/gcrypt/Enabling-FIPS-mode.html

Where are we using libgcrypt? We shouldn't use it as it's a known insecure crypto library ....

Firstyear:
Oh lord that's a bit suboptimal ... so long as we aren't directly including it that's fine, but wow ....

tbordaz (commented):
The patch looks very nice to me.
Just a question, the fix description says that the default scheme is now PBKDF2 in FIPS as well. I did not find this change in the patch. I would expect a change in pw_name2scheme.

tbordaz (approved):
ACK

PR #4955, opened 2021-10-20
Issue 4954 - pytest is killed by OOM killer when the whole test suite…

### LINKED ISSUE #4954: pytest is killed by OOM killer when the whole test suite is executed
**Issue Description**
After 2a9df10303c4902a816a64b805448f31380a2728 we started to collect dirsrv logs and attach them to html report. This works fine in PR CI when a single test module is executed. But when *all* tests are executed, pytest starts to consume a lot of memory, which on 4GB VMs leads to OOM kill event, leaving test run without results. 

We should read logs into memory only when html report is requested.

vashirov:
… is executed

Bug Description:
After 2a9df10 we started to collect dirsrv logs and attach them to
html report. This works fine in PR CI when a single test module is
executed. But when all tests are executed, pytest starts to consume
a lot of memory, which on 4GB VMs leads to OOM kill event, leaving
test run without results.

Fix Description:
We should read logs into memory only when html report is requested.

Fixes: https://github.com/389ds/389-ds-base/issues/4954

Reviewed by: ???

### DIFF
--- a/dirsrvtests/conftest.py
+++ b/dirsrvtests/conftest.py
@@ -106,18 +106,20 @@ def pytest_runtest_makereport(item, call):
     report = outcome.get_result()
     extra = getattr(report, 'extra', [])
     if report.when == 'call' and pytest_html is not None:
-        for f in glob.glob(f'{p.run_dir}/ns-slapd-*san*'):
-            with open(f) as asan_report:
-                text = asan_report.read()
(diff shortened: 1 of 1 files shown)


Firstyear (approved):

droideck (approved):
LGTM too!

PR #4960, opened 2021-10-25
Issue 4959 - BUG - Invalid /etc/hosts setup can cause isLocalHost

### LINKED ISSUE #4959: Invalid /etc/hosts setup can cause isLocalHost to fail.
```
DEBUG: no ip address for b'susetest'
DEBUG: Allocate <class 'lib389.DirSrv'> with b'susetest':636
DEBUG: Allocate <class 'lib389.DirSrv'> with b'susetest':636
DEBUG: Instance allocated
DEBUG: 'NoneType' object has no attribute 'get'
  File "/sbin/dsctl", line 142, in <module>
    result = args.func(inst, log, args)
  File "/usr/lib/python3.8/site-packages/lib389/cli_ctl/instance.py", line 55, in instance_status
    if inst.status() is True:
  File "/usr/lib/python3.8/site-packages/lib389/__init__.py", line 1185, in status
    if self.with_systemd():
  File "/usr/lib/python3.8/site-packages/lib389/__init__.py", line 1678, in with_systemd
    return self.ds_paths.with_systemd
  File "/usr/lib/python3.8/site-packages/lib389/paths.py", line 221, in __getattr__
    return ensure_str(self._config.get(SECTION, name).format(instance_name=self._serverid))
AttributeError: 'NoneType' object has no attribute 'get'
ERROR: Error: 'NoneType' object has no attribute 'get'

```

This was found in the suse openqa environment, which appears to have a very odd method to setup the hostname where the hostname is never set into /etc/hosts. I think we can't fully defend from a broken setup like this, but we also shouldn't raise an exception like this.

Firstyear:
Bug Description: During dsctl localhost status, this can fail
to detect if the instance isLocal if /etc/hosts is not correctly
configured. This starts inside of utils with no ip for "hostname"
but then causes paths.py to fail as there is no config.

Fix Description: Improve the error condition when _config
is not present either due to remote instance is offline, or
an invalid dns setup with /etc/hosts

fixes: https://github.com/389ds/389-ds-base/issues/4959

Review by: ???

### DIFF
--- a/src/lib389/lib389/paths.py
+++ b/src/lib389/lib389/paths.py
@@ -217,17 +217,23 @@ class Paths(object):
                 v = v.split('/')[1].split()[0]
             return v
         # Else get from the config
-        if self._serverid is not None:
-            return ensure_str(self._config.get(SECTION, name).format(instance_name=self._serverid))
+        if self._config:
+            if self._serverid is not None:
+                return ensure_str(self._config.get(SECTION, name).format(instance_name=self._serverid))
+            else:
+                return ensure_str(self._config.get(SECTION, name))
         else:
-            return ensure_str(self._config.get(SECTION, name))
+            raise LookupError("""Invalid state
+No paths config (defaults.inf) could be found.
+This may be because the remote instance is offline, or your /etc/hosts is misconfigured for a local instance
+""")
 
     @property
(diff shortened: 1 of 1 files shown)


progier389 (approved):
ACK

Issue #4731, opened 2021-04-21
Demoting a supplier to a consumer crashes the server


tbordaz:
**Issue Description**
When demoting a supplier to a consumer, the server calls cl5CleanRUV on a replica that has no more changelog. cl5CleanRuv dereference the changelog that is NULL and crash

**Package Version and Platform:**
It happens F34/rawhide

**Steps to Reproduce**
Running the following command crash the server
dsconf -D "cn=Directory Manager" ldap://server.example.com replication demote --suffix="dc=example,dc=com" --newrole="consumer"

**Expected results**
It should not crash

tbordaz:
The crash occurs when the replica is modified to a consumer role:
update MOD_REPL
nsDS5ReplicaType: 2
nsDS5Flags: 0
nsDS5ReplicaId: 65535

<pre>
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[0]
$19 = {mod_op = 130, mod_type = 0x7fc4a5006e40 "nsDS5ReplicaType", mod_vals = {modv_strvals = 0x7fc4a50355d0,
    modv_bvals = 0x7fc4a50355d0}}
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[0]->mod_vals.modv_bvals[0]
$25 = {bv_len = 1, bv_val = 0x7fc4a5032240 "2"}
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[1]
$26 = {mod_op = 130, mod_type = 0x7fc4a5035650 "nsDS5Flags", mod_vals = {modv_strvals = 0x7fc4a5035660, modv_bvals = 0x7fc4a5035660}}
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[1]->mod_vals.modv_bvals[0]
$27 = {bv_len = 1, bv_val = 0x7fc4a5032248 "0"}
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[2]
$28 = {mod_op = 130, mod_type = 0x7fc4a5035680 "nsDS5ReplicaId", mod_vals = {modv_strvals = 0x7fc4a5035690, modv_bvals = 0x7fc4a5035690}}
(gdb) print *pb->pb_op->o_params.p.p_modify.modify_mods[1]->mod_vals.modv_bvals[0]
$29 = {bv_len = 1, bv_val = 0x7fc4a5032248 "0"}
[...]

PR #4963, opened 2021-10-26
Issue 4962 - Fix various UI bugs part 1

mreynolds389:
Fix Description:

Bug 2016022 - Cockpit UI: UI is incorrectly saying "Create the Sub Suffix entry"
Bug 2015951 - Cockpit UI: Database tab ---> Export Database/replicaton
data
Bug 2015221 - Cockpit UI: UX Bugs Server Settings ->Tuning and Limits
Bug 2015139 - Configuration for Import Cache Settings is not saved
Bug 2015127 - No message when configuring Global Database Configuration
Bug 2014924 - Cockpit UI: UX Bugs and other cockpit GUI related defects

relates: https://github.com/389ds/389-ds-base/issues/4962

Reviewed by: ?

PR #4968, opened 2021-10-27
Issue 4731 - Promoting/demoting a replica can crash the server

mreynolds389:
Bug Description:  

The server will crash if you demote a supplier with no changelog.

Fix Description:  

Check if the changelog pointer is NULL before dereferencing it.

relates: https://github.com/389ds/389-ds-base/issues/4731

### DIFF
--- a/dirsrvtests/tests/suites/replication/promote_demote_test.py
+++ b/dirsrvtests/tests/suites/replication/promote_demote_test.py
@@ -1,0 +1,67 @@
+import logging
+import pytest
+import os
+from lib389._constants import DEFAULT_SUFFIX,  ReplicaRole
+from lib389.topologies import topology_m1h1c1 as topo
+from lib389.replica import Replicas, ReplicationManager, Agreements
+
+pytestmark = pytest.mark.tier1
+
+log = logging.getLogger(__name__)
+
+def test_promote_demote(topo):
+    """Test promoting and demoting a replica
+
+    :id: 75edff64-f987-4ed5-a03d-9bee73c0fbf0
+    :setup: 2 Supplier Instances
+    :steps:
+        1. Promote Hub to a Supplier
+        2. Test replication works
+        3. Demote the supplier to a consumer
+        4. Test replication works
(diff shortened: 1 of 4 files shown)


ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -4403,6 +4414,10 @@ void
 cl5CleanRUV(ReplicaId rid, Replica *replica)
 {
     cldb_Handle *cldb = replica_get_cl_info(replica);
+
+    if (cldb == NULL) {
+        return;
```
Firstyear, line 4419: Do we need a error message here?

mreynolds389: I don't think so because it will probably get logged in one of the other functions that checks for cldb.  I just don't want the errors log filling up with redundant/duplicate messages

Firstyear (approved):

dirsrvtests/tests/suites/replication/promote_demote_test.py
```diff
@@ -0,0 +1,65 @@
+import logging
+import pytest
+import os
+from lib389._constants import DEFAULT_SUFFIX,  ReplicaRole
+from lib389.topologies import topology_m1h1c1 as topo
+from lib389.replica import Replicas, ReplicationManager, Agreements
+
```
droideck, line 7: Small note, if you want this test to run as a PR CI, you can add `pytestmark = pytest.mark.tier1` line here.

droideck: Oh, actually, it had run... (and failed) Discard my previous comment! (or may be @sgouvern can fix me, in case we need the pytestmark for something else here)

P.S. so if I get you correctly, we merge #4963 first, and then this PR (#4968) will PASS. Do I get it right?

mreynolds389: Correct and I just merged it

Issue #4973, opened 2021-10-31
Failure during ipa-server-install related to /run owned by dirsrv


flo-renaud:
**Issue Description**
The nightly tests for IPA, using the copr repository @389ds/389-ds-base-nightly,  started failing with 389-ds-base-2.1.0-20211030.fc34.x86_64 in `ipa-server-install` call.

**Package Version and Platform:**
 - Platform: Fedora 34
 - Package and version: 389-ds-base-2.1.0-20211030.fc34.x86_64

**Steps to Reproduce**
```
ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --auto-forwarders -a $PASSWORD -p $PASSWORD -U
...
Configuring the web interface (httpd)
  [1/21]: stopping httpd
  [2/21]: backing up ssl.conf
  [3/21]: disabling nss.conf
  [4/21]: configuring mod_ssl certificate paths
  [5/21]: setting mod_ssl protocol list
  [6/21]: configuring mod_ssl log directory
  [7/21]: disabling mod_ssl OCSP
  [8/21]: adding URL rewriting rules
  [9/21]: configuring httpd
  [10/21]: setting up httpd keytab
  [11/21]: configuring Gssproxy
  [12/21]: setting up ssl
  [13/21]: configure certmonger for renewals
  [14/21]: publish CA cert
  [15/21]: clean up any existing httpd ccaches
  [error] CalledProcessError: CalledProcessError(Command ['/usr/bin/systemd-tmpfiles', '--create', '--prefix', '/run/ipa/ccaches'] returned non-zero exit status 73: 'Detected unsafe path transition /run (owned by dirsrv) → /run/ipa (owned by root) during canonicalization of /run/ipa.\n')
CalledProcessError(Command ['/usr/bin/systemd-tmpfiles', '--create', '--prefix', '/run/ipa/ccaches'] returned non-zero exit status 73: 'Detected unsafe path transition /run (owned by dirsrv) → /run/ipa (owned by root) during canonicalization of /run/ipa.\n')
The ipa-server-install command failed. See /var/log/ipaserver-install.log for more information
Exit code: 1
```

**Expected results**
The command should succeed.

**Additional context**
Tests were succeeding last week, the issue seems related to this recent [commit](https://github.com/389ds/389-ds-base/commit/63243a4af49e29234d629f72e1513bc891de7ec1) on 389ds master branch: Issue 4092 - systemd-tmpfiles warnings .

[...]

PR #4975, opened 2021-11-01
Issue 4973 - update snmp to use /run/dirsrv for PID file

mreynolds389:
Description:  

Previously SNMP would write the agent PID file directly under /run (or /var/run), but this broke a CI test after updating lib389/defaults.inf to use /run/dirsrv.

Instead of hacking the CI test, I changed the path snmp uses to:  /run/dirsrv/  Which is where it should really be written anyway.

relates: https://github.com/389ds/389-ds-base/issues/4973


vashirov:
systemd service template for snmp agent also has an older path: 
https://github.com/389ds/389-ds-base/blob/master/wrappers/systemd-snmp.service.in#L12

mreynolds389:
> systemd service template for snmp agent also has an older path:
> 
> https://github.com/389ds/389-ds-base/blob/0552081fe736b5c86d3c7e3d640f8a8b170d1dc6/wrappers/systemd-snmp.service.in#L12

@vashirov thanks!  Fixed

vashirov (approved):
LGTM!

Issue #4978, opened 2021-11-02
Improve installer robustness


mreynolds389:
**Issue Description**

When run in a container the server can fail to start because the installer set the db_home_dir to /dev/shm, but in containers the default size of /dev/shm is too small for libdb.  We should detect if we are in a container and not set db_home_dir to /dev/shm.

During instance removal, if an instance was not properly created then it can not be removed either.  Make the uninstall more robust to accept some errors and continue removing the instance.

mreynolds389:
b0e890bf6..b1efe0d4f  389-ds-base-2.0 -> 389-ds-base-2.0
1b8306956..599bb15b2  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
9a6cfddee..1d2e85439  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
Fixed porting issues with python command subprocess.run() 

5f92e1d6f..92ad2b917  master -> master
608d4b372..33c85c560  389-ds-base-2.0 -> 389-ds-base-2.0
f75d68b20..2499a379d  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
e8e910b64..e7bc3b530  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #4980, opened 2021-11-02
Issue 4978 - make installer robust

mreynolds389:
Description:  

When run in a container the server can fail to start because the installer sets the db_home_dir to /dev/shm, but in containers the default size of /dev/shm is too small for libdb. We should detect if we are in a container and not set db_home_dir to /dev/shm.

During instance removal, if an instance was not properly created then it can not be removed either. Make the uninstall more robust to accept some errors and continue removing the instance.

relates: https://github.com/389ds/389-ds-base/issues/4978


Firstyear (approved):

tbordaz (approved):
LGTM

Issue #4972, opened 2021-10-28
gecos with IA5 introduces a compatibility issue with previous releases where it was DirectoryString


tbordaz:
**Issue Description**
For years 'gecos' was DirectoryString (UTF8), with #[50933](https://github.com/389ds/389-ds-base/commit/0683bcde1b667b6d0ca6e8d1ef605f17c51ea2f7#diff-8868991d59b4f41e540fa8cfe5d44561086e8e9ec81db1379e1f6bd37f38151bR21) it was restricted to IA5 (ascii)

This IA5 definition conforms rfc2307 but is a problem for existing deployments where 'gecos' attribute already contains UTF8. 

In addition, because of schema learning IA5 does not overwrite UTF8 and schema will not be propagated. I would even expect IA5 to be eventually overwritten by UTF8.

The change of syntax of 'gecos' from less restrictive to more restrictive is not compatible with existing deployment and needs to be reverted.

**Package Version and Platform:**
  since F33 (1.4.4.4) and 2.0.1

**Steps to Reproduce**
Creates M2, provision entries. Stop M1-M2, update 99users.ldif on M1 with former definition

    ( 1.3.6.1.1.1.1.2 NAME 'gecos' DESC 'Standard LDAP attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN 'RFC 2307' )

Enable replication log. update M1 schema with a dummy definition (update nsschemacsn). Update an entry on M1 with UTF8 of gecos.
The schema will not be replicated ('gecos' definition diverges) although replication should be fine.
We should see UTF8 gecos value on both sides, although on M2 it breaks the schema.

**Expected results**
Replication of the schema should be done

Firstyear:
Likely this happened with the 2307bis changes. I'm okay to put this back to utf8 though.

tbordaz:
aeed1de95..5e00102e2  master
b1efe0d4f..7570259aa  389-ds-base-2.0
599bb15b2..f75d68b20  389-ds-base-1.4.4
1d2e85439..edb23b870  389-ds-base-1.4.3

PR #4981, opened 2021-11-03
Issue 4972 - gecos with IA5 introduces a compatibility issue with pre…

tbordaz:
…vious

 releases where it was DirectoryString

Bug description:
       For years 'gecos' was DirectoryString (UTF8), with #50933 it was restricted to IA5 (ascii)
       https://github.com/389ds/389-ds-base/commit/0683bcde1b667b6d0ca6e8d1ef605f17c51ea2f7#

       IA5 definition conforms rfc2307 but is a problem for existing deployments
       where entries can have 'gecos' attribute value with UTF8.

Fix description:
       Revert the definition to of 'gecos' being Directory String

relates: https://github.com/389ds/389-ds-base/issues/4972

Reviewed by:

Platforms tested: F34

### DIFF
--- a/ldap/schema/10rfc2307compat.ldif
+++ b/ldap/schema/10rfc2307compat.ldif
@@ -21,9 +21,9 @@ attributeTypes: (
 attributeTypes: (
   1.3.6.1.1.1.1.2 NAME 'gecos'
   DESC 'The GECOS field; the common name'
-  EQUALITY caseIgnoreIA5Match
-  SUBSTR caseIgnoreIA5SubstringsMatch
-  SYNTAX 1.3.6.1.4.1.1466.115.121.1.26
+  EQUALITY caseIgnoreMatch
+  SUBSTR caseIgnoreSubstringsMatch
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15
   SINGLE-VALUE
   )
 attributeTypes: (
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM.  , just a detail:I would have preferred that you  used wait_for_replication rather than   sleep ...

jchapma (commented):
If you make the changes suggested by @progier389 and @mreynolds389, we wont need the time and pdb import. Other than that it LGTM

ldap/schema/10rfc2307compat.ldif
```diff
@@ -21,9 +21,7 @@ attributeTypes: (
 attributeTypes: (
   1.3.6.1.1.1.1.2 NAME 'gecos'
   DESC 'The GECOS field; the common name'
-  EQUALITY caseIgnoreIA5Match
-  SUBSTR caseIgnoreIA5SubstringsMatch
-  SYNTAX 1.3.6.1.4.1.1466.115.121.1.26
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15
```
Firstyear, line 26: I think we need to leave case ignore matching here though if there is a utf8 equivalent.

Firstyear (changes requested):

ldap/schema/10rfc2307compat.ldif
tbordaz: You are right, [RFC2307](https://datatracker.ietf.org/doc/html/rfc2307#section-3) says it is ignore case matching. Change done

Firstyear (approved):

Issue #4970, opened 2021-10-28
Add support for recursively deleting subentries


antbob:
**Is your feature request related to a problem? Please describe.**
If a parent entries contains any subentry object children there is no way to recursively delete such entry because neither the Subentries Control [1.3.6.1.4.1.4203.1.10.1] (used by OpenLDAP's ldapdelete) nor Tree Delete Control [1.2.840.113556.1.4.805] are supported and so the only resort is to delete such subentries first 1 by 1 before their parent entry can be deleted.

**Describe the solution you'd like**
Implementing support for Subentries Control [1.3.6.1.4.1.4203.1.10.1] as per RFC3676 should be pretty easy and straightforward and would allow OpenLDAP's ldapdelete pruning iteration to work properly.

Implementing support for Tree Delete Control [1.2.840.113556.1.4.805] as per draft-armijo-ldap-treedelete-02 is a bit more involved however it can be extremely useful for recursive delete operations on large subtrees mainly due to performance considerations.

Ideally it would be nice to have both implemented of course.

**Describe alternatives you've considered**
Currently the only alternative I know of is to grab the list of subentry objects DNs with ldapsearch and then feed them to ldapdelete with some slicing and dicing in between so those DNs are acceptable.

**Additional context**
N/A

jchapma:
Need more info from Anton, is this a nice to have feature or is it a request from someone.

antbob:
@jchapma just a nice to have features for a modern directory server. to give you some context this request in particular came about due to RHDS production use and running into issue of having to manually delete a bunch of sub-entries and while in this specific case the workaround was ok as there were around only 50+ subentries to deal with and it was more or less a one off thing in other scenarios it would be very beneficial to have these features for the reasons i have described.

antbob:
Something I forgot to link here is that OpenLDAP ldapdelete currently has a bug where it fails in some cases. I have filed this ITS and submitted a patch to address it as well https://bugs.openldap.org/show_bug.cgi?id=9737 then I totally forgot about it but it looks like it got picked up by Howard over there so hopefully a fix will land eventually.

mreynolds389:
> Something I forgot to link here is that OpenLDAP ldapdelete currently has a bug where it fails in some cases. I have filed this ITS and submitted a patch to address it as well https://bugs.openldap.org/show_bug.cgi?id=9737 then I totally forgot about it but it looks like it got picked up by Howard over there so hopefully a fix will land eventually.

[...]

PR #4985, opened 2021-11-04
Issue 4970 - Add support for recursively deleting subentries

antbob:
Bug Description:

Currently LDAP Subentries Control [1.3.6.1.4.1.4203.1.10.1] (used by OpenLDAP's ldapdelete among others)
is not implemented so any LDAP client attempting to use it would fail.

Fix Description:

This patch implements LDAP Subentries Control support in accordance with RFC 3672 on the server side.
On the client side OpenLDAP client tools already support this control.

relates: https://github.com/389ds/389-ds-base/issues/4970

Author: antbob

Reviewed by: ???


antbob:
Some additional comments related to this patch.

I see some replication tests failures, potentially they could be related to this patch but please check if those are something that you might have seen before and thus unrelated eg timing or setup issues.

I did do some basic manual testing but did not write any unit tests. The main reason is that I cannot get your test setup to work on my local machine. Perhaps the instructions I found are outdated ? Also writing unit tests would be a bit more involved in this case because I'm not familiar with pytest and also python-ldap does not support subentries control either so it would need a custom one tho it should be easy to do.

So I'm leaving this patch as is for now just to see if you are ok with general approach taken and if so then we can take it from there I guess.

mreynolds389:
Yeah I'm seeing the same regressions caused by this PR.  Obviously this is messing up how the RUV tombstone entry is behaving.  I suggest adding some debug logging to figure out why it's failing and run some of the failing tests like so:

cd 389-ds-base/dirsrvtests/tests/suites/replication

# sudo DEBUGGING=true py.test ./acceptance_test.py::test_modrdn_entry

If you need help running the tests just let us know.  Thanks!

antbob:
@mreynolds389 I'm getting this

E               subprocess.CalledProcessError: Command '['/opt/dirsrv/bin/pwdhash', 'Zkry5y2EzPXlaUV11cZcf.iy5yLXdKgKDJKPf3luiNNTwWmm1q9jUsaA1Ao3KHax8']' returned non-zero exit status 1.

running /opt/dirsrv/bin/pwdhash manually yields this

/opt/dirsrv/bin/pwdhash Zkry5y2EzPXlaUV11cZcf.iy5yLXdKgKDJKPf3luiNNTwWmm1q9jUsaA1Ao3KHax8
The plugin entry [cn=PBKDF2,cn=Password Storage Schemes,cn=plugins,cn=config] in the configfile /opt/dirsrv/share/dirsrv/data/template-dse.ldif was invalid.  Failed to load plugin's init function.

antbob:
@mreynolds389  and if i remove related entries from ldif then pwdhash would just crash like this

Process 3079625 (pwdhash) of user 0 dumped core.

Stack trace of thread 3079625:
#0  0x0000000000402657 main (pwdhash + 0x2657)
#1  0x00007f8bfeac1b75 __libc_start_main (libc.so.6 + 0x27b75)
#2  0x00000000004027ae _start (pwdhash + 0x27ae)

and this is from gdb

#0  0x0000000000402657 in main (argc=2, argv=0x7fffc42e2278) at ldap/servers/slapd/tools/pwenc.c:230

any suggestions ?!

mreynolds389:
@antbob how are you producing this crash?  Is this when you run the py.test command I suggested?  

Looks like you are using a PREFIX build, which I never personally use.  So I think you need to provide "PREFIX=/opt" to py.test.  Sometimes turning off selinux helps with the CI tests, but I haven't had to do that in a while.

Maybe @tbordaz might have some suggestions as he uses PREFIX builds.

vashirov:
I get this on builds *without* Rust:
```
The plugin entry [cn=PBKDF2,cn=Password Storage Schemes,cn=plugins,cn=config] in the configfile /usr/share/dirsrv/data/template-dse.ldif was invalid.  Failed to load plugin's init function.
```

antbob:
@mreynolds389  with 'sudo DEBUGGING=true py.test ./acceptance_test.py::test_modrdn_entry' you suggested.

this is not a prefix build. i have tried using prefix build before and it failed horribly with god knows what. so this one is done like this

make clean
make distclean
./autogen.sh
./configure
make && make install

then run pytest. i have even tried doing rpm build and re-installing rpms it produced just to see if that might help but it did not.

i also spoke to Thierry about prefix builds and he said hes not doing them anymore bc they are broken.

antbob:
@vashirov how do you resolve it ?

vashirov:
I think it's a regression from a1680a2b89239ac03a46c9bfe5898c1b52ab3be6. It added password storage schemes to dse.ldif template which use libpwdchan-plugin written in Rust.

antbob:
@vashirov thanx! i did configure --enable-rust and pytest is slowly crawling now, will see how it goes.

mreynolds389:
Another issue is that "make install" does not build rust.  You have to build the rpms using:

make -f rpm.mk  rpms

If this fails to build because of rust failures then do:

make -f rpm.mk update-cargo-dependencies download-cargo-dependencies rpms
  

Guess we need a need ticket for "make" to do the rust stuff too.  But for me this all works as of today, but I think I start with a proper rpm (with rust) and when I run "make install" it does not break anything.

Now as for PREFIX builds, your output shows its looking for "/opt/dirsrv/bin/pwdhash", but on my system its "/usr/sbin/pwdhash".  So it looks like you are using a prefix build...

antbob:
@mreynolds389 if i run make clean make distclean and then autogen and configure without a prefix why would it do a prefix build ? in any case i have never used '/opt/dirsrv' as prefix ever so i have no idea where that comes from. spooky remnants of Halloween most likely.

vashirov:
I think some of the autotools artifacts are not removed completely. Try to nuke them with `git clean -fxd`.

antbob:
@vashirov yeah i have noticed that, it doesnt remove some stuff that got generated from .in but /opt/dirsrv is some sort of default prefix for some reason. this is is your configure.ac file:

# the default prefix - override with --prefix or --with-fhs
AC_PREFIX_DEFAULT([/opt/$PACKAGE_NAME])

# If we have no prefix specified, we need to fix the prefix variable.
# If we don't what happens is $prefixdir ends up as NONE, and then
# later configure changes $prefix to $ac_default_prefix underneath us.
if test "$prefix" = "NONE"; then
    prefix=$ac_default_prefix
fi

and the PACKAGE_NAME defined to be 'dirsrv'.

antbob:
Thanx for your help with build/test stuff guys! I have finally managed to run those tests properly with right binaries and figure what the problem was.

Apparently you have tombstones implemented as normal entries and not subentries and that is why you have to do explicit filtering for them. So that is fixed now in the patch and all replication acceptance tests pass now.

antbob:
Added some unit tests for subentries. The tests are mostly working fine for me now however some are failing badly with 'server down' type errors which seems to be related to my hostname lookups and i dunno why it happens to some tests and not others. The whole suite tests take about 3 hours to complete on my laptop so i'm not very keen on re-running them locally. The test framework seems quite inefficient as it setups/tearsdown for ?each test? instead of quick cleanup which should suffice in most cases but i digress. Anyways, please check it out if you can kick another test run on a proper build machine to see if there are any further regressions from this patch.

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -0,0 +1,201 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 RED Hat, Inc.
```
vashirov: nit: capitalization

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -0,0 +1,201 @@
...
+from lib389._constants import *
+from lib389.topologies import topology_st as topo
+from lib389.idm.user import UserAccount, UserAccounts
+
+log = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.tier0
+
+"""
+This BooleanControl class is taken from python-ldap,
```
vashirov, line 29: I think it would be better to put it in https://github.com/389ds/389-ds-base/blob/master/src/lib389/lib389/_controls.py in case we'll want to use it for lib389/ds* tools, until the fixed version of python-ldap appears in distros.

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -0,0 +1,201 @@
...
+import ldap
+from ldap.controls import LDAPControl
+from lib389.utils import *
+from lib389._constants import *
+from lib389.topologies import topology_st as topo
+from lib389.idm.user import UserAccount, UserAccounts
+
+log = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.tier0
```
vashirov, line 26: I'd also add a skipif mark for tests to avoid running it on versions that don't support this control (see https://github.com/389ds/389-ds-base/blob/master/src/lib389/lib389/rootdse.py).

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -0,0 +1,201 @@
...
+        user1 = user.create_test_user(uid=i)
+
+    # Add subentries.
+    user = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    for i in range(nentries, nentries * 2):
+        user1 = user.create_test_user(uid=i)
+        user1.add("objectclass", "ldapsubentry")
+
+    # Test normal entries first.
+    search_filter = "(objectclass=inetorgperson)"
```
vashirov: I believe this test can be parametrized on search_filter and request_ctrl. 
We have some examples in [paged results test](https://github.com/389ds/389-ds-base/blob/master/dirsrvtests/tests/suites/paged_results/paged_results_test.py).

vashirov (commented):
I've reviewed only the python part, see my comments there.
IIRC, large chunk of time at setup/teardown was SELinux labeling/relabeling ports. This should help a bit by prelabeling ports that are used for tests:
```
semanage port -a -t ldap_port_t -p tcp 38900-39299
semanage port -a -t ldap_port_t -p tcp 63600-63999
``` 

Constant setup/teardown provides a huge benefit: tests can be run isolated and in parallel (we used them this way for GH actions). Cleanups can be done via pytest fixtures. You can have topology set up for the test module once and run setup/cleanup fixtures for each test function in the module.

Believe me, this is way better than troubleshooting TET where most of the test suites used one default instance. I still get PTSD flashbacks from that :D

dirsrvtests/tests/suites/subentries/subentries_test.py
antbob: I sent a pull request to python-ldap, see https://github.com/python-ldap/python-ldap/pull/441 so hopefully this wont be necessary.

dirsrvtests/tests/suites/subentries/subentries_test.py
antbob: That should teach me not to copy paste existing copyright notices ;)

dirsrvtests/tests/suites/subentries/subentries_test.py
antbob: done

dirsrvtests/tests/suites/subentries/subentries_test.py
antbob: It sure can be and its gonna be neat and compact but i dont feel its worth a stretch in this unit test. If you feel strongly about it feel free to chip in and redo it however you see fit, kept patch changes flag allowed.

dirsrvtests/tests/suites/subentries/subentries_test.py
vashirov: I don't see the change, was the new commit pushed?

dirsrvtests/tests/suites/subentries/subentries_test.py
vashirov: That's fine for me, thanks.

antbob:
@vashirov Thanx for review! Just ti make sure I do things correctly wrt yr comments. I realized that for pytest.mark.skipif it would need to create a new instance, start it, test skip condition then stop/teardown the instance. Is this really needed ? Are you really gonna run this test on previous versions ? And if so is there a better, read more efficient way, to perform this check eg inside the unit test there is already an instance so can it be checked and skipped from there instead ? I cant seem to find a single example with LDAP controls that does a similar check.

antbob:
@vashirov jfyi, without skip test check the test completes in ~10 seconds, with skip check ~20 seconds so thats the reason i'm asking. btw you have mentioned these tests can run in parallel, how can i take advantage of that ? is there a flag or something to go parallel ? like i have mentioned before it took ~3 hours for my laptop to go thru suites tests which is not fun.

antbob:
@vashirov oh, one more thing, sorry for barrage of questions btw, you dont seem to have any existing LDAP subentry tests, at least no direct ones so maybe it would make sense to split my test into pure subentries part and subentries control part and the skip test would only apply to the later ? (altho it would still run 2x slower).

vashirov:
> Just ti make sure I do things correctly wrt yr comments. I realized that for pytest.mark.skipif it would need to create a new instance, start it, test skip condition then stop/teardown the instance. Is this really needed ?

> oh, one more thing, sorry for barrage of questions btw, you dont seem to have any existing LDAP subentry tests, at least no direct ones so maybe it would make sense to split my test into pure subentries part and subentries control part and the skip test would only apply to the later ? (altho it would still run 2x slower).

This is where parametrized tests would be handy: 2 of the tests that explicitly require subentries control would be skipped ( `pytest.skip` can be called in the test function body), but the rest would be executed. Splitting tests in 2 test functions also should work.

For such small number of cases this (creating an instance just to skip the tests) might be an overkill. Another way is to check for DS version (see usage of `ds_is_older` in the tests), *but* it's a hit or miss depending on the downstream patch level version. And we often see in the downstream testing that tests fail only because the version is slightly different, it takes some time to troubleshoot and verify. So if it only takes 10s overhead to check if the server *definitely* supports this control, I'm gladly taking that.

> btw you have mentioned these tests can run in parallel, how can i take advantage of that ? is there a flag or something to go parallel ? like i have mentioned before it took ~3 hours for my laptop to go thru suites tests which is not fun.

You can enable GH Actions on your fork. We generate matrix of tests based on the filesystem structure:
https://github.com/389ds/389-ds-base/blob/master/.github/workflows/pytest.yml#L21-L23
So every test suite receives its own VM/container to run in parallel. It takes about 40 mins to complete the tests here (17 of which is building rpms). 
 
I used to run tests in containers in parallel on my machine, but it was a CPU/RAM hog, so I delegated that to GH :) 

And no worries, these are good questions, keep them coming :)

vashirov:
@antbob, thank you, really appreciated! If you don't mind, I will modify the test a bit to use fixtures and @pytest.mark.parametrize.

antbob:
@vashirov Thanx and sure, go for it! Like I said I flagged this to allow modifications. Same goes for C code.

vashirov:
@antbob, I've applied changes to the test.

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -58,156 +54,108 @@ def encodeControlValue(self):
...
+            # Search for normal entries
+            (search_entries, False, None, nentries, False),
+            # Search for subentries
+            (search_subentries, False, None, nentries, True),
+            # Search for normal entries and subentries
+            (f'(|{search_entries}{search_subentries})', False, None,
+                nentries * 2, None)
+        ]
+
+@pytest.mark.parametrize('search_filter, use_control, controlValue,'\
```
antbob, line 110: i see what you really meant by parametrizing now. thats much neater altho it would add/delete the same entries for each invokation scenario.

dirsrvtests/tests/suites/subentries/subentries_test.py
```diff
@@ -58,156 +54,108 @@ def encodeControlValue(self):
...
+        request_ctrl = None
+
+    entries = topo.standalone.search_ext_s("ou=people," +
+        DEFAULT_SUFFIX,
+        ldap.SCOPE_SUBTREE,
+        search_filter,
+        serverctrls=request_ctrl)
+
+    assert len(entries) == expected_nentries
+    if expected_subentries:
```
antbob: it need to test that normal entries are not subentries too so the 'not in' check you have removed is actually needed to test this functionality properly.

antbob (commented):
Cool, thanx for doing this! This is much neater but there is a problem with some logic you have removed, see comments.

dirsrvtests/tests/suites/subentries/subentries_test.py
vashirov: The scope of the `setup_test_entries` fixture is "module", so the entries are created at the beginning of the first test that uses it in the module and removed only after the last one. In case of the parametrized test case, fixture is not executed on each new parameter. You can check the access logs to see what I mean.

dirsrvtests/tests/suites/subentries/subentries_test.py
vashirov: You're right, I'll add the missing logic.

dirsrvtests/tests/suites/subentries/subentries_test.py
antbob: ah ok, thats good then and i'm sorry i'm just not familiar with pytest at all and this is literally my 1st time looking into it.

Firstyear:
@mreynolds389 @vashirov Is there anything else that needs to be done here?

vashirov:
@antbob, not sure if you've seen it, but I pushed the change to add missing logic to the test. Let me know if you're happy with it. And if yes, please rebase to to the latest master, squash commits into one and update the commit message. Thanks!

@Firstyear, besides that the only concern was issues with `test_modrdn_entry` test, but looks like it is resolved now. But I'll let @mreynolds389 to say the final word :)

mreynolds389:
Tests are passing, ASAN appears to be passing as well:

==3665003==LeakSanitizer has encountered a fatal error.
==3665003==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==3665003==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)

I am using those LSAN options so this might be okay?

Anyway, Ack.

antbob:
@mreynolds389 that LSAN message is kinda confusing and it most likely means LSAN itself cannot ptrace. it needs ptrace to do the leak check on the process. make sure your selinux stuff allows that. it needs

getsebool deny_ptrace --> off AND sysctl kernel.yama.ptrace_scope --> 0

otherwise it cant ptrace and thinks its under another ptrace or debugger hence the error message.

antbob:
@vashirov sorry, just catching up on this. anyways, i did a rebase and squashed everything into a single commit and hopefully didnt screw it up too bad in the process :) if you need anything else done on my side just gimme a shout.

vashirov:
@antbob, hey, no worries. 
I see that your PR contains additional 17 commits on top of your squashed commit, so something did gone wrong.

vashirov:
Thanks, now the branch looks good! I kicked off the tests. Once they're done, we can merge this.

mreynolds389 (approved):
LGTM

Firstyear:
@mreynolds389 We need to do the merge for them.

antbob:
Thank you everyone for all your help and reviews, really stocked to see this landing!

PR #4986, opened 2021-11-05
Issue 4984 - BUG - pid file handling

### LINKED ISSUE #4984: DS does not error if unable to write pid file
During startup, ns-slapd can be requested to write a pid file. However, if 389 is started as dirsrv directly (such as in a container), when ns-slapd is unable to write the pid to the destination (ie /run/slapd-instance.pid) it silently continues.

In this situation, we should correctly raise that we could not write the pid file.

Firstyear:
Bug Description: When starting up in a container, if the pid
file location is not writeable, the dscontainer/lib389 tools can't
detect that the pid is running correctly, which causes the start
up to fail. This happened because the pid file location was
incorrectly set due to a previous change.

Fix Description: This forces the pid file path when we know
that we are in container mode, and also causes 389-ds to exit
with an error if we can not write to the pid file location.

fixes: https://github.com/389ds/389-ds-base/issues/4984

Review by: ???

### DIFF
--- a/ldap/servers/slapd/main.c
+++ b/ldap/servers/slapd/main.c
@@ -499,15 +499,22 @@ write_start_pid_file(void)
      * admin programs. Please do not make changes here without
      * consulting the start/stop code for the admin code.
      */
-    if ((start_pid_file != NULL) && (fp = fopen(start_pid_file, "w")) != NULL) {
+    if (start_pid_file == NULL) {
+        return 0;
+    } else if ((fp = fopen(start_pid_file, "w")) != NULL) {
         fprintf(fp, "%d\n", getpid());
         fclose(fp);
         if (chmod(start_pid_file, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) != 0) {
+            int rerrno = errno;
+            slapi_log_err(SLAPI_LOG_ERR, "write_start_pid_file", "file (%s) chmod failed (%d)\n", start_pid_file, rerrno);
             unlink(start_pid_file);
+            return -2;
         } else {
             return 0;
         }
     }
+    int rerrno = errno;
+    slapi_log_err(SLAPI_LOG_ERR, "write_start_pid_file", "file (%s) fopen failed (%d)\n", start_pid_file, rerrno);
     return -1;
 }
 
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -387,6 +387,9 @@ def __init__(self, verbose=False, external_log=None):
         self.uuid = str(uuid.uuid4())
         self.verbose = verbose
         self.serverid = None
+        # Are we a container? We get containerised in setup.py during SetupDs, and
+        # in other cases we use the marker to tell us.
+        self._containerised = os.path.exists(DSRC_CONTAINER) or containerised
```
progier389, line 392: IMHO we should also be in that case if getuid() is not 0 (for prefixed install running as plain user)
   ( I hit the issue on Friday (and FYI, I also got in trouble with ldapi sockets that could not be created) 
   
   I wonder if we should not only check that the directory is writable ...

ldap/servers/slapd/main.c
```diff
@@ -499,15 +499,22 @@ write_start_pid_file(void)
      * admin programs. Please do not make changes here without
      * consulting the start/stop code for the admin code.
      */
-    if ((start_pid_file != NULL) && (fp = fopen(start_pid_file, "w")) != NULL) {
+    if (start_pid_file == NULL) {
+        return 0;
+    } else if (fp = fopen(start_pid_file, "w") != NULL) {
```
progier389: line 504 is wrong as the following warning shows. 
Should be ((fp = fopen(...)) != ... )

progier389 (changes requested):
Hi William, there is one mandatory change and one remark

src/lib389/lib389/__init__.py
Firstyear: No - I think this isn't a correct way to determine the containerisation state of the python tools. A user might do "sudo -u dirsrv dsconf ..." for ldapi, and this would confuse that on a local install.

ldap/servers/slapd/main.c
Firstyear: That's an easy fix :)

src/lib389/lib389/__init__.py
progier389: I am not sure that "sudo -u dirsrv dsconf" works nicely anyway (well dsconf may work but dsctl does not (semanage calls will fail)  but if you would rather check for the prefix variable I have not problem with that.

progier389 (approved):
LGTM

src/lib389/lib389/__init__.py
Firstyear: It does work pretty well dsconf at least :) dsctl is a bit of a weirder one.

PR #4988, opened 2021-11-05
Use local_simple_allocate in dsctl so that isLocal is always set properly

spike77453:
This changes the method that's used to allocate an instance in `dsctl`.  
Assuming `dsctl` is always used to manage local instances (hope that's correct) `local_simple_allocate` seems to be the right method to call. It (among other things) ensures that `isLocal` is always set properly (see also #4959)

### DIFF
--- a/src/lib389/cli/dsctl
+++ b/src/lib389/cli/dsctl
@@ -155,7 +155,7 @@ if __name__ == '__main__':
             log.error("Unable to access instance information. Are you running as the correct user? (usually dirsrv or root)")
         sys.exit(1)
 
-    inst.allocate(insts[0])
+    inst.local_simple_allocate(insts[0]['server-id'])
     log.debug('Instance allocated')
 
     try:


Firstyear:
I agree, this is the right thing to do. @droideck @mreynolds389 ?

droideck (approved):
Looks good!

When I try to run 'dsctl' locally with just an instance name, I get this now:

```
DEBUG: Allocate local instance <class 'lib389.DirSrv'> with None
DEBUG: Instance allocated
```

Instead of this:

```
DEBUG: Allocate <class 'lib389.DirSrv'> with None
DEBUG: Allocate <class 'lib389.DirSrv'> with b'www.example.com':636
DEBUG: Allocate <class 'lib389.DirSrv'> with b'www.example.com':636
DEBUG: Instance allocated
```

mreynolds389:
We do need to merge this, but the commit message needs to comply with our guidelines:   

@spike77453 can you please amend the "git commit" message to follow these guidelines?  Then we will merge it, thanks!

mreynolds389:
Thanks @spike77453, merging...

Issue #4992, opened 2021-11-08
slapd-socket incorrect in container.


Firstyear:
It looks like a recent fix has actually fixed our slapd ldapi socket handling in dscontainer, which exposed a mismatch in the dscontainer config. This fixes that behaviour.

Firstyear:
```
To github.com:389ds/389-ds-base.git
   d051255a1..9c2815dc3  389-ds-base-2.0 -> 389-ds-base-2.0
   0ebb39e05..05b0adcb6  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

PR #4993, opened 2021-11-08
4992 fix slapd socket

Firstyear:
Issue 4992 - BUG - slapd.socket container fix

    Bug Description: A recent fix exposed that we were incorrectly
    setting the container socket for ldapi.

    Fix Description: Correct this to be consistent.

    fixes: https://github.com/389ds/389-ds-base/issues/4992

    Review by: ???


Firstyear:
This depends on https://github.com/389ds/389-ds-base/issues/4984 so we should wait for that, then I'll rebase it to be a correct merge.

droideck (approved):
LGTM!

Issue #4734, opened 2021-04-22
ldif2db - import of entry with no parent doesnt generate a warning


jchapma:
**Issue Description**
Online import of ldif file that contains an entry with no parent doesnt generate a task warning.

**Package Version and Platform:**
 - Platform: 5.6.6-300.fc32.x86_64
 - Package and version: 389-Directory/2.0.4

**Steps to Reproduce**
1. Import ldif file with an entry that has no parent entry
2. Notice import success message
3. Check the server error logs to notice an import warning

```
...  Skipping entry "uid=demo,ou=People,dc=example,dc=com" which has no parent, ...
```

**Expected results**
The user should get a warning message saying some entries were skipped and the that they should check the logs for more info.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=1951537
Fix provided by Viktor Ashirov

PR #4996, opened 2021-11-09
Issue 4734 - ldif2db - import of entry with no parent doesnt generate…

bsimonova:
… a warning

Description:
Added a test to check if warning message is displayed
when there are skipped entries during backend import.

Relates: https://github.com/389ds/389-ds-base/issues/4734

Reviewed by: ???


droideck (approved):
Looks good!

PR #5003, opened 2021-11-14
Ensure prop dictionary only contains strings

spike77453:
https://github.com/389ds/389-ds-base/blob/829ea4113fee15f9bf5779734f9d2ddf0281e49b/src/lib389/lib389/__init__.py#L559

Since `self.ldapi_enabled` is a bytes object, `self.ldapi_enabled == 'on'` will be `False` even if `self.ldapi_enabled` is `b'on'`. Bytes objects need to be decoded before comparing them to strings.

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -696,7 +696,7 @@ class DirSrv(SimpleLDAPObject, object):
             ldifconn = LDIFConn(filename)
             configentry = ldifconn.get(DN_CONFIG)
             for key in args_dse_keys:
-                prop[key] = configentry.getValue(args_dse_keys[key])
+                prop[key] = ensure_str(configentry.getValue(args_dse_keys[key]))
                 # SER_HOST            (host) nsslapd-localhost
                 # SER_PORT            (port) nsslapd-port
                 # SER_SECURE_PORT     (sslport) nsslapd-secureport


spike77453:
Thinking about this some more, it might be a good idea to make all values of the `args` dictionary stings, instead of some of them being bytes and some being strings. So maybe changing

https://github.com/389ds/389-ds-base/blob/829ea4113fee15f9bf5779734f9d2ddf0281e49b/src/lib389/lib389/__init__.py#L699

to `prop[key] = configentry.getValue(args_dse_keys[key]).decode()` is the more sane approach?

Firstyear:
We have a wrapper for this that's python2/3 compaible for lol reasons. I think it's ensure_str or ensure_bytes which you can use to wrap the getValue.

spike77453:
> We have a wrapper for this that's python2/3 compaible for lol reasons. I think it's ensure_str or ensure_bytes which you can use to wrap the getValue.

I've updated this PR accordingly.

progier389 (approved):
LGTM

spike77453:
I there anything I can do or still need to do to get this merged?

progier389:
Since you got the approval,  I think that you only have  to hit the "Squash and merge" button.

Firstyear:
@progier389 They won't have commit rights to squash/merge themself, one of us need to do it.

progier389:
Thank you @FirstYear for the information, so I merged it ...

Issue #5006, opened 2021-11-16
UI - LDAP editor tree not being properly updated


mreynolds389:
**Issue Description**

The LDAP editor, while still under construction, has some odd behavior when updating the tree view verses the table view.  Tree updates were not applied to the treeview, and changes to the table view were not reflected in the treeview.

mreynolds389:
a69bd6111..f974ec393  389-ds-base-2.0 -> 389-ds-base-2.0
ea5c61a88..a55c381f2  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

PR #5007, opened 2021-11-16
Issue 5006 - UI - LDAP editor tree not being properly updated

mreynolds389:
Description:  

Deleting an entry was the tree view did not update the treeview.  Updates to table view were also not seen in the tree view.  The views should now be in synch

Also, replaced some console logging with our "log_cmd" function in the editor utils file.

relates: https://github.com/389ds/389-ds-base/issues/5006

### DIFF
--- a/src/cockpit/389-console/src/lib/ldap_editor/treeView.jsx
+++ b/src/cockpit/389-console/src/lib/ldap_editor/treeView.jsx
@@ -113,6 +113,10 @@ class EditorTreeView extends React.Component {
 
         this.handleNodeOnClick = (treeViewItem) => {
             console.log('handleNodeOnClick - treeViewItem: ', treeViewItem);
+            if (treeViewItem && treeViewItem.dn === this.state.entryDn) {
+                // Clicking on already selected node, just return
+                return;
+            }
             this.setState({
                 searching: true,
             }, () => {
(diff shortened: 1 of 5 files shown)


src/cockpit/389-console/src/lib/ldap_editor/treeView.jsx
```diff
@@ -113,6 +113,10 @@ class EditorTreeView extends React.Component {
 
         this.handleNodeOnClick = (treeViewItem) => {
             console.log('handleNodeOnClick - treeViewItem: ', treeViewItem);
+            if (treeViewItem.dn === this.state.entryDn) {
```
droideck: Weirdly enough, a browser crashes here when I:

1. Go to the Tree view and select a user under ou=People;
2. Go to the Table view and select the same user there;
3. Go back to the Tree view and open the same user view (may be an unnecessary step)
4. Go back to the Table view and delete the user.

Result - Browser crashed.

droideck (changes requested):

src/cockpit/389-console/src/lib/ldap_editor/treeView.jsx
mreynolds389: I can not reproduce the crash

mreynolds389: fixed

droideck (approved):
Looks good! Thanks!

Issue #5008, opened 2021-11-17
If a non critical plugin can not be loaded/initialized, bootstrap should succeeds


tbordaz:
**Issue Description**
During startup, the server parses the config (dse.ldif),  loads plugins and retrieves their init function. If the server fails to load a plugin/init fct, the startup fails.

If a non-critical plugin can not be loaded/initialized, the server should continue its startup.

This issue can occur if for example, dse.ldif contains legacy plugin entry which lib/init fct is no longer delivered

**Package Version and Platform:**
All

**Steps to Reproduce**
Need to craft dse.ldif so that a plugin entry refer to a not delivered init fct, for example
<pre>
dn: cn=GOST_YESCRYPT,cn=Password Storage Schemes,cn=plugins,cn=config
objectClass: top
objectClass: nsSlapdPlugin
cn: GOST_YESCRYPT
nsslapd-pluginPath: libpwdstorage-plugin
nsslapd-pluginInitfunc: missing_function_for_init
nsslapd-pluginType: pwdstoragescheme
nsslapd-pluginEnabled: on
nsslapd-pluginId: none
nsslapd-pluginVersion: none
nsslapd-pluginVendor: none
nsslapd-pluginDescription: none

</pre>

**Expected results**
Failure to init a non-critical plugin does not prevent DS to start

mreynolds389:
Please cherry-pick down to 1.4.3

tbordaz:
6100985dd..792197708  master
237913e86..cb9980a06  389-ds-base-2.0
b633f727b..083fd6a84  389-ds-base-1.4.4
c600ddcab..ff76d6c63  389-ds-base-1.4.3

aivanov389:
With standard dse.ldif i still have errors (even if the server starts):

```
[01/Dec/2021:12:54:04.459308888 +0100] - ERR - symload_report_error - Netscape Portable Runtime error -5977: /Local/dirsrv/lib/dirsrv/plugins/libpwdchan-plugin.so: cannot open shared object file: No such file or directory
[01/Dec/2021:12:54:04.460194603 +0100] - ERR - symload_report_error - Could not open library "/Local/dirsrv/lib/dirsrv/plugins/libpwdchan-plugin.so" for plugin PBKDF2
[01/Dec/2021:12:54:04.460833312 +0100] - ERR - plugin_setup - "PBKDF2" plugin in library "libpwdchan-plugin" not initialized and ignored
[...]

PR #5009, opened 2021-11-17
Issue 5008 - If a non critical plugin can not be loaded/initialized, …

tbordaz:
…bootstrap should succeeds

Bug description:
	If a non-critical plugin can not be loaded/initialized, the server should continue its startup.

Fix description:
	During plugin_setup, if the server fails to initialize a
	non-critical plugin then it just log an error:

	plugin_setup - "GOST_YESCRYPT" plugin in library "libpwdstorage-plugin" not initialized and ignored

	The non-critical plugins are statically listed in
        plugin_critical(). ATM, only GOST_YESCRYPT is seen as non-critical

relates: #5008

Reviewed by:

Platforms tested: F34, CentOS8

### DIFF
--- a/ldap/servers/slapd/plugin.c
+++ b/ldap/servers/slapd/plugin.c
@@ -2680,6 +2680,36 @@ plugin_free(struct slapdplugin *plugin)
     slapi_ch_free((void **)&plugin);
 }
 
+/*
+ * Check if a plugin is critical or not.
+ * The plugins not critical are statically listed either
+ * with their name or with the plugin path
+ *  - GOST_YESCRYPT (name)
+ *  - entryuuid (name)
+ *  - libpwdchan (rust
+ * Returns:
+ *   1 - critical
+ *   0 - non critical
+ */
+static int
+plugin_load_critical(struct slapdplugin *plugin)
+{
+    char *non_critical_plugin_name[] = {"GOST_YESCRYPT", "entryuuid", NULL };
+    char *non_critical_plugin_libpath[] = {"libpwdchan-plugin", NULL};
+    for (int32_t i = 0; non_critical_plugin_name[i]; i++) {
+        if (strcasecmp(plugin->plg_name, non_critical_plugin_name[i]) == 0) {
+            /* this plugin in *not* critical */
+            return 0;
+        }
+    }
+    for (int32_t i = 0; non_critical_plugin_libpath[i]; i++) {
+        if (strcasecmp(plugin->plg_libpath, non_critical_plugin_libpath[i]) == 0) {
+            /* this plugin in *not* critical */
+            return 0;
+        }
+    }
+    return 1;
+}
 /***********************************
 This is the main entry point for plugin configuration.  The plugin_entry argument
 should already contain the necessary fields required to initialize the plugin and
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/plugin.c
```diff
@@ -2680,6 +2680,24 @@ plugin_free(struct slapdplugin *plugin)
...
+ * Check if a plugin is critical or not
+ * Returns:
+ *   1 - critical
+ *   0 - non critical
+ */
+static int
+plugin_critical(struct slapdplugin *plugin)
+{
+    char *non_critical_plugins[] = {"GOST_YESCRYPT", NULL };
+    for (int32_t i = 0; non_critical_plugins[i]; i++) {
```
mreynolds389: I feel this list could also include the RUST plugins (entryuuid, pbdfk2 256 (libpwdchan-plugin), etc).  Not everyone builds with Rust and I think people have complained about it recently.

tbordaz: Thanks @mreynolds389  for your review !!
@progier389 also reported this difficulties with RUST plugin. I agree with entryuuid. With libpwdchan I have a doubt because pbkdf2 is our default storage schema, if it does not start most simple authentication will fail so it looks critical to me.

Firstyear: The pkbdf2 in pwdchan is openldap compat pbkdf2 not the same as our inbuilt C one.

There will come a point where we have to draw a line and say "rust is required though", and I think that's a seperate issue. 

But for now I'd say pwdchan can be "optional" if you want. It's just a hard requirement for us at suse with openldap to 389 migrations though. And it's up to RH if you want to support openldap password hash imports when you start to migrate people over.

progier389: To complete what William is saying. That is the PBKDF2_SHA256 implemented by libpwdstorage-plugin which is mandatory. So I think that we can safely mark pwdchan as "optional" and do not prevent the server to start if the plugin cannot be loaded. (That said if there are some #ifdef specific to SUZE we should use conditional compilation to insure the plugin is mandatory in suze)

progier389: BTW for the plugin relying on Rust I would rather check the plg_libpath rather than the name

ldap/servers/slapd/plugin.c
```diff
@@ -2893,12 +2911,22 @@ plugin_setup(Slapi_Entry *plugin_entry, struct slapi_componentid *group, slapi_p
...
-                goto PLUGIN_CLEANUP;
+                                                                         loadNow, loadGlobal);
+            if (initfunc == NULL) {
+                /* failure to load that plugin or find its initfunct, let's fail only if it is critical */
+                if (plugin_critical(plugin)) {
+                    PR_snprintf(returntext, SLAPI_DSE_RETURNTEXT_SIZE, "Failed to load plugin's init function.");
+                    status = -1;
+                    goto PLUGIN_CLEANUP;
+                } else {
+                    /* This plugin is not critical, just do not ignore it and continue */
```
progier389: Comment should be "just ignore it and continue" (i.e without the "do not")

stanislavlevin:
Hi,
there are 2 cases I see:
- plugins disabled by downstream for any reason
- no longer shipped plugins by upstream

The GOST_YESCRYPT plugin in this PR is disabled in downstream only.
What are the formal criteria for the critical or uncritical plugin then?

tbordaz:
> Hi, there are 2 cases I see:
> 
>     * plugins disabled by downstream for any reason
> 
>     * no longer shipped plugins by upstream
> 
> 
> The GOST_YESCRYPT plugin in this PR is disabled in downstream only. What are the formal criteria for the critical or uncritical plugin then?

Before starting a plugin, the plugin is initialized. The startup mechanism is controlled by the value of attribute 'nsslapd-pluginEnabled' and the PR does not change it.
The PR only impacts initialization and more specifically the retrieval/upload of the initialization callback 'nsslapd-pluginInitfunc'. When 'nsslapd-pluginInitfunc' callback can not be loaded, because library is no longer delivered or initfunc is removed/renamed, DS startup fails.
The PR allows successful DS startup, even if non critical plugin initialization fct is not retrieved.

The list of non critical plugin is just a static list (build time). A better option would be that 'nsSlapdPlugin' objectclass would support a new attributetype like 'nsslapd-pluginCritical: true/false'.

progier389 (approved):
LGFM

Issue #5014, opened 2021-11-19
UI - Add group creation to the LDAP editor


mreynolds389:
The LDAP editor is lacking an option to create "groups"

mreynolds389:
f974ec393..18a127491  389-ds-base-2.0 -> 389-ds-base-2.0
a55c381f2..151ac7efb  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

PR #5015, opened 2021-11-19
Issue 5014 - UI - Add group creation to LDAP editor

mreynolds389:
Description:  

Added group creation to LDAP editor via the "New ..." menu option.

Did some Form cleanup with autoComplete="off"

relates: https://github.com/389ds/389-ds-base/issues/5014


droideck (approved):
LGTM!

The only thing - it's not possible to Edit or Rename entries. But it's true for users too, so I assume that it's a 'TODO' feature. If so - you have my ack!

image

mreynolds389:
Yeah editing and renaming will come later.  Basically I'm trying to get all the "Add entry" functions working.  Then Edit for each type of entry, and finally modrdn.

Issue #5001, opened 2021-11-12
Fix next round of UI bugs


mreynolds389:
**Issue Description**

Addressing a series of bugs found by QE:

Bug 2016526 - LDAPI & Autobind save btn misbehaving
Bug 2016481 - Disabling Security leaves the pop-up window open
Bug 2016026 - Selecting existing certificate in Security Configuration crashes browser
Bug 2017402 - Adding several allowed SASL mechanisms does not behave correctly
Bug 2017411 - cockpit crashes because invalid SASL mapping regex was saved
Bug 2022117 - Cockpit UI: Editing an Objectclass name causes an error in Cockpit UI
Bug 2021194 - Searching "matching rules" in the "Schema" Tab crashes browser
Bug 2021591 - cockpit : audit and audit failure log enablement status is not persistent

mreynolds389:
f53793d37..8a2b4c7d5  389-ds-base-2.0 -> 389-ds-base-2.0
78b5bc996..9658e9367  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

mreynolds389:
18a127491..a033e0269  389-ds-base-2.0 -> 389-ds-base-2.0
151ac7efb..b633f727b  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

droideck:
5ba2b8b21..706d95fa1  389-ds-base-2.0 -> 389-ds-base-2.0
   8bff82793..adc2478d0  389-ds-base-1.4.4 -> 389-ds-base-1.4.4

PR #5016, opened 2021-11-20
Issue 4962 - Fix various UI bugs - Settings and Monitor

droideck:
Description:

Bug 2014924 - Cockpit UI: UX Bugs and other cockpit GUI related defects
Bug 2017441 - cockpit : Export changelog allows to check both 'Export to LDIF For Debugging' options but only takes one into account
Bug 2018101 - cockpit : impossible to create credentials or aliases for replication monitoring synchronization report
Bug 2021250 - cockpit : logging setting entered values for rotation and deletion policies should be checked
Bug 2021278 - Cockpit UI: Unable to Edit Attributes without first searching for the attribute to edit

Related: https://github.com/389ds/389-ds-base/issues/5001

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/dsModals.jsx
+++ b/src/cockpit/389-console/src/dsModals.jsx
@@ -435,7 +435,7 @@ export class CreateInstanceModal extends React.Component {
             >
                 <div className={loadingCreate ? "ds-disabled" : ""}>
                     <Form isHorizontal autoComplete="off">
-                        <Grid className="ds-margin-top" title="The instance name, this is what gets appended to 'slapi-'. The instance name can only contain letters, numbers, and:  # @ : - _">
+                        <Grid className="ds-margin-top" title="The instance name, this is what gets appended to 'slapi-'. The instance name can only contain letters, numbers, and: # @ : - _">
                             <GridItem className="ds-label" span={4}>
                                 Instance Name
                             </GridItem>
@@ -456,7 +456,7 @@ export class CreateInstanceModal extends React.Component {
                                 </FormHelperText>
                             </GridItem>
                         </Grid>
-                        <Grid title="The server port number">
+                        <Grid title="The server port number should be in the range of 1 to 65534.">
                             <GridItem className="ds-label" span={4}>
                                 Port
                             </GridItem>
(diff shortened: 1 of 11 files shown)


src/cockpit/389-console/src/dsModals.jsx
```diff
@@ -569,12 +569,12 @@ export class CreateInstanceModal extends React.Component {
...
                                 />
                                 <FormHelperText isError isHidden={!errObj.createDMPasswordConfirm}>
-                                    Confirmation password must be set and it must match the first password
+                                    Confirmation password must be set and it must match the first password.
                                 </FormHelperText>
                             </GridItem>
                         </Grid>
                         <hr />
-                        <Grid title="Create the Database">
+                        <Grid title="Create a database during the instalation.">
```
mreynolds389: typo, should be "installation"

mreynolds389:
You changed all the "Create New Log Every ..." and "Log File is Older Than ..." settings appear to be a mix of a dropdown and NumberInputs

Screenshot from 2021-11-20 10-20-27

mreynolds389:
In replication monitor, "Add Replication Credntials", the port number is not set, it should have a default value.  And when I increment the number I get a console warning:

checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `port` of type `number` supplied to `ReportCredentialsModal`, expected `string`.
    in ReportCredentialsModal (created by ReplMonitor)

mreynolds389:
Also, not sure if it was always broken, but if the root DN passwords do not match the TextInput is not set to correct validated state.  It never turns red when the passwords don't match

droideck:
> You changed all the "Create New Log Every ..." and "Log File is Older Than ..." settings appear to be a mix of a dropdown and NumberInputs
> 
> Screenshot from 2021-11-20 10-20-27

This part is really weird. It looks okay in full-width mode, but when you shrink the screen - it merges like on your picture...
I may need to rethink the design here as it seems that PF4 fails to Grid properly such a chain of elements (label + numberinput + select on the same line).

mreynolds389:
> This part is really weird. It looks okay in full-width mode, but when you shrink the screen - it merges like on your picture...
> I may need to rethink the design here as it seems that PF4 fails to Grid properly such a chain of elements (label + numberinput + select on the same line).

I think the issue is that the GridItem span's do not add up to 12.  If you reduce the size in this case, thinks do get goofy.  Maybe try setting the very last gridItem's span so that it adds up to 12 total for the Grid.  That might not be enough in all cases and screen sizes, in that case you probably have to equally spread out the spans (it still needs to total 12).

mreynolds389:
Ack!

PR #5023, opened 2021-11-23
Revert slapi-plugin.h change from PR 4924

### LINKED ISSUE #4923: issue about LMDB dbi versus txn handling.

progier389:
Fix slapi-plugin.h change because although it does not break compatibility with old plugin, 
  it cause trouble to some tester tools that mimic the roles of ns-slapd server (and refine slapi_log_error and slapi_log_error_ext)
Using by default the old interface and using the new interface only when including slapd.h (a private include).
(that avoid being plagued by warning about const getting removed from string literal)
BTW: In term of binaries both interfaces are equivalent (i.e compatible when compilation does not fail) so we can use safely plugin using old or new interface with ns-slapd using the other one.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -59,7 +59,6 @@ typedef unsigned char u_int8_t;
 typedef unsigned int u_int32_t;
 typedef unsigned short u_int16_t;
 #endif
-#include "dbimpl.h"
 
 #define dptr  data
 #define dsize size
@@ -76,6 +75,7 @@ typedef unsigned short u_int16_t;
 #include "slap.h"
 #include "slapi-plugin.h"
 #include "slapi-private.h"
+#include "dbimpl.h"
 #include "avl.h"
 #include "portable.h"
 #include "proto-slap.h"
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -2483,7 +2483,7 @@ vslapd_log_error(
  * severity - LOG_ERR, LOG_WARNING, LOG_INFO, etc
  */
 int
-slapi_log_error(int loglevel, char *subsystem, const char *fmt, ...)
+slapi_log_error(int loglevel, const char *subsystem, const char *fmt, ...)
 {
     va_list ap_err;
     va_list ap_file;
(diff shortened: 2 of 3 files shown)


tbordaz (approved):
Thanks for the fix. It looks good (and clever) to me. ACK if PR-CI are successful

progier389:
PR-CI are OK but the compile still show lots of warning  like
 Warning: ldap/servers/slapd/back-ldbm/ancestorid.c:43:63: warning: passing argument 3 of 'slapi_log_error' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]

The reason is that dbimpl.h is including slapi-plugin without including slapd.h ==> I will fix that

progier389:
Ok this time most of the warning disappeared (there is still a few ones but not in the backend and I feel we will have to do a global warning cleanup in another issue (some other functions are impacted)
So I am merging now.

PR #5026, opened 2021-11-24
Issue 5025 - RFE - remove useless logging

### LINKED ISSUE #5025: Remove useless, noisy, perfomance degrading debug logs.
Sometimes when we attempt to use gdb or ask customers to debug an issue, we ask for a higher log level, sometimes -d 0 or similar. This causes thousands of useless messages on console, that cause huge amounts of delay and latency, changing server startup and shutdown from seconds, to minutes. In one case a customer saw startup take 45 minutes due to this excess, pointless logging.

Many of these trace log records serve no value, they don't help with debugging, they create noise, performance issues, and just shouldn't exist.

This removes some of the worst offenders in the code base.

Firstyear:
Bug Description: Many of these trace log records serve no value, they don't help with debugging,
they create noise, performance issues, and just shouldn't exist.

Fix Description: Delete the offending lines.

fixes: https://github.com/389ds/389-ds-base/issues/5025

Review by: ???

### DIFF
--- a/ldap/servers/slapd/attr.c
+++ b/ldap/servers/slapd/attr.c
@@ -958,9 +958,6 @@ attr_get_value_cmp_fn(const Slapi_Attr *attr, value_compare_fn_type *compare_fn)
 {
     int rc = LDAP_PROTOCOL_ERROR;
 
-    slapi_log_err(SLAPI_LOG_TRACE, "attr_get_value_cmp_fn",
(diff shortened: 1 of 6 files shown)


tbordaz (approved):
Good idea to cleanup those logs.
I had an initial doubt about slapi_vattr_filter_test_ext_internal cleanup but others logs help much more (like get_filter_internal) to follow filter parsing.

PR #5027, opened 2021-11-24
Issue 5024 - BUG - windows ro replica sigsegv

### LINKED ISSUE #5024: windows sync can cause SIGSEGV when replica is configured as hub or consumer.
After 1.4.3, the changelog moves into the main database rather than being a seperate entity. This caused a situation where it was possible to create and configure a changelog while also acting as a read-only replica. This allows individuals to create a windows sync agreement, however during the upgrade to 1.4.4 where we move the changelog into the main DB, I can only assume we either delete or ignore the CL if the replica is readonly.

As a result, this caused a situation where the replica information and agreement existed, but the changelog did not. During startup as the agreement was processed, an attempt to open the CL was made, but caused a NULL pointer dereference which prevented server startup. 

This is currently blocking and affecting a SUSE customer so I would appreciate a quick review so we can send a hotfix.

Firstyear:
Bug Description: After 1.4.3, the changelog moves into the main database rather than being a
seperate entity. This caused a situation where it was possible to create and configure a
changelog while also acting as a read-only replica. This allows individuals to create a
windows sync agreement, however during the upgrade to 1.4.4 where we move the changelog into
the main DB, I can only assume we either delete or ignore the CL if the replica is readonly.

As a result, this caused a situation where the replica information and agreement existed, but
the changelog did not. During startup as the agreement was processed, an attempt to open the
CL was made, but caused a NULL pointer dereference which prevented server startup.

Fix Description: This fixes the issue on a number of fronts. First, we remove the original
NULL pointer dereference in cl5_api.c. We correct a bug in promote in lib389 that prevented
consumer to supplier promotion. And finally, this adds a hardening check to better communicate
to users in this situation what steps they need to take.

fixes: https://github.com/389ds/389-ds-base/issues/5024

Review by: ???

### DIFF
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -897,7 +897,7 @@ cl5CreateReplayIteratorEx(Private_Repl_Protocol *prp, const RUV *consumerRuv, CL
 
     replica = prp->replica;
     if (replica == NULL || consumerRuv == NULL || iterator == NULL) {
-        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl,
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
                       "cl5CreateReplayIteratorEx - Invalid parameter\n");
         return CL5_BAD_DATA;
     }
@@ -905,10 +905,20 @@ cl5CreateReplayIteratorEx(Private_Repl_Protocol *prp, const RUV *consumerRuv, CL
     *iterator = NULL;
 
     cldb = replica_get_cl_info(replica);
+    if (cldb == NULL) {
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
+                      "cl5CreateReplayIteratorEx - Changelog is not available (NULL) for %s\n",
+                      replica_get_name(replica)
+        );
+        return CL5_BAD_STATE;
+    }
+
     pthread_mutex_lock(&(cldb->stLock));
     if (cldb->dbState != CL5_STATE_OPEN) {
-        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl,
-                      "cl5CreateReplayIteratorEx - Changelog is not initialized\n");
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
+                      "cl5CreateReplayIteratorEx - Changelog is not available for %s (dbState: %d)\n",
+                      replica_get_name(replica),
+                      cldb->dbState);
         pthread_mutex_unlock(&(cldb->stLock));
         return CL5_BAD_STATE;
     }
@@ -946,7 +956,7 @@ cl5CreateReplayIterator(Private_Repl_Protocol *prp, const RUV *consumerRuv, CL5R
 
     replica = prp->replica;
     if (replica == NULL || consumerRuv == NULL || iterator == NULL) {
-        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl,
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
                       "cl5CreateReplayIterator - Invalid parameter\n");
         return CL5_BAD_DATA;
     }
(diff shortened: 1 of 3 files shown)


ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -897,18 +897,28 @@ cl5CreateReplayIteratorEx(Private_Repl_Protocol *prp, const RUV *consumerRuv, CL
...
-        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name_cl,
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
                       "cl5CreateReplayIteratorEx - Invalid parameter\n");
         return CL5_BAD_DATA;
     }
 
     *iterator = NULL;
 
     cldb = replica_get_cl_info(replica);
+    if (cldb == NULL) {
```
tbordaz, line 908: Good catch, it is another flavor of https://github.com/389ds/389-ds-base/issues/4731

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -482,8 +482,20 @@ agmt_new_from_entry(Slapi_Entry *e)
 
     /* DBDB: review this code */
     if (slapi_entry_attr_hasvalue(e, "objectclass", "nsDSWindowsReplicationAgreement")) {
-        ra->agreement_type = REPLICA_TYPE_WINDOWS;
-        windows_init_agreement_from_entry(ra, e);
+        if (replica_get_type(replica) == REPLICA_TYPE_PRIMARY) {
```
tbordaz, line 485: type PRIMARY (1) or UPDATABLE (3) should have a CL and we can have a RA.
Shouldn't we fail only for READONLY replica, else create the RA ?

mreynolds389:
I have a minor concern here.  Many log levels were changed to log by default, so now I'm a bit worried about the noise this might make in the log.  So if the changelog is not present, its going to give you like 10 error messages basically saying the same thing.  Just wondering if there is a common path for these failures and if we can reduce it to the fewest amount of SLAPI_LOG_ERR messages as possible.  If its really needed that's fine, but if it can be reduced then that would be great.

ldap/servers/plugins/replication/cl5_api.c
Firstyear: I didn't catch it. A customer is in production.

ldap/servers/plugins/replication/repl5_agmt.c
Firstyear: But it's then valid to have a topology which is:

```
PRIMARY -> HUB <- WINSYNC
```

Which means that HUB no longer is consistent, nor is the topology. All changes from winsync must go to a full supplier.

Firstyear:
> I have a minor concern here. Many log levels were changed to log by default, so now I'm a bit worried about the noise this might make in the log. So if the changelog is not present, its going to give you like 10 error messages basically saying the same thing. Just wondering if there is a common path for these failures and if we can reduce it to the fewest amount of SLAPI_LOG_ERR messages as possible. If its really needed that's fine, but if it can be reduced then that would be great.

These errors only appear during the reload of a config for that backend though. And if the changelog isn't present, we have a bigger issue ... this is a genuine error in the config / system state, and it shouldn't be hidden behind a repl log flag. In the current situation, we'd end up with silent failure instead.

ldap/servers/plugins/replication/repl5_agmt.c
tbordaz: replicaType (lib389) comes from 'nsDS5ReplicaType'. It looks lib389 only supports REPLICA_TYPE_UPDATABLE and REPLICA_TYPE_READONLY. I do not know where the value REPLICA_TYPE_PRIMARY is coming from (during config).

Anyway we want to create/init RA if there is a CL. It exists CL for REPLICA_TYPE_UPDATABLE and possibly also for REPLICA_TYPE_PRIMARY. 
So I wonder if the test should rather be 
<pre>
if (replica_get_type(replica) == REPLICA_TYPE_PRIMARY || replica_get_type(replica) == REPLICA_TYPE_UPDATABLE) {
...
}
</pre>

Firstyear: I tested this. Primary is when it is a supplier. When we set this as a hub it's NOT a primary, and IS uptable. It works. 

As mentioned, we *can not* match on updateable here because that's an invalid replication topology.

tbordaz: Sorry @Firstyear for my slow understanding. I missed that HUB case is an invalid replication topology. So this change is good to me.

Side note, IIUC this part of code is to handle outgoing RA to windows. It is not clear to me why hub can not act as a supplier.

Side note, during upgrade (_cl5_upgrade_replica), I do not find any specificity supplier vs hub. It looks to me that in both cases the changelog file and config is move to the main DB.

tbordaz (approved):

ldap/servers/plugins/replication/repl5_agmt.c
Firstyear: It handles incoming replication too @tbordaz .... it does both directions. 

The upgrade issue is for consumer vs supplier, not hub.

mreynolds389:
LGTM if @tbordaz is happy.

Firstyear:
ping @tbordaz :)

tbordaz:
> ping @tbordaz :)

Sure I am very happy with the fix and the detailed explanations. I already approved the patch. ACK

Firstyear:
Thank you @tbordaz ! @mreynolds389 I'll do the same and backport these through 1.4.3

Firstyear:
```
   2a0edec19..43ac925f9  389-ds-base-2.0 -> 389-ds-base-2.0
   b65515248..4290afd68  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
   76f56b28a..e6068c07a  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
```

PR #5028, opened 2021-11-24
5011 - test_replica_backup_and_restore random failure

### LINKED ISSUE #5011: test_replica_backup_and_restore random failure
**Issue Description**
CI test dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py::test_replica_backup_and_restore sometime fails during nighty tests::

```
        # Check that the updates (DEL) are no longer there
        for i in users.list():
            testuser = UserAccount(topo_m4.ms["supplier1"], i.dn)
>           assert testuser.exists()
E           assert False
E            +  where False = <bound method DSLdapObject.exists of <lib389.idm.user.UserAccount object at 0x7f6612653a60>>()
E            +    where <bound method DSLdapObject.exists of <lib389.idm.user.UserAccount object at 0x7f6612653a60>> = <lib389.idm.user.UserAccount object at 0x7f6612653a60>.exists

```
IMHO the issue is a simple timing issue in the test case:
  we should disable the agmt toward supplier1 before starting it 
  and re-enable them after checking that user exists

progier389:
fourway_mmr CI test is sometime failing I suspect it is because the replication may replay the DEL before we test the entries
Solution: disable all agreements toward the tested supplier before restarting it, then reenable them after the entry test.

### DIFF
--- a/dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py
+++ b/dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py
@@ -430,6 +430,7 @@ def test_bob_acceptance_tests(topo_m4):
 
 
 def list_agmt_towards(topo_m4, serverid):
+    # Note: all instances must be started to use that.
     res = []
     for inst in topo_m4:
         for agmt in Agreements(inst).list():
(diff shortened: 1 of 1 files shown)


tbordaz (approved):
ACK

droideck (approved):
LGTM too!

PR #5033, opened 2021-11-26
Issue 5032 - OpenLDAP is not shipped with non-threaded version of lib…

### LINKED ISSUE #5032: OpenLDAP is not shipped with non-threaded version of libldap
**Issue Description**
389-ds-base is built with libldap_r shared library. The non-threaded version of libldap is fully depricated since Fedora 34, libldap_r supplied as symlinks. OpenLDAP 2.6 won't have libldap_r at all. Hence, we should rewrite our Makefile.

droideck:
…ldap

Bug Description: 389-ds-base is built with libldap_r shared library.
The non-threaded version of libldap is fully depricated since Fedora 34,
libldap_r supplied as symlinks. OpenLDAP 2.6 won't have libldap_r at all.

Fix Description: Use lldap instead of lldap_r in Makefile.

Reviewed by: ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -868,6 +868,24 @@ AC_SUBST(PACKAGE_BASE_VERSION)
 
 AM_CONDITIONAL(OPENLDAP,test "$with_openldap" = "yes")
 
+# check for --with-libldap-r
+AC_MSG_CHECKING(for --with-libldap-r)
+AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use libldap_r.so shared library (default: if OpenLDAP version is less than 2.6, then libldap_r.so will be used, else - libldap.so)]),
+[
+  if test "$withval" = "no"; then
+    AC_MSG_RESULT(no)
+  else
+    with_libldap_r=yes
+    AC_MSG_RESULT(yes)
+    AC_SUBST(with_libldap_r)
+  fi
+],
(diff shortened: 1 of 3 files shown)


Firstyear (changes requested):
We need to check it the configure which version of openldap it is, else we are linking to the non-threaded version in 2.4.

droideck:
Good catch!
It is okay for Fedora - as Fedora 34 has already gotten rid of the non-threaded version.

But generally, we need a way to configure it.
Hence, the update :)

configure.ac
```diff
@@ -868,6 +868,24 @@ AC_SUBST(PACKAGE_BASE_VERSION)
...
+    AC_MSG_RESULT(no)
+  else
+    with_libldap_r=yes
+    AC_MSG_RESULT(yes)
+    AC_SUBST(with_libldap_r)
+  fi
+],
+OPENLDAP_VERSION=`ldapsearch -VV 2> >(sed -n '/ldapsearch/ s/.*ldapsearch \([0-9]\+\.[0-9]\+\.[0-9]\+\) .*/\1/p')`
+AC_MSG_RESULT([$OPENLDAP_VERSION])
+AX_COMPARE_VERSION([$OPENLDAP_VERSION], [lt], [2.6], [ with_libldap_r=no ], [ with_libldap_r=yes ]))
```
Firstyear, line 885: How does this interact with the ARG_WITH above it? Does this kick in and set the value to yes if not set?

droideck: Yep, it's a part of ARG_WITH call.
The code is executed if "--with-libldap-r" is not set - in that case, we check the OpenLDAP version on the build system and use the appropriate shared library.

Firstyear (approved):
Great, thank you!

PR #5041, opened 2021-11-30
Issue 4758 - Add tests for WebUI

bsimonova:
Description:
Just added additional iframe assignment checking, because
some tests ocassionaly failed at instance removal.

Relates: https://github.com/389ds/389-ds-base/issues/4758

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/webui/__init__.py
+++ b/dirsrvtests/tests/suites/webui/__init__.py
@@ -19,12 +19,19 @@ pytest.importorskip('playwright')
 RHEL = 'Red Hat Enterprise Linux'
 
 
-# in cockpit-250 some selectors got renamed so this function returns True if cockpit version is 250 or higher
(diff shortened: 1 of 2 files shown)


bsimonova:
I see it failing again, will have to investigate further on Fedora 35.
I checked it in headed mode and for some reason, the page is loaded with visible and enabled #no-inst-create-btn after the instance removal but the selector id is not picked even though the script is continuously checking for it and then it timeouts.

I tried to add page.reload() during the checking but it still won't pick it up in the browser instance. Will check another possible solution with browser instances.

vashirov (approved):

tbordaz:
FYI @bsimonova congratulations this very common failure has not happened since you pushed the patch. Nice !

bsimonova:
@tbordaz 
Thanks, I appreciate it.
I was monitoring the results during the week to check it, so hopefully it will stay that way.

PR #5054, opened 2021-12-10
Issue 5053 - Improve GitHub Actions debugging

### LINKED ISSUE #5053: Improve GitHub Actions debugging
**Issue Description**
Sometimes tests fail only in PR CI, but it's hard to troubleshoot them without remote access. There is no built-in implementation, but this can be solved by using "[Debugging with tmate](https://github.com/marketplace/actions/debugging-with-tmate)" action.

Also, currently is not possible to rerun just one test suite or test module. GitHub Actions support [`workflow_dispatch`](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) that can be parametrized (up to 10 inputs), so we can take advantage of that.

vashirov:
Description:
This commit adds `workflow_dispatch` trigger with 2 inputs.
First input specifies what test suites or test modules should be
executed. Paths should be delimited by space and be relative to
dirsrvtests/tests/suites/, i.e. input
```
basic replication/acceptance_test.py
```
will execute tests under `basic` test suite directory and
`replication/acceptance_test.py` module.
Only existing paths are accepted, everything else will be filtered out.

Second input is a boolean that enables debugging with tmate:
https://github.com/marketplace/actions/debugging-with-tmate
By default it's `false`, so set it to `true` when you need to rerun the
test with remote access.

Action is configured to use SSH key(s) registered with your GitHub
profile.

Connection string will be logged in "Setup tmate session" step in the
test logs.

Once you inside the tmate session, run `touch continue` to continue the
workflow. When ontainer with pytest is started, run
`docker exec -ti $(docker ps -q) /bin/bash` to get inside the container
to inspect it.

Fixes: https://github.com/389ds/389-ds-base/issues/5053

Reviewed by: ???

### DIFF
--- a/.github/scripts/generate_matrix.py
+++ b/.github/scripts/generate_matrix.py
@@ -1,19 +1,37 @@
 import os
+import sys
 import glob
 import json
 
-suites = next(os.walk('dirsrvtests/tests/suites/'))[1]
+# If we have arguments passed to the script, use them as the test names to run
+if len(sys.argv) > 1:
+    suites = sys.argv[1:]
+    valid_suites = []
+    # Validate if the path is a valid file or directory with files
+    for suite in suites:
+        test_path = os.path.join("dirsrvtests/tests/suites/", suite)
+        if os.path.exists(test_path) and not os.path.islink(test_path):
+            if os.path.isfile(test_path) and test_path.endswith(".py"):
+                valid_suites.append(suite)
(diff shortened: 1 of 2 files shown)


vashirov:
See attached screenshot on how to use these new parameters:
image

droideck (approved):
LGTM!

PR #5056, opened 2021-12-10
Issue 5055 - Improve core dump detection and collection in PR CI

### LINKED ISSUE #5055: Improve core dump detection and collection in PR CI
**Issue Description**
We have core file logging during the test run:
https://github.com/389ds/389-ds-base/blob/master/dirsrvtests/report.py#L44-L53

Unfortunately, this doesn't work on GitHub Actions runners. We use Ubuntu VM and run tests in a Fedora container with systemd. When container is started, systemd configures `/proc/sys/kernel/core_pattern` to use `systemd-coredump`, which by default is not installed on the runner. So when a core dumps, it's not piped to an existing binary. Even if we install it, the core would be generated on the host, not inside the container, so our logging won't be able to access it.

I propose to set `core_pattern` to location inside the container and collect cores as part of the pytest artifacts.

vashirov:
Description:
Set `core_pattern` to location inside the container so that cores can be
collected as part of the pytest artifacts.

Fixes: https://github.com/389ds/389-ds-base/issues/5055

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -64,12 +64,14 @@ jobs:
     - name: Run pytest in a container
       run: |
         set -x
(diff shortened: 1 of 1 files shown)


mreynolds389:
LGTM

PR #5057, opened 2021-12-12
Issue 4299 - UI LDAP editor - add "edit" and "rename" functionality

mreynolds389:
Description:

Reworked the entry edit wizard to be one form to edit all
aspects of the entry.  Also add the ability to do modrdns.

Reviewed by: ?


droideck (changes requested):
Looks really good!
I have only two concerns.
First is minor:
1. I think it may be worth adding some visual aid to the fields that require values. Like if you add a new attribute, on 'Edit Values' step, button 'Next' is gray and there is no indication why. Can we make the field 'red' or something when it's empty? (like it doesn't allow you to remove value from the field)
image

2. userPassword attribute displayed as empty even when I've added a value. Hence, it's not possible to proceed with the edit if you try to set a userPassword attribute.

The rest is awesome and intuitive. :)

tmihinto:
@mreynolds389 @droideck 
You might want to add something like a label to provide a more information.

For instance:

`
title: (value, rowIndex, cellIndex, props) => (
           <React.Fragment>
             <EditableTextCell
               value={value}
               rowIndex={rowIndex}
               cellIndex={cellIndex}
               props={props}
               handleTextInputChange={this.handleTextInputChange}
               inputAriaLabel={'_' + value} // To avoid empty property when value is empty.
             />
             {
               value === '' &&
               <Label color="red" icon={<InfoCircleIcon />}>
                Empty value!
               </Label>
             }
`
Screenshot_empty_values

mreynolds389:
Thanks @tmihinto !!!  I can't believe that actually works - putting an object under an editableTableCell makes it look like it's in the cell.

mreynolds389:
@droideck @tmihinto - all changes applied, please review...

droideck (changes requested):
Looks good!
One small issue and I think we are good to go :)

When I try to add a user and choose a user password I can't proceed with the creation (Next button is not active):

image

UserPassword works in Edit mode though!

tmihinto:
@mreynolds389 @droideck ,
Hi

The issue is that some attributes ( userPassword, jpegPhoto ) are handled differently.
For instance the userPassword is saved without having to trigger the "Save" action of the editable table.
Therefore the check for empty value(s) is skipped and the "Next" button stays inactive.

I have a working fix that you might want to try.
I have updated the function handlePwdSave() in EditableTable.jsx as below:


this.handlePwdSave = () => {
      const newRows = [...this.state.tableRows];
      if (this.state.pwdRowIndex !== -1) {
        newRows[this.state.pwdRowIndex].cells[1].props.value = this.state.pwdValue;
       
        this.setState({
          tableRows: newRows,
          isPasswordField: false
        }, () => {
          const rowDataToSave = this.buildRowDataToSave();
          this.props.saveCurrentRows(rowDataToSave, this.state.namingRowID);
          // Check for empty value(s) and update the "Next" button.
          const foundEmptyValue = newRows.find(el => el.cells[1].props.value === '');
          this.props.enableNextStep(foundEmptyValue === undefined);
        });
      }
    };

tmihinto:
Screenshot_userPassword_issue

mreynolds389:
@tmihinto - sorry I also got this all working, exactly like this.  Updated patch will be out later this afternoon (just need to revise the generic LDAP entry add" wizard.

tmihinto:
@mreynolds389 , Cool :-) Thanks for the update!

droideck (changes requested):
Everything works from start to finish!
I just found one small thing:

image

UserPassword is displayed in cleartext when I try to create an entry.

Also, I am not sure if it's intended but I see a "Bump version to 2.1.1" commit in this PR https://github.com/389ds/389-ds-base/pull/5057/commits/18ca3f6e45bb01040811e55af133a3bd5fd4c2e9.

droideck (approved):
Looks good! Ack

PR #5060, opened 2021-12-14
Issue 5052 - BUG - Custom filters prevented entry deletion

### LINKED ISSUE #5052: Winsync OneWay fromWindows not replicate deletion
**Issue Description**
After added winSyncWindowsFilter parameter in winsync agreement, the deletion of an entry in AD does not reflect D389

 - Platform: Debian Bullseye
 - Package and version: 389-Directory/1.4.4.11 B2021.139.1122

**Steps to Reproduce**
1. Add for example this winSyncWindowsFilter: (&(objectCategory=Person)(sAMAccountName=*)(memberOf=CN=Portal Users,OU=Groups,DC=lab,DC=local))
2. Try to delete a synced entry in AD
3. The corresponding entry in D389 will not be deleted

**Expected results**
Expected deletion as for winSyncWindowsFilter is not set

Firstyear:
Bug Description: When a custom filter was provided, entries
which were deleted in AD did not have that event correctly
reflected in 389-ds. This was due to the behaviour that when
an entry in AD is deleted, it is marked with a "deleted" flag
which the objectClass=* filter would (accidentally) collect
when it did a search. However, a custom user filter being
specified would in some cases (such as a memberOf filter)
NOT show up the deletion since the entry was considered
to have moved out of scope rather than being a full delete.

Fix Description: In the case that we have a userfilter, we
wrap it in an OR condition that always requests isDeleted
flags so that we can correctly reflect the delete status.

fixes: https://github.com/389ds/389-ds-base/issues/5052

Review by: ???

### DIFF
--- a/ldap/servers/plugins/replication/windows_connection.c
+++ b/ldap/servers/plugins/replication/windows_connection.c
@@ -776,14 +776,26 @@ send_dirsync_search(Repl_Connection *conn)
         slapi_log_err(SLAPI_LOG_REPL, windows_repl_plugin_name, "send_dirsync_search - Calling dirsync search request plugin\n");
         userfilter = windows_private_get_windows_userfilter(conn->agmt);
         if (userfilter) {
-            filter = slapi_ch_strdup(userfilter);
+            /*
+             * When we have a userfilter, we encounter an issue where a previously
+             * matching object that is *deleted* that we had synced, was not being
+             * deleted. This is because in the unfiltered case, we relied on the
+             * objectClass=*  to get everything, but when we apply a filter we are
+             * removing items that were deleted, especially if they were members of
+             * a group. As a result, we need to *always* request the isDeleted flag
+             * so that we can correct delete any remnants on our side.
+             */
+            size_t buflen = 18 + strlen(userfilter);
+            filter = slapi_ch_calloc(1, buflen);
+            snprintf(filter, buflen, "(|(isDeleted=*)%s)", userfilter);
         } else {
             filter = slapi_ch_strdup("(objectclass=*)");
         }
 
         winsync_plugin_call_dirsync_search_params_cb(conn->agmt, old_dn, &dn, &scope, &filter,
                                                      &attrs, &server_controls);
-        slapi_log_err(SLAPI_LOG_REPL, windows_repl_plugin_name, "send_dirsync_search - Sending dirsync search request\n");
+        slapi_log_err(SLAPI_LOG_REPL, windows_repl_plugin_name, "send_dirsync_search - Sending dirsync search request %s %d %s\n",
+            dn, scope, filter);
 
         rc = ldap_search_ext(conn->ld, dn, scope, filter, attrs, PR_FALSE, server_controls,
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/replication/windows_connection.c
```diff
@@ -776,14 +776,26 @@ send_dirsync_search(Repl_Connection *conn)
...
+             * matching object that is *deleted* that we had synced, was not being
+             * deleted. This is because in the unfiltered case, we relied on the
+             * objectClass=*  to get everything, but when we apply a filter we are
+             * removing items that were deleted, especially if they were members of
+             * a group. As a result, we need to *always* request the isDeleted flag
+             * so that we can correct delete any remnants on our side.
+             */
+            size_t buflen = 18 + strlen(userfilter);
+            filter = slapi_ch_calloc(1, buflen);
+            snprintf(filter, buflen, "(|(isDeleted=*)%s)", userfilter);
```
tbordaz, line 790: As I am not expert of Winsync, sorry for this dummy question.
This new filters looks a bit weird to me, because it will return all the matching entries (userfilter) and all the deleted entries. Is it what you expect or rather something like "(&(I(isDeleted=*)(objectclass=*))%s)".
Also are we sure that isDeleted attribute is indexed ?

Firstyear: Good catch @tbordaz it looks like https://docs.microsoft.com/en-us/windows/win32/adschema/a-isdeleted is not indexed. But because of how the dirsync works, it's similar to syncrepl so it only sends us what has changed between the previous cookie and the current. 

The other possible way to do this filter is:

```
(| (userFilter ...)  (&(isDeleted=True)(userFilter ...))  )
```

That way we can lean on the userFilter to be indexed, but also means that we'll limit the isDeleted results to our userFilter set. What do you think?

tbordaz: My understanding is that 'isDeleted' is similar to tombstone entry. So the initial filter would match all tombstones whether or not they match the userFilter.
IIUC the need, the second filter you proposed looks good to me
<pre>
(| (userFilter ...)  (&(isDeleted=True)(userFilter ...))  )
</pre>

Firstyear: @tbordaz  See https://github.com/389ds/389-ds-base/issues/5052#issuecomment-1005388589 I think there is no way to fix this ....

mreynolds389: @tbordaz @Firstyear - any update on this?  Thierry do you approve this after Williams remarks?  Or do we still have concerns?

tbordaz: Absolutely, Sorry I missed that it was pending on my update

tbordaz (approved):
LGTM

Firstyear:
```
   148ad351d..13e66bfa5  389-ds-base-2.3 -> 389-ds-base-2.3
   ee0ba3cec..f6c639768  389-ds-base-2.2 -> 389-ds-base-2.2
```

Issue #5049, opened 2021-12-10
ns-slapd crash in replication/acceptance_test.py


vashirov:
**Issue Description**
replication/acceptance_test.py sometimes crashes the server. I was able to reproduce it somewhat reliably only in PR CI runs.

[Stacktrace](https://github.com/389ds/389-ds-base/files/7689168/stacktrace.txt)

Full core is available at https://github.com/vashirov/389-ds-base/suites/4610865242/artifacts/124635010 under assets/cores/.
RPMs with debuginfo are at https://github.com/vashirov/389-ds-base/suites/4610865242/artifacts/124635048

tbordaz:
The problem is dynamic and ATM only occurs, from time to time, on CI (not on laptop/1MT).

The scenario of tests is to send replicated updates to a  supplier and while the replication session has not yet completed, the tests removes the replica and then the mapping tree. The crash occurs during a replicated req (extopEnd) that triggers a writedown of the RUV, but the replica structure has already been freed (via MT factory extnode).

Note the replica->repl_lock being overwritten '0x7373616c43746365'
<pre>
Thread 1 (Thread 0x7f5d9b1f8640 (LWP 322)):
#0  ___pthread_mutex_lock (mutex=mutex@entry=0x7373616c4374636d) at pthread_mutex_lock.c:76
#1  0x00007f5dda8d0c45 in PR_EnterMonitor (mon=0x7373616c43746365) at ../../../../nspr/pr/src/pthreads/ptsynch.c:613
#2  0x00007f5dd651b589 in replica_lock (lock=<optimized out>) at ldap/servers/plugins/replication/repl5_replica.c:108
#3  replica_update_ruv_consumer (r=0x7f5da2242dc0, supplier_ruv=0x7f5dd6d4bd60) at ldap/servers/plugins/replication/repl5_replica.c:3776
#4  0x00007f5dd6500ad5 in multisupplier_extop_EndNSDS50ReplicationRequest (pb=0x7f5daee07580) at ldap/servers/plugins/replication/repl_extop.c:1194
#5  0x0000563218701dfd in do_extended (pb=0x7f5daee07580) at ldap/servers/slapd/extendop.c:454
#6  0x00005632186f709a in connection_dispatch_operation (pb=0x7f5daee07580, op=<optimized out>, conn=<optimized out>) at ldap/servers/slapd/connection.c:689
#7  connection_threadmain () at ldap/servers/slapd/connection.c:1785
</pre>

[...]

PR #5063, opened 2021-12-15
Issue 5049 - ns-slapd crash in replication/acceptance_test.py

tbordaz:
Bug description:
	The pointer to a replica structure is stored in
	object extension in the mapping tree (mtn_extension)
	and in incoming replication connection (connect_ext).
	If the replica structure is freed by mapping tree
	the replication connection can access the freed
	replica structure (at session termination)
	and most probably will crash.

Fix description:
	There is no easy way to update the replication
	connection. The fix is just to reduce the window
	of possibility of use after free.
	When an adminstrative task deletes a mapping tree
	(that is quite rare event), the replica_destroy
	function will pause for 3 seconds.
	This will likely give the time to complete
	the end of the replication session.

relates: https://github.com/389ds/389-ds-base/issues/5049

Reviewed by:

Platforms tested: f34

### DIFF
--- a/ldap/servers/plugins/replication/repl5_replica.c
+++ b/ldap/servers/plugins/replication/repl5_replica.c
@@ -308,6 +308,18 @@ replica_destroy(void **arg)
 
     slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name, "replica_destroy\n");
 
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

Issue #5037, opened 2021-11-26
in OpenQA changelog trimming can crashes


abbra:
OpenQA shows that a replica deployment crashes in https://openqa.fedoraproject.org/tests/1071548#step/realmd_join_sssd/33

I am attaching logs which also can be found in a "Logs & Assets" tab at the link above.

There is a crash and then AVCs when systemd-coredump attempts to create a stack trace. AVCs are post crash, not a reason for crash.

Package version: 389-ds-base-2.0.11-1.fc36

Fedora bug with logs and coredumps: https://bugzilla.redhat.com/show_bug.cgi?id=2026980

I suspect it is because there is no check for NULL of cldb in cl5TrimMain():

static int
_cl5TrimMain(void *param)
{
    struct timespec current_time = {0};
    struct timespec prev_time = {0};
    Replica *replica = (Replica *)param;
    cldb_Handle *cldb = replica_get_cl_info(replica);
    int32_t trimInterval = cldb->clConf.trimInterval;

    /* Get the initial current time for checking the trim interval */
    clock_gettime(CLOCK_MONOTONIC, &prev_time);

    /* Lock the CL state, and bump the thread count */
    pthread_mutex_lock(&(cldb->stLock));
    slapi_counter_increment(cldb->clThreads);

    while (cldb->dbState == CL5_STATE_OPEN)
    {
        pthread_mutex_unlock(&(cldb->stLock));

        clock_gettime(CLOCK_MONOTONIC, &current_time);
        if (current_time.tv_sec - prev_time.tv_sec >= trimInterval) {
            /* time to trim */
            prev_time = current_time;
            _cl5TrimReplica(replica);
        }

        pthread_mutex_lock(&(cldb->clLock));
        /* While we have the CL lock get a fresh copy of the trim interval */
        trimInterval = cldb->clConf.trimInterval;
        current_time.tv_sec += trimInterval;
        pthread_cond_timedwait(&(cldb->clCvar), &(cldb->clLock), &current_time);
        pthread_mutex_unlock(&(cldb->clLock));

        pthread_mutex_lock(&(cldb->stLock));
    }
    slapi_counter_decrement(cldb->clThreads);

    pthread_mutex_unlock(&(cldb->stLock));

    return 0;
}

where 

void *
replica_get_cl_info(Replica *r)
{
    return r ? r->cldb : NULL;
}

so if r->cldb is NULL, we'd crash on trimInterval initialization.

tbordaz:
[...]

PR #5070, opened 2021-12-17
Issue 5037 - in OpenQA changelog trimming can crashes

tbordaz:
Bug description:
	The changelog trimming thread is launched
	upon various conditions (changelog open,
	create replica, check RUVs, enable replication,
	...).
	The trimming thread is stopped upon
	various conditions (import, changelog close,
	delete replica, reload RUVs, disable replication
	demote supplier,...)
	There are two issues:
	In case the trimming is stopped while the thread
	has not yet started, the trimming thread can crash
	because some required data (cldb) have been
	cleared under it.

	In case the trimming is restarted while the thread
	has not yet started, there is a possiblity of
	starting several trimming threads

Fix description:
	The fix to prevent the first issue, just checks that
        the required data (cldb) is set.
	The second fix is to use a flag (trimmingOnGoing)
	to prevent multiple trimming threads. The flag is
	protected by stLock.

relates: https://github.com/389ds/389-ds-base/issues/5037

Reviewed by:

Platforms tested: F34

foo

### DIFF
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -125,6 +125,9 @@ struct cl5DBFileHandle
     Slapi_Counter *clThreads; /* track threads operating on the changelog */
     pthread_mutex_t clLock; /* controls access to trimming configuration  and */
                             /* lock associated to clVar, used to notify threads on close */
+    int32_t trimmingOnGoing; /* it is a flag to indicate that a trimming thread is started
+                              * and to prevent another trimming thread to start
+                              */
     pthread_cond_t clCvar; /* Condition Variable used to notify threads on close */
     pthread_condattr_t clCAttr; /* the pthread condition attr */
     void *clcrypt_handle;   /* for cl encryption */
@@ -1274,6 +1277,7 @@ cldb_SetReplicaDB(Replica *replica, void *arg)
     }
     cldb->clThreads = slapi_counter_new();
     cldb->dbState = CL5_STATE_OPEN;
+    cldb->trimmingOnGoing = 0;
 
     if (pthread_mutex_init(&(cldb->stLock), NULL) != 0) {
         slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name_cl,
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -2192,13 +2196,33 @@ _cl5TrimMain(void *param)
...
     pthread_mutex_lock(&(cldb->stLock));
+
+    /* First check that no other trimming thread is running */
+    if (cldb->trimmingOnGoing) {
+        pthread_mutex_unlock(&(cldb->stLock));
+        return 0;
+    }
+
+    /* Now trimming thread can start */
+    cldb->trimmingOnGoing = 1;
```
droideck, line 2225: I am not sure if it's a legit concern but do we also need to take into an account other places where we call `slapi_counter_increment(cldb->clThreads);` and `slapi_counter_decrement(cldb->clThreads);`? Like cl5CreateReplayIteratorEx(), cl5CreateReplayIterator(), etc.

droideck (commented):
The rest looks good :)

ldap/servers/plugins/replication/cl5_api.c
tbordaz: Good point !

One part of the patch is prevent the creation of several trimming threads. It uses a cldb flag (trimmingOnGoing) protected by stLock. This part does not really or change the clThread mechanism (see below)

A second part of the patch (test cldb != NULL) is to address a rare race condition (depending on thread dispatching) that can crash the server.
clThreads is a way to prevent to close/remove cldb under others threads (remove changelog file, read clRUV, export/import, write update, trimming/RA cursor, ...). This mechanism is not perfect and cldb could be cleared from the replica while _cl5TrimMain is called. We could improve clThreads, for example moving it from cldb to the replica, but I found this solution a bit overkill compare to the test cldb != NULL. Do you prefer to remove the test and change the clThreads mechanism ?

droideck: Okay, thank you for the explanations! I had concerns regarding the trimmingOnGoing flag, actually, and you explained it. :)
I agree that (test cldb != NULL) is the way to go for the second issue.

droideck:
As your PR doesn't touch Rust code and there is no mentioning the commit message about that - I think it's better to remove the `src/Cargo.lock` change from this PR.

droideck (approved):
Besides the minor `src/Cargo.lock` thingy, - Looks Good to me! Ack.

Issue #5085, opened 2022-01-04
Race condition about snmp collator at startup


progier389:
**Issue Description**
Got random crash while starting the server:

Thread 3 "ns-slapd" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff6e765640 (LWP 8458)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
49	  return ret;
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff6b898a4 in __GI_abort () at abort.c:79
#2  0x00007ffff6de543e in PR_Assert (s=s@entry=0x7ffff7518240 "mutex != NULL", file=file@entry=0x7ffff75181c0 "/home/progier/sb/i4939/tst/source/389-ds-base/ldap/servers/slapd/slapi2runtime.c", ln=ln@entry=91) at ../../../../nspr/pr/src/io/prlog.c:571
#3  0x00007ffff74773ea in slapi_unlock_mutex (mutex=<optimized out>) at /home/progier/sb/i4939/tst/source/389-ds-base/ldap/servers/slapd/slapi2runtime.c:91
#4  0x00007ffff7479cb1 in set_snmp_interaction_row (host=host@entry=0x6050009f4d80 "linux.home", port=port@entry=39002, error=error@entry=0) at /home/progier/sb/i4939/tst/source/389-ds-base/ldap/servers/slapd/snmp_collator.c:252
#5  0x00007ffff73a714e in slapi_ldap_init_ext (ldapurl=ldapurl@entry=0x0, hostname=0x6050009f4d80 "linux.home", port=39002, secure=secure@entry=0, shared=shared@entry=1, ldapi_socket=ldapi_socket@entry=0x0) at /home/progier/sb/i4939/tst/source/389-ds-base/ldap/servers/slapd/ldaputil.c:750
#6  0x00007ffff03aea3a in conn_connect_with_bootstrap (conn=conn@entry=0x60f000262bd0, bootstrap=bootstrap@entry=0) at /home/progier/sb/i4939/tst/source/389-ds-base/ldap/servers/plugins/replication/repl5_connection.c:1183
[...]

PR #5086, opened 2022-01-04
Issue 5085 - Race condition about snmp collator at startup

progier389:
Trivial fix to avoid trying to update snmp statistics before snmp framework get fully initialized.

Issue: [5085](https://github.com/389ds/389-ds-base/issues/5085)

Reviewd by:


droideck (approved):
LGTM!

Issue #4812, opened 2021-06-24
Listener thread does not scale when dealing with a high number of established connections


tbordaz:
**Issue Description**
Listener thread poll established connections. When the number of connections is high and/or there is a high incoming traffic, this single thread becomes a bottleneck. The thread is eating a lot of cpu to go through the huge array of established connection but it is not running fast enough if there is a big incoming traffic and finally limits the capacity of the server.

The idea is to split the connection table into several portions. Each portion being poll by a dedicated listener thread. 
This solution would also help to scale for 10K problem

**Package Version and Platform:**
 all version

**Steps to Reproduce**
Configure the server with a connection table with 5000 slots. Create 5000 clients and hammer the server. The overall throughput is lower than with 100 clients hammering the server (TBC)
One symptom would be high CPU on listener.

**Expected results**
High CPU should be on workers and throughput with large number of client should be higher than with smaller number of clients

jchapma:
Design doc for review - https://github.com/mreynolds389/389wiki/pull/64

tbordaz:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1975930

progier389:
Hi @jchapma
 I am afraid that there is a regression with latest commit ! 
Instance creation fails on ASAN build (server fails to start)
Problem is due to a buffer overflow 
#0  __sanitizer::Die () at ../../../../libsanitizer/sanitizer_common/sanitizer_termination.cpp:50
#1  0x00007ffff7694dbc in __asan::ScopedInErrorReport::~ScopedInErrorReport (this=0x7fffffffb8a6, __in_chrg=<optimized out>) at ../../../../libsanitizer/asan/asan_report.cpp:190
#2  0x00007ffff7694686 in __asan::ReportGenericError (pc=4364047, bp=bp@entry=140737488340240, sp=sp@entry=140737488340232, addr=108095737118976, is_write=is_write@entry=true, access_size=access_size@entry=8, exp=0, fatal=true) at ../../../../libsanitizer/asan/asan_report.cpp:478
#3  0x00007ffff769563f in __asan::__asan_report_store8 (addr=<optimized out>) at ../../../../libsanitizer/asan/asan_rtl.cpp:126
#4  0x000000000042970f in connection_table_new (table_size=table_size@entry=1024) at /home/progier/sb/ai1/tst/source/389-ds-base/ldap/servers/slapd/conntable.c:198
[...]

PR #5090, opened 2022-01-06
Issue 4812 - Scalability with high number of connections

jchapma:
Description: Listener thread poll established connections. When the
number of connections is high and/or there is a high incoming traffic,
this single thread becomes a bottleneck. The thread is eating a lot of
cpu to go through the huge array of established connection but it is not
running fast enough if there is a big incoming traffic and finally
limits the capacity of the server.

Fix: Create multi connection table active lists where each list is
managed by a dedicated listener thread.

Relates: https://github.com/389ds/389-ds-base/issues/4812

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -124,7 +124,7 @@ Connection_Table *
 connection_table_new(int table_size)
 {
     Connection_Table *ct;
-    int i = 0;
+    size_t i = 0;
     int ct_list = 0;
     int free_idx = 0;
     ber_len_t maxbersize = config_get_maxbersize();
@@ -132,6 +132,8 @@ connection_table_new(int table_size)
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
     ct->list_num = 2;
+    /* Round up table_size to balance the CT lists */
+    table_size += (table_size % ct->list_num);
     ct->list_size = table_size/ct->list_num;
     ct->list_select = 0;
     ct->c = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection));
@@ -309,7 +311,7 @@ connection_table_get_connection(Connection_Table *ct, int sd)
 /* active connection iteration functions */
 
 Connection *
-connection_table_get_first_active_connection(Connection_Table *ct, int listnum)
+connection_table_get_first_active_connection(Connection_Table *ct, size_t listnum)
 {
     return ct->c[listnum][0].c_next;
 }
@@ -329,7 +331,7 @@ connection_table_iterate_active_connections(Connection_Table *ct, void *arg, Con
      */
     Connection *current_conn = NULL;
     int ret = 0;
-    int i = 0;
+    size_t i = 0;
     PR_Lock(ct->table_mutex);
     for (i = 0; i < ct->list_num; i++) {
         current_conn = connection_table_get_first_active_connection(ct, i);
@@ -510,7 +512,7 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
 #endif
 }
 /* Get which active list to use */
-int
+size_t
 connection_table_get_list(Connection_Table *ct)
 {
     int ret = -1;
@@ -536,7 +538,9 @@ connection_table_as_entry(Connection_Table *ct, Slapi_Entry *e)
     char maxthreadbuf[BUFSIZ];
     struct berval val;
     struct berval *vals[2];
-    int i, ct_list, nconns, nreadwaiters;
+    size_t ct_list;
+    size_t i;
+    int nconns, nreadwaiters;
     struct tm utm;
 
     vals[0] = &val;
@@ -613,7 +617,7 @@ connection_table_as_entry(Connection_Table *ct, Slapi_Entry *e)
                     bufptr = newbuf;
                 }
 
-                sprintf(bufptr, "%d:%s:%d:%d:%s%s:%s:%s:%" PRIu64 ":ip=%s",
+                sprintf(bufptr, "%zu:%s:%d:%d:%s%s:%s:%s:%" PRIu64 ":ip=%s",
                         i,
                         buf2,
                         ct->c[ct_list][i].c_opsinitiated,
@@ -667,8 +671,8 @@ connection_table_as_entry(Connection_Table *ct, Slapi_Entry *e)
 void
 connection_table_dump_activity_to_errors_log(Connection_Table *ct)
 {
-    int i;
-    int l;
+    size_t i;
+    size_t l;
     for (l = 0; l < ct->list_num; l++) {
         for (i = 0; i < ct->list_size; i++) {
             Connection *c = &(ct->c[l][i]);
@@ -681,7 +685,7 @@ connection_table_dump_activity_to_errors_log(Connection_Table *ct)
                     int r = ct->fd[l][ct_list].out_flags & SLAPD_POLL_FLAGS;
                     if (r) {
                         slapi_log_err(SLAPI_LOG_CONNS, "connection_table_dump_activity_to_errors_log",
-                                    "activity on %d%s\n", i, r ? "r" : "");
+                                    "activity on %zu%s\n", i, r ? "r" : "");
                     }
                 }
                 pthread_mutex_unlock(&(c->c_mutex));
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -127,15 +127,18 @@ connection_table_new(int table_size)
     size_t i = 0;
     int ct_list = 0;
     int free_idx = 0;
+    int num_hw_threads = 0;
     ber_len_t maxbersize = config_get_maxbersize();
-
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
-    ct->list_num = 2;
-    /* Round up table_size to balance the CT lists */
-    table_size += (table_size % ct->list_num);
-    ct->list_size = table_size/ct->list_num;
     ct->list_select = 0;
+    /* Number of processors determine if we use two or four CT lists */
+    num_hw_threads = util_get_capped_hardware_threads(MIN_CT_HW_THREADS, MAX_CT_HW_THREADS);
+    ct->list_num = (num_hw_threads >= FLEX_POINT_NUM_CT_HW_THREADS) ? MAX_NUM_CT_LISTS : MIN_NUM_CT_LISTS;
+    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new",
+                  "number of lists: %d\n", ct->list_num);
+
+    ct->list_size = table_size/ct->list_num + 1; /* +1 to avoid rounding issue */
     ct->c = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection));
     ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
     ct->table_mutex = PR_NewLock();
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -127,22 +127,15 @@ connection_table_new(int table_size)
     size_t i = 0;
     int ct_list = 0;
     int free_idx = 0;
-    int num_hw_threads = 0;
     ber_len_t maxbersize = config_get_maxbersize();
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
-    ct->list_select = 0;
-    #if 0
-    /* Number of processors determine if we use two or four CT lists */
-    num_hw_threads = util_get_capped_hardware_threads(MIN_CT_HW_THREADS, MAX_CT_HW_THREADS);
-    ct->list_num = (num_hw_threads >= FLEX_POINT_NUM_CT_HW_THREADS) ? MAX_NUM_CT_LISTS : MIN_NUM_CT_LISTS;
-    #else
-    ct->list_num = config_get_conntable_numlists();
-    #endif
+    ct->list_num = config_get_num_listeners();
 
     slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "number of lists: %d\n", ct->list_num);
 
     ct->list_size = table_size/ct->list_num + 1; /* +1 to avoid rounding issue */
+    ct->num_active = (int *)slapi_ch_calloc(1, ct->list_num * sizeof(int));
     ct->c = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection));
     ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
     ct->table_mutex = PR_NewLock();
@@ -227,6 +220,7 @@ connection_table_free(Connection_Table *ct)
     slapi_ch_free((void **)&ct->c);
     slapi_ch_free((void **)&ct->fd);
     PR_DestroyLock(ct->table_mutex);
+    slapi_ch_free((void *)&ct->num_active);
     slapi_ch_free((void **)&ct);
 }
 
@@ -445,6 +439,10 @@ connection_table_move_connection_out_of_active_list(Connection_Table *ct, Connec
     /* We need to lock here because we're modifying the linked list */
     PR_Lock(ct->table_mutex);
 
+    /* Decrement the number of active connections on the ct list this connection was assigned. */
+    (*(ct->num_active + c->c_ct_list))--;
+    c->c_ct_list = -1;
+
     c->c_prev->c_next = c->c_next;
 
     if (c->c_next) {
@@ -502,8 +500,9 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
     connection_table_dump_active_connection(c);
 #endif
 
-    /* Map new connection to connection table list */
+    /* Get the least used ct list and incremant its number of active connections. */
     c->c_ct_list = connection_table_get_list(ct);
+    (*(ct->num_active + c->c_ct_list))++;
 
     c->c_next = ct->c[c->c_ct_list][0].c_next;
     if (c->c_next != NULL) {
@@ -518,15 +517,21 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
     connection_table_dump_active_connections(ct);
 #endif
 }
-/* Get which active list to use */
-size_t
+
+/* Find a connection table list with the lowest number of connections. */
+int
 connection_table_get_list(Connection_Table *ct)
 {
-    int ret = -1;
-    /* No need to lock here as calling function will have the ct lock */
-    ret = ct->list_select;
-    ct->list_select = (ct->list_select + 1) % ct->list_num;
-    return ret;
+    size_t i;
+    int list = 0;
+    int lowest = ct->num_active[0];
+    for (i = 1; i < ct->list_num; i++) {
+        if (*(ct->num_active + i) < lowest) {
+            lowest = *(ct->num_active + i);
+            list = i;
+        }
+    }
+    return list;
 }
 
 /*
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -125,12 +125,17 @@ connection_table_new(int table_size)
 {
     Connection_Table *ct;
     int i = 0;
+    int ct_list = 0;
+    int free_idx = 0;
     ber_len_t maxbersize = config_get_maxbersize();
 
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
-    ct->c = (Connection *)slapi_ch_calloc(1, table_size * sizeof(Connection));
-    ct->fd = (struct POLL_STRUCT *)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
+    ct->list_num = 2;
+    ct->list_size = table_size/ct->list_num;
+    ct->list_select = 0;
+    ct->c = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection));
+    ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
     ct->table_mutex = PR_NewLock();
     /* Allocate the freelist */
     ct->c_freelist = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection *));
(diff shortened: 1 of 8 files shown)


ldap/servers/slapd/conntable.c
```diff
@@ -125,12 +125,17 @@ connection_table_new(int table_size)
...
     int i = 0;
+    int ct_list = 0;
+    int free_idx = 0;
     ber_len_t maxbersize = config_get_maxbersize();
 
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
-    ct->c = (Connection *)slapi_ch_calloc(1, table_size * sizeof(Connection));
-    ct->fd = (struct POLL_STRUCT *)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
+    ct->list_num = 2;
```
Firstyear: For bonus marks, consider https://man7.org/linux/man-pages/man3/numa.3.html  numa_num_configured_nodes() to get how many numa regions we have. Alternately, numa_num_configured_cpus()  / 4.

ldap/servers/slapd/conntable.c
```diff
@@ -125,12 +125,17 @@ connection_table_new(int table_size)
...
+    ct->list_size = table_size/ct->list_num;
```
Firstyear: If the user has configured a list_size that is not a multiple of table_size or list_num, we'll end up with orhapned connections that can't be accessed. So I think there needs to be a way to assert that these are valid sizes.

ldap/servers/slapd/conntable.c
```diff
@@ -314,14 +329,17 @@ connection_table_iterate_active_connections(Connection_Table *ct, void *arg, Con
...
     Connection *current_conn = NULL;
     int ret = 0;
+    int i = 0;
     PR_Lock(ct->table_mutex);
-    current_conn = connection_table_get_first_active_connection(ct);
-    while (current_conn) {
-        ret = f(current_conn, arg);
-        if (ret) {
-            break;
+    for (i = 0; i < ct->list_num; i++) {
```
Firstyear, line 335: size_t i in the for loop. Int is not the correct type for array indexing.

ldap/servers/slapd/conntable.c
```diff
@@ -141,53 +146,57 @@ connection_table_new(int table_size)
     pthread_mutexattr_t monitor_attr = {0};
     pthread_mutexattr_init(&monitor_attr);
     pthread_mutexattr_settype(&monitor_attr, PTHREAD_MUTEX_RECURSIVE);
+    for (ct_list = 0; ct_list < ct->list_num; ct_list++) {
+        ct->c[ct_list] = (Connection *)slapi_ch_calloc(1, ct->list_size * sizeof(Connection));
+        ct->fd[ct_list] = (struct POLL_STRUCT *)slapi_ch_calloc(1, ct->list_size * sizeof(struct POLL_STRUCT));
+        /* We rely on the fact that we called calloc, which zeros the block, so we don't
+        * init any structure element unless a zero value is troublesome later
+        */
+        for (i = 0; i < ct->list_size; i++) {
```
Firstyear, line 155: size_t i

ldap/servers/slapd/conntable.c
```diff
@@ -294,9 +309,9 @@ connection_table_get_connection(Connection_Table *ct, int sd)
 /* active connection iteration functions */
 
 Connection *
-connection_table_get_first_active_connection(Connection_Table *ct)
+connection_table_get_first_active_connection(Connection_Table *ct, int listnum)
```
Firstyear: size_t listnum

ldap/servers/slapd/conntable.c
```diff
@@ -471,19 +493,32 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
...
+    ct->c[c->c_ct_list][0].c_next = c;
 
     PR_Unlock(ct->table_mutex);
 
 #ifdef FOR_DEBUGGING
     connection_table_dump_active_connections(ct);
 #endif
 }
+/* Get which active list to use */
+int
```
Firstyear, line 520: size_t

ldap/servers/slapd/conntable.c
```diff
@@ -510,90 +545,92 @@ connection_table_as_entry(Connection_Table *ct, Slapi_Entry *e)
     attrlist_delete(&e->e_attrs, "connection");
     nconns = 0;
     nreadwaiters = 0;
-    for (i = 0; i < (ct != NULL ? ct->size : 0); i++) {
-        PR_Lock(ct->table_mutex);
-        if (ct->c[i].c_state == CONN_STATE_FREE) {
+    for (ct_list = 0; ct_list < (ct != NULL ? ct->list_num : 0); ct_list++) {
```
Firstyear, line 562: size_t

ldap/servers/slapd/conntable.c
```diff
@@ -631,22 +668,24 @@ void
...
-            /* Find the connection we are referring to */
-            int j = c->c_fdi;
-            pthread_mutex_lock(&(c->c_mutex));
-            if ((c->c_sd != SLAPD_INVALID_SOCKET) &&
-                (j >= 0) && (c->c_prfd == ct->fd[j].fd)) {
-                int r = ct->fd[j].out_flags & SLAPD_POLL_FLAGS;
-                if (r) {
-                    slapi_log_err(SLAPI_LOG_CONNS, "connection_table_dump_activity_to_errors_log",
-                                  "activity on %d%s\n", i, r ? "r" : "");
+    int l;
```
Firstyear: size_t

Firstyear (changes requested):

ldap/servers/slapd/conntable.c
jchapma: Hi @Firstyear 
Bonus marks are always welcome :)

I understand the benefit of a process being executed on a NUMA node local to its CPU, does the scheduler policy not try make sure this happens. But, I don't quiet get  how the number of CT lists can be determined by the number of NUMA nodes on a system.

ldap/servers/slapd/conntable.c
jchapma: Good point.

ldap/servers/slapd/conntable.c
progier389: > Hi @Firstyear Bonus marks are always welcome :)
> 
> I understand the benefit of a process being executed on a NUMA node local to its CPU, does the scheduler policy not try make sure this happens. But, I don't quiet get how the number of CT lists can be determined by the number of NUMA nodes on a system.

Since you tested that 2 CT are optimum on a machine with a single numa node, IMHO we should just use twice the number of nodes

ldap/servers/slapd/conntable.c
progier389: Nice catch!     Having:
 ct->list_size = table_size/ct->list_num + 1; /* +1 to avoid rounding issue */
should simply solve that issue 
No orphan connection because ct->list_size * ct->list_num > table_size 
And we do not really care about having list size slightly too long.

ldap/servers/slapd/conntable.c
Firstyear: Yes, but we also then need to ensure that we are spawning the threads on differnt nodes ....

progier389: Good point but that would means splitting the worker thread pool too.
Another point is that in the test Ludwig had done years ago on Solaris using machine with 64 cores, performance started to decrease when having more than 4 CTs .
So I suspect we are trying to get too far.  Maybe we should just have 2 CT if less than 8 CPUS and 4 otherwise. 
And do not bother about numa  for now. Anyway I do not think we could seriously improve things on numa without thinking globally.
(And to be honest, my feeling on that point, is to try to let the kernel handles the optimizations (including affinity handling) as far as possible rather than having the application doing it)

Firstyear: Soundsgood, and it looks like @jchapma has done this, so we can resolve this point :)

jchapma:
@tbordaz 
Rebased

tbordaz:
> @tbordaz Rebased

Thanks @jchapma for the rebase. Unfortunately it crashes on 2.0 or master with that backtrace
<pre>
[Current thread is 1 (Thread 0x7f8874df9640 (LWP 2529930))]
(gdb) where
#0  0x0000000000000000 in ??? ()
#1  0x00007f88e81a21f2 in _pr_poll_with_poll () at /lib64/libnspr4.so
#2  0x000055d10f78f8e0 in ct_list_thread (threadnum=<optimized out>) at ldap/servers/slapd/daemon.c:1293
#3  0x00007f88e81a7524 in _pt_root () at /lib64/libnspr4.so
#4  0x00007f88e813a299 in start_thread () at /lib64/libpthread.so.0
#5  0x00007f88e84da353 in clone () at /lib64/libc.so.6
</pre>

Does it ring any bell ?

jchapma:
> It doesn't, as you know each CT list has a dedicated thread, so there is a mapping between CT lists and thread. Thread 0 uses CT->list[0], etc. I see from the stack trace that threadnum is not set, which is bad news. Let me have a look.

@tbordaz 
I ran it on me dev vm and it runs fine, if you get a chance could you see how many CT lists have been configured:
sudo grep "connection_table_new" /var/log/dirsrv/slapd-inst01/errors

tbordaz:
> > It doesn't, as you know each CT list has a dedicated thread, so there is a mapping between CT lists and thread. Thread 0 uses CT->list[0], etc. I see from the stack trace that threadnum is not set, which is bad news. Let me have a look.
> 
> @tbordaz I ran it on my dev vm and it runs fine, if you get a chance could you see how many CT lists have been configured: sudo grep "connection_table_new" /var/log/dirsrv/slapd-inst01/errors

It looks one  of the CT struct (the slot 2) is wrongly init
<pre>
(gdb) print the_connection_table->list_num
$1 = 4
(gdb) print the_connection_table->fd[0]
$2 = (struct PRPollDesc *) 0x7f8878bfe000
(gdb) print the_connection_table->fd[1]
$3 = (struct PRPollDesc *) 0x7f8878bff400
(gdb) print the_connection_table->fd[2]
$4 = (struct PRPollDesc *) 0x7f8878c00800
(gdb) print the_connection_table->fd[3]
$5 = (struct PRPollDesc *) 0x7f8878c01c00
(gdb) print the_connection_table->fd[4]
$6 = (struct PRPollDesc *) 0x0
(gdb) print *the_connection_table->fd[0]
$7 = {fd = 0x7f8878c7d2e0, in_flags = 1, out_flags = 1}
(gdb) print *the_connection_table->fd[1]
$8 = {fd = 0x7f8878c7d3a0, in_flags = 1, out_flags = 1}
(gdb) print *the_connection_table->fd[2]
$9 = {fd = 0x0, in_flags = 1, out_flags = 0}   <===== invalid fd
(gdb) print *the_connection_table->fd[3]
$10 = {fd = 0x7f88e6eec110, in_flags = 1, out_flags = 0}
</pre>

Firstyear:
@jchapma have you run a stress test under an ASAN build? that can often show you where the issue is.

jchapma:
> (gdb) print *the_connection_table->fd[0]
> $7 = {fd = 0x7f8878c7d2e0, in_flags = 1, out_flags = 1}
> (gdb) print *the_connection_table->fd[1]
> $8 = {fd = 0x7f8878c7d3a0, in_flags = 1, out_flags = 1}
> (gdb) print *the_connection_table->fd[2]
> $9 = {fd = 0x0, in_flags = 1, out_flags = 0}   <===== invalid fd
> (gdb) print *the_connection_table->fd[3]
> $10 = {fd = 0x7f88e6eec110, in_flags = 1, out_flags = 0}

Hi @tbordaz 
The first element in each fd contains a signal handler, so in the 4 CT list case we should have 4. The reason you only see 2 configured fd's is because **someone** didn't remove a #define used when creating the signal table. Initially the num of CT lists were hardcoded to 2, but now we let the number of system HW threads determine the number of CT lists.

tbordaz:
@jchapma thanks for the tip, indeed changing NUM_CT_LISTS=4 made the job. It builds/install/run on my laptop.
Unfortunately running on the large machine I can see a big drop in both throughput and response time
<pre>

searchrate --hostname _server_ --port 389 --bindDN "cn=Directory manager" --bindPassword password --baseDN dc=example,dc=com --scope sub --filter "(uid=user[0:10000%00000])" --attribute dn --numThreads 50 -I 5

389-ds-base-2.0.14-1.el9.x86_64  (15workers) 
     Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
  150633.809        0.331        1.000        0.000   152188.151        0.327

389-ds-base-2.0.14-1_4812_1.el9_0.x86_64 (15workers)
      Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
   55825.346        0.895        1.000        0.000    56872.064        0.878

</pre>

continue investigation

jchapma:
> changing NUM_CT_LISTS=4 made the job. It builds/install/run on my laptop.
> Unfortunately running on the large machine I can see a big drop in both throughput and response time

I removed the #define NUM_CT_LISTS with the latest PR push, it was used when creating the signal pipe table, we dont use it anymore.

That is a big drop :(

Firstyear:
Given the work on epoll at the moment, is there a need to continue this? Epoll will likely eliminate a lot of our bottlenecks around the CT.

tbordaz:
@jchapma , @Firstyear Sorry for my mistake, after checking the various tuning I realize that my initial numbers were wrong. The multi-listener is still showing a drop but less severe -17%
<pre>
searchrate --hostname _server_ --port 389 --bindDN "cn=Directory manager" --bindPassword password --baseDN dc=example,dc=com --scope sub --filter "(uid=user[0:10000%00000])" --attribute dn --numThreads 50 -I 5

389-ds-base-2.0.14-1.el9.x86_64  (15workers) 
     Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
  150633.809        0.331        1.000        0.000   152188.151        0.327

389-ds-base-2.0.14-1_4812_1.el9_0.x86_64 (15workers)
      Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
  124892.739        0.399        1.000        0.000   124167.984        0.401

</pre>

The goal of the patch is to get a better handling with a high number of connections (10K). The tests above are not testing this. They are just testing with 50 clients.

progier389:
> Given the work on epoll at the moment, is there a need to continue this? Epoll will likely eliminate a lot of our bottlenecks around the CT.

Hi @Firstyear, FYI: we wondered about that point a few weeks ago, but so far @tbordaz's tests shows that performance with epoll are still significantly worse than those with poll, so we decided to continue with this PR at least until the performance degradation with epoll get understood and fixed ...

tbordaz:
@Firstyear, I agree with @progier389 explanations. we have two improvements to evaluate: epoll (that we expect will outperform 'poll' with low/medium/high number of connections), multi-listener (that ticket that is specifically design for a better handling of high number of connections).

ATM both are showing performance regressions with low number of connections.

- epoll (~-40%) https://github.com/389ds/389-ds-base/pull/5182#issuecomment-1073948670
- multi-listeners (~-17%) https://github.com/389ds/389-ds-base/pull/5090#issuecomment-1081785519

For high number of connection we are still working on a tool to measure perf.

I think we should continue investigation of the RC of those regressions because this is possibly a simple issue to fix. Note that I think we can easily achieve multi-listener+epoll so both are valuable.

tbordaz:
@jchapma, I changed the patch to be able to tune the number of listeners. Running your patch with a **single** listener is still showing a small regression (-2%)

<pre>
searchrate --hostname _server_ --port 389 --bindDN "cn=Directory manager" --bindPassword password --baseDN dc=example,dc=com --scope sub --filter "(uid=user[0:10000%00000])" --attribute dn --numThreads 50 -I 5

389-ds-base-2.0.14-1.el9.x86_64  (15workers) 
     Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
  150633.809        0.331        1.000        0.000   152188.151        0.327

      Recent       Recent       Recent       Recent      Overall      Overall
Searches/Sec   Avg Dur ms Entries/Srch   Errors/Sec Searches/Sec   Avg Dur ms
------------ ------------ ------------ ------------ ------------ ------------
  149466.627        0.334        1.000        0.000   149290.828        0.334

</pre>

The new patch introduces a so small overhead (with a single listener), that it may be a false negative. With multi-listener (4 in my previous run), I suspect there may be some contention on writing to the work_queue.

tbordaz:
Likely the regression is due to contention on work_queue
<pre>
perf c2c
  -------------------------------------------------------------
      9        0      129      173        0      0x55614f6ee5c0
  -------------------------------------------------------------
           0.00%    0.78%    0.00%    0.00%                0x20     0       1      0x7f635439d424         0       213         0        1         1  [.] __GI___lll_lock_wait        libc.so.6  __GI___lll_lock_wait+4           0
            |
            ---__GI___lll_lock_wait
               ___pthread_mutex_lock (inlined)
               add_work_q (inlined)
               connection_activity
               handle_pr_read_ready (inlined)
               ct_list_thread
               0x7f63540d3983
               start_thread
               __clone3 (inlined)
~                                   
</pre>

So we have multi-listeners (4), detecting activity on established connections, trying to add a work_item into the work-queue. Under heavy load, listeners are in competition to the get the work-queue access, as well as worker to read from it. A possible fix would be to create a work_queue dedicated to each listeners, only the notification will use a common lock/cv.

jchapma:
> So we have multi-listeners (4), detecting activity on established connections, trying to add a work_item into the work-queue. Under heavy load, listeners are in competition to the get the work-queue access, as well as worker to read from it. A possible fix would be to create a work_queue dedicated to each listeners, only the notification will use a common lock/c

Hi @tbordaz 
Thanks for the profiling work, let me look at adding a work_queue for each listener.

tbordaz:
@jchapma, I wonder if working on the several work_queue, could be done in a separated ticket. I think we can make work_q being in an array, having the array with a single slot.

Then we will apply multi-listener patch on top of multi-work_queue, "just" adding the glue to link a work_queue with a listener.

Regarding multi-listener, the tests with low number of connections is done. IMHO a priority would rather be to work on the test to verify the gain of multi-listener with high number of connections.

jchapma:
Added patch to allow for the configuration of connection table lists for testing. We can remove this later.

ldap/servers/slapd/conntable.c
```diff
@@ -471,19 +502,32 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
...
 
     PR_Unlock(ct->table_mutex);
 
 #ifdef FOR_DEBUGGING
     connection_table_dump_active_connections(ct);
 #endif
 }
+/* Get which active list to use */
+size_t
+connection_table_get_list(Connection_Table *ct)
```
tbordaz, line 521: This function selects which listener thread will poll a new established connection.
It is based on modulo, so depending on connection closure we may end in listeners with unbalanced number of connections. Any reason you prefer this option rather that selecting a listener if its number of connection is low ?

ldap/servers/slapd/conntable.c
```diff
@@ -471,19 +502,32 @@ connection_table_move_connection_on_to_active_list(Connection_Table *ct, Connect
     connection_table_dump_active_connection(c);
 #endif
 
-    c->c_next = ct->c[0].c_next;
+    /* Map new connection to connection table list */
+    c->c_ct_list = connection_table_get_list(ct);
```
tbordaz, line 502: listener thread is assigned here when the connection is active. Shouldn't we reset c_ct_list to some specific value during connection_table_move_connection_out_of_active_list

ldap/servers/slapd/slap.h
```diff
@@ -281,6 +281,8 @@ typedef void (*VFPV)(); /* takes undefined arguments */
 #define SLAPD_DEFAULT_MAXSIMPLEPAGED_PER_CONN_STR "-1"
 /* We'd like this number to be prime for the hash into the Connection table */
 #define SLAPD_DEFAULT_CONNTABLESIZE 4093 /* connection table size */
+#define SLAPD_DEFAULT_CONNTABLE_NUMLISTS 2 /* connection table lists */
```
tbordaz: According to my test there are some drops in throughput/resp_time with any value > 1. I think it should be 1 ATM, until we fix others bottlenecks

ldap/servers/slapd/slap.h
```diff
@@ -2184,6 +2187,7 @@ typedef struct _slapdEntryPoints
 #define CONFIG_MAXTHREADSPERCONN_ATTRIBUTE "nsslapd-maxthreadsperconn"
 #define CONFIG_MAXDESCRIPTORS_ATTRIBUTE "nsslapd-maxdescriptors"
 #define CONFIG_CONNTABLESIZE_ATTRIBUTE "nsslapd-conntablesize"
+#define CONFIG_CONNTABLENUMLISTS_ATTRIBUTE "nsslapd-conntablenumlists"
```
tbordaz: I would prefer a name like nsslapd-numlisteners

ldap/servers/slapd/fe.h
```diff
@@ -73,24 +73,41 @@ int connection_call_io_layer_callbacks(Connection *c);
...
+#define MAX_NUM_CT_LISTS 4
+
 /*
  * Note: the correct order to use when acquiring multiple locks is
  * c[i]->c_mutex followed by table_mutex.
  */
 struct connection_table
 {
     int size;
+    int list_size;
```
tbordaz: It would be good to have precise comment of the purpose of those variable.

tbordaz (commented):
Very nice code and huge improvement. Will continue review

ldap/servers/slapd/libglobs.c
```diff
@@ -4852,6 +4857,41 @@ config_set_conntablesize(const char *attrname, char *value, char *errorbuf, int
     return retVal;
 }
 
+int
+config_set_conntable_numlists(const char *attrname, char *value, char *errorbuf, int apply)
+{
+    int retVal = LDAP_SUCCESS;
+    long nValue = 0;
+    int minVal = 1;
+    int maxVal = 4;
```
tbordaz: Should be MAX_NUM_CT_LISTS.

A comment describing the logic of the setting would be helpful. IMHO we should be allowed to set 0 to 4
0 means that you let the server decide the appropriate value. Currently IIUC this is SLAPD_DEFAULT_CONNTABLE_NUMLISTS (but in the future we may rely on a more complex computation like nsslapd-threadnumber). 
1-4 are acceptables values
outside of this range, you set it to SLAPD_DEFAULT_CONNTABLE_NUMLISTS.

Also, as metrics are showing a performance hit when numlist > 1, I suggest that SLAPD_DEFAULT_CONNTABLE_NUMLISTS=1 and dse.ldif.template also define nsslapd-conntablenumlists=1

ldap/servers/slapd/libglobs.c
```diff
@@ -4852,6 +4857,41 @@ config_set_conntablesize(const char *attrname, char *value, char *errorbuf, int
...
+    }
+
+    errno = 0;
+    nValue = strtol(value, &endp, 0);
+    if (*endp != '\0' || errno == ERANGE || nValue < minVal || nValue > maxVal) {
+        slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE,
+                              "%s: invalid value \"%s\", connection table num lists must range from %d to %d. "
+                              "Server will use a setting of %d.",
+                              attrname, value, minVal, maxVal, defVal);
+        if (nValue > maxVal || nValue < minVal) {
```
tbordaz: I think this test is useless, if anything is weird with the setting just fallback to the default value

ldap/servers/slapd/conntable.c
tbordaz: Just an additional comment, in handle_pr_read_ready we are scanning the all active list of the listener. We could easily compute the number of connections for that listener and update a new field like ct->c[list_num][0].num_active. Later connection_table_get_list would select the listener with the fewer num_active

ldap/servers/slapd/daemon.c
```diff
@@ -74,17 +75,18 @@ int slapd_accept_wakeup_timer = SLAPD_ACCEPT_WAKEUP_TIMER; /* time in ms to wake
...
 
+static signal_pipe signalpipes[MAX_NUM_CT_LISTS]; /* One signal pipe per connection table list */
+static PRInt32 ct_shutdown = 0;
 static PRThread *disk_thread_p = NULL;
 static PRThread *accept_thread_p = NULL;
 static pthread_cond_t diskmon_cvar;
 static pthread_mutex_t diskmon_mutex;
 
 void disk_monitoring_stop(void);
+void init_ct_list_threads(void);
```
tbordaz: both are static functions

ldap/servers/slapd/daemon.c
```diff
@@ -1110,33 +1112,10 @@ slapd_daemon(daemon_ports_t *ports)
...
-            /*
-             *  Now done in accept_thread
-             * handle_listeners(the_connection_table);
-             */
-            /* handle new data ready */
-            handle_pr_read_ready(the_connection_table, connection_table_size);
-            clear_signal(the_connection_table->fd);
-            break;
-        }
+        usleep(250);
```
tbordaz: I think the sleep could be longer. If the server is going to shutdown, usleep(1000) (1ms) looks good enough

ldap/servers/slapd/daemon.c
```diff
@@ -1176,7 +1156,7 @@ slapd_daemon(daemon_ports_t *ports)
     threads = g_get_active_threadcnt();
     if (threads > 0) {
         slapi_log_err(SLAPI_LOG_INFO, "slapd_daemon",
-                      "slapd shutting down - waiting for %" PRIu64 " thread%s to terminate\n",
+                       "slapd shutting down - waiting for %" PRIu64 " thread%s to terminate\n",
```
tbordaz: indentation

ldap/servers/slapd/daemon.c
```diff
@@ -1185,28 +1165,32 @@ slapd_daemon(daemon_ports_t *ports)
...
          * blocked on it. */
-        xpd.fd = signalpipe[0];
-        xpd.in_flags = PR_POLL_READ;
-        xpd.out_flags = 0;
-        spe = PR_Poll(&xpd, 1, PR_INTERVAL_NO_WAIT);
-        if (spe > 0) {
-            spe = PR_Read(signalpipe[0], &x, 1);
-            if (spe < 0) {
+        for(i = 0; i < the_connection_table->list_num; i++)
+        {
```
tbordaz: To follow the indentation of the existing code, you may put '{' on the same line as 'for '

ldap/servers/slapd/fe.h
```diff
@@ -73,24 +73,41 @@ int connection_call_io_layer_callbacks(Connection *c);
  * conntable.c
  */
 
+/* Used to determine the number connection table lists */
+#define MIN_CT_HW_THREADS 1
```
tbordaz: Those 3 define are not use ATM. You may put it with #ifdef

ldap/servers/slapd/conntable.c
```diff
@@ -124,13 +124,27 @@ Connection_Table *
...
+    int free_idx = 0;
+    int num_hw_threads = 0;
     ber_len_t maxbersize = config_get_maxbersize();
-
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->size = table_size;
-    ct->c = (Connection *)slapi_ch_calloc(1, table_size * sizeof(Connection));
-    ct->fd = (struct POLL_STRUCT *)slapi_ch_calloc(1, table_size * sizeof(struct POLL_STRUCT));
+    ct->list_select = 0;
+    #if 0
```
tbordaz: You may choose a more meaningful define

tbordaz (commented):
I completed the review. Thanks

ldap/servers/slapd/conntable.c
jchapma: > Any reason you prefer this option rather that selecting a listener if its number of connection is low ?

I used a simple round robin method because I believe it would be faster than checking which list has the lowest number of connections. But, taking connection closure into consideration, we could end up with lists that are unbalanced. I like your idea, let me look into it.

ldap/servers/slapd/conntable.c
jchapma: connection_table_move_connection_out_of_active_list is called when a connection is inactive, where it is moved back to the freelist for re-use. The next time a connection is taken from the freelist it is assigned a ct list. IMHO the c->ct_list doesnt need to be reset.

tbordaz: I agree it does not need a reset but I am a bit concern that a connection on the freelist keeps its previous listener assigment. I would prefere a setting like: free => c->ct_list == -1, active => c->ct_list = [0-3]

jchapma: Sure, I can do that.

ldap/servers/slapd/libglobs.c
jchapma: > dse.ldif.template also define nsslapd-conntablenumlists=1

If the default config is set to return nsslapd-numlisteners=1, I dont understand why we need to add it to template-dse.ldif

tbordaz: That is a good question, should we make this configuration attribute visible in dse.ldif or not. With CLI we should never edit dse.ldif and rather use dsconf to get/replace the value.
I am fine, not adding this attribute in the template.

Firstyear: No values should be added to dse.ldif.template unless they are REQUIRED for initial server bring up. All other values should be derived from the libglobs.c defaults in all cases to allow config reset.

jchapma:
Here are some search results with a large number of connections. Not very exciting, as the increase in throughput is negligible. 

However, when profiling the sever while searching with 10k connections, the contention on the work_q increases as the number of lists increase. For  example, with a single CT list, perf rates the adding of a connection to the work_q at position 16, when using 4 lists it jumps into the top 10. 

I'm currently working on reducing this contention.

Screenshot from 2022-05-12 16-22-31

Firstyear:
> I'm currently working on reducing this contention.

Yeah, it's pretty common that when you unblock one bottleneck, you find the next one. So I'm not surprised you hit something else here, but it shows you're on the right track.

To stop this getting "too overboard", I think since there are no regresions here, but you find a new bottleneck, should we consider merging this as is so that we don't delay or let the patch go for to long? Can't let perfect be the enemy of good.

jchapma:
> * Limit the number of CT to 1 by default
     This is now hardcoded to 1 CT list, see latest commit: 

> * prevent dsconf to set this number, so that only editing the dse.ldif will be used
      Removed option to configure the number of CT lists, see latest commit: 

> * describe the tests behavior (thread/connection, barrier, duration of run...)
     The aim of the perf client was to perform concurrent search operations on a remote 389 ds instance. The effect of the python global interpreter lock means that all threads are not really concurrent. As the timing observed from the access logs are so small this behaviour is accepted.
     Performance client behaviour:
     Uses a single daemon thread to create a number of connections and keep them open.
     Mode 0 - Create a search thread that searches on a defined number of connections.
     Mode 1 - Create a search thread for each connection.
     In both modes, a barrier is used to synchronise all search threads. Ensuring all search threads begin searching at the same   time. Another barrier is used to sync search completion, so  make sure we don't process the results till all threads have completed. 
  
> * possibly rerun the test master vs master+patch
   The performance machines we are using are running rhel-9.0, so I had to use 389 ds v 2.0.14 instead of the master branch. Here are performance figures comparing vanilla 2.0.14 vs 2.0.14 + a single CT lists.
Screenshot from 2022-05-24 13-50-23

tbordaz (approved):
Thank @jchapma for the very clear status. The patch looks very nice to me (need a rebase on master)

jchapma:
Hi @Firstyear , i you are happy with the changes could you approve the requested changes. I just rebased with master. Thanks

Firstyear (approved):

PR #5091, opened 2022-01-07
Issue 5032 - Fix OpenLDAP version check

### LINKED ISSUE #5032: OpenLDAP is not shipped with non-threaded version of libldap

droideck:
Description: In the Upstream, libldap_r has been merged with libldap
starting from openldap-2.5.
Fix the version check in configure.ac and the comments.

Reviewed by: ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -870,7 +870,7 @@ AM_CONDITIONAL(OPENLDAP,test "$with_openldap" = "yes")
 
 # check for --with-libldap-r
 AC_MSG_CHECKING(for --with-libldap-r)
-AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use libldap_r.so shared library (default: if OpenLDAP version is less than 2.5, then libldap_r.so will be used, else - libldap.so)]),
+AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use lldap_r shared library (default: if OpenLDAP version is less than 2.5, then lldap_r will be used, else - lldap)]),
 [
   if test "$withval" = "no"; then
     AC_MSG_RESULT(no)
--- a/configure.ac
+++ b/configure.ac
@@ -870,7 +870,7 @@ AM_CONDITIONAL(OPENLDAP,test "$with_openldap" = "yes")
 
 # check for --with-libldap-r
 AC_MSG_CHECKING(for --with-libldap-r)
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -870,7 +870,7 @@ AM_CONDITIONAL(OPENLDAP,test "$with_openldap" = "yes")
 
 # check for --with-libldap-r
 AC_MSG_CHECKING(for --with-libldap-r)
-AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use libldap_r.so shared library (default: if OpenLDAP version is less than 2.6, then libldap_r.so will be used, else - libldap.so)]),
+AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use libldap_r.so shared library (default: if OpenLDAP version is less than 2.5, then libldap_r.so will be used, else - libldap.so)]),
```
Firstyear: This relies on how the distro is packaging 2.5 though because libldap.so is from 2.6 not 2.5, and that would require the lib link to be present. So I think this needs to handle that case?

droideck: This change is about lldap_r and lldap only.

In the Upstream, libldap_r has been merged with libldap starting from openldap-2.5:
https://lists.openldap.org/hyperkitty/list/openldap-announce@openldap.org/thread/BH3VDPG6IYYF5L5U6LZGHHKMJY5HFA3L/

So lldap_r should be used before 2.5. (And lldap will be used for 2.5 and newer)

But you are correct that libldap.so and  libldap_r.so wording is wrong (as it refers to 2.6 change). I'll change the comment.

Firstyear: I thought that lldap.so in 2.5 still had a version, such as lldap-2_5.so?

droideck: > I thought that lldap.so in 2.5 still had a version, such as lldap-2_5.so?

I mean `-lldap` and `-lldap_r` parameters that we supply to GCC.
This commit https://github.com/389ds/389-ds-base/commit/a3375a08a506c3b72a8a23df5238848017691350 doesn't touch anything *.so related so the OpenLDAP 2.6 change is not touched.
And if I understood you correctly...

Firstyear (approved):

PR #5093, opened 2022-01-07
Issue 5092 - BUG - pin concread for el8.5

### LINKED ISSUE #5092: Pin concread to avoid el8.5 build issue
Pin concread for el8.5. @vashirov is this for 1.4.4 or 2.0 only?

Firstyear:
Bug Description: el8.5 doesn't have a new enough
rust compiler to build anything with edition 2021

Fix Description: pin concread to an earlier version

fixes: https://github.com/389ds/389-ds-base/issues/5092

Review by: ???

### DIFF
--- a/src/librslapd/Cargo.toml
+++ b/src/librslapd/Cargo.toml
@@ -15,7 +15,7 @@ crate-type = ["staticlib", "lib"]
 [dependencies]
 slapd = { path = "../slapd" }
 libc = "0.2"
-concread =  "^0.2.20"
+concread =  "=0.2.20"
 
 [build-dependencies]
 cbindgen = "0.9"


Firstyear:
Note I'm only going to cherry-pick and apply this to the affected branches, not to main/master.

vashirov (approved):

tbordaz:
> Note I'm only going to cherry-pick and apply this to the affected branches, not to main/master.

This patch allows me to build on master branch, without that patch it fails to download concread. Why do you want to skip master branch ?

Firstyear:
@tbordaz This is pinning to an older version of a dependency, which we shouldn't do. It's only needed due to limitations of EL8.5 and it's rust version. So if you aren't able to build on master without this, that means you probably need to update your environment in someway. If you want I can help you out with this :)

Firstyear:
```
   7518d0468..a46217a05  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

Issue #5095, opened 2022-01-07
sync-repl with openldap may send truncated syncUUID


tbordaz:
**Issue Description**
sync-repl generates syncUUID ([SyncInfoMessage](https://datatracker.ietf.org/doc/html/rfc4533#section-2.5), or [syncStateControl](https://datatracker.ietf.org/doc/html/rfc4533#page-10))

According to [syncUUID](https://datatracker.ietf.org/doc/html/rfc4533#section-2.1.1) definition the UUID is a 16bytes octet string value. So it can contain a byte with NULL value

But when sending a syncUUID value containing a NUL bytes it truncated the syncUUID
<pre>
[3] (3 elem)
  OCTET STRING (52 byte) rid=123,csn=21000101020105.000000Z#000000#000#000000
  BOOLEAN true
  SET (1 elem)
    OCTET STRING (5 byte) 7FEF4573FB  <---- syncUUID

syncUUID[0] = 0x7f
syncUUID[1] = 0xef
syncUUID[2] = 0x45
syncUUID[3] = 0x73
syncUUID[4] = 0xfb
syncUUID[5] = 0x00
syncUUID[6] = 0x4a
syncUUID[7] = 0x91
syncUUID[8] = 0xa4
syncUUID[9] = 0x4f
syncUUID[10] = 0x60
syncUUID[11] = 0x0c
syncUUID[12] = 0xa2
syncUUID[13] = 0x87
syncUUID[14] = 0xbc
syncUUID[15] = 0xa1
</pre>

pyasn1 decoding fails when handling a synInfoMessage with a truncated syncUUID

**Package Version and Platform:**
since 1.4.3

**Steps to Reproduce**
run in a loop tests/suites/syncrepl_plugin/openldap_test.py

**Expected results**

Firstyear:
@tbordaz Did you get this from wireshark? Do you have the packet trace? I suspect it's a bug in pyasn or python-ldap since it looks like the data is intact from what you have shown here.

mreynolds389:
This should be cherry-picked down to 1.4.3

tbordaz:
19570bf3d..06dace1d5  master
c2f6c23ba..cb21e1a20  389-ds-base-2.0
3ab042db6..a2a55e8b0  389-ds-base-1.4.4
7a351d036..730977955  389-ds-base-1.4.3

PR #5099, opened 2022-01-10
Issue 5095 - sync-repl with openldap may send truncated syncUUID

tbordaz:
Bug description:
	When using sync_repl from openldap, syncUUID (that identify an
	entry) is retrieved from targetEntryUUID rather than nsuniqueid.
	syncUUID is a 16 bytes long representation of targetEntryUUID.
        TargetEntryUUID can contain '00' so syncUUID can contain a
	byte with 0x00.
	When creating a syncInfo
(https://datatracker.ietf.org/doc/html/rfc4533#section-2.5)
	syncUUIDS is ber encoded with '[v]' taking a null terminated
	array of (char*). In such case the 0x00 char truncates the
	syncUUID.

Fix description:
	Instead of using a null terminated array of (char*), the
	fix uses a null terminated array of berval.

relates: https://github.com/389ds/389-ds-base/issues/5095

Reviewed by:

Platforms tested: F34

### DIFF
--- a/ldap/servers/plugins/sync/sync_refresh.c
+++ b/ldap/servers/plugins/sync/sync_refresh.c
@@ -745,6 +745,7 @@ void
 sync_send_deleted_entries(Slapi_PBlock *pb, Sync_UpdateNode *upd, int chg_count, Sync_Cookie *cookie)
 {
     char *syncUUIDs[SYNC_MAX_DELETED_UUID_BATCH + 1] = {0};
+    struct berval *ber_syncUUIDs[SYNC_MAX_DELETED_UUID_BATCH + 1] = {0};
     size_t uuid_index = 0;
 
     PR_ASSERT(cookie);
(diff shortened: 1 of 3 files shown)


Firstyear (approved):
Well dang! Nice find! :)

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -764,9 +765,15 @@ sync_send_deleted_entries(Slapi_PBlock *pb, Sync_UpdateNode *upd, int chg_count,
...
-                sync_intermediate_msg(pb, LDAP_TAG_SYNC_ID_SET, cookie, (char **)&syncUUIDs);
+                for (size_t i = 0; i < uuid_index; i++) {
+                    ber_syncUUIDs[i] = (struct berval *) slapi_ch_malloc(sizeof(struct berval));
+                    ber_syncUUIDs[i]->bv_val = syncUUIDs[i];
+                    ber_syncUUIDs[i]->bv_len = 16;
+                }
+                sync_intermediate_msg(pb, LDAP_TAG_SYNC_ID_SET, cookie, ber_syncUUIDs);
                 for (size_t i = 0; i < uuid_index; i++) {
                     slapi_ch_free((void **)&syncUUIDs[i]);
+                    slapi_ch_free((void **)&ber_syncUUIDs[i]);
```
droideck, line 776: I wonder if we should use `ber_bvfree()` here?
And add `ber_syncUUIDs[i] = NULL;` afterwards.

droideck (commented):
Besides the minor concern, looks very good!

ldap/servers/plugins/sync/sync_refresh.c
tbordaz: It could work but I am not confident on what ber_bvfree will do. If it just frees the BerValue (ber_syncUUIDs[i]) that would work but then it is identical to slapi_ch_free as we previously allocated it with slapi_ch_malloc. My fear is that it may in addition free ber_syncUUIDs[i].bv, that is syncUUIDs[i] we just free before. In that case it would lead to a double free.

I prefer to manage the struct in that  function rather than relying on ber_bvfree.

Issue #5105, opened 2022-01-14
During a bind, if the target entry is not reachable the operation may complete without sending result


tbordaz:
**Issue Description**
The issue was found with tests/suites/mapping_tree/referral_during_tot_init_test.py

During a bind, if the target entry is not reachable (backend, referral, bind callback) then the bind operation complete but does not send a reponse. So the client hang (until connection timeout).

Note This type of bug may exist with others types of operation

**Package Version and Platform:**
 - all version

**Steps to Reproduce**
Run in loop tests/suites/mapping_tree/referral_during_tot_init_test.py

**Expected results**
ldap client should receive a bind reponse

tbordaz:
db514a1cf..05961eb97  master
cb21e1a20..389188cda  389-ds-base-2.0
a2a55e8b0..2562d284e  389-ds-base-1.4.4
730977955..5e2404fd1  389-ds-base-1.4.3
9a9292f64..e1cd8b036  389-ds-base-1.3.10

PR #5107, opened 2022-01-14
Issue 5105 - During a bind, if the target entry is not reachable the …

tbordaz:
…operation may complete without sending result

Bug description:
	A bind operation can skip sending back operation result.
	This can happen in rare condition like backend is not available
	or in referral mode or did not define a bind callback.

Fix description:
	Catch those errors condition and send an operation result.

relates: https://github.com/389ds/389-ds-base/issues/5105

Reviewed by:

Platforms tested: F34


progier389 (approved):
LGTM

PR #5109, opened 2022-01-14
Issue 5098 - Multiple issues around replication and CI test test_online_reinit_may_hang

### LINKED ISSUE #5098: Multiple issues around replication and CI test test_online_reinit_may_hang
**Issue Description**

Import and bulk import may break the replication (RUV may be lost - and changelog starting csn may not be created.)

Investigation of CI test failure when working on issue 4939 (LMDB import redesign) showed that the problem was not because of lmdb import code but that there were several issue around: 
1. the test itself. ( Should not have to reinitialize the replication after the import (i.e: replication should not be broken by the import)
2. the bdb import ( RUV entry is skipped)
3. the changelog initialization code (if changelog is empty either the starting csn of the database RUV should be updated or dummy changes about the max csn(s) should be created

1) replication should not be broken when reimporting a master from an ldif including replication data. 
    so IMHO a step is missing in the CI test (tocheck that replication is still working after the import and before the total update 
    (Total update may still be needed because originally the hang that leaded to create the bug was occurring in the total update)
2) while running the test on supplier1 in bdb mode:
[07/Jan/2022:22:34:49.975351682 +0100] - WARN - bdb_import_foreman - import userRoot: Skipping entry "nsuniqueid=ffffffff-ffffffff-ffffffff-ffffffff,dc=example,dc=com" which has no parent, ending at line 15 of file "/home/progier/sb/i4939/tst/ci-install/var/lib/dirsrv/slapd-supplier1/ldif/supplier1.ldif"
...
[07/Jan/2022:22:34:50.647405077 +0100] - INFO - bdb_public_bdb_import_main - import userRoot: Import complete.  Processed 20 entries (1 were skipped) in 1 seconds. (20.00 entries/sec)
...
[07/Jan/2022:22:34:55.111294779 +0100] - DEBUG - NSMMReplicationPlugin - changelog max RUV: {replicageneration} 61d8b1e2000000010000
[07/Jan/2022:22:34:55.114265534 +0100] - DEBUG - NSMMReplicationPlugin - changelog max RUV: {replica 1} 61d8b1e2000100010000 61d8b1e9000100010000 00000000
[07/Jan/2022:22:34:55.117366657 +0100] - DEBUG - NSMMReplicationPlugin - changelog max RUV: {replica 2} 61d8b1e6000100020000 61d8b1ea000200020000 00000000
[07/Jan/2022:22:34:55.120676337 +0100] - DEBUG - NSMMReplicationPlugin - database RUV: {replicageneration} 61d8b1ff000000010000
[07/Jan/2022:22:34:55.123860710 +0100] - DEBUG - NSMMReplicationPlugin - database RUV: {replica 1 ldap://linux.home:39001}

==> So the database RUV was recreated empty (with another generation number by the import) 
 which is clearly wrong as it breaks the replication.
3) while running the test on supplier1 in lmdb mode:
[07/Jan/2022:22:35:58.666675869 +0100] - INFO - dbmdb_public_dbmdb_import_main - import userRoot: Import complete.  Processed 20 entries in 1 seconds. (20.00 entries/sec)
...
[07/Jan/2022:22:36:02.388680247 +0100] - DEBUG - NSMMReplicationPlugin - changelog program - cldb_SetReplicaDB: cldb is set
[07/Jan/2022:22:36:02.430453701 +0100] - DEBUG - NSMMReplicationPlugin - changelog max RUV: {replicageneration} 61d8b229000000010000
[07/Jan/2022:22:36:02.433352533 +0100] - DEBUG - NSMMReplicationPlugin - database RUV: {replicageneration} 61d8b229000000010000
[07/Jan/2022:22:36:02.436294638 +0100] - DEBUG - NSMMReplicationPlugin - database RUV: {replica 1 ldap://linux.home:39001} 61d8b22a000000010000 61d8b231000100010000 00000000
[07/Jan/2022:22:36:02.439217513 +0100] - DEBUG - NSMMReplicationPlugin - database RUV: {replica 2 ldap://linux.home:39002} 61d8b22e000100020000 61d8b232000100020000 00000000
...
[07/Jan/2022:22:36:13.797250968 +0100] - ERR - agmt="cn=002" (linux:39002) - clcache_load_buffer - Can't locate CSN 61d8b231000100010000 in the changelog (DB rc=-12797). If replication stops, the consumer may need to be reinitialized.
[07/Jan/2022:22:36:13.800517424 +0100] - DEBUG - agmt="cn=002" (linux:39002) - clcache_load_buffer - rc=-12797
[07/Jan/2022:22:36:13.803327610 +0100] - DEBUG - NSMMReplicationPlugin - changelog program - _cl5CheckMissingCSN - the change with 61d8b231000100010000 csn was never logged because it was imported during replica initialization

 ==> This time the database RUV is rightly preserved but replication fails because thes supllier1 maxcsn is not in the changelog. 
 (either the mincsn should have the same value to ignore the error (to be confirmed) or some dummy changelog records should exists in the changelog 

 - Platform: Fedora 34 but probably all platforms are impacted. 
 - Package and version: master branch but IMHO all versions are impacted

**Steps to Reproduce**
1. Go to 389-ds-base/dirsrvtests/tests/suites/replication
2. Run DEBUGGING=1 qtst389 -vvvvv -x $PWD -k test_online_reinit_may_hang
3. Look at suppllier1 error log.

**Expected results**
Test should test that replication is still working after the import
Test should be PASS both on bdb and lmdb

**Additional context**
With buggy code the test is pass in bdb case: this is due to the fact that 
- the test does not check that replication is still working after the import 
- and to the fact that database RUV is wrongly reinitialized

progier389:
Missing CSN Issue: 
  replica_check_for_data_reload calls replica_log_ruv_elements (that creates the dummy changelog starting entries) when erasing the changelog or if the changelog RUV is missing
  But the issue is that the changelog RUV is not missing when the changelog is empty because _cl5ReadRUV creates a new one (by callng _cl5ConstructRUV)  if the RUV entry is not found in changelog.

 Solution is to check if changelog ruv is empty by using ruv_replica_count
    in replica_check_for_data_reload:
    if (upper_bound_ruv && ruv_replica_count(upper_bound_ruv) > 0) {

The missing csn issue is hidden in bdb case because of a second problem: 
   RUV defined before the suffix get skipped.
In fact this case was already coded in bdb but the test to handle it was buggy (entry dn was compared to RUV rdn instead of entry rdn)

reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -845,12 +845,14 @@ def test_online_reinit_may_hang(topo_with_sigkill):
         1. Export the database
         2. Move RUV entry to the top in the ldif file
         3. Import the ldif file
-        4. Online replica initializaton
+        4. Check that replication is still working
+        5. Online replica initializaton
     :expectedresults:
         1. Ldif file should be created successfully
         2. RUV entry should be on top in the ldif file
         3. Import should be successful
-        4. Server should not hang and consume 100% CPU
+        4. Replication should work
+        5. Server should not hang and consume 100% CPU
     """
     M1 = topo_with_sigkill.ms["supplier1"]
     M2 = topo_with_sigkill.ms["supplier2"]
@@ -863,6 +865,11 @@ def test_online_reinit_may_hang(topo_with_sigkill):
     M1.ldif2db(DEFAULT_BENAME, None, None, None, ldif_file)
     M1.start()
     # After this server may hang
+    # Exporting idle server with replication data and reimporting
+    # should not break replication (Unless we hit issue 5098)
+    # So let check that replication is still working.
+    repl = ReplicationManager(DEFAULT_SUFFIX)
+    repl.test_replication_topology(topo_with_sigkill)
     agmt = Agreements(M1).list()[0]
     agmt.begin_reinit()
     (done, error) = agmt.wait_reinit()
--- a/ldap/servers/plugins/replication/repl5_replica.c
+++ b/ldap/servers/plugins/replication/repl5_replica.c
@@ -1714,7 +1714,7 @@ replica_check_for_data_reload(Replica *r, void *arg __attribute__((unused)))
             return -1;
         }
 
-        if (upper_bound_ruv) {
+        if (upper_bound_ruv && ruv_replica_count(upper_bound_ruv) > 0) {
             ruv_obj = replica_get_ruv(r);
             r_ruv = object_get_data(ruv_obj);
             PR_ASSERT(r_ruv);
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_import_threads.c
@@ -2576,7 +2576,7 @@ bdb_import_foreman(void *param)
                  */
 #define RUVRDN SLAPI_ATTR_UNIQUEID "=" RUV_STORAGE_ENTRY_UNIQUEID
                 if (!slapi_be_issuffix(inst->inst_be, backentry_get_sdn(fi->entry)) &&
-                    strcasecmp(backentry_get_ndn(fi->entry), RUVRDN) /* NOT nsuniqueid=ffffffff-... */) {
+                    strcasecmp(slapi_entry_get_nrdn_const(fi->entry->ep_entry), RUVRDN) /* NOT nsuniqueid=ffffffff-... */) {
                     import_log_notice(job, SLAPI_LOG_WARNING, "bdb_import_foreman",
                                       "Skipping entry \"%s\" which has no parent, ending at line %d "
                                       "of file \"%s\"",


progier389:
Still working on the bdb RUV skipped issue.

progier389:
Waiting until https://github.com/389ds/389-ds-base/pull/5071 get merged to be able to test on mdb

tbordaz (commented):
The fix looks good to me but to be honest I do not fully understand the description when it occurs.

cldb_SetReplicaDB/_cl5ReadRUV are updating the in memory cldb  RUVs (purge and maxcsn) from the replica. While replica_log_ruv_elements set the START_ITERATION_ENTRY on database cldb. So I miss how those functions are stepping on each others.

progier389:
@tbordaz Not much to say on this area: (the changelog RUV is read before its consistency with the backend RUV is tested.
  i.e: _cl5ReadRUV is called before replica_check_for_data_reload (that calls replica_log_ruv_elements) get called.
If replica_check_for_data_reload detects, changelog is emptied and replica_log_ruv_elements is called (which is OK)
But if there is no entry in the changelog, _cl5ReadRUV has created an empty RUV (with generation ID but no replica) 
and replica_check_for_data_reload was (wrongly) checking that RUV was NULL (rather than having a RUV without replica) so replica_log_ruv_elements was not called in that case.

tbordaz (approved):
Thanks for the explanations, the fix makes sense ACK.

I was initially thinking that the problematic _cl5ReadRUV was called https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/replication/repl5_replica.c#L1642 and replica_log_ruv_elements the line after. That confused me.

PR #5112, opened 2022-01-17
Issue 4299 - UI - Add ACI editing features

mreynolds389:
Description:  Add ACI management features to UI


mreynolds389:
Added modal to create ACI manually...

droideck:
I'm not sure if it's this PR issue but still, it'll be good to fix:
If there is no suffix and no backend, the Users & Groups tab goes into infinite loading.

droideck (changes requested):
Generally, I've encountered no issues with the ACI adding/editing/removing but I was able to crash UI once...

I've added the ACI through the wizard with the choices like that:

```
(target="ldap:///uid=demo_user,ou=people,dc=example,dc=com")(targetattr="description")(version 3.0; acl "NewACI"; allow(read) userdn="ldap:///uid=demo_user,ou=people,dc=example,dc=com";)
```

But UI also crashes if I add the above ACI manually.

P.S. I know it doesn't make much sense but it may show the existing issue with more valid ACIs

mreynolds389:
> ```
> (target="ldap:///uid=demo_user,ou=people,dc=example,dc=com")(targetattr="description")(version 3.0; acl "NewACI"; allow(read) userdn="ldap:///uid=demo_user,ou=people,dc=example,dc=com";)
> ```

This works for me.  No crash.  What is the console error you are getting?

droideck:
> > ```
> > (target="ldap:///uid=demo_user,ou=people,dc=example,dc=com")(targetattr="description")(version 3.0; acl "NewACI"; allow(read) userdn="ldap:///uid=demo_user,ou=people,dc=example,dc=com";)
> > ```
> 
> This works for me. No crash. What is the console error you are getting?

Oh, sorry, I forgot to mention, I try to add the ACI to `uid=demo_user,ou=people,dc=example,dc=com`
```
Uncaught TypeError: Cannot read properties of undefined (reading 'property')
    at Object.assign.row.reduce.secretTableRowKeyId (index.js:34020:75)
    at Array.reduce (<anonymous>)
    at ContextBody.mapCells (index.js:33997:36)
    at index.js:34039:119
    at Array.map (<anonymous>)
    at ContextBody.render (index.js:34039:31)
    at finishClassComponent (index.js:163638:31)
    at updateClassComponent (index.js:163588:24)
    at beginWork (index.js:165226:16)
    at HTMLUnknownElement.callCallback (index.js:150098:14)
```

mreynolds389:
@droideck - fixed the first round of issues.  Anything else?

mreynolds389:
> @droideck - fixed the first round of issues. Anything else?

Actually I want to add something else.  When there is no suffix data the table view presents an option to create the top suffix entry (although it's not actually implemented).  So I want to make it work, and add the same thing to the tree view when there is suffix data.

droideck:
When I try to add an invalid ACI through ACI Wizard, the browser crashes.
For example, I try to set a wrong filter **(wron&&&)** - but only through the Wizard (manually - it fails successfully):

```
(target="ldap:///cn=demo_group,ou=groups,dc=example,dc=com")(targetattr="description")(targetfilter="(wron&&&)")(version 3.0; acl "New ACI"; allow(read,compare,search,write) groupdn="ldap:///cn=demo_group,ou=groups,dc=example,dc=com";)
```

Then I have this error:

```
aciNew.jsx:1335 Uncaught ReferenceError: CardTitle is not defined
    at AddNewAci.render (index.js:67903:70)
    at finishClassComponent (index.js:163638:31)
    at updateClassComponent (index.js:163588:24)
    at beginWork (index.js:165226:16)
    at HTMLUnknownElement.callCallback (index.js:150098:14)
    at Object.invokeGuardedCallbackDev (index.js:150147:16)
    at invokeGuardedCallback (index.js:150209:31)
    at beginWork$1 (index.js:170112:7)
    at performUnitOfWork (index.js:168927:12)
    at workLoopSync (index.js:168855:5)
```

droideck:
Another thing I was thinking about, - if we need a **"target"** field somewhere?

If I haven't missed it, I see that it currently sets **'target="ldap:///DN_WHERE_ADD_YOUR_ACI"'**.
But we can set **targetfilter** without the **target**. Or we can set wildcards like **target="ldap:///cn=*,dc=example,dc=com"**.

I think it makes sense to add an additional step to the ACI Wizard where the user can set it (by default, it will be the DN as it's now, but it can be removed as it's an optional field).

What do you think? Or do I miss something, and it's already like this in the current UI somewhere?

mreynolds389:
> Another thing I was thinking about, - if we need a **"target"** field somewhere?
> 
> If I haven't missed it, I see that it currently sets **'target="ldap:///DN_WHERE_ADD_YOUR_ACI"'**. But we can set **targetfilter** without the **target**. Or we can set wildcards like **target="ldap:///cn=*,dc=example,dc=com"**.

What do you mean you can set a targetfilter without a target?  Do you mean you don't need target keyword set?  This is true, it will use the entry the aci is located at.  But I'm trying to make the ACI editor a common denominator wizard.  Trying to keep it simple, and a little opinionated, since so many customers have a hard time with ACI's.  That's why I don't want to expose too many options because I feel it will get confusing if we try and add all the aci options...  More on this below...

> 
> I think it makes sense to add an additional step to the ACI Wizard where the user can set it (by default, it will be the DN as it's now, but it can be removed as it's an optional field).
> 
> What do you think? Or do I miss something, and it's already like this in the current UI somewhere?

I thought of this but it makes the UI much more complicated to allow these various wildcards and other options.  Which is one of the reasons I added the manual tweaking steps at the end so you can fine tune an ACI.  So if a customer wants an advanced specialized ACI, then they probably know what they are doing and they can manually edit the aci on the last step before adding it, or skip the wizard and add the entire ACI manually.

I hear what you are saying though, I'm just worried if I provide too many options customers will get confused and possibly create "incorrect" aci's.  In most customer cases I've seen customers use really simple ACI's (except for IDM, which uses macro ACI's which I completely left out of the editor).  

Anyway I will add the target as a settable value, but this is a slippery slope where I don't want to start adding other more complicated scenarios into the wizard.  The wizard is just there to help Admins who want to quickly create basic ACI's.  Trust me there are really complicated ways to create ACI's like using booleans and groupings, etc.  But we can't get that all into the editor and it still be useful and easy to use :-)

droideck:
@mreynolds389, sounds good!

Yeah, I agree, we shouldn't add all of the ACI options and combinations to the UI Wizard. It'll make it crazy complex and the UI benefit of simplicity will be gone.

mreynolds389:
@droideck - ok added option to change target, fixed crash, and got the empty suffix stuff all working.  Please review...

droideck:
Looks awesome!

I have one more thought. In the ACI Wizard, can we bind the field states and the resulting constructed ACI?
I'll explain.
Currently, it works like this, if I understood correctly:

1. User goes through all of the Wizard steps one at a time;
2. On Review & Edit step, he clicks 'Add ACI'
3. If there's an error -> the error is displayed, and the user can fix the issues (nice!)
4. User can edit it here, or he can go back to the previous steps and change it there (because of the error or because he just decided that he wants something else)
5. When the value is changed in the previous step, the user click "Next" button and the change is applied to the "Review & Edit" field;
6. Then the user "jumps" between the steps, fixes it and goes to the result once again

The issue I see here is that naturally, the UI allows you to "jump" between steps, but you still need to click on the "Next" button if you want to edit it. I think it's a bit less intuitive as the interface doesn't restrict from "jumping" back and forth. And the interface doesn't notify that without "Next" button clicked, it won't update the result.

So I think we should restrict the editing so it's impossible to "jump" between steps without consciously going through all of the steps.

Or we can "bind" all of the fields so when one of them is updated -> the final result is updated too. So the final ACI is always up to date.

mreynolds389:
@droideck - yeah I disabled the back button once you manually edit the aci (unless you undo the changes), but I didn't stop the "jump".  Nice catch.  I'll get that fixed :-)

mreynolds389:
@droideck - changes made...

jchapma (commented):
@mreynolds389 
Had a good play around with this, and had no issues with it. This is a very nice feature. LGTM

droideck (approved):
Looks great! Thank you!

PR #5114, opened 2022-01-18
Issue #5113 - Increase timestamp precision for development builds

### LINKED ISSUE #5113: Increase timestamp precision for development builds
**Issue Description**
We use `%Y%m%d` timestamp format for development and copr builds. Copr builds are done on each commit. If multiple commits were done during the same day, packages would have the same version.
Additionally, git is missing in the buildroot, this makes prerel version to be empty and copr builds do not contain commit hash information in their verison.

vashirov:
Bug Description:

We use %Y%m%d timestamp format for development and copr builds.
Copr builds are done on each commit. If multiple commits were done during
the same day, packages would have the same version.

Additionally, git is missing in the buildroot, this makes prerel version
to be empty and copr builds do not contain commit hash information in their
verison.

Fix Description:

* Change timestamp format to %Y%m%d%H%M.
* Modify rpm.mk to evaluate version information only once, instead of
  calling shell script every time and get different results.
* Update copr Makefile to install git as a dependency.

Fixes: https://github.com/389ds/389-ds-base/issues/5113

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,4 +1,6 @@
 srpm:
+	# Install git in the buildroot to correctly generate commit hash
+	dnf install -y git
 	# Generate spec file
 	make -f rpm.mk rpmroot
 	# Install build dependencies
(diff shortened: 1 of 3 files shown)


vashirov:
Copr build works: https://download.copr.fedorainfracloud.org/results/vashirov/389ds/fedora-35-x86_64/03192942-389-ds-base/
But `make -f rpm.mk rpms` failed, investigating.

mreynolds389 (approved):
LGTM

PR #5130, opened 2022-01-24
Issue 5129 - BUG - Incorrect fn signature in add_index

### LINKED ISSUE #5129: Backend add index may fail as no MRs are specified
```
> DEBUG: updated dn: cn=displayName,cn=index,cn=userroot,cn=ldbm database,cn=plugi
> ns,cn=config with {'cn': [b'displayName'], 'nsSystemIndex': [b'False'], 'nsIndex
> Type': [b'pres'], 'nsMatchingRule': []}
> DEBUG: {'desc': 'Protocol error'}
> Traceback (most recent call last):
>   File "/usr/sbin/openldap_to_ds", line 250, in <module>
>     result = do_migration(inst, log, args, skip_overlays)
>   File "/usr/sbin/openldap_to_ds", line 209, in do_migration
>     migration.execute_plan(log)
>   File "/usr/lib/python3.6/site-packages/lib389/migrate/plan.py", line 656, in e
> xecute_plan
>     item.apply(self.inst)
>   File "/usr/lib/python3.6/site-packages/lib389/migrate/plan.py", line 77, in ap
> ply
>     be.add_index(self.attr, self.type)
...
```

This is due to incorrect fn signature in backend.py

Firstyear:
Bug Description: Due to an incorrect function signature,
it was possible to cause add index to fail by trying to
add an empty mr set.

Fix Description: Fix the function signature and make
the function more robust.

fixes: https://github.com/389ds/389-ds-base/issues/5129

Review by: ???

### DIFF
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -719,7 +719,7 @@ class Backend(DSLdapObject):
                 return
         raise ValueError("Can not delete index because it does not exist")
 
-    def add_index(self, attr_name, types, matching_rules=[], reindex=False):
+    def add_index(self, attr_name, types, matching_rules=None, reindex=False):
         """ Add an index.
 
         :param attr_name - name of the attribute to index
@@ -736,7 +736,9 @@ class Backend(DSLdapObject):
             mrs = []
             for mr in matching_rules:
                 mrs.append(mr)
-            props['nsMatchingRule'] = mrs
+            # Only add if there are actually rules present in the list.
+            if len(mrs) > 0:
+                props['nsMatchingRule'] = mrs
         new_index.create(properties=props, basedn="cn=index," + self._dn)
 
         if reindex:


mreynolds389 (commented):
In the CLI it definitely treats the matching rules as a list:

lib389/cli_conf/backend.py:    be.add_index(args.attr, args.index_type, args.matching_rule, reindex=args.reindex)
...
lib389/cli_conf/backend.py:    add_index_parser.add_argument('--matching-rule', action='append', help='Sets the matching rule for the index')

Why change the function params from list ([]) to None, and then still treat it as a list?  Looks like the default value should still be a list and not None.

Firstyear:
@mreynolds389 Since the fn sig is "mrs=[]" then `if matching_rules is not None:` is always true since [] is not None. This means we try to add an empty list during the create which causes ldap to reject the op.

mreynolds389:
> @mreynolds389 Since the fn sig is "mrs=[]" then `if matching_rules is not None:` is always true since [] is not None. This means we try to add an empty list during the create which causes ldap to reject the op.

You are right, I did a poor review of the fix and did not look at the raw source.

mreynolds389 (approved):
LGTM

Firstyear:
```
   8a0dc0c19..72eb93ac9  389-ds-base-2.0 -> 389-ds-base-2.0
   e2bebb49d..7b7167420  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
```

PR #5139, opened 2022-01-25
Issue 4678 - Added test cases

aadhikar:
Bug Description: Added test cases to check nsslapd-ignore-virtual-attrs
is set back to ON when cos/roles are deleted. Added test cases to check if
it works in replication setup.

Relates: https://github.com/389ds/389-ds-base/issues/4678

Reviewed by: ??

### DIFF
--- a/dirsrvtests/tests/suites/config/config_test.py
+++ b/dirsrvtests/tests/suites/config/config_test.py
@@ -352,6 +352,7 @@ def test_ignore_virtual_attrs(topo):
     """Test nsslapd-ignore-virtual-attrs configuration attribute
 
     :id: 9915d71b-2c71-4ac0-91d7-92655d53541b
+    :customerscenario: True
     :setup: Standalone instance
     :steps:
          1. Check the attribute nsslapd-ignore-virtual-attrs is present in cn=config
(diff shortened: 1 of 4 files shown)


aadhikar:
not sure we need all roles in replication scenario, can remove the test case.

dirsrvtests/tests/suites/replication/virtual_attribute_test.py
```diff
@@ -0,0 +1,214 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
```
droideck: Please, change the year to 2022 :)

droideck (approved):
The rest looks good! Ack

dirsrvtests/tests/suites/replication/virtual_attribute_test.py
aadhikar: oops! still live in 2021 huh! fixing it right away!

sgouvern (commented):
could you please add the :customerscenario: True docstring to these tests please ?
Appart from that, it looks good !

sgouvern (approved):
Thanks ! Ack from me

PR #5141, opened 2022-01-27
ds_selinux_restorecon.sh: always exit 0

AdamWill:
We don't want to error out and give up on starting the service
if the restorecon fails - it might just be that the directory
doesn't exist and doesn't need restoring. Issue identified and
fix suggested by Simon Farnsworth.

https://bugzilla.redhat.com/show_bug.cgi?id=2047323

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/wrappers/ds_selinux_restorecon.sh.in
+++ b/wrappers/ds_selinux_restorecon.sh.in
@@ -29,5 +29,6 @@ then
     exit 0
 fi
 
-# Now run restorecon
-restorecon ${DS_HOME_DIR}
+# Now run restorecon, but don't die if it fails (could be that the
+# directory doesn't exist)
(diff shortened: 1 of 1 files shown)


mreynolds389:
@AdamWill Can you change the git commit message to:

```
Issue 5127 - ds_selinux_restorecon.sh: always exit 0

Description: 

We don't want to error out and give up on starting the service
if the restorecon fails - it might just be that the directory
doesn't exist and doesn't need restoring

relates: https://github.com/389ds/389-ds-base/issues/5127

Reviewed by: adamw & mreynolds
```

mreynolds389 (approved):
LGTM

PR #5143, opened 2022-01-28
Issue 2790 - RFE - set db home directory to /dev/shm by default

bsimonova:
Description:
Added set of test cases to check if db_home_dir is set
to /dev/shm/slapd-instance/ by default and the contents
of the directory.

Relates: https://github.com/389ds/389-ds-base/issues/2790

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/setup_ds/db_home_test.py
+++ b/dirsrvtests/tests/suites/setup_ds/db_home_test.py
@@ -1,0 +1,205 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+
+import pytest
+
+from lib389.utils import *
+from lib389.topologies import topology_st as topo
+from lib389._constants import *
+from lib389.dseldif import *
+from lib389.cli_conf.backend import *
+from lib389.config import BDB_LDBMConfig
+from .... conftest import get_rpm_version
+from lib389.paths import DEFAULTS_PATH
+
+# Check if we are in a container
+container_result = subprocess.run(["systemd-detect-virt", "-c"], stdout=subprocess.PIPE)
+
+pytestmark = [pytest.mark.tier1,
+              pytest.mark.skipif(get_rpm_version("selinux-policy") <= "3.14.3-79" or
+                                 get_rpm_version("selinux-policy") <= "34.1.19-1",
+                                 reason="Will fail because of incorrect selinux labels"),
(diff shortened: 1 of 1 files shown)


progier389:
Beware that the new test case failed (See this PR checks result) !

bsimonova:
@progier389 
Thanks, I am investigating the issue.
It seems the dbhome directory is created in the old location on F35 but I had no issues when I developed these tests on F34 so this escaped me.

bsimonova:
Strange, I reserved a F35 with 389-ds-base: 2.0.12-1.fc35. It passed.
So I build the package manually like here and tested on same machine and it also passed.
Have to investigate more.

dirsrvtests/tests/suites/setup_ds/db_home_test.py
```diff
@@ -0,0 +1,202 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
```
vashirov: 2022

dirsrvtests/tests/suites/setup_ds/db_home_test.py
```diff
@@ -0,0 +1,202 @@
...
+    :steps:
+        1. Create instance
+        2. Check nsslapd-db-home-directory is set to /dev/shm/slapd-instance in cn=config
+    :expectedresults:
+        1. Success
+        2. Success
+    """
+
+    standalone = topo.standalone
+    dbhome_value = '/dev/shm/slapd-{}'.format(standalone.serverid)
```
vashirov, line 56: In a container it won't be in /dev/shm, it falls back to /var/lib/...
So perhaps this test module should be skipped when it runs in a container?

vashirov (changes requested):
Failures are because in PR CI we run tests in a container. Lib389 detects that and falls back to `/var/lib` instead of `/dev/shm`:
https://github.com/389ds/389-ds-base/blob/af20dce9eba1059eb42c7c45d4e36a071e358f0b/src/lib389/lib389/instance/setup.py#L736-L739

dirsrvtests/tests/suites/setup_ds/db_home_test.py
bsimonova: Thanks, I added another skipif for container.
I build latest package and retested (all passing).
Please review.

vashirov: It still fails in PR CI. The test is not skipped, because systemd is present, even though it is running in a container. See https://github.com/389ds/389-ds-base/blob/af20dce9eba1059eb42c7c45d4e36a071e358f0b/src/lib389/lib389/instance/setup.py#L736-L743
on how to detect if the test is running in a container.

vashirov (approved):

PR #5144, opened 2022-01-31
Issue 4299 - UI - fix minor issues with ldap editor

mreynolds389:
Description:  Improved how treeview handles loading subtrees with large
number of entries.  Previously, the parent entry would not be
displayed while loading its child entries, and if a timeout occurred
then the paretn entry would not be loading inthe UI, nad you could not
do or see anything with it.

In the search form, the dropdown list showing the backend was not updated 
when multiple backends are available, it will now update to the correct suffix 
when "Search" is selected on a subtree from the tree/table view.

Also added a pop modal when an error occurs when searching the database.

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/LDAPEditor.jsx
+++ b/src/cockpit/389-console/src/LDAPEditor.jsx
@@ -209,7 +209,8 @@ export class LDAPEditor extends React.Component {
                 const params = {
                     serverId: this.props.serverId,
                     baseDn: suffixDN,
-                    parentId: parentId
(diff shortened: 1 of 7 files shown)


droideck (commented):
I have one issue which I think can be related to this PR:

Create a standard instance with the default sample directory tree.
In the tree view, click quickly on ou=permissions and ou=groups.
ou=permissions loading hangs indefinitely.

droideck (approved):
Thanks!
Ack

PR #5146, opened 2022-02-01
Issue 4299 - UI - fix minor issues with ldap editor (table view)

mreynolds389:
Description:

Imrpove th4e react handling of the table view while searching is going
on.  Also adjusted the default size limit to 2000.  The search was also
not doing any notifications if an error happened.

### DIFF
--- a/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
+++ b/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
@@ -238,6 +238,8 @@ export function getSearchEntries (params, resultCallback) {
     params.searchScope,
     '-l',
     params.timeLimit,
+    '-z',
+    params.sizeLimit,
     params.searchFilter,
     '*',
     '+'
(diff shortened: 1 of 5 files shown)


src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
```diff
@@ -258,7 +260,10 @@ export function getSearchEntries (params, resultCallback) {
       if (err.exit_status === 4) {
         console.log('Size limit hit'); // Use the partial data.
         searchResult = data;
-        // TODO: Check other relevant error codes ( 32, 53 ...)
+        params.addNotification(
+            "info",
+            `Size limit of ${params.sizeLimit} was exceeded.  The child entries of "${params.searchBase}" have been truncated.`
```
droideck, line 265: I think it makes sense to mention in the help message that "Full list of entries can be accessed through search or table view". Otherwise, I think, we can hear a lot of questions like - "Can we configure the limit?", "Is it a planned feature to expand the tree view", etc.
So a more detailed info message will help here, IMHO.

droideck (approved):
Otherwise, LGMT!

src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
mreynolds389: Actually it can only be done through the search page.  Table and tree both use the hardcoded size limit

mreynolds389: I'll update the other message when the search is done through the the tree/table.

mreynolds389: changes made, merging....

Issue #5142, opened 2022-01-27
dsctl dbgen is broken


mreynolds389:
**Issue Description**

There was a regression somewhere that removed the userid from the DirSrv object:

```
dsctl -v localhost ldifgen users --number 1010 --parent ou=people,dc=example,dc=com --suffix dc=example,dc=com
DEBUG: The 389 Directory Server Administration Tool
DEBUG: Inspired by works of: ITS, The University of Adelaide
DEBUG: Called with: Namespace(verbose=True, instance='localhost', json=False, list=False, remove_all=False, number='1010', suffix='dc=example,dc=com', parent='ou=people,dc=example,dc=com', generic=False, start_idx=0, rdn_cn=False, localize=False, ldif_file='users.ldif', func=<function dbgen_create_users at 0x7fae5eec4a60>)
DEBUG: Allocate local instance <class 'lib389.DirSrv'> with None
DEBUG: Instance allocated
INFO: 
Generating LDIF with the following options:
INFO:  - number=1010
INFO:  - suffix=dc=example,dc=com
INFO:  - parent=ou=people,dc=example,dc=com
INFO:  - generic=False
INFO:  - start-idx=0
INFO:  - rdn-cn=False
INFO:  - localize=False
INFO:  - ldif-file=/var/lib/dirsrv/slapd-localhost/ldif/users.ldif
INFO: 
Writing LDIF ...
DEBUG: DirSrv has no attribute 'userid'
Traceback (most recent call last):
  File "/usr/sbin/dsctl", line 162, in <module>
    result = args.func(inst, log, args)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/cli_ctl/dbgen.py", line 195, in dbgen_create_users
    dbgen_users(inst, args.number, args.ldif_file, args.suffix, generic=args.generic, parent=args.parent, startIdx=args.start_idx, rdnCN=False, pseudol10n=args.localize)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/dbgen.py", line 296, in dbgen_users
    finalize_ldif_file(instance, ldif_file)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/dbgen.py", line 135, in finalize_ldif_file
    userid = ensure_str(instance.userid)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 362, in __getattr__
    raise AttributeError('%s has no attribute %s' % (
[...]

PR #5152, opened 2022-02-03
Issue 5142 - CLI - dsctl dbgen is broken

mreynolds389:
Description:

Changes to dsctl broke dbgen which requires instance.userid to
set the permissions of the ldif file. It occurred when we added:
local_simple_allocate(). The fix is add userid in this allocate
function.

relates: https://github.com/389ds/389-ds-base/issues/5142


progier389 (approved):
LGTM

Issue #5126, opened 2022-01-20
Memory leak in slapi_ldap_get_lderrno


progier389:
**Issue Description**
Clone of https://bugzilla.redhat.com/show_bug.cgi?id=2032742

The leak is due to a change in libldap API:
Digging in the old man page:
Note  LDAP_OPT_ERROR_STRING returns a pointer to an internal static string table, and ldap_memfree should not be called when using this session option.
And the new one:
LDAP_OPT_DIAGNOSTIC_MESSAGE
Sets/gets a string containing the error string associated to the LDAP handle. This option was formerly known as LDAP_OPT_ERROR_STRING. outvalue must be a char **, and the caller is responsible of freeing the returned string by calling ldap_memfree(3), while invalue must be a char *; the library duplicates the corresponding string.

So it looks like we will have to review the whole code for all slapi_ldap_get_lderrno calls:
  if the call has a message address it should be freed after use  by a call to a new slapi_ldap_free_lderrno_msg function
  that free the message if LDAP_OPT_DIAGNOSTIC_MESSAGE is defined.

progier389:
I am able to reproduce the leak (setting up replication over SASL/Digest-MD5 and exporting SASL_PATH=foo and restarting the instances then perform a change and loop poking on the repl-agmt) but the weird thing is that libasan does not report the leak even if I count more than 200000 occurence of the leaking string when looking in /proc/xxxx/mem  (Furthermore using the debbuger to set up a breakpoint on slapi_ldap_get_lderrno then following the code step by step clearly  shows that here is a leaks (string is  duped then printed and reference is lost because the variable get out of scope ... )

mreynolds389:
I wonder if valgrind might show the leak?  ASAN is "suppose" to be better than valgrind, but this wouldn't be the first time ASAN is not reporting on things it should.

progier389:
Indeed valgrind is slower but it detects the leaks:

==1987883== 15,848 bytes in 283 blocks are definitely lost in loss record 826 of 849
==1987883==    at 0x484086F: malloc (vg_replace_malloc.c:381)
==1987883==    by 0x4BA6F64: ber_memalloc_x (memory.c:228)
==1987883==    by 0x4BA9167: ber_strdup_x (memory.c:643)
==1987883==    by 0x4B7F37F: ldap_get_option (options.c:320)
==1987883==    by 0x48C8BA6: slapi_ldap_get_lderrno (ldaputil.c:381)
[...]

PR #5153, opened 2022-02-03
Issue 5126 - Memory leak in slapi_ldap_get_lderrno

progier389:
The problem is that some time ago libldap API replaced  ​LDAP_OPT_ERROR_STRING whose data should not be freed by
LDAP_OPT_DIAGNOSTIC_MESSAGE whose data must be freed.
slapi_ldap_get_lderrno was adapted to use the new option but the callers were not modified to free the value.

The Solution:
 Insure that we also need to free slapi_ldap_get_lderrno value if legacy LDAP_OPT_ERROR_STRING is used (by duping the value)
 Insure that the callers free the value.

Added test case about replication using SASL/Digest-md5 authentication
Added test case to check this leak 
Also updated test case about SASL/GSSAPI to be comapatible with current lib389 framework but marked as skipped because it requires a specific configuration (This path should be tested by IPA tests)
Fixed valgrind lib389 function to run on prefixed installation without needing to be root.
At last I also improved lib389 mapped object to have a better diagnostic when LDAP operation fails (by adding the request within the exception)
 
 issue: [5126](https://github.com/389ds/389-ds-base/issues/5126)

Reviewd by: @droideck

### DIFF
--- a/dirsrvtests/tests/suites/replication/sasl_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/sasl_m2_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2017 Red Hat, Inc.
+# Copyright (C) 2022 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
@@ -91,6 +91,8 @@ def use_valgrind(topo_m2, request):
     set_sasl_md5_client_auth(m2, m1)
     m1.stop()
     m2.stop()
+    m1.systemd_override = False
+    m2.systemd_override = False
     valgrind_enable(m1.ds_paths.sbin_dir, gen_valgrind_wrapper(m1.ds_paths.sbin_dir))
 
     def fin():
@@ -128,6 +130,7 @@ def test_repl_sasl_md5_auth(topo_m2):
     repl.test_replication_topology(topo_m2)
 
 
+@pytest.mark.skipif(not os.path.exists('/usr/bin/valgrind'), reason="valgrind is not installed.")
 def test_repl_sasl_leak(topo_m2, use_valgrind):
     """Test replication with SASL digest-md5 authentication
 
--- a/dirsrvtests/tests/suites/replication/sasl_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/sasl_m2_test.py
@@ -1,0 +1,182 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2017 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import os
+import pytest
+import ldap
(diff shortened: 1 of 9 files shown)


progier389:
Looks like there is a problem with the new test (it fails in CI tests although it works in my environment)
Guess I will have to run it on a 1minutetips instance

dirsrvtests/tests/suites/replication/sasl_m2_test.py
```diff
@@ -0,0 +1,185 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2017 Red Hat, Inc.
```
droideck: Wrong date :)

src/lib389/lib389/utils.py
```diff
@@ -583,7 +600,14 @@ def valgrind_disable(sbin_dir):
...
 
     # Enable selinux
-    os.system('setenforce 1')
+    if os.geteuid() == 0:
+        os.system('setenforce 1')
+
+    # Restore .inf file (for systemd)
+    new_path = Paths()._get_defaults_loc(DEFAULTS_PATH)
+    old_path = f'{new_path}.orig'
+    if os.exists(old_path):
```
droideck: Should be `os.path.exists()`, I think.

droideck (changes requested):
@progier389, could you please run 'dirsrvtests/tests/suites/gssapi_repl/gssapi_repl_test.py' with the kerberos setup? Or, may be, @sgouvern, you can direct me how you run these tests?
   When I try to run, it says `Test disabled because it requires specific kerberos requirement (server principal, keytab, etc ...`

@Firstyear could you please check the `lib389/_mapped_object` change? You may be interested in the change that @progier389 made :)

progier389:
> @progier389, could you please run 'dirsrvtests/tests/suites/gssapi_repl/gssapi_repl_test.py' with the kerberos setup? Or, may be, @sgouvern, you can direct me how you run these tests? When I try to run, it says `Test disabled because it requires specific kerberos requirement (server principal, keytab, etc ...`
> 
> @Firstyear could you please check the `lib389/_mapped_object` change? You may be interested in the change that @progier389 made :)

@droideck,  Short answer is no. 
  That is a test that has been disabled for some time. 
  when I removed the "return that deactivated it, I got several errors at lib389 level so I decided to fix it, then I realized later on that I needed a specific kerberos environment and disabled it again with the skip decorator 
   IMHO if we want to do gssapI test we have better to run the freeIPA setup that properly set up Kerberos and DS with GSS API over Kerberos ...
   I just kept the changes to avoid having to redo it if we somehow decide to fully fix that test but that is serious work that deserve a ticket by itself 
And that is why I finally ended to test this fix by using  GSSAPI over DigestMD5.

src/lib389/lib389/utils.py
progier389: Indeed, I just wonder why the test that I run on 1 minutetips was PASS instead of getting an AttributeError exception
 (probably because it was the finalizer that was impacted .. )

droideck (approved):
I guess @Firstyear is not about to check lib389 part he might be interested in.
And the code looks good to me!
Ack :)

Issue #5155, opened 2022-02-07
RFE - Provide an option to abort an Auto Member rebuild task


mreynolds389:
Depending of the size of the groups along with the complexity of the rules and the interactions with other plugins ( typically in an IPA context ), the Auto Member rebuild task can generate performance issues ( for instance a high CPU usage when the Schema Compatibility plugin is called ).  Therefore an option to abort the rebuild task would be beneficial.

See:  https://bugzilla.redhat.com/show_bug.cgi?id=2018153

mreynolds389:
e13faa2b4..5716a22ca  389-ds-base-2.0 -> 389-ds-base-2.0
a48da4403..dca050edc  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
bacf9d80a..29a0ce9d7  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

tbordaz:
@mreynolds389  is not backported in 1.3.10 and cherry-pick looks not immediate. Could you backport it, thanks

tbordaz:
7eca61ace..78d2ef45d  389-ds-base-1.3.10

PR #5157, opened 2022-02-07
Issue 5155 - RFE - Provide an option to abort an Auto Member rebuild task

mreynolds389:
Description:  Add an abort task for the autommeber rebuild task.  There
are cases where IPA can start spinning up schema compat search during
the rebuild which can bog down an entire system.  If this happens the
task can be aborted to prevent an outage.

Also found that in cleanAllRUV we wre trying to write to the task entry
in the task add callback function (which is too early to start updating
the task and triggers error 32 messages in the errors log).  So that
was fixed as well.

relates: https://github.com/389ds/389-ds-base/issues/5155

### DIFF
--- a/ldap/servers/plugins/automember/automember.c
+++ b/ldap/servers/plugins/automember/automember.c
@@ -21,6 +21,7 @@
  */
 static PRCList *g_automember_config = NULL;
 static Slapi_RWLock *g_automember_config_lock = NULL;
+static uint64_t abort_rebuild_task = 0;
 
 static void *_PluginID = NULL;
 static Slapi_DN *_PluginDN = NULL;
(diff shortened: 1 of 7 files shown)


ldap/servers/plugins/automember/automember.c
```diff
@@ -2434,19 +2510,19 @@ automember_rebuild_task_thread(void *arg)
...
         }
     }
-    automember_config_unlock();
 
 out:
+    automember_config_unlock();
+
     if (plugin_is_betxn && fixup_pb) {
-        if (i == 0 || result != 0) { /* no updates performed */
+        if ((i == 0 || result != 0) && aborted == 0) { /* no updates performed */
```
progier389: Should abort txn if i==0 (in other words it should be: if (i==0 || (result !=0 && aborted==0))

progier389 (approved):
LGTM except a change in a condition

ldap/servers/plugins/automember/automember.c
mreynolds389: You're right, but @tbordaz mentioned that the rebuild task actually starts over from the beginning with each rebuild, so there is no real benefit to "keeping" the transaction.  The next rebuild will be just as expensive.  He also said we could change that transaction model to use smaller transactions.  Not sure if that's going to impact performance but I'm going to try it out next...

mreynolds389:
Removed transaction around fixup task (experimental).  Please review...

tbordaz (approved):
LGTM as well

droideck (approved):
Ack from me too!

PR #5163, opened 2022-02-11
Issue 4299 - UI - Add Role funtionality

droideck:
Description:  Add Role management features to UI.
Improve CLI role functionality.

Relates: #4299

Reviewed by: ?


mreynolds389:
When creating a filtered role, you are allowed to click "next" even though the filter attribute is empty.  

Nested roles has the same issue.  Also, you can proceed without choosing an existing role, is that by design?  If you must select a role then "next" should not be enabled, but if you don't an existing role then ignore this comment.

droideck:
> When creating a filtered role, you are allowed to click "next" even though the filter attribute is empty.
> 
> Nested roles has the same issue. Also, you can proceed without choosing an existing role, is that by design? If you must select a role then "next" should not be enabled, but if you don't need an existing role then ignore this comment.

Fixed.
Now, the Next button is disabled while the attribute value is empty.

It's still possible to proceed with no Roles selected in the Roles search (nested role type). But later it will block during the attribute values stage anyway (if the value is empty).

mreynolds389 (approved):
Thanks, looks good to me

PR #5164, opened 2022-02-11
Issue 5162 - BUG - error on importing chain files

### LINKED ISSUE #5162: Importing a ca chain only imports the first certificate
A ca chain pem file is just a series of pem certs concatinated. NSS of course, when presented with a chain file will only import the first certificate, and will silently ignore the rest.

This can create user confusion who expect to be able to import a chain file.

While we can't fix NSS, we CAN fix dsctl instance tls import commands such that when a chain file is presented we can inform the user they need to split it and import the components individually.

Firstyear:
Bug Description: Nss can't import pem chain files which can
confuse users why they have missing certificates when they try
to import a chain.

Fix Description: Error out on chain files in any of the import
paths since they are ambiguous.

fixes: https://github.com/389ds/389-ds-base/issues/5162

Review by: ???


droideck (approved):
LGTM!

Firstyear:
```
   9f410afa6..11adb34df  389-ds-base-2.0 -> 389-ds-base-2.0
```

Issue #4721, opened 2021-04-12
Cockpit's 'plugins' tab doesn't load if 'attribute uniqueness' configuration record is deleted


spike77453:
**Issue Description**
After deleting the plug-in configuration record 'attribute uniqueness' (att-uniq plugin) the plugins tab in cockpit's '389 Directory Server Management' tool can't be loaded

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: cockpit-389-ds-1.4.4.15-1.fc33.noarch
 - Browser: firefox

**Steps to Reproduce**
1. Create new instance
```
[general]
config_version = 2

[slapd]
root_password = mysecret

[backend-userroot]
sample_entries = yes
suffix = dc=example,dc=com
```
```dscreate from-file /root/instance.inf```

2. Run ```dsconf localhost  plugin attr-uniq delete "attribute uniqueness"```
3. Got to `<389-ds-host>:9090` and log in
4. Click on '389 Directory Server' -> 'Plugins'

**Expected results**
A properly loaded plugins tab

**Screenshots**
![Screenshot_20210412_085707](https://user-images.githubusercontent.com/25122064/114354022-fd661b80-9b6d-11eb-9079-59463103c88b.png)

**Additional context**
Problem seems to be the same on the Fedora 34 preview with cockpit-389-ds 2.0.3. Happens even if additional attr-uniq plugin configuration records are present. After re-adding a plugin configuration with the name 'attribute uniqueness' (even with different attr-name or subtree than the default) makes the plugin tab load again.

mreynolds389:
UI now crashes when deleting the last attr uniqueness configuration

mreynolds389:
5716a22ca..23f1ba21a  389-ds-base-2.0 -> 389-ds-base-2.0
dca050edc..3c37d957a  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
29a0ce9d7..3de8ec355  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5166, opened 2022-02-16
Issue 4721 - UI - attribute uniqueness crashes UI when there are no configs

mreynolds389:
Description:

The UI crashes if there are no attribute uniqueness configurations.  So
improved the robustness for plugins that only exist as separate config
entries (currently just attribute uniqueness).  Did some other plugin
cleanup with action menus and spacing.

relates: https://github.com/389ds/389-ds-base/issues/4721


droideck (approved):
LGTM!

PR #5168, opened 2022-02-21
filter optimizer

mreynolds389:
Just testing Williams old filter optimization code, this is not ready for review.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1946,6 +1946,7 @@ TESTS = test_slapd
 test_slapd_SOURCES = test/main.c \
 	test/libslapd/test.c \
 	test/libslapd/counters/atomic.c \
+	test/libslapd/filter/optimise.c \
 	test/libslapd/pblock/analytics.c \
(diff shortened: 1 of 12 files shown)


mreynolds389:
There is a regression with ACLsforn one-level scope searches.  This PR adds a parentid component to the filter (create_onelevel_filter() during build_candidate_list()), which in turn gets evaluated by the ACL code and causes regressions.

Firstyear:
Okay, I'll see if I can investigate this today.

mreynolds389:
> Okay, I'll see if I can investigate this today.

I think we just need to preserve the original filter and use that for ACI evaluation.  Because right now it adds parentId and objectclass filter components and those should be skipped from the access evaluation.

Firstyear:
Yep, tottaly agree with you on that. I was going to see if we already preserved it, if not I'll add that to the pb.

Firstyear:
See https://github.com/389ds/389-ds-base/pull/5171

PR #5171, opened 2022-02-22
Enable filter optimisation

Firstyear:
Supercedes https://github.com/389ds/389-ds-base/pull/5168

### DIFF
--- a/ldap/servers/slapd/back-ldbm/filterindex.c
+++ b/ldap/servers/slapd/back-ldbm/filterindex.c
@@ -701,6 +701,9 @@ list_candidates(
     struct berval *vpairs[2] = {NULL, NULL};
     int is_and = 0;
     IDListSet *idl_set = NULL;
+    back_search_result_set *sr = NULL;
+
+    slapi_pblock_get(pb, SLAPI_SEARCH_RESULT_SET, &sr);
 
     slapi_log_err(SLAPI_LOG_TRACE, "list_candidates", "=> 0x%x\n", ftype);
 
@@ -891,6 +894,7 @@ list_candidates(
              * and allids - we should not process anymore, and fallback to full
              * table scan at this point.
              */
+            sr->sr_flags |= SR_FLAG_MUST_APPLY_FILTER_TEST;
             goto apply_set_op;
         }
 
@@ -899,11 +903,12 @@ list_candidates(
              * If we encounter a zero length idl, we bail now because this can never
              * result in a meaningful result besides zero.
              */
+            sr->sr_flags |= SR_FLAG_MUST_APPLY_FILTER_TEST;
             goto apply_set_op;
         }
     }
 
-/*
+    /*
      * Do the idl_set operation if required.
      * these are far more efficient than the iterative union and
      * intersections we previously used.
--- a/ldap/servers/slapd/back-ldbm/ldbm_search.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_search.c
@@ -31,7 +31,7 @@
 /* prototypes */
 static int build_candidate_list(Slapi_PBlock *pb, backend *be, struct backentry *e, const char *base, int scope, int *lookup_returned_allidsp, IDList **candidates);
 static IDList *base_candidates(Slapi_PBlock *pb, struct backentry *e);
-static IDList *onelevel_candidates(Slapi_PBlock *pb, backend *be, const char *base, struct backentry *e, Slapi_Filter *filter, int managedsait, int *lookup_returned_allidsp, int *err);
+static IDList *onelevel_candidates(Slapi_PBlock *pb, backend *be, const char *base, Slapi_Filter *filter, int *lookup_returned_allidsp, int *err);
 static back_search_result_set *new_search_result_set(IDList *idl, int vlv, int lookthroughlimit);
 static void delete_search_result_set(Slapi_PBlock *pb, back_search_result_set **sr);
 static int can_skip_filter_test(Slapi_PBlock *pb, struct slapi_filter *f, int scope, IDList *idl);
@@ -922,17 +922,19 @@ ldbm_back_search(Slapi_PBlock *pb)
             tmp_desc = "Filter is not set";
             goto bail;
         }
-        if (can_skip_filter_test(pb, filter, scope, candidates)) {
-            sr->sr_flags |= SR_FLAG_CAN_SKIP_FILTER_TEST;
+        if (can_skip_filter_test(pb, filter, scope, candidates) == 0) {
+            sr->sr_flags |= SR_FLAG_MUST_APPLY_FILTER_TEST;
         }
     }
 
     /* if we need to perform the filter test, pre-digest the filter to
        speed up the filter test */
-    if (!(sr->sr_flags & SR_FLAG_CAN_SKIP_FILTER_TEST) ||
+    if ((sr->sr_flags & SR_FLAG_MUST_APPLY_FILTER_TEST) ||
         li->li_filter_bypass_check) {
         int rc = 0, filt_errs = 0;
         Slapi_Filter *filter = NULL;
+
+        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_search", "Applying Filter Test\n");
 
         slapi_pblock_get(pb, SLAPI_SEARCH_FILTER, &filter);
         if (NULL == filter) {
@@ -956,6 +958,8 @@ ldbm_back_search(Slapi_PBlock *pb)
                 tmp_desc = "Could not compile regex for filter matching";
             }
         }
+    } else {
+        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_search", "Skipped Filter Test\n");
     }
 bail:
     /* Fix for bugid #394184, SD, 05 Jul 00 */
@@ -981,8 +985,10 @@ build_candidate_list(Slapi_PBlock *pb, backend *be, struct backentry *e, const c
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
     int managedsait = 0;
     Slapi_Filter *filter = NULL;
+    Slapi_Filter *filter_exec = NULL;
     int err = 0;
     int r = 0;
+    char logbuf[1024] = {0};
 
     slapi_pblock_get(pb, SLAPI_SEARCH_FILTER, &filter);
     if (NULL == filter) {
@@ -999,13 +1005,35 @@ build_candidate_list(Slapi_PBlock *pb, backend *be, struct backentry *e, const c
         break;
 
     case LDAP_SCOPE_ONELEVEL:
-        *candidates = onelevel_candidates(pb, be, base, e, filter, managedsait,
-                                          lookup_returned_allidsp, &err);
+        /* Now optimise the filter for use */
+        slapi_filter_optimise(filter);
+        /* modify the filter to be: (&(|(originalfilter)(objectclass=referral))(parentid=idofbase)) */
+        filter_exec = create_onelevel_filter(filter, e, managedsait);
+
+        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_search", "Optimised ONE filter to - %s\n",
+             slapi_filter_to_string(filter_exec, logbuf, sizeof(logbuf)));
+
+        *candidates = onelevel_candidates(pb, be, base, filter_exec, lookup_returned_allidsp, &err);
+
+        slapi_pblock_set(pb, SLAPI_SEARCH_FILTER, filter_exec);
+        slapi_pblock_set(pb, SLAPI_SEARCH_FILTER_INTENDED, filter);
+
         break;
 
     case LDAP_SCOPE_SUBTREE:
-        *candidates = subtree_candidates(pb, be, base, e, filter, managedsait,
-                                         lookup_returned_allidsp, &err);
+        /* Now optimise the filter for use */
+        slapi_filter_optimise(filter);
+        /* make (|(originalfilter)(objectclass=referral)) */
+        filter_exec = create_subtree_filter(filter, managedsait);
+
+        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_search", "Optimised SUB filter to - %s\n",
+             slapi_filter_to_string(filter_exec, logbuf, sizeof(logbuf)));
+
+        *candidates = subtree_candidates(pb, be, base, e, filter_exec, lookup_returned_allidsp, &err);
+
+        slapi_pblock_set(pb, SLAPI_SEARCH_FILTER, filter_exec);
+        slapi_pblock_set(pb, SLAPI_SEARCH_FILTER_INTENDED, filter);
+
         break;
 
     default:
@@ -1064,15 +1092,15 @@ base_candidates(Slapi_PBlock *pb __attribute__((unused)), struct backentry *e)
  * non-recursively when done with the returned filter.
  */
 static Slapi_Filter *
-create_referral_filter(Slapi_Filter *filter, Slapi_Filter **focref, Slapi_Filter **forr)
+create_referral_filter(Slapi_Filter *filter)
 {
     char *buf = slapi_ch_strdup("objectclass=referral");
 
-    *focref = slapi_str2filter(buf);
-    *forr = slapi_filter_join(LDAP_FILTER_OR, filter, *focref);
+    Slapi_Filter *focref = slapi_str2filter(buf);
+    Slapi_Filter *forr = slapi_filter_join(LDAP_FILTER_OR, filter, focref);
 
     slapi_ch_free((void **)&buf);
-    return *forr;
+    return forr;
 }
 
 /*
@@ -1086,20 +1114,20 @@ create_referral_filter(Slapi_Filter *filter, Slapi_Filter **focref, Slapi_Filter
  * This function is exported for the VLV code to use.
  */
 Slapi_Filter *
-create_onelevel_filter(Slapi_Filter *filter, const struct backentry *baseEntry, int managedsait, Slapi_Filter **fid2kids, Slapi_Filter **focref, Slapi_Filter **fand, Slapi_Filter **forr)
+create_onelevel_filter(Slapi_Filter *filter, const struct backentry *baseEntry, int managedsait)
 {
     Slapi_Filter *ftop = filter;
     char buf[40];
 
     if (!managedsait) {
-        ftop = create_referral_filter(filter, focref, forr);
+        ftop = create_referral_filter(filter);
     }
 
     sprintf(buf, "parentid=%lu", (u_long)(baseEntry != NULL ? baseEntry->ep_id : 0));
-    *fid2kids = slapi_str2filter(buf);
-    *fand = slapi_filter_join(LDAP_FILTER_AND, ftop, *fid2kids);
+    Slapi_Filter *fid2kids = slapi_str2filter(buf);
+    Slapi_Filter *fand = slapi_filter_join(LDAP_FILTER_AND, ftop, fid2kids);
 
-    return *fand;
+    return fand;
 }
 
 /*
@@ -1110,37 +1138,15 @@ onelevel_candidates(
     Slapi_PBlock *pb,
     backend *be,
     const char *base,
-    struct backentry *e,
     Slapi_Filter *filter,
-    int managedsait,
     int *lookup_returned_allidsp,
     int *err)
 {
-    Slapi_Filter *fid2kids = NULL;
-    Slapi_Filter *focref = NULL;
-    Slapi_Filter *fand = NULL;
-    Slapi_Filter *forr = NULL;
-    Slapi_Filter *ftop = NULL;
     IDList *candidates;
 
-    /*
-     * modify the filter to be something like this:
-     *
-     *    (&(parentid=idofbase)(|(originalfilter)(objectclass=referral)))
-     */
-
-    ftop = create_onelevel_filter(filter, e, managedsait, &fid2kids, &focref, &fand, &forr);
-
-    /* from here, it's just like subtree_candidates */
-    candidates = filter_candidates(pb, be, base, ftop, NULL, 0, err);
+    candidates = filter_candidates(pb, be, base, filter, NULL, 0, err);
 
     *lookup_returned_allidsp = slapi_be_is_flag_set(be, SLAPI_BE_FLAG_DONT_BYPASS_FILTERTEST);
-
-    /* free up just the filter stuff we allocated above */
-    slapi_filter_free(fid2kids, 0);
-    slapi_filter_free(fand, 0);
-    slapi_filter_free(forr, 0);
-    slapi_filter_free(focref, 0);
 
     return (candidates);
 }
@@ -1156,12 +1162,12 @@ onelevel_candidates(
  * This function is exported for the VLV code to use.
  */
 Slapi_Filter *
-create_subtree_filter(Slapi_Filter *filter, int managedsait, Slapi_Filter **focref, Slapi_Filter **forr)
+create_subtree_filter(Slapi_Filter *filter, int managedsait)
 {
     Slapi_Filter *ftop = filter;
 
     if (!managedsait) {
-        ftop = create_referral_filter(filter, focref, forr);
+        ftop = create_referral_filter(filter);
     }
 
     return ftop;
@@ -1178,13 +1184,9 @@ subtree_candidates(
     const char *base,
     const struct backentry *e,
     Slapi_Filter *filter,
-    int managedsait,
     int *allids_before_scopingp,
     int *err)
 {
-    Slapi_Filter *focref = NULL;
-    Slapi_Filter *forr = NULL;
-    Slapi_Filter *ftop = NULL;
     IDList *candidates;
     PRBool has_tombstone_filter;
     int isroot = 0;
@@ -1193,13 +1195,8 @@ subtree_candidates(
     Operation *op = NULL;
     PRBool is_bulk_import = PR_FALSE;
 
-    /* make (|(originalfilter)(objectclass=referral)) */
-    ftop = create_subtree_filter(filter, managedsait, &focref, &forr);
-
     /* Fetch a candidate list for the original filter */
-    candidates = filter_candidates_ext(pb, be, base, ftop, NULL, 0, err, allidslimit);
-    slapi_filter_free(forr, 0);
-    slapi_filter_free(focref, 0);
+    candidates = filter_candidates_ext(pb, be, base, filter, NULL, 0, err, allidslimit);
 
     /* set 'allids before scoping' flag */
     if (NULL != allids_before_scopingp) {
@@ -1406,6 +1403,7 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
     int managedsait;
     Slapi_Attr *attr;
     Slapi_Filter *filter;
+    Slapi_Filter *filter_intent;
     back_search_result_set *sr;
     ID id;
     struct backentry *e;
@@ -1439,6 +1437,7 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
     slapi_pblock_get(pb, SLAPI_SEARCH_SCOPE, &scope);
     slapi_pblock_get(pb, SLAPI_MANAGEDSAIT, &managedsait);
     slapi_pblock_get(pb, SLAPI_SEARCH_FILTER, &filter);
+    slapi_pblock_get(pb, SLAPI_SEARCH_FILTER_INTENDED, &filter_intent);
     slapi_pblock_get(pb, SLAPI_NENTRIES, &nentries);
     slapi_pblock_get(pb, SLAPI_SEARCH_SIZELIMIT, &slimit);
     slapi_pblock_get(pb, SLAPI_SEARCH_TIMELIMIT, &tlimit);
@@ -1711,34 +1710,52 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
                         filter_test = 0;
                     }
                 }
-                if (filter_test == 0) {
-                    /* check if the entry matches the filter, and passes the ACL check */
-                    filter_test = -1;
-                    if (0 != (sr->sr_flags & SR_FLAG_CAN_SKIP_FILTER_TEST)) {
-                        /* Since we do access control checking in the filter test (?Why?) we need to check access now */
-                        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry",
-                                    "Bypassing filter test\n");
-                        if (ACL_CHECK_FLAG) {
-                            filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
-                        } else {
-                            filter_test = 0;
-                        }
-                        if (li->li_filter_bypass_check) {
-                            int ft_rc;
 
-                            slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry", "Checking bypass\n");
-                            ft_rc = slapi_vattr_filter_test(pb, e->ep_entry, filter,
-                                                            ACL_CHECK_FLAG);
-                            if (filter_test != ft_rc) {
-                                /* Oops ! This means that we thought we could bypass the filter test, but noooo... */
-                                slapi_log_err(SLAPI_LOG_ERR, "ldbm_back_next_search_entry",
-                                            "Filter bypass ERROR on entry %s\n", backentry_get_ndn(e));
-                                filter_test = ft_rc; /* Fix the error */
-                            }
-                        }
+                /* it's a regular entry, check if it matches the filter, and passes the ACL check */
+                /*
+                 * Remember, MUST_APPLY is set during a shortcut condition from the IDL backend,
+                 * which means we can NOT ignore it! When it's 0, we assume that IDL fully resolved
+                 * which means we then check the ACL only, we have a decision about if we do the
+                 * test based on the configuration.
+                 */
+                if (0 == (sr->sr_flags & SR_FLAG_MUST_APPLY_FILTER_TEST)) {
+                    /*
+                     * Since we do access control checking in the filter test we need to check access now
+                     * This checks access to the filter as INTENDED by the user - not the query that
+                     * we have messed with internally - remember, our internal changes are safe!
+                     */
+                    slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry",
+                                  "Bypassing filter test\n");
+                    if (ACL_CHECK_FLAG) {
+                        filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
                     } else {
-                        /* Old-style case---we need to do a filter test */
-                        filter_test = slapi_vattr_filter_test(pb, e->ep_entry, filter, ACL_CHECK_FLAG);
+                        filter_test = 0;
+                    }
+
+                    /* If we don't check this, we could stomp the filter_test aci denied result. */
+                    if (filter_test == 0 && li->li_filter_bypass_check) {
+                        slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry", "Checking bypass\n");
+                        filter_test = slapi_vattr_filter_test(pb, e->ep_entry, filter, 0);
+                        if (filter_test != 0) {
+                            /* Oops ! This means that we thought we could bypass the filter test, but noooo... */
+                            slapi_log_err(SLAPI_LOG_ERR, "ldbm_back_next_search_entry",
+                                          "Filter bypass ERROR on entry %s\n", backentry_get_ndn(e));
+                        }
+                    }
+                } else {
+                    /* MUST APPLY - This occurs when we have a partial candidate set */
+                    /*
+                     * IMPORTANT - there is a large and important difference between "filter as intended"
+                     * and filter as executed. The filter as executed is what we optimised to, and this
+                     * can importantly include items like parentid in a one level search. The filter as
+                     * intended however is what the user ASKED for. We shouldn't penalise the user because
+                     * we mucked with their filter, so we check the ACI with the "filter as intended", but
+                     * we need to STILL apply the filter test with "as executed" in case of a test threshold
+                     * shortcut (lest we accidentally prevent the user seeing what they wanted ....)
+                     */
+                    filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
+                    if (filter_test == 0) {
+                        filter_test = slapi_vattr_filter_test(pb, e->ep_entry, filter, 0);
                     }
                 }
             }
--- a/Makefile.am
+++ b/Makefile.am
@@ -1947,6 +1947,7 @@ TESTS = test_slapd
 test_slapd_SOURCES = test/main.c \
 	test/libslapd/test.c \
 	test/libslapd/counters/atomic.c \
+	test/libslapd/filter/optimise.c \
 	test/libslapd/pblock/analytics.c \
 	test/libslapd/pblock/v3_compat.c \
 	test/libslapd/schema/filter_validate.c \
--- a/dirsrvtests/tests/suites/filter/filter_onelevel_aci_test.py
+++ b/dirsrvtests/tests/suites/filter/filter_onelevel_aci_test.py
@@ -1,0 +1,49 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 William Brown <william@blackhats.net.au>
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ----
+
+import pytest, os, ldap
+
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.topologies import topology_st
+
+from lib389.idm.account import Anonymous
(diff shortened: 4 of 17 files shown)


Firstyear:
Can you test this with FreeIPA (or ask someone from IPA to test it?) This should resolve the onelevel issue as mentioned.

mreynolds389:
We have another regression in the filters suite:   filters/filter_test.py::test_filter_scope_one

```
        log.info('Search user using ldapsearch with scope one')
        results = topology_st.standalone.search_s(DEFAULT_SUFFIX, ldap.SCOPE_ONELEVEL,'ou=services',['ou'] )
        log.info(results)
    
        log.info('Search should only have one entry')
>       assert len(results) == 1
E       assert 6 == 1
E         +6
E         -1
```

A one level search is returning a lot of entries that do not match the filter:

```
ldapsearch -xLLL -D "cn=directory manager" -h localhost -p 38901 -s one  -w password -b "dc=example,dc=com" ou=services ou
dn: ou=groups,dc=example,dc=com
ou: groups

dn: ou=people,dc=example,dc=com
ou: people

dn: ou=permissions,dc=example,dc=com
ou: permissions

dn: ou=services,dc=example,dc=com
ou: services

dn: uid=test_entry,dc=example,dc=com

dn: uid=test_entry2,dc=example,dc=com

dn: uid=ext-test-entry,dc=example,dc=com
```

Firstyear:
I had a different error on that test (something about pytest going bang) so I missed that one. I'll take a look later :)

Firstyear:
@mreynolds389 I have pushed a second fix for this. This now passes all 713 filter tests :) 

It could be that the second patch MAY need to be backported down to 1.4.3 and 1.4.4 because it MAY affect those - it's not that the query optimiser was broken, it's just showing that something else was!

mreynolds389:
> @mreynolds389 I have pushed a second fix for this. This now passes all 713 filter tests :)
> 
> It could be that the second patch MAY need to be backported down to 1.4.3 and 1.4.4 because it MAY affect those - it's not that the query optimiser was broken, it's just showing that something else was!

Well this would not go into 1.4.x anyway, only in 2.x.  Next we want to do performance tests on this.  Any filters you can suggest that would benefit most of this optimization?

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -999,13 +1005,33 @@ build_candidate_list(Slapi_PBlock *pb, backend *be, struct backentry *e, const c
...
+
+        *candidates = onelevel_candidates(pb, be, base, filter_copy, lookup_returned_allidsp, &err);
+        slapi_filter_free(filter_copy, 1);
         break;
 
     case LDAP_SCOPE_SUBTREE:
-        *candidates = subtree_candidates(pb, be, base, e, filter, managedsait,
-                                         lookup_returned_allidsp, &err);
+        /* Dup before optimise to prevent issues with aci and filter as intended vs optimised. */
+        filter_copy = slapi_filter_dup(filter);
```
mreynolds389: This filter duping & freeing for every search concerns me from a performance & memory frag perspective.  I wonder if create_*_filter() functions could be modified to only dup the filter if the filter is actually modified.  Return NULL if not modified, and use the original filter, otherwise use the copy.  That would at least reduce this overhead for cases where the optimization is not needed.

Firstyear: But there aren't many cases where this would happen. Create onelevel always has to add the parentid element. Create subtree always has to add the objectClass referral (unless we get the managedsait control, which practically no one sends). There additions were what caused the issue before. 

About the only way to potential min-max this to avoid alloc/dealloc would be to pull out the optimised inner, but then we loose the ability to optimise around parentid= and oc=referral. 

Even with the suggestion to have a filter as intended vs as optimised, we have to do an extra dup/free. So that's unavoidable unless we do some weird tracking of the inner filter that we optimise, and then the wrapper that we don't, and we free them separately or something. Which risks more memory corruption and weirdness. Generally this all gets very messy ....

mreynolds389: Sure, makes sense.  I was just hoping it could be improved, but doesn't look that way.  Ok, thanks for listening :-)  This fix is going through the entire IPA test suite at the moment.  And hopefully this week I can do some testing numbers.  

I know @tbordaz had some concerns that this fix will impact a feature where we skip acl evaluation of filter components.  He thought this might actually break that - which in turn would hurt performance (or worse?).  I asked him to go into detail and comment on this PR.  He hasn't had a chance to do so yet, but I'd like his concerns looked into (once they are accurately expressed).

In the meantime I hope to run some performance tests.  What filters do you suggest I test that will best showcase this feature's improvements?

Firstyear: Wellllll after a walk I think there MIGHT be a way to do what you want ... so let me think a bit and come up with a new patch on top today.

The classic example is "(&(objectClass=person)(uid=foo))". But generally most of the filters that SSSD emits would benefit too, I remember one of the filters was pathologically bad, but I don't recall what it was now.


Since this is only a first pass, it'll only improve things like the above, but later we can make more interesting optimisations.

Firstyear:
@mreynolds389 Extra patch that does what you want, it doesn't dup, but we take a ref to the inner filter "as intended" by the user. Then we check teh aci against the intent filter, but the attributes of the entry against the full filter. Comments help explain (especially re the onelevel case). I'm seeing a single test failure in the filter cert test, but it appears to be internal to pytest so I have no idea where to start .....

mreynolds389:
> I'm seeing a single test failure in the filter cert test, but it appears to be internal to pytest so I have no idea where to start .....

It just passed github CI tests 100%

Firstyear:
> > I'm seeing a single test failure in the filter cert test, but it appears to be internal to pytest so I have no idea where to start .....
> 
> It just passed github CI tests 100%

It must be an issue with pytest on SUSE then :)

ldap/servers/slapd/back-ldbm/filterindex.c
```diff
@@ -899,11 +903,12 @@ list_candidates(
              * If we encounter a zero length idl, we bail now because this can never
              * result in a meaningful result besides zero.
              */
+            sr->sr_flags |= SR_FLAG_MUST_APPLY_FILTER_TEST;
```
tbordaz, line 906: Not sure to understand, in such case we have no candidate. So why do we need to enforce the filter ?

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1695,31 +1694,41 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
...
                 /* it's a regular entry, check if it matches the filter, and passes the ACL check */
-                if (0 != (sr->sr_flags & SR_FLAG_CAN_SKIP_FILTER_TEST)) {
-                    /* Since we do access control checking in the filter test (?Why?) we need to check access now */
+                if (0 == (sr->sr_flags & SR_FLAG_MUST_APPLY_FILTER_TEST)) {
+                    /* Since we do access control checking in the filter test we need to check access now */
                     slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry",
                                   "Bypassing filter test\n");
                     if (ACL_CHECK_FLAG) {
-                        filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
+                        filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
```
tbordaz, line 1730: I am missing something, filter_intent is the original and optimized filter but it does not contain (objectclass=referral). If the candidate is a referral, this filter will not match and the referral will be skipped. ?

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1695,31 +1694,41 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
...
+                    /*
+                     * IMPORTANT - there is a large and important difference between "filter as intended"
+                     * and filter as executed. The filter as executed is what we optimised to, and this
+                     * can importantly include items like parentid in a one level search. The filter as
+                     * intended however is what the user ASKED for. We shouldn't penalise the user because
+                     * we mucked with their filter, so we check the ACI with the "filter as intended", but
+                     * we need to STILL apply the filter test with "as executed" in case of a test threshold
+                     * shortcut (lest we accidentally prevent the user seeing what they wanted ....)
+                     */
+                    filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG, 1 /* Only perform access checking, thank you */);
```
tbordaz, line 1756: To be honest I am a bit scared here. Evaluation of the filter is expensive, here instead of one filter matching we may do two matching. So the gain on candidate list may be lost into a loss in more expensive filter matching.

tbordaz (commented):
The overall patch looks nice, however I have some minor comments.

Regarding index_subsystems.c looks huge and I do not know the purpose of this functions, could you document in the PR its purpose and the entry points ?

There is a large set of tests to result into filter optimisation. I would like to see some user case where, for a given filter, we can verify gains (throughput or CPU/IO rate or response time or...)

mreynolds389:
Maybe I'm not testing this correctly, but I don't see any search improvement.  I created 20000 entries under ou=people,dc=example,dc=com, idscanlimit is set to 4000, and I ran searches like this:

ldapsearch -xLLL -D "cn=mark,dc=example,dc=com" -w password -b "dc=example,dc=com"  "(&(objectclass=person)(uid=HBodkin6))" dn

mreynolds389:
I tested with 100k entries under ou=people, and now I do see an improvement in some cases, but not all.  The **optimes** below represent a range of optimes I saw.

All searches were performed by a regular user (not root DN)
     
    ldapsearch -xLLL -D "cn=mark,dc=example,dc=com" -W  -b ...

## 2 Filter Components:  "(&(objectClass=person)(uid=WDuplacey9868))"

#### Optimized

Subtree:
- optime=0.000317126
- optime=0.000425608

One Level:
- optime=0.000343993
- optime=0.000487508

#### Not Optimized

Subtree:
- optime=0.001060876
- optime=0.001578461

One Level:
- optime=0.000551731
- optime=0.000933796


## 3 Filter Components:  "(&(objectClass=person)(ou=product testing)(uid=WDuplacey9868))"  

Added (ou=product testing) which matches 14000+ entries

#### Optimized

Subtree: 
- optime=0.000263894 
- optime=0.000314080 

One Level:
- optime=0.001025355
- optime=0.001075440


#### Not Optimized

Subtree: 
- optime=0.000678650
- optime=0.000884481


One Level:
- optime=0.001016183
- optime=0.001029509 
  
## Conclusion

In small databases I don't see any noticeable improvement, but in 100K database I see a noticeable improvement with subtree searches, at least double.  One level searches didn't really show an improvement.  

I realize that this kind of testing with searches that are already really fast make the results harder to measure accurately.  I feel more testing is needed.  Just need to come up with other search scenarios to try.  I'd like to see a case where a search that takes 5 seconds, now takes 2 seconds.  Something like that, but when the searches are already completing in nanoseconds, I feel the results might be a tad unreliable.

I'll try to do more testing next week, if anyone wants any kind of filters tested please let me know...

ldap/servers/slapd/back-ldbm/filterindex.c
Firstyear: Because in the future we may add other optimisations to the OR case, that would still require filter eval when there are partial sets. Also remember, that provided the candidate set is less than test threshold, we early return and thus MUST do a filter test.

ldap/servers/slapd/back-ldbm/ldbm_search.c
Firstyear: we skip checking it against *aci* specifically. because the user may not have the aci to read referals. The same way the user may not have the aci to see parentid. But we still need to enforce this condition in execution ....

ldap/servers/slapd/back-ldbm/filterindex.c
Firstyear: Which by the way is the current behaviour, it's just that now we flag this in the filterindex.c code instead, because before it was flagged by lookup_returned_allidsp in onelevel_candidates().

ldap/servers/slapd/back-ldbm/ldbm_search.c
Firstyear: The whole theory of the filter test threshold is that filter eval is faster than evaling idls because we don't have an idl cache so they are always IO bound. That's why we have the ability to shortcut out and then evaluate the remaining conditions of the filter on the entry rather than doing full-indexed queries and bypassing the filter test. 


I'm actually reading the current master branch code and I have no idea how this it actually working correctly for current queries. 

Looking at this we have `onelevel_candidates()` which wraps the filter as intended by the user, with `(&(filter_intended)(parendid=<id>))`.  Lets say the filter_intended is `uid=a`. We then call into filter_candidates, which will look into the first elements of the filter_indended, and see it's equality, we'll hit the test_threshold and then return out with an idlist of just uid=a,. 

But then in onelevel, we discard the wrapper filter, and return the flag for forcing the filter test. So we then filter test only on filter_intended. 

I'm trying to think of ways to corrupt this, because I think it may be possible to ask for a query that would have subtree elements accidentally included in our partial candidate set since we never actually evaluated parentid=, and we don't apply that condition in our filter test.


So that's why we MUST absolutely test the filter twice! We have to check "once" for the access controls of "what did the user intend" and a second of "What is the correct *query* that gets the right results". Both onelevel and referrals matter here because the user may not have read on objectClass or parentid, and because we inject both, we need to check these are properly in the result sets else we are excluding/over-including in our results.

Firstyear:
> ## Conclusion
> In small databases I don't see any noticeable improvement, but in 100K database I see a noticeable improvement with subtree searches, at least double. One level searches didn't really show an improvement.
> 
> I realize that this kind of testing with searches that are already really fast make the results harder to measure accurately. I feel more testing is needed. Just need to come up with other search scenarios to try. I'd like to see a case where a search that takes 5 seconds, now takes 2 seconds. Something like that, but when the searches are already completing in nanoseconds, I feel the results might be a tad unreliable.
> 
> I'll try to do more testing next week, if anyone wants any kind of filters tested please let me know...

Yes, you'll see this more with larger databases, OR complex queries. Consider something like .... `'(&(objectClass=nsPerson)(objectClass=nsAccount)(objectClass=posixAccount)(uid=a))'` and raise your idlistscanlimit to something really high as well. 

Also remember, this current optimiser is a framework to let us do basic re-arrangements. Later we can add more advanced ones but for now it's only simple things! 

I seem to recall that SSSD had some super nasty queries too ....

mreynolds389:
Latest test results with 100K users and 100k groups (optimized vs original code)

## Filter (with unindexed component):  (&(objectClass=person)(objectClass=inetuser)(l=Milpitas)(uid=group_entry43481-1))

"l" is NOT indexed

### idscanlimit: 4000

Subtree search was 3x faster
One Level:  almost 2x faster


### idscanlimit 99999999

Subtree search was 4x faster
One Level:  almost 2x faster

Definitely starting to see bigger swings.  Increasing allidsthreshold to have all entries scanned didn't really impact the new optimizer (although it looks like it's a little faster), but the old code suffered greatly (5x worse)  when increasing the scanlimit.  

The old code also saw each search as unindexed, while the optimizer code did not.


## Filter (fully indexed):  (&(objectClass=person)(objectClass=inetuser)(l=Milpitas)(uid=group_entry43481-1))

All attributes in filter are now indexed 

### idscanlimit: 4000

Subtree search was 7x faster
One Level:  almost 3x faster


### idscanlimit 99999999

Subtree search was 4x faster
One Level:  almost 2x faster


One thing I noticed is that with the optimization, the larger the idscanlimit the **worse** the one-level scope searches are with the optimization.  With the old/original code one-level searches **improve** with increased idscanlimit.  So there is sort of a *regression* with one-level searches with the optimization code and the idscanlimit.  However, we are still faster than the original code in all cases.

## Single Search Filter:  (uid=group_entry43481-1)

Subtree searches were about the same, but one-level searches are **2x slower** with the optimization fix than the original code.  This should be investigated...

Firstyear:
> One thing I noticed is that with the optimization, the larger the idscanlimit the **worse** the one-level scope searches are with the optimization. With the old/original code one-level searches **improve** with increased idscanlimit. So there is sort of a _regression_ with one-level searches with the optimization code and the idscanlimit. However, we are still faster than the original code in all cases.
> 
> ## Single Search Filter: (uid=group_entry43481-1)
> Subtree searches were about the same, but one-level searches are **2x slower** with the optimization fix than the original code. This should be investigated...

I wonder if that means that `parentid` is not indexed. That would be the only reason I could think of for such a regression here .... 

Have you tried with the updated PR? That re-arranges how we apply parentid to make it a tail element so it should generally be bypassed in favour for filter testing. But we also need to check parentid is indexed.

Anyway, at the end of the day, the optimiser will probably mean we can remove idlistscanlimit because idlistscan limit is a naive attempt to "optimise" to smaller search terms, where this is a more aware one.

mreynolds389:
I wasn't sure if I had the latest patch so I started over.  Still seeing a big discrepancy:

Master:

    [25/Feb/2022:14:17:19.186661811 -0500] conn=10 op=1 SRCH base="ou=people,dc=example,dc=com" scope=1 filter="(uid=group_entry43481-1)" attrs="distinguishedName"
    [25/Feb/2022:14:17:19.187135160 -0500] conn=10 op=1 RESULT err=0 tag=101 nentries=1 wtime=0.000140460 optime=0.000481984 etime=0.000617927

Your PR:

    [27/Feb/2022:13:01:29.212615420 -0500] conn=4 op=1 SRCH base="ou=people,dc=example,dc=com" scope=1 filter="(uid=group_entry43481-1)" attrs="distinguishedName"
    [27/Feb/2022:13:01:29.219901187 -0500] conn=4 op=1 RESULT err=0 tag=101 nentries=1 wtime=0.000251033 optime=0.007293483 etime=0.007538561

The optime is much larger with your PR :-(  I should probably test base searches too (next week....)

I created the ldif using:  

    # dsctl localhost ldifgen groups myGroup

I specified 100000 groups, and to create a user and add it to a group. (200k entries total)

parentid is indexed (by default), and config(dse.ldif) is default.

In my scenario I then created a user to bind with, so ACI's are in the mix:   **cn=mark,dc=example,dc=com**

Here is my search:

    # ldapsearch -xLLL -D "cn=mark,dc=example,dc=com" -W -b "ou=people,dc=example,dc=com" -s one  "(uid=group_entry43481-1)" dn

Firstyear:
@mreynolds389 Fixed. 

```
SUB
RESULT err=0 tag=101 nentries=1 wtime=0.000317893 optime=0.000532981 etime=0.000848102
ONE
RESULT err=0 tag=101 nentries=1 wtime=0.000506704 optime=0.000402270 etime=0.000906464
```

The issue was that the internal filtel was `(&(parentid)(|(filter)(referal)))` when it should have been `(&(|(filter)(referal))(parentid))` which is logically equivalent, but prefers out optimised filter elements first, then we test-threshold on the parentid in the later stages.

mreynolds389:
> @mreynolds389 Fixed.
> 
> ```
> SUB
> RESULT err=0 tag=101 nentries=1 wtime=0.000317893 optime=0.000532981 etime=0.000848102
> ONE
> RESULT err=0 tag=101 nentries=1 wtime=0.000506704 optime=0.000402270 etime=0.000906464
> ```
> 
> The issue was that the internal filtel was `(&(parentid)(|(filter)(referal)))` when it should have been `(&(|(filter)(referal))(parentid))` which is logically equivalent, but prefers out optimised filter elements first, then we test-threshold on the parentid in the later stages.

I actually woke up in the middle of the night and had the same idea (I remembered from the aci test that parentid was looked at first - which is wrong).  I almost got up and worked on it at like 3am.  I was hoping you would not have looked at it so I could have fixed it before you haha!

mreynolds389:
Base searches appear to be faster as well.  This is looking pretty good now.  I think @tbordaz wanted to run more tests, so I will chat with him next, but I'm feeling pretty good about this!

ldap/servers/slapd/back-ldbm/ldbm_search.c
tbordaz: > The whole theory of the filter test threshold is that filter eval is faster than evaling idls because we don't have an idl cache so they are always IO bound. That's why we have the ability to shortcut out and then evaluate the remaining conditions of the filter on the entry rather than doing full-indexed queries and bypassing the filter test.

filter threshold is to interrupt the process of building (from the filter) the candidate list when the #candidate is low (10 IIRC) and move immediately to filter evaluation. IIRC the main issue with filter threshold is when we have a 'OR' that force to fetch candidates for all filter components.

tbordaz:
@mreynolds389 , those results are impressive. I think this ticket is an improvement rather than a fix, if we want to verify that we would need some test case (topology, indexing, command to provision entries, commands to verify gain...). Regarding the tests, it looks the SRCH returns 1 entry, would be interesting to check the gain if the filter returns 10 entries (|(uid=foo100000)...(uid=foo100009))

Firstyear:
I think it's worth pointing out what I mean about the "naive filter optimise". 

Let's say (since we are human and have this insight that our server does ...) we have a search like `(&(objectClass=group)(cn=foo))`. Let's assume our 200k entry db, with 100k groups and 100k users. 

This means, that when translated, the size of the idls is:

```
(&(objectClass=group = 100,000)(cn=foo=1))
```

If we executed this query "as is", we'll load 100,000 32bit ints from the db (391kb of data) for objectClass=group.  (ASIDE: Since we DO NOT have IDL caching, this comes from the BDB / LMDB cache, which is a bit more precarious since it shares the same hot path as our entry db). Now we then step forward and load cn=foo, which is a single 4 byte load since it's a single value. We then and the two sets, and return the single value of cn=foo. 

The goal to speed this up, is to *avoid loading objectClass=group if possible*.

entrylistscanlimit achieves this by *early termination of the idl load* if it is too large.

As we start to load the objectClass=group idl, entrylistidlscanlimit notices we loaded 2000 (I think is the default) values from the idl (about 8kb), and says "okay this is too large, skip this". Then we consult cn=foo, and get the single value idl, and do a filter test for the missing assertion over objectClass=group. 

The query optimiser does this by re-arranging the query to put cn=foo first. Since we then have `(&(cn=foo=1)(objectClass=group = 100,000))`, we load the cn=foo idl first now. We then see it is a single value which is less than a filter_test_threshold so we shortcut and return and apply the filter test.

This is what I mean when I say entrylistidlscanlimit is a naive-filter-optimiser. It is attempting to do the same thing (skip large idls if possible) but in a way that is frankly, stupid. Why stupid? Because entryidllistscanlimit forces *full table scans* on any query that has a large result set.


Let's change our query. This time we want to search for `(objectClass=group)`. 

With the filter optimiser, we see there is nothing to optimise, so we load the (large, but correct) idl for objectClass=group. This actually allows us to bypass filter testing the entries since all elements of the filter were executed and asserted!

However, with entryidlistscanlimit, since we shortcut during the load, we force return ALLIDS, which then means we have to filter test *against every entry of the database* - in this example, 200,000 entries. This means instead of just loading the objectClass idl, we end up loading the full database instead. If we had a small entry cache, we would likely end up cache-busting ourselves with a full scan like this. 


You can consider that one something like freeipa, which has many different object classes, you could expect at a guess maybe 1/4 or 1/3 of entries in the DB to be groups, and of course, entrylistidlscanlimit means that we have to full table scan anyway. 

To put it another way - entryidllistscanlimt means than any index which is "large" is now useless, and does nothing. This means objectClass, parentid, and many many more. 

So with the filter optimiser we also now have the opportunity to *remove* entryidllistscanlimit, which means we benefit on both small AND large queries. 


That's why I've been asking you to test with entryidllistscanlimit tunings :) 

This is also why to "test" the changes you have to check eitherside of the filter-test-threshold. If you are below it or above it, you'll hit different penalties (mainly incurred by entryidlistscanlimit. )

Firstyear:
@tbordaz I think we are just waiting on your review to merge this? @mreynolds389 do you agree?

mreynolds389:
> @tbordaz I think we are just waiting on your review to merge this? @mreynolds389 do you agree?

Not yet, Thierry wanted me to run some other types of queries first.  I haven't had a chance yet because I'm working on several hot/important issues at the moment.  I plan to wrap it up by the end of the week.

Firstyear:
Cool cool, no rush, we can leave it for next week

mreynolds389:
One side observation, increasing the idscanlimit to "cover all entries" is faster in **all** cases than having a scan limit hit.  

I'm really starting to think the global idscanlimit causes more harm than good.  We still have the per index idscanlimit which I feel still serves a useful purpose, but I think the global limit should be removed IMHO or set to "unlimited" by default.  I'm not sure how LMDB will respond to these varying ID scan limits, but bdb(libdb) performs better without a scan limit.

Firstyear:
It's likely that LMDB will be faster again as it has many similar behaviours to BDB. And yes, I fully agree. :)

Firstyear:
> Well I'm a bit surprised, but all of these filters are at least twice as slow with the "optimization" verses plain master branch:
> 
> ```
> "(&(objectClass=top)(|(uid=group_entry2-1)(uid=group_entry3-1)(uid=group_entry4-1)(uid=group_entry5-1)(uid=group_entry6-1)(uid=group_entry7-1)(uid=group_entry8-1)(uid=group_entry9-1)(uid=group_entry10-1)(uid=group_entry11-1)(uid=group_entry12-1)))"
> 
> "(&(objectClass=top)(objectClass=person)(|(uid=group_entry2-1)(uid=group_entry3-1)(uid=group_entry4-1)(uid=group_entry5-1)(uid=group_entry6-1)(uid=group_entry7-1)(uid=group_entry8-1)(uid=group_entry9-1)(uid=group_entry10-1)(uid=group_entry11-1)(uid=group_entry12-1)))"
> 
> "(&(objectClass=top)(uid=*)(|(uid=group_entry2-1)(uid=group_entry3-1)(uid=group_entry4-1)(uid=group_entry5-1)(uid=group_entry6-1)(uid=group_entry7-1)(uid=group_entry8-1)(uid=group_entry9-1)(uid=group_entry10-1)(uid=group_entry11-1)(uid=group_entry12-1)))"
> 
> "(&(objectClass=top)(uid=group*)(|(uid=group_entry2-1)(uid=group_entry3-1)(uid=group_entry4-1)(uid=group_entry5-1)(uid=group_entry6-1)(uid=group_entry7-1)(uid=group_entry8-1)(uid=group_entry9-1)(uid=group_entry10-1)(uid=group_entry11-1)(uid=group_entry12-1)))"
> ```
> 
> These filters all return 11 entries.
> 
> To reproduce create and import this ldif:
> 
> ```
> # dsctl localhost ldifgen groups create --number 100000 --parent ou=groups,dc=example,dc=com --num-members 1 --create-members --member-parent ou=people,dc=example,dc=com --member-attr uniquemember --ldif-file /tmp/users.ldif --suffix dc=example,dc=com
> ```
> 
> Then run subtree searches (binding as a user in the database), and check the "optime" in the access log.

That's because they are above filter-test-threshold, which means we then also execute the remaining elements of the filter such as the objectClass calls etc. That then hits the "worst case" which is loading the larger query. 

So it's likely we either need to raise filter-test-threshold to find the cross over point, OR we need to look into compressed indexes, because it's likely IOPS on objectClass indexes that is bottlenecking larger indexes.

mreynolds389:
> That's because they are above filter-test-threshold, which means we then also execute the remaining elements of the filter such as the objectClass calls etc. That then hits the "worst case" which is loading the larger index.

I'm assuming you mean that magic limit of "10".  But why are we slower?  Shouldn't we be the same as bdb then? 

I'll try searches tomorrow that are under that limit (just for curiosity).  I wonder if that limit is still needed then?  Or bump it to 100?  < just picking a random number :-) >

> 
> So it's likely we either need to raise filter-test-threshold to find the cross over point, OR we need to look into compressed indexes, because it's likely IOPS on objectClass indexes that is bottlenecking larger indexes.

Didn't you already have C code for that?

> 
> OR we need an IDL cache so we aren't having to continually access these. I think we also don't do binary search on these where IDLset does ( https://github.com/kanidm/idlset )

Definitely something to look into. 

For now, I don't think this PR can be merged (yet)...

Firstyear:
> > That's because they are above filter-test-threshold, which means we then also execute the remaining elements of the filter such as the objectClass calls etc. That then hits the "worst case" which is loading the larger index.
> 
> I'm assuming you mean that magic limit of "10". But why are we slower? Shouldn't we be the same as bdb then?

See above: Specifically:

>> The query optimiser does this by re-arranging the query to put cn=foo first. Since we then have (&(cn=foo=1)(objectClass=group = 100,000)), we load the cn=foo idl first now. We then see it is a single value which is less than a filter_test_threshold so we shortcut and return and apply the filter test.

The moment we are above filter-test-threshold we don't shortcut return, and we start to load the objectClass=group IDL which is ~391kb. So we hit the BDB cache and have to read the full thing in (unless we hit entryidlistscanlimit, but then we are hurting on larger queries). 

This is why we need to either:

* Reduce the amount of IO - compressed IDL's can be 6% of the size in the optimal case, and 1/8th of the size in worst case. This means we'd be between ~24kb to ~50kb which is a huge reduction in IO AND it means the BDB cache can cache MORE
* Implement an IDL cache, so we don't do any IO at all, and the IDL's are ready to go this is what Kanidm does, so we have zero-IO to hit an IDL unless we cache miss.
* Add binary search scanning. I think we currently do a linear scan of the full 100,000 id's in the IDL to perform the union. In Kanidm + IDLSet, when one set is 1/8th the size of the other or less, we do a binary search of the elements of the smaller into the larger. This is MUCH faster.

> 
> I'll try searches tomorrow that are under that limit (just for curiosity). I wonder if that limit is still needed then? Or bump it to 100? < just picking a random number :-) >

That's what I meant by the cross-over point - we need to test where it becomes faster to shortcut vs not. In Kani we found that a test threshold of 2 is actually faster, but we have the full suite of the above improvements, the optimiser, and a stats analyser of the indexes, so there is fair bit to potentially do to get to that point where we could actually lower test threshold down. 

> 
> > So it's likely we either need to raise filter-test-threshold to find the cross over point, OR we need to look into compressed indexes, because it's likely IOPS on objectClass indexes that is bottlenecking larger indexes.
> 
> Didn't you already have C code for that?

Nope, it's all in Rust. 

> 
> > OR we need an IDL cache so we aren't having to continually access these. I think we also don't do binary search on these where IDLset does ( https://github.com/kanidm/idlset )
> 
> Definitely something to look into.
> 
> For now, I don't think this PR can be merged (yet)...

Even today, we would hit this problem. It's just that entryidlistscanlimit masks it by triggering the second shortcut return case. 

So this PR can be merged, but we can't drop entryidlistscanlimit yet I think is the correction here.

Firstyear:
PS: https://github.com/kanidm/kanidm/blob/master/kanidmd/src/lib/be/idl_arc_sqlite.rs#L709 

This is the stats analyser I'm talking about btw.

mreynolds389:
> Even today, we would hit this problem. It's just that entryidlistscanlimit masks it by triggering the second shortcut return case.
> 
> So this PR can be merged, but we can't drop entryidlistscanlimit yet I think is the correction here.

Well, I don't exactly see it that way.  I see it as the optimization helps a lot for queries, but also hurts other queries.  You can argue that it helps the more common use cases, but in reality it's hard to say what really is the common case for everyone.  I'm going to do more testing with this shortcut threshold and see how high we can increase it before it hurts things.  Once we find a better number I want include it with this PR.

As for the things kanidm does I think we should look into similar design changes, but we don't have time right now to work on them.  So they are longer term goals.  For now I'd like to get this filter optimization as finely tuned as possible.

Firstyear:
> > Even today, we would hit this problem. It's just that entryidlistscanlimit masks it by triggering the second shortcut return case.
> > So this PR can be merged, but we can't drop entryidlistscanlimit yet I think is the correction here.
> 
> Well, I don't exactly see it that way. I see it as the optimization helps a lot of queries, but also hurts other queries. You can argue that it helps the more common use cases, but in reality it's hard to say what really is the common case for everyone. I'm going to do more testing with this shortcut threshold and see how high we can increase it before it hurts things. Once we find a better number I want include it with this PR.

Sure, but remember entryidlistscanlimit is currently "our poor attempt" to do the same as the filter optimiser. So it *also* helps some queries and hurts many others. The goal is to try and get it to be at least similar (which I think it already is). 

The test threshold is already something we have to deal with, and would affect the same queries above with the 11 users in an OR, but the idllistscanlimit then shortcuts on the objectClass term at the end. It's when you raise that limit, you are see the cross over in what's occuring. So we choose, do we hurt larger or smaller queries with the idlistscanlimit. That's not the filter optimiser causing this issue - it's because now we are looking at it in detail we are seeing the edge cases. 

Anyway, feel free to tweak with the test threshold. :) I look forward to the results.

Firstyear:
Any further comments on this?

mreynolds389:
> Any further comments on this?

@tbordaz had some concerns and didn't want to merge it (yet?).  I'll let him comment on this...

Firstyear:
Okay, I'll wait for @tbordaz then :)

tbordaz:
Sorry I have not look at the PR for a long time.
Only think that is not clear to me is what @mreynolds389 mentionned https://github.com/389ds/389-ds-base/pull/5171#issuecomment-1058753981

My understanding is that the filter "(&(objectClass=top)(|(uid=1)...(uid=12))" is slower with optimizer code. Is it ?
I would expect the optimizer to craft the filter to "(&(|(uid=1)...(uid=12)(objectClass=top)), then missing the "less than 10 candidate" it will evaluate '(objectClass=top)', I would expect no benefit (optimizer code vs regular code) in such case.
 Is it really slower ?, is it expected with this very specific filter  and if so why ?

Firstyear:
> Sorry I have not look at the PR for a long time. Only think that is not clear to me is what @mreynolds389 mentionned [#5171 (comment)](https://github.com/389ds/389-ds-base/pull/5171#issuecomment-1058753981)
> 
> My understanding is that the filter "(&(objectClass=top)(|(uid=1)...(uid=12))" is slower with optimizer code. Is it ? I would expect the optimizer to craft the filter to "(&(|(uid=1)...(uid=12)(objectClass=top)), then missing the "less than 10 candidate" it will evaluate '(objectClass=top)', I would expect no benefit (optimizer code vs regular code) in such case. Is it really slower ?, is it expected with this very specific filter and if so why ?

The problem has nothing to do with the optimiser. This would affect us WITHOUT the optimiser if you just provided that query today. 

The issue is the filter test threshold, where when we have a partial candidate set (IE we've looked at some filter bits, but not all), when the threshold is lower than that limit, it will shortcut.

So because of the OR condition what happens is:

```
(&(|(uid=1)(uid=2)(uid=3)(uid=4)(uid=5)(uid=6)(uid=7)(uid=8)(uid=9))(objectClass=top))
     ^-- each uid= is an idl of length 1

(&(|(uid=1)...(uid=9))(objectClass=top))
   ^-- which means this OR becomes an IDL of length 9

(&(|(uid=1)...(uid=10))(objectClass=top))
 ^-- so now we have the OR complete we ask "are we more or
       less than the filter test threshold".
       9 < 10, so we return and objectClass=top is never checked.
       It's checked by filter testing


(&(|(uid=1)(uid=2)(uid=3)(uid=4)(uid=5)(uid=6)(uid=7)(uid=8)(uid=9)(uid=10)(uid=11))(objectClass=top))
     ^-- each uid= is an idl of length 1

(&(|(uid=1)...(uid=11))(objectClass=top))
   ^-- which means this OR becomes an IDL of length 11

(&(|(uid=1)...(uid=11))(objectClass=top))
 ^-- so now we have the OR complete we ask "are we more or
       less than the filter test threshold".
       11 > 10, so we continue.


(&(|(uid=1)...(uid=11))(objectClass=top))
  ^-- idl of len 11    ^
                       \-- is now processed - mark's test, this is 100,000



(&(|(uid=1)...(uid=11))(objectClass=top))
  ^-- idl of len 11    ^
                       \-- idl of len 100,000
  At this point the two idls are now anded.

```

Because the test is set to 10, that means that when you have an OR with 11 or more items, it will keep going to eval objectClass top.

This is *not a fault of the query optimiser*. This is how the current code works. This happens today, which is why in my view, this is not a blocker.

tbordaz:
@Firstyear thanks for the description. I agree that when the first component has more than 10 IDs then we evaluate the second component (objectclass=top) that is a killer.

What I do not understand is why it is slower with  PR (according to https://github.com/389ds/389-ds-base/pull/5171#issuecomment-1058753981) ?

With the original filter '(&(objectclass=top)(|(uid=1)... (uid=11))', the patch reorders the filter to '(&(|(uid=1)... (uid=11))(objectclass=top))'.

I would expect '(&(objectclass=top)(|(uid=1)... (uid=11))' to be as bad as '(&(|(uid=1)... (uid=11))(objectclass=top))'.

Firstyear:
> @Firstyear thanks for the description. I agree that when the first component has more than 10 IDs then we evaluate the second component (objectclass=top) that is a killer.
> 
> What I do not understand is why it is slower with PR (according to [#5171 (comment)](https://github.com/389ds/389-ds-base/pull/5171#issuecomment-1058753981)) ?
> 
> With the original filter '(&(objectclass=top)(|(uid=1)... (uid=11))', the patch reorders the filter to '(&(|(uid=1)... (uid=11))(objectclass=top))'.
> 
> I would expect '(&(objectclass=top)(|(uid=1)... (uid=11))' to be as bad as '(&(|(uid=1)... (uid=11))(objectclass=top))'.

I'm going to ignore the filter which is "(&(|(uid=1)... (uid=11))(objectclass=top))" because that's optimal. What we care about here is when it's the suboptimal form of "(&(objectclass=top)(|(uid=1)... (uid=11)))"

```
-- CURRENT CODE - default entryidlistscan limit

(&(objectClass=top)(|(uid=1)...(uid=11)))
  ^-- idl begins to be loaded. 2000 id's of 100,000 are loaded.
        entryidlistscanlimit engages. IDL is shortcut to ALLIDS.
        ALLIDS does not trigger filter test threshold.

(&(objectClass=top)(|(uid=1)...(uid=11)))
                      ^-- each uid= is an idl of length 1

(&(objectClass=top)(|(uid=1)...(uid=11)))
                    ^-- which means this OR becomes an IDL of length 11

(&(objectClass=top)(|(uid=1)...(uid=11)))
 ^ ALLIDS & IDL(11) shortcuts to return the IDL(11), filter test is applied.


-- CURRENT CODE - entryidlistscanlimit of 999999999

(&(objectClass=top)(|(uid=1)...(uid=11)))
  ^-- idl of length 100,000 is loaded
      100,000 > 10 does not trigger filter test threshold.

(&(objectClass=top)(|(uid=1)...(uid=11)))
                      ^-- each uid= is an idl of length 1

(&(objectClass=top)(|(uid=1)...(uid=11)))
                    ^-- which means this OR becomes an IDL of length 11

(&(objectClass=top)(|(uid=1)...(uid=11)))
 ^ IDL(100,000) & IDL(11),
    filter test is NOT applied because this was fully resolved.


-- QUERY OPTIMISER - default entryidlistscan limit

(&(objectClass=top)(|(uid=1)...(uid=11)))
   ^-- objectClass eq is always demoted.

(&(|(uid=1)...(uid=11))(objectClass=top))
    ^-- each uid= is an idl of length 1

(&(|(uid=1)...(uid=11))(objectClass=top))
   ^-- which means this OR becomes an IDL of length 11
       11 > 10 does not trigger filter test threshold.

(&(|(uid=1)...(uid=11))(objectClass=top))
                       ^-- idl begins to be loaded. 2000 id's of 100,000 are loaded.
                       entryidlistscanlimit engages. IDL is shortcut to ALLIDS.

(&(|(uid=1)...(uid=11))(objectClass=top))
 ^ IDL(11) & ALLIDS shortcuts to return the IDL(11), filter test is applied.


-- QUERY OPTIMISER - entryidlistscanlimit of 999999999

(&(objectClass=top)(|(uid=1)...(uid=11)))
   ^-- objectClass eq is always demoted.

(&(|(uid=1)...(uid=11))(objectClass=top))
    ^-- each uid= is an idl of length 1

(&(|(uid=1)...(uid=11))(objectClass=top))
   ^-- which means this OR becomes an IDL of length 11
       11 > 10 does not trigger filter test threshold.

(&(|(uid=1)...(uid=11))(objectClass=top))
                       ^-- idl of length 100,000 is loaded

(&(|(uid=1)...(uid=11))(objectClass=top))
 ^ IDL(11) & IDL(100,000)
   filter test is NOT applied because this was fully resolved.

```

This is what I mean when I say "entryidlistscan is a naive attempt to query optimise". See how in the first case, the objectClass=top is skipped because it's "too large"? That's *trying* to do the same as the query optimiser where we demote the objectClass eq filter to the end. What happens is that when you have a situation where you want to get a list of lots of items, you end up forcing everything to full table scans. 

The only issue here is that the filter test threshold is likely "too low" at the moment.

Remember though, this is a pretty pathologically bad query. It's the edge case of edge cases. 

The query optimiser is designed to take queries that are extremely bad today, and fix them (i.e. pretty much anything SSSD generates). 

So that's why there is a difference here. Today I'm certainly advocating for the query optimiser to be added. The entryidlistscan limit being removed is a future goal, and not feasible today, because many other improvements are needed to make this viable (compressed indexes, idl caching, etc).

tbordaz (approved):
I failed to reproduce https://github.com/389ds/389-ds-base/pull/5171#issuecomment-1058753981. All etimes are below 1ms so it is uncertain to compare the numbers (ranging 0.16ms to 0.58ms). I tested with idlscanlimit - 4K  to100K

Thanks @Firstyear for your great explanations. My concern was to get similar etimes when the idl threshold (10) was missed and that is the case.

Sorry for this long discussion I am happy to ack your patch.

Firstyear:
Okay, just up to @mreynolds389 then! And I'll rebase the conflicts and merge once that's all good :)

mreynolds389:
> Okay, just up to @mreynolds389 then! And I'll rebase the conflicts and merge once that's all good :)

ACK!

Firstyear:
Okay, I'll rebase and merge then. Thank you!

PR #5172, opened 2022-02-23
Remove stale libevent(-devel) dependency

Flowdalic:
It appears that the last user of libevent disappeared with
44e92dc8b900 ("Ticket 50669 - remove nunc-stans").

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -803,8 +803,6 @@ AM_CONDITIONAL([FREEBSD],[test "$platform" = "freebsd"])
 AM_CONDITIONAL([SPARC],[test "x$TARGET" = xSPARC])
 
 # Check for library dependencies
-PKG_CHECK_MODULES([EVENT], [libevent])
-
 if $PKG_CONFIG --exists nspr; then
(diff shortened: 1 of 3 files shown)


Firstyear:
I think we are working to re-include libevent soon, so maybe we might leave this one ....

Flowdalic:
> I think we are working to re-include libevent soon, so maybe we might leave this one ....

Well,. or just revert the commit once you re-introduce libevent in 389ds? The main issue right now is `PKG_CHECK_MODULES([EVENT], [libevent])`, which means you can't build 389ds without libevent being present (or without patching this line out).

Flowdalic:
Ping. I just ran into this again when bumping 389-ds-base in Gentoo.

mreynolds389:
> Ping. I just ran into this again while bumping 389-ds-base in Gentoo.

Sorry we were going to actually use it for epoll() but that work is indefinitely stalled, so we can back this out for now...

mreynolds389 (approved):

mreynolds389:
@Flowdalic - do you need this in earlier versions?  Or is "main" fine?

Issue #5176, opened 2022-02-24
CI rewriter fails when libslapd.so.0 does not exist


tbordaz:
**Issue Description**
Should improve robustness of tests/suites/rewriters/basic_test.py::test_foo_filter_rewriter, so that it checks if libslapd.so or libslapd.so.0 exists, before using it into adding the path to the config entry

**Package Version and Platform:**
Since 1.43

**Steps to Reproduce**
Run tests/suites/rewriters/basic_test.py::test_foo_filter_rewriter on 2.0

**Expected results**
Test should check which libslapd exists

PR #5177, opened 2022-02-24
Issue 5176 - CI rewriter fails when libslapd.so.0 does not exist

tbordaz:
Bug description:
	rewriter test assumes that the library libslapd.so.0
	exists (/usr/lib64/dirsrv/libslapd.so.0).
	On 389-ds-base-2.0, only libslapd.so exists.

Fix description:
	The test case should test if libslapd.so exists
        If not then fall back to libslapd.so.0

relates: https://github.com/389ds/389-ds-base/issues/5176

Reviewed by:

Platforms tested: F35


mreynolds389 (approved):
Ack

Issue #5184, opened 2022-03-01
memberOf does not work correctly with multiple include scopes


mreynolds389:
**Issue Description**

MemberOf Plugin only looks at the first include scope, and the rest are ignored.  So if multiple "memberOfEntryScope" attributes are set then the plugin will not work as expected.

mreynolds389:
11adb34df..1f6646d7e  389-ds-base-2.0 -> 389-ds-base-2.0
33452a7c5..a56201425  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
3b397371b..1d3969b51  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5185, opened 2022-03-01
Issue 5184 - memberOf does not work correctly with multiple include scopes

mreynolds389:
Bug Description:

MemberOf Plugin only looks at the first include scope, and the rest are
ignored. So if multiple "memberOfEntryScope" attributes are set then the
plugin will not work as expected.

Fix Description:

The fix is to read all the memberOfEntryScope attributes and update the
group cache.

relates: https://github.com/389ds/389-ds-base/issues/5184

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -626,21 +626,6 @@ memberof_del_dn_type_callback(Slapi_Entry *e, void *callback_data)
     return rc;
 }
 
-/* Check if the the entry include scope is a child of the sdn */
-static Slapi_DN *
-memberof_scope_is_child_of_dn(MemberOfConfig *config, Slapi_DN *sdn)
-{
-    int i = 0;
-
-    while (config->entryScopes && config->entryScopes[i]) {
-        if (slapi_sdn_issuffix(config->entryScopes[i], sdn)) {
-            return config->entryScopes[i];
-        }
-        i++;
-    }
-    return NULL;
-}
-
 static void
 add_ancestors_cbdata(memberof_cached_value *ancestors, void *callback_data)
 {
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/memberof/memberof.c
```diff
@@ -813,16 +811,25 @@ memberof_call_foreach_dn(Slapi_PBlock *pb __attribute__((unused)), Slapi_DN *sdn
...
+                        /* Search each include scope */
+                        slapi_search_internal_set_pb(search_pb, slapi_sdn_get_dn(config->entryScopes[i]),
+                                                     LDAP_SCOPE_SUBTREE, filter_str, 0, 0, 0, 0,
+                                                     memberof_get_plugin_id(), 0);
+                        slapi_search_internal_callback_pb(search_pb, callback_data, 0, callback, 0);
+                    }
+                }
+                if (!all_backends) {
+                    goto done;
+                }
```
tbordaz: in case 'all_backend'==True, is there a risk that it goes to the internal search below (line 836). In such case is base_sdn valid and is it processing the base_sdn twice ?

mreynolds389: This function is actually rather confusing.  I've revised it again, and it passes CI tests, but please review...

Firstyear:
can you add a test for a "broken" case like `dc=example,dc=com` and `ou=group,dc=example,dc=com`?

ldap/servers/plugins/memberof/memberof.c
mreynolds389: @tbordaz could you review this again?  Remember there is second part to this that is waiting for this to be merged first.

tbordaz (commented):
The patch looks good to me. I admit that this part of code handling scoping/exclude/only one backend is a bit messy but not sure how to write in a better way.

tbordaz (approved):
ACK

Issue #5189, opened 2022-03-04
memberOf plugin exclude subtree not cleaning up groups on modrdn


mreynolds389:
**Issue Description**

If a user is moved from one subtree to an "excluded" subtree, the group still has the member attribute pointing to the entry that is now out of scope.  You could argue RI plugin should do this, but both RI & MO plugins have their own include/exclude scopes.  They are not required to be in sync, so in the modrdn case the MO plugin should do the cleanup.

mreynolds389:
3329b180b..94afd1dd7  389-ds-base-2.1 -> 389-ds-base-2.1
3de6dda19..a0d1658cf  389-ds-base-2.0 -> 389-ds-base-2.0
967f228da..821db3487  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
2ce9321d8..0a150ec01  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5190, opened 2022-03-04
Issue 5189 - memberOf plugin exclude subtree not cleaning up groups on modrdn

mreynolds389:
Bug Description:

If MO plugin is configured to exclude membership to certain subtrees
then when a member is moved (moddn) to the "excluded" subtree the group
still maintains the member/uniquemember attribute.

Fix Description:

After modrdn if the user is out of scope then all the groups that it
belonged to will have the membrship attribute also removed from them.

This allows an admin to move entries to "special" locations, like a
"disabled ou" container, and all its memberships are properly updated
automatically.

relates: https://github.com/389ds/389-ds-base/issues/5189

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -575,7 +575,10 @@ memberof_del_dn_from_groups(Slapi_PBlock *pb, MemberOfConfig *config, Slapi_DN *
 
         groupattrs[0] = config->groupattrs[i];
 
-        slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM, "memberof_del_dn_from_groups: Ancestors of %s\n", slapi_sdn_get_dn(sdn));
+        slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM,
+                "memberof_del_dn_from_groups: Ancestors of %s   attr: %s\n",
+                slapi_sdn_get_dn(sdn),
+                groupattrs[0]);
         rc = memberof_call_foreach_dn(pb, sdn, config, groupattrs,
                                       memberof_del_dn_type_callback, &data, &cached, PR_FALSE);
     }
@@ -618,9 +621,7 @@ memberof_del_dn_type_callback(Slapi_Entry *e, void *callback_data)
     slapi_pblock_destroy(mod_pb);
 
     if (rc == LDAP_NO_SUCH_ATTRIBUTE && val[0] == NULL) {
-        /* if no memberof attribut exists
-         * handle as success
-         */
+        /* if no memberof attribute exists handle as success */
         rc = LDAP_SUCCESS;
     }
     return rc;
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/slapi-plugin.h
```diff
@@ -1858,6 +1858,17 @@ PRBool slapi_entry_attr_get_bool(const Slapi_Entry *e, const char *type);
  */
 PRBool slapi_entry_attr_get_bool_ext(const Slapi_Entry *e, const char *type, PRBool default_value);
 
+/**
+ * Gets a array of Slapi_Values for a particular attribute.  These values are
+ * not copies and should not be freed.
+ *
+ * \param e Entry from which you want to get the boolean value.
+ * \param type Attribute type from which you want to get the values.
```
droideck: Minor compiler complaint...

droideck (approved):
Besides the warning, looks good to me!

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -958,6 +1025,10 @@ memberof_postop_modrdn(Slapi_PBlock *pb)
                         slapi_log_err(SLAPI_LOG_ERR, MEMBEROF_PLUGIN_SUBSYSTEM,
                                       "memberof_postop_modrdn - Delete dn callback failed for (%s), error (%d)\n",
                                       slapi_entry_get_dn(post_e), ret);
+                    } else {
+                        /* Need to check subtree exclusion and clean up the
+                         * entry's groups if needed */
+                        memberof_cleanup_after_modrdn(&configCopy, pre_e, post_e);
```
tbordaz: I am a bit lost here.
My understanding is that here the post_e gets out of the scope, and this function updates all groups with 'member: post_e.dn'. Is that correct ?

In which case this can happen ? A group have a member that did not yet exist ?

mreynolds389: Ok, yeah I realize this is confusing.  MemebrOf (not RI plugin) does a lot of manipulation before it hits this code.  By the time we get here the "Post entry" has already been stripped of the memberOf attributes.  The groups the Pre-entry belonged to have already had the uniquemember DN renamed to the Post DN:

    uid=mark,ou=included,o=test
    memberOf: cn=group1,ou=groups,o=test

    cn=group,ou=groups,o=test
    uniquemember: uid=mark,ou=included,o=test

Do the modrdn of uid=mark --->  uid=mark,ou=excluded,o=test

Then in MemberOf modrdn postop when we get to my code the entries look like this:

    uid=mark,ou=excluded,o=test

    cn=group,ou=groups,o=test
    uniquemember: uid=mark,ou=excluded,o=test

So to cleanup the groups the user used to belong to we need to look at the user's "pre-entry" so we can read the "memberOf" attributes and find the groups the "pre entry" belonged to.  But now these groups already had the uniquemember renamed to the new post entry DN.  So we use the "post entry" DN to cleanup the groups that were found in the "pre entry".  

I hope that clears this up a little.  The handling of the excluded subtree is a corner case and the include/exclude feature was added after the MO plugin was originally designed and written.  So I'm sure this could be cleaner, but it would require rewriting the function (risky!).  The safest approach is to just to deal with the quirkiness of the MO behavior, and handle it this way IMHO.

tbordaz: Thanks Mark for the explanation, so your change is good to me. ACK.
I have a comment, if at this point groups have a reference to post_E, what is the purpose of the function memberof_del_dn_from_groups few lines above ?

Regarding perf pov, include/exclude is a hit, triggering at least 2 times more internal MOD that strictly required.
Assume the entry E, in include scope, belongs to N groups. Then the entry is move in an exclude scope.
MO updates  the N groups to add 'member: E.post', then updates E to remove membership attribute (like 'member') (memberof_del_attr_list) , then updates E to remove 'memberof' (memberof_del_dn_type_callback), then updates the N groups to remove 'member: E.post'. That is 2*(N+1) updates where in theory, only N+1 were required.
We should revisit MO to make it more efficient and compare with what @Firstyear suggested long time ago.

tbordaz (approved):
Thanks for your feedback. The patch is good to me

ldap/servers/plugins/memberof/memberof.c
mreynolds389: > Thanks Mark for the explanation, so your change is good to me. ACK. I have a comment, if at this point groups have a reference to post_E, what is the purpose of the function memberof_del_dn_from_groups few lines above ?

Thanks for making me revisit this function.  So after further investigation I found that the group contains the post & pre DN's as members.  In the code we were only cleaning up the pre-entry DN, and not the post-entry DN.  So I changed the fix to do both.  Please review...

tbordaz: Indeed it is much simpler to read. ACK

PR #5192, opened 2022-03-07
Issue 5051 - RFE - ADSync flatten tree

### LINKED ISSUE #5051: Winsync fromAD preserve the tree structure of Active Directory
Hi,
doing a OneWay replication with WinSync It looks like 389 tries to preserve the tree structure of AD, when it's probably
actually not needed because of AD's always-unique RDN policy.
If OUs are not defined in destination D389 the replication will fails.

Firstyear:
Bug Description: ADSync would try to preserve the tree
structure within AD when replicated to 389. This often
led to sync failing unless the administrator manually
recreated the OU structure since we DID NOT create
those OU's automatically

Fix Description: Allow flattening the content of the AD
tree into the ds-subtree. Since AD guarantees that ALL
RDN's are unique across the directory, this is safe
because there can never be a conflict. Additionally
due to how this is implemented, it means that when an
AD entry is relocated in the AD tree, it does not move
in 389, but continues to receive updates and changes

fixes: https://github.com/389ds/389-ds-base/issues/5051

Review by: ???


tbordaz (approved):
This patch looks good to me. Note that we need to update  OID repos for the new attribute winSyncFlattenTree.

ldap/servers/plugins/replication/windows_protocol_util.c
```diff
@@ -4087,6 +4098,10 @@ map_entry_dn_inbound_ext(Slapi_Entry *e, Slapi_DN **dn, const Repl_Agmt *ra, int
              * which is normalized. Thus, we can use _normdn_.
              */
             new_dn = slapi_sdn_new_normdn_passin(new_dn_string);
+            slapi_log_err(SLAPI_LOG_REPL, windows_repl_plugin_name,
+                                  "map_entry_dn_inbound - %s - mapped entry to dn [%s]\n",
+                                  agmt_get_long_name(ra),
+                                  slapi_sdn_get_dn(new_dn));
```
mreynolds389: Indentation :-p

ldap/servers/plugins/replication/windows_private.c
```diff
@@ -889,6 +900,39 @@ windows_private_set_subtreepairs(const Repl_Agmt *ra, char **parray)
...
+    Dirsync_Private *dp;
+
+    slapi_log_err(SLAPI_LOG_TRACE, windows_repl_plugin_name, "=> windows_private_get_flatten_tree\n");
+
+    PR_ASSERT(ra);
+    dp = (Dirsync_Private *)agmt_get_priv(ra);
+    PR_ASSERT(dp);
+
+    slapi_log_err(SLAPI_LOG_TRACE, windows_repl_plugin_name, "<= windows_private_get_flatten_tree\n");
+
```
mreynolds389: I know this is how trace logging should be used, but trace logging has been abused & over-used, and it grinds the server to a halt when in use.  I'm not saying we need to change it, but I wonder if it should be plugin (or replication) logging instead?

I guess this is a future goal of mine to rewrite the debug logging.  For example, replication logging should cover all debug logging in replication - no mixing of debug log levels in a plugin or feature area.  And adding new logging levels for Password Policy (ticket already opened), SASL, etc.

ldap/schema/02common.ldif
```diff
@@ -154,6 +154,7 @@ attributeTypes: ( 2.16.840.1.113730.3.1.2357 NAME 'nsTaskCurrentItem' DESC 'Slap
 attributeTypes: ( 2.16.840.1.113730.3.1.2358 NAME 'nsTaskTotalItems' DESC 'Slapi Task total items' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2359 NAME 'nsTaskCreated' DESC 'Slapi Task creation date' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2375 NAME 'nsTaskWarning' DESC 'Slapi Task warning code' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.2385 NAME 'winSyncFlattenTree' DESC 'When set to on, will flatten tree structure in AD replication'  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 #
```
mreynolds389, line 158: OID is already in use in older branches (1.4.3)

     2.16.840.1.113730.3.1.2385     nsslapd-changelogcompactdb-time

It was removed in 1.4.4 when the changelog was moved into the main backend.  But this schema could be found during replication with mixed versions of DS.  

Instead can you use this oid:

    2.16.840.1.113730.3.1.2387

ldap/servers/plugins/replication/windows_private.c
Firstyear: This is how tdhe "other code" near by was, but yes, it's needlessly noisy IMO. I don't think we need these traces in here at all. But there is much more than just these two functions that create that noise.

Firstyear: So in other words - do I leave it to stay the same as the surrounding code, or do I do the cleanup and remove the traces from all the nearby functions?

Firstyear:
Applied the other feedback, needed to use 2.16.840.1.113730.3.1.2389 as the OID

ldap/servers/plugins/replication/windows_private.c
mreynolds389: You can leave it or just strip it out.  I was just raising the issue so we don't all forget about how bad trace function call logging is :-)

mreynolds389 (approved):
Ack

mreynolds389:
> Applied the other feedback, needed to use 2.16.840.1.113730.3.1.2389 as the OID

Ok, our internal OID database is definitely out of sink.  We'll work on that.

ldap/servers/plugins/replication/windows_private.c
Firstyear: I love deleting code, so I'll strip it then :)

Firstyear:
@mreynolds389 Okay, did the cleanup, can you double check before I merge?

mreynolds389:
Haha I didn't mean strip all the trace logging haha!  But I love it!  Ack

Issue #5193, opened 2022-03-07
Incomplete ruv occasionally returned  from ruv search


jchapma:
**Issue Description**
An intermittent condition occurs during cleanallruv (force) CI tests which results in an incomplete ruv being returned to the client. This generates an "IndexError" in lib389 because of the ruv->replica_purl being NULL. 

**Package Version and Platform:**
 - Platform: 5.6.6-300.fc32.x86_64 and 5.14.0-63.el9.x86_64
 - Package and version: 389-Directory/2.0.14 

**Steps to Reproduce**
1. Run cleanallruv CI test (This issue occurs occasionally, so you might need to run it more than once)
```
 sudo py.test -s -v dirsrvtests/tests/suites/replication/cleanallruv_test.py::test_clean_force
```

2. Notice CI test error
```
self = <lib389.replica.RUV object at 0x7f9ad7b5c2b0>
ruvs = ['{replicageneration} 6204409b000000010000', 
'{replica 1 ldap://julia32:39001} 6204409b000100010000 620440e90000000100...000 620440c5000000020000', '{replica 3 ldap://julia32:39003} 620440a6000100030000 620440c8000100030000', '{replica 4}']
logger = None

    def __init__(self, ruvs=[], logger=None):
        if logger is not None:
            self._log = logger
        else:
            self._log = logging.getLogger(__name__)
        self._rids = []
        self._rid_url = {}
        self._rid_rawruv = {}
        self._rid_csn = {}
        self._rid_maxcsn = {}
        self._rid_modts = {}
        self._data_generation = None
        self._data_generation_csn = None
        # Process the array of data
        for r in ruvs:
            pr = r.replace('{', '').replace('}', '').split(' ')
            if pr[0] == 'replicageneration':
                # replicageneration 5a2ffd0f000000010000
                self._data_generation = pr[1]
            elif pr[0] == 'replica':
                self._log.info("jc pr%s" % (pr))
                # replica 1 ldap://ldapkdc.example.com:39001 5a2ffd0f000100010000 5a2ffd0f000200010000
                # Don't add rids if they have no csn (no writes) yet.
                rid = pr[1]
                self._rids.append(rid)
>               self._rid_url[rid] = pr[2]
E               IndexError: list index out of range

/usr/local/lib/python3.8/site-packages/lib389-1.4.0.1-py3.8.egg/lib389/replica.py:853: IndexError
```
[...]

PR #5194, opened 2022-03-07
Issue 5193 - Incomplete ruv occasionally returned from ruv search

jchapma:
Bug Description:
An intermittent condition occurs during cleanallruv (force) CI tests
which results in an incomplete ruv being returned to the client. This
generates an "IndexError" in lib389 because of the ruv->replica_purl
being NULL.

Fix Description:
During an ruv search we iterate over the in memory ruv list. Skip over
an ruv if we detect ruv->replica_purl == NULL.

Fixes: https://github.com/389ds/389-ds-base/issues/5193

Reviewed by: ?

### DIFF
--- a/ldap/servers/plugins/replication/repl5_ruv.c
+++ b/ldap/servers/plugins/replication/repl5_ruv.c
@@ -1145,10 +1145,10 @@ ruv_to_valuearray(RUV *ruv)
     for (ruv_e = dl_get_first(ruv->elements, &cookie);
          NULL != ruv_e;
          ruv_e = dl_get_next(ruv->elements, &cookie)) {
-             /* Skip over an ruv with no purl */
-            if (NULL == ruv_e->replica_purl) {
-                continue;
-            }
+        /* Skip over an ruv with no purl */
+        if (NULL == ruv_e->replica_purl) {
+            continue;
+        }
 
         ruv_element_to_string(ruv_e, &bv, NULL, 0);
         value = slapi_value_new_berval(&bv);
--- a/ldap/servers/plugins/replication/repl5_ruv.c
+++ b/ldap/servers/plugins/replication/repl5_ruv.c
@@ -1145,6 +1145,10 @@ ruv_to_valuearray(RUV *ruv)
     for (ruv_e = dl_get_first(ruv->elements, &cookie);
          NULL != ruv_e;
          ruv_e = dl_get_next(ruv->elements, &cookie)) {
+             /* Skip over an ruv with no purl */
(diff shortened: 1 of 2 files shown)


progier389 (approved):
I would have rather done the test on lib389 side, but LGTM anyway ...

ldap/servers/plugins/replication/repl5_ruv.c
```diff
@@ -1145,6 +1145,10 @@ ruv_to_valuearray(RUV *ruv)
     for (ruv_e = dl_get_first(ruv->elements, &cookie);
          NULL != ruv_e;
          ruv_e = dl_get_next(ruv->elements, &cookie)) {
+             /* Skip over an ruv with no purl */
+            if (NULL == ruv_e->replica_purl) {
+                continue;
+            }
```
mreynolds389: Indentation is off :-)

mreynolds389 (changes requested):

mreynolds389:
> I would have rather done the test on lib389 side, but LGTM anyway ...

Perhaps lib389 should also be hardened as well.  CLI tools should never crash.

jchapma:
> > I would have rather done the test on lib389 side, but LGTM anyway ...
> 
> Perhaps lib389 should also be hardened as well. CLI tools should never crash.

Now that we check for ruv url in the server, lib389 should never hit this particular IndexError condition. But to be sure I added a similar check in lib389.

src/lib389/lib389/replica.py
```diff
@@ -848,7 +848,10 @@ def __init__(self, ruvs=[], logger=None):
                 # Don't add rids if they have no csn (no writes) yet.
                 rid = pr[1]
                 self._rids.append(rid)
-                self._rid_url[rid] = pr[2]
+                try:
+                    self._rid_url[rid] = pr[2]
+                except IndexError:
+                    continue
```
mreynolds389, line 855: Don't "continue" here because we've already added the rid to Replica "_rids".  Instead set it to some default value, like what we do with other settings when they are missing.  Or, remove the rid from the "_rids" list before continuing.

jchapma: > Don't "continue" here because we've already added the rid to Replica "_rids". Instead set it to some default value, like what we do with other settings when they are missing. Or, remove the rid from the "_rids" list before continuing.

I would prefer to remove the rid as I'm afraid setting a default url might break something down the line.

mreynolds389: Agreed

mreynolds389 (approved):
ACK

PR #5198, opened 2022-03-11
Issue 5197 - Build break in lib389 with INSTALL_PREFIX

### LINKED ISSUE #5197: Broken build in make lib389-install when using INSTALL_PREFIX
**Issue Description**
lib389 build is now failing when using custom directory
(Attempt to overwrite system library)

 - Platform: Fedora
 - Package and version: 389-ds-base-1.4.4.4-20200721git5d41dc5a4.fc32.x86_64

**Steps to Reproduce**
1. Set and export INSTALL_PREFIX variable so more directory you have write access
2. Configure a build with prefix directory: i.e configure ... --prefix=$INSTALL_PREFIX
3. Run: make lib389-install
5. See error

**Expected results**
Build should be successful and data located in INSTALL_PREFIX directory

**Additional context**
IMHO the issue is that src/lib389/lib389 is ignoring the INSTALL_PREFIX variable
.Note: it is a regression as this was working a few weeks ago ...

progier389:
Issue: make lib389 and lib389-install tries to install binaries in / instead of in $INSTALL_PREFIX
 so the build fails when following https://www.port389.org/docs/389ds/development/building.html#building-into-a-prefix-location-non-standardnot-root instructions

Solution: Add the prefix to the data 
(Note there is a trick: to get same result than previous builds and the other parts of the Makefile /usr should be removed when using a prefix)
 
 Issue: [5197](https://github.com/389ds/389-ds-base/issues/5197)
 
 Reviewd by: ?

### DIFF
--- a/src/lib389/setup.py
+++ b/src/lib389/setup.py
@@ -13,7 +13,8 @@
 #
 
 from setuptools import setup, find_packages
-from os import path
+from os import path, environ
+from sys import argv
 import build_manpages as bm
 if bm.__version__ < '2.1':
     from build_manpages import build_manpages as bm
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
LGTM

PR #5201, opened 2022-03-14
Issue 5200 - dscontainer should use environment variables with DS_ prefix

### LINKED ISSUE #5200: `dscontainer` should use environment variables with `DS_` prefix
**Issue Description**
`dscontainer` accepts several environment variables, but some of them don't have `DS_` prefix, such as `ERRORLOG_LEVEL` and `SUFFIX_NAME`. It would be good to use a uniform namespaced notation to avoid generic names that can possibly conflict with other environment variables (for example, when DS runs in a pod with other containers, that can also use these generic variable names).

Additionally, `DS_MEMORY_PERCENTAGE` is no longer applicable when server uses MDB. We should log an error/warn message to notify the user.

vashirov:
Description:
dscontainer accepts several environment variables, but some of them
don't have DS_ prefix, such as ERRORLOG_LEVEL and SUFFIX_NAME.
It would be good to use a uniform namespaced notation to avoid
generic names that can possibly conflict with other environment
variables (for example, when DS runs in a pod with other containers,
that can also use these generic variable names).

Additionally, DS_MEMORY_PERCENTAGE is no longer applicable when server
uses MDB. We should log an error/warn message to notify the user.

Fixes: https://github.com/389ds/389-ds-base/issues/5200

Reviewed by: ???

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -43,6 +43,7 @@ from lib389.passwd import password_generate
 from lib389.nss_ssl import NssSsl, CERT_NAME
 from lib389.paths import Paths
 from lib389.config import LDBMConfig
+from lib389.utils import get_default_db_lib
 from lib389._constants import (
     DSRC_CONTAINER,
     CONTAINER_TLS_SERVER_KEY,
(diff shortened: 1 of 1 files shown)


src/lib389/cli/dscontainer
```diff
@@ -89,14 +90,17 @@ def _begin_environment_config():
...
-        log.debug("Setting LDBM Autotune Percentage to: %s", autotune_pct)
-        ldbmconfig = LDBMConfig(inst)
-        ldbmconfig.set("nsslapd-cache-autosize", str(autotune_pct))
+        if get_default_db_lib() == "mdb":
+            log.error("DS_MEMORY_PERCENTAGE is present, but setting entry cache is not supported for MDB")
+        else:
+            try:
+                autotune_pct = int(autotune_pct)
+            except:
+                log.error("Invalid DS_MEMORY_PERCENTAGE - resetting to system default value")
```
vashirov: I'm not sure if this should be an error or warn, feedback is welcome.

mreynolds389: It's not technically breaking anything, so I think it should be a warning

Firstyear: Agreed, should be a warning.

Firstyear (approved):

Issue #5214, opened 2022-03-17
CI Test tests/suites/replication/virtual_attribute_replication_test.py


tbordaz:
**Issue Description**
tests/suites/replication/virtual_attribute_replication_test.py is failing quite frequently. The Reason is that an update done on supplier is not yet propagated to the consumer when the consumer is restarted. The update is to delete a COS and so revers nsslapd-ignore-virtual-attrs=on at restart

**Package Version and Platform:**
 master only

**Steps to Reproduce**
run tests/suites/replication/virtual_attribute_replication_test.py::test_vattr_on_cos_definition_with_replication

**Expected results**
It should not fail on master

tbordaz:
d2cc9401e..55cbd275f  master

PR #5215, opened 2022-03-17
Issue 5214 - CI Test tests/suites/replication/virtual_attribute_repli…

tbordaz:
…cation_test.py

Bug description:
	The test test_vattr_on_cos_definition_with_replication creates
	a COS in a supplier/consumer topology.
        With that virtual attribute, nsslapd-ignore-virtual-attrs gets
        the value OFF.
        Then it deletes the COS on the supplier but does not wait
        that the DEL is propagated to the consumer before restarting
        the consumer.
        The consequence is that at startup, consumer may keep
        nsslapd-ignore-virtual-attrs=OFF

Fix description:
        Wait that replication is in sync before restarting the consumer

relates: #5214

Reviewed by:

Platforms tested: F34


mreynolds389 (approved):
ack

Issue #5202, opened 2022-03-14
CLI - dump-changelog does not work on 1.4.3


mreynolds389:
**Issue Description**

In 1.4.3 there is a missing import for "logging", and the restore changelog has a copy and paste error where it does a changelog export instead of an import.

There are also additional accidental output added to the end of the dumped changelog (this probably impacts Master branch - need to check).

PR #5216, opened 2022-03-18
Issue - CLI - dump-changelog not working correctly

mreynolds389:
Bug Description:

There are several erros when trying to use dump-changelog.  First there
is an import error with "logging", second the restore changelog
accidentally calls cl2ldif instead of ldif2cl, third we log the
command result output in the exported LDIF, fourth we don't log the
correct output file name in the LDIF file.

Fix Description:  Import logging, call the correct import command, pass
the custom output file name down the correct function so we can write it
to the file, and last remove the log handler so we don't log unexpected
messages in the LDIF file.

relates: https://github.com/389ds/389-ds-base/issues/5202

### DIFF
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -10,6 +10,7 @@ import re
 import os
 import json
 import ldap
+import logging
 import stat
 from shutil import copyfile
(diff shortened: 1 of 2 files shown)


droideck (approved):
LGTM!

P.S. tested on RHEL 8

PR #5219, opened 2022-03-21
Issue 5218 - double-free of the virtual attribute context in persiste…

tbordaz:
…nt search

description:
	A search is processed by a worker using a private pblock.
	If the search is persistent, the worker spawn a thread
	and kind of duplicate its private pblock so that the spawn
        thread continue to process the persistent search.
	Then worker ends the initial search, reinit (free) its private pblock,
        and returns monitoring the wait_queue.
	When the persistent search completes, it frees the duplicated
	pblock.
	The problem is that private pblock and duplicated pblock
        are referring to a same structure (pb_vattr_context).
        That can lead to a double free

Fix:
	When cloning the pblock (slapi_pblock_clone) make sure
	to transfert the references inside the original (private)
	pblock to the target (cloned) one
        That includes pb_vattr_context pointer.

Reviewed by: Mark Reynolds, James Chapman, Pierre Rogier (Thanks !)

### DIFF
--- a/ldap/servers/slapd/pblock.c
+++ b/ldap/servers/slapd/pblock.c
@@ -292,6 +292,12 @@ _pblock_assert_pb_deprecated(Slapi_PBlock *pblock)
     }
 }
 
+/* It clones the pblock
+ * the content of the source pblock is transfered
+ * to the target pblock (returned)
+ * The source pblock should not be used for any operation
+ * it needs to be reinit (slapi_pblock_init)
+ */
 Slapi_PBlock *
 slapi_pblock_clone(Slapi_PBlock *pb)
 {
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
LGTM

ldap/servers/slapd/pblock.c
```diff
@@ -312,28 +318,32 @@ slapi_pblock_clone(Slapi_PBlock *pb)
     if (pb->pb_task != NULL) {
         _pblock_assert_pb_task(new_pb);
         *(new_pb->pb_task) = *(pb->pb_task);
+        memset(pb->pb_task, 0, sizeof(slapi_pblock_task));
```
Firstyear, line 321: This doesn't make sense. If we are cloning the pblock, we need to clone all it's elements. This is just zeroing them all instead, so pblock clone is now effectively pblock new.

Issue #5186, opened 2022-03-02
UI crashes because invalid SASL mapping regex was saved


mreynolds389:
**Description of problem**

Editing an existing SASL mapping, you can change the SASL mapping regular expression and enter an invalid mapping. You can then save it without problem.

But then entering a random expression in 'test regex' will make cockpit crash 

**How reproducible**

Steps to Reproduce:
1.Go to 'Server settings' / 'SASL settings and mappings'
2.In 'SASL mappings' table, click on the 3 points on the right and Edit mapping
3. Change the SASL mapping Regular Expression to something invalid
4. Click 'Save mapping'
5. Now enter a random expression in 'Test regex' and click 'Test it'

**Actual results**

cockpit crashes : Ooops!  appears in red a the top right

mreynolds389:
bdeeca56c..9f410afa6  389-ds-base-2.0 -> 389-ds-base-2.0
d84894150..33452a7c5  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
7a9c8e3ad..26b2d9b07  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2017411

mreynolds389:
b7fd028e5..b5c24f5a6  389-ds-base-2.1 -> 389-ds-base-2.1
e6431d959..d12f31d52  389-ds-base-2.0 -> 389-ds-base-2.0
feb4a55d8..350f07357  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
64001477b..27f10e6c5  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5220, opened 2022-03-21
Issue 5186 - UI - Fix SASL Mapping regex test feature

mreynolds389:
Description:

If the regex is invalid you are still able to click on the "Test"
button which then crashes the UI.  THe Test button needs to be disabled
if the regex is invalid

relates: https://github.com/389ds/389-ds-base/issues/5186


jchapma (approved):
Had a play around with your fix and I failed to break it. LGTM you have my ack.

droideck (approved):
LGTM too

Issue #5221, opened 2022-03-21
user still has privedlges and can login even if password is expired


mreynolds389:
**Issue Description**

An LDAP client can bind to DS, and even if an error 49 (expired password) is returned, the session is still "authenticated" and the user can proceed as "normal".  When the user's password is fully expired (grace logins used up) the user needs to be considered anonymous (not authenticated).

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2064781

mreynolds389:
00385645f..b7fd028e5  389-ds-base-2.1 -> 389-ds-base-2.1
3553bce44..e6431d959  389-ds-base-2.0 -> 389-ds-base-2.0
cf2ad7345..feb4a55d8  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
269dfcfc7..64001477b  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
91c4f3e78..1e0a3e0aa  389-ds-base-1.3.10 -> 389-ds-base-1.3.10

tbordaz:
Fix covscan (https://github.com/389ds/389-ds-base/pull/5359)

05c0a45f9..37101d995  master
6280e2f4c..7b6ff508f  389-ds-base-2.1
0e5044ece..d8f1a7c74  389-ds-base-2.0
f519ef753..8ba2ea21f  389-ds-base-1.4.4
93ff55814..970adbe74  389-ds-base-1.4.3
78d2ef45d..f3f930863  389-ds-base-1.3.10

PR #5222, opened 2022-03-21
Issue 5221 - User with expired password can still login with full privledges

mreynolds389:
Bug Description:

A user with an expired password can still login and perform operations
with its typical access permissions.  But an expired password means the
account should be considered anonymous.

Fix Description:

Clear the bind credentials if the password is expired

relates: https://github.com/389ds/389-ds-base/issues/5221


progier389 (approved):
LGTM

PR #5224, opened 2022-03-22
Issue 5217 - Simplify instance creation and administration by non root

### LINKED ISSUE #5217: Simplify instance creation and administration by non root user
Customer reports that open ldap allowed a non root user to create instance (for example to run tests) and complained that it is not possible with 389DS. (In fact it is possible but that is cumbersome)

I propose to add a dsconf option to prepare a subtree and wrappers to the  dscreate, dsctl and dsconf command 
 to be able to create/run/administrate 389ds instances.
The idea is very simple: copy what is needed from the standard package installation (as root) to generate the subtree.
  optionally generate wrapper to the standard administration commands (Running the standard command after setting 
   the PREFIX and INSTALL_PREFIX environment variables) and update the template files (typically the dse.ldif and the inf files) so  that dscreate defaults get adjusted for a prefixed install. 

**Limitation**
1. requires that packages get previously installed as root (by using: dnf install 389-ds-base) 
2. selinux not handled (But as we are running as an user in a subtree owned by the user, this should not be an issue)
3. Rust plugins are disabled. 
4. snmp is not supported.

I wondered if it would be possible to fully install 389ds in a prefixed variables (to be able to handle different versions of 389ds)
as non root but handling the libraries dependencies became tricky so I rather use a simple solution

progier389:
Goal: 
    Allow any user to easily create and administrate directory server instances

Solution:
    Add a ds-root subcommand to dscreate:
```
       ds-root             Prepare a root directory in which non root user can create, run and administrate instances.
       usage: dscreate ds-root [-h] root_dir [bin_dir]

       positional arguments:
         root_dir    A directory that will be used as virtual root.
         bin_dir     A directory in which administration wrappers are installed. (Should be in PATH)

      Example of install by a non root user:
          PATH=$HOME/bin:$PATH # Should be in .profile
          dscreate ds-root $HOME/mydsroot $HOME/bin
          hash -d dscreate # bash command to insure new dscreate wrapper will be used
          dscreate interactive
          # Note: Make sure to use non priviledged port number (i.e > 1000)

```
   
   This command prepare a subtree root directory for the directory server instances) and optionally a set of wrapper around administrative commands.
   

Issue: [5217](https://github.com/389ds/389-ds-base/issues/5217)
Reviewed by: mreynolds

### DIFF
--- a/dirsrvtests/tests/suites/setup_ds/dscreate_test.py
+++ b/dirsrvtests/tests/suites/setup_ds/dscreate_test.py
@@ -7,6 +7,11 @@
 # --- END COPYRIGHT BLOCK ---
 import sys
 import pytest
+import subprocess
+import logging
+import pwd
+import re
+from tempfile import TemporaryDirectory
 from lib389 import DirSrv
 from lib389.cli_base import LogCapture
 from lib389.instance.setup import SetupDs
@@ -18,10 +23,17 @@ from lib389.utils import ds_is_older
 pytestmark = [pytest.mark.tier0,
               pytest.mark.skipif(ds_is_older('1.4.1.2'), reason="Needs a compatible systemd unit, see PR#50213")]
 
+DEBUGGING = os.getenv('DEBUGGING', False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
 INSTANCE_PORT = 54321
 INSTANCE_SECURE_PORT = 54322
 INSTANCE_SERVERID = 'standalone'
-DEBUGGING = True
+#DEBUGGING = True
 
 MAJOR, MINOR, _, _, _ = sys.version_info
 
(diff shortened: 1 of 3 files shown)


src/lib389/cli/dscreate
```diff
@@ -40,9 +40,13 @@ interactive_parser.set_defaults(func=cli_instance.instance_create_interactive)
 template_parser = subparsers.add_parser('create-template', help="Display an example inf answer file, or provide a file name to write it to disk.")
 template_parser.add_argument('--advanced', action='store_true', default=False,
     help="Add advanced options to the template - changing the advanced options may make your instance install fail")
-template_parser.add_argument('template_file', nargs="?", default=None, help="Write example template to this file")
+template_parser.add_argument('template_file', nargs="?", default='Foo', help="Write example template to this file")
```
progier389: Ouch a change related to a test stayed by mistake. default should be None

mreynolds389:
Can you use a different term for "prepare-subtree"?   It makes me think it's a DS suffix "subtree", and not a FS directory.  It could be confusing for others.

dirsrvtests/tests/suites/setup_ds/dscreate_test.py
```diff
@@ -115,3 +127,133 @@ def test_setup_ds_minimal(topology):
...
+                '}',
+                'trap savedir exit',
+                'trap savedir err',
+            )
+        else:
+            dbg_script = ( )
+        # Export current path and python path in test script
+        # to insure that right binary/libraries are used
+        path = os.environ['PATH']
+        pythonpath = os.environ['PYTHONPATH']
```
progier389: Should use os.getenv('PYTHONPATH', '') to handle the case where PYTHPATH is not defined.

mreynolds389:
> > Can you use a different term for "prepare-subtree"? It makes me think it's a DS suffix "subtree", and not a FS directory. It could be confusing for others.
> 
> The sub command and options names could obviously be changed. For my part I do not like prepare-virtual-root (Too long for my taste) nor prepare-prefix (Meaningless for anyone that has not tried to build the product) (That said, it is a mild dislike not an hard veto .... ) prepare-directory is even more confusing than prepare subtree and folder is a window terminology rather than a linux one Well finally prepare-virtual-root seems the better among these. Anyway iI am open to any suggestion!

What about:

- prepare-root
- prepare-root-dir
- prepare-ds
- prepare-install
- prepare-custom
- prepare-non-root

Just some ideas.  Maybe others on the team will have some input...

vashirov:
We have a verb `create` in `dscreate`, so I'd drop `prepare` completely. 
My candidates:
* `dscreate inst-path` or `dscreate instance-path`
* `dscreate inst-dir` or `dscreate instance-dir`

progier389:
I am now thinking about simply using:  virtual-root because "dscreate virtual-root ..."  is meaningful by itself.
Any comments ?

progier389:
> We have a verb `create` in `dscreate`, so I'd drop `prepare` completely. My candidates:
> 
> * `dscreate inst-path` or `dscreate instance-path`
> * `dscreate inst-dir` or `dscreate instance-dir`

Seeing your update just after posting mine !
I see that we are both thinking that we should drop the "prepare" verb, 
but I do not like the "instance" part  because several instances may be created using the same prefix install.

progier389:
Thinking more about @vashirov's proposal I hesitate now between:
- dscreate virtual-root
- dscreate instances-root

vashirov:
It's not immediately clear to me what 'virtual root' might be, it brings memories of virtual ldap directories :) 
And I don't want to assume how many instances would be created, so I'd shorten `instances-root` to `inst-root` which can also have a meaning of 'installation root'.

mreynolds389:
> Ok, so lets drop 'virtual'. server-root is a good idea and along the same line of thought, there is also: ds-root

I like ds-root too.  Short and sweet :-)

progier389:
Committed new version with ds-root sub command.

src/lib389/cli/dscreate
```diff
@@ -40,9 +51,13 @@ interactive_parser.set_defaults(func=cli_instance.instance_create_interactive)
 template_parser = subparsers.add_parser('create-template', help="Display an example inf answer file, or provide a file name to write it to disk.")
 template_parser.add_argument('--advanced', action='store_true', default=False,
     help="Add advanced options to the template - changing the advanced options may make your instance install fail")
-template_parser.add_argument('template_file', nargs="?", default=None, help="Write example template to this file")
+template_parser.add_argument('template_file', nargs="?", default='None', help="Write example template to this file")
 template_parser.set_defaults(func=cli_instance.instance_example)
 
+subtree_parser = subparsers.add_parser('ds-root', help="Prepare a root directory in which non root user can create, run and administrate instances.")
```
mreynolds389: This grammar is a little off,  "administrate" should probably be "administer"
Thanks!

mreynolds389 (approved):
Minor grammar issue, but ack from me

progier389:
Fixed the grammar issue.

Issue #5225, opened 2022-03-22
UI - can not exit autotuned cache settings


mreynolds389:
**Issue Description**

An invalid comparator was used to check if cache autotuning was enabled, which resulted in the UI thinking it is ALWAYS set.  This makes it impossible to manually set the entry cache from the UI.

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2066790

mreynolds389:
b5c24f5a6..c93463ed7  389-ds-base-2.1 -> 389-ds-base-2.1
d12f31d52..f4a80df67  389-ds-base-2.0 -> 389-ds-base-2.0
350f07357..e39d17040  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
27f10e6c5..5a18bbb45  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5226, opened 2022-03-22
Issue 5225 - UI - impossible to manually set entry cache

mreynolds389:
Bug description:  The UI thinks cache auto-tuning is always set which
prevents the user from manaully setting the entry cache.

Fix Description:  The UI was comparing a value to an array, which always
returned false and kept the UI thinking autotunning was set.

relates: https://github.com/389ds/389-ds-base/issues/5225

### DIFF
--- a/src/cockpit/389-console/package-lock.json
+++ b/src/cockpit/389-console/package-lock.json
@@ -5591,9 +5591,9 @@
       }
     },
     "node_modules/minimist": {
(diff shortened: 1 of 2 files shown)


mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2066790

droideck (approved):
LGTM!

PR #5243, opened 2022-03-30
Issue 5242- Craft message may crash the server

tbordaz:
Bug description:
	A craft request can result in DoS

Fix description:
	If the server fails to decode the ber value
	then return an Error

relates: 5242

Reviewed by: Pierre Rogier

Platforms tested:  F34


mreynolds389 (approved):
LGTM

PR #5246, opened 2022-04-01
Issue 5245 - RFE - default add krb schema.

### LINKED ISSUE #5245: Default install of krb kdc schema
We ship a number of schemas as examples, including 60kerberos.ldif. This is used when someone wants to store their kdc db into ldap. To make it easier for people to use this configuration, we should install it by default into /usr/share/dirsrv/schema

Firstyear:
Bug Description: We ship a number of schemas as examples, including 60kerberos.ldif. This is used
when someone wants to store their kdc db into ldap. To make it easier for people to use this
configuration, we should install it by default into /usr/share/dirsrv/schema

Fix Description: Enable it as a default

fixes: https://github.com/389ds/389-ds-base/issues/5245

Review by: ???

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -634,7 +634,6 @@ sampledata_DATA = $(srcdir)/ldap/ldif/Ace.ldif \
 	$(srcdir)/ldap/schema/60changelog.ldif \
 	$(srcdir)/ldap/schema/60inetmail.ldif \
 	$(srcdir)/ldap/schema/60krb5kdc.ldif \
-	$(srcdir)/ldap/schema/60kerberos.ldif \
 	$(srcdir)/ldap/schema/60nis.ldif \
 	$(srcdir)/ldap/schema/60qmail.ldif \
 	$(srcdir)/ldap/schema/60radius.ldif \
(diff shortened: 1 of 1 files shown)


tbordaz:
There will be collision between 389-ds 60kerberos.ldif and freeipa 60kerberos.ldif. Freeipa is a superset of 389-ds.
In addition, freeipa install script may have to take into account this new schema file.

Checking with the ipa team

tbordaz:
Following custom schema [recommendations](https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/deployment_guide/designing_the_directory_schema-customizing_the_schema#Customizing_the_Schema-Creating_Custom_Schema_Files), the custom definitions in IPA will likely be overwritten by default definitions.

Initial thoughts: On fresh install, IPA installer should remove /usr/share/dirsrv/schema/60kerberos.ldif. On upgrade IPA install should do the same and 389 upgrade function should do the same.

@Firstyear, this change has a significant impact both in dev and qe. Is it justified ?
Why not simply documenting that customer with their kdc ldap have to copy the file from 'data' to 'schema' ?

PR #5248, opened 2022-04-04
Issue 5247 - BUG - Missing attributes in samba schema

### LINKED ISSUE #5247: Missing attributes in samba schema
When using 389ds as an authentication backend for samba, a specific set of attributes and objectclasses is required. Some of these were missing which currently prevents correct operation of the backend.

Firstyear:
Bug Description: When using 389ds as an authentication backend for samba, a specific set of
attributes and objectclasses is required. Some of these were missing which currently prevents
correct operation of the backend.

Fix Description: Update to the latest sambav3 schema.

fixes: https://github.com/389ds/389-ds-base/issues/5247

Review by: ???

### DIFF
--- a/ldap/schema/60samba3.ldif
+++ b/ldap/schema/60samba3.ldif
@@ -1,566 +1,190 @@
-# 60samba3.ldif - Samba v3 schemas
-################################################################################
-#
+## schema file for Fedora/RedHat Directory Server
+##
+## NOTE: this file can be copied as 60samba.ldif into your instance schema
+## directory:
+## cp samba-schema-FDS.ldif /etc/dirsrv/slapd-<instance-name>/schema/60schema.ldif
+##
+## Schema for storing Samba user accounts and group maps in LDAP
+## OIDs are owned by the Samba Team
+##
+## Prerequisite schemas - uid         (cosine.schema)
+##                      - displayName (inetorgperson.schema)
+##                      - gidNumber   (nis.schema)
+##
+## 1.3.6.1.4.1.7165.2.1.x - attributeTypes
+## 1.3.6.1.4.1.7165.2.2.x - objectClasses
+##
+## Printer support
+## 1.3.6.1.4.1.7165.2.3.1.x - attributeTypes
+## 1.3.6.1.4.1.7165.2.3.2.x - objectClasses
+##
+## Samba4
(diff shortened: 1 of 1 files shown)


ldap/schema/60samba3.ldif
```diff
@@ -1,566 +1,190 @@
...
+##
+## 1.3.6.1.4.1.7165.2.1.x - attributeTypess
+## 1.3.6.1.4.1.7165.2.2.x - objectClasseses
+##
+## Printer support
+## 1.3.6.1.4.1.7165.2.3.1.x - attributeTypess
+## 1.3.6.1.4.1.7165.2.3.2.x - objectClasseses
+##
+## Samba4
+## 1.3.6.1.4.1.7165.4.1.x - attributeTypess
```
droideck: I'm not sure if it's done intentionally or because of some paste error, but I think it's a mistake to add 's' here (and 'es' to objectClasses)
Example from Samba repo - https://git.samba.org/?p=samba.git;a=blob_plain;f=examples/LDAP/samba.schema

droideck (commented):
In the commit message, you say 'Fix Description: Update to the latest sambav3 schema.'
But I thought we already support Samba v3 (as it's in a file header).
Also, I see Samba4 mentioned in the new lines.

Could you please clear this confusion for me?

ldap/schema/60samba3.ldif
Firstyear: Lol, I actually copied this direct from the samba repo, but from the 389 compatible schema. I can fix these typos up though.

Firstyear:
> In the commit message, you say 'Fix Description: Update to the latest sambav3 schema.' 

So we had a schema to support samba using an ldap authentication backend in sambav3.ldif. However, the samba project has added more attributes and objectclasses since, and those are missing and current cause our schema to not work. 

> But I thought we already support Samba v3 (as it's in a file header). Also, I see Samba4 mentioned in the new lines.

Yes, because samba is at version 4 now, but in fileserver mode with the ldap backend it's more "samba 3". There is a lot of weird history in the samba project that caused this .... 

>
> 
> Could you please clear this confusion for me?

droideck (approved):
Thanks for the explanation!
LGTM!

Firstyear:
```
   b7f891097..e988a3f55  389-ds-base-2.1 -> 389-ds-base-2.1
```

PR #5251, opened 2022-04-06
Issue 5241 - UI - Add account locking missing functionality

### LINKED ISSUE #5241: UI - Add account locking missing functionality
**Issue Description**
Add a missing ability to lock and unlock Roles/Accounts to LDAP Editor:

- Add Lock/Unlock button to Dropdown when choose entry in Table or Directory Tree;
- Add 'Locked' indicator with 'lock' icon beside entry name;
- Call account functions from dsconf for these actions;

droideck:
Description: Add a missing ability to lock and unlock Roles/Accounts to LDAP Editor.
Add Lock/Unlock button to Dropdown when choose entry in Tree View tab;
Add 'Locked' indicator with 'lock' icon beside entry name;
Add the same functionality to Search tab. But also add Enable Locking button
which enables this functionality (disabled by default). We need it because
it vastly impacts performance.
Hence, don't add the functionality to Table View tab because it runs the
same kind of search as Search tab but on the whole directory tree.

Fixes: https://github.com/389ds/389-ds-base/issues/5241

Reviewed by: ?


mreynolds389:
Can you add the lock icon to the table view?  Then when searching or browsing the table view you can see which entries are locked.  I did something similar for ldapSubentry's (I added an info icon to make it stand out).

What about changing the "lock" color to "red" instead of blue?  Just a thought, not requiring it. :-)

Also, when locking/unlocking the modal shows a vertical scroll bar.  Can you clean that up?  There is no need for a scroll bar, maybe just add some margins to the text field?

droideck:
> Can you add the lock icon to the table view? Then when searching or browsing the table view you can see which entries are locked. I did something similar for ldapSubentry's (I added an info icon to make it stand out).

I've added the 'Locking Disabled' switch o Table View. Please, check if you'll like the design.
I'm also not sure how to name it (if we decide to keep it). So any suggestions are welcome. :)
 
> 
> What about changing the "lock" color to "red" instead of blue? Just a thought, not requiring it. :-)

I've changed: Triangle (warning) color to Yellow; Circle (error) to Red; and the Lock color was left as Blue because I think it's better to distinguish these three.

> 
> Also, when locking/unlocking the modal shows a vertical scroll bar. Can you clean that up? There is no need for a scroll bar, maybe just add some margins to the text field?

Fixed.

Thanks!

mreynolds389:
The "lock" icon is still blue.  In the table view there is still no way to tell the entry is locked, and there is no way to unlock it

mreynolds389:
> The "lock" icon is still blue. In the table view there is still no way to tell the entry is locked, and there is no way to unlock it

Sorry I see you still want to keep the lock blue, that is fine.  But we need something better in the table view.

mreynolds389:
> > The "lock" icon is still blue. In the table view there is still no way to tell the entry is locked, and there is no way to unlock it
> 
> Sorry I see you still want to keep the lock blue, that is fine. But we need something better in the table view.

Also I don't see this "lock switch" you are talking about in the table view.

mreynolds389:
> > > The "lock" icon is still blue. In the table view there is still no way to tell the entry is locked, and there is no way to unlock it
> > 
> > 
> > Sorry I see you still want to keep the lock blue, that is fine. But we need something better in the table view.
> 
> Also I don't see this "lock switch" you are talking about in the table view.

And....  :-)

In the search page the "enable locking" switch is confusing to me.  I think it should be "show" or "display" locking not "enabling" it.  That's all for now ;-)

mreynolds389:
Ok I have an idea for the table view.  We can drop the icon from the row, and the "switch".

When you expand the row to see the entry details we can then call "account status".  While that is running we could add a spinner in front of a new field in the expanded row:

"Account Locked:  <Spinner>"
...
"Account Locked:  <LOCKED>"

Because even in the tree view, you still have to "open" the entry to see if its locked.  So in the table view we can do that processing once the entry is opened(row expanded).  Something like that - just an idea...

droideck:
I've made the Table view act completely similar to the Tree view.
And Search tab will be for fine-grained searching as discussed on IRC.

mreynolds389:
OK, two issues.  [1]  I meant to bring this up before.  In the Tree view I can lock the root suffix "dc=example,dc=com" or any entry I want, but in the table view I can not.  Not sure you should be able to lock the root suffixes. I thought you should only be able to lock users and roles.

[2] In Table view, I think the Padlock should either be duplicated in the expanded row (next to dn), or _only_ displayed in the expanded row.

[3?] Maybe this is for another ticket, but in the old java console you could search/filter the database by users who are locked or unlocked.  This would be a nice feature in the search page.

mreynolds389:
> [3?] Maybe this is for another ticket, but in the old java console you could search/filter the database by users who are locked or unlocked. This would be a nice feature in the search page.

I'm going to open a new ticket for this, because I have some interesting ideas on how to incorporate this (and other specialty searches)...

droideck:
> OK, two issues. [1] I meant to bring this up before. In the Tree view I can lock the root suffix "dc=example,dc=com" or any entry I want, but in the table view I can not. Not sure you should be able to lock the root suffixes. I thought you should only be able to lock users and roles.

Good catch! I've also fixed the CLI part.

> [2] In Table view, I think the Padlock should either be duplicated in the expanded row (next to dn), or _only_ displayed in the expanded row.

Sounds good. Done.

> I'm going to open a new ticket for this, because I have some interesting ideas on how to incorporate this (and other specialty searches)...

Looking forward to seeing what's the idea:)

mreynolds389 (approved):
ACK!

mreynolds389:
Great job!  Very happy with this.

PR #5253, opened 2022-04-11
lib389 prerequisite for ansible-ds

progier389:
Some neeeded changes to lib389 to support ansible-ds:
-   DSEldif class:
    -       Allow support of custom path
    -       Add globalSubstitute function that replace a pattern on every lines
       
Issue [1](https://github.com/389ds/ansible-ds/issues/1)

Reviewed by:


droideck (approved):
LGTM!

Issue #5252, opened 2022-04-11
During DEL, vlv search can erroneously return NULL candidate


tbordaz:
**Issue Description**
During a vlv search, the vlv index iterates from starting to last records. During that iteration the last record may change if for example deleting entries reduces the number of indexed records.
In such case, not retrieving as much records as expected, the candidate list is NULL and errors are logged.

This is not correct because it is normal that the effective number of records fluctuates during iteration. It should not be processed as an error.

If the problem occurs the following errors are logged
<pre>
 ERR - vlv_build_idl - Can't follow db cursor (err -12797)
 ERR - vlv_find_index_by_filter_txn - vlv find index: err 1
</pre>

**Package Version and Platform:**
all versions

**Steps to Reproduce**
See BZ. In short

- create a vlv index
- provision a large number of entries matching the vlv index
- run several vlv search and at the same time remove entries
- Check that the messages (this error does *not* trigger the failure of the search, just candidates=null)

**Expected results**
Those error messages should not be logged

PR #5256, opened 2022-04-11
Issue 5252 - During DEL, vlv search can erroneously return NULL candi…

tbordaz:
…date

Bug description:
	vlv_build_idl builds a candidate list from a starting to ending
	points. By the time the ending point is computed (like in
        vlvIndex_get_indexlength) and is finally used, the actual
        number of record may reduced. In such case the ending point
        does no longer exist, and cursor(NEXT) hits the end
        of the index (DBI_RC_NOTFOUND).
        The if that occurs, the current candidate list is valid
        and should be returned.

Fix description:
	On DBI_RC_NOTFOUND, exit from the start to end loop
        without error.

relates: https://github.com/389ds/389-ds-base/issues/5252

Reviewed by:

Platforms tested:


progier389 (approved):
LGTM

Issue #5257, opened 2022-04-11
nsslapd-subtree-rename-switch option broken


mreynolds389:
**Issue Description**

Trying to disable subtree rename breaks the server in several ways, mainly around import/export, but also in replication.

This setting is going away in 389-ds-base-2.x (as it doesn't work with LMDB), but it should be working in 1.4.x (for RHEL 8).

PR #5258, opened 2022-04-11
Issue 5257 - nsslapd-subtree-rename-switch option broken

mreynolds389:
Bug Description:

Disabling subtree renaming breaks imports and replication.

Fix Description:

The ID list management changed, and in most cases we need to check if
b_nids is greater than zero.  With this setting we also swqitch to
entrydn (from entryrdn), and we were not handling tobmstones correctly
when using the entrydn index.

fixes: https://github.com/389ds/389-ds-base/issues/5257

### DIFF
--- a/dirsrvtests/tests/suites/replication/moddn_off_and_entrydn_test.py
+++ b/dirsrvtests/tests/suites/replication/moddn_off_and_entrydn_test.py
@@ -1,0 +1,107 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import pytest
+import os
+import time
+from lib389._constants import (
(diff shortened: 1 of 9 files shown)


progier389 (approved):
LGTM 
looks like you had fun with the tombstones ! -;)

dirsrvtests/tests/suites/replication/moddn_off_and_entrydn_test.py
```diff
@@ -0,0 +1,98 @@
...
+import time
+from lib389._constants import (
+    DN_CONFIG_LDBM, DEFAULT_SUFFIX, DEFAULT_BENAME,
+    REPLICA_PRECISE_PURGING, REPLICA_PURGE_DELAY, REPLICA_PURGE_INTERVAL)
+from lib389.topologies import topology_m1 as topo
+from lib389.dseldif import DSEldif
+from lib389.tombstone import Tombstones
+from lib389.idm.user import UserAccounts, TEST_USER_PROPERTIES
+
+
```
droideck, line 21: I think we need to add `pytestmark = pytest.mark.tier1` here.

And could you also please add Copyright to the file?

droideck (approved):
Besides the small issues and a minor typo in the commit message (swqitch) - Looks Good!

Issue #5259, opened 2022-04-11
dsctl --remove-all can be triggered with partial arg name


mreynolds389:
**Issue Description**

Apparently you don't need to enter the full arg parse option (--remove-all) for it to be run:

```
# dsctl --r

About to remove all Directory Server instances!
If this is not what you want, press ctrl-c now ...
        
5 ...
4 ...
3 ...
2 ...
1 ...
Removing instance: slapd-supplier1
All instances have been successfully removed
```

If an admin accidentally entered "--r", or '--re', etc, it will trigger "--remove-all" and blow away the entire installation unexpectedly.   We should fix that sooner than later.

Issue #5232, opened 2022-03-23
Remove --durable-txn from dsconf


mmuehlfeldRH:
**Issue Description**
nsslapd-db-durable-transaction can only be changed when the instance is stopped. The `dsconf backend config set --durable-txn` command makes no sense. Please remove `--durable-txn`.

**Package Version and Platform:**
 - Platform: RHEL 9
 - Package and version: 389-ds-base-2.0.14-3.module+el9dsrv+14387+a6c313de.x86_64

**Steps to Reproduce**
```
dsconf -D "cn=Directory Manager" ldap://server.example.com backend config set --durable-txn off
Error: 103 - 4 - 53 - Server is unwilling to perform - [] - nsslapd-db-durable-transaction can't be modified while the server is running.
```

**Expected results**
The `--durable-txn` should be removed because dsconf requires a running instance, and the setting can't be changed when the server is running.

tbordaz:
I have not look in details why it was unwilling to perform. IMHO, this setting could be done even if the server is running but should only be taken into account at next restart.

Note: this setting --durable-txn=off seem so risky to me, that I see no reason to propose it in the UI/CLI. For risky tuning like this, we may assume that it is done under support control and then experts agree to update dse.ldif.

Firstyear:
Yep, I agree with @tbordaz here :)

mreynolds389:
d50340ba5..f7f0986de  389-ds-base-2.1 -> 389-ds-base-2.1
4d2d010b8..0ee456ed3  389-ds-base-2.0 -> 389-ds-base-2.0

Issue #5195, opened 2022-03-08
ldifgen: make required parameters more specific 


vashirov:
**Issue Description**
When ldifgen is used without required parameters, it drops into interactive mode without telling which required parameters are missing:
```
# dsctl localhost ldifgen mod-load
Missing required parameters, switching to Interactive mode ...
```
It's possible to specify that some of the parameters are required by using [`required=True`](https://docs.python.org/3/library/argparse.html#required) keyword.

The downside of this approach is that it doesn't drop into interactive mode anymore.
Another option is to update `log.info` message to state what parameters are required.

mreynolds389:
d50340ba5..f7f0986de  389-ds-base-2.1 -> 389-ds-base-2.1
4d2d010b8..0ee456ed3  389-ds-base-2.0 -> 389-ds-base-2.0

Issue #5169, opened 2022-02-21
Remove debug logging line from standard error logging


mreynolds389:
**Issue Description**

Affects 2.1.x:

    [21/Feb/2022:16:08:52.623283312 -0500] - INFO - bdb__import_file_name - DBG: fname=/var/lib/dirsrv/slapd-standalone1/db/.import_userRoot

This should have a different log level and not be written by default

PR #5263, opened 2022-04-12
Issue 4904 - Fix various small issues

mreynolds389:
- Issue 4904 - Set the plugin identity before calling internal op.
- Issue 5232 - Remove --durable-txn from dsconf
- Issue 5195 - Improve error messages to state with settings are missing
- Issue 5169 - During LDIF import we log a debug message by default when we should
not
- Issue 5259 - dsctl --remove-all can be triggered with abbreviated arg
 name

relates: https://github.com/389ds/389-ds-base/issues/4904
relates: https://github.com/389ds/389-ds-base/issues/5232
relates: https://github.com/389ds/389-ds-base/issues/5195
relates: https://github.com/389ds/389-ds-base/issues/5169
relates: https://github.com/389ds/389-ds-base/issues/5259

### DIFF
--- a/src/lib389/cli/dsctl
+++ b/src/lib389/cli/dsctl
@@ -33,7 +33,7 @@ from lib389.cli_base import (
     format_error_to_dict)
 from lib389._constants import DSRC_CONTAINER
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(allow_abbrev=False)
 parser.add_argument('-v', '--verbose',
         help="Display verbose operation tracing during command execution",
         action='store_true', default=False
--- a/ldap/servers/slapd/add.c
+++ b/ldap/servers/slapd/add.c
@@ -386,7 +386,7 @@ slapi_exists_or_add_internal(
         0,
         NULL,
         NULL,
-        NULL,
+        plugin_get_default_component_id(),
         0);
 
     slapi_search_internal_pb(search_pb);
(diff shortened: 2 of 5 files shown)


Firstyear (approved):

droideck (approved):
LGTM!

tbordaz (approved):
LGTM

src/lib389/cli/dsctl
```diff
@@ -33,7 +33,7 @@ from lib389.cli_base import (
     format_error_to_dict)
 from lib389._constants import DSRC_CONTAINER
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(allow_abbrev=False)
```
progier389, line 36: I am afraid that this change will have unwanted side effect  (unable to pack the short options together) (for example  dsconf -vvvvv ... will have to be chamged to dsconf -v -v -v -v -v ... ) that may impact customer scripts ... 
IMHO we have better double check the options by ourself by checking that '--removal-all' is in sys.argv

progier389 (changes requested):

src/lib389/cli/dsctl
mreynolds389: I didn't think we used "-vvvvv", or any other arg option in that way.  I thought it was only -v.  I see no difference in the output:

[root@fedora cli]# dsconf -vvvvvv localhost config get | wc -l
362
[root@fedora cli]# dsconf -v localhost config get | wc -l
362

I don't mind changing the patch, but I don't "think" this is a concern.  Am I missing something?

progier389: Sorry for the noise, 
    finally there is no concern at all:   dsctl -jl is still working.
I just misread the Argparse documentation:
  In fact the short option issue was fixed in python 3.8 (I initially read (too quickly) the opposite):
Changed in version 3.8: In previous versions, allow_abbrev also disabled grouping of short flags such as -vv to mean -v -v.

progier389 (approved):
LGTM

src/lib389/cli/dsctl
mreynolds389: All good, I appreciate the review, and apparently it really was a problem not that long ago.  Glad it's fixed now :-)

Issue #5262, opened 2022-04-12
high contention in find_entry_internal_dn on mixed load


tbordaz:
**Issue Description**
Under mixed load (updates and indexed searches), the pressure on the entry cache is so high that during an update by the time the entry is added into the cache,  it is then removed from the cache before it is locked for the update.
If such event occurs 1000 times in a raw, the update fails.

**Package Version and Platform:**
all versions

**Steps to Reproduce**
Was not able to reproduce

**Expected results**
Should not see messages like "Retry count exceeded"

**Screenshots**

**Additional context**

tbordaz:
Reverting #4324 (https://github.com/389ds/389-ds-base/commit/abb93243b427ee122a5a80ed095c0118d7b6616e) prevent that lookup-lock fails 1000 times in a raw (msg 'Retry count exceeded').

#4324 improves performance for search (~8%)
It was not tested for write. This should be a first test to confirm there is no regression on write.

A second phase would be to implement an atomic lookup-lock

tbordaz:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=2051476

Firstyear:
The problem is that as we go to LMDB we will be moving to a concurrently readable transactional model. Having a "lock around the entry cache" basicly nerfs any performance gain from LMDB, and if we moved to a parallel hashmap, then we break our transactional guarantees. 

There are already transactional, concurrently readable caches that exist for exactly this scenario. I specifically wrote and designed one for the future use of 389-ds ( https://github.com/kanidm/concread/blob/master/CACHE.md , https://docs.rs/concread/latest/concread/arcache/index.html ) which you probably could consider.

droideck:
Associated PR and an issue:
https://github.com/389ds/389-ds-base/pull/5463
https://github.com/389ds/389-ds-base/issues/4324

tbordaz:
1011ad1fd..810fc8775 main
21314ec9e..3cc5e3829  389-ds-base-2.2
6c95604f6..00f5cd25d  389-ds-base-2.1
7ba233bf7..8d7d19faf  389-ds-base-2.0
8564399c5..1312c2796  389-ds-base-1.4.4
6d5537900..2a6870bcc  389-ds-base-1.4.3

PR #5264, opened 2022-04-13
Issue 5262 - high contention in find_entry_internal_dn on mixed load

tbordaz:
Bug description:
	Under high mixed load (update/read), the pressure on the entry cache is
        so high that during an update by the time the entry is added into the
	cache, it is then removed from the cache before it is locked for the update.
	If such event occurs 1000 times in a raw, the update fails.

	Another issue is that when updating the parent entry of a
	deleted entry (numsubordinates), if it fails to lock the parent
        it does not return the parent entry. So refcnt becomes invalid.

Fix description:
        To reduce the likelyhood of failure, if an entry is removed from
        the entry cache before it is locked, then pause 100ms before
        retrying to upload/lock it.
        In the case of the failure to lock the parent entry, the entry
        should be returned.

relates: #5262

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_delete.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_delete.c
@@ -542,6 +542,7 @@ replace_entry:
                             if (cache_retry == RETRY_CACHE_LOCK &&
                                 cache_retry_count < LDBM_CACHE_RETRY_COUNT) {
                                 /* try again */
+                                CACHE_RETURN(&(inst->inst_cache), &parent);
                                 DS_Sleep(PR_MillisecondsToInterval(100));
                                 cache_retry_count++;
                                 continue;


progier389:
A few questions: ( i wonder about them because 100ms seems a awfully long time compared to average operation time. )
    did you see performance degradation related to this fix with mixed load ? 
    have you tried other values for the time out ?

mreynolds389:
> A few questions: ( i wonder about them because 100ms seems a awfully long time compared to average operation time. ) did you see performance degradation related to this fix with mixed load ? have you tried other values for the time out ?

Yeah adding "sleeps" to server code makes me uneasy, and I agree with Pierre that it seems like a long sleep.  I wonder how 10ms would do?

tbordaz:
@progier389, I missed one of your remark. I did not do an performance measure of the fix. I was unable to reproduce such high pressure on entrycache, but I can retry.

Firstyear:
A better idea could be to add sched_yield to allow thread backoff and rescheduling, which is likely to help since we over-provision threads by default.

I'd be curious to see if this was "better" if we set threads to match cores instead of our current over provision strategy.

droideck (approved):
LGTM!
At least, for the `CACHE_RETURN(&(inst->inst_cache), &parent);` part as discussed previously.

droideck:
@tbordaz could you please merge this PR as it's verified that helped he customer? (at lease the part with `CACHE_RETURN(&(inst->inst_cache), &parent);` should go in, for sure, I think)

Issue #5271, opened 2022-04-19
Serialization of pam_passthrough causing high etimes


tbordaz:
**Issue Description**
libpam was not thread-safe and 389-ds [serialized](https://github.com/389ds/389-ds-base/commit/87e9a9fc6d8a98e421f2363f97b33afd75f4158e) the calls to pam authentication.
Since a long time now libpam is thread-safe, at the condition each worker uses its own pam_handle. 389-ds workers are using their own pam_handle (obtained from pam_start) so there is no reason to keep the serialization.

This serialization create contention and high etime when pam_authenticate response time is higher than incoming rate of pam_passthrough requests.

**Package Version and Platform:**
all version (since 2008)

**Steps to Reproduce**
Apply parallel pam_passthrough calls

**Expected results**
should not be serialized

**Additional context**

mreynolds389:
49b0ae530..17cc58066  389-ds-base-2.2 -> 389-ds-base-2.2
94d124b0a..7e1d99182  389-ds-base-2.1 -> 389-ds-base-2.1
e05369cbc..4a7dbb44e  389-ds-base-2.0 -> 389-ds-base-2.0
d0a91891c..08c18b373  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
bd566957f..d08a0e414  389-ds-base-1.3.10 -> 389-ds-base-1.3.10

droideck:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2170223

PR #5272, opened 2022-04-19
Issue 5271 - Serialization of pam_passthrough causing high etimes

tbordaz:
Bug description:
	calls to PAM authentication are serialized by DS.
	libpam is thread-safe and some module may also be
        thread-safe.
        Because serialization impacts performance, an administator
        should be allowed to remove the serialization in case
        the pam module is thread safe

Fix description:
	If the pam configuration entry contains 'pamThreadSafe: TRUE'
        then futher calls to pam_start/pam_authenticate/pam_end are not
        serialized

relates: #5271

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/pam_passthru/pam_passthru.h
+++ b/ldap/servers/plugins/pam_passthru/pam_passthru.h
@@ -79,6 +79,8 @@ typedef struct pam_passthruconfig
     PRBool pamptconfig_fallback;              /* if false, failure here fails entire bind */
                                               /* if true, failure here falls through to regular bind */
     PRBool pamptconfig_secure;                /* if true, plugin only operates on secure connections */
+    PRBool pamptconfig_thread_safe;           /* if true, the underlying pam module is thread safe */
+                                              /* if false, the module is not thread safe => serialize calls */
     char *pamptconfig_pam_ident_attr;         /* name of attribute in user entry for ENTRY map method */
     int pamptconfig_map_method1;              /* how to map the BIND DN to the PAM identity */
     int pamptconfig_map_method2;              /* how to map the BIND DN to the PAM identity */
@@ -101,6 +103,7 @@ typedef struct pam_passthruconfig
 #define PAMPT_MAP_METHOD_ATTR "pamIDMapMethod"       /* single valued */
 #define PAMPT_FALLBACK_ATTR "pamFallback"            /* single */
 #define PAMPT_SECURE_ATTR "pamSecure"                /* single */
+#define PAMPT_THREAD_SAFE_ATTR "pamModuleIsThreadSafe" /* single */
 #define PAMPT_SERVICE_ATTR "pamService"              /* single */
 #define PAMPT_FILTER_ATTR "pamFilter"                /* single */
 
--- a/ldap/servers/plugins/pam_passthru/pam_ptconfig.c
+++ b/ldap/servers/plugins/pam_passthru/pam_ptconfig.c
@@ -579,6 +579,7 @@ pam_passthru_apply_config(Slapi_Entry *e)
     char *dn = NULL;
     PRBool fallback;
     PRBool secure;
+    PRBool thread_safe;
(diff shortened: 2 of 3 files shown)


ldap/servers/plugins/pam_passthru/pam_passthru.h
```diff
@@ -79,6 +79,8 @@ typedef struct pam_passthruconfig
     PRBool pamptconfig_fallback;              /* if false, failure here fails entire bind */
                                               /* if true, failure here falls through to regular bind */
     PRBool pamptconfig_secure;                /* if true, plugin only operates on secure connections */
+    PRBool pamptconfig_thread_safe;           /* if true, the underlying pam module is thread safe */
+                                              /* if flase, the module is not thread safe => serialize calls */
```
mreynolds389: typo, "flase"

Firstyear:
Are there any examples of thread safe pam modules? 

Additionally, with SSSD even if pam is thread safe, SSSD is single threaded and this will bottle neck. So I'm a bit skeptical on this because it carries a lot of risks of subtle breakage for admins, even if a smoke test we do "looks okay".

tbordaz:
> Are there any examples of thread safe pam modules?
> 
> Additionally, with SSSD even if pam is thread safe, SSSD is single threaded and this will bottle neck. So I'm a bit skeptical on this because it carries a lot of risks of subtle breakage for admins, even if a smoke test we do "looks okay".

You are absolutely right. The more I am investigating this the more risky it looks to me. DS pam plugin locking makes safer the use of unknown underlying mechanism.
ATM it exists pam_radius (freeradius) that is thread safe, so at least for this mechanism it will help scaling.
For others interesting pam mechanism it still requires investigations.

Firstyear:
If we are going to go ahead with this, maybe we can make the tunable named something liek "danger_pam_pt_module_assume_thread_safe"

tbordaz:
@Firstyear , upstream pam_radius and pam_sss can get benefit of relaxing this serialization. 
"pamModuleIsThreadSafe" is it enough clear or does it require to be more alarming ?

tbordaz:
Internal heavy tests shows that pam_radius 2.0 supports multi-thread auth. pam_sss is in the process to become thread safe (new dev), the same heavy tests did not break it.
In both tests auth average response time improved (more than 2 times faster).

I fully agree that not serializing a non thread-safe module is a killer. In such case DS likely crash and the victim thread may not be the guilty one. However relaxing the serialization is not the default behavior, it is the administrator decision.

Is your concern related to support difficulty, naming of the attribute,... ?

There are others tunable (like txn batch) that are also risky one.

Firstyear:
> Internal heavy tests shows that pam_radius 2.0 supports multi-thread auth. pam_sss is in the process to become thread safe (new dev), the same heavy tests did not break it. In both tests auth average response time improved (more than 2 times faster).
> 
> I fully agree that not serializing a non thread-safe module is a killer. In such case DS likely crash and the victim thread may not be the guilty one. However relaxing the serialization is not the default behavior, it is the administrator decision.
> 
> Is your concern related to support difficulty, naming of the attribute,... ?
> 
> There are others tunable (like txn batch) that are also risky one.

This assumes that all the modules in the stack are thread safe. 

Still, I think the naming of the attribute needs to be something like "pamDangerModuleIsThreadSafe"

tbordaz:
Hi @Firstyear, if you have some bandwidth to look at my last update I would appreciate your feedback . Thanks

tbordaz:
@Firstyear , sorry to bother you again and again but you have been the only reviewer of that fix. Because of pam_sss and pam_radius being thread safe, a fix in pam_passthrough is required to get benefit of those improved pam module.
My last update was https://github.com/389ds/389-ds-base/pull/5272#issuecomment-1181928701. Does it answer your concerns or did I miss something ?

Firstyear (approved):
Okay, given @tbordaz last comments I'm okay with this.

Issue #5273, opened 2022-04-22
CLI - add arg completer for instance name


mreynolds389:
**Issue Description**

Add arg completer for the instance names to the CLI tools:

```
# dsconf slapd-l  <hit tab>
# dsconf slapd-localhost 
```

mreynolds389:
beb382e1b..865b3aacc  389-ds-base-2.1 -> 389-ds-base-2.1
6dc6430b1..498ec3e93  389-ds-base-2.0 -> 389-ds-base-2.0
a2eebf990..fc58e2f30  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
1b31519c5..cf44ab82f  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5274, opened 2022-04-22
Issue 5273 - CLI - add arg completer for instance name

mreynolds389:
Description:

For convenience it would be nice to have auto compeletion for the instance
name when using the CLI tools.

relates: https://github.com/389ds/389-ds-base/issues/5273

Reviewed by: ?


droideck (approved):
Nice addition! Ack!

PR #5275, opened 2022-04-26
Issue 3 - ansible-ds - Prefix handling fix

droideck:
Desciption: When we run 389-ds in Ansible, its processing sets
PREFIX to "", so we should check for both - None and "".

Related: https://github.com/389ds/ansible-ds/issues/3

Reviewed by: ?


progier389 (approved):
LGTM

PR #5282, opened 2022-05-05
Issue 5281 - HIGH - basic test does not run

### LINKED ISSUE #5281: HIGH - basic test does not run
In a development environment, where you did a chroot install, basic test no longer functions. This appears to be due to:

pytestmark = pytest.mark.skipif(get_rpm_version("389-ds-base-snmp") == "not installed", reason="389-ds-base-snmp package is not present")

This also makes a lot of assumptions about the distro, which excludes anything that isn't fedora/rhel - we don't have a 389-ds-base-snmp package, in opensuse and suse we have 389-ds-snmp.

In a chroot everything for snmp is present:

```
find /opt/dirsrv | grep -i ldap-agent
/opt/dirsrv/sbin/ldap-agent
/opt/dirsrv/etc/dirsrv/config/ldap-agent.conf
/opt/dirsrv/share/man/man1/ldap-agent.1
```

So I think this needs to be reverted/ fixed.

vashirov:
Bug Description:
test_basic_ldapagent has an incorrect pytestmark that checks for presence
of 389-ds-base-snmp package that might not exist on other distros than
Fedora/RHEL. It's also a second declaration of the pytestmark that
overrides previous one at the beginning of the test.

Fix Description:
Instead of checking for package presence, check for `ldap-agent` binary.

Fixes: https://github.com/389ds/389-ds-base/issues/5281

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -985,7 +985,6 @@ def test_basic_systemctl(topology_st, import_example_ldif):
     log.info('test_basic_systemctl: PASSED')
 
 
-pytestmark = pytest.mark.skipif(get_rpm_version("389-ds-base-snmp") == "not installed", reason="389-ds-base-snmp package is not present")
 def test_basic_ldapagent(topology_st, import_example_ldif):
     """Tests that the ldap agent starts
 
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
ack

droideck (approved):
LGTM too!

PR #5283, opened 2022-05-06
Issue 379 - RFE - Compress rotated logs

mreynolds389:
Description:

After rotating a log compress it using gzip.  Required updating the
rotation/deletion policy to account for the longer log name (.gz)

relates: https://github.com/389ds/389-ds-base/issues/379

This requires an update to selinux to work on RHEL.  See:  https://bugzilla.redhat.com/show_bug.cgi?id=2081797

### DIFF
--- a/dirsrvtests/conftest.py
+++ b/dirsrvtests/conftest.py
@@ -5,6 +5,7 @@ import shutil
 import glob
 import ldap
 import os
+import gzip
 
 from .report import getReport
 from lib389.paths import Paths
@@ -115,7 +116,13 @@ def pytest_runtest_makereport(item, call):
                     text = asan_report.read()
                     extra.append(pytest_html.extras.text(text, name=os.path.basename(f)))
             for f in glob.glob(f'{p.log_dir.split("/slapd",1)[0]}/*/*'):
-                if 'rotationinfo' not in f:
+                if f.endswith('gz'):
+                    with gzip.open(f, 'rb') as dirsrv_log:
+                        text = dirsrv_log.read()
+                        log_name = os.path.basename(f)
+                        instance_name = os.path.basename(os.path.dirname(f)).split("slapd-",1)[1]
+                        extra.append(pytest_html.extras.text(text, name=f"{instance_name}-{log_name}"))
+                elif 'rotationinfo' not in f:
                     with open(f) as dirsrv_log:
                         text = dirsrv_log.read()
                         log_name = os.path.basename(f)
--- a/dirsrvtests/tests/suites/logging/logging_compression_test.py
+++ b/dirsrvtests/tests/suites/logging/logging_compression_test.py
@@ -1,0 +1,121 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import glob
+import ldap
+import logging
+import pytest
+import os
+import time
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.dseldif import DSEldif
+from lib389.topologies import topology_st as topo
+from lib389.idm.domain import Domain
+
+log = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.tier1
+
+def log_rotated_count(log_type, log_dir, check_compressed=False):
+    # Check if the log was rotated
+    log_file = f'{log_dir}/{log_type}.2*'
+    if check_compressed:
+        log_file += ".gz"
+    return len(glob.glob(log_file))
+
+
+def update_and_sleep(suffix, sleep=True):
+    for loop in range(2):
+        for count in range(10):
+            suffix.replace('description', str(count))
+            suffix.get_attr_val('description')
+            with pytest.raises(ldap.OBJECT_CLASS_VIOLATION):
+                suffix.add('doesNotExist', 'error')
+        if sleep:
+            # log rotation smallest unit is 1 minute
+            time.sleep(61)
+        else:
+            # should still sleep for a little bit
+            time.sleep(1)
+
+
+def test_logging_compression(topo):
+    """Test logging compression works, and log rotation/deletion is still
+    functional.  This also tests a mix of non-compressed/compressed logs.
+
+    :id: 15b5ed0e-628c-48e5-a61e-43908590c9f1
+    :setup: Standalone Instance
+    :steps:
+        1. Enable all the logs (audit,and auditfail)
+        2. Set an aggressive rotation/deletion policy
+        3. Make sure all logs are rotated at least once
+        4. Enable log compression on all logs
+        5. Make sure all logs are rotated again and are compressed
+        6. Make sure log deletion is working
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Success
+    """
+
+    inst = topo.standalone
+    suffix = Domain(topo.standalone, DEFAULT_SUFFIX)
+    timeunit = "minute"  # This is the smallest time unit available
+    log_dir = inst.get_log_dir()
+
+    # Enable all the logs (audit, and auditfail)
+    inst.stop()
+    dse_ldif = DSEldif(inst)
+    dse_ldif.replace('cn=config', 'nsslapd-auditfaillog', log_dir + "/auditfail")
+    inst.start()
+
+    inst.config.set('nsslapd-auditlog-logging-enabled', 'on')
+    inst.config.set('nsslapd-auditfaillog-logging-enabled', 'on')
+    inst.config.set('nsslapd-accesslog-logbuffering', 'off')
+    inst.config.set('nsslapd-errorlog-level', '64')
+
+    # Set an aggressive rotation/deletion policy for all logs
+    for ds_log in ['accesslog', 'auditlog', 'auditfaillog', 'errorlog']:
+        inst.config.set('nsslapd-' + ds_log + '-logrotationtime', '1')
+        inst.config.set('nsslapd-' + ds_log + '-logrotationtimeunit', timeunit)
+        inst.config.set('nsslapd-' + ds_log + '-maxlogsize', '1')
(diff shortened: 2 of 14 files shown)


dirsrvtests/conftest.py
```diff
@@ -115,7 +115,7 @@ def pytest_runtest_makereport(item, call):
                     text = asan_report.read()
                     extra.append(pytest_html.extras.text(text, name=os.path.basename(f)))
             for f in glob.glob(f'{p.log_dir.split("/slapd",1)[0]}/*/*'):
-                if 'rotationinfo' not in f:
+                if 'rotationinfo' not in f and not f.endswith('.gz'):
```
progier389: IMHO we should handle compressed log too here rather than just ignoring it:
```
-                if 'rotationinfo' not in f:
+                if f.endswith('gz'):
+                    with subprocess.Popen(["/usr/bin/zcat", f], stdout=PIPE) as dirsrv_log:
+                        text = dirsrv_log.read()
+                        log_name = os.path.basename(f)
+                        instance_name = os.path.basename(os.path.dirname(f)).split("slapd-",1)[1]
+                        extra.append(pytest_html.extras.text(text, name=f"{instance_name}-{log_name}"))
+                elif 'rotationinfo' not in f:
```

ldap/servers/slapd/log.c
```diff
@@ -156,6 +156,22 @@ get_syslog_loglevel(int loglevel)
...
+    char cmd[BUFSIZ] = {0};
+
+    PR_snprintf(cmd, sizeof(cmd), "/usr/bin/gzip %s", filename);
+    if (PR_SUCCESS != PR_Access(filename, PR_ACCESS_EXISTS)) {
+        PR_snprintf(cmd, sizeof(cmd), "Can not compress log file because it does not exist (%s)\n", filename);
+        log__error_emergency(cmd, 0, 0);
+        return;
+    }
+    /* gzip it */
+    system(cmd);
```
progier389: I am a bit wary about using system with ns-slapd. I wonder if we should not use zlib instead ?

mreynolds389: zlib!!  Didn't know about zlib.  :-)  I also did not want to use system(), but I didn't know there was an alternative.  So I will definitely look into this.  This will also remove the selinux issue I ran into.

ldap/servers/slapd/log.c
```diff
@@ -3363,31 +3488,42 @@ log__fix_rotationinfof(char *pathname)
...
+                   NULL != strchr(p, '-')) /* e.g., errors.20051123-165135 */
+        {
             struct logfileinfo *logp;
             char *q;
             int ignoreit = 0;
 
-            for (q = p; q && *q; q++) {
-                if (*q != '-' && !isdigit(*q))
+            for (q = ++p; q && *q; q++) {
+                if (*q != '-' &&
```
progier389, line 3535: Regression risk: some files that are currently rejected like error.zz2022...  will be seen as valid
 if would break the loop if strcmp(q,".gz")==0)  and I will double check at the end of the loop that q-p is 15 :
    if (ignoreit or q-p != 15)

ldap/servers/slapd/log.c
```diff
@@ -3549,21 +3687,26 @@ log__check_prevlogs(FILE *fp, char *pathname)
...
         if (0 == strncmp(log_type, dirent->name, strlen(log_type)) &&
             (p = strrchr(dirent->name, '.')) != NULL &&
-            15 == strlen(++p) &&
             NULL != strchr(p, '-')) { /* e.g., errors.20051123-165135 */
             char *q;
             int ignoreit = 0;
 
-            for (q = p; q && *q; q++) {
-                if (*q != '-' && !isdigit(*q))
+            for (q = ++p; q && *q; q++) {
```
progier389, line 3723: same remark than previous one ...

ldap/servers/slapd/log.c
```diff
@@ -3660,6 +3803,9 @@ log__extract_logheader(FILE *fp, long *f_ctime, PRInt64 *f_size)
         if (PR_SUCCESS != PR_Access(p, PR_ACCESS_EXISTS)) {
             return LOG_ERROR;
         }
+        if (strcmp(p + strlen(p) - 3, ".gz") == 0) {
```
progier389, line 3835: Should check that p+strlen(p)-3 >= buf before trying to compare

progier389 (changes requested):
IMHO A few minor changes are needed, and maybe a more important one (using zlib instead of system?)

ldap/servers/slapd/log.c
mreynolds389: Well there was "++p" called in the old code before it got to this for loop.  So the behavior should be the same as before.  So perhaps there has always been this regression?

ldap/servers/slapd/log.c
mreynolds389: I'll add that check, but all these strings should have at least a log name in them (access, audit, auditfail, errors)

ldap/servers/slapd/log.c
progier389: I do not think that we speak of the same thing here:
  My point is that you now accept z and g as valid character in the middle of error file name rather than checking that it is at the end -;) 
  The ++ is ok as you moved it from the now removed strlen test to here, I will just add the q-p test to check that the date lenght is still ok ..

ldap/servers/slapd/log.c
progier389: Good point. and there is also the LOGINFO prefix so the test is useless as there is no risk of getting out of the buffer here

ldap/servers/slapd/log.c
mreynolds389: @progier389 - all changes made, please review...

dirsrvtests/conftest.py
mreynolds389: Sorry didn't get to this one yet.  Looking at it now...

mreynolds389: I fixed this by using python's gzip command instead (but it's basically the same thing) :-)

dirsrvtests/tests/suites/logging/logging_compression_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+def update_and_sleep(suffix, sleep=True):
+    for loop in range(2):
+        for count in range(10):
+            suffix.replace('description', str(count))
+            suffix.get_attr_val('description')
+            with pytest.raises(ldap.OBJECT_CLASS_VIOLATION):
+                suffix.add('doesNotExist', 'error')
+        if sleep:
+            time.sleep(61)
```
droideck, line 41: Could you please put a comment here on why we sleep 61 seconds?
While reviewing PR I haven't noticed the logrotationtime is set to 1 minute and I got confused for a bit. It's a long sleep for tests. :)

dirsrvtests/tests/suites/logging/logging_compression_test.py
```diff
@@ -0,0 +1,119 @@
...
+import pytest
+import os
+import time
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.dseldif import DSEldif
+from lib389.topologies import topology_st as topo
+from lib389.idm.domain import Domain
+
+log = logging.getLogger(__name__)
+
```
droideck, line 21: Could you please add `pytestmark = pytest.mark.tier1` here so the tests are run in the GitHub CI?

src/cockpit/389-console/src/lib/server/accessLog.jsx
```diff
@@ -685,6 +699,21 @@ export class ServerAccessLog extends React.Component {
                                     />
                                 </GridItem>
                             </Grid>
+                            <Grid title="Compress (gzip) the log after it's rotated.">
+                                <GridItem className="ds-label" span={3}>
+                                    Compress Rotated Logs
+                                </GridItem>
+                                <GridItem span={1}>
+                                    <Checkbox
```
droideck: It's a subjective thing but I think 'Switch' will look better here.
Now it looks very small and "stealthy" compared to other elements on the page.

image

The `Switch` will make it more visible and nice for perception.
Still, it's just my opinion and I leave the final decision fully to you.

droideck (approved):
Looks great!
Besides the minor comments and one suggestion - you have my ack.

dirsrvtests/tests/suites/logging/logging_compression_test.py
mreynolds389: > Could you please put a comment here on why we sleep 61 seconds? While reviewing PR I haven't noticed the logrotationtime is set to 1 minute and I got confused for a bit.

Fixed

dirsrvtests/tests/suites/logging/logging_compression_test.py
mreynolds389: > Could you please add `pytestmark = pytest.mark.tier1` here do the tests are run in the GitHub CI?

Done

src/cockpit/389-console/src/lib/server/accessLog.jsx
mreynolds389: I switched to the Switch.  I had thought about it, but never got around to trying it out.  It is definitely better!  Thanks!

All changes are made, please review...

progier389 (approved):
LGTM

tbordaz:
@mreynolds389 , sorry for this late question. My understanding is that the compression (access, audit, errors) occurs on the fly when we need to rotate the log file. Log of the operation occurs after sending the result so, compression should only delay the worker to be back to the work queue. That looks fine.
Now if a debug log is enabled, my understanding is that compression may randomly slowdown an operation. Am I correct ?
Does it worth to be documented ?

mreynolds389:
> @mreynolds389 , sorry for this late question. My understanding is that the compression (access, audit, errors) occurs on the fly when we need to rotate the log file. Log of the operation occurs after sending the result so, compression should only delay the worker to be back to the work queue. That looks fine. Now if a debug log is enabled, my understanding is that compression may randomly slowdown an operation. Am I correct ? Does it worth to be documented ?

It would not hurt to document it.  However, if you are using debug logging then operations are already going to be slowed down :-)  So I'm not sure if documenting it is necessary, but I am not opposed to mentioning it in the release notes.

PR #5286, opened 2022-05-09
Issue 5284 - Replication broken after password change

### LINKED ISSUE #5284: Replication broken until server restart after password change
**Issue Description**
BZ1956987: After changing the replication agreement and the replication manager password, 
replication fails until the  server is restarted (connection fails with Invalid Credential error).

 - Platform: Fedora
 - Package and version:  Master branch

**Steps to Reproduce**
 - On supplier, Change password of replication agreement
 - On consumer: Change password of the replication manager entry
 - On supplier modify an entry within replicated suffix
 - Wait a few seconds
 - On consumer: Check that the entry modification has been replicated.
If the bug exists. 

**Expected results**
- Replication should be still working (after maybe a transient error on replication session started while password where changed).

**Additional context**
The issue is that a clear version of the agmt password is cached within the replication connection struct and not updated 
 when the password is changed.
A solution is to also store a copy of the encrypted password in that struct so it can be quickly compared to the agmt password, then both passwords values get recomputed if the compare failed.

progier389:
Problem: A cached version of decoded password within agmt connection
 is not updated when password get changed.

Solution: Store also the encoded version beside the decoded one
 and replace both password verfsion if the encoded one does not
 match the agmt password.

Also fixed two issue related to ensure_state within lib389 mapped entries
 to support attribute deletion (using None value) and avoid  replacing
 values that are already present (which can generate error in cn=config)

Issue: 5284

Reviewed by ?

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -18,11 +18,13 @@ from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts
 from lib389.pwpolicy import PwPolicyManager
 from lib389.utils import *
 from lib389._constants import *
+from lib389.idm.domain import Domain
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccount
 from lib389.idm.group import Groups, Group
 from lib389.idm.domain import Domain
 from lib389.idm.directorymanager import DirectoryManager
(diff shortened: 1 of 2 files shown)


progier389:
Fixed test order (because of a fixture conflict . Also reversed my change in lib389 that created a regression (and now using replace_many/remove_all to revert entries to original state rather than broken ensure_state)

Firstyear (approved):
Seems straight forward enough to me :)

PR #5292, opened 2022-05-10
Issue 5291 - Harden ReplicationManager.wait_for_replication

### LINKED ISSUE #5291: CI Test: Harden ReplicationManager.wait_for_replication lib389 function
**Issue Description**
To verify that replication is working , lib389 ReplicationManager.wait_for_replication 
 modify the replication manager group description with a random value on the supplier
then loops until that value is seen on the consumer side.
Unfortunately that is not a good way to check replication (Because of URP there is no way to be absolutly sure that the change we have done will be seen as the ,last change (an change done the same second on another supplier may take precedence.
To harden this code and offer a better diagnostic, I propose: to also check the description value on supplier side
   If the changed, supplier descriptions and consumer description values are the same, Then replication is OK.
   if supplier and consumer values are different then replica are not in sync ==> wait and loop 
   if change and supplier values are different it is probable that change get overwritten ==>
      Redo the change (As we waited 1 second, it should not be overwritten again)
      wait and loop 

 - Platform: Aby
 - Package and version: Master branch

**Steps to Reproduce**
We sometime see error in CI test result (and so far we cannot determine if it is really an error or not)

**Expected results**
Having different debug/info message for the three case so that we  can easilly determine if replication is really broken or not.

progier389:
ReplicationManager.wait_for_replication may wrongly fails without no way to determine if the error is normal or not.
Solution to harden the function is to also check the description value ob the "from" side
And if servers are in sync but not having the expected value

LOG A DIFFERENT MESSAGE
REDO THE CHANGE (because previous one may have been overwritten by URP)

If the timeout (now set to 60 seconds by default) expires then log last 30 lines from replication plugin in error log
 on both instances.  (to capture the last replication session)
This change helped me to understand why test_change_repl_passwd[using-bind-group] fails while it pass when running alone.
 So I also fixed test_fetch_bindDnGroup test to restore the configuration and avoid impacting the other tests.

Issue 5291

Reviewed by: @mreynolds389

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -589,14 +589,19 @@ def test_fetch_bindDnGroup(topo_m2):
     repl = ReplicationManager(DEFAULT_SUFFIX)
     repl.wait_for_replication(M1, M2, timeout=20)
 
+    save_values = [] # Save original value to restore them at the end of the test
     # Define the group as the replication manager and fetch interval as 60sec
     replicas = Replicas(M1)
     replica = replicas.list()[0]
+    save_values.append((replica, 'nsDS5ReplicaBindDnGroupCheckInterval', replica.get_attr_val_utf8('nsDS5ReplicaBindDnGroupCheckInterval')))
+    save_values.append((replica, 'nsDS5ReplicaBindDnGroup', replica.get_attr_val_utf8('nsDS5ReplicaBindDnGroup')))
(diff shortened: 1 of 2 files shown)


mreynolds389:
Can we bump the default timeout from 20 seconds to 60?

mreynolds389 (approved):
Besides my request about changing the default timeout the rest looks great!

progier389:
Changed the timeout to 60
Also limited the number of error log added in pytest log to the 30 last ones
And finally fixed test_fetch_bindDnGroup to restore the initial configuration after the test to avoid impacting the other tests

PR #5303, opened 2022-05-23
Issue 5302 - Release tarballs don't contain cockpit webapp

### LINKED ISSUE #5302: Release tarballs don't contain cockpit webapp
**Issue Description**
Ubuntu package since 389-ds-base-2.0.15 doesn't contain cockpit webapp: 
https://bugs.launchpad.net/ubuntu/+source/389-ds-base/+bug/1974329

The issue is that release tarballs on GitHub do not contain precompiled webapp. They are generated automatically from the git tags. Prior to that we distributed release tarballs on Pagure, where they were prepared with `dist-bz` target and contained all the required files.

Also, current release tarballs contain top-level directory with the duplicated name such as `389-ds-base-389-ds-base-2.0.15`, because we also use `389-ds-base` in our tag names.

We should generate proper release tarballs and provide source-only and source + vendored dependencies tarballs so that they can be easily consumed by the downstream distributions.

vashirov:
Description:
Add a new GitHub action to create a release tarball when a new tag
is created. The created source tarball also contains Rust vendored
dependencies and precompiled Cockpit assets.

Fixes: https://github.com/389ds/389-ds-base/issues/5302

Reviewed by: ???

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,31 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "389-ds-base-*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/389ds/ci-images:test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Create tarball
+        run: TAG=${{ github.ref_name}} make -f rpm.mk dist-bz2
+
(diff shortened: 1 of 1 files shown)


vashirov:
Example of the created release:
https://github.com/vashirov/389-ds-base/releases/tag/389-ds-base-2.0.16

mreynolds389 (approved):
LGTM

PR #5306, opened 2022-05-23
Issue 5305 - OpenLDAP version autodetection doesn't work

### LINKED ISSUE #5305: OpenLDAP version autodetection doesn't work
**Issue Description**
I've updated CI images to F36 and noticed that all compile jobs failed. 
https://github.com/389ds/389-ds-base/runs/6556076219?check_suite_focus=true
```
checking for --with-libldap-r... 
...
/usr/bin/ld: cannot find -lldap_r
make[1]: *** [Makefile:5795: libslapd.la] Error 1
make: *** [Makefile:4009: all] Error 2
Error: Process completed with exit code 2.
```

F36 has openldap-2.6.1, in these jobs rpm build is not used. Only ```autoreconf -fvi && ./configure``` is executed. 

Looking at the generated `configure`, `OPENLDAP_VERSION` doesn't have the correct regex, because square brackets are special characters and should be escaped. Moreover, version comparison logic is inverted, resulting in `--with-libldap-r=yes` on systems where `ldap_r` doesn't exist.

vashirov:
Fix Description:
* Escape regex for `OPENLDAP_VERSION`.
* Invert logic for the version comparison.

Relates: https://github.com/389ds/389-ds-base/issues/5032
Fixes: https://github.com/389ds/389-ds-base/issues/5305

Reviewed by: ???

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -881,9 +881,9 @@ AC_ARG_WITH(libldap-r, AS_HELP_STRING([--with-libldap-r],[Use lldap_r shared lib
     AC_SUBST(with_libldap_r)
   fi
 ],
-OPENLDAP_VERSION=`ldapsearch -VV 2> >(sed -n '/ldapsearch/ s/.*ldapsearch \([0-9]\+\.[0-9]\+\.[0-9]\+\) .*/\1/p')`
-AC_MSG_RESULT([$OPENLDAP_VERSION])
-AX_COMPARE_VERSION([$OPENLDAP_VERSION], [lt], [2.5], [ with_libldap_r=no ], [ with_libldap_r=yes ]))
+OPENLDAP_VERSION=`ldapsearch -VV 2> >(sed -n '/ldapsearch/ s/.*ldapsearch \([[[0-9]]]\+\.[[[0-9]]]\+\.[[[0-9]]]\+\) .*/\1/p')`
+AX_COMPARE_VERSION([$OPENLDAP_VERSION], [lt], [2.5], [ with_libldap_r=yes ], [ with_libldap_r=no ])
+AC_MSG_RESULT($with_libldap_r))
 
 AM_CONDITIONAL([WITH_LIBLDAP_R],[test "$with_libldap_r" = yes])
 


mreynolds389 (approved):
Ack

vashirov:
Build failed because `VERSION_PREREL` is not empty and set to 0. This causes version to be `389-ds-base-2.2.00` when build is done out of tree and `.git` directory is not present.
I'm not sure why it worked before, but I think we should unset it.

VERSION.sh
```diff
@@ -12,7 +12,7 @@ VERSION_MAJOR=2
 VERSION_MINOR=2
 VERSION_MAINT=0
 # NOTE: VERSION_PREREL is automatically set for builds made out of a git tree
-VERSION_PREREL=0
+VERSION_PREREL=
 VERSION_DATE=$(date -u +%Y%m%d%H%M)
```
mreynolds389: Does this mess with the version number when doing "make install" or "make -f rpm.mk rpms"?

vashirov: Yes, it breaks `make -f rpm.mk rpms` build when `.git` directory is not present. For example, from tarball:
```
cp /workspace/ds/rpmbuild/RPMS/*/*2.2.00*.rpm dist/rpms/
cp: cannot stat '/workspace/ds/rpmbuild/RPMS/*/*2.2.00*.rpm': No such file or directory
make: *** [rpm.mk:140: rpms] Error 1
```

`VERSION_PREREL` should be empty for the real releases. And it is automatically populated with git commit hash if it's a development build. It was empty until b4d22b67463586561f4621dce4b3b198f3858373.

vashirov: The reason we started to see failures now: for quite some time CI images were not updated automatically, they had an older version of Git. I've updated CI images to F36, but they broke Compile action (this issue fixes it). I've reverted back to F35, but the updated image now contains new version of Git that has a new behaviour after CVE-2022-24765 was fixed (see https://github.blog/2022-04-12-git-security-vulnerability-announced/ and https://github.blog/2022-04-18-highlights-from-git-2-36/). This caused `git` commands not to be executed after checkout, for example, we use them to set `VERSION_PREREL` to the commit hash.
I'll open a new issue for this and move `VERSION_PREREL` change there. And this PR should be merged after.

droideck (approved):
LGTM!

PR #5308, opened 2022-05-24
Issue 5307 - VERSION_PREREL is not set correctly in CI builds

### LINKED ISSUE #5307: `VERSION_PREREL` is not set correctly in CI builds
**Issue Description**
`VERSION_PREREL` is used to set pre-release version (`.a1`, `.rc1`, etc). If the build is done inside the git tree, it is set to `VERSION_PREREL=.${VERSION_DATE}git$COMMIT`. 
But since we don't do any Alpha or RC releases, it should be empty by default and populated by date and git commit hash in the development builds.
It was updated in https://github.com/389ds/389-ds-base/commit/b4d22b67463586561f4621dce4b3b198f3858373, but it stayed hidden until git-2.36 which fixed CVE-2022-24765 (see https://github.blog/2022-04-12-git-security-vulnerability-announced/ and https://github.blog/2022-04-18-highlights-from-git-2-36/). This caused git commands not to be executed after checkout in our CI (because UIDs were different on the host and in the container that we use for building RPMs). We should add $GITHUB_WORKSPACE as a safe directory.

vashirov:
Bug Description:
VERSION_PREREL is used to set pre-release version (.a1, .rc1, etc).
If the build is done inside the git tree, it is set to
VERSION_PREREL=.${VERSION_DATE}git$COMMIT.
But since we don't do any Alpha or RC releases, it should be
empty by default and populated by date and git commit hash in the
development builds.
Additionally, in our CI git commands stopped working after git-2.36
resulting in an incorrect VERSION_PREREL.

Fix Description:
* Set VERSION_PREREL to an empty value
* Update GH actions to explicitly add $GITHUB_WORKSPACE directory to a
  list of safe directories.

Fixes: https://github.com/389ds/389-ds-base/issues/5307

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -53,9 +53,9 @@ jobs:
       image: ${{ matrix.image }}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Checkout and configure
-        run: cd $GITHUB_WORKSPACE && autoreconf -fvi && ./configure
+        run: autoreconf -fvi && ./configure
         env:
           CC: ${{ matrix.compiler }}
           CXX: ${{ matrix.cpp-compiler }}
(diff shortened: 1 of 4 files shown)


bsimonova (approved):
LGTM!

PR #5310, opened 2022-05-26
Issue 5304: Need a compatibility option about sub suffix handling

### LINKED ISSUE #5304: Need a compatibility option about sub suffix handling
**Issue Description**

Some time ago [Issue 4373](https://github.com/389ds/389-ds-base/issues/4373)
changed the mapping tree behavior to avoid having to configure explicitly the parent relationship.
Although that change is correct as it avoid a lot of configuration errors and inconsistencies, it breaks some users deployment that relies on the old behavior, so we need to had a new  parameter that explicitly break the parent relationship. 

 Since Fedora 1.4.2 or RHDS 11

**Steps to Reproduce**
1. Create dc=parent and dc=child,dc=parent suffix, populate them
2. Search all entries under dc=parent
3. You should see entries of both suffix, but with version before Issue 4373 you do not see the entries under dc=child,dc=parent

**Expected results**
Although the current behavior is the correct one, we need a way to revert to the old behavior to avoid breaking some user deployment.

**External links**
- [Bugzilla 2083589](https://bugzilla.redhat.com/show_bug.cgi?id=2083589)
- [JIRA Issue IDMDS-2289](https://issues.redhat.com/browse/IDMDS-2289)

progier389:
Issue #4373 has obsoleted the nsslapd-parent-suffix config attribute 
(this attribute is now ignored )
This fixed a common configuration error but that cause regression 
to some users deployment that requires that search does not get 
forwarded to sub suffix (as it was the case in old versions 
if nsslapd-parent-suffix was not configured.)

The solution is to add an "orphan" compatibility attribute to the 
mapping tree entry that force the parent suffix to be the root entry 
(ignoring the dn relationship)

This attribute may be configured by using
  dsconf instance backend suffix --enable-orphan bename
  dsconf instance backend suffix -disable-orphan bename

Issue #5304

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/mapping_tree/regression_test.py
+++ b/dirsrvtests/tests/suites/mapping_tree/regression_test.py
@@ -1,0 +1,130 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+
+import ldap
+import logging
+import os
+import pytest
+from lib389.backend import Backends, Backend
+from lib389.dbgen import dbgen_users
+from lib389.mappingTree import MappingTrees
+from lib389.topologies import topology_st
+
+try:
+    from lib389.backend import BackendSuffixView
+    has_orphan_attribute = True
+except ImportError:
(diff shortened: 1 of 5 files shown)


progier389:
Test showed a small error in the test case cleanup phase: Should order the backend list such a way that the 
 children backends  get deleted before the parent one.

mreynolds389 (approved):
LGTM

PR #5312, opened 2022-05-27
Issue 5311 - Missing Requires for acl in the spec file

### LINKED ISSUE #5311: Missing `Requires` for acl in the spec file
**Issue Description**
We have fixed this in Fedora downstream spec files (see https://bugzilla.redhat.com/show_bug.cgi?id=2057436, https://bugzilla.redhat.com/show_bug.cgi?id=2083552) but not in the upstream spec file. This causes tests to error out on F36 in GH actions: 
https://github.com/389ds/389-ds-base/actions/runs/2389723539/attempts/1

vashirov:
Description:
Add missing Requires for acl package that contains setfacl used by
ds_systemd_ask_password_acl script.

Fixes: https://github.com/389ds/389-ds-base/issues/5311

Reviewed by: ???

### DIFF
--- a/rpm/389-ds-base.spec.in
+++ b/rpm/389-ds-base.spec.in
@@ -163,6 +163,7 @@ Requires:         policycoreutils-python-utils
 Requires:         libsemanage-python%{python3_pkgversion}
 # the following are needed for some of our scripts
 Requires:         openldap-clients
+Requires:         acl
 # this is needed to setup SSL if you are not using the
 # administration server package
 Requires:         nss-tools


vashirov:
Without this change tests are failing in PR CI using F36 image. Once this is merged, we can switch from F35 to F36.

droideck (approved):
LGTM!

mreynolds389:
> Without this change tests are failing in PR CI using F36 image. Once this is merged, we can switch from F35 to F36.

I had made this change to rawhide a month or two ago, but it is not present in F36 and older.  Not sure what happened there (I don't think I did new Fedora builds since that change).

vashirov:
Tests in PR CI use rpm package built from source, not from Fedora repos. So we need to update our spec file in tree as well.

Issue #5170, opened 2022-02-22
Filter optimiser


Firstyear:
LDAP queries are evaluated by our backend in the order of provided terms, which can cause surprising performance issues. These have tried to be mitigated with tools like the idl scan limit, but a proper query optimiser will help us with non optimal searches.

mreynolds389:
PR:  https://github.com/389ds/389-ds-base/pull/5171

mreynolds389:
@Firstyear can you cherry pick this down to 2.0?

Firstyear:
Sure, if you'd like :)

Firstyear:
```
   31f6c444a..b6d2fa4e4  389-ds-base-2.1 -> 389-ds-base-2.1
   0a1dc2a86..a444d3454  389-ds-base-2.0 -> 389-ds-base-2.0
```

Firstyear:
It'll be worth testing after that cherry-pick on the two branches just to be sure - there was a conflict in ldbm_search.c that I was careful in how I resolved it, but it'd be good for someone else to double check.

mreynolds389:
We have a regression, and it was picked in the github CI but it was overlooked:

py.test dirsrvtests/tests/suites/filter/complex_filters_test.py

This patch now returns ldapsubentries when it should not:

```
dn: cn=389_ds_system,dc=example,dc=com
objectClass: top
objectClass: nscontainer
objectClass: ldapsubentry
cn: 389_ds_system
```

Firstyear:
I did say there was a conflict and it was about subentry code, but I can't look today because I have a security issue to deal with :(

mreynolds389:
> I did say there was a conflict and it was about subentry code, but I can't look today

This is on master branch, so the cherry pick issues you had are unrelated.  If don't mind looking into it next week that would be great. 

> because I have a security issue to deal with :(

And are you dealing with a SQL injection attack? Fun fun...

Firstyear:
> > I did say there was a conflict and it was about subentry code, but I can't look today
> 
> This is on master branch, so the cherry pick issues you had are unrelated. If don't mind looking into it next week that would be great.

Yeah, I can look next week. 

> 
> > because I have a security issue to deal with :(
> 
> And are you dealing with a SQL injection attack? Fun fun...

That would be the one :(

mreynolds389:
Ok, so I started investigating the last regression with this mentioned in:

https://github.com/389ds/389-ds-base/issues/5289

[...]

PR #5315, opened 2022-05-30
Issue 5170 - BUG - incorrect behaviour of filter test

Firstyear:
Bug Description: In the filter test during access only
checks, OR conditions were not correctly evaluated. They
would have their access checked, but it was not confirmed
if this was the element that the entry matched. This mean
that queries could incorrectly reduce entries.

Fix Description: In the "access only" check mode, do not
handle it as a "special case", and apply the full OR check
algorithm to ensure that the component we are matching on
is indeed, the one we are accessing.

fixes: https://github.com/389ds/389-ds-base/issues/5170

Author: William Brown <william@blackhats.net.au>

Review by: ???

PR #5316, opened 2022-05-30
5170 revert fix CVE in access

Firstyear:
Bug Description: It was possible to bypass access controls
    by causing a shortcut query to 389-ds which would prematurely
    return a result set, where it would then not correctly
    enforce access over the entries. This may have allowed
    passwords to be readable.

    Fix Description: This sets the filter test threshold to 0
    which prevents an access control bypass condition. It also
    resolves a possible corrupt access check in attr filter
    test by properly enforcing what attributes we check are
    matching

    fixes: https://github.com/389ds/389-ds-base/issues/5170

    Review by: ???


Firstyear:
This CONFLICTS with #5315

Only ONE of these PR's can be merged!

Firstyear:
I would really rather we DO NOT merge this, but we use it as a template for backports to resolve the ACI bypass issue.

Issue #5324, opened 2022-06-01
plugin acceptance test needs hardening 


mreynolds389:
**Issue Description**

When running the dynamic plugins testsuite it calls the plugin acceptnace test: dirsrvtests/tests/suites/plugins/acceptance_test.py

```
_____________________________ test_stress ____________________________________________________

topology_m2 = <lib389.topologies.TopologyMain object at 0x7f5c866e4280>

    @pytest.mark.flaky(max_runs=2, min_passes=1)
    @pytest.mark.tier2
    def test_stress(topology_m2):
        """Test plugins while under a big load. Perform the test 5 times
    
        :id: 96136538-0151-4b09-9933-0e0cbf2c7863
        :setup: 2 Supplier Instances
        :steps:
            1. Pause all replication
            2. Set nsslapd-dynamic-plugins to on
            3. Try to update LDBM config entry
            4. Do one run through all tests
            5. Enable Referential integrity and MemberOf plugins
            6. Launch three new threads to add a bunch of users
            7. While we are adding users restart the MemberOf and
               Linked Attributes plugins many times
            8. Wait for the 'adding' threads to complete
            9. Now launch three threads to delete the users
            10. Restart both the MemberOf, Referential integrity and
                Linked Attributes plugins during these deletes
            11. Wait for the 'deleting' threads to complete
            12. Now make sure both the MemberOf and Referential integrity plugins still work correctly
            13. Cleanup the stress tests (delete the group entry)
            14. Perform 4-13 steps five times
            15. Resume replication
            16. Go through 4-14 steps once more
            17. Check that data in sync and replication is working
        :expectedresults:
            1. Success
            2. Success
            3. Success
            4. Success
            5. Success
            6. Success
            7. Success
            8. Success
            9. Success
            10. Success
            11. Success
            12. Success
            13. Success
            14. Success
            15. Success
            16. Success
            17. Success
        """
    
        m1 = topology_m2.ms["supplier1"]
        msg = ' (no replication)'
        replication_run = False
        stress_max_runs = 5
    
        # First part of the test should be without replication
        topology_m2.pause_all_replicas()
    
        # First enable dynamic plugins
        m1.config.replace('nsslapd-dynamic-plugins', 'on')
    
        # Test that critical plugins can be updated even though the change might not be applied
        ldbm_config = LDBMConfig(m1)
        ldbm_config.replace('description', 'test')
    
        while True:
            # First run the tests with replication disabled, then rerun them with replication set up
    
[...]

PR #5325, opened 2022-06-01
Issue 5324 - plugin acceptance test needs hardening

mreynolds389:
Description:

On slow VM's the schema & config update nmeedsa sleep after it before we
do the bind.

relates: https://github.com/389ds/389-ds-base/issues/5324


droideck (approved):
LGTM!

PR #5326, opened 2022-06-02
Issue 5323 - BUG - Fix issue in mdb tests with monitor

### LINKED ISSUE #5323: openldap to ds when a monitoring database is present causes an error
An error occurs as the "monitoring" databes looks like a database, but does not have a true suffix, and so can't be converted. In this case, when the db type is monitor, it should be skipped.

Firstyear:
Bug Description: due to the way that the monitor check worked
it would fail if the database name did not have a } in it.

Fix Description: Change the check to be a bit more robust to
various DB types.

fixes: https://github.com/389ds/389-ds-base/issues/5323

Review by: ???

### DIFF
--- a/src/lib389/lib389/migrate/openldap/config.py
+++ b/src/lib389/lib389/migrate/openldap/config.py
@@ -107,7 +107,8 @@ class olDatabase(object):
         assert len(entries) == 1
         self.config = entries.pop()
         self.log.debug(f"{self.config}")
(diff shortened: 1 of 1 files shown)


Firstyear:
@vashirov fixs the tests

@OttoHollmann If you can supply the slapd.d dir, we can make a new test suite to ensure monitors are correctly handled.

vashirov (approved):
LGTM, thanks!

PR #5328, opened 2022-06-03
Issue 5327 - Validate test metadata

### LINKED ISSUE #5327: Validate test metadata
**Issue Description**
Metadata in our tests' docstrings is used for test case and requirements import in our test case management system. But it's prone to copy-paste issues (duplicate unique test case IDs), formatting issues (RST parser is not always happy) or it's simply missing. We should use automatic validation for test metadata as part of our PR CI so that all new test cases have valid docstrings and unique IDs.

vashirov:
Bug Description:
Metadata in our tests' docstrings is used for test case and requirements
import in our test case management system. But it's prone to copy-paste
issues (duplicate unique test case IDs), formatting issues (RST parser
is not always happy) or it's simply missing. We should use automatic
validation for test metadata as part of our PR CI so that all new test
cases have valid docstrings and unique IDs.

Fix Description:
* Add a new Validate workflow to run on each PR. It runs testimony that
  validates docstrings
* Add missing docstrings.
* Rename test_user fixtures to avoid confusion with pytest test functions.
* Fix token names and indentation.
* Add a python script to check for duplicate `:id:` tokens
* Fix duplicate `:id:` tokens.

Fixes: https://github.com/389ds/389-ds-base/issues/5327

Reviewed by: ???

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,22 @@
+name: Validate tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/389ds/ci-images:test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Run testimony
+        if: always()
+        run: testimony validate -c dirsrvtests/testimony.yaml dirsrvtests/tests/suites
+
+      - name: Check for duplicate IDs
+        if: always()
+        run: python3 dirsrvtests/check_for_duplicate_ids.py dirsrvtests/tests/suites
--- a/dirsrvtests/check_for_duplicate_ids.py
+++ b/dirsrvtests/check_for_duplicate_ids.py
@@ -1,0 +1,38 @@
+import os
+import subprocess
+import sys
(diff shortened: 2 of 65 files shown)


vashirov:
I didn't fix one test intentionally to show the validation output and ask a question: https://github.com/389ds/389-ds-base/blob/419461df8c327c86c7778257ef9b6e877d280724/dirsrvtests/tests/suites/config/config_test.py#L287-L304

It shows that the code should be moved to a different test once the bugzilla is closed. Bugzilla is closed, but this test case still fails. Should I open a new bug for this?

droideck:
> It shows that the code should be moved to a different test once the bugzilla is closed. Bugzilla is closed, but this test case still fails. Should I open a new bug for this?

Looks like the behaviour was changed with the fix: https://github.com/389ds/389-ds-base/commit/fb3be041002931b94aa583646a170985783ae823

Now it's:

```
As for the default naming context, if there is a second suffix configured, it will be
automatically set as the new default naming context, otherwise the attribute is not
modified.
```

So the test is not valid and can be removed, IMHO.

droideck (approved):
PR looks good to me!
Thanks for implementing the change!

Issue #5329, opened 2022-06-03
Improve replication extended op logging


mreynolds389:
**Issue Description**

We are seeing some cases where the replication extended op payload can not be parsed.  We do not have any logging that provides any insight as to what is wrong with the payload

mreynolds389:
fd6c9514e..213cb5b6e  389-ds-base-2.1 -> 389-ds-base-2.1
3bec6edb4..e6da41762  389-ds-base-2.0 -> 389-ds-base-2.0
9d455da30..5b9933c05  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5330, opened 2022-06-03
Issue 5329 - Improve replication extended op logging

mreynolds389:
Description:

We need logging around parsing extended op payload, right now when it
fails we have no idea why.

relates: https://github.com/389ds/389-ds-base/issues/5329

### DIFF
--- a/ldap/servers/plugins/replication/repl_extop.c
+++ b/ldap/servers/plugins/replication/repl_extop.c
@@ -73,6 +73,18 @@ done:
     return rc;
 }
 
+static void
+ruv_dump_to_log(const RUV *ruv, char *log_name)
+{
+    if (!ruv) {
+        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name, "%s: RUV: None\n", log_name);
+    } else {
+        slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name, "%s: RUV:\n", log_name);
+        ruv_dump(ruv, log_name, NULL);
+    }
+}
+
+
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM, 
I just wonder if we should not also have a way to log the payloads on both side (maybe with a new log level?)
Because I remember having to deal with issues related to external smart network hardware
 that decoded the ldap protocol and re-encoded it wrongly. (we had to snoop the line on both way to find the root cause  but as encryption get the rule it will be harden to diagnose these days) anyway that could be a separate RFE and I am ok with current commit.

Firstyear:
Yeah, I think that's a different RFE @progier389. It's a shame we can't hide checksums in the messages that we send for verification of this. 

Anyway, ack from me, always on board for better logging.

mreynolds389:
> LGTM, I just wonder if we should not also have a way to log the payloads on both side (maybe with a new log level?) Because I remember having to deal with issues related to external smart network hardware that decoded the ldap protocol and re-encoded it wrongly. (we had to snoop the line on both way to find the root cause but as encryption get the rule it will be harden to diagnose these days) anyway that could be a separate RFE and I am ok with current commit.

@progier389  Ok I improved the logging to include more payload encoding/decoding debug info.  It might not be exactly what you were talking about but I feel its still a nice improvement.  Please review...

droideck (approved):
LGTM!
But I'll let folks check too. :)

Firstyear (approved):

progier389 (approved):
LGTM

Issue #5343, opened 2022-06-15
Various fixes around winsync


mreynolds389:
**Issue Description**

The logging could use more details when sync events fail

Also entries coming from AD are missing "sn" which is required by objectclass "person" and this prevents the syncing of those users

mreynolds389:
7c658cd63..fd6c9514e  389-ds-base-2.1 -> 389-ds-base-2.1
cc5aeb14e..3bec6edb4  389-ds-base-2.0 -> 389-ds-base-2.0
9fd81249b..9d455da30  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5344, opened 2022-06-15
Issue 5343 - Various improvements to winsync

mreynolds389:
Description:

In some sync cases the objectclass person is present, but we do not add
the required attribute "sn" to the mapped entry.  So this has been fixed.

Revised some of the logging to provide more details about sync failures

Updated an error with CLI around winysnc agreement configuration

Updated UI to better handle replication agreements

relates: https://github.com/389ds/389-ds-base/issues/5343

### DIFF
--- a/ldap/servers/plugins/replication/windows_protocol_util.c
+++ b/ldap/servers/plugins/replication/windows_protocol_util.c
@@ -1231,8 +1231,8 @@ process_replay_add(Private_Repl_Protocol *prp, Slapi_Entry *add_entry, Slapi_Ent
                 mapped_entry = NULL;
                 if (NULL == entryattrs) {
                     slapi_log_err(SLAPI_LOG_ERR, windows_repl_plugin_name,
-                                  "process_replay_add - %s - Cannot convert entry to LDAPMods.\n",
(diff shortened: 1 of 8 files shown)


ldap/servers/plugins/replication/windows_protocol_util.c
```diff
@@ -4381,8 +4371,9 @@ windows_create_local_entry(Private_Repl_Protocol *prp, Slapi_Entry *remote_entry
         /* Fatal error : need the guid */
         goto error;
     }
-    /* Hack for NT4, which has no surname */
-    if (is_nt4 && is_user) {
+
```
Firstyear, line 4374: Just think of all those sad nt4 dcs from 1996 you are just throwing out support for 🥺

Firstyear (approved):
All looks good, always good to have better logging.

ldap/servers/plugins/replication/windows_protocol_util.c
mreynolds389: > Just think of all those sad nt4 dcs from 1996 you are just throwing out support for pleading_face

I know, it's a sad day indeed ;-)

Issue #5353, opened 2022-06-23
dsconf backend tasks can break if there are multiple backends


mreynolds389:
**Issue Description**

If there are multiple backends "dsconf backend export" fails with a decoding error:

```
DEBUG: list filter = (&(objectclass=nsBackendInstance))
DEBUG: cn=changelog,cn=ldbm database,cn=plugins,cn=config getVal('cn')
DEBUG: cn=changelog,cn=ldbm database,cn=plugins,cn=config getVal('nsslapd-suffix')
DEBUG: 
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 138, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.9/site-packages/lib389/cli_conf/backend.py", line 282, in backend_export
    dn = _search_backend_dn(inst, be_name)
  File "/usr/lib/python3.9/site-packages/lib389/cli_conf/backend.py", line 97, in _search_backend_dn
    if cn == del_be_name or str2dn(suffix) == str2dn(del_be_name):
  File "/usr/lib64/python3.9/site-packages/ldap/dn.py", line 52, in str2dn
    return ldap.functions._ldap_function_call(None,_ldap.str2dn,dn,flags)
  File "/usr/lib64/python3.9/site-packages/ldap/functions.py", line 55, in _ldap_function_call
    result = func(*args,**kwargs)
ldap.DECODING_ERROR
ERROR: Error: 
```

mreynolds389:
7d9a7b52c..6280e2f4c  389-ds-base-2.1 -> 389-ds-base-2.1
a542b44f0..8daae835b  389-ds-base-2.0 -> 389-ds-base-2.0
3634fee51..deed28c5f  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
5b9933c05..93ff55814  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5354, opened 2022-06-23
Issue 5353 - CLI - dsconf backend export breaks with multiple backends

mreynolds389:
Description:

If there are multiple backends "dsconf backend export" fails with a decoding error
because we don't process mutliple backends correctly, and we attempt to
process a suffix/dn that is actually the backend name.

Fix Description:

Properly handle checking a backend name or suffix when attempting to
find the backend.  Also did some code cleanup.

relates: https://github.com/389ds/389-ds-base/issues/5353


droideck (approved):
LGTM!

PR #5357, opened 2022-06-27

bsimonova:
Description:
Add an additional run of tests with LMDB enabled in github actions.

Reviewed by: ???

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -93,11 +93,12 @@ jobs:
         sudo docker exec $CID sh -c "mkdir -p /workspace/assets/cores && chmod 777 /workspace{,/assets{,/cores}}"
         sudo docker exec $CID sh -c "echo '/workspace/assets/cores/core.%e.%P' > /proc/sys/kernel/core_pattern"
         sudo docker exec -e WEBUI=1 -e DEBUG=pw:api -e PASSWD="${PASSWD}" $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml --html=pytest.html --browser=firefox --browser=chromium -v dirsrvtests/tests/suites/${{ matrix.suite }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pytest.yml
```diff
@@ -93,11 +93,12 @@ jobs:
         sudo docker exec $CID sh -c "mkdir -p /workspace/assets/cores && chmod 777 /workspace{,/assets{,/cores}}"
         sudo docker exec $CID sh -c "echo '/workspace/assets/cores/core.%e.%P' > /proc/sys/kernel/core_pattern"
         sudo docker exec -e WEBUI=1 -e DEBUG=pw:api -e PASSWD="${PASSWD}" $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest.xml --html=pytest.html --browser=firefox --browser=chromium -v dirsrvtests/tests/suites/${{ matrix.suite }}
+        sudo docker exec -e NSSLAPD_DB_LIB=mdb -e WEBUI=1 -e DEBUG=pw:api -e PASSWD="${PASSWD}" $CID py.test  --suppress-no-test-exit-code  -m "not flaky" --junit-xml=pytest_lmdb.xml --html=pytest_lmdb.html --browser=firefox --browser=chromium -v dirsrvtests/tests/suites/${{ matrix.suite }}
```
vashirov, line 96: Correct me if I'm wrong: if the previous pytest run exits with non-zero code, tests with mdb won't be executed? 
It's not a blocker, but something we need to keep in mind.

I will investigate if we can parametrize our workflow in a better way without blowing up the test matrix. Maybe extract pytest workflow in its own action and reuse it.

bsimonova: I am not sure how it will behave with previous failed pytest run but I could make it fail on purpose with commit to see the results.

I also checked the failing tests in LMDB mode and they pass with build package, so I need to investigate more.

progier389:
Closing this PR because it is handled by [PR 5627](https://github.com/389ds/389-ds-base/pull/5627)

PR #5359, opened 2022-07-04
Issue 5221 - fix covscan

tbordaz:


progier389 (approved):
The change is harmless (No need to reset the connection credentials if there is no connection), but it really miss some context ( It is really hard to understand why the change is needed.
 I would have liked to have a non empty description ! -;)  )

Issue #5335, opened 2022-06-06
RFE - Create a Security Audit Log


mreynolds389:
Create a new server log that records "security" events:

- Authentication attacks
- Authorization issues
- DOS/TCP attacks
- etc

See this design doc for more details:

https://www.port389.org/docs/389ds/design/security-audit-log-design.html

Bugzilla:

https://bugzilla.redhat.com/show_bug.cgi?id=2093981

progier389:
> Other “high profile” accounts, how to specify them (multi-valued config attribute?

 I do not like too much the multi valued config idea
  because that means a somewhat costly lookup as we need to normalize the dns to compare the values
  I would rather add an auxilliary objectclass and an optional attribute to the entries
  since the entry is read during the authentication, cgecking the attribute would not be very costly ...
  (directory manager is an exception as there is no associated entry ... )
 That said it is a personal point of view and arguable ! -;)

Log format: 
  utc_time should be a floating number (i.e: 168485945.828151054) if you plan to use it as sorting key.

mreynolds389:
@progier389 - this is still very much a rough draft at that stage (I was just throwing out ideas in the doc).  Thanks for your comments, I'll update the design doc soon.

Firstyear:
The high profile account thing also can be "future" since any external log consumption system can apply that information

PR #5370, opened 2022-07-17
Issue 5335 - RFE - Add Security Audit Log

mreynolds389:
Description:

Create a new log using a JSON format that captures in a single line all
the relevant info (client, server, bind dn, event type, etc) related to
security events: Failed/success logins, TCP errors, authorization and
account lockout.

https://www.port389.org/docs/389ds/design/security-audit-log-design.html

relates: https://github.com/389ds/389-ds-base/issues/5335

ASAN: passed

### DIFF
--- a/dirsrvtests/tests/suites/logging/security_basic_test.py
+++ b/dirsrvtests/tests/suites/logging/security_basic_test.py
@@ -1,0 +1,241 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import ldap
+import json
+import logging
+import pytest
+import os
+import re
+import signal
+import subprocess
+import time
+from lib389._constants import DEFAULT_SUFFIX, PASSWORD, DN_DM
+from lib389.topologies import topology_st as topo
+from lib389.idm.user import UserAccount, UserAccounts
+from lib389.dirsrv_log import DirsrvSecurityLog
+from lib389.utils import ensure_str
+from lib389.idm.domain import Domain
+
+
+log = logging.getLogger(__name__)
+
+DN = "uid=security,ou=people," + DEFAULT_SUFFIX
+DN_NO_ENTRY = "uid=fredSomething,ou=people," + DEFAULT_SUFFIX
+DN_NO_BACKEND = "uid=not_there,o=nope," + DEFAULT_SUFFIX
+DN_QUOATED = "uid=\"cn=mark\",ou=people," + DEFAULT_SUFFIX
+DN_QUOATED_ESCAPED = "uid=cn\\3dmark,ou=people," + DEFAULT_SUFFIX
+DN_LONG = "uid=" + ("z" * 520) + ",ou=people," + DEFAULT_SUFFIX
+DN_LONG_TRUNCATED = "uid=" + ("z" * 508) + "..."
+
+
+@pytest.fixture
+def setup_test(topo, request):
+    """Disable log buffering"""
+    topo.standalone.config.set('nsslapd-securitylog-logbuffering', 'off')
+
+    """Add a test user"""
+    users = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    try:
+        users.create(properties={
+            'uid': 'security',
+            'cn': 'security',
+            'sn': 'security',
+            'uidNumber': '3000',
+            'gidNumber': '4000',
+            'homeDirectory': '/home/security',
+            'description': 'test security logging with this user',
+            'userPassword': PASSWORD
+        })
+    except ldap.ALREADY_EXISTS:
+        pass
+
+
+def check_log(inst, event_id, msg, dn=None):
+    """Check the security log
+    """
+    time.sleep(1)  # give a little time to flush to disk
+
+    security_log = DirsrvSecurityLog(inst)
+    log_lines = security_log.readlines()
+    for line in log_lines:
+        if re.match(r'[ \t]', line):
+            # Skip log title lines
+            continue
+
+    event = json.loads(line)
+    if dn is not None:
+        if event['dn'] == dn.lower() and event['event'] == event_id and event['msg'] == msg:
+            # Found it
+            return
+    elif event['event'] == event_id and event['msg'] == msg:
+            # Found it
+            return
+
+    assert False
+
+
+def test_invalid_binds(topo, setup_test):
+    """Test the various bind scenarios that should be logged in the security log
+
+    :id: b82e3fb9-f1af-4a75-8d96-5e5d284f31c5
+    :setup: Standalone Instance
+    :steps:
+        1. Test successful bind is logged
+        2. Test bad password is logged
+        3. Test no such entry is logged
+        4. Test no such entry is logged (quoated dn)
+        5. Test no such entry is logged (truncated dn)
+        6. Test no such backend is logged
+        7. Test account lockout is logged
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Success
+        7. Success
+    """
+
+    inst = topo.standalone
+    user_entry = UserAccount(inst, DN)
+
+    # Good bind
+    user_entry.bind(PASSWORD)
+    check_log(inst, "BIND_SUCCESS", "", DN)
+
+    # Bad password
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        user_entry.bind("wrongpassword")
+    check_log(inst, "BIND_FAILED", "INVALID_PASSWORD", DN)
+
+    # No_such_entry
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        UserAccount(inst, DN_NO_ENTRY).bind(PASSWORD)
+    check_log(inst, "BIND_FAILED", "NO_SUCH_ENTRY", DN_NO_ENTRY)
+
+    # No_such_entry (quoted)
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        UserAccount(inst, DN_QUOATED).bind(PASSWORD)
+    check_log(inst, "BIND_FAILED", "NO_SUCH_ENTRY", DN_QUOATED_ESCAPED)
+
+    # No such entry (truncated dn)
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        UserAccount(inst, DN_LONG).bind(PASSWORD)
+    check_log(inst, "BIND_FAILED", "NO_SUCH_ENTRY", DN_LONG_TRUNCATED)
+
+    # No_such_backend
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        UserAccount(inst, DN_NO_BACKEND).bind(PASSWORD)
+    check_log(inst, "BIND_FAILED", "NO_SUCH_ENTRY", DN_NO_BACKEND)
+
+
+def test_authorization(topo, setup_test):
+    """Test the authorization event by performing a modification that the user
+    is not allowed to do.
+
+    :id: 17a62670-f86d-4b39-9ee7-e7d36b973ec8
+    :setup: Standalone Instance
+    :steps:
+        1. Bind as a unprivileged user
+        2. Attempt to modify restricted resource
+        3. Security authorization event is logged
+    :expectedresults:
+        1. Success
+        2. Should fail
+        3. Success
+    """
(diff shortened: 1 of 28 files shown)


ldap/servers/slapd/bind.c
```diff
@@ -349,9 +349,9 @@ do_bind(Slapi_PBlock *pb)
             while (1) {
                 if (*pmech == NULL) {
                     /*
-                 * As we call the safe function, we receive a strdup'd saslmechanisms
-                 * charray. Therefore, we need to remove it instead of NULLing it
-                 */
+                     * As we call the safe function, we receive a strdup'd saslmechanisms
```
Firstyear, line 352: Should we run clang-format again?

ldap/servers/slapd/log.c
```diff
@@ -3013,137 +3204,969 @@ log__open_accesslogfile(int logfile_state, int locked)
...
+        /* MSG may be truncated */
+        PR_snprintf(local_time, sizeof(local_time), "vslapd_log_security, Unable to format system time");
+        log__error_emergency(local_time, 1, 0);
+        return -1;
+    }
+
+
+    /* build the format string, and truncate really long DN's */
+    strcpy(fmtstr, "{\"date\": \"%s\", \"utc_time\": %ld.%.09ld, \"event\": \"%s\", \"dn\": ");
+    if (slapi_sdn_get_size(bind_sdn) > 512) {
```
Firstyear: This opens us up to json injection and tampering attacks due to the fact that dn (bindsdn) is remotely controlled, meaning that this can cause log forgery. We can't use strcat/fmtstrs with remotely controlled inputs like this, we probably need to use a proper json library so write these strings so they are escaped correctly.

Firstyear (changes requested):

ldap/servers/slapd/log.c
mreynolds389: I thought since all these values have come through openldap ber decoding they should be safe.  But using a real library for the JSON building would be nice.  Have to look into that.

ldap/servers/slapd/log.c
```diff
@@ -3013,137 +3204,969 @@ log__open_accesslogfile(int logfile_state, int locked)
...
+        log_flush_buffer(lbi, SLAPD_SECURITY_LOG, 0 /* do not sync to disk */);
+    }
+    insert_point = lbi->current;
+    lbi->current += size;
+    /* Increment the copy refcount */
+    slapi_atomic_incr_64(&(lbi->refcount), __ATOMIC_RELEASE);
+    PR_Unlock(lbi->lock);
+
+    /* Now we can copy without holding the lock */
+    memcpy(insert_point, msg, size);
```
tbordaz, line 3782: Just curious, is there a risk that the log file get rotated between unlock and memcpy so that insert_point is invalid ?

ldap/servers/slapd/upgrade.c
```diff
@@ -146,6 +146,61 @@ upgrade_201_remove_des_rever_pwd_scheme(void)
     return UPGRADE_SUCCESS;
 }
 
+/*
+ * The replication plugin was renamed, this function cleans up all the
```
tbordaz: The comment is likely a bad cut/paste

ldap/servers/slapd/log.c
progier389: I am afraid that ber decoding function cannot help against injection because you are able to encode/decode any binary data through ber.  @Firstyear is right, the DN should be escaped.

mreynolds389: I wasn't disagreeing with @Firstyear, I really thought openldap validated DN's and would reject them if they were invalid. 
 Well then our other logs are subject to this because this is exactly how we print the DN in the access log (from the pblock target bind dn).  Or is this specific to JSON?

ldap/servers/slapd/bind.c
mreynolds389: We should, but not in this PR :-)

ldap/servers/slapd/log.c
mreynolds389: @tbordaz I'd have to check, but this is how the old code did it.  I just renamed the function, so if there is a problem, it has always been there.

ldap/servers/slapd/upgrade.c
mreynolds389: Fixed

ldap/servers/slapd/log.c
progier389: Although libber does not perform any check about tthe DN (it is an octet string at this level), the string that is stored in the security record is a valid DN because slapi_sdn_get_dn does normalize the DN and returns NULL if the DN is not parsable.
But a valid DN may contain characters that requires escape like \ otherwise it will be lost when decoding the JSON value and the security application wanting to use this value dn will not be able to parse the dn.

BTW It should probably be a good idea to add a test user like cn=Test with \"quotes\" in bind dn,dc=example, dc=foo 
    in your pytest test then to bind on this dn and see if the dn within the security log could be parsed as a dn 
     and is the expected dn.

mreynolds389: Thanks @progier389   I'm currently reworking the json stuff now.  Using "json-c" to build the object, but I have no idea yet if it will escape special characters.  I think it's supposed to.   I am also making sure we are using the normalized DN from the pblock (switched to slapi_sdn_get_ndn() vs slapi_sdn_get_dn()).  But yes I will try some complicated DN's in the CI test and see how it responds.   I won't have any new test results until tomorrow as I have meetings until the end of my work day today.

mreynolds389:
All requested changes have been made, please review...

dirsrvtests/tests/suites/logging/security_basic_test.py
```diff
@@ -0,0 +1,241 @@
...
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+    """
+
+    inst = topo.standalone
+
+    # Start interactive ldapamodfy command
+    ldap_cmd = ['ldapmodify', '-x', '-D', DN_DM, '-w', PASSWORD,
```
mreynolds389, line 223: Note to all, this test fails in github CI.  Trying to launch ldapmodify is not working and not sure why....

vashirov: There is no `ldapmodify` in the image. I'll add it.

ldap/servers/slapd/log.c
```diff
@@ -3013,137 +3205,995 @@ log__open_accesslogfile(int logfile_state, int locked)
...
+        slapi_pblock_get(pb, SLAPI_CONN_DN, &mod_dn);
+        PR_snprintf(binddn, sizeof(binddn), "%s", mod_dn ? mod_dn : "" /* anonymous */);
+        slapi_ch_free_string(&mod_dn);
+    } else {
+        /* For authz events we need the connection DN, but for all other events we use
+         * normalized bind target */
+        PR_snprintf(binddn, sizeof(binddn), "%s", slapi_sdn_get_ndn(bind_sdn));
+    }
+
+    if (binddn == NULL) {
```
progier389: Clang is right: these 3 lines are useless (starting at: if (binddn == NULL) { ...  )

progier389 (approved):
LGTM (But before merging you should remove that useless code that generates a Clang warning. -;) )

dirsrvtests/tests/suites/logging/security_basic_test.py
vashirov: Hmm, the test still fails, investigating...

dirsrvtests/tests/suites/logging/security_basic_test.py
```diff
@@ -0,0 +1,241 @@
...
+        '-H', f'ldap://{inst.host}:{inst.port}']
+    subprocess.Popen(ldap_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
vashirov: `ldapmodify` expects some input, but since in `Popen` `stdin=None` by default, it exits. And by the time `pidof` tries to find it, it's long gone.
```suggestion
    subprocess.Popen(ldap_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE)
```

ldap/servers/slapd/log.c
```diff
@@ -3013,137 +3205,991 @@ log__open_accesslogfile(int logfile_state, int locked)
...
+    conn_id = pb_conn->c_connid,
+    op_id = operation->o_opid,
+    client_ip = pb_conn->c_ipaddr;
+    server_ip = pb_conn->c_serveripaddr;
+    ldap_version = pb_conn->c_ldapversion;
+    if (saslmech) {
+        external_bind = !strcasecmp(saslmech, LDAP_SASL_EXTERNAL);
+    }
+
+    slapi_log_err(SLAPI_LOG_ALERT, "MARK", "event: %p\n", event_type); // DEBUG
```
vashirov: This logs 
```
[20/Jul/2022:15:32:22.010146800 +0000] - ALERT - MARK - event: 0x563919d14386
```
Should be removed, I guess?

mreynolds389: Whoops yeah that needs to come out.

dirsrvtests/tests/suites/logging/security_basic_test.py
mreynolds389: it doesn't exit on laptop though, but I will gladly make this change right now...  Give me one minute...

vashirov:
Security log has a header that makes the JSON non-compliant:
```
        389-Directory/2.2.2.202207201408git97e48bc B2022.201.1410
        server.example.com:38901 (/etc/dirsrv/slapd-standalone1)

{ "date": "[20\/Jul\/2022:15:32:05.603258906 +0000] ", "utc_time": "1658331125.603258906", "event": "BIND_SUCCESS", "dn": "cn=directory manager", "bind_method": "LDAPI", "root_dn": true, "client_ip": "local", "server_ip": "\/run\/slapd-standalone1.socket", "ldap_version": 3, "conn_id": 1, "op_id": 0, "msg": "" }
```
This makes it difficult to process it with tools just as `jq` or `gron`.

ldap/servers/slapd/log.c
Firstyear: @mreynolds389 openldap makes sure the DNs are valid for ldap, but that doesn't mean they can't contain characters that have meanings in json. :)

ldap/servers/slapd/bind.c
Firstyear: 👍

Firstyear:
> Anyway all our other logs have it, but I am fine removing it. Since it is a JSON log it makes sense to me to remove it, but then you do lose that info.

If we want the date string, we could format it specifically for this, or we rely on the utc time field

mreynolds389:
> > Anyway all our other logs have it, but I am fine removing it. Since it is a JSON log it makes sense to me to remove it, but then you do lose that info.
> 
> If we want the date string, we could format it specifically for this, or we rely on the utc time field

Well I don't know if the title is actually being used by anyone.  I know with DSEE we did **not** have a title, so it was added since Red Hat started work on 389.  I thought about adding it to the log name, but that is actually pretty complicated because of the way the logging code works.  If we really want to keep that "title" info in this log then perhaps we just add a JSON title, and use the "msg" element to store the host and port?  I don't have a strong opinion either way since I don't know if customers are actually using the current title

Firstyear:
> Well I don't know if the title is actually being used by anyone. I know with DSEE we did **not** have a title, so it was added since Red Hat started work on 389. I thought about adding it to the log name, but that is actually pretty complicated because of the way the logging code works. If we really want to keep that "title" info in this log then perhaps we just add a JSON title, and use the "msg" element to store the host and port? I don't have a strong opinion either way since I don't know if customers are actually using the current title

I think we should try to make these logs as "json as possible" rather than relying on json and then an extra "title parser" :)

ldap/servers/slapd/log.c
mreynolds389: Ok ok you guys can stop beating me up on this.  I thought openldap did all the necessary escaping.  I was wrong  :-)

ldap/servers/slapd/bind.c
mreynolds389: @Firstyear have I completed your requested changes?

ldap/servers/slapd/log.c
Firstyear: 🔨

Firstyear: It's all good mate, there are tons of layers involved so it's good we picked it up now and not later!

mreynolds389:
@firstyear - you officially requested changes, and it's blocking me from merging this.  Can you please finish the "review"?

Firstyear (approved):
Sorry for the delay, I didn't notice the pushed commit that added the json_object wrapper. All good :)

Issue #5375, opened 2022-07-21
CI - disable strict hostname checking 


mreynolds389:
**Issue Description**

Our testing environment is fragile, and sometimes there are hostname issues which causes all tests that use TLS to fail.  we should disable it by default for CI tests

mreynolds389:
3576e79e8..b90cf6901  389-ds-base-2.1 -> 389-ds-base-2.1
c0c099622..1f0217cee  389-ds-base-2.0 -> 389-ds-base-2.0
01a3d566b..3a358de00  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5376, opened 2022-07-21
Issue 5375 - CI - disable TLS hostname checking

mreynolds389:
Description:  
For CI tests we should disable TLS hostname checking
because our testing environments are fragile and TLS issues occur often.

relates: https://github.com/389ds/389-ds-base/issues/5375

### DIFF
--- a/src/lib389/lib389/cli_conf/backend.py
+++ b/src/lib389/lib389/cli_conf/backend.py
@@ -707,7 +707,6 @@ def backend_list_vlv(inst, basedn, log, args):
                     log.info(' - dn: ' + idx.dn)
                     for k, v in list(entry.items()):
                         log.info(' - {}: {}'.format(ensure_str(k), ensure_str(v[0])))
-                    log.info()
 
     if args.json:
         log.info(json.dumps({"type": "list", "items": results}, indent=4))
--- a/src/lib389/lib389/topologies.py
+++ b/src/lib389/lib389/topologies.py
@@ -19,6 +19,7 @@ from lib389.nss_ssl import NssSsl
 from lib389._constants import *
 from lib389.cli_base import LogCapture
 
+TLS_HOSTNAME_CHECK = os.getenv('TLS_HOSTNAME_CHECK', default=True)
 DEBUGGING = os.getenv('DEBUGGING', default=False)
 if DEBUGGING:
     logging.getLogger(__name__).setLevel(logging.DEBUG)
@@ -107,6 +108,10 @@ def _create_instances(topo_dict, suffix):
             # configured in a FIPS compliant way
             if is_fips():
                 instance.enable_tls()
+
+            # Disable strict hostname checking for TLS
+            if not TLS_HOSTNAME_CHECK:
+                instance.config.set('nsslapd-ssl-check-hostname', 'off')
             if DEBUGGING:
                 instance.config.set('nsslapd-errorlog-level','8192')
                 instance.config.set('nsslapd-accesslog-level','260')


tbordaz (commented):
I wonder if the yaml file used for the CI could add an environment variable and disable hostname checking only in CI.

mreynolds389:
> I wonder if the yaml file used for the CI could add an environment variable and disable hostname checking only in CI.

But topologies/py is only used by the CI - isn't that the same thing?

droideck:
> > I wonder if the yaml file used for the CI could add an environment variable and disable hostname checking only in CI.
> 
> But topologies/py is only used by the CI - isn't that the same thing?

I think Thierry meant GitHub CI...

And I agree. We may still want to have it enabled for other ways to run the test (and with the environment variable we can adjust it there too - for example, in downstream CI).

And if we run tests locally with the correct TLS hostname setup, we want to check that this feature works by default without modifying the topologies.py code.

mreynolds389:
@tbordaz @droideck do we know if we can pass an ENV var in the OSCI testing framework?

My previous argument was that hostname checking is a feature of openldap (not really DS).  I don't think it's something we actually "need" to test because all we do is a set a flag in LD struct, but I agree if we don't have to force it off for all scenarios then we shouldn't.

vashirov:
Yes, we can pass an environment variable. It should go in the `run` statement here:
https://src.fedoraproject.org/rpms/389-ds-base/blob/rawhide/f/tests/tests.yml#_22 (this is the upstream definition, but the same works for the downstream).

tbordaz:
> Yes, we can pass an environment variable. It should go in the `run` statement here: https://src.fedoraproject.org/rpms/389-ds-base/blob/rawhide/f/tests/tests.yml#_22 (this is the upstream definition, but the same works for the downstream).

do you mean that if we have this directive
run: "{{ pytest }} {{ pytest_args }} {{ pytest_tests }} OSCI=1"

Then in the launched test using topology.py, os.environ['OSCI'] can retrieve this variable ?

vashirov:
I was thinking of something like this:
```
  vars:
    ds_repo_url: https://github.com/389ds/389-ds-base.git
    ds_repo_dir: ds
    ds_tests: "{{ ds_repo_dir }}/dirsrvtests/tests"
    pytest: py.test-3
    pytest_args: "-v --continue-on-collection-errors"
    pytest_tests: "suites/basic"
    pytest_env_vars: "TLS_HOSTNAME_CHECK=False"  <---- New var
    artifacts: ./artifacts
```

and then add it to `run`:
```
run: "{{ pytest_env_vars }} {{ pytest }} {{ pytest_args }} {{ pytest_tests }}"
```

mreynolds389:
SO OSCI actually uses

> Yes, we can pass an environment variable. It should go in the `run` statement here: https://src.fedoraproject.org/rpms/389-ds-base/blob/rawhide/f/tests/tests.yml#_22 (this is the upstream definition, but the same works for the downstream).

Just to confirm this is what OSCI uses?  I thought OSCI was downstream/internal.

mreynolds389:
Added ENV var  support

vashirov:
OSCI is the name of the internal team/initiative that uses [Standard Test Interface aka STI](https://docs.fedoraproject.org/en-US/ci/standard-test-interface/) (and other interfaces) to run gating tests. STI can be used for Fedora gating tests too. So I'd rather not use OSCI as the name for the variable. I'd prefer a descriptive name for it.

In the downstream yaml file we actually have this:
```yaml
    pytest_run_command: "PYTHONPATH=../../src/lib389 {{ pytest }} {{ pytest_args }}"
```
So we already use some environment variables to alter pytest behaviour.

mreynolds389:
> OSCI is the name of the internal team/initiative that uses [Standard Test Interface aka STI](https://docs.fedoraproject.org/en-US/ci/standard-test-interface/) (and other interfaces) to run gating tests. STI can be used for Fedora gating tests too. So I'd rather not use OSCI as the name for the variable. I'd prefer a descriptive name for it.

I updated the PR to use TLS_HOSTNAME_CHECK as you suggested

tbordaz (approved):
LGTM

PR #5379, opened 2022-07-22
Issue 4656 - Remove problematic language from source code

mreynolds389:
Description:  Remove problematic language from source code


progier389 (approved):
LGTM

droideck (approved):

docker.mk
```diff
@@ -1,6 +1,6 @@
 
 suse:
-	docker build -t 389-ds-suse:master -f docker/389-ds-suse/Dockerfile .
```
vashirov, line 3: Here it's a tag that IIRC corresponds to a branch name. There is a special tag `latest` that I think would make more sense here.

ldap/schema/60kerberos.ldif
```diff
@@ -111,7 +111,7 @@ attributetypes: ( 2.16.840.1.113719.1.301.4.36.1 NAME 'krbPwdPolicyReference' EQ
 ##### The time at which the principal's password expires
 attributetypes: ( 2.16.840.1.113719.1.301.4.37.1 NAME 'krbPasswordExpiration' EQUALITY generalizedTimeMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE)
 ##### This attribute holds the principal's key (krbPrincipalKey) that is encrypted with
-##### the master key (krbMKey).
```
vashirov, line 114: In Kerberos terminology there is no primary key, see for example docs https://web.mit.edu/kerberos/krb5-1.20/doc/admin/admin_commands/kdb5_util.html
I think it would be confusing to refer to a non-existing thing, even though it's just a comment.
Also, our schema is a bit out of date with the upstream (we miss some new attributes):
https://github.com/krb5/krb5/blob/master/src/plugins/kdb/ldap/libkdb_ldap/kerberos.schema

docker.mk
mreynolds389: Ahh I was waiting for a container expert to comment on this :-)  I'll get this fixed

ldap/schema/60kerberos.ldif
mreynolds389: Yeah I wasn't sure what to do here.   Maybe I can just remove the comments?

I'll look into the new schema.  As long as nothing changed in existing attributes it should be ok to update it.  Have to check with IPA first as they might already be adding the updated schema...

docker.mk
Firstyear: latest makes more sense, that's okay :)

ldap/schema/60kerberos.ldif
Firstyear: I think we should keep the kerberos language because else it could be confusing. :(

mreynolds389: I agree changing the text is not a good idea.  Or, do we just remove the comments?  Odd to have design concepts in a schema file (not that it's a bad thing, it's just not typical and I wonder if anyone actually reads them).

docker.mk
mreynolds389: Since we renamed master to main in our repo, should this "main" now?

mreynolds389: > Since we renamed master to main in our repo, should this "main" now?

@vashirov @Firstyear I am going to change the tag to "main" first thing tomorrow, so if there are any objections let me know sooner than later :-)

Firstyear: No problems here.

vashirov: > Since we renamed master to main in our repo, should this "main" now?

`latest` is fine, thanks!

vashirov (approved):

Issue #5362, opened 2022-07-07
Nightly failure when re-installing a replica


flo-renaud:
**Issue Description**
FreeIPA nightly tests detected a failure in the test `test_topology_TestTopologyOptions`. See PR #[1841](https://github.com/freeipa-pr-ci2/freeipa/pull/1841) with the following [logs](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7d904abc-fd59-11ec-89b4-fa163e472ef6/) and [report](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7d904abc-fd59-11ec-89b4-fa163e472ef6/report.html).

**Package Version and Platform:**
 - Platform: Fedora 36
 - Package and version: 389-ds-base-2.1.3-1.fc36.x86_64
 Full packages list available [here](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7d904abc-fd59-11ec-89b4-fa163e472ef6/installed_packages/installed_packages_master.log.gz)

**Steps to Reproduce**
1. Install ipa server on master
2. Install client on replica1, promote to replica
3. Install client on replica2, promote to replica
4. Uninstall replica2 with ipa-replica-manage del + ipa-server-install --uninstall
5. Install client on replica2, promote to replica
The promotion fails:
```
RUN ['ipa-replica-install', '--admin-password', 'Secret.123', '--password', 'Secret.123', '-U', '--ip-address', '192.168.122.192', '--realm', 'IPA.TEST', '--domain', 'ipa.test', '--dirsrv-config-file', '/ipatests/ipatests_dse.ldif']
Lookup failed: Preferred host replica2.ipa.test does not provide DNS.
Reverse DNS resolution of address 192.168.122.34 (master.ipa.test) failed. Clients may not function properly. Please check your DNS setup. (Note that this check queries IPA DNS directly and ignores /etc/hosts.)
IPA client is already configured on this system, ignoring the --domain, --server, --realm, --hostname, --password and --keytab options.
Run connection check to master
Connection check OK
Disabled p11-kit-proxy
Configuring directory server (dirsrv). Estimated time: 30 seconds
  [1/42]: creating directory server instance
Validate installation settings ...
Create file system structures ...
Perform SELinux labeling ...
Create database backend: dc=ipa,dc=test ...
Perform post-installation tasks ...
  [2/42]: tune ldbm plugin
[...]

PR #5382, opened 2022-07-27
Issue 4324 - Change entry monitor to recursive pthread mutex

droideck:
Description: We've changed the entry cache monitor from PR_Monitor
to pthread recursive mutex. The entry lock is also protected
with with recursive mutex, which is implemented using PR_Monitor (NSPR).
Now, when we try to lock entry in cache_lock_entry we may stumble on
another bottleneck as entry lock still uses NSPR while entry cache
mutex uses a more performant pthread recursive mutex.

Fix Description:
Change the ep_mutexp from PR_Monitor to pthread recursive mutex.

Related: https://github.com/389ds/389-ds-base/issues/4324

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -337,7 +337,7 @@ struct backentry
     void *ep_dn_link;               /* linkage for the 3 hash */
     void *ep_id_link;               /*     tables used for */
     void *ep_uuid_link;             /*     looking up entries */
-    PRMonitor *ep_mutexp;           /* protection for mods; make it reentrant */
+    pthread_mutex_t *ep_mutexp __attribute__((__aligned__(64))); /* protection for mods; make it reentrant */
 };
 
 /* From ep_type through ep_create_time MUST be identical to backcommon */
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -1652,7 +1652,7 @@ cache_lock_entry(struct cache *cache, struct backentry *e)
         /* make sure only one thread does this */
         PR_Lock(cache->c_emutexalloc_mutex);
         if (!e->ep_mutexp) {
-            e->ep_mutexp = PR_NewMonitor();
+            e->ep_mutexp = slapi_pthread_mutex_alloc(PTHREAD_MUTEX_RECURSIVE);
             if (!e->ep_mutexp) {
                 PR_Unlock(cache->c_emutexalloc_mutex);
                 LOG("<= cache_lock_entry (DELETED)\n");
@@ -1667,13 +1667,13 @@ cache_lock_entry(struct cache *cache, struct backentry *e)
     }
 
     /* wait on entry lock (done w/o holding the cache lock) */
-    PR_EnterMonitor(e->ep_mutexp);
+    pthread_mutex_lock(e->ep_mutexp);
 
     /* make sure entry hasn't been deleted now */
     cache_lock(cache);
     if (e->ep_state & (ENTRY_STATE_DELETED | ENTRY_STATE_NOTINCACHE | ENTRY_STATE_INVALID)) {
         cache_unlock(cache);
-        PR_ExitMonitor(e->ep_mutexp);
+        pthread_mutex_unlock(e->ep_mutexp);
         LOG("<= cache_lock_entry (DELETED)\n");
         return RETRY_CACHE_LOCK;
     }
@@ -1688,8 +1688,8 @@ void
 cache_unlock_entry(struct cache *cache __attribute__((unused)), struct backentry *e)
 {
     LOG("=> cache_unlock_entry\n");
-    if (PR_ExitMonitor(e->ep_mutexp)) {
-        LOG("=> cache_unlock_entry - monitor was not entered!!!\n");
+    if (pthread_mutex_unlock(e->ep_mutexp)) {
+        LOG("=> cache_unlock_entry - pthread mutex was not unlocked!!!\n");
     }
 }
 
--- a/ldap/servers/slapd/back-ldbm/backentry.c
+++ b/ldap/servers/slapd/back-ldbm/backentry.c
@@ -29,7 +29,7 @@ backentry_free(struct backentry **bep)
         slapi_entry_free(ep->ep_entry);
     }
     if (ep->ep_mutexp != NULL) {
-        PR_DestroyMonitor(ep->ep_mutexp);
+        slapi_pthread_mutex_free(&ep->ep_mutexp);
     }
     slapi_ch_free((void **)&ep);
     *bep = NULL;


progier389:
I wonder if we should not directly use a pthread_mutex_t ep_mutexp; instead of a pointer 
  then init it instead of alloc it and destroy it instead of freeing it. 
It  avoids a needless malloc and will probably have better performance ...

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -337,7 +337,7 @@ struct backentry
     void *ep_dn_link;               /* linkage for the 3 hash */
     void *ep_id_link;               /*     tables used for */
     void *ep_uuid_link;             /*     looking up entries */
-    PRMonitor *ep_mutexp;           /* protection for mods; make it reentrant */
+    pthread_mutex_t *ep_mutexp __attribute__((__aligned__(64))); /* protection for mods; make it reentrant */
```
Firstyear, line 340: Why the allignment here?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1652,7 +1652,7 @@ cache_lock_entry(struct cache *cache, struct backentry *e)
         /* make sure only one thread does this */
         PR_Lock(cache->c_emutexalloc_mutex);
         if (!e->ep_mutexp) {
-            e->ep_mutexp = PR_NewMonitor();
+            e->ep_mutexp = slapi_pthread_mutex_alloc(PTHREAD_MUTEX_RECURSIVE);
```
Firstyear, line 1655: Recursive mutexes are really risky and generally point to a bigger architectural flaw - and can lead more easily to deadlocks if not handled correctly. 

Especially in the entry cache, which is very high contention, this may actually end up also causing more stalls.

So I'm very skeptical of making this recursive without at least exploring other non-recursive solutions.

Firstyear (changes requested):

ldap/servers/slapd/back-ldbm/cache.c
droideck: It's an interesting point...
I've done this PR because I wanted to bring consistency between this change: https://github.com/389ds/389-ds-base/pull/4569
And this function (where we use both entry cache mutex and entry mutex):
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/cache.c#L1647

This change is an attempt to fix this issue (and improve performance): https://github.com/389ds/389-ds-base/issues/5362 (which is not 100% reproducible, BTW, - I've tried a lot, and with high entry cache pressure - no luck)

Which happens (if I understand correctly) because we are not able to lock the loaded entry in time. And **LDBM_CACHE_RETRY_COUNT if** blocks us because we tried 1000 times:
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/findentry.c#L142

```
[06/Jul/2022:19:05:34.236431632 +0000] - ERR - find_entry_internal_dn - Retry count exceeded (cn=services,cn=accounts,dc=ipa,dc=test)
```

So I thought it may happen because NSPR monitor and thread recursive mutex have very different dynamics.

But... Could it be that the initial change caused it - https://github.com/389ds/389-ds-base/pull/4569 - because the recursive mutex has this ownership thing when the thread that grabs the mutex must be the same thread that releases the mutex? So I wonder if we may have a rare deadlock because of the recursive mutex nature.

I also thought about using atomics with AQUIRE/RELEASE so we have load/lock here - https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/findentry.c#L103 - happen atomically. But... It may cause bigger performance problems.

Firstyear: > It's an interesting point... I've done this PR because I wanted to bring consistency between this change: #4569 And this function (where we use both entry cache mutex and entry mutex): https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/cache.c#L1647
> 
> This change is an attempt to fix this issue (and improve performance): #5362 (which is not 100% reproducible, BTW, - I've tried a lot, and with high entry cache pressure - no luck)
> 
> Which happens (if I understand correctly) because we are not able to lock the loaded entry in time. And **LDBM_CACHE_RETRY_COUNT if** blocks us because we tried 1000 times: https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/findentry.c#L142
> 
> ```
> [06/Jul/2022:19:05:34.236431632 +0000] - ERR - find_entry_internal_dn - Retry count exceeded (cn=services,cn=accounts,dc=ipa,dc=test)
> ```
> 
> So I thought it may happen because NSPR monitor and thread recursive mutex have very different dynamics.

NSPR montior and pthread locks do have different dynamics, but being recursive may not help .... 

> 
> But... Could it be that the initial change caused it - #4569 - because the recursive mutex has this ownership thing when the thread that grabs the mutex must be the same thread that releases the mutex? So I wonder if we may have a rare deadlock because of the recursive mutex nature.

Yes, it's possible to deadlock like this. 

> 
> I also thought about using atomics with AQUIRE/RELEASE so we have load/lock here - https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/findentry.c#L103 - happen atomically. But... It may cause bigger performance problems.

Nooooppppppeeeeeee don't do that, almost every use of atomics from older code in the code base is wrong. It won't just be performance issues, but stability ones.

progier389: A few points:
  - Do not be worried about the "thread that grabs the mutex must be the same thread that releases the mutex" condition:
    In fact this condition must be true with pthread lock whether is mutex is recursive or not.
     ( The "Unlock When Not Owner" column in pthread_mutex_lock man page is quite clear! )
    And as far as I know ns-slapd and the plugin respects this condition
  - I do not share @firstyear concerns about recursive mutex (as far as the mutex implementation is correct)
     there is not much differences between a regular mutex and a recursive one 
     (IMHO, the main concern is that deadlocks may be a bit more painful to debug)
     but personally I am much more warry of read write locks than of recursive mutexes.
  - Like @firstyear, I do not think that atomic operation is the right solution to your problem:
     using atomic operation usually means having a retry loop and IMHO you risk to have the same issue 
     than the one you are trying to solve.
  - I am a bit surprise that you speak about deadlock as possible cause because that does not match the 
     customer symptoms (Is it what you got after replacing NSPR per pthread mutex ? )

droideck: So I've tried to reproduce the entry cache issue in different ways...

1. Modify a user entry's description attribute with a lot of modify operations. At the same time as there was a tombstone trimming set to a very small value.
2. Do add/delete operations in a loop and then try to search the same entry (still while tombstone trimming is active). No error.
3. I've run the FreeIPA test a few times, but still no luck.

We saw the error on FreeIPA PR CI tests with promoting client to replica, and it may indicate that the issue is not about heavy load but because something is wrong in the entry cache mutex interactions. And we know for sure that the issue doesn't happen if we go back to NSPR Monitor (so away from pthread recursive mutex).

We may revert all of the pthread recursive changes back to NSPR Monitor. We improve the stability but lose the performance gain (we had 8% search improvement after moving to pthread recursive).
I need to think more...

ldap/servers/slapd/back-ldbm/back-ldbm.h
droideck: I've done it according to c_mutex change.
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/back-ldbm.h#L374

Also, Thierry mentioned some time ago that CPU cache loads memory per lines. (and i7 cache line is 64 bytes)
So it makes sense to align the mutex data as it's a really sensitive area.

ldap/servers/slapd/back-ldbm/cache.c
Firstyear: > * I do not share @Firstyear concerns about recursive mutex (as far as the mutex implementation is correct)
>   there is not much differences between a regular mutex and a recursive one
>   (IMHO, the main concern is that deadlocks may be a bit more painful to debug)
>   but personally I am much more warry of read write locks than of recursive mutexes.

Given that the PR_Lock is not recursive, why do we need to add recursion now? 

The recursive counter adds some overhead I thought. I think to start with changing to pthread mutex is the right start, then add recursion *later* if it's actually required. 

> * Like @Firstyear, I do not think that atomic operation is the right solution to your problem:
>   using atomic operation usually means having a retry loop and IMHO you risk to have the same issue
>   than the one you are trying to solve.

^- this. And it's also infinitely harder to get write on weakly ordered CPU arches. 

> * I am a bit surprise that you speak about deadlock as possible cause because that does not match the
>   customer symptoms (Is it what you got after replacing NSPR per pthread mutex ? )

droideck: @Firstyear but we've changed PRMonitor, not PRLock... And PRMonitor works with counts, the same as pthread mutex recursive. Or do I miss something?

Regarding the atomic part - sure, makes 100% sense. :) I agree.

ldap/servers/slapd/back-ldbm/back-ldbm.h
Firstyear: You might find that re-ordering and aligning the struct as a whole is more effective than just sticking this allignment in the end ebcause this will just add overhead and padding.

ldap/servers/slapd/back-ldbm/cache.c
Firstyear: You're right, that's my mistake. I thought it was PR_lock.

ldap/servers/slapd/back-ldbm/back-ldbm.h
droideck: Do you mean something like this? (that we do in `slapi_pthread_mutex_alloc` function which I call later)
https://github.com/389ds/389-ds-base/blob/7efbeb3b75a5f879dec87a90ff925b436216c4e6/ldap/servers/slapd/ch_malloc.c#L402-L404

Or do you mean something additional?

Firstyear: No I mean that you should re-order the fields in struct backentry, such as:

```

struct backentry __attribute__((__aligned__(64)))
{
    pthread_mutex_t *ep_mutexp ;
    int32_t ep_type;                /* to distinguish backdn from backentry */
    int32_t ep_refcnt;              /* entry reference cnt */
    struct backcommon *ep_lrunext;  /* for the cache */
    struct backcommon *ep_lruprev;  /* for the cache */
    size_t ep_size;                 /* for cache tracking */
    Slapi_Entry *ep_entry;          /* real entry */
    Slapi_Entry *ep_vlventry;
    void *ep_dn_link;               /* linkage for the 3 hash */
    void *ep_id_link;               /*     tables used for */
    void *ep_uuid_link;             /*     looking up entries */
    struct timespec ep_create_time; /* the time the entry was added to the cache */
    ID ep_id;                       /* entry id */
    uint8_t ep_state;               /* state in the cache */

```

This aligns the whole struct, puts the mutex as the first element so it inherits the alignment, and then re-arranges the other fields so that they minimise padding between them. For a better explainer - http://www.catb.org/esr/structure-packing/

progier389: @Firstyear idea makes sense especially if we directly wrap the mutex within the backentry rather than allocating it: i.e:  pthread_mutex_t ep_mutexp;
( Otherwise changing the position of a pointer would not change much in term of alignment (even if it still  limits the padding ...))

droideck: > This aligns the whole struct, puts the mutex as the first element so it inherits the alignment, and then re-arranges the other fields so that they minimise padding between them. For a better explainer - http://www.catb.org/esr/structure-packing/

Awesome read, thank you so much! And I loved his style too! (somehow reminded me of the docs style that Rust is written with)

I think it's a great idea and we should look into it. I'm just not sure that it will solve the issue I'm trying to resolve with this PR.
The issue is when we are stuck in the loop and get 1000 runs which results in `Retry count exceeded` error.
https://github.com/389ds/389-ds-base/blob/7efbeb3b75a5f879dec87a90ff925b436216c4e6/ldap/servers/slapd/back-ldbm/findentry.c#L102-L144

I'll play around with it a bit more but I think it's worth a separate issue as we need carefully observe the structs within structs so we can properly order them. So, in my opinion, it'll be better to check and fix other "connected" structs too.

For now, I'd like to change PRMonitor to thread recursive mutex.
I'm also searching for another place where we may get stuck with an invalid entry, - maybe something is not decreasing ep_refcnt properly at some point... But it's hard to say as I still was unable to reproduce, - the issue is pretty rare. 

> idea makes sense especially if we directly wrap the mutex within the backentry rather than allocating it: i.e: pthread_mutex_t ep_mutexp;

If I understand correctly, @tbordaz initially made the change so we can benefit from the alignment as discussed in these comments:
https://github.com/389ds/389-ds-base/pull/4569#pullrequestreview-575651860

So if I understand correctly, it may be a downgrade if I remove slapi_ch_memalign (which is basically malloc with some extra guarantees).

Firstyear: It's more that to do this properly, we shouldn't just "throw in" an alignment in the middle of the struct, we should actually put this in *correctly*.

tbordaz: The purpose of this ticket is to change a PR_MONITOR (nspr) with a pthread mutex (recursive). Some tests showed that pthread mutex were faster.

Alignment is separated subject. For example ep_uuid_link (or ep_id_link) are used by any thread doing a cache_find_. So a thread doing a search, can access ep_id_link on a CPU while another thread doing an update (cache_lock_entry) on another CPU accesses ep_mutexp. Without alignment, those two fields ep_id_link and ep_mutexp are in the same CPU cacheline (64 bytes). So those two threads are creating a fake contention because they are competing for the same cacheline. ep_id_link should be accessible independently of ep_mutexp.
I think we should review all those structures to make sure there is no fake contention but it is another ticket. Here the alignment pragma looks good to me because there is a risk of fake contention.

tbordaz: I agree with @Firstyear  and @progier389 suggestion to not allocate ep_mutexp and rather wrap it into the structure.

mreynolds389:
Ok in regards to the alignment of backentry, I just want to point out that the backcommon struct expects backentry struct and backdn struct to have the same initial struct layout.  They must all be kept in sync or we will have entry cache issues or worse.

Firstyear:
> Ok in regards to the alignment of backentry, I just want to point out that the backcommon struct expects backentry struct and backdn struct to have the same initial struct layout. They must all be kept in sync or we will have entry cache issues or worse.

If they expect the same struct layout, then why isn't backentry including backcommon, rather than rely on the headers to be kept insync? Seems like  ahuge error waiting to happen ....

mreynolds389:
> > Ok in regards to the alignment of backentry, I just want to point out that the backcommon struct expects backentry struct and backdn struct to have the same initial struct layout. They must all be kept in sync or we will have entry cache issues or worse.
> 
> If they expect the same struct layout, then why isn't backentry including backcommon, rather than rely on the headers to be kept insync? Seems like ahuge error waiting to happen ....

backcommon is used to represent both a backdn and backentry before it goes into the cache.  You can see it in action in cache_add().

ldap/servers/slapd/back-ldbm/back-ldbm.h
droideck: @Firstyear @progier389 @tbordaz I've created an issue for the discussed improvement:
https://github.com/389ds/389-ds-base/issues/5402

For this PR, I'd like to keep it just about fixing the dynamics where we have a mix of NSPR and pthread mutex (backentry and cache structs, respectively)

Also, I'll continue to look into other reasons why we may get... https://github.com/389ds/389-ds-base/blob/7efbeb3b75a5f879dec87a90ff925b436216c4e6/ldap/servers/slapd/back-ldbm/findentry.c#L143
I'll look into more options where we may get INVALID or NOTINCACHE state - maybe something around the replication area...

Firstyear: > For this PR, I'd like to keep it just about fixing the dynamics where we have a mix of NSPR and pthread mutex (backentry and cache structs, respectively)

All good :)

droideck:
As we've decided to go with the revert solution (see https://github.com/389ds/389-ds-base/pull/5463), dropping this PR.

PR #5387, opened 2022-07-29
Issue 5386 - BUG - Update sudoers schema to correctly support UTF-8

### LINKED ISSUE #5386: Update sudoers schema to correctly support UTF-8
It was found by a SUSE customer, that an inconsistency between ldap core schemas and sudoers existed. This was due to items like uid and cn being UTF-8, but sudoUser and other types were IA5 only. This created a scenario where sudo rules in ldap were failing due to incorrect matching being applied (trying to match a UTF8 type with IA5 rules.). 

Since UTF-8 is a superset of IA5 the sudoers schema can be expanded to support UTF-8 without any other changes. This was discussed with the sudo maintainers at SUSE, and the sudo project itself accepted the changes to the schema. Since we ship a copy of the schema, we should update to be in line.

As this affected a SUSE customer, I'd like to apply this to 2.0 as well which is our currently shipping version in SP4. 

https://github.com/sudo-project/sudo/pull/163

Firstyear:
Bug Description: It was found by a SUSE customer, that an inconsistency between ldap core schemas
and sudoers existed. This was due to items like uid and cn being UTF-8, but sudoUser and other
types were IA5 only. This created a scenario where sudo rules in ldap were failing due to incorrect
matching being applied (trying to match a UTF8 type with IA5 rules.).

Fix Description: Since UTF-8 is a superset of IA5 the sudoers schema can be expanded to support
UTF-8 without any other changes. This was discussed with the sudo maintainers at SUSE, and the sudo
project itself accepted the changes to the schema. Since we ship a copy of the schema, we should
update to be in line.

fixes: https://github.com/389ds/389-ds-base/issues/5386

Review by: ???

### DIFF
--- a/ldap/schema/60sudo.ldif
+++ b/ldap/schema/60sudo.ldif
@@ -9,9 +9,9 @@ attributeTypes: (
   1.3.6.1.4.1.15953.9.1.1
   NAME 'sudoUser'
   DESC 'User(s) who may  run sudo'
-  EQUALITY caseExactIA5Match
-  SUBSTR caseExactIA5SubstringsMatch
-  SYNTAX 1.3.6.1.4.1.1466.115.121.1.26
+  EQUALITY caseExactMatch
+  SUBSTR caseExactSubstringsMatch
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15
   X-ORIGIN 'SUDO' )
 #
 ################################################################################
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

droideck (approved):
Ack from me too:)

Firstyear:
```
   9c30ed063..fbcd472dd  389-ds-base-2.1 -> 389-ds-base-2.1
   d606a2bb8..e2a858a86  389-ds-base-2.0 -> 389-ds-base-2.0
```

PR #5390, opened 2022-08-01
Issue 4308 - checking if an entry is a referral is expensive

mreynolds389:
Description:

When updating the entry cache check if the entry has a smart referral, if
so set a flag so we don't have to do the referral check for future
searches.

relates: https://github.com/389ds/389-ds-base/issues/4308

### DIFF
--- a/ldap/servers/slapd/back-ldbm/findentry.c
+++ b/ldap/servers/slapd/back-ldbm/findentry.c
@@ -32,8 +32,13 @@ check_entry_for_referral(Slapi_PBlock *pb, Slapi_Entry *entry, char *matched, co
     struct berval **url = NULL;
 
     /* if the entry is a referral send the referral */
+    if (!slapi_entry_flag_is_set(entry, SLAPI_ENTRY_FLAG_REFERRAL)) {
+        return 0;
+    }
+
     if (slapi_entry_attr_find(entry, "ref", &attr)) {
-        // ref attribute not found
+        // ref attribute not found (should not happen)
+        PR_ASSERT(0);
         goto out;
     }
 
--- a/ldap/servers/plugins/replication/repl5_total.c
+++ b/ldap/servers/plugins/replication/repl5_total.c
@@ -705,7 +705,7 @@ decode_total_update_extop(Slapi_PBlock *pb, Slapi_Entry **ep)
         bv.bv_val = "ldapsubentry";
         bv.bv_len = strlen(bv.bv_val);
         if (slapi_attr_value_find(attr, &bv) == 0) {
-            slapi_entry_set_flag(e, SLAPI_ENTRY_LDAPSUBENTRY);
+            slapi_entry_set_flag(e, SLAPI_ENTRY_FLAG_LDAPSUBENTRY);
         }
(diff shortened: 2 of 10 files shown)


ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -32,8 +32,12 @@ check_entry_for_referral(Slapi_PBlock *pb, Slapi_Entry *entry, char *matched, co
     struct berval **url = NULL;
 
     /* if the entry is a referral send the referral */
+    if (!slapi_entry_flag_is_set(entry, SLAPI_ENTRY_FLAG_REFERRAL)) {
+        return 0;
+    }
+
     if (slapi_entry_attr_find(entry, "ref", &attr)) {
-        // ref attribute not found
+        // ref attribute not found (should not happen)
```
Firstyear, line 40: If this shouldn't happen, can we add a DEBUG_ASSERT here?

Firstyear (changes requested):

ldap/servers/slapd/back-ldbm/findentry.c
mreynolds389: > If this shouldn't happen, can we add a DEBUG_ASSERT here?

Fixed

mreynolds389:
I was worried about a performance hit when adding/replacing entries in the cache.  So I ran a basic performance test:  10K adds, 10k mods, and 10k deletes.  I did not see any noticeable performance hit.

ldap/servers/slapd/back-ldbm/findentry.c
mreynolds389: @Firstyear - I made your requested change, can you approve it?

Firstyear (approved):

Firstyear:
Sorry, for some reason I totally missed this :(

mreynolds389:
@tbordaz - I know you wanted to look at this PR.  Can you check it over before I merge it?

ldap/servers/slapd/back-ldbm/findentry.c
tbordaz: I like ASSERT but it hides the benefit of the fix.

The issue is that most of the time the entry has no "ref" attribute. slapi_entry_attr_find("ref) will go through the complete list of attributes, doing strcasecmp on each of them to finally find nothing and return 0. Avoiding slapi_entry_attr_find to test referral is the purpose of that fix. Can we put the test in #ifdef DEBUG ?

tbordaz (commented):
The fix looks good to me except the test of "ref" for assert. That should only be done in debug mode.

ldap/servers/slapd/back-ldbm/findentry.c
tbordaz: Sorry after a second thought I realize my mistake. slapi_entry_attr_find is only called when the entry is a referral, that is quite rare. No problem in such case to call slapi_entry_attr_find

tbordaz (approved):
After a second thought the fix looks good to me. Sorry for the noise

PR #5396, opened 2022-08-03
Issue 3903 - Supplier should do periodic updates

mreynolds389:
Description:

On suppliers update the keep alive entry periodically to keep the RUV up
to date in case a replica is neglected for al ong time.  This prevents
very long changelog scans when finally processing updates.


ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -22,6 +22,7 @@
 #include "slap.h"
 
 #define RUV_SAVE_INTERVAL (30 * 1000) /* 30 seconds */
+#define REPLICA_UPDATE_INTERVAL (60 * 1000) /* 1 minute */
```
tbordaz: This value seems too frequent to me. In a 4 suppliers topology, the keep alive traffic will be +200 updates per hour that is high compare to regular traffic.

I think a low value is useful for test reason. An option would be to have a replica configuration parameter with a higher default value for example an update per hour. Then for the test we can set it to an update per minute and verify that it occurs.

The rest of the patch looks good to me

tbordaz (commented):
a small concern regarding the frequency of keepalive update.

ldap/servers/plugins/replication/repl5_replica.c
Firstyear: Active Directory defaults to once every 5 minutes or 20 times per hour, which I think would be enough.

mreynolds389: Sorry I wanted to mention that about the interval.  I had it set to 1 minute for easier testing.  

The only issue we have is the CI tests. If we bump it to 10, 30, or 60 minutes its going to add a lot of time to the CI test suite.  Or, I adjust the test suite and not check for the next interval.  Our CI tests already take a very long time to run so I am just being cautious.

tbordaz: If you add a new replica configuration parameter with a 3600s default value, you may use it to select 60s for testing.

mreynolds389: @tbordaz true we could do that, but is there value in a new config setting for customers?  There could be I suppose.  

We definitely need to test this feature in our CI, and it would not be a dangerous config setting if misused by an admin (if we set a minimum requirement on the interval).  So I will work on that...

mreynolds389:
@tbordaz @Firstyear 

Set the default to 1 hour, made it configurable with a minimum of 60 seconds, updated CLI/UI, and updated several CI tests.  Please review...

Firstyear (approved):

tbordaz (approved):
The config parameter is nsds5ReplicaRefreshRUVInterval while internally the define/var dropped the 'RUV' part (REPLICA_REFRESH_INTERVAL_MIN, DEFAULT_REPLICA_REFRESH_INTERVAL, type_replicaRefreshInterval and repl_refresh_interval). I would prefer a consistent naming.

Regarding the name itself I was initially not very fan of 'refresh' the RUV but I did not find a better option. With a rapid read we may think the RUV is out of sync and needs to be refreshed. The goal being to get a recent RUVelement but should we express that and how ? I have no idea.

Except those minor points the patch looks good to me

mreynolds389:
> The config parameter is nsds5ReplicaRefreshRUVInterval while internally the define/var dropped the 'RUV' part (REPLICA_REFRESH_INTERVAL_MIN, DEFAULT_REPLICA_REFRESH_INTERVAL, type_replicaRefreshInterval and repl_refresh_interval). I would prefer a consistent naming.
> 
> Regarding the name itself I was initially not very fan of 'refresh' the RUV but I did not find a better option. With a rapid read we may think the RUV is out of sync and needs to be refreshed. The goal being to get a recent RUVelement but should we express that and how ? I have no idea.

I agree, I don't like any of the names either (and they should be consistent).  Let me see what I can come up with...

mreynolds389:
> > The config parameter is nsds5ReplicaRefreshRUVInterval while internally the define/var dropped the 'RUV' part (REPLICA_REFRESH_INTERVAL_MIN, DEFAULT_REPLICA_REFRESH_INTERVAL, type_replicaRefreshInterval and repl_refresh_interval). I would prefer a consistent naming.
> > Regarding the name itself I was initially not very fan of 'refresh' the RUV but I did not find a better option. With a rapid read we may think the RUV is out of sync and needs to be refreshed. The goal being to get a recent RUVelement but should we express that and how ? I have no idea.
> 
> I agree, I don't like any of the names either (and they should be consistent). Let me see what I can come up with...

Renamed the following:

```
nsds5ReplicaRefreshRUVInterval     -> nsds5ReplicaKeepAliveUpdateInterval
DEFAULT_REPLICA_REFRESH_INTERVAL   -> DEFAULT_REPLICA_KEEPALIVE_UPDATE_INTERVAL
REPLICA_REFRESH_INTERVAL_MIN       -> REPLICA_KEEPALIVE_UPDATE_INTERVAL_MIN
Slapi_Eq_Context repl_eqcxt_update -> Slapi_Eq_Context repl_eqcxt_ka_update;
int64_t repl_refresh_interval      -> int64_t keepalive_update_interval;
type_replicaRefreshInterval        -> type_replicaKeepAliveUpdateInterval
```

mreynolds389:
This is causing regressions with replication.  Even crashes:

```
0x00007f16498d5caf in csn_init_by_csn (csn1=0x0, csn2=0x7f1619826d90) at ../389-ds-base/ldap/servers/slapd/csn.c:94
94	        memcpy(csn1, csn2, sizeof(CSN));
(gdb) where
#0  0x00007f16498d5caf in csn_init_by_csn (csn1=0x0, csn2=0x7f1619826d90) at ../389-ds-base/ldap/servers/slapd/csn.c:94
#1  0x00007f1645182bf6 in clcache_initial_anchorcsn (buf=0x7f16198c4800, dbop=0x7f16167fae9c) at ../389-ds-base/ldap/servers/plugins/replication/cl5_clcache.c:705
#2  0x00007f1645181d8e in clcache_load_buffer (buf=0x7f16198c4800, anchorCSN=0x7f16167faf30, continue_on_miss=0x7f16167fafb4, initial_starting_csn=0x0)
    at ../389-ds-base/ldap/servers/plugins/replication/cl5_clcache.c:338
#3  0x00007f16451803e7 in _cl5PositionCursorForReplay (consumerRID=65535, consumerRuv=0x7f161985f820, replica=0x7f16270ddb40, iterator=0x7f16167fb040, continue_on_missing=0x7f16167fafb4)
    at ../389-ds-base/ldap/servers/plugins/replication/cl5_api.c:3982
#4  0x00007f16451792cb in cl5CreateReplayIterator (prp=0x7f1619841000, consumerRuv=0x7f161985f820, iterator=0x7f16167fb040) at ../389-ds-base/ldap/servers/plugins/replication/cl5_api.c:995
#5  0x00007f16451a4fc1 in send_updates (prp=0x7f1619841000, remote_update_vector=0x7f161985f820, num_changes_sent=0x7f16167fb150)
    at ../389-ds-base/ldap/servers/plugins/replication/repl5_inc_protocol.c:1559
#6  0x00007f16451a3d7c in repl5_inc_run (prp=0x7f1619841000) at ../389-ds-base/ldap/servers/plugins/replication/repl5_inc_protocol.c:1046
#7  0x00007f16451ac4e9 in prot_thread_main (arg=0x7f1627147ef0) at ../389-ds-base/ldap/servers/plugins/replication/repl5_protocol.c:252
#8  0x00007f16493f1413 in _pt_root () from target:/lib64/libnspr4.so
#9  0x00007f16490f9e2d in start_thread () from target:/lib64/libc.so.6
#10 0x00007f164917f1b0 in clone3 () from target:/lib64/libc.so.6
```

PR #5400, opened 2022-08-05
Issue 5399 - UI - LDAP Editor is not updated when we switch instances

### LINKED ISSUE #5399: UI - LDAP Editor is not updated when we switch instances
**Issue Description**
We don't refresh LDAP Editor when we switch instances. It may lead to unpleasant errors.

**Steps to Reproduce**
Steps to reproduce the behaviour:
1. Go to the LDAP Editor tab
2. Switch instance
3. The list of entries is not updated

**Expected results**
The list of entries is updated.

droideck:
Description: We don't refresh LDAP Editor when we switch instances.
It may lead to unpleasant errors.

Add componentDidUpdate function with the appropriate processing and
properties.

Fixes: https://github.com/389ds/389-ds-base/issues/5399

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/LDAPEditor.jsx
+++ b/src/cockpit/389-console/src/LDAPEditor.jsx
@@ -60,6 +60,7 @@ export class LDAPEditor extends React.Component {
 
         this.state = {
             activeTabKey: 0,
+            firstLoad: true,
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
Ack!

Issue #5397, opened 2022-08-04
Memory leak in filter optimizer


mreynolds389:
**Issue Description**

Looks like we have some memory leaks caused by the filter optimizer RFE:

```
Indirect leak of 120 byte(s) in 10 object(s) allocated from:
    #0 0x7fc6a5cae377 in strdup (/lib64/libasan.so.6+0x5a377)
    #1 0x7fc6a54db0bb in slapi_ch_strdup ldap/servers/slapd/ch_malloc.c:196
    #2 0x7fc6a54ca26c in slapi_attr_syntax_normalize ldap/servers/slapd/attrsyntax.c:539
    #3 0x7fc6a5535bac in filter_normalize_ava ldap/servers/slapd/filter.c:1046
    #4 0x7fc6a5536a4f in filter_normalize_ext ldap/servers/slapd/filter.c:1144
    #5 0x7fc6a5536916 in filter_normalize_list ldap/servers/slapd/filter.c:1118
    #6 0x7fc6a5536ca0 in filter_normalize_ext ldap/servers/slapd/filter.c:1165
    #7 0x7fc6a5536916 in filter_normalize_list ldap/servers/slapd/filter.c:1118
    #8 0x7fc6a5536c64 in filter_normalize_ext ldap/servers/slapd/filter.c:1162
    #9 0x7fc6a5536d8b in slapi_filter_normalize ldap/servers/slapd/filter.c:1184
    #10 0x7fc69d54d083 in ldbm_back_search ldap/servers/slapd/back-ldbm/ldbm_search.c:959
    #11 0x7fc6a55c0eb1 in op_shared_search ldap/servers/slapd/opshared.c:821
    #12 0x7fc6a5607e6a in search_internal_callback_pb ldap/servers/slapd/plugin_internal_op.c:730
    #13 0x7fc6a5606ea7 in search_internal_pb ldap/servers/slapd/plugin_internal_op.c:594
    #14 0x7fc6a56067ba in slapi_search_internal_pb ldap/servers/slapd/plugin_internal_op.c:504
    #15 0x7fc6a56bf134 in vattr_check_thread ldap/servers/slapd/vattr.c:183
    #16 0x7fc6a4a48553 in _pt_root ../../.././nspr/pr/src/pthreads/ptthread.c:201


Indirect leak of 180 byte(s) in 10 object(s) allocated from:
    #0 0x7fc6a5d04627 in __interceptor_malloc (/lib64/libasan.so.6+0xb0627)
[...]

PR #5401, opened 2022-08-05
Issue 5397 - Fix various memory leaks

mreynolds389:
Description:

Fixed memory leaks in:

- Upgrade code when we check if an expected plugin is present, we didn't
  free the search results.
- Filter optimizer introduced sr_norm_filter_intent which dupped a filter
  but never freed it.
- Replication connections would leak the replication manager's
  credentials.

relates: https://github.com/389ds/389-ds-base/issues/5397


progier389 (approved):
LGTM

jchapma (approved):
LGTM.
I ran an asan instance with this PR, all filter optimisation related mem leaks fixed, all filter tests run with no issue.

PR #5405, opened 2022-08-08
Issue 3903 - keep alive update event starts too soon

mreynolds389:
Description: The keep alive update needs a little more time to start to
allow changelog, and other replication protocols to startup


tbordaz (approved):
LGTM

Issue #4588, opened 2021-02-02
Gost yescrypt may fail to build on some older versions of glibc.


Firstyear:
**Issue Description**
```
/home/william/development/389ds/ds/ldap/servers/plugins/pwdstorage/gost_yescrypt.c:23:12: warning: implicit declaration of function 'crypt_rn' is invalid in C99 [-Wimplicit-function-declaration]
    hash = crypt_rn(userpwd, dbpwd, &output, (int) sizeof(output));
           ^
/home/william/development/389ds/ds/ldap/servers/plugins/pwdstorage/gost_yescrypt.c:23:10: warning: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
    hash = crypt_rn(userpwd, dbpwd, &output, (int) sizeof(output));
         ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/william/development/389ds/ds/ldap/servers/plugins/pwdstorage/gost_yescrypt.c:41:15: error: use of undeclared identifier 'CRYPT_GENSALT_OUTPUT_SIZE'
    char salt[CRYPT_GENSALT_OUTPUT_SIZE];
              ^
/home/william/development/389ds/ds/ldap/servers/plugins/pwdstorage/gost_yescrypt.c:47:10: warning: implicit declaration of function 'crypt_gensalt_rn' is invalid in C99 [-Wimplicit-function-declaration]
    if (!crypt_gensalt_rn(prefix, 0, ((void*)0), 0, salt, (int) sizeof(salt))) {
         ^
/home/william/development/389ds/ds/ldap/servers/plugins/pwdstorage/gost_yescrypt.c:53:12: warning: implicit declaration of function 'crypt_rn' is invalid in C99 [-Wimplicit-function-declaration]
    hash = crypt_rn(pwd, salt, &output, (int) sizeof(output));
           ^
4 warnings and 1 error generated.
```

This appears to be due to missing defines in crypt.h on older glibc. 

**Package Version and Platform:**
SLE 15.2 glibc-2.26-13.51.1.x86_64

does not affect SUSE TW glibc-2.32-4.1.x86_64

Firstyear:
Ahhhh it's not crypt versions, it's that tw has xcrypt.

Firstyear:
```
[...]

PR #5410, opened 2022-08-10
Issue 4588 - Gost yescrypt may fail to build on some older versions o…

aadhikar:
…f glibc

Description: Fixed the test case for older version, also fixed another test
case which had wrong attribute name: telephonenumber

Relates: https://github.com/389ds/389-ds-base/issues/4588

Reviewed by: bsimonova (Thanks!)


bsimonova:
LGTM!

Issue #5407, opened 2022-08-09
sync_repl crashes if enabled while dynamic plugin is enabled


tbordaz:
**Issue Description**
Dynamic plugin (nsslapd-dynamic-plugins) allows to initialize and start a plugin while the server is running.  sync_repl expects that pre/post operation callbacks are called for each operation. For a MOD that enables the plugin, the preop is not called (plugin not init/started), while the postop is called as the plugin as been init/started.
Because of this the postop callback is missing the primary operation and crash

**Package Version and Platform:**
Since 1.3.9

**Steps to Reproduce**
run tests/tickets/ticket48013_test.py

**Expected results**
The server should not crash

tbordaz:
e867d1451..ff1d65d23  389-ds-base-2.2
a2f60b806..692c4cec6  389-ds-base-1.4.3

PR #5411, opened 2022-08-10
Issue 5407 - sync_repl crashes if enabled while dynamic plugin is ena…

tbordaz:
…bled

Bug description:
	When dynamic plugin is enabled, if a MOD enables sync_repl plugin
	then sync_repl init function registers the postop callback
        that will be called for the MOD itself while the preop
        has not been called.
        postop expects preop to be called and so primary operation
        to be set. When it is not set it crashes

Fix description:
	If the primary operation is not set, just return

relates: #5407

Reviewed by:


mreynolds389 (approved):
LGTM

Issue #5412, opened 2022-08-10
lib389 - backend creation always sets suffix to lowercase


mreynolds389:
**Issue Description**

We normalize the suffix by default when creating the backend configuration, but some customers want the root suffix to have a particular case.

mreynolds389:
66ccacf2c..581426cba  389-ds-base-2.1 -> 389-ds-base-2.1
4b34c0453..129c42a1e  389-ds-base-2.0 -> 389-ds-base-2.0
0f1f3d45f..f4ba4f810  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2117025

PR #5414, opened 2022-08-10
Issue 5412 - lib389 - do not set backend name to lowercase

mreynolds389:
Description:

There is no reason to set a new suffix to lowercase.  The server
will correctly handle the case, and some customers, especially
with migrations, want to have the base suffix a certain case.

relates: https://github.com/389ds/389-ds-base/issues/5412


droideck (approved):
LGTM!

PR #5417, opened 2022-08-15
Issue 3903 - fix repl keep alive event interval

mreynolds389:
Previously we passed the interval as seconds to the event queue, but it is supposed to be milliseconds.

Fixed a crash with repl logging and decoding extended op playload (referrals).

Also reworked a lot of the replication CI tests that were flaky.


droideck (approved):
Ack from me for the tests and lib389! C code looks good to me to the best of my replication code understanding. Still, I'd like @tbordaz to look at it too. :)

tbordaz (approved):
The change looks good to me.
Reading one of your change it raise a question. When the backend switch to read-only (for example during an admin task (for example import/export/backup/restore/demote...), should we stop the keep alive thread ?

mreynolds389:
> The change looks good to me. Reading one of your change it raise a question. When the backend switch to read-only (for example during an admin task (for example import/export/backup/restore/demote...), should we stop the keep alive thread ?

Hmmm, we would need to extend how the eventq works.  Add new flags to set which events should be paused and resumed during special "events" like going into read-only mode (or changelog export/import, etc).

Perhaps we should do that in a different ticket?

tbordaz:
> > The change looks good to me. Reading one of your change it raise a question. When the backend switch to read-only (for example during an admin task (for example import/export/backup/restore/demote...), should we stop the keep alive thread ?
> 
> Hmmm, we would need to extend how the eventq works. Add new flags to set which events should be paused and resumed during special "events" like going into read-only mode (or changelog export/import, etc).
> 
> Perhaps we should do that in a different ticket?

Absolutely this PR is fine to me and no other concern is raised I agree to push it upstream.

I think we may also use statechange plugin to capture backend moving to readonly state.

PR #5419, opened 2022-08-16
Issue 5012 - Migrate pcre to pcre2

### LINKED ISSUE #5012: Migrate to PCRE2
389 is still using the long-obsolete PCRE version, it should migrate to PCRE2. Debian will soon remove the old PCRE.

SSSD ported to PCRE2 three years ago, and dropped support for the old PCRE in 2.6.0.

mreynolds389:
Description:  PCRE is deprecated and is being removed, need to use the
new PCRE2 lkbrary

fixes: https://github.com/389ds/389-ds-base/issues/5012

### DIFF
--- a/ldap/servers/slapd/regex.c
+++ b/ldap/servers/slapd/regex.c
@@ -1,5 +1,5 @@
 /** BEGIN COPYRIGHT BLOCK
- * Copyright (C) 2009 Red Hat, Inc.
+ * Copyright (C) 2022 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -10,20 +10,18 @@
 #include <config.h>
 #endif
 
-/* number of elements in the output vector */
-#define OVECCOUNT 30 /* should be a multiple of 3; store up to \9 */
-
 #include "slap.h"
 #include "slapi-plugin.h"
+#include <pcre2.h>
 
-/* Perl Compatible Regular Expression */
-#include <pcre.h>
+#define OVEC_MATCH_LIMIT 30 /* should be a multiple of 3; store up to \9 */
+#define CPRE_ERR_MSG_SIZE 120
 
 struct slapi_regex_handle
 {
-    pcre *re_pcre;    /* contains the compiled pattern */
-    int *re_ovector;  /* output vector */
-    int re_oveccount; /* count of the elements in output vector */
+    pcre2_code *re_pcre;
+    pcre2_match_data *match_data; /* Contains the output vector */
+    pcre2_match_context *mcontext; /* Stores the max element limit */
 };
 
 /**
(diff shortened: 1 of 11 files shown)


ldap/servers/slapd/regex.c
```diff
@@ -123,33 +125,35 @@ slapi_re_exec(Slapi_Regex *re_handle, const char *subject, time_t time_up)
...
-                   subject,                  /* the subject string */
-                   strlen(subject),          /* the length of the subject */
-                   0,                        /* start at offset 0 in the subject */
-                   0,                        /* default options */
-                   re_handle->re_ovector,    /* output vector for substring info */
-                   re_handle->re_oveccount); /* number of elems in the ovector */
+    rc = pcre2_match(re_handle->re_pcre,
+                     (PCRE2_SPTR)subject,
+                     strlen(subject),
+                     0,
```
Firstyear: The comments for what each argument was for was nice to have before, could be good to re-add them.

m4/libpcre.m4
```diff
@@ -0,0 +1,30 @@
...
+    ],
+    [
+        found_libpcre=no
+    ]
+)
+
+# SSS_AC_EXPAND_LIB_DIR()
+AS_IF([test x"$found_libpcre" != xyes],
+    [AC_CHECK_HEADERS([pcre2.h],
+        [AC_CHECK_LIB([libpcre2-8],
```
Firstyear: You should use pkgconfig here rather than checking for a specific version IMO. In pcre2-devel there should be: `/usr/lib64/pkgconfig/libpcre2-posix.pc`

Firstyear (changes requested):

ldap/servers/slapd/regex.c
```diff
@@ -36,18 +34,20 @@ struct slapi_regex_handle
...
+    PCRE2_UCHAR errormsg[CPRE_ERR_MSG_SIZE];
 
-    re = pcre_compile(pat, 0, &myerror, &erroffset, NULL);
-    if (error) {
-        *error = myerror;
-    }
-    if (re) {
+    re = pcre2_compile((PCRE2_SPTR)pat, strlen(pat), 0,
+                       &myerror, &erroffset, NULL);
+    if (re == NULL) {
```
tbordaz, line 47: In case the caller did not initialize 'error = NULL', it would be safer to systematically do '*error = NULL'.

tbordaz (approved):
Except a minor comment, the code looks good to me. The more pcre to pcre2 was not that immediate.. nice job.

m4/libpcre.m4
mreynolds389: I'm not familiar with what you asking in regards to pkgconfig (looking into it though).  I just used the same approach that SSSD used when they migrated to pcre2.  So do you meaning adding this to the specfile and removing the "AS_IF" section from the m4 file?  

    BuildRequires: pkgconfig(libpcre2-8)

The only concern I have here (if that's what you meant) is that no one uses our specifle but us

ldap/servers/slapd/regex.c
mreynolds389: > The comments for what each argument was for was nice to have before, could be good to re-add them.

Fixed

ldap/servers/slapd/regex.c
mreynolds389: > In case the caller did not initialize 'error = NULL', it would be safer to systematically do '*error = NULL'.

I reworked it, let me know if you are okay with it.

tbordaz (approved):
LGTM

m4/libpcre.m4
mreynolds389: Fixed

mreynolds389:
@Firstyear - I changed the packaging config around, so you have to approve the "change request" if it's acceptable.  

FYI, pcre2-8, the "-8" is not a rpm version, but the "8-bit" flavor.  See https://www.pcre.org/readme.txt section "The PCRE2 APIs"

Firstyear (approved):

Firstyear:
Sorry for the really late approval, I injured myself last week :(

Issue #5418, opened 2022-08-16
Sync_repl may crash while managing invalid cookie


tbordaz:
**Issue Description**
If a ldapclient sends specific cookie without meta character '#' then sync_repl may crash

**Package Version and Platform:**
since 1.3.9

**Steps to Reproduce**
testcase provided in patch

**Expected results**
It should not crash

Firstyear:
@tbordaz @mreynolds389 this will probably need to be applied to 1.4 as well as 2.0?

tbordaz:
@Firstyear thanks for your review. You are correct it applies to several branches, I set milestone to 1.3.10

tbordaz:
3cda974c1..513a763b5 main
8837f3137..874d2de7e  389-ds-base-2.1
85965d1d0..19632a4b8  389-ds-base-2.0
7a611298d..be599560c  389-ds-base-1.4.4
eebd91b04..b45bdc53c  389-ds-base-1.4.3
5594c3dac..bd566957f  389-ds-base-1.3.10

mreynolds389:
This was not properly cherry-picked to 1.4.3.  The actual fix was removed from the commit.  This is going to impact a lot of releases :-(

tbordaz:
I am really sorry of that. I do not know how I could have wrongly cherry-pick this simple fix. It looks to me that upstream, only 1.4.3 was impacted by this error (fixed now with ac351252f..f5cafd3b5 389-ds-base-1.4.3).

PR #5420, opened 2022-08-17
Issue 5418 - Sync_repl may crash while managing invalid cookie

tbordaz:
Bug description:
	If the servers receives an invalid cookie without separator '#',
	it parses it into an empty cookie (Sync_Cookie) instead of a NULL
	cookie (failure).
	Later it sigsegv when using the empty cookie.

Fix description:
	If the parsing fails return NULL

relates: #5418

Reviewed by: Viktor Ashirov, Mark Reynolds


Firstyear (approved):

droideck (approved):
Looks great:)
Just checked, and the test perfectly fails without the fix.

Issue #5356, opened 2022-06-24
pbkdf2 hardcoded parameters not secure anymore


furznir:
This is a security issue. It leads to weak password storage compared to current standards and computation times.
[This commit](https://github.com/389ds/389-ds-base/commit/ee25b881225f788fc35d8ebdfa126e797960760c) in 2017 reduced the security parameters for pbkdf2 algorithm. In 2022, it's time to follow the [code comment's](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c#L39-L42) recommendation and raise the default parameters.

These parameters are not considered secure anymore, [OWASP recommends](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2) 310.000 iterations for the algorithm (in HMAC case).

The current implementation targets [2 ms](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c#L53) computation time for one hash, so attackers won't be able to easily brute-force the password hash. The number of iteratons is dinamically set: If the benchmark functions for this target yields more time on the current machine, the number of iterations will be reduced. The hard-minimum is [2048](https://github.com/389ds/389-ds-base/blob/master/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c#L59) iterations.

I ran the benchmark myself on a laptop with 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz CPU. On this laptop 1500 iterations yield the target 2 ms computation time, so the code set the 2048 minimum, I saw this in the logs: "Based on CPU performance, chose 2048 rounds", and the base64 decoded user password hash in 389ds also contained the 2048 iteration parameter.

With some calculation, we can see that with a password of 12 characters (A-Z a-z 0-9 and 10 possible specials) with 2048 iterations can be broken on this laptop (8 threads) with 50% chance in 726 days, with 64 threads in 91 days.
Calculation time for 2048 iterations on my computer with 1 thread: 0,003 sec
Number of all possible input passwords: (26+26+10+10)^12
Trials needed for 50% chance of password break due to [birthday paradox](https://mathworld.wolfram.com/BirthdayAttack.html): 1,2×√((26+26+10+10)^12) = 167176883404,8
Needed time in seconds with 8 cores: 167176883404,8*0,003/8 = 62691331,2768 sec
[...]

PR #5431, opened 2022-08-23
Issue 5356 - Make Rust non-optional and update default password storage

mreynolds389:
Make Rust non-optional and update default password storage scheme

Description:

We need a stronger default storage scheme which comes from our Rust
plugins, but to do this  Rust needs to be non-optional.  It will be
a requirement moving forward.

relates: https://github.com/389ds/389-ds-base/issues/5356

Reviewed by: ?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -56,9 +56,6 @@ SYSTEMTAP_DEFINES = @systemtap_defs@
 NSPR_INCLUDES = $(NSPR_CFLAGS)
 
 # Rust inclusions.
-if RUST_ENABLE
-# Rust enabled
-RUST_ON = 1
 CARGO_FLAGS = @cargo_defs@
 
 if CLANG_ENABLE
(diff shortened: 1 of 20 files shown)


Firstyear (approved):
Huge and enthusiastic yes from me!

droideck:
I see a few new failures. I wonder if our CI went crazy or if something is up with the PR...
Rerunning.

vashirov:
healthcheck and [tests](https://github.com/389ds/389-ds-base/blob/main/dirsrvtests/tests/suites/healthcheck/health_security_test.py#L71
) for it complain that our new default storage scheme is weak. I think lint.py should be updated too:
https://github.com/389ds/389-ds-base/blob/dea9f6a7cd6d1c4029705559c69c273584a53561/src/lib389/lib389/lint.py#L97-L98



`pwp_test.py` expects a different default password storage scheme:
https://github.com/389ds/389-ds-base/blob/dea9f6a7cd6d1c4029705559c69c273584a53561/dirsrvtests/tests/suites/password/pwp_test.py#L30

and [`dscreate_test.py::test_setup_ds_as_non_root`](https://github.com/389ds/389-ds-base/blob/main/dirsrvtests/tests/suites/setup_ds/dscreate_test.py#L182) fails with
```
[24/Aug/2022:04:20:50.731427271 +0000] - ERR - slapd_bootstrap_config - The default password storage scheme could not be read or was not found in the file /tmp/tmphu0fk5_5/root/etc/dirsrv/slapd-standalone/dse.ldif. It is mandatory.

```

mreynolds389:
I don't see any regressions in CI tests with my last commit.  Merging...

PR #5432, opened 2022-08-24
Issue 5356 - For RUST build update the default password storage scheme

mreynolds389:
Description:

If Rust is enabled use PBKDF2-SHA512 as the default password storage
scheme

relates: 389ds#5356

### DIFF
--- a/src/lib389/lib389/password_plugins.py
+++ b/src/lib389/lib389/password_plugins.py
@@ -31,7 +31,7 @@ class PasswordPlugin(Plugin):
         self._protected = True
 
 class PBKDF2Plugin(PasswordPlugin):
-    def __init__(self, instance, dn="cn=PBKDF2_SHA256,cn=Password Storage Schemes,cn=plugins,cn=config"):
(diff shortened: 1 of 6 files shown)


src/lib389/lib389/password_plugins.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, instance, dn=None):
         self._protected = True
 
 class PBKDF2Plugin(PasswordPlugin):
-    def __init__(self, instance, dn="cn=PBKDF2_SHA256,cn=Password Storage Schemes,cn=plugins,cn=config"):
+    def __init__(self, instance, dn="cn=PBKDF2-SHA256,cn=Password Storage Schemes,cn=plugins,cn=config"):
```
droideck: Should we change it to PBKDF2-SHA512 here too?

droideck (approved):
Otherwise, looks good!
I've built and run a few tests on my machine and everything's working.

src/lib389/lib389/password_plugins.py
mreynolds389: I'm not sure why @Firstyear chose SHA256 over SHA512, as this is used by his openldap migration work.  Well I'll change it and see if he yells about it :-)

PR #5439, opened 2022-09-06
Issue 5158: entryuuid fixup tasks fails in replicated topology

### LINKED ISSUE #5158: entryuuid fixup tasks fails in replicated topology
**Issue Description**
During a entryuuid fixup task, if an entry is updated the updates fails because entryUUID attribute is NO-USER-MODIFICATION

since 1.4.3

**Steps to Reproduce**
create  entries without entryuuid, run fixup task

**Expected results**
Should fix all entries

progier389:
Problem: 
 entryuuid plugin fixup task fails in replicated topology because o CSN is generated and the replication plugin aborts the operation.
 This happen because the operation flags are set to 0x400000  which is OP_FLAG_TOMBSTONE_FIXUP
 (a flags that should only be set to handle  replication plugin internal operations)

Solution: 
   Do not any flags 

Issue: [5158](https://github.com/389ds/389-ds-base/pull/5158)

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/entryuuid/replicated_test.py
+++ b/dirsrvtests/tests/suites/entryuuid/replicated_test.py
@@ -15,6 +15,8 @@ from lib389.paths import Paths
 from lib389.utils import ds_is_older
 from lib389._constants import *
 from lib389.replica import ReplicationManager
+from lib389.plugins import EntryUUIDPlugin
+from lib389.tasks import EntryUUIDFixupTask
 
 default_paths = Paths()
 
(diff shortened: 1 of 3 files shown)


Firstyear:
I think there was a reason for this flag to be added in the first place, because it resolved an issue, could we git blame that line and check specifically?

progier389:
it puzzled me too, so I have looked at it before creating the PR -;)
It is Issue 4326 and PR 4328 (i.e when modify.rs file was added) but there is no specific mention about this flag 
The only hint that I have is the way the flag was named (bypass referral ?) but even that does not make sense 
  in an internal modify operation, so I just do not know.
But I am absolutely sure is that this flag value cannot be used here. (Or in any plugins except the replication one)
So my feeling is that unless you remember why you used this flag we should just remove it, then wait until someone discover a problem that we could fix it properly.

Firstyear:
Perhaps it was during the review during that. I do seem to recall there was a problem that required that flag to exist, and there was no simple work around, so you can remove the flag but perhaps comment it and say that it used to be present?

Firstyear:
Okay fair enough :) I'm happy with that :)

progier389:
Thanks, could you then formally approve the review. please ?

Firstyear (approved):
Sorry about that, totally slipped my mind :)

Issue #5440, opened 2022-09-08
memberof is slow on update/fixup if there are several 'groupattr'


tbordaz:
**Issue Description**
Memberof can be configured with several membership attribute (member, uniquemember,..)
When fixing (create memberof values) an entry, memberof plugin searches the groups owning that entry using a filter that contains all the membership attributes. The drawbacks are

- it mixes the relation, so G2 --member--> G1 -- uniquemember--> U1, then U1.memberof=[G1,G2]
- it forces the internal search to reevaluate the filter (|(member=U1)(uniquemember=U2) although both attributes are index

**Package Version and Platform:**
all version

**Steps to Reproduce**
To be done

**Expected results**
Fixup is very slow

tbordaz:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=2123525

tbordaz:
7e701e2c3..7283a57b7 main
0ebbb26f8..78ee3a5a4  389-ds-base-2.2
8fdbe5947..b5e5afb7f  389-ds-base-2.1
d15a0a7f1..ef2ec82e2  389-ds-base-2.0
136fc8448..ce0bda67e  389-ds-base-1.4.4
620f96fb4..9ba2cd33d  389-ds-base-1.4.3

still fighting for 1.3.10...

tbordaz:
d08a0e414..813c8cd6f  389-ds-base-1.3.10

droideck:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2170218

PR #5441, opened 2022-09-08
Issue 5440 - memberof is slow on update/fixup if there are several 'g…

tbordaz:
…roupattr'

Bug description:
	Tentative fix to improve memberof fixup
	When there are several groupattr in memberof config
        To fixup an entry we are doing an internal search
	"(|(attr_1=e1)(attr_2=e1)...(attr_n=e1))"
	This is not valid regarding membership relation and in
        addition it prevents the server to shortcut filter evaluation.

Fix description:
	To fixup an entry iterate several internal searches
	"(attr_1=e1)" , then "(attr_2=e1)", then "(attr_n=e1)"

relates: #5440

Reviewed by:

Platforms tested:

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -744,7 +744,7 @@ memberof_call_foreach_dn(Slapi_PBlock *pb __attribute__((unused)), Slapi_DN *sdn
     } else {
         escaped_filter_val = (char *)slapi_sdn_get_dn(sdn);
     }
-
+#if 0
     if (num_types > 1) {
         int bytes_out = 0;
         int filter_str_len = types_name_len + (num_types * (3 + dn_len)) + 4;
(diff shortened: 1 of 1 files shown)


progier389:
Just wondering whether the fact that an entry may get processed several time (for example if it match (attr_1=e1)" and "(attr_2=e1)" could be a problem

progier389:
Another point (even if understand that this PR goal  is to a build an hotfix for a customer that needs it urgently. ) 
IMHO the filter evaluation bypass logic is wrong
  We should be able to bypass the check in most OR cases and some AND cases. 
   as far as the candidate build is accurate (in other words: no allids no NOT no shortcut because candidate list is small) and no need to evaluate the acls.

tbordaz:
Good point. 
memberof_call_foreach_dn is a generic function called in different conditions. The problematic one is when the callback is memberof_get_groups_callback. In seach case the function/callback build a set of groups, if the group has already been found in the set, it is not added again. So this change should not be problem for  memberof_get_groups_callback.

Now I have not look at the others cases when memberof_call_foreach_dn

tbordaz:
> Good point. memberof_call_foreach_dn is a generic function called in different conditions. The problematic one is when the callback is memberof_get_groups_callback. In seach case the function/callback build a set of groups, if the group has already been found in the set, it is not added again. So this change should not be problem for memberof_get_groups_callback.
> 
> Now I have not look at the others cases when memberof_call_foreach_dn

I absolutely agree. It looks very strict to skip the shortcut just because of a OR. Now changing the filter code is sensitive and need some more time.

tbordaz:
> Another point (even if understand that this PR goal is to a build an hotfix for a customer that needs it urgently. ) IMHO the filter evaluation bypass logic is wrong We should be able to bypass the check in most OR cases and some AND cases. as far as the candidate build is accurate (in other words: no allids no NOT no shortcut because candidate list is small) and no need to evaluate the acls.

I absolutely agree. It looks very strict to skip the shortcut just because of a OR. Now changing the filter code is sensitive and need some more time.

progier389:
My feeling too (clearly, we should not  change the filter for the urgent hotfix, but we should think about it for the final fix or even another issue ...)

tbordaz:
Closing this PR to use https://github.com/389ds/389-ds-base/pull/5455

Issue #5413, opened 2022-08-10
Only allow one fixup task at a time


mreynolds389:
**Issue Description**

We keep seeing issues where customers incorrectly create 100's or 1000's of memberOf fixup tasks which leads to overall terrible server performance.  We should only allow one fixup task per plugin.

mreynolds389:
198d921ca..fc205ec7c  389-ds-base-2.2 -> 389-ds-base-2.2
814f8b583..d1b16781e  389-ds-base-2.1 -> 389-ds-base-2.1
6083fd136..4401b66ca  389-ds-base-2.0 -> 389-ds-base-2.0
e665ae06a..5dd7043b3  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

tbordaz:
The detection of concurrent memberof task should be done consider base/filter.
IPA is doing valid parallel fixup tasks

<pre>
dn: cn=Update PBAC memberOf $TIME, cn=memberof task, cn=tasks, cn=config
objectClass: top
objectClass: extensibleObject
cn: IPA PBAC memberOf $TIME
basedn: cn=privileges,cn=pbac,$SUFFIX
filter: (objectclass=*)
ttl: 10

dn: cn=Update Role memberOf $TIME, cn=memberof task, cn=tasks, cn=config
objectClass: top
objectClass: extensibleObject
cn: Update Role memberOf $TIME
basedn: cn=roles,cn=accounts,$SUFFIX
filter: (objectclass=*)
ttl: 10
</pre>

Firstyear:
This may not be true @tbordaz because IPA has nested groups. so you could have cn=some_priv,cn=privileges,cn=pbac,$SUFFIX which includes a member cn=myrole,cn=roles,cn=accounts,$SUFFIX. Since these are nested, this would trigger the members of cn=myrole to be adjusted too. Since there is *also* the second fixup in update role, it could conflict on the changes and cause loss of member of records.

I think that nested groups pretty much means you have to consider only one memberof fixup at a time.

tbordaz:
I absolutely agree with your point. With nested groups it is unpredictable if parallel fixup will or not steps on each other. In addition I would say that parallel fixup will increase DB pressure (even if accessing different entries) with a higher risk of retries so from performance pov, there is a possibility that parallel fixups will take more time that serialized one.

Now IMHO it should be in the hand of the administrators/applications to decide/experiment if parallel fixups bring some benefits.

[...]

PR #5442, opened 2022-09-09
Issue 5413 - Allow only one MemberOf fixup task at a time

mreynolds389:
Description:  only allow one fixup task to run at a time, and improve the task logging

relates: https://github.com/389ds/389-ds-base/issues/5413

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -52,7 +52,12 @@ static Slapi_DN* _pluginDN = NULL;
 MemberOfConfig *qsortConfig = 0;
 static int usetxn = 0;
 static int premodfn = 0;
-
+static PRBool fixup_running = PR_FALSE;
+static PRLock *fixup_lock = NULL;
+static int32_t fixup_progress_count = 0;
+static int64_t fixup_progress_elapsed = 0;
+static int64_t fixup_start_time = 0;
+#define FIXUP_PROGRESS_LIMIT 1000
 
 typedef struct _memberofstringll
 {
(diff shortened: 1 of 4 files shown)


progier389 (approved):
LGTM

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -421,6 +432,8 @@ memberof_postop_close(Slapi_PBlock *pb __attribute__((unused)))
     slapi_sdn_free(&_pluginDN);
     slapi_destroy_rwlock(config_rwlock);
     config_rwlock = NULL;
+    PR_DestroyLock(fixup_lock);
```
tbordaz, line 438: It is initialized in init fct, I would prefer not to destroy it. Either init it in start (memberof_postop_start) and destroy it in close or simply init it in memberof_postop_init and do not bother to free it.

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -2820,9 +2833,16 @@ memberof_fixup_task_thread(void *arg)
     if (!task) {
         return; /* no task */
     }
+
+    PR_Lock(fixup_lock);
+    fixup_running = PR_TRUE;
```
tbordaz, line 2841: Only 'fixup_running' looks protected by the lock. I would unlock just after the setting of fixup_running to make it clear that the others fixup_* are not protected by the lock.

tbordaz (approved):
The code looks good to me. Only minor comments

ldap/servers/plugins/memberof/memberof.c
mreynolds389: only "fixup_running" needs to be protected

ldap/servers/plugins/memberof/memberof.c
mreynolds389: I'll move it to the start function, that way dynamic plugins will work correctly and not break anything.

Issue #5443, opened 2022-09-10
UI - save buttons should be disabled while saving


mreynolds389:
**Issue Description**

You can click the save button over and over while its spinning and it will trigger the same operation over and over.  It should be disabled while it's "working/spinning".

mreynolds389:
fc205ec7c..bb6e19927  389-ds-base-2.2 -> 389-ds-base-2.2
d1b16781e..029f9056c  389-ds-base-2.1 -> 389-ds-base-2.1
4401b66ca..acb3bda63  389-ds-base-2.0 -> 389-ds-base-2.0
5dd7043b3..aadb3556d  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
Fix regression with replication changelog

990ffe26e..f69ba1bfa  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5444, opened 2022-09-10
Issue 5443 - UI - disable save button while saving

mreynolds389:
Description:

You can click the save button over and over while its spinning and it will trigger the same operation over and over. It should be disabled while it's "working/spinning".

relates: https://github.com/389ds/389-ds-base/issues/5443

Reviewed by: ?


droideck (approved):
LGTM!

Issue #5453, opened 2022-09-19
When changing the Root DN also update the LDAPI mapping


mreynolds389:
**Issue Description**

If you change "cn=directory manager" to something else the UI will crash and LDAPI breaks.  We should update the LDAPi settings to match the new root DN

To reproduce:

    # dsconf -D "cn=Directory Manager" ldap://localhost config replace nsslapd-rootdn="cn=newdirmgr"

Workaround:

    # dsconf -D "cn=newdirmgr" ldap://localhost  config replace nsslapd-ldapimaprootdn="cn=newdirmgr"

We should set this by default, and fix UI crash.

PR #5454, opened 2022-09-20
Issue 5453 - UI/CLI - Changing Root DN breaks UI

mreynolds389:
Bug Description:

When you change the root DN it breaks LDAPI connectivity because the LDAPI mapping for the root DN is out of sync.  This also causes the UI to crash when accessing the server.

Fix Description:

When you change the root dn via dsconf it will automatically update the LDAPI configuration.  The UI will also gracefully handle the situation when things are out sync.

relates:  https://github.com/389ds/389-ds-base/issues/5453

### DIFF
--- a/src/cockpit/389-console/src/ds.jsx
+++ b/src/cockpit/389-console/src/ds.jsx
@@ -48,14 +48,14 @@ const staticStates = {
     ),
     noInsts: (
         <TextContent>
-            <Text className="ds-margin-top-xlg" component={TextVariants.h2}>
+            <Text className="ds-margin-top-xlg ds-indent-md" component={TextVariants.h2}>
                 There are no Directory Server instances to manage
             </Text>
         </TextContent>
     ),
     notRunning: (
         <TextContent>
-            <Text className="ds-margin-top-xlg" component={TextVariants.h2}>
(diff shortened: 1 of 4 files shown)


src/cockpit/389-console/src/ds.jsx
```diff
@@ -48,26 +48,34 @@ const staticStates = {
...
+            <Text className="ds-margin-top-xlg ds-indent-md" component={TextVariants.h2}>
                 This server instance is running, but we can not connect to it. Check LDAPI is properly
                 configured on this instance.
             </Text>
         </TextContent>
+    ),
+    ldapiIssue: (
+        <TextContent>
+            <Text className="ds-margin-top-xlg ds-indent-md" component={TextVariants.h2}>
+                Problem accessing required server configuration.  Check LDAPI is properly
```
droideck: Extra space before 'Check' :)

droideck (approved):
Besides the very minor issue, looks good!
Ack

src/cockpit/389-console/src/ds.jsx
mreynolds389: > Extra space before 'Check' :)

I usually add two spaces after a period, but the surrounded code uses one space, and html truncates it to one space anyway :-)  Fixed...

PR #5456, opened 2022-09-22
Issue 5032 - OpenLDAP is not shipped with non-threaded version of libldap

droideck:
Bug Description: 389-ds-base is built with libldap_r shared library. The non-threaded version of libldap is fully depricated since Fedora 34, libldap_r supplied as symlinks. OpenLDAP 2.6 won't have libldap_r at all.

Fix Description: Use lldap instead of lldap_r in Makefile. Add "--with-libldap-r" parameter to configure.ac.
If it's not set, we check the OpenLDAP version on the build system and use the appropriate shared library.

Reviewed by: ?


progier389 (approved):
LGTM

PR #5463, opened 2022-09-28
Issue 4324 - Revert recursive pthread mutex change

droideck:
Bug description:
	The entry cache is protected with recursive mutex. Currently it is
	implemented using pthread recursive mutex. Very rarely, we may
	encounter 'Retry count error' (we weren't able to reproduce it reliably).

Fix description:
	As per investigation and team discussions, we decided to go with the 'revert' option,
	as it reliably helped the user who had the issue on their environment.

	Changing the c_mutex from pthread recursive mutex back to PR_Monitor.

Related: https://github.com/389ds/389-ds-base/issues/4324

Reviewed by: ?


progier389 (approved):
LGTM

PR #5468, opened 2022-09-30
Issue 2669 - RFE - Commit draft rfcs for StartTLS deprecation

### LINKED ISSUE #2669: Deprecation of StartTLS, and usage of LDAPS instead.
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/49610
- Created at 2018-03-19 03:39:01 by [firstyear](https://pagure.io/user/firstyear) (@Firstyear)

---

#### Issue Description
It's been well known for some time to us as a community that StartTLS has a number of flaws. The usage of TLS via LDAPS is a far better option.

As a result, to correctly advertise and move this forward in our community, we should create an rfc draft that provides a reference, similar to the deprecation issued by smtp/imap communities.

We will keep the source of the draft rfc in our repo, but otherwise this is purely a formality to help shift our consumers and community attitudes to secure communication protocols.

Firstyear:
Bug Description: StartTLS should be considered deprecated.

Fix Description: This commit adds rfc source pertaining to this.

fixes: https://github.com/389ds/389-ds-base/issues/2669

Review by: ???


mreynolds389:
Not to be the bad guy here, but Microsoft and Openldap have declared LDAPS deprecated.  If you want to deprecate startTLS then what is left to use that is supported across the industry?

Firstyear:
The idea here is to revert / counter that deprecation. LDAPS should no longer be deprecated, and it's the other way around. Remember, MS probably made that kind of statement when they were using GSSAPI wrap for all their transport security and couldn't run GSSAPI over TLS, but now they are able to do so. 

Even MS now enforce LDAPS only for some services ( https://learn.microsoft.com/en-us/azure/active-directory-domain-services/tutorial-configure-ldaps ) and other protocols have already had similar moments ( imap/smtp https://nostarttls.secvuln.info/ )

Saying that I think making that clearer in the text that this is "reverting" that deprecation is probably a good idea? 

Regardless, even that all considered this will probably always only be a draft - the ldap ietf wg has fallen apart so there may be no way to advance this to a true rfc anyway. But we'll see.

Firstyear:
PS: You're not the bad guy :)

mreynolds389:
> The idea here is to revert / counter that deprecation. LDAPS should no longer be deprecated, and it's the other way around. Remember, MS probably made that kind of statement when they were using GSSAPI wrap for all their transport security and couldn't run GSSAPI over TLS, but now they are able to do so.
> 
> Even MS now enforce LDAPS only for some services ( https://learn.microsoft.com/en-us/azure/active-directory-domain-services/tutorial-configure-ldaps ) and other protocols have already had similar moments ( imap/smtp https://nostarttls.secvuln.info/ )
> 
> Saying that I think making that clearer in the text that this is "reverting" that deprecation is probably a good idea?
> 
> Regardless, even that all considered this will probably always only be a draft - the ldap ietf wg has fallen apart so there may be no way to advance this to a true rfc anyway. But we'll see.

There is definitely a consensus in the field that LDAPS should not be used.  They continually quote MS and Openldap docs (perhaps those docs are outdated).  I hear this from a lot of different groups (not just Freeipa).  Regardless I'm fine filing this RFC draft, it's just going against the currently accepted practice so I think it's going to be hard for people to switch their perspective on this.  That doesn't mean we shouldn't try.

antbob:
I was just pointed here by a colleague of mine and since I have never heard of this draft before I'm just curious if anyone can clarify these points:

a) was this draft published anywhere in the standards track and if so where ? I cant seem to find it anywhere apart from this patch.

b) while the points presented in the draft appear to be valid why get rid of the startTLS op instead of improving it ? it seems to me that at least some points made in the draft can be addressed by imposing stricter requirements on startTLS while others would always ultimately be up to clients following stricter security requirements/guidelines.

Note that LDAPS/636 was never part of any standard but startTLS is. One of the reasons for startTLS is to make it easier for clients to discover and talk to LDAP service. Historically the problems with LDAP/LDAPS were that there was no way for a client to tell if LDAP service listening on some port had already security layer installed or not which lead to confusion and weird errors (eg client talking LDAP over LDAPS and vise-versa) most clients couldnt handle properly. startTLS does address that nicely and it still allows for plain LDAP in cases where it is preferable eg for performance reasons.

Perhaps the next iteration/evolution of the startTLS should follow HTTP Upgrade-Insecure-Requests way ? Or something along those lines ?

Firstyear:
@antbob I intend to publish it as a draft from these sources once the other team members reviewed it here.

Secondly, there is no way to improve startTLS as an operation since it is fundamentally flawed as a cryptographic concept to start plaintext then upgrade - time and time again, it has been shown that to properly secure communications you must start with encryption, rather than layer it in later.

I think that the performance issues that existed when startTLS was recommended in the 2000s no longer exist either. Today startTLS is actually a negative on performance due to the extra round trips before communication can begin. 

So IMO there is no way to fix StartTLS, and this has already been highlighted in other protocols (IMAP, SMTP, POP).

Firstyear:
> There is definitely a consensus in the field that LDAPS should not be used. They continually quote MS and Openldap docs (perhaps those docs are outdated). I hear this from a lot of different groups (not just Freeipa). Regardless I'm fine filing this RFC draft, it's just going against the currently accepted practice so I think it's going to be hard for people to switch their perspective on this. That doesn't mean we shouldn't try.

Those docs were from the 2000's though, and they kinda just .. stuck because no serious examinations of the issues occured until the last few years starting with IMAP/SMTP. I probably did the first examination of StartTLS in LDAP and the issues with human interaction failure and referral/error injection. 

Where we see that any serious modern protocol starts with encryption/cryptography from the start, and doesn't try to mix plaintext and upgrade. We can even see this with HTTP that HTTPS is now "the norm", and most cases is faster than HTTP infact. As well, when we look at MS and their Azure AD offerings, they are LDAPS only. 

FreeIPA if I recall only have the plaintext/startTLS thing because they never fixed MIT KRB to do GSSAPI over TLS (where MS-AD did fix this).

I also need to double check, but there may be a priv-esc in SSSD related to this, but I want to confirm it first and report it properly rather than dropping 0-days on github 🤣

Firstyear:
@mreynolds389 If you're okay with the sources being here, once you approve I'll submit it to the draft track.

antbob:
@Firstyear so let me just counter your points BUT keep in mind that i reckon the goal you have is undoubtedly a good one its just how you are going about it is questionable and unlikely to yield any industry wide consensus in terms of standards and adoption assuming this is probably what you want as the end result.

> Secondly, there is no way to improve startTLS as an operation since it is fundamentally flawed as a cryptographic concept to start plaintext then upgrade - time and time again, it has been shown that to properly secure communications you must start with encryption, rather than layer it in later.
> 

LDAP server implementations already have some form of minimum connection security requirements eg 389ds has minssf. Enforcing this alone and actually making it part of the standard would solve all the issues highlighted in your draft:

4.1.  Error Handling or Injection
4.2.  Referral Injection
4.3.  Incorrect Client Handling
4.4.  Configuration Error

because the only operations allowed would be startTLS or SASL bind (as long as the later install sufficient security layer/s). So this way you get what you want as a STEP towards the bright future of LDAPS only. On the other hand the service is easily discoverable (clients know its LDAP) and can provide unencrypted anonymous access and server information eg rootDSE contexts.

> I think that the performance issues that existed when startTLS was recommended in the 2000s no longer exist either. Today startTLS is actually a negative on performance due to the extra round trips before communication can begin.
> 

That is certainly not true. Even with modern CPU support for additional instructions to speed up crypto related functions and various optimizations made any additional layer on top of existing one would cause a slow down, by simple arithmetic. Even if you just XOR it is still O(N)+M or whatever. In some cases the difference will be peanuts but at a high volume and specific access patterns it can be huge and this has 2 main cost factors: TLS session establishment (many short lived connections) + encrypt/decrypt of already established sessions.

Quick google https://www.f5.com/company/blog/stop-just-stop-https-is-not-faster-than-http explains how that myth came about in HTTP world.

> So IMO there is no way to fix StartTLS, and this has already been highlighted in other protocols (IMAP, SMTP, POP).

Those are very different and not a good comparison to LDAP. LDAP has anonymous access and things like root DSE where LDAP serves information that has no need to be encrypted and in fact encrypting things like that is a hindrance and not an actual advantage. The email protocols you mention serve no such information ie all information they serve or exchange has to be protected. HTTP would be closer to LDAP in that sense eg accessing a public weather service over unencrypted connection as one example.

You propose a deprecation of startTLS without LDAPS being standardized in any way. The way it should be done is via LDAP standard actually declaring LDAP protocol over secure connection/transport a MUST ie evolving RFC4511 section 5.2 to require what is known as LDAPS. Then startTLS can be deprecated/obsolete. Without changing the LDAP standard itself this probably won't go anywhere.

Firstyear:
> @Firstyear so let me just counter your points BUT keep in mind that i reckon the goal you have is undoubtedly a good one its just how you are going about it is questionable and unlikely to yield any industry wide consensus in terms of standards and adoption assuming this is probably what you want as the end result.

What would be a better way to go about it then? 

> 
> > Secondly, there is no way to improve startTLS as an operation since it is fundamentally flawed as a cryptographic concept to start plaintext then upgrade - time and time again, it has been shown that to properly secure communications you must start with encryption, rather than layer it in later.
> 
> LDAP server implementations already have some form of minimum connection security requirements eg 389ds has minssf. Enforcing this alone and actually making it part of the standard would solve all the issues highlighted in your draft:

MinSSF is a server *response* to a message, by which time it's too late and the disclosure already occured. 

Error injection forcing StartTLS to not start - MinSSF won't resolve because the client sends the bind after the failed StartTLS meaning they plaintext credentials. Attacker wins.

Referral injection - the client is directed away from the server implementing MinSSF, so it is no longer enforced. Attacker wins.

Human error (forgetting StartTLS) - LDAP will send a bind first, which the server then rejections. Credentials were plaintexted. Attacker wins.

Clients that delay StartTLS operation - MinSSF *may* protect here depending on the content of the first operation (likely an anonymous bind). However, it would be better to just use LDAPS rather than rely on a leaky patch like MinSSF.

https://fy.blackhats.net.au/blog/html/2016/11/23/the_minssf_trap.html?highlight=minssf

> 
> 4.1. Error Handling or Injection 4.2. Referral Injection 4.3. Incorrect Client Handling 4.4. Configuration Error
> 
> because the only operations allowed would be startTLS or SASL bind (as long as the later install sufficient security layer/s). So this way you get what you want as a STEP towards the bright future of LDAPS only. On the other hand the service is easily discoverable (clients know its LDAP) and can provide unencrypted anonymous access and server information eg rootDSE contexts.

Incorrect, the bind operation would be sent plaintext, and then the server responds that minssf wasn't met. The leak has already occured. Additionally, minssf can't be enforced on startTLS meaning the referral attacks are still possible. 

> 
> > I think that the performance issues that existed when startTLS was recommended in the 2000s no longer exist either. Today startTLS is actually a negative on performance due to the extra round trips before communication can begin.
> 
> That is certainly not true. Even with modern CPU support for additional instructions to speed up crypto related functions and various optimizations made any additional layer on top of existing one would cause a slow down, by simple arithmetic. Even if you just XOR it is still O(N)+M or whatever. In some cases the difference will be peanuts but at a high volume and specific access patterns it can be huge and this has 2 main cost factors: TLS session establishment (many short lived connections) + encrypt/decrypt of already established sessions.
> 
> Quick google https://www.f5.com/company/blog/stop-just-stop-https-is-not-faster-than-http explains how that myth came about in HTTP world.

LDAPS to StartTLS, LDAPS is faster due to a reduction in overall round trips.

> 
> > So IMO there is no way to fix StartTLS, and this has already been highlighted in other protocols (IMAP, SMTP, POP).
> 
> Those are very different and not a good comparison to LDAP. LDAP has anonymous access and things like root DSE where LDAP serves information that has no need to be encrypted and in fact encrypting things like that is a hindrance and not an actual advantage. The email protocols you mention serve no such information ie all information they serve or exchange has to be protected. HTTP would be closer to LDAP in that sense eg accessing a public weather service over unencrypted connection as one example.

LDAP is an authority on groups and identities in unix environments and the ability to modify those responses can allow privilege escalation. Even hijacking a calendar response over ldap could allow a user to be tricked to email sensitive data to the incorrect recipient. 

There is really no good justification for plaintext LDAP given the position LDAP holds for security and identity management today. 

> 
> You propose a deprecation of startTLS without LDAPS being standardized in any way. The way it should be done is via LDAP standard actually declaring LDAP protocol over secure connection/transport a MUST ie evolving RFC4511 section 5.2 to require what is known as LDAPS. Then startTLS can be deprecated/obsolete. Without changing the LDAP standard itself this probably won't go anywhere.

I'll have to consider submitting that too then.

antbob:
@Firstyear You take min SSF literally as its related to 389ds where I think its only applicable to SASL/GSSAPI which is not the case for all LDAP server implementations. The notion of minimum security strength on a connection/session goes beyond that eg OpenLDAP has overall SSF and then it can be further dialed for SASL and TLS separately. We had a similar notion in OpenDS back when. If you have an idiotic client that sends credentials over insecure channels that wont help of course. Same goes for referrals. Although most clients I have seen do not follow referrals automatically or send credentials in clear those that do should not do so blindly.

On performance point I mean TLS (startTLS or LDAPS) vs PLAIN. There are cases (eg LDAP use in telco space) where nobody wants TLS in closed circuit systems bc its a CPU burner and just too damn slow vs PLAIN on specific access patterns and workloads eg cases where low response time and high thruput are requirements eg when you saturate network card/s bc you can go that fast if you are not burning CPU on useless churn. So when you say no good justification it just means "no good justification IMHO".

I have asked around how this can be done and it appears that if you want to evolve the LDAP standard ie LDAPS declaration and deprecation of startTLS you would have to call a new working group which i guess would be quite a hassle. This would also raise various issues like LDAP URL schemes, lack of backward compatibility, SASL/GSSAPI awareness of assumption of the transport layer (?2xTLS wrap on the same connection?) and i dunno what else.

If you just want to publish this as separate draft to raise awareness that is fine of course but making this into RFC i dont understand how that would work. You cannot obsolete LDAP RFCs that are part of the Standard Track already by publishing this and then anyone who implements this would then find themselves no longer able to claim they are LDAP standard compliant. So its like, whatever man.

In practical sense, without the long road of a working group, it would be more beneficial to everyone, based on the issues you raise, to tighten the startTLS implementation and configuration recommendations the following way:

for rootDSE and anonymous requests allow PLAIN

for everything else require startTLS

This won't address security for idiotic client implementations of course (as you have pointed out) but it would be better than what we have today and it would be inline with existing standards.

Firstyear:
> @Firstyear You take min SSF literally as its related to 389ds where I think its only applicable to SASL/GSSAPI which is not the case for all LDAP server implementations. 

You claimed minSSF resolves all these issues, and I pointed out it doesn't. For SASL, yes it does something. But the majority of the world doesn't use SASL. They use simple binds. 

> The notion of minimum security strength on a connection/session goes beyond that eg OpenLDAP has overall SSF and then it can be further dialed for SASL and TLS separately. We had a similar notion in OpenDS back when. If you have an idiotic client that sends credentials over insecure channels that wont help of course.

It's not okay to blame an "idiotic client" when so much of ldaps "difficulty" to use is from it's obtuse and complex user interface and experience. If people makes mistakes it's because *we let them* make the mistake. What did we do to confuse the user? What constraints were missing? 

> Same goes for referrals. Although most clients I have seen do not follow referrals automatically or send credentials in clear those that do should not do so blindly.
> 
> On performance point I mean TLS (startTLS or LDAPS) vs PLAIN. There are cases (eg LDAP use in telco space) where nobody wants TLS in closed circuit systems bc its a CPU burner and just too damn slow vs PLAIN on specific access patterns and workloads eg cases where low response time and high thruput are requirements eg when you saturate network card/s bc you can go that fast if you are not burning CPU on useless churn. So when you say no good justification it just means "no good justification IMHO".
> 
> I have asked around how this can be done and it appears that if you want to evolve the LDAP standard ie LDAPS declaration and deprecation of startTLS you would have to call a new working group which i guess would be quite a hassle. This would also raise various issues like LDAP URL schemes, lack of backward compatibility, SASL/GSSAPI awareness of assumption of the transport layer (?2xTLS wrap on the same connection?) and i dunno what else.
> 
> If you just want to publish this as separate draft to raise awareness that is fine of course but making this into RFC i dont understand how that would work. You cannot obsolete LDAP RFCs that are part of the Standard Track already by publishing this and then anyone who implements this would then find themselves no longer able to claim they are LDAP standard compliant. So its like, whatever man.
> 
> In practical sense, without the long road of a working group, it would be more beneficial to everyone, based on the issues you raise, to tighten the startTLS implementation and configuration recommendations the following way:
> 
> for rootDSE and anonymous requests allow PLAIN
> 
> for everything else require startTLS
> 
> This won't address security for idiotic client implementations of course (as you have pointed out) but it would be better than what we have today and it would be inline with existing standards.

So the tl;dr that you are saying here is "it's not possible to improve ldap because the WG is dead and too hard to revive, therefore we are stuck with what we have".

Firstyear:
Sounds like there is no point to trying to improve LDAP as a protocol then. 🙃

antbob:
@Firstyear sorry man, i feel bad now. i didnt mean for my comments to be negative or discouraging. i think your draft is absolutely worth publishing bc it highlights related security issues that are not really acknowledged or addressed by the current standard/s. in terms of other stake holders actually doing something about it (other than improving startTLS constraints) it just has to be addressed via standards track but you maybe right, maybe its time for a new protocol, LDAP been around for so long now it accumulated too much baggage and been too slow to evolve over the years with folks driving it back then have either retired or moved on. i think there is still want/need for LDAP data model but the protocol itself has always been the adoption blocker so maybe something modern like gRPC/protobuf could be an evolution tho those web thingies are transient/stateless and could be tricky to re-design around but yeah.

Firstyear:
> @Firstyear sorry man, i feel bad now. i didnt mean for my comments to be negative or discouraging. i think your draft is absolutely worth publishing bc it highlights related security issues that are not really acknowledged or addressed by the current standard/s. in terms of other stake holders actually doing something about it (other than improving startTLS constraints)

Yeah, and it just seems like there is no interest there. :( 

> it just has to be addressed via standards track but you maybe right, maybe its time for a new protocol, LDAP been around for so long now it accumulated too much baggage and been too slow to evolve over the years with folks driving it back then have either retired or moved on. i think there is still want/need for LDAP data model but the protocol itself has always been the adoption blocker so maybe something modern like gRPC/protobuf could be an evolution tho those web thingies are transient/stateless and could be tricky to re-design around but yeah.

I think that's a bigger topic, but https://datatracker.ietf.org/doc/html/rfc7644 fits many areas of the datamodel at least.

Issue #5469, opened 2022-09-30
Increase the default value of nsslapd-conntablesize


jchapma:
**Description of problem**
There are quite often customer cases about shortage of file descriptors and LDAP client requests not being processed. The shortage is usually caused by a low value set for the nsslapd-conntablesize parameter.

**Solution**
Increase the default value for nsslapd-conntablesize 

**Additional context**
https://issues.redhat.com/browse/IDMDS-2526

mreynolds389:
We also dynamically set the max fd's based off systemd config.  You can see it happening in the errors log:

    [30/Sep/2022:08:39:38.323920871 -0400] - INFO - main - Setting the maximum file descriptor limit to: 524288

Perhaps look into how we are doing it here so that we are consistent with conn table size.

jchapma:
> As far as I understand thing the conntablesize is set to the maximum number of open file for the process, (The one that you get with ulimit -n ) so IMHO we should increase the soft limit to the hard limit (using getrlimit and setrlimit ) when statring up the process

Currently the default conntablesize is obtained from the system fd soft limit, by default this is 1024. Yes we could set the process soft limit to match the hard limit, but as conntable memory is preallocated that might result in a lot of unused memory. Each connection takes roughly 40B, so to match the hard limit on my system would use ~20M of memory. Maybe that's not a big deal, would a default conntablesize of 10k not be enough ?

mreynolds389:
> > As far as I understand thing the conntablesize is set to the maximum number of open file for the process, (The one that you get with ulimit -n ) so IMHO we should increase the soft limit to the hard limit (using getrlimit and setrlimit ) when statring up the process
> 
> Currently the default conntablesize is obtained from the system fd soft limit, by default this is 1024. Yes we could set the process soft limit to match the hard limit, but as conntable memory is preallocated that might result in a lot of unused memory. Each connection takes roughly 40B, so to match the hard limit on my system would use ~20M of memory. Maybe that's not a big deal, would a default conntablesize of 10k not be enough ?

IMHO I think that would be a sufficient default for most deployments, but I'd be curious what other teammates think...

PR #5472, opened 2022-10-03
Issue 5469 - Increase the default value of nsslapd-conntablesize

jchapma:
Description:  There are quite often customer cases about shortage of file
	             descriptors and LDAP client requests not being processed.
                     The shortage is usually caused by a low value set for the
                     nsslapd-conntablesize parameter.

Relates:       https://github.com/389ds/389-ds-base/issues/5469

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/slap.h
+++ b/ldap/servers/slapd/slap.h
@@ -279,8 +279,7 @@ typedef void (*VFPV)(); /* takes undefined arguments */
 #define SLAPD_DEFAULT_MAXBERSIZE_STR "2097152"
 #define SLAPD_DEFAULT_MAXSIMPLEPAGED_PER_CONN (-1)
 #define SLAPD_DEFAULT_MAXSIMPLEPAGED_PER_CONN_STR "-1"
-/* We'd like this number to be prime for the hash into the Connection table */
-#define SLAPD_DEFAULT_CONNTABLESIZE 4093 /* connection table size */
+#define SLAPD_DEFAULT_CONNTABLESIZE 65536 /* connection table size */
 #define SLAPD_DEFAULT_NUM_LISTENERS 1 /* connection table lists */
 #define SLAPD_DEFAULT_LDAPSSOTOKEN_TTL 3600
 #define SLAPD_DEFAULT_LDAPSSOTOKEN_TTL_STR "3600"
--- a/dirsrvtests/tests/suites/config/config_delete_attr_test.py
+++ b/dirsrvtests/tests/suites/config/config_delete_attr_test.py
@@ -89,7 +89,6 @@ def test_reset_attributes(topology_st):
         'nsslapd-certmap-basedn',
         'nsslapd-port',
         'nsslapd-secureport',
-        'nsslapd-conntablesize',
         'nsslapd-rootpw',
         'nsslapd-hash-filters',
         'nsslapd-requiresrestart',
--- a/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
+++ b/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
@@ -3,6 +3,7 @@ import pytest
 import os
 import ldap
 import resource
+from lib389.backend import Backends
 from lib389._constants import *
 from lib389.topologies import topology_st
 from lib389.utils import ds_is_older, ensure_str
@@ -14,9 +15,11 @@ logging.getLogger(__name__).setLevel(logging.INFO)
 log = logging.getLogger(__name__)
 
 FD_ATTR = "nsslapd-maxdescriptors"
+RESRV_FD_ATTR = "nsslapd-reservedescriptors"
 GLOBAL_LIMIT = resource.getrlimit(resource.RLIMIT_NOFILE)[1]
 SYSTEMD_LIMIT = ensure_str(check_output("systemctl show -p LimitNOFILE dirsrv@standalone1".split(" ")).strip()).split('=')[1]
 CUSTOM_VAL = str(int(SYSTEMD_LIMIT) - 10)
+RESRV_DESC_VAL = str(10)
 TOO_HIGH_VAL = str(GLOBAL_LIMIT * 2)
 TOO_HIGH_VAL2 = str(int(SYSTEMD_LIMIT) * 2)
 TOO_LOW_VAL = "0"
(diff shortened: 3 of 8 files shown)


ldap/servers/slapd/slap.h
```diff
@@ -279,8 +279,7 @@ typedef void (*VFPV)(); /* takes undefined arguments */
 #define SLAPD_DEFAULT_MAXBERSIZE_STR "2097152"
 #define SLAPD_DEFAULT_MAXSIMPLEPAGED_PER_CONN (-1)
 #define SLAPD_DEFAULT_MAXSIMPLEPAGED_PER_CONN_STR "-1"
-/* We'd like this number to be prime for the hash into the Connection table */
-#define SLAPD_DEFAULT_CONNTABLESIZE 4093 /* connection table size */
+#define SLAPD_DEFAULT_CONNTABLESIZE 65536 /* connection table size */
```
Firstyear, line 282: This seems like a really large number - remind me, is this conntable size bound by max fd's somewhere? Else we'll hit a different set of resource exhaustion issues.

Firstyear (changes requested):

ldap/servers/slapd/slap.h
jchapma: Hi @Firstyear 
I had a chat with the team and it seems a connection table size of 65k is not an uncommon deployment, so we went with that. 

Regarding resource limits, as long as the soft limit is greater than conntablesize then we are good, during startup. after the config is loaded the server increases the soft limit to match the hard limit.

mreynolds389: I did suggest, but I think it was overlooked, that we can dynamically set the conntable size, like how we set the FD limit in DS.  We do it start up, and it can be seen in the errors log.

jchapma: @mreynolds389 
Apologies if you think I ignored your suggestion, that wasn't my intention. 
Currently the conntablesize is set dynamically to the system fd soft limit, by default this is 1024 and the root of our problem. From what I understand the fd limits in DS are based on the systems soft and hard fd limits so I don't see how we can make the conntablesize dynamic, unless we do something to calculate what we think it should be.

Firstyear: There are no such thing as root causes, only contributing factors :) 

A dynamic conntable would be a nightmare given the need to realloc the table. I think it would be dynamic with regard to restarts and reading the fd limit and choosing a %of that fd limit to set the conntable amount. But that has other risks I think if the number is too high.

jchapma: We could have a self tuning conntablesize, where the server initially starts up with a default value, gets a count of the maximum number of connections the server had to handle, on shutdown/restart the server sets the conntablesize attribute to the max number of connections handled. Or is this overkill ?

mreynolds389: It is overkill, why can't we just handle this like we handle the FD limit in config_set_maxdescriptors()?

progier389: ==> IMHO in this specific case, self tuning is a good way to get in trouble. I think that administrator would want to be able to control on the limit:  
The number of simultaneous connection is very dynamic and a simple slow down of the server tend to generate connection peaks (if the incoming connection rate stay stable but connection last longer because the server is a slower to answer than usual then the average number of open connection increase) so IMHO you can easily have a peak that is several time the average value 

And I agree that changing dynamically the connection table size after its creation will be problematic with the current design. But I am not sure that we really need a connection table (we may want to let the worker thread alloc the connection context and free it once the connection is fully closed (and no more thread is using it) 
 It may be better on numa machine because the connection memory will be near the cpu running the worker threads
  or it may be worse because of the heap access serialization -;)
 Anyway my feeling is that it had to much impact to do that know, but that may be something to remember if we  do
  another change that impact the connection code design (like if we ever move from nss to openssl)

progier389: Hi @mreynolds389 , 
 I am not sure what you mean exactly:
 As I understand it, the idea behind config_set_maxdescriptors is to the the fi;e descriptor limit according to the dse.ldif value (and decrease the configured value if the system cannot use that value) 
 But here we want to compute the connection default value (BTW: in the libglob,c code that is removed
 in this PR, The config parameter are not yet even read) 
  Now I do not think that we could set  the conn table size to the config_set_maxdescriptors
    value (because there is a serious risk of db corruption if we run out of fds) 
But IMHO the proposed code is buggy and will leads to trouble if config_set_maxdescriptors set the value below 64K ( I think that we need to check before creating the conntectiom table  that 
  maxconn + reserveddesc <= maxdesc and decrease maxconn if it is not the case 
    and log an info message in that case )

mreynolds389: @progier this what I was referring to

```
    if (getrlimit(RLIMIT_NOFILE, &rlp) == 0) {
            conntable =  (int64_t)rlp.rlim_max - 64  // something like this
    }
```

progier389: Hum ! would rather use config_get_reservedescriptors() instead of 64 
(Using an hardcoded value of 64 seems dangerous as some config have more than 64 indexes) 
and I would also add a second test to limit the value to 64K 
But I agree with the idea ( and we should also fix the doc because the nsslapd-conntablesize value will then be ignored.)
I would also add a test to increase the reserveddescriptor value if it is below 64 + 3*nbBackends + nbIndexes

mreynolds389: > Hum ! would rather use config_get_reservedescriptors() instead of 64 (Using an hardcoded value of 64 seems dangerous as some config have more than 64 indexes) 

Sorry again it was a rough example, like I said "something like this"  :-)  I said 64 because that is what that reserve descriptors are, but I was just trying to highlight my approach since I've been previously misunderstood.  

> and I would also add a second test to limit the value to 64K But I agree with the idea ( and we should also fix the doc because the nsslapd-conntablesize value will then be ignored.) I would also add a test to increase the reserveddescriptor value if it is below 64 + 3*nbBackends + nbIndexes

+1

jchapma: This commit removes a configurable conntablesize, we now calculate it using (maxdesc - reservedesc), A check has been added for the process 64k soft limit. 

Regarding the reserveddescriptor value, I noticed that on my dev vm the non client connection fds used by the server are  greater than 64. So how about we have a dynamic reserve descriptors  value using the formula from the RH DS docs.
image

progier389: Not sure we should  make reserve descriptor fully dynamic (as unknown plugins may need file descriptors)
IMHO we should dynamically increase the configured value if the formula shows it is too small 
otherwise I am ok with your formula ...

jchapma: > Not sure we should make reserve descriptor fully dynamic (as unknown plugins may need file descriptors)
> IMHO we should dynamically increase the configured value if the formula shows it is too small
> otherwise I am ok with your formula ...

Yes this makes sense, I never considered future plugins that might require fds. I will  create a separate PR for the reserve-descriptor change.

progier389 (approved):
LGTM

Firstyear (approved):

PR #5473, opened 2022-10-04
Issue 5302 - Release tarballs don't contain cockpit webapp

### LINKED ISSUE #5302: Release tarballs don't contain cockpit webapp

vashirov:
Bug Description:
It was found that this action was producing incomplete tarballs **only** containing the cockpit webapp and rust vendored dependencies. The reason is the same as in #5307 - current working directory was not marked as a safe directory.

Fix Description:
* Add $GITHUB_WORKSPACE as a safe directory.
* Add a manual trigger that allows you to specify a tag and whether to skip npm audit-ci or not.

Fixes: https://github.com/389ds/389-ds-base/issues/5302

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,15 @@ on:
   push:
     tags:
       - "389-ds-base-*"
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Specify tag to generate a tarball
+        required: true
+      skip-audit-ci:
+        description: Skip npm audit-ci
+        type: boolean
+        default: false
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
Ack, thanks Viktor!!!!

PR #5474, opened 2022-10-04
Issue 3061 - RFE - Add password policy debug log level

mreynolds389:
Description:

Added new logging level (1048576) for password policy debugging where the failure message, entry dn, and pwd policy is recorded (it's either the global policy or the DN of the local policy).

relates: https://github.com/389ds/389-ds-base/issues/3061


Firstyear (approved):
This seems like an excellent change. Given the relatively low frequency of pw resets, and the value this would give to administrators, is there a reason not to enable this by default?

droideck (changes requested):
Some minor test message text failure:

```
        err_msg = "PWDPOLICY_DEBUG - invalid password syntax - password must be at least 6 characters long:  Entry " + \
                  "\\(uid=local_user,ou=people,dc=example,dc=com\\) Policy \\(cn="
>       assert inst.searchErrorsLog(err_msg)
E       AssertionError: assert False
E        +  where False = <bound method DirSrv.searchErrorsLog of <lib389.DirSrv object at 0x7ff423203df0>>('PWDPOLICY_DEBUG - invalid password syntax - password must be at least 6 characters long:  Entry \\(uid=local_user,ou=people,dc=example,dc=com\\) Policy \\(cn=')
E        +    where <bound method DirSrv.searchErrorsLog of <lib389.DirSrv object at 0x7ff423203df0>> = <lib389.DirSrv object at 0x7ff423203df0>.searchErrorsLog
```

Also, the validation test suite has failed because of on minor some invalid docstrings.

mreynolds389:
> This seems like an excellent change. Given the relatively low frequency of pw resets, and the value this would give to administrators, is there a reason not to enable this by default?

Well I'd like to keep it a log level level for a few reasons:

- People always get upset when the server starts logging new messages by default.  Trust me every time someone complains about it
- The logging is only useful if the password policy is not working as they think it should
- The logging could evolve into more messages for benign events (creating more log bloat that not everyone would want).

What do others think?  To me this is debug logging and doesn't serve a good purpose if password policy is working as intended.

mreynolds389:
> Some minor test message text failure:
> 
> ```
>         err_msg = "PWDPOLICY_DEBUG - invalid password syntax - password must be at least 6 characters long:  Entry " + \
>                   "\\(uid=local_user,ou=people,dc=example,dc=com\\) Policy \\(cn="
> >       assert inst.searchErrorsLog(err_msg)
> E       AssertionError: assert False
> E        +  where False = <bound method DirSrv.searchErrorsLog of <lib389.DirSrv object at 0x7ff423203df0>>('PWDPOLICY_DEBUG - invalid password syntax - password must be at least 6 characters long:  Entry \\(uid=local_user,ou=people,dc=example,dc=com\\) Policy \\(cn=')
> E        +    where <bound method DirSrv.searchErrorsLog of <lib389.DirSrv object at 0x7ff423203df0>> = <lib389.DirSrv object at 0x7ff423203df0>.searchErrorsLog
> ```
> 
> Also, the validation test suite has failed because of on minor some invalid docstrings.

Of course at the last minute I slightly tweaked the log message which broke the test.  The test is now fixed.

dirsrvtests/tests/suites/password/pwdPolicy_logging_test.py
```diff
@@ -0,0 +1,151 @@
...
+    :id: cc152c65-94e0-4716-a77c-abdd2deec00d
+    :setup: Standalone Instance
+    :steps:
+        1. Set password policy logging level
+        2. Add database entries
+        3. Configure local and global policies
+        4. Test syntax checking on local policy
+        5  Test syntax checking on global policy
+        6. Test account lockout on local policy
+        7. Test account lockout on global policy
```
droideck, line 77: I think an empty line is needed between ':steps:' and ':expectedresults:'

droideck (approved):
Minor validation issue. But otherwise - ack from me!

And I agree with the reasons and I also think it's better to leave it as a separate log level.

dirsrvtests/tests/suites/password/pwdPolicy_logging_test.py
```diff
@@ -0,0 +1,152 @@
...
+    """Enable password policy logging
+
+    :id: cc152c65-94e0-4716-a77c-abdd2deec00d
+    :setup: Standalone Instance
+    :steps:
+        1. Set password policy logging level
+        2. Add database entries
+        3. Configure local and global policies
+        4. Test syntax checking on local policy
+        5  Test syntax checking on global policy
```
droideck: Oh, sorry, I think, the issue is in this line.
It should be `5. Test syntax checking on global policy` - so you need to add a dot after `5`, otherwise, the Validate test fails. Ugh...

Firstyear:
> > This seems like an excellent change. Given the relatively low frequency of pw resets, and the value this would give to administrators, is there a reason not to enable this by default?
> 
> Well I'd like to keep it a log level level for a few reasons:
> 
> * People always get upset when the server starts logging new messages by default.  Trust me every time someone complains about it
> * The logging is only useful if the password policy is not working as they think it should
> * The logging could evolve into more messages for benign events (creating more log bloat that not everyone would want).
> 
> What do others think? To me this is debug logging and doesn't serve a good purpose if password policy is working as intended.

These are all fair reasons. Sometimes I think people complain for silly reasons, but it's all good. Still glad the extra debugging is there.

PR #5477, opened 2022-10-06
Issue 5476 - RFE - add memberUid read aci by default

### LINKED ISSUE #5476: Add memberUid to aci by default
Some legacy methods of group lookup use memberUid which by default we don't allow. Because these searches look for member AND memberUid, the lack of MemberUid access causes the search to fail. 

Add an aci for reading this by default.

Firstyear:
Bug Description: Some legacy client issue searches for (|(member=...)(memberUid=...)). Because of how our aci's work, the lack of access to memberUid (even if empty) would block the legitimate search to member=.

Fix Description: Allow memberUid to be searched by default.

fixes: https://github.com/389ds/389-ds-base/issues/5476

Review by: ???

### DIFF
--- a/src/lib389/lib389/configurations/__init__.py
+++ b/src/lib389/lib389/configurations/__init__.py
@@ -11,17 +11,22 @@ from lib389._constants import INSTALL_LATEST_CONFIG
 from .config_001003006 import c001003006, c001003006_sample_entries
 from .config_001004000 import c001004000, c001004000_sample_entries
 from .config_001004002 import c001004002, c001004002_sample_entries
+from .config_002003000 import c002003000, c002003000_sample_entries
 
 
 def get_config(version):
     # We do this to avoid test breaking on older version that may
     # not expect the new default layout.
+    if (version == INSTALL_LATEST_CONFIG and ds_is_newer('2.3.0')):
+        return c002003000
     if (version == INSTALL_LATEST_CONFIG and ds_is_newer('1.4.2')):
         return c001004002
     elif (version == INSTALL_LATEST_CONFIG and ds_is_newer('1.4.0')):
         return c001004000
     elif (version == INSTALL_LATEST_CONFIG):
         return c001003006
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
ack

src/lib389/lib389/configurations/__init__.py
```diff
@@ -11,17 +11,22 @@
...
+    if (version == INSTALL_LATEST_CONFIG and ds_is_newer('2.3.0')):
+        return c002003000
     if (version == INSTALL_LATEST_CONFIG and ds_is_newer('1.4.2')):
         return c001004002
     elif (version == INSTALL_LATEST_CONFIG and ds_is_newer('1.4.0')):
         return c001004000
     elif (version == INSTALL_LATEST_CONFIG):
         return c001003006
+    elif (version == '002003000' and ds_is_newer('2.3.0')):
+        return c001004002
```
droideck: Shouldn't it be `return c002003000`?

src/lib389/lib389/configurations/config_002003000.py
```diff
@@ -0,0 +1,134 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2020 Red Hat, Inc.
```
droideck: It was not the best year, lets forget it :D

droideck (approved):
Besides a couple of minor issues, LGTM!

src/lib389/lib389/configurations/__init__.py
Firstyear: Opps! Good catch, fixed

src/lib389/lib389/configurations/config_002003000.py
Firstyear: Also fixed :)

droideck (approved):

Firstyear:
Thanks!

PR #5480, opened 2022-10-06
Issue 5356 - Update Cargo.lock and bootstrap PBKDF2-SHA512

droideck:
Description: We've changed our default password storage scheme to PBKDF2-SHA512. Change the bootstrap logic to use the scheme when Rust is enabled.

Related: https://github.com/389ds/389-ds-base/issues/5356

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/config.c
+++ b/ldap/servers/slapd/config.c
@@ -40,16 +40,16 @@ char *rel2abspath(char *);
  * see fedse.c instead!
  */
 static char *bootstrap_plugins[] = {
-    "dn: cn=PBKDF2_SHA256,cn=Password Storage Schemes,cn=plugins,cn=config\n"
+#ifdef RUST_ENABLE
+    "dn: cn=PBKDF2-SHA512,cn=Password Storage Schemes,cn=plugins,cn=config\n"
     "objectclass: top\n"
     "objectclass: nsSlapdPlugin\n"
-    "cn: PBKDF2_SHA256\n"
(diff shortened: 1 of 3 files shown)


droideck:
I'm also looking into a weird entryuuid failure in 389-ds-base-2.0 branch... 

```
        # 4. Check the entryuuid is valid
        euuid = account.get_attr_val_utf8('entryUUID')
        print(f"❄️   account entryUUID -> {euuid}");
>       assert(euuid != "INVALID_UUID")
E       AssertionError: assert 'INVALID_UUID' != 'INVALID_UUID'
```

But still, I wasn't able to find the reason why it fails here but works on 389-ds-base-2.1. The code seems identical. Continue looking...

droideck:
> I'm also looking into a weird entryuuid failure in 389-ds-base-2.0 branch...
> 
> ```
>         # 4. Check the entryuuid is valid
>         euuid = account.get_attr_val_utf8('entryUUID')
>         print(f"❄️   account entryUUID -> {euuid}");
> >       assert(euuid != "INVALID_UUID")
> E       AssertionError: assert 'INVALID_UUID' != 'INVALID_UUID'
> ```
> 
> But still, I wasn't able to find the reason why it fails here but works on 389-ds-base-2.1. The code seems identical. Continue looking...

It takes more time than I though as I can't easily pin down the commit that causes it it fail (it works on 389-ds-base-2.1 and main).
I'll do another PR for entryuuid (and other failures I'll find).

mreynolds389 (approved):
Ack

Issue #5482, opened 2022-10-09
With Case sensitive rootDn, 389ds - enbling replication fails.


cooldharma06:
**Issue Description**
When 389ds ldap server was configured with Case-sensitive values (ou=Example,o=orgz) for backend suffix. dsconf cli fails to enable replication with "No backend associated with nsslapd-suffix" suffix error message. But in backend, replication got enabled.

**Package Version and Platform:**
 - Platform: 
 - Package and version: 389-ds-base-2.3.0

**Steps to Reproduce**
1. Install 389ds with Case sensitive suffix.
2. Enable replication with dsconf cli

**Expected results**
Replication got enabled successfully for DN.

**Actual result**
No backend associated with nsslapd-suffix "{ou=Example,o=orgz}

**Additional context**
When checking with debug/verbose option we could find that in replica.py, Changelog specific code is throwing error while evaluating the suffix.

https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/replica.py line no:1069 if condition couldn't handle the Case-sensitive suffix.

mreynolds389:
2a2262723..c284b34c5  389-ds-base-2.2 -> 389-ds-base-2.2
41c8e4101..993fe84ed  389-ds-base-2.1 -> 389-ds-base-2.1
0aa5e12e1..63166c176  389-ds-base-2.0 -> 389-ds-base-2.0
2a6870bcc..203085c1c  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

progier389:
No the server will accept the request (as required by the standard), what @mreynolds389 was speaking about was the dn that are returned by search requests.
  Historically the server was returning dn as it was when the entry was created
  A few years ago the server was returning dn that is rebuilt from the entry and parent entries rdn (so case may be different)
  @mreynolds389 is speaking about being able to return again the dn as it was when the entry get added (because if broke compatibility issue with some apllication that are not compliant to the ldap standard 
   (they compare dn as string rather than comparing them properly as specified in the standard) 
  As it is easier to add yet another configuration parameter to mimic the old behavior rather than fixing these old applications we have done so !-;) )

cooldharma06:
@progier389 Thank you for clarifying me. I am completely fine with this.

PR #5485, opened 2022-10-10
Issue 5482 - lib389 - Can not enable replication with a mixed case suffix

mreynolds389:
Can not enable replication with a mixed case suffix using dsconf because not all of the code expects mixed case.  Need to "lower" some values before comparing them.

Also did a little code cleanup

relates: https://github.com/389ds/389-ds-base/issues/5482


progier389 (approved):
LGTM

droideck (approved):
Wow, the chain of imports with an asterisk finally backfired on us... :D 
Some tests fail because of the missing 'os' here, DEFAULT_SUFFIX there.
Could you please bring the missing imports to the failed tests? :)

Otherwise, the change looks good! And it's a nice clean-up!

Issue #5487, opened 2022-10-13
Fix various isses with logconv.pl


mreynolds389:
**Issue Description**

[Bug 2110577](https://bugzilla.redhat.com/show_bug.cgi?id=2110577) - logconv.pl - "Connection Latency Details" has different results with options -y and -V

The array that holds the latency stats values was not initialized leading to inconsistent results

[Bug 2095448](https://bugzilla.redhat.com/show_bug.cgi?id=2095448) - Logconv.pl CSV file contains mismatched header and data columns

mreynolds389:
c284b34c5..ccf3a485c  389-ds-base-2.2 -> 389-ds-base-2.2
993fe84ed..2c3ddfbec  389-ds-base-2.1 -> 389-ds-base-2.1
63166c176..78af26f10  389-ds-base-2.0 -> 389-ds-base-2.0
b165bdfb4..71396b5ba  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

jchapma:
68de69a1f..b840c7737  389-ds-base-2.5 -> 389-ds-base-2.5
50d5912ff..69a2aee3b  389-ds-base-2.4 -> 389-ds-base-2.4
a4f7633fb..9da508584  389-ds-base-2.3 -> 389-ds-base-2.3
347aae6f7..d49d5fb29  389-ds-base-2.2 -> 389-ds-base-2.2
c1daaabde..27ad77e7c  389-ds-base-2.1 -> 389-ds-base-2.1
30a91b58b..b7bc23563  389-ds-base-2.0 -> 389-ds-base-2.0
692c4cec6..b204c9fb2  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5488, opened 2022-10-13
Issue 5487 - Fix various issues with logconv.pl

mreynolds389:
Description:

Latency stats were inconsistent depending if "-V" was used. This was due to a missing usage check and uninitialized latency array

Fixed CSV format

relates: https://github.com/389ds/389-ds-base/issues/5487


droideck (approved):
Looks good to me!

Issue #5267, opened 2022-04-13
RFE - store full DN in database record


mreynolds389:
While testing "nsslapd-subtree-rename-switch=off" there was noticeable performance improvement.  The current implementation generates the DN dynamically by using the entryrdn attribute in the database record, and then obtaining the parent portion of the DN from other database indexes.  

Disabling subtree-rename (nsslapd-subtree-rename-switch=off) improves throughput/response_time performance especially when the number of clients increases.

The range of improvement on highest load (high number of clients) ranges up to ~10%

```
subtree-rename=OFF 
----------------------------------------------------
Workers         Clients         SRCH/s          GAIN
15              40              142302.868      + 0%
15              45              157062.354      + 4%    
15              50              165993.278      +12%
15              55              168967.106      +15%

20              40              133749.551      + 2%
20              45              148529.026      + 0%
20              50              150216.182      + 1%
20              55              114247.201      +13%

25              40              123818.029      + 1%
25              45              141934.678      + 1%
25              50              153096.810      + 0%
25              55              109355.695      +10%


subtree-rename=ON
----------------------------------------------------
Workers         Clients         SRCH/s
15              40              142297.329
15              45              150931.420
15              50              147570.526
15              55              146769.262

20              40              130967.392
20              45              148365.746
20              50              148903.385
20              55              121286.399

25              40              122408.771
25              45              140797.527
25              50              151812.222
25              55               98683.023
```

We could also maintain the existing operational attribute "entrydn" and only use it for DN "generation" on searches.  It would need to be updated on adds, and modrdns.  There could also be an impact on tombstones queries as well.

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2075017

mreynolds389:
b1ed566c4..4b4bb9705  389-ds-base-2.2 -> 389-ds-base-2.2

mreynolds389:
CI fix:
26eaa1b7d..86fb2b998  389-ds-base-2.2 -> 389-ds-base-2.2

PR #5494, opened 2022-10-20
RFE - store full DN in database record 

mreynolds389:
While testing "nsslapd-subtree-rename-switch=off" there was noticeable performance improvement. The current implementation generates the DN dynamically by using the entryrdn attribute in the database record, and then obtaining the parent portion of the DN from other database indexes.

Disabling subtree-rename (nsslapd-subtree-rename-switch=off) improves throughput/response_time performance especially when the number of clients increases.

The range of improvement on highest load (high number of clients) ranges up to ~10%

-------------------------------------------------------------------------------------------------------------------

This PR brings in the short-cut of already having a DN, and not needing to call entryrdn_lookup_dn() which is very expensive.   This shows a 35% improvement when moving entries from disk into the entry cache, There is also an improvement once the cache is primed, but a full performance test run is needed.  Regardless this definitely provides a nice perf boost...

Note - this new attribute "dsEntryDN" stores the DN in the exact case that was used when the entry was added.  This helps with migrations from ODSEE.

https://github.com/389ds/389-ds-base/issues/5267

### DIFF
--- a/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
+++ b/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
@@ -1,0 +1,96 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import logging
+import pytest
+import os
+import time
+from lib389.topologies import topology_st as topo
+from lib389.idm.user import UserAccount, UserAccounts
+from lib389.idm.organizationalunit import OrganizationalUnit
+
+log = logging.getLogger(__name__)
+
+SUFFIX = "dc=Example,DC=COM"
+SUBTREE = "ou=People,dc=Example,DC=COM"
+NEW_SUBTREE = "ou=humans,dc=Example,DC=COM"
+USER_DN = "uid=tUser,ou=People,dc=Example,DC=COM"
+NEW_USER_DN = "uid=tUser,ou=humans,dc=Example,DC=COM"
+NEW_USER_NORM_DN = "uid=tUser,ou=humans,dc=example,dc=com"
+
+
+def test_dsentrydn(topo):
+    """Test that the dsentrydn attribute is properly maintained and preserves the case of the DN
+
+    :id: f0f2fe6b-c70d-4de1-a9a9-06dda74e7c30
+    :setup: Standalone Instance
+    :steps:
+        1. Create user and make sure dsentrydn is set to the correct value/case
+        2. Moddn of "ou=people" and check dsentrydn is correct for parent and the children
+        3. Check the DN matches dsEntryDN
+        4. Disable nsslapd-return-original-entrydn
+        5. Check the DN matches normalized DN
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+    """
+
+    inst = topo.standalone
+    inst.config.replace('nsslapd-return-original-entrydn', 'on')
+
+    # Create user and makes sure "dsEntryDN" is set correctly
+    users = UserAccounts(inst, SUFFIX)
+    user_properties = {
+        'uid': 'tUser',
+        'givenname': 'test',
+        'cn': 'Test User',
+        'sn': 'user',
+        'userpassword': 'password',
+        'uidNumber': '1000',
+        'gidNumber': '2000',
+        'homeDirectory': '/home/tUser'
+    }
+    user = users.create(properties=user_properties)
+    assert user.get_attr_val_utf8('dsentrydn') == USER_DN
+
+    # Move subtree ou=people to ou=humans
+    ou = OrganizationalUnit(inst, SUBTREE)
+    ou.rename("ou=humans", SUFFIX)  # NEW_SUBTREE
+
+    # check dsEntryDN is properly updated to new subtree
+    ou = OrganizationalUnit(inst, NEW_SUBTREE)
+    assert ou.get_attr_val_utf8('dsentrydn') == NEW_SUBTREE
+
+    user = UserAccount(inst, NEW_USER_DN)
+    assert user.get_attr_val_utf8('dsentrydn') == NEW_USER_DN
+
+    # Check DN retruend to client matches "dsEntryDN"
+    users = UserAccounts(inst, SUFFIX).list()
+    for user in users:
+        if user.dn.startswith("tUser"):
+            assert user.dn == NEW_USER_DN
+            break
+
+    # Disable 'nsslapd-return-original-entrydn' andcheck DN is normalized
+    inst.config.replace('nsslapd-return-original-entrydn', 'on')
+    users = UserAccounts(inst, SUFFIX).list()
+    for user in users:
+        if user.dn.startswith("tUser"):
+            assert user.dn == NEW_USER_NORM_DN
+            break
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
+++ b/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
@@ -656,7 +656,7 @@ def test_internal_log_level_516(topology_st, add_user_log_level_516, disable_acc
                                     r'SRCH base="cn=group,ou=Groups,dc=example,dc=com".*')
     # (Internal) op=10(1)(2) ENTRY dn="cn=group,ou=Groups,dc=example,dc=com"
     assert topo.ds_access_log.match(r'.*\(Internal\) op=[0-9]+\([0-9]+\)\([0-9]+\) '
-                                    r'ENTRY dn="cn=group,ou=groups,dc=example,dc=com".*')
+                                    r'ENTRY dn="cn=group,ou=Groups,dc=example,dc=com".*')
     # (Internal) op=10(1)(2) RESULT err=0 tag=48 nentries=1*')
     assert topo.ds_access_log.match(r'.*\(Internal\) op=[0-9]+\([0-9]+\)\([0-9]+\) RESULT err=0 tag=48 nentries=1*')
     # (Internal) op=10(1)(1) RESULT err=0 tag=48
--- a/ldap/schema/01core389.ldif
+++ b/ldap/schema/01core389.ldif
@@ -329,6 +329,8 @@ attributeTypes: ( 2.16.840.1.113730.3.1.2374 NAME 'nsDS5ReplicaBootstrapTranspor
 attributeTypes: ( 2.16.840.1.113730.3.1.2387 NAME 'nsslapd-tcp-fin-timeout' DESC 'Netscape defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN 'Netscape Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2388 NAME 'nsslapd-tcp-keepalive-time' DESC 'Netscape defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN 'Netscape Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2390 NAME 'nsds5ReplicaKeepAliveUpdateInterval' DESC '389 defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.9998 NAME 'dsEntryDN' DESC '389 Directory Server defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 NO-USER-MODIFICATION SINGLE-VALUE USAGE directoryOperation X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.9999 NAME 'nsslapd-return-original-entrydn' DESC '389 Directory Server defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
 #
 # objectclasses
 #
--- a/ldap/servers/slapd/back-ldbm/id2entry.c
+++ b/ldap/servers/slapd/back-ldbm/id2entry.c
@@ -67,6 +67,10 @@ id2entry_add_ext(backend *be, struct backentry *e, back_txn *txn, int encrypt, i
         int options = SLAPI_DUMP_STATEINFO | SLAPI_DUMP_UNIQUEID;
         Slapi_Entry *entry_to_use = encrypted_entry ? encrypted_entry->ep_entry : e->ep_entry;
         memset(&data, 0, sizeof(data));
+        entrydn = slapi_entry_get_dn(entry_to_use);
+        slapi_entry_attr_set_charptr(entry_to_use, SLAPI_ATTR_DS_ENTRYDN,
+                entrydn);
+
         if (entryrdn_get_switch()) {
             struct backdn *oldbdn = NULL;
             Slapi_DN *sdn =
@@ -141,7 +145,7 @@ id2entry_add_ext(backend *be, struct backentry *e, back_txn *txn, int encrypt, i
                         myparentdn = slapi_dn_parent_ext(
                             slapi_entry_get_dn_const(e->ep_entry),
                             is_tombstone);
-                        if (myparentdn && PL_strcmp(parentdn, myparentdn)) {
+                        if (myparentdn && PL_strcasecmp(parentdn, myparentdn)) {
                             Slapi_DN *sdn = slapi_entry_get_sdn(e->ep_entry);
                             char *newdn = NULL;
                             CACHE_LOCK(&inst->inst_cache);
@@ -350,21 +354,28 @@ id2entry(backend *be, ID id, back_txn *txn, int *err)
                 CACHE_RETURN(&inst->inst_dncache, &bdn);
             } else {
                 Slapi_DN *sdn = NULL;
-                rc = entryrdn_lookup_dn(be, rdn, id, &normdn, &srdn, txn);
-                if (rc) {
-                    slapi_log_err(SLAPI_LOG_TRACE, ID2ENTRY,
-                                  "id2entry: entryrdn look up failed "
-                                  "(rdn=%s, ID=%d)\n",
-                                  rdn, id);
-                    /* Try rdn as dn. Could be RUV. */
-                    normdn = slapi_ch_strdup(rdn);
-                } else if (NULL == normdn) {
-                    slapi_log_err(SLAPI_LOG_ERR, ID2ENTRY,
-                                  "id2entry( %lu ) entryrdn_lookup_dn returned NULL. "
-                                  "Index file may be deleted or corrupted.\n",
-                                  (u_long)id);
-                    goto bail;
+                if (config_get_return_orig_dn() &&
+                    !get_value_from_string((const char *)data.dptr, SLAPI_ATTR_DS_ENTRYDN, &normdn))
+                {
+                    srdn = slapi_rdn_new_all_dn(normdn);
+                } else {
+                    rc = entryrdn_lookup_dn(be, rdn, id, &normdn, &srdn, txn);
+                    if (rc) {
+                        slapi_log_err(SLAPI_LOG_TRACE, ID2ENTRY,
+                                      "id2entry: entryrdn look up failed "
+                                      "(rdn=%s, ID=%d)\n",
+                                      rdn, id);
+                        /* Try rdn as dn. Could be RUV. */
+                        normdn = slapi_ch_strdup(rdn);
+                    } else if (NULL == normdn) {
+                        slapi_log_err(SLAPI_LOG_ERR, ID2ENTRY,
+                                      "id2entry( %lu ) entryrdn_lookup_dn returned NULL. "
+                                      "Index file may be deleted or corrupted.\n",
+                                      (u_long)id);
+                        goto bail;
+                    }
                 }
+
                 sdn = slapi_sdn_new_normdn_byval((const char *)normdn);
                 bdn = backdn_init(sdn, id, 0);
                 if (CACHE_ADD(&inst->inst_dncache, bdn, NULL)) {
@@ -395,6 +406,7 @@ id2entry(backend *be, ID id, back_txn *txn, int *err)
 
         /* All entries should have uniqueids */
         PR_ASSERT(slapi_entry_get_uniqueid(ee) != NULL);
+
         /* ownership of the entry is passed into the backentry */
         e = backentry_init(ee);
         e->ep_id = id;
--- a/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
@@ -23,6 +23,8 @@ static IDList *moddn_get_children(back_txn *ptxn, Slapi_PBlock *pb, backend *be,
 static int moddn_rename_children(back_txn *ptxn, Slapi_PBlock *pb, backend *be, IDList *children, Slapi_DN *dn_parentdn, Slapi_DN *dn_newsuperiordn, struct backentry *child_entries[]);
 static int modrdn_rename_entry_update_indexes(back_txn *ptxn, Slapi_PBlock *pb, struct ldbminfo *li, struct backentry *e, struct backentry **ec, Slapi_Mods *smods1, Slapi_Mods *smods2, Slapi_Mods *smods3, Slapi_Mods *smods4);
 static void mods_remove_nsuniqueid(Slapi_Mods *smods);
+static int32_t dsentrydn_modrdn_update(backend *be, const char *newdn, struct backentry *e, struct backentry **ec, back_txn *txn);
+static int dsentrydn_moddn_rename(back_txn *ptxn, backend *be, ID id, IDList *children, const Slapi_DN *dn_parentdn, const Slapi_DN *dn_newsuperiordn, struct backentry *child_entries[]);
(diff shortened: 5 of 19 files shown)


mreynolds389:
Added a config setting to control how the DN is returned to a client (case intact, or normalized), but regardless what DN is returned to the client there is still the perf boost.  So the config setting just controls what DN the client sees.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -23,6 +23,8 @@ static IDList *moddn_get_children(back_txn *ptxn, Slapi_PBlock *pb, backend *be,
 static int moddn_rename_children(back_txn *ptxn, Slapi_PBlock *pb, backend *be, IDList *children, Slapi_DN *dn_parentdn, Slapi_DN *dn_newsuperiordn, struct backentry *child_entries[]);
 static int modrdn_rename_entry_update_indexes(back_txn *ptxn, Slapi_PBlock *pb, struct ldbminfo *li, struct backentry *e, struct backentry **ec, Slapi_Mods *smods1, Slapi_Mods *smods2, Slapi_Mods *smods3, Slapi_Mods *smods4);
 static void mods_remove_nsuniqueid(Slapi_Mods *smods);
+static int32_t dsentrydn_modrdn_update(backend *be, const char *olddn, const char *newdn, struct backentry *e, struct backentry **ec, ID id, back_txn *txn);
+static int dsentrydn_moddn_rename(back_txn *ptxn, backend *be, ID id, IDList *children, const Slapi_DN *dn_parentdn, const Slapi_DN *dn_newsuperiordn, struct backentry *child_entries[]);
```
droideck, line 27: Do we need also to declare `dsentrydn_moddn_rename_child` function here?

dirsrvtests/tests/suites/basic/ds_entrydn_test.py
```diff
@@ -0,0 +1,64 @@
+import logging
```
droideck, line 9: Probably, we need a copyright here. :)

droideck (approved):
Besides the minor issues, looks good to me.

ldap/servers/slapd/slap.h
```diff
@@ -2345,6 +2346,7 @@ typedef struct _slapdEntryPoints
 #define CONFIG_LISTEN_BACKLOG_SIZE "nsslapd-listen-backlog-size"
 #define CONFIG_DYNAMIC_PLUGINS "nsslapd-dynamic-plugins"
 #define CONFIG_RETURN_DEFAULT_OPATTR "nsslapd-return-default-opattr"
+#define CONFIG_RETURN_ENTRY_DN "nsslapd-return-original-entrydn"
```
droideck, line 2348: Oh, and could you please also add a simple test case for this setting?

Firstyear:
Just to check is this going to disable subtree rename by default? It's a pretty rare operation IMO so I'd be okay with us optimising around the general (and simpler/faster) cases.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
Firstyear: In a c file, functions can only be called if the declaration is "before" the first usage. Since dsentrydn_moddn_rename_child's full function impl is *before* it's called at all, there is no need for the declaration here. 

Remember a c file is "one pass" from top to bottom, and if a "name" isn't defined it's an error. So the following would fail.

```

void foo() {
    bar()
}

void bar() {
}
```

Because as void foo is read, bar() isn't defined yet. These both below would work.

```
void bar(); 

void foo() {
    bar()
}

void bar() {
}
```

```
void bar() {
}

void foo() {
    bar()
}
```

Hope that helps :)

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    char **olddns;
+    int olddncomps = 0;
+    int need = 1; /* For the '\0' */
+    size_t i;
+
+    olddn = slapi_entry_attr_get_charptr((*ec)->ep_entry, SLAPI_ATTR_DS_ENTRYDN);
+    if (NULL == olddn) {
+        return retval;
+    }
+    olddns = slapi_ldap_explode_dn(olddn, 0);
```
Firstyear, line 1589: 💥

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    /*
+     * Construct the new DN for the entry by taking the old DN
+     * excluding the old parent entry DN, and adding the new
+     * superior entry DN.
+     */
+    int retval = 0;
+    char *olddn;
+    char *newdn;
+    char **olddns;
+    int olddncomps = 0;
```
Firstyear: size_t due to use as array offset in for loops

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    olddns = slapi_ldap_explode_dn(olddn, 0);
+    if (NULL == olddns) {
+        goto out;
+    }
+    for (; olddns[olddncomps] != NULL; olddncomps++);
+
+    for (i = 0; i < olddncomps - parentdncomps; i++) {
+        need += strlen(olddns[i]) + 1; /* For the "," */
+    }
+    for (i = 0; i < newsuperiordncomps; i++) {
```
Firstyear, line 1598: any reason not to scope the size_t i = 0; here per loop rather than once outside? IIRC there is some compiler optimisations that can occur if you scope it per loop rather than once and try to reuse it.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+        need += strlen(newsuperiordns[i]) + 1; /* For the " " */
+    }
+    need--; /* We don't have a comma on the end of the last component */
+    newdn = slapi_ch_malloc(need);
```
Firstyear, line 1602: calloc not malloc.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    struct backentry *child_entries[])
+{
+    /* Iterate over the children list renaming every child */
+    struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    ldbm_instance *inst = (ldbm_instance *)be->be_instance_info;
+    struct backentry **child_entry_copies = NULL;
+    int retval = -1;
+    char **newsuperiordns = NULL;
+    int newsuperiordncomps = 0;
+    int parentdncomps = 0;
```
Firstyear: size_t

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,204 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    }
+
+    /*
+     * Iterate over the child entries renaming them.
+     */
+    child_entry_copies = (struct backentry **)slapi_ch_calloc(sizeof(struct backentry *), nids + 1);
+    for (i = 0; i <= nids; i++) {
+        child_entry_copies[i] = backentry_dup(child_entries[i]);
+    }
+    for (i = 0, retval = 0; retval == 0 && child_entries[i]; i++) {
```
Firstyear: similar loop bound optimisation here for size_t i

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: ??  This explodes?

mreynolds389:
@Firstyear ok so in the modrdn code it is adapted from working code (dsentrydn_* functions mirrored moddn functions).  I'll have to go through it all.  All my ASAN tests are good though, but I will clean this up.  Just give me some time...

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
Firstyear: It's a joke comment. ldap_explode_dn -> 💥

Firstyear:
No stress mate, beside those minor nits, I think it's a good idea and a good change :)

mreynolds389:
> Just to check is this going to disable subtree rename by default? It's a pretty rare operation IMO so I'd be okay with us optimising around the general (and simpler/faster) cases.

No, this does not affect subtree rename.  You can still move DN's between subtree's (hence all that extra modrdn code), but we do have all this code "switching" around entryrdn_get_switch().  With this patch we can eliminate the need for that old code, and we can stick with entryrdn.   The only customers who had wanted this option ("nsslapd-subtree-rename-switch") were ones who needed to preserve the DN case (AFAIK)

So this patch allows us to get a perf boost, and optionally preserve the case of DN's (which has been a pain point with some migrations to 389).

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: > It's a joke comment. ldap_explode_dn -> boom

I really appreciate that!  Just wasn't expecting it haha

Firstyear: Sometimes I'm funny :)

dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
```diff
@@ -656,7 +656,7 @@ def test_internal_log_level_516(topology_st, add_user_log_level_516, disable_acc
                                     r'SRCH base="cn=group,ou=Groups,dc=example,dc=com".*')
     # (Internal) op=10(1)(2) ENTRY dn="cn=group,ou=Groups,dc=example,dc=com"
     assert topo.ds_access_log.match(r'.*\(Internal\) op=[0-9]+\([0-9]+\)\([0-9]+\) '
-                                    r'ENTRY dn="cn=group,ou=groups,dc=example,dc=com".*')
+                                    r'ENTRY dn="cn=group,ou=Groups,dc=example,dc=com".*')
```
progier389, line 659: I would rather use match(..., flags=re.IGNORECASE)
because the test are sometime run against an older DS version

progier389 (approved):
LGTM except a few minor comments about  py tests

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -566,8 +569,8 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
              * instead.  Otherwise, a ModRDN operation will potentially change an
              * entry's entire DN (at least with respect to case and spacing).
              */
-            slapi_sdn_copy(slapi_entry_get_sdn_const(e->ep_entry), sdn);
-            slapi_pblock_set(pb, SLAPI_MODRDN_TARGET_SDN, sdn);
+            //slapi_sdn_copy(slapi_entry_get_sdn_const(e->ep_entry), sdn); // MARK This resets new superior case
+            //slapi_pblock_set(pb, SLAPI_MODRDN_TARGET_SDN, sdn);
```
mreynolds389: @progier389 @tbordaz  Ok so please look at this code and the long comment about changing the case.  I am sort of breaking what the code was originally trying to do.  In my case I want the new superior components (the ones that are not changing) to match the case of the original entry.  So resetting the superior would break what I am trying to do (preserver the case of the entire DN where possible).  Anyway I am hesitant to change the behavior, but I need to in order to conform with what I am trying to do.   Perhaps I am not seeing the potential risk of changing this behavior?  

Let me know what you guys think about this code/comment above...

dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
mreynolds389: This doesn't work since lib389 is overloading the match function.  It doesn't accept any other positional args.  So it would require a change to lib389 to get this to work, and that should be done in a separate ticket IMHO.

mreynolds389:
Just found out the config setting has no effect.  The DN case is always preserved in the entry.  This could cause issues that expect it to follow the old behavior.  Will investigate...

mreynolds389:
> Just found out the config setting has no effect. The DN case is always preserved in the entry. This could cause issues where client expects it to follow the old behavior. Will investigate...

This is fixed now, but it means the performance improvement will only happen if we configure the exact case to be returned.  I was hoping the perf improvement would work regardless of how we return the DN to the client, but that's just not how it works.  So if we want the perf improvement by default then the DN case behavior will change by default.  We can discuss this later...

ldap/servers/slapd/slap.h
mreynolds389: done!

ldap/schema/01core389.ldif
```diff
@@ -329,6 +329,8 @@ attributeTypes: ( 2.16.840.1.113730.3.1.2374 NAME 'nsDS5ReplicaBootstrapTranspor
 attributeTypes: ( 2.16.840.1.113730.3.1.2387 NAME 'nsslapd-tcp-fin-timeout' DESC 'Netscape defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN 'Netscape Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2388 NAME 'nsslapd-tcp-keepalive-time' DESC 'Netscape defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN 'Netscape Directory Server' )
 attributeTypes: ( 2.16.840.1.113730.3.1.2390 NAME 'nsds5ReplicaKeepAliveUpdateInterval' DESC '389 defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 SINGLE-VALUE X-ORIGIN '389 Directory Server' )
+attributeTypes: ( 2.16.840.1.113730.3.1.9998 NAME 'dsEntryDN' DESC '389 Directory Server defined attribute type' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 NO-USER-MODIFICATION SINGLE-VALUE USAGE directoryOperation X-ORIGIN '389 Directory Server' )
```
tbordaz: Should not it be distinguished name syntax 1.3.6.1.4.1.1466.115.121.1.12

mreynolds389: it should

ldap/servers/slapd/back-ldbm/id2entry.c
```diff
@@ -350,21 +355,28 @@ id2entry(backend *be, ID id, back_txn *txn, int *err)
...
-                                  rdn, id);
-                    /* Try rdn as dn. Could be RUV. */
-                    normdn = slapi_ch_strdup(rdn);
-                } else if (NULL == normdn) {
-                    slapi_log_err(SLAPI_LOG_ERR, ID2ENTRY,
-                                  "id2entry( %lu ) entryrdn_lookup_dn returned NULL. "
-                                  "Index file may be deleted or corrupted.\n",
-                                  (u_long)id);
-                    goto bail;
+                if (config_get_return_orig_dn() &&
```
tbordaz, line 357: I fail to understand when we will run those lines :(
Few lines above there is a test that the entry (in id2entry.db) contains "rdn". (get_value_from_string). This 'rdn' attribute exists in all entries (before your fix) correct ? So during migration (before the entry is updated) we will never run those lines correct ?
During an update, is the RDN attribute written ? I missed in the patch where "rdn" is/was added ?

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,200 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+
+    /*
+     * Iterate over the child entries renaming them.
+     */
+    child_entry_copies = (struct backentry **)slapi_ch_calloc(sizeof(struct backentry *), nids + 1);
+    for (size_t i = 0; i <= nids; i++) {
+        child_entry_copies[i] = backentry_dup(child_entries[i]);
+    }
+    for (size_t i = 0; retval == 0 && child_entries[i]; i++) {
+        retval = dsentrydn_moddn_rename_child(ptxn, be, li, child_entries[i], &child_entry_copies[i],
```
tbordaz, line 1679: This function is similar to moddn_rename_children except that it does not set opcsn to the updated child. is it intentional ?

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,200 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    slapi_ch_free((void **)&child_entry_copies);
+    slapi_ldap_value_free(newsuperiordns);
+
+    return retval;
+}
+
+static int32_t
+dsentrydn_modrdn_update(
+    backend *be,
+    const char *olddn,
```
tbordaz: oldrdn is not used

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,200 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    back_txn *txn)
+{
+    int32_t ret = 0;
+    int32_t cache_rc = 0;
+    ldbm_instance *inst = (ldbm_instance *)be->be_instance_info;
+
+    /* We have the entry, so update id2entry */
+    slapi_entry_attr_set_charptr((*ec)->ep_entry, SLAPI_ATTR_DS_ENTRYDN, newdn);
+    slapi_entry_set_dn((*ec)->ep_entry, (char *)newdn);
+    ret = id2entry_add_ext(be, *ec, txn, 1, &cache_rc);
```
tbordaz, line 1717: Is it updating the entryrdn index ? Should not entryrdn be updated ?

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,200 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+}
+
+static int32_t
+dsentrydn_modrdn_update(
+    backend *be,
+    const char *olddn,
+    const char *newdn,
+    struct backentry *e,
+    struct backentry **ec,
+    ID id,
```
tbordaz: I think id is useless

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,200 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+                      "id2entry_add deadlock\n");
+        goto out;
+    }
+    if (ret != 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "modrdn_rename_entry_update_indexes",
+                      "id2entry_add failed, err=%d\n",
+                      ret);
+        goto out;
+    }
+    id = e->ep_id;
```
tbordaz: If this function should return 'id' then it should be '*id = e->ep_id". But I have the feeling 'id' is a leftover of intermediate version and could be removed.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: So I adapted the old entyrdn code to work with the new attribute "dsEntryDN"  So the dsentry* functions are all modified in this way.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: No, the dsentry* functions are only meant to work on the "dsEntryDN" attribute (not entryrdn).  All the entryrdn functions are still called.

ldap/servers/slapd/back-ldbm/id2entry.c
mreynolds389: Your questions have nothing to do with my patch :-), but we write the rdn when the entry is added to id2entry (entry2str* functions).

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: you're right

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: correct

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: Yup, cleaned it up

mreynolds389:
So I did a test with MODDN of a subtree with 100K entries.  There is a performance drop, as expected, because we need to update "dsEntryDN" in all the children.  I see 35% drop in performance.  100K subtree move goes from 12 seconds to 19 seconds.  While this is a big drop, subtree moves should be very rare, and IMHO the gain we get in searches is worth it.

ldap/servers/slapd/result.c
```diff
@@ -1571,9 +1572,13 @@ send_ldap_search_entry_ext(
         goto cleanup;
     }
 
-    rc = ber_printf(ber, "{it{s{", operation->o_msgid,
-                    LDAP_RES_SEARCH_ENTRY, slapi_entry_get_dn_const(e));
-
+    if (return_entrydn && e->e_dn) {
+        rc = ber_printf(ber, "{it{s{", operation->o_msgid,
+                        LDAP_RES_SEARCH_ENTRY, e->e_dn);
```
tbordaz: I have a mix feeling about that new slapi_entry field e_dn. It is a reference into another field e_sdn and I feel risky to get two references. I doubt the e_sdn could be freed while still accessing e_dn but who knows.
I would prefer keep using slapi_entry_get_dn_const(e).

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
tbordaz: My concern here is that the children DN are updated during the MODRDN_1. If another MODRDN_2 comes in (from replication), I would expect URP to decide which DN is the good one. If the DN or RDN or maxcsn have not been updated then the last MODRDN will systematically win.
Anyway this it not related to your code, it has been like this for years

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1557,6 +1566,197 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
+    }
+    for (size_t i = 0; retval == 0 && child_entries[i]; i++) {
+        retval = dsentrydn_moddn_rename_child(ptxn, be, li, child_entries[i], &child_entry_copies[i],
+                                              parentdncomps, newsuperiordns,
+                                              newsuperiordncomps);
+    }
+
+    if (0 == retval) {
+        for (size_t i = 0; child_entries[i]; i++) {
+            CACHE_REMOVE(&inst->inst_cache, child_entry_copies[i]);
```
tbordaz, line 1686: On success, is not it surprising to remove the copied entry from the hash rather original entry ?

tbordaz (commented):
The code looks good to me.
two minor questions

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
mreynolds389: :-)  I tried playing around with this code and every time I did it caused heap-use-after-free or memory leaks, or other odd errors.  Yeah I thought it seemed odd, yet it seems this is exactly what it needs to be.

ldap/servers/slapd/result.c
mreynolds389: yeah it's not needed, and removing it allowed me to clean up more code.  Thanks!

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
tbordaz: Thanks for the testing. I would say it is part of the entry cache magic :)

tbordaz (approved):
LGTM

mreynolds389:
Last question @progier389 @tbordaz   Do we enable this by default?  Support thinks we should, but what about us?

progier389:
Since it has a positive impact on the search performance my feeling is that we should better enable it by default

Firstyear:
@tbordaz So long as the "upgrade" is clean and doesn't require user intervention. Remember we don't do update-ds.pl anymore, so this has to work to upgrade from version X to Y without any user intervention.

mreynolds389:
> @tbordaz So long as the "upgrade" is clean and doesn't require user intervention. Remember we don't do update-ds.pl anymore, so this has to work to upgrade from version X to Y without any user intervention.

Well to fully utilize it you need to reimport the database, but nothing will break if the new attribute it not present in the entry.  What you can NOT do is downgrade as the schema will be different but we don't support downgrades anyway.

Firstyear:
Okay, so long as it "just works" on upgrades, then that's fine by me. And yep, no downgrade is no issue IMO.

tbordaz:
> Last question @progier389 @tbordaz Do we enable this by default? Support thinks we should, but what about us?

My only concern is about the change of returned DN. All current 389ds clients are use to receive transformed DN (normalized) the risk would be that some clients could rely on that normalization. If that is the case, then they can disable the feature. Assuming the feature ran for some time and some entries do contain dsentryDN, if we disable the feature it will return the normalized DN (from entryrdn) ignoring dsentryDN. Am I correct ?
If yes, I also agree it is good to enable it by default.

mreynolds389:
> > Last question @progier389 @tbordaz Do we enable this by default? Support thinks we should, but what about us?
> 
> My only concern is about the change of returned DN. All current 389ds clients are use to receive transformed DN (normalized) the risk would be that some clients could rely on that normalization. 

This was my concern as well...

> If that is the case, then they can disable the feature. Assuming the feature ran for some time and some entries do contain dsentryDN, if we disable the feature it will return the normalized DN (from entryrdn) ignoring dsentryDN. Am I correct ? If yes, I also agree it is good to enable it by default.

Correct, if the feature is disabled then the entyrdn index is used to generate the DN (See the change in id2entry.c)

Issue #5339, opened 2022-06-08
Fix GitHib CI 'cannot find -lldap_r' issue


droideck:
**Issue Description**
The error 'cannot find -lldap_r' happens when GitHub CI is run on a non-master branch which doesn't have the https://github.com/389ds/389-ds-base/commit/a3375a08a506c3b72a8a23df5238848017691350 fix.

The issue happens if you build older branches on Fedora 36 (where we have OpenLDAP 2.6).

- Building 2.1 on F36 (where we have the libldap 2.6) - is okay.
- And Building 2.0, 1.4.4 and 1.4.3 on F35 should be fine, too, as it has libldap_r.

We should be careful with backporting the fix as we may find ourselves in the "dangerous" territory when we build 389-ds-base with libdap from openldap-2.6, but it will be used with openldap-2.4.

We shouldn't have issues on Fedora 35, but we may get issues on RHEL 8 or SUSE if they don't have libldap/libldap_r fix from https://fedoraproject.org/wiki/Changes/OpenLDAPwithoutNonthreadedLibraries

So we should always build with libldap_r the package that should be used on a system where we use libldap_r.

So we may need to find another way to fix the github CI. Simple cherry-pick to older branches may cause problems because of the issue described above.

droideck:
Also, it may not be worth the hustle as we don't usually run CI on anything besides master. The rare cases we need it, we can do it manually and check it on the correct OS (RHEL, SUSE, older Fedora).

progier389:
> Also, it may not be worth the hustle as we don't usually run CI on anything besides master. 

@droideck, I am afraid that you are wrong on that point: 
I received the test report failures after having cherry-picked a fix and pushed it on the github repository.
As I understand things the tests are triggered automatically through the github actions defined in
https://github.com/389ds/389-ds-base/tree/master/.github/workflows when a commit is merged or pushed on any branches ...

droideck:
> > Also, it may not be worth the hustle as we don't usually run CI on anything besides master.
> 
> @droideck, I am afraid that you are wrong on that point: I received the test report failures after having cherry-picked a fix and pushed it on the github repository. As I understand things the tests are triggered automatically through the github actions defined in https://github.com/389ds/389-ds-base/tree/master/.github/workflows when a commit is merged or pushed on any branches ...

Sorry, mate, I wasn't clear enough on my point.
I meant that our GitHub CI is built for testing the main branch.
[...]

PR #4457, opened 2020-11-24
Issue 4446 RFE - openldap password hashers

Firstyear:
Bug Description: To allow easier migrations, we need to support
some password types that OpenLDAP has that we do not. This
is mainly PBKDF2 types. OpenLDAP's hashers are
based on python passlib, so they also store the values in
a different way than our PBDKF2 module.

Fix Description: This adds passlib style PBKDF2 support, written
in Rust. It extends the slapi_r_plugin shim to support password
extensions for Rust plugins, as well as providing a number of
small improvements to the build system and testing for rust
plugins.

fixes: #4446 

Author: William Brown <william@blackhats.net.au>

Review by: ???

PR #5499, opened 2022-10-20
Issue 5507 - Fix 1.4.3 GitHub PR CI

droideck:
**Description**
Our CI fails in many places because of the missing backports and Fedora package changes.

Backport the missing peaces and fix CI images.
Update package-lock.json
Add Requires: acl.
Remove failing perl_enable invocation.
Don't set max age if changelog doesn't exist.
Fix dscreate, ssl_test, ssl_version, healcheck, fourwaymmr tests

Fixes: https://github.com/389ds/389-ds-base/issues/5339
Fixes: https://github.com/389ds/389-ds-base/issues/5507

Reviewed by: ?


dirsrvtests/tests/suites/healthcheck/healthcheck_test.py
```diff
@@ -65,11 +58,10 @@ def run_healthcheck_and_flush_log(topology, instance, searched_code=None, json=F
...
 
 def set_changelog_trimming(instance):
-    log.info('Get the changelog enteries')
-    inst_changelog = Changelog5(instance)
-
     log.info('Set nsslapd-changelogmaxage to 30d')
-    inst_changelog.add('nsslapd-changelogmaxage', '30')
+
+    cl = Changelog5(instance)
+    cl.replace('nsslapd-changelogmaxage', '30')
```
mreynolds389: I think this is supposed to be "30d"

mreynolds389:
@droideck I think you're missing changes to fedse.c and template-dse.ldif.in.  I think there should be the new pwdchan password plugins listed in these two files.

droideck:
> @droideck I think you're missing changes to fedse.c and template-dse.ldif.in. I think there should be the new pwdchan password plugins listed in these two files.

I've reverted the "PBKDF2-SHA512 by default" change for now because it requires a massive Rust backport.

One thing is not fully clear to me. Do we want to make Rust in 1.4.3 mandatory?
If we add PBKDF2-SHA512 functionality with this [PR](https://github.com/389ds/389-ds-base/pull/4457/files), we'll have to always build with Rust enabled. So I want to clarify if it's our intention...

mreynolds389:
> > @droideck I think you're missing changes to fedse.c and template-dse.ldif.in. I think there should be the new pwdchan password plugins listed in these two files.
> 
> I've reverted the "PBKDF2-SHA512 by default" change for now because it requires a massive Rust backport.
> 
> One thing is not fully clear to me. Do we want to make Rust in 1.4.3 mandatory? If we add PBKDF2-SHA512 functionality with this [PR](https://github.com/389ds/389-ds-base/pull/4457/files), we'll have to always build with Rust enabled. So I want to clarify if it's our intention...

1.4.3. is only used for RHEL builds now, so it is fine if it's RUST by default.  I thought this is what you were doing actually.  If it's still a lot of work then it can be skipped for now.

dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py
```diff
@@ -439,13 +450,16 @@ def test_replica_backup_and_restore(topo_m4):
         5. Should success
     """
     # Testing bug #830335: Taking a replica backup and Restore on M1 after deleting few entries from M1 nad M2
+    # Make sure to update S1 to avoid useless replication delay
```
progier389, line 453: I would not change the supplier in that test. : We do not have lot of 4 supplier tests and IMHO it is a good idea to have a test in which the changes are not issued on supplier 1 .
IMHO we should rather keep creating the user on supplier 3 but wait 
repl.wait_for_replication(topo_m4.ms["supplier3"], topo_m4.ms["supplier1"]) before waiting for supplier1 -> supplier2

droideck: Closing the conversation for now, as we've discussed in the meeting that this part of the PR is a pure backport from the main. And we can change it through the usual PR process main->backports in the future.
And probably, we should, as fourwaymmr test suite still has its issues and can fail from time to time.

progier389 (approved):
Better have Viktor's feeling about the .github subtree changes,
But with the clarification we got in the meeting, that is OK on my side.

vashirov (approved):
I'm happy with the changes, thanks for syncing GH actions workflows!

Issue #5502, opened 2022-10-27
RFE - Allow specifying entry attributes to include in audit log


mreynolds389:
There are use cases where having more fine grained identifying attributes in the audit log entry is useful.

    nsslapd-auditlog-display-attrs: cn <attribute> <attribute> <"*" for all attributes>

```
time: 20221014125914
dn: uid=demo_user,ou=people,dc=example,dc=com
#cn: demo user
result: 0
changetype: modify
replace: displayName
displayName: Demo Entry
-
replace: modifiersname
modifiersname: cn=dm
-
replace: modifytimestamp
modifytimestamp: 20221014165914Z
```

So the displayed attributes will be prefixed with a "#" to effectively make them comments so parsing tools will not have an issue replaying the audit log.

mreynolds389:
c1ec31b2b..e495cad6c  389-ds-base-2.2 -> 389-ds-base-2.2
3d4f6751a..3e027d7a0  389-ds-base-2.1 -> 389-ds-base-2.1
f06a9b59f..984b286bf  389-ds-base-2.0 -> 389-ds-base-2.0
25ec767f1..ac351252f  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2136610

PR #5503, opened 2022-10-27
Issue 5502 - RFE - Add option to display entry attributes in audit log

mreynolds389:
Description:

Add a new config setting to specify specifc, or all, attributes from the entry being updated.  These attributes are prefixed with a '#' to make them comments so parsing tools will still work.

relates: https://github.com/389ds/389-ds-base/issues/5502

### DIFF
--- a/dirsrvtests/tests/suites/ds_logs/audit_log_test.py
+++ b/dirsrvtests/tests/suites/ds_logs/audit_log_test.py
@@ -1,0 +1,104 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import logging
+import pytest
+import os
+import time
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.topologies import topology_st as topo
+from lib389.idm.user import UserAccounts
+
+log = logging.getLogger(__name__)
+
+
+def test_auditlog_display_attrs(topo):
+    """Test "display attributes" feature of the audit log
+
+    :id: 01beaf71-4cb5-4943-9774-3210ae5d68a2
+    :setup: Standalone Instance
+    :steps:
+        1. Test "cn" attribute is displayed
+        2. Test multiple attributes are displayed
+        3. Test modrdn updates log
+        4. Test all attributes are displayed
+        5. Test delete updates log
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Sucecss
+    """
+
+    inst = topo.standalone
+    inst.config.replace('nsslapd-auditlog-logging-enabled', 'on')
+
+    # Test "cn" attribute
+    inst.config.replace('nsslapd-auditlog-display-attrs', 'cn')
+    users = UserAccounts(inst, DEFAULT_SUFFIX)
+    user = users.ensure_state(properties={
+        'uid': 'test_audit_log',
+        'cn': 'test',
+        'sn': 'user',
+        'uidNumber': '1000',
+        'gidNumber': '1000',
+        'homeDirectory': '/home/test',
+        'userPassword': 'pppppppp'
+    })
+    user2 = users.ensure_state(properties={
+        'uid': 'test_modrdn_delete',
+        'cn': 'modrdn_delete',
+        'sn': 'modrdn_delete',
+        'uidNumber': '1001',
+        'gidNumber': '1001',
+        'homeDirectory': '/home/modrdn_delete',
+        'userPassword': 'pppppppp'
+    })
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#cn: test")
+    assert not inst.ds_audit_log.match("#uid: test_audit_log")
+
+    # Test multiple attributes
+    inst.config.replace('nsslapd-auditlog-display-attrs', 'uidNumber gidNumber, homeDirectory')
+    user.replace('sn', 'new value')
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#uidNumber: 1000")
+    assert inst.ds_audit_log.match("#gidNumber: 1000")
+    assert inst.ds_audit_log.match("#homeDirectory: /home/test")
+    assert not inst.ds_audit_log.match("#uid: test_audit_log")
+    assert not inst.ds_audit_log.match("#uidNumber: 1001")
+    assert not inst.ds_audit_log.match("#sn: modrdn_delete")
+
+    # Test modrdn
+    user2.rename("uid=modrdn_delete", DEFAULT_SUFFIX)
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#uidNumber: 1001")
+    assert inst.ds_audit_log.match("#gidNumber: 1001")
+
+    # Test ALL attributes
+    inst.config.replace('nsslapd-auditlog-display-attrs', '*')
+    user.replace('sn', 'new value again')
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#uid: test_audit_log")
+    assert inst.ds_audit_log.match("#cn: test")
+    assert inst.ds_audit_log.match("#uidNumber: 1000")
+    assert inst.ds_audit_log.match("#objectClass: top")
+
+    # Test delete
+    user2.delete()
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#sn: modrdn_delete")
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -33,7 +33,7 @@ static int audit_hide_unhashed_pw = 1;
 static int auditfail_hide_unhashed_pw = 1;
 
 /* Forward Declarations */
-static void write_audit_file(int logtype, int optype, const char *dn, void *change, int flag, time_t curtime, int rc, int sourcelog);
+static void write_audit_file(Slapi_Entry *entry, int logtype, int optype, const char *dn, void *change, int flag, time_t curtime, int rc, int sourcelog);
 
 static const char *modrdn_changes[4];
 
@@ -46,6 +46,7 @@ write_audit_log_entry(Slapi_PBlock *pb)
     void *change;
     int flag = 0;
     Operation *op;
+    Slapi_Entry *entry = NULL;
 
     /* if the audit log is not enabled, just skip all of
        this stuff */
@@ -58,14 +59,17 @@ write_audit_log_entry(Slapi_PBlock *pb)
     switch (operation_get_type(op)) {
     case SLAPI_OPERATION_MODIFY:
         slapi_pblock_get(pb, SLAPI_MODIFY_MODS, &change);
+        slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &entry);
         break;
     case SLAPI_OPERATION_ADD:
         slapi_pblock_get(pb, SLAPI_ADD_ENTRY, &change);
+        slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &entry);
         break;
     case SLAPI_OPERATION_DELETE: {
         char *deleterDN = NULL;
         slapi_pblock_get(pb, SLAPI_REQUESTOR_DN, &deleterDN);
         change = deleterDN;
+        slapi_pblock_get(pb, SLAPI_ENTRY_PRE_OP, &entry);
     } break;
 
     case SLAPI_OPERATION_MODDN: {
@@ -86,6 +90,7 @@ write_audit_log_entry(Slapi_PBlock *pb)
             modrdn_changes[2] = NULL;
         }
         change = (void *)modrdn_changes;
+        slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &entry);
         break;
     }
     default:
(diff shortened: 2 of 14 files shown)


mreynolds389:
Design doc:

https://www.port389.org/docs/389ds/design/audit-log-entry-attrs-design.html

dirsrvtests/tests/suites/ds_logs/audit_log_test.py
```diff
@@ -0,0 +1,77 @@
...
+        'gidNumber': '1000',
+        'homeDirectory': '/home/test',
+        'userPassword': 'pppppppp'
+    })
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#cn: test")
+    assert not inst.ds_audit_log.match("#uid: test_audit_log")
+
+    # Test multiple attributes
+    inst.config.replace('nsslapd-auditlog-display-attrs', 'uidNumber gidNumber')
```
droideck: In the design doc, you've also mentioned that commas can be used as a delimiter.
Could you please add a test for two things: comma-separated attr list, and a mix case (something like "uidNumber gidNumber, cn") even though it sounds wrong?

ldap/servers/slapd/auditlog.c
```diff
@@ -163,15 +172,71 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+                addlenstr(l, val);
+                addlenstr(l, "\n");
+            }
+        }
+    }
+    slapi_ch_free_string(&display_attrs);
+}
 
 /*
  * Function: write_audit_file
```
droideck, line 277: Minor thing, `Slapi_Entry *entry` probably should be in the comment's description.

ldap/servers/slapd/auditlog.c
```diff
@@ -118,20 +124,22 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
         break;
     case SLAPI_OPERATION_ADD:
         slapi_pblock_get(pb, SLAPI_ADD_ENTRY, &change);
+        slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &entry);
         break;
     case SLAPI_OPERATION_DELETE: {
         char *deleterDN = NULL;
         slapi_pblock_get(pb, SLAPI_REQUESTOR_DN, &deleterDN);
         change = deleterDN;
+        slapi_pblock_get(pb, SLAPI_ENTRY_PRE_OP, &entry);
```
droideck: If I try to restart after the test case, it crashes for me:

```
Stack trace of thread 59926:
#0  0x00007ffa2eaed87f write_audit_file (libslapd.so.0 + 0xed87f)
#1  0x00007ffa2eaee3e0 write_auditfail_log_entry (libslapd.so.0 + 0xee3e0)
#2  0x00007ffa2eafea8c op_shared_delete (libslapd.so.0 + 0xfea8c)
#3  0x00007ffa2eafed9d delete_internal_pb (libslapd.so.0 + 0xfed9d)
#4  0x00007ffa2eb7fbc4 upgrade_server (libslapd.so.0 + 0x17fbc4)
#5  0x0000562372ae9927 main (ns-slapd + 0x14927)
#6  0x00007ffa2e629510 __libc_start_call_main (libc.so.6 + 0x29510)
#7  0x00007ffa2e6295c9 __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x295c9)
#8  0x0000562372aea945 _start (ns-slapd + 0x15945)
ELF object binary architecture: AMD x86-64
```

But looking at the stack trace, I think, it happens just because `nsslapd-auditlog-logging-enabled` is on during init. And exactly delete operation, odd...

dirsrvtests/tests/suites/ds_logs/audit_log_test.py
```diff
@@ -0,0 +1,77 @@
...
+    user.replace('sn', 'new value')
+    time.sleep(1)
+    assert inst.ds_audit_log.match("#uidNumber: 1000")
+    assert inst.ds_audit_log.match("#gidNumber: 1000")
+    assert not inst.ds_audit_log.match("#uid: test_audit_log")
+
+    # Test ALL attributes
+    inst.config.replace('nsslapd-auditlog-display-attrs', '*')
+    user.replace('sn', 'new value again')
```
droideck, line 86: Could you also, please, add more tests for 'user.remove' and 'user.add' also?

src/cockpit/389-console/src/lib/server/auditLog.jsx
```diff
@@ -447,6 +556,45 @@ export class ServerAuditLog extends React.Component {
...
                         </Form>
+                        <Form className="ds-margin-top-lg ds-margin-left" isHorizontal autoComplete="off">
+                            <FormGroup
+                                label="Display Attributes"
+                                fieldId="nsslapd-auditlog-display-attrs"
+                                title="Display attributes from the entry in the audit log (nsslapd-auditlog-display-attrs)."
+                            >
+                                <div className={this.state.displayAllAttrs ? "ds-hidden" : "ds-margin-bottom"}>
+                                    <Select
+                                        variant={SelectVariant.typeaheadMulti}
```
droideck, line 567: Hmm, it doesn't work as multi-valued for me... If I add two attributes and click `Save Log Setting` - it says `Success!` but after refreshing shows only one attribute.

droideck (changes requested):
Also, could you please clarify the design a bit?

I see that `auditfaillog` also is involved but the setting `nsslapd-auditlog-display-attrs` mentions only `auditlog`.

Does it mean that the setting propagates over auditfaillog all-together when set to auditlog?

ldap/servers/slapd/auditlog.c
```diff
@@ -163,15 +172,71 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+    }
+
+    if (strcmp(display_attrs, "*")) {
+        /* Return specific attributes */
+        for (req_attr = ldap_utf8strtok_r(display_attrs, ", ", &last); req_attr;
+             req_attr = ldap_utf8strtok_r(NULL, ", ", &last))
+        {
+            char **vals = slapi_entry_attr_get_charray(entry, req_attr);
+            for(size_t i = 0; vals && vals[i]; i++) {
+                addlenstr(l, "#");
```
tbordaz, line 225: Should also skip unhashedpassword. What about 'userpassword' ?

ldap/servers/slapd/auditlog.c
```diff
@@ -163,15 +172,71 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+            if (strcmp(attr, PSEUDO_ATTR_UNHASHEDUSERPASSWORD) == 0) {
+                /* Do not write the unhashed password */
+                continue;
+            }
+            for(size_t i = 0; vals && vals[i]; i++) {
+                val = slapi_value_get_string(vals[i]);
+                addlenstr(l, "#");
+                addlenstr(l, attr);
+                addlenstr(l, ": ");
+                addlenstr(l, val);
```
tbordaz: I think it would be good to set a limit of size. It makes no real sense to store big attributes in audit log. i.e. 256 chars ?

tbordaz (commented):
The code looks good. minor comments

ldap/servers/slapd/auditlog.c
mreynolds389: @droideck I'm not sure I follow what you asking?  Do you mean describe the two function parameters?  I'll do that sure, but I'm not sure if that's what you asking :-)

ldap/servers/slapd/auditlog.c
mreynolds389: @tbordaz We don't ever hide userpassword from the auditlog, so didn't see it as important to remove it from the entry attributes.  I can remove it, and I did think about it, but I didn't see it as necessary.  Thoughts?

ldap/servers/slapd/auditlog.c
mreynolds389: @tbordaz yes I like limiting the size of the value

ldap/servers/slapd/auditlog.c
mreynolds389: > Also, could you please clarify the design a bit?
> 
> I see that `auditfaillog` also is involved but the setting `nsslapd-auditlog-display-attrs` mentions only `auditlog`.
> 
> Does it mean that the setting propagates over auditfaillog all-together when set to auditlog?

In the design doc, the first sentence, it does say this is for both the audit and auditfail log.  I'll will try and make this more obvious that the config setting is for both audit logs.

ldap/servers/slapd/auditlog.c
tbordaz: Well this comment was to skip unhashedpassword when configured attribute are '*'

Regarding others attributes,  it is quite common that administrators do not want to see sensitive data in the logs (including audit). That restrict a lot the use of '*'. Another possibility is to use a syntax like ciphers, with '-' something like " *,-userpassword, -certificate". But it looks a bit overkill.

ldap/servers/slapd/auditlog.c
mreynolds389: > If I try to restart after the test case, it crashes for me:
> 
> ```
> Stack trace of thread 59926:
> #0  0x00007ffa2eaed87f write_audit_file (libslapd.so.0 + 0xed87f)
> #1  0x00007ffa2eaee3e0 write_auditfail_log_entry (libslapd.so.0 + 0xee3e0)
> #2  0x00007ffa2eafea8c op_shared_delete (libslapd.so.0 + 0xfea8c)
> #3  0x00007ffa2eafed9d delete_internal_pb (libslapd.so.0 + 0xfed9d)
> #4  0x00007ffa2eb7fbc4 upgrade_server (libslapd.so.0 + 0x17fbc4)
> #5  0x0000562372ae9927 main (ns-slapd + 0x14927)
> #6  0x00007ffa2e629510 __libc_start_call_main (libc.so.6 + 0x29510)
> #7  0x00007ffa2e6295c9 __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x295c9)
> #8  0x0000562372aea945 _start (ns-slapd + 0x15945)
> ELF object binary architecture: AMD x86-64
> ```
> 
> But looking at the stack trace, I think, it happens just because `nsslapd-auditlog-logging-enabled` is on during init. And exactly delete operation, odd...

I can not reproduce the crash....

dirsrvtests/tests/suites/ds_logs/audit_log_test.py
mreynolds389: > Could you also, please, add more tests for 'user.remove' and 'user.add' also?
 
There is a user add already, but I can add a delete and modrdn...

dirsrvtests/tests/suites/ds_logs/audit_log_test.py
mreynolds389: @droideck - ah python's split() doesn't work as I thought it did.  So I just removed the comma option.  So the attribute list must be space separated only.

ldap/servers/slapd/auditlog.c
mreynolds389: I think that is overkill, and in fact the "\*" option might be overkill as well, but Anton thought it was a good idea.  I don't disagree with him, but if you want to see everything, then we should show everything.  However, I don't see that option "\*" being used often, as the main use case is displaying a single attribute to help identify the entry.

mreynolds389:
Everything should be fixed except for the crash (that I can not reproduce), and the question about whether or not to display userpassword.  One option for the password would just be to simply mask the value "******"

ldap/servers/slapd/auditlog.c
mreynolds389: Reproduced crash, you need auditfail log enabled :-)  Investigating...

ldap/servers/slapd/auditlog.c
droideck: Yep, something like:

```
/*
 * Function: write_audit_file
 * Arguments:
 *            entry - my awesome entry 🎉
 *            logtype - Destination where the message will go.
 *            optype - type of LDAP operation being logged
 *            dn     - distinguished name of entry being changed
```

tbordaz (approved):
I agree with your point, if an admin set ` * `as wished attributes we should log everything (except unhashedpassword. ACK (anticipating the fix for the crash)

mreynolds389:
> > In the design doc, the first sentence, it does say this is for both the audit and auditfail log. I'll will try and make this more obvious that the config setting is for both audit logs.
> 
> Oh, okay, got it! I was just confused a bit when in UI, I saw the `nsslapd-auditlog-display-attrs` field only in Auditlog menu. Does it make sense to duplicate it in the Auditfail log UI too? Or maybe, it will make sense to put a comment there that "You can configure nsslapd-auditlog-display-attrs values in Auditlog menu, and they will affect this log also"?
> 
> I'm okay either way, so it's up to you. :)

Ok found out it does not work at all in the auditfail log,so I removed all references to it (code anddesign doc).  

All issues have now been fixed!  @tbordaz @droideck please review...

ldap/servers/slapd/auditlog.c
```diff
@@ -156,22 +160,116 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+            if (strcmp(attr, PSEUDO_ATTR_UNHASHEDUSERPASSWORD) == 0) {
+                /* Do not write the unhashed clear-text password */
+                continue;
+            }
+
+            if (strcasecmp(attr, SLAPI_USERPWD_ATTR) == 0) {
+                /* userpassword - mask the value */
+                addlenstr(l, "#");
+                addlenstr(l, attr);
+                addlenstr(l, ": *********************************\n");
```
droideck: If we do `userPassword` this way, should we also process `nsslapd-rootpw` similarly? 

And just to confirm, is it by design that we don't replace the userPassword with `****` when userPassword is requested directly? (i.e. `nsslapd-auditlog-display-attrs: userPassword cn`)

mreynolds389: Yes if you are explicitly asking for an attribute then you get its value.  Doesn't make sense to ask for it and then the server hide the value.  

And yes I will add nsslapd-rootpw

droideck (approved):
Ack! Nicely done feature!

PR #5508, opened 2022-11-08
Issue 3729 - RFE Extend log of operations statistics in access log

tbordaz:
Bug description:
	Create a per operation framework to collect/display
	statistics about internal ressource consumption

Fix description:
	The fix contains 2 parts
	The framework registers a per operation object extension
        (op_stat_init). The extension is used to store/retrieve
	collected statistics.
        To reduce the impact of collecting/logging it uses a toggle
        with config attribute 'nsslapd-statlog-level' that is a bit mask.

	An exemple of collected data from the indexes during a search
	filterindex.c (store), retrieve/log (result.c). This path uses
	LDAP_STAT_READ_INDEX=0x1.

relates: #3729

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
+++ b/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
@@ -20,6 +20,7 @@ from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389._constants import DEFAULT_SUFFIX, LOG_ACCESS_LEVEL, PASSWORD
 from lib389.utils import ds_is_older, ds_is_newer
 from lib389.config import RSA
+from lib389.dseldif import DSEldif
 import ldap
 import glob
 import re
@@ -1143,7 +1144,7 @@ def test_enable_external_libs_debug_log(topology_st):
 @pytest.mark.skipif(ds_is_older('1.4.3'), reason="Might fail because of bug 1895460")
 @pytest.mark.bz1895460
 @pytest.mark.ds4593
-def test_cert_personality_log_help(topology_st):
+def test_cert_personality_log_help(topology_st, request):
     """Test changing the nsSSLPersonalitySSL attribute will raise help message in log
 
     :id: d6f17f64-d784-438e-89b6-8595bdf6defb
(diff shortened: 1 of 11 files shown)


ldap/servers/slapd/back-ldbm/filterindex.c
```diff
@@ -1054,13 +1054,55 @@ keys2idl(
     int allidslimit)
 {
     IDList *idl = NULL;
+    Op_stat *op_stat;
+    PRBool collect_stat;
 
     slapi_log_err(SLAPI_LOG_TRACE, "keys2idl", "=> type %s indextype %s\n", type, indextype);
+
+    /* Before reading the index take the start time */
+    op_stat = op_stat_get_operation_extension(pb);
```
progier389: Sould we get the op_stat if collect_stat is false ?

ldap/servers/slapd/result.c
```diff
@@ -1999,6 +1999,64 @@ notes2str(unsigned int notes, char *buf, size_t buflen)
...
+        case LDAP_REQ_MODRDN:
+        case LDAP_REQ_MODIFY:
+        case LDAP_REQ_COMPARE:
+            break;
+        case LDAP_REQ_SEARCH:
+            if ((LDAP_STAT_READ_INDEX & config_get_statlog_level()) &&
+                op_stat->search_stat) {
+                struct component_keys_lookup *key_info;
+                for (key_info = op_stat->search_stat->keys_lookup; key_info; key_info = key_info->next) {
+                    slapi_log_stat(LDAP_STAT_READ_INDEX,
```
progier389, line 2039: slapi_log_stat proto is missing in slapi-proto.h  file

ldap/servers/slapd/result.c
```diff
@@ -1999,6 +1999,64 @@ notes2str(unsigned int notes, char *buf, size_t buflen)
...
+log_op_stat(Slapi_PBlock *pb)
+{
+
+    Connection *conn = NULL;
+    Operation *op = NULL;
+    Op_stat *op_stat;
+    struct timespec duration;
+    char stat_etime[ETIME_BUFSIZ] = {0};
+
+    slapi_pblock_get(pb, SLAPI_CONNECTION, &conn);
```
progier389, line 2017: IMHO We should just return here if config_get_statlog_level is 0

progier389 (changes requested):
Sound globally good except a few issues that requires changes:
- Missing slapi_log_stat proto.
- Minimizing a bit more the impact when stat logging is disabled.

ldap/servers/slapd/back-ldbm/filterindex.c
tbordaz: I agree. I changed the patch

ldap/servers/slapd/result.c
tbordaz: done

ldap/servers/slapd/result.c
tbordaz: absolutely

tbordaz:
I changed the PR and added a testcase

ldap/servers/slapd/operation.c
```diff
@@ -652,6 +652,92 @@ slapi_operation_time_expiry(Slapi_Operation *o, time_t timeout, struct timespec
...
+        struct component_keys_lookup *keys, *next;
+
+        /* free all the individual key counter */
+        keys = op_statp->search_stat->keys_lookup;
+        while (keys) {
+            next = keys->next;
+            slapi_ch_free((void **) &keys->attribute_type);
+            slapi_ch_free((void **) &keys->key);
+            slapi_ch_free((void **) &keys->index_type);
+            slapi_ch_free((void **) &keys);
```
mreynolds389, line 716: Very minor, but could you use  slapi_ch_free_string() here.

ldap/servers/slapd/result.c
progier389: Weird there is still a warning ... Maybe you forgot to commit the slapi-plugin.h file ?

progier389 (approved):
You should fix the warning in log_op_stat but otherwise LGTM

dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
```diff
@@ -1175,6 +1175,73 @@ def test_cert_personality_log_help(topology_st):
...
                                          r"is set to '{}'\)\..*".format(WRONG_NICK))
 
+def test_stat_index(topology_st, request):
+    """Testing nsslapd-statlog-level with indexing statistics
+
+    :id: fcabab05-f000-468c-8eb4-02ce3c39c902
+    :setup: Standalone instance
+    :steps:
+         1. Check that nsslapd-statlog-level is 0 (default)
+         2. Create 20 users with 'cn' starting with 'user_'
```
mreynolds389: Docstring test is complaining about "user" in this message.  Maybe it doesn't like the single quote?  Not sure, but it should be fixed

vashirov: It doesn't like `user_` because in RST it is considered a reference notation, see https://docutils.sourceforge.io/docs/user/rst/quickref.html
Underscore in this case should be escaped with backslash, i.e. `user\_`.

PR #5511, opened 2022-11-08
Issue 5510 - Backport `twalk_r` from glibc

### LINKED ISSUE #5510: Backport `twalk_r` from glibc
**Issue Description**
Since version 2.1 we use `twalk_r` function that is available in glibc-2.30 and newer. This prevents building 389-ds on platforms with older glibc such as EL8.

We should backport it from https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/tsearch.c;h=852cadf7b93e880099a58111c4823df68a450ba4;hb=HEAD

vashirov:
Bug Description
Since version 2.1 we use `twalk_r` function that is available in glibc-2.30 and newer. This prevents building 389-ds on platforms with older glibc such as EL8.

Fix Description
Backport `twalk_r` from
https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/tsearch.c;h=852cadf7b93e880099a58111c4823df68a450ba4;hb=HEAD

Fixes: https://github.com/389ds/389-ds-base/issues/5510

Reviewed by:

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1252,6 +1252,7 @@ libback_ldbm_la_SOURCES = ldap/servers/slapd/back-ldbm/ancestorid.c \
 	ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c \
 	ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c \
 	ldap/servers/slapd/back-ldbm/db-bdb/bdb_import_threads.c \
+	ldap/servers/slapd/back-ldbm/db-mdb/twalk_r.c \
 	ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c \
 	ldap/servers/slapd/back-ldbm/db-mdb/mdb_debug.c \
 	ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c \
--- a/configure.ac
+++ b/configure.ac
@@ -79,7 +79,6 @@ AC_CHECK_FUNCS([endpwent ftruncate getcwd getaddrinfo inet_pton inet_ntop localt
 
 # These functions are *required* without option.
 AC_CHECK_FUNCS([clock_gettime], [], AC_MSG_ERROR([unable to locate required symbol clock_gettime]))
-AC_CHECK_FUNCS([twalk_r], [], AC_MSG_ERROR([cannot locate required symbol twalk_r: use glibc-2.30 or higher ]))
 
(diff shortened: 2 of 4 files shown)


progier389:
I do not much like the backport idea of having a module using opaque struct from withing glibc 
IMHO we should better get rid of the twalk_r call and use twalk and a global variable instead ...  
 ( Especially since the call is already protected by a global mutex so there will not be any race condition )

progier389:
FYI: Change would then be something like:

```
diff --git a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
index 56937efde..ba6154603 100644
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
@@ -90,6 +90,26 @@ typedef struct {
 static dbmdb_dbi_t *dbi_slots;    /* The alloced slots */
 static int dbi_nbslots;           /* Number of available slots in dbi_slots */
 
+/*
+ * twalk_r is not available before glibc-2.30 so lets replace it by twalk
+ * and a global variable (it is possible because there is a single call
+ *   and it is protected by global mutex dbmdb_ctx->dbis_lock )
+ */
+
+static struct {
+    void *closure;
+    void (*cb)(const void *nodep, VISIT which, void *closure);
+} twalk_ctx;
+
+static void twalk_cb(const void *nodep, VISIT which, int depth)
+{
+    twalk_ctx.cb(nodep, which, twalk_ctx.closure);
+}
+
+#define twalk_r(_tree, _cb, _closure) { twalk_ctx.cb = (_cb); twalk_ctx.closure = (_closure); twalk(_tree, twalk_cb); }
+
+
+
 /*
  * Rules:
  * NULL comes before anything else.

```

vashirov:
@progier389, I'm happy with any solution that allows us to do builds on EL8 :)
If you (and others) are satisfied with change you proposed, I'd gladly replace the backport code with yours.
Please let me know if you want to open a separate PR to replace mine or I can use your code as part of this PR.
Thanks!

progier389:
Created PR https://github.com/389ds/389-ds-base/pull/5516 
 (added #if to activate the fix only for glibc < 2.30)

vashirov:
Thanks, closing this in favor of https://github.com/389ds/389-ds-base/pull/5516.

PR #5518, opened 2022-11-10
Issue 5517 - Replication conflict CI test sometime fails

### LINKED ISSUE #5517: Replication conflict CI test sometime fails
**Issue Description**
dirsrvtests/tests/suites/replication/conflict_resolve_test.py::TestTwoSuppliers::test_conflict_attribute_multi_valued FAILED [ 75%]
dirsrvtests/tests/suites/replication/conflict_resolve_test.py::TestTwoSuppliers::test_conflict_attribute_single_valued FAILED [ 87%]

Sometime the test fails. 
I suspect that the 10s timeout is too small

progier389:
replication conflict test failure (both in bdb and mdb mode):
Several causes:
  - Test issues:
      - Cleanup issues
      - Timing issues
  - A server bug about MODRDN conflict resolution that triggered "entryrdn_rename_subtree failed - Failed to read the target element ..." message.
  
Fix: 
   - Test case:
       - Fixed the test cleanup to avoid alarming error messages in the error logs  
       - Replacing the sleep in test case by  Replicationmanager().test_replication
       - Increasing number of attempt when testing for entry from 10 to 60 (i.e having a ! minute timeout)
   - Modrdn operation:
       - Used correct dn (i.e The target entry one) rather than the MODRDN operation one to rename the subtree in
        entryrdn index (They are different when handling renaming conflict)
   - Search operation: 
       - Fixed a compilation warning in keys2idl about new search statistics
   - Entry cache:
       - Fix debug mode issues
       - Add a test about entry pointer validation when walking the hash table.

Note:  There are some other server bugs that the  Replication conflict CI test could triggers:
     - [Issue 5620](https://github.com/389ds/389-ds-base/issues/5620)
     - A SIGSEGV triggered by a corruption of the entry cache (I added code to help diagnose that we hit these bug)
     but something changed that last week (either the test machine dynamic changed or code rebase 
      somehow fixed the issue) and I was no more able to see that bug this week while it triggered in most than 30% of
      the tests before 

Issue: [5517](https://github.com/389ds/389-ds-base/issues/5517)

Reviewed by:  @tbordaz, thanks

### DIFF
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -44,7 +44,8 @@
         if (!(_x)) {                                                                    \
             slapi_log_err(SLAPI_LOG_ERR, "cache", "BAD CACHE ASSERTION at %s/%d: %s\n", \
                           __FILE__, __LINE__, #_x);                                     \
-            *(char *)0L = 23;                                                           \
+            slapi_log_backtrace(SLAPI_LOG_ERR);                                         \
+            *(char *)23L = 1;                                                           \
         }                                                                               \
     } while (0)
 
@@ -514,6 +515,30 @@ flush_remove_entry(struct timespec *entry_time, struct timespec *start_time)
     }
 }
 
+static inline void
+dbgec_test_if_entry_pointer_is_valid(void *e, void *prev, int slot, int line)
+{
+    /* Check if the entry pointer is rightly aligned and crash loudly otherwise */
+    if ( ((uint64_t)e) & ((sizeof(long))-1) ) {
+        /* If this message occurs, it means that we have reproduced the elusive entry cache corruption
+         * seen first while fixing replication conflict_resolution CI test 
+         * FYI some debug attempt have been stored in https://github.com/progier389/389-ds-base.git
+         * in branches:
+         *  debug-stuff1 (older try with log of debugging trick in slapd/dbgec*) 
+         *  debug-stuff2: Log the dn associated with backentries in a mmap file and retrieve the
+         *  dn of the corrupted entry. 
+         *   Note: if we are able to reproduce using this debug stuff and if it is always the same entry
+         *   we may catch the issue by adding a watchpoint when adding that backentry 
+         *   Note: you should disable the setuid to be able to use watchpoint 
+         */ 
+        slapi_log_err(SLAPI_LOG_FATAL, "dbgec_test_if_entry_pointer_is_valid", "cache.c[%d]: Wrong entry address: %p Previous entry address is: %p hash table slot is %d\n", line, e, prev, slot);
+        slapi_log_backtrace(SLAPI_LOG_FATAL);
+        *(char*)23 = 1;   /* abort() somehow corrupt gdb stack backtrace so lets generate a SIGSEGV */
+        abort();
+    }
+}
+
+
 /*
  * Flush all the cache entries that were added after the "start time"
  * This is called when a backend transaction plugin fails, and we need
@@ -536,6 +561,7 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
 
     for (size_t i = 0; i < ht->size; i++) {
         e = ht->slot[i];
+        dbgec_test_if_entry_pointer_is_valid(e, NULL, i, __LINE__);
         while (e) {
             struct backcommon *entry = (struct backcommon *)e;
             uint64_t remove_it = 0;
@@ -547,6 +573,7 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
             }
             laste = e;
             e = HASH_NEXT(ht, e);
+            dbgec_test_if_entry_pointer_is_valid(e, laste, i, __LINE__);
 
             if (remove_it) {
                 /* since we have the cache lock we know we can trust refcnt */
@@ -577,6 +604,7 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
 
         for (size_t i = 0; i < ht->size; i++) {
             e = ht->slot[i];
+            dbgec_test_if_entry_pointer_is_valid(e, NULL, i, __LINE__);
             while (e) {
                 struct backcommon *entry = (struct backcommon *)e;
                 uint64_t remove_it = 0;
@@ -588,6 +616,7 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                 }
                 laste = e;
                 e = HASH_NEXT(ht, e);
+                dbgec_test_if_entry_pointer_is_valid(e, laste, i, __LINE__);
 
                 if (remove_it) {
                     /* since we have the cache lock we know we can trust refcnt */
@@ -2165,7 +2194,7 @@ check_entry_cache(struct cache *cache, struct backentry *e)
     struct backentry *debug_e = cache_find_dn(cache,
                                               slapi_sdn_get_dn(sdn),
                                               slapi_sdn_get_ndn_len(sdn));
-    in_cache = cache_is_in_cache(cache, (void *)e);
+    int in_cache = cache_is_in_cache(cache, (void *)e);
     if (in_cache) {
         if (debug_e) { /* e is in cache */
             CACHE_RETURN(cache, &debug_e);
--- a/dirsrvtests/tests/suites/replication/conflict_resolve_test.py
+++ b/dirsrvtests/tests/suites/replication/conflict_resolve_test.py
@@ -17,9 +17,10 @@ from lib389.idm.nscontainer import nsContainers
 from lib389.idm.user import UserAccounts, UserAccount
 from lib389.idm.group import Groups
 from lib389.idm.organizationalunit import OrganizationalUnits
-from lib389.replica import ReplicationManager
+from lib389.replica import ReplicationManager, Replicas
 from lib389.agreement import Agreements
 from lib389.plugins import MemberOfPlugin
+from lib389.dirsrv_log import DirsrvErrorLog
 
 pytestmark = pytest.mark.tier1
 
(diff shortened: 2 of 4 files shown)


droideck:
The test still fails in the GitHub CI though...

```
        # Waiting fo the user to be replicated
        for i in range(0,30):
            time.sleep(1)
            if test_users_m2.exists():
                break
>       assert(test_users_m2.exists())
E       assert False
E        +  where False = <bound method DSLdapObject.exists of <lib389.idm.user.UserAccount object at 0x7fa344ca93c0>>()
E        +    where <bound method DSLdapObject.exists of <lib389.idm.user.UserAccount object at 0x7fa344ca93c0>> = <lib389.idm.user.UserAccount object at 0x7fa344ca93c0>.exists
```

I wonder if it makes sense to set something like `@pytest.mark.flaky(max_runs=2, min_passes=1)` or, could be the problem is somewhere else...

progier389:
I do not think this test should be flaky as it basically test that replication is working 
  bit in a different way than  wait_for_replication.
I have noticed that sometime the CI virtual machine get frozen for quite long
I see that the  wait_for_replication default timeout is 1 minute, I will try to set the timeout to 1 minute too 
replica.py:    def wait_for_replication(self, from_instance, to_instance, timeout=60):

progier389:
Still failing with 60s ! -:(
The bad thing is that I was able to run the test in loop on my laptop for a few hours without failure.

progier389:
Adding the error log when the test is failing shows that there is indeed an issue:
 Replication is broken:
 conflict_resolve_test.py:139 [18/Nov/2022:23:41:34.053819879 +0000] - ERR - entryrdn_delete_key - Failed to remove cn=test_container; has a child nsuniqueid=e780f99b-679911ed-8069fb40-53b262c7+uid=test_user_1118
conflict_resolve_test.py:139 [18/Nov/2022:23:41:34.054373185 +0000] - ERR - index_addordel_entry - database index operation failed BAD 1031, err=-1 Unexpected dbimpl error code
conflict_resolve_test.py:139 [18/Nov/2022:23:41:34.054940292 +0000] - ERR - ldbm_back_delete - index_addordel_entry(cn=test_container,dc=example,dc=com, 0x26) failed (-1)
conflict_resolve_test.py:139 [18/Nov/2022:23:41:34.061831571 +0000] - DEBUG - NSMMReplicationPlugin - process postop - conn=19 op=4 csn=63781775001d00010000 canceling operation csn
conflict_resolve_test.py:139 [18/Nov/2022:23:41:34.062763281 +0000] - ERR - NSMMReplicationPlugin - process_postop - Failed to apply update (63781775001d00010000) error (1).  Aborting replication session(conn=19 op=4)

It looks like a replication bug.
IMHO a replicated delete operation should not fail because children exists. 
URP should handle this case (probably by tombstoning the children entries)

progier389:
I will try to add a new test that reproduce this issue systematically.

progier389:
Test are still failing later on because of server crashes:

(gdb) #0  slapi_timespec_diff
    (a=a@entry=0x653d63642c72659e, b=b@entry=0x7f93dd1f9800, diff=diff@entry=0x7f93dd1f9510) at ldap/servers/slapd/time.c:259
#1  0x00007f941a8db321 in flush_remove_entry
    (start_time=0x7f93dd1f9800, entry_time=0x653d63642c72659e)
    at ldap/servers/slapd/back-ldbm/cache.c:509
#2  flush_hash
    (cache=cache@entry=0x7f941dbf3950, start_time=start_time@entry=0x7f93dd1f9800, type=type@entry=0) at ldap/servers/slapd/back-ldbm/cache.c:583
#3  0x00007f941a8db513 in revert_cache
    (start_time=0x7f93dd1f9800, inst=0x7f941dbf38c0)
    at ldap/servers/slapd/back-ldbm/cache.c:620
#4  revert_cache (inst=0x7f941dbf38c0, start_time=0x7f93dd1f9800)
    at ldap/servers/slapd/back-ldbm/cache.c:615
#5  0x00007f941a90988d in ldbm_back_modrdn (pb=0x7f941b161280)
    at ldap/servers/slapd/back-ldbm/ldbm_modrdn.c:1323
#6  0x00007f941f13abda in op_shared_rename.constprop.0
    (pb=pb@entry=0x7f941b161280, passin_args=0)
    at ldap/servers/slapd/modrdn.c:625
#7  0x00007f941f0e339c in do_modrdn (pb=pb@entry=0x7f941b161280)
    at ldap/servers/slapd/modrdn.c:225
#8  0x0000562e75dbd1fc in connection_dispatch_operation
    (pb=0x7f941b161280, op=<optimized out>, conn=<optimized out>)
    at ldap/servers/slapd/connection.c:648
#9  connection_threadmain (arg=<optimized out>)
    at ldap/servers/slapd/connection.c:1805
#10 0x00007f941e74f413 in _pt_root () at /lib64/libnspr4.so

progier389:
That is a bit weird it seems that the cache table moved under us although it is locked:

```
(gdb) up 2
(gdb) #2  flush_hash (cache=cache@entry=0x7f3bc8ff3950, 
    start_time=start_time@entry=0x7f3b86dec800, type=type@entry=0)
    at ldap/servers/slapd/back-ldbm/cache.c:583
583	                if (flush_remove_entry(&entry->ep_create_time, start_time)) {
(gdb) print *cache->c_idtable
(gdb) $1 = {offset = 88, size = 2490373, hashfn = 0x0, 
  testfn = 0x7f3bc5ddd950 <entry_same_id>, slot = {0x0}}
(gdb) print $rbp
(gdb) $2 = (void *) 0x653d63642c72656e

```
dissassembling shows that entry is %rdb aka 0x653d63642c72656e which does not looks like a valid struct address
    (I would expect struct address to be long aligned  (i.e ending by 0,4,8,c )
BTW: in the stack entry_time = 0x653d63642c72659e and entry_time = &entry->ep_create_time aka entry + 0x30 
(which confirm that entry is indeed 0x653d63642c72656e when the crash occurs. )

Another surprising thing is that cache->c_idtable->slot (aka ht->slot)  is NULL which should not be possible as it was dereferenced to get the entry ...

So know the point is to understand how the cache could be corrupted while we are in flush_hash function.

progier389:
I can confirm that the cache get corrupted:
```
[13/Jan/2023:20:16:32.197298492 +0000] - ERR - test_if_entry_pointer_is_valid - cache.c[625]: Wrong entry address: 0x653d63642c72656e Previous entry address is: 0x7f01a67abca0 hash table slot is 17582

x /s $cache->c_dntable->slot[$slot]
(gdb) 0x7f01a67abca0:   "nsuniqueid=0907f70a-937f11ed-bf88ffbb-c34c9a02,uid=test_user_1100,cn=test_container,dc=example,dc=com"
```

**Shows that a backdn entry get either overwritten or freed and reused (by a dn)**

Added trace to see if backentry_free or backdn_free is called with the problematic address.

The bad news is that entry test_user_1100 had a complex story:
```
[13/Jan/2023:20:16:30.502021746 +0000] - DEBUG - conn=3 op=6 csn=63c1bbf5000100010000 - urp_naming_conflict_removal - Found nsuniqueid=0efdd80b-937f11ed-896dfe19-14f2ec64+uid=test_user_1110,cn=test_container,dc=example,dc=com
[13/Jan/2023:20:16:30.507037064 +0000] - ERR - cache - cache.c[1340]: Freeing cache entry 0x7f01ac1f40f8
..
[13/Jan/2023:20:16:32.146099816 +0000] - ERR - NSMMReplicationPlugin - tombstone_to_glue - conn=3 op=30 csn=63c1bc19000000010000 - Resurrected tombstone uid=test_user_1100,cn=test_container,dc=example,dc=com to glue reason 'renameTombstone'
[13/Jan/2023:20:16:32.146683318 +0000] - DEBUG - conn=3 op=30 csn=63c1bc19000000010000 - urp_modrdn_operation  - Target_entry nsuniqueid=0907f70a-937f11ed-bf88ffbb-c34c9a02,uid=test_user_1100,cn=test_container,dc=example,dc=com is a tombstone; Renaming since delete was after rename.
...
[13/Jan/2023:20:16:32.193600080 +0000] - DEBUG - conn=3 op=31 csn=63c1bc1e000000010000 - urp_modrdn_operation - nsuniqueid=0907f70b-937f11ed-bf88ffbb-c34c9a02,uid=test_user_1101,cn=test_container,dc=example,dc=com - Target entry is a tombstone.
[13/Jan/2023:20:16:32.197298492 +0000] - ERR - test_if_entry_pointer_is_valid - cache.c[625]: Wrong entry address: 0x653d63642c72656e Previous entry address is: 0x7f01a67abca0 hash table slot is 17582
....
[13/Jan/2023:20:16:32.161028967 +0000] - DEBUG - conn=3 op=30 csn=63c1bc19000000010000 - urp_fixup_add_cenotaph - addinng cenotaph: cenotaphID=0907f70a-937f11ed-bf88ffbb-c34c9a02+uid=test_user_1100,cn=test_container,dc=example,dc=com
```
So that means lot of code could be the root cause ...
I will see if I can use watchpoint to trap the thread that is writing at that address.
  adding the watchpoint in  add_hash if the key lenght > 100 and contains: test_user_1100 
  Freeing the watchpoint in backdn_free

tbordaz:
> 

If I read it correctly test_if_entry_pointer_is_valid is called during flush_cache/revert_cache that is called during an error during TXN_BE_PRE/POST or backend error.

A possibility is to try to simulate an backend update failure on MODRDN(glue_entry) ?

progier389:
I think you are right: the error is detected in the error_return case .
Which probably means that there are two bugs. 
 [1] It should never occur in the case of a replicated operation ( a replicated operation is not supposed to fail)
      That said, I am not sure about why it is there (Only that the failure is not logged.) 
        Should probably add a trace before all goto error_return without SLAPI_LOG_ERR trace to find the path
 [2] The cache should not be corrupted anyway.
       As the dn that overwrite the backdn is the one of a previous completed modrdn operation I do no know 
      what is the path that created the corruption

And I would rather understand why [2] occurs first because it may be harder to reproduce the problem once [1] will be fixed.

About [2] 
My feeling is that there is a confusion between dn and backdn at some point and we are freeing a backdn thinking it is a dn (so the backdn is not removed from the hash)
And the random behavior depends if the address was reused or not in a malloc unfortunately the test PASS with asan build so finding the invalid free (i.e if the backdn next field was overwritten or not)
The bad thing is that asan build does not help as the test pass ...

progier389:
I tried to run with a deterministic uuid generator, so entries have the same nsuniqueid in different runs
I was hoping that the hash table slot that contains the broken  entry would stay the same between runs (to be able to set a watchpoint)
But apparently it is not the case: There should be another cause creating the randomness. 
Trying now a run that checks if freed block is anchored in the cache ...

progier389:
@tbordaz FYI: I found why modrdn get in error_return case while no error is logged: 
 The mmr preop plugin has returned SLAPI_PLUGIN_NOOP 
   (likely after detecting that the entry was renamed/deleted by an operation with an higher csn
     (because the access logs shows that replication has not already played the operation)  
 That said, It seems a bit cumbersome to run in the error path in this case, I would rather expected to be in the common path ( I think it is to  run the postop plugin but it seems strange anyway ... )

progier389:
Sorry, I did not made it ( I have not yet found why it crash, 
My update was just about a point that we discussed (The fact that the crash occurs in error case code but no error was recorded. ( I found it by just adding trace in all goto error_return then once I knew that mmr plugin was triggering the goto, I logged the error code)
About the main issue I guess I will try for writing data in mmaped file (as log)
But these last two days I got in trouble to reproduce the issue ...

tbordaz:
opss, I misunderstood your update. Anyway knowing the reason of error condition (NOOP) explains why we have this crash so rarely.
There are several cache update/flush around error handling. There are a potential double call of rever_cache in line [1322](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c#L1322) and line [1405](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c#L1405) . With the same timestamp...

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -514,6 +515,30 @@ flush_remove_entry(struct timespec *entry_time, struct timespec *start_time)
     }
 }
 
+static inline void
+dbgec_test_if_entry_pointer_is_valid(void *e, void *prev, int slot, int line)
+{
+    /* Check if the entry pointer is rightly aligned and crash loudly otherwise */
+    if ( ((uint64_t)e) & ((sizeof(long))-1) ) {
```
tbordaz, line 522: Reading this I was wondering if it was valid in little/big endian. I think it is. Am I correct ?

tbordaz (approved):

ldap/servers/slapd/back-ldbm/cache.c
progier389: you are right: in this case, Endianess does not matter:
Endianess determines where the bits are stored in memory but neither casting a pointer to an integer nor performing binary operations on an integer is impacted by endianess
 (In short as far as there is no (explicit or implicit) cast to (char*)  there should not be any endianess issue.)

PR #5519, opened 2022-11-10
Run CI test with lmdb

progier389:
Please, do not review this PR:
 This change is not intended to be committed. 
It only trigger CI tests on lmdb


droideck:
I think you don't need to create a PR for just the CI run. If you push your branch to your fork, it should trigger the CI there (you can see the symbol on the left side of your commit).

The yellow circle on the screenshot.

image

PR #5523, opened 2022-11-14
Issue 5521 - BUG - Pam PTA multiple issues

### LINKED ISSUE #5521: Pam Pass Through - Multiple issues in setup/config
Setting up Pam Pass Through in it's current form is extremely rocky as a setup process. 

-- rh docs list wrong plugin name.
PAM Pass-Through Auth Plugin
PAM Pass Through Auth

-- dsconf enables the wrong plugin
dsconf localhost plugin pass-through-auth enable
Enables dn: cn=Pass Through Authentication,cn=plugins,cn=config not cn=PAM Pass Through Auth,cn=plugins,cn=config

-- pamMissingSuffix requires a value by default, will be accepted during run time and cause server to fail to start
[11/Nov/2022:11:50:21.048486118 +1000] - ERR - pam_passthru-plugin - pam_passthru_validate_config - Valid values for pamMissingSuffix are ERROR, ALLOW, IGNORE

-- pamSecure when required only emits a plugin trace warn not an error.
[11/Nov/2022:13:35:40.318908742 +1000] - DEBUG - pam_passthru-plugin - pam_passthru_bindpreop - Connection not secure (secure connection required; check config)

Firstyear:
Bug Description: Pam PTA and the lib389 cli had numerous issues that were affecting administration and configuration.

Fix Description: This fixes many issues:

* add pam-[enable,disable,show] seperate to pta enable. We can't combine these into one because they are seperate plugins. They also still needs ways to enable them outside of the direct config attribute manipulation.
* Make pamMissingSuffix return a default of IGNORE on NONE. This is because many of the current tools don't actually set this value and it can block server restarts.
* pamSecure would not warn properly on lack of TLS connections which can trick users into thinking the plugin is not working.

fixes: https://github.com/389ds/389-ds-base/issues/5521

Review by: ???


progier389:
If I understand rightly,  you propose to add new sub command action (pam-enable, pam-disable, .. ) to the PTA plugin menu. 
I wonder if we should not rather create a new plugin menu for PAM pass-though plugin and use the standard (enable, disable) actions. (it is a different plugin, so why not fully handling it as a different plugin?) 
@mreynolds389, do you have an opinion on that point ?  cockpit impact of both solution  ?

mreynolds389:
> If I understand rightly, you propose to add new sub command action (pam-enable, pam-disable, .. ) to the PTA plugin menu. I wonder if we should not rather create a new plugin menu for PAM pass-though plugin and use the standard (enable, disable) actions. (it is a different plugin, so why not fully handling it as a different plugin?) @mreynolds389, do you have an opinion on that point ? cockpit impact of both solution ?

Yes we need to separate it in Cockpit as well.  I was trying to simplify/combine these plugins in the past.   I don't recall all the reasons now (I think they are linked in some way though), but with this change we will need to separate it.  A new Issue should be opened to handle the UI part (SUSE doesn't have Cockpit so I don't expect William to make those changes).

@Firstyear what versions of 389 do you need this backported to?  Just "main", or further down?

progier389:
About 'I think they are linked in some way though'
They are linked in the sense that you should not  enable both plugins on the same suffixes
   (but IMHO we should handle that with an healthcheck rule and have a simple configuration scheme)

Firstyear:
> > If I understand rightly, you propose to add new sub command action (pam-enable, pam-disable, .. ) to the PTA plugin menu. I wonder if we should not rather create a new plugin menu for PAM pass-though plugin and use the standard (enable, disable) actions. (it is a different plugin, so why not fully handling it as a different plugin?) @mreynolds389, do you have an opinion on that point ? cockpit impact of both solution ?
> 
> Yes we need to separate it in Cockpit as well. I was trying to simplify/combine these plugins in the past. I don't recall all the reasons now (I think they are linked in some way though), but with this change we will need to separate it. A new Issue should be opened to handle the UI part (SUSE doesn't have Cockpit so I don't expect William to make those changes).

I think that as they are seperate plugins they should be seperate. The only reason I "jammed" in the pam-enable/disable like this is because I think the dsconf cli is bound to the cockpit ui, so we can't "un-glue" these two from each other without also changing cockpit at the same time.

Otherwise, they should be seperate plugins IMO. ldap-pass-through and pam-pass-through. I would have much preferred that but, as said, I'm trying to "not break things". 

> 
> @Firstyear what versions of 389 do you need this backported to? Just "main", or further down?

I'm of two minds. It would "be nice" to have in 2.0 for SLE15SP4, but we can also work around all the issues resolved here with other commands.

I think that perhaps:

* If we merge it as is, we backport down to 2.0 since this is "not disruptive".
* If we decide to split to two plugin handler commands here, then we just leave it on main to allow you time to do cockpit fixes.

Firstyear:
@mreynolds389 What do you prefer here? Should I split the command or leave it as is?

Firstyear:
@mreynolds389 Hey there, what would you like me to do on this one?

Also probably need to decide what versions we want this and https://github.com/389ds/389-ds-base/pull/5528 backported for (if any).

mreynolds389:
> @mreynolds389 Hey there, what would you like me to do on this one?
> 
> Also probably need to decide what versions we want this and #5528 backported for (if any).

Split it, and cherry pick down to 1.4.3.  We just need to update the UI at the same time (I can handle that)

Firstyear:
Okay, I'll do the split and testing of the cli today and will leave the ui to you. Thank you! <3

Firstyear:
@mreynolds389 This splits the commands. I have tested both work with test setups and can enable/disable the plugins and fully reconfigure them :)

Firstyear:
@mreynolds389 I know you're busy, can you have a look at this? @droideck too maybe.

mreynolds389 (approved):
LGTM

droideck (approved):
The code looks good to me!

I have one minor concern though.
Will it make more sense to rename the main commit and the issue to include its RFE nature?
It'll make it easier for users later to see in changelog that the behaviour has changed.
Maybe, even split into two commits...

Also, the commit adds a couple of CLI subtools (and we'll need to document them).

Firstyear:
> The code looks good to me!
> 
> I have one minor concern though. Will it make more sense to rename the main commit and the issue to include its RFE nature? It'll make it easier for users later to see in changelog that the behaviour has changed. Maybe, even split into two commits...
> 
> Also, the commit adds a couple of CLI subtools (and we'll need to document them).

Yep, I'm happy to reword the commit message.

Firstyear:
@droideck I reworded the commit and rebased, is this more what you wanted?

droideck:
> @droideck I reworded the commit and rebased, is this more what you wanted?

Looks great! Thanks!

Firstyear:
> > @droideck I reworded the commit and rebased, is this more what you wanted?
> 
> Looks great! Thanks!

You want it left as two commits? Or squash? There isn't much reason to keep the first commit as is since it's just an intermediate state.

droideck:
> > Looks great! Thanks!
> 
> You want it left as two commits? Or squash? There isn't much reason to keep the first commit as is since it's just an intermediate state.

I think two commits are better as we create the package's changelog out of commits and this way it'll be nice and visible, and we won't miss it for sure.

Firstyear:
Thanks everyone!

PR #5525, opened 2022-11-15
Issue 5162 - RFE - CLI allow adding CA certificate bundles

mreynolds389:
Description:  Parse PEM file and add each CA cert separately

relates: https://github.com/389ds/389-ds-base/issues/5162

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/clu/ca_cert_bundle_test.py
+++ b/dirsrvtests/tests/suites/clu/ca_cert_bundle_test.py
@@ -1,0 +1,166 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import logging
+import pytest
+import os
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.topologies import topology_st as topo
+from lib389.cli_base import FakeArgs
+from lib389.cli_conf.security import cacert_add, cacert_list, cert_del
+from lib389.cli_ctl.tls import import_ca
+from lib389.cli_base import LogCapture
+
+log = logging.getLogger(__name__)
+
+PEM_CONTEXT = """
+-----BEGIN CERTIFICATE-----
+MIIFZjCCA06gAwIBAgIFAL18JOowDQYJKoZIhvcNAQELBQAwZTELMAkGA1UEBhMC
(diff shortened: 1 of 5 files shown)


dirsrvtests/tests/suites/clu/ca_cert_bundle_test.py
```diff
@@ -0,0 +1,158 @@
+import logging
```
droideck, line 9: Copyright is missing. :)

src/lib389/lib389/cli_ctl/tls.py
```diff
@@ -33,34 +38,102 @@ def import_client_ca(inst, log, args):
...
+
     tls = NssSsl(dirsrv=inst)
-    cert_path = args.cert_path
-    nickname = args.nickname
-    if nickname.lower() == CERT_NAME.lower() or nickname.lower() == CA_NAME.lower():
-        log.error("You may not import a CA with the nickname %s or %s" % (CERT_NAME, CA_NAME))
-        return
-    tls.add_cert(nickname=nickname, input_file=cert_path)
-    tls.edit_cert_trust(nickname, "C,,")
+    try:
```
droideck: The code here and in `src/lib389/lib389/cli_conf/security.py` is very similar.

I think it may be worth encapsulating the code to `NssSsl()` class:
https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/nss_ssl.py#L53

`add_cert` is already there and it'll be natural to have `add_cert_bundle` (to something), IMO.

What do you think?

src/lib389/lib389/utils.py
```diff
@@ -308,7 +308,7 @@ def selinux_label_file(path, label):
         raise ValueError(f'Cannot change file context for {path} because it is defined in seLinux policy. Please choose another path.')
     if label:
         try:
-            log.info(f"Setting label {label} in seLinux file context {path}.")
+            log.debug(f"Setting label {label} in seLinux file context {path}.")
```
droideck: Very minor nitpick... I think officially it's SELinux instead of seLinux :D

droideck (approved):
A couple of minor things and one suggestion.
Otherwise, looks good!

src/lib389/lib389/utils.py
mreynolds389: > Very minor nitpick... I think officially it's SELinux instead of seLinux :D

You can yell at Pierre for that :-)

progier389: Culprit ! And the raise ValueError message in selinux_label_file also have the issue.
    The good new is that only these 3 cases seems impacted.

PR #5528, opened 2022-11-16
Issue 5526 - RFE - Improve saslauthd migration options

### LINKED ISSUE #5526: Saslauthd migration paths
OpenLDAP is commonly used with saslauthd for external authentication ( https://openldap.org/doc/admin24/security.html#Configuring%20saslauthd ). This is denoted on openldap servers by a config of:

```
 pwcheck_method: saslauthd
 saslauthd_path: /var/run/sasl2/mux
```

Users are then bound to external auth services by a userPassword hash type:

```
 userPassword: {SASL}user@ad.example.com
```

I have now written https://github.com/Firstyear/pam_saslauthd as a "glue" between 389-ds and it's pam pass through infrastructure, and saslauthd. Since it is possible now, we can improve the process of allowing migrations to 389-ds from openldap with these components.

I propose that during an openldap to ds migration, that if a user is detected with the SASL userPassword scheme, then userPassword is removed from the account and we add a "saslauthd_id" field to the entry (along with an extra objectClass). 

We can then use pam pass through with map_method ENTRY and id-attr of saslauthd_id.

The file `/etc/pam.d/saslauthd` would come from the pam_saslauthd package, so it's out of scope here. We just need to do "most of" the configuration at the 389-ds layer.

Firstyear:
Bug Description: We should improve our migration paths from openldap to allow the commonly used saslauthd plugin.

Fix Description: This adds the import transform to convert users to use the nsSaslauthId field. This also adds a helper in migration to enable the plugins as needed. Finally this also adds some hardening to pam_pta.

fixes: https://github.com/389ds/389-ds-base/issues/5526

Review by: ???


src/lib389/lib389/plugins.py
```diff
@@ -1532,7 +1532,7 @@ class PAMPassThroughAuthConfigs(DSLdapObjects):
 
     def __init__(self, instance, basedn="cn=PAM Pass Through Auth,cn=plugins,cn=config"):
         super(PAMPassThroughAuthConfigs, self).__init__(instance)
-        self._objectclasses = ['top', 'extensibleObject', 'nsslapdplugin', 'pamConfig']
+        self._objectclasses = ['top', 'extensibleObject', 'pamConfig']
```
progier389, line 1535: It seem a bit strange that  nsslapdplugin objectclass must be removed while this entry is a plugin config entry
as there will be trouble if one of the nsslapdplugin mandatory attribute is missing ...

Firstyear: But they aren't needed or mandatory. They should never have been added in the first place. These entries are children to the pam passthrough plugin's dn.

progier389: Ok. now I understand!

progier389 (approved):
LGTM

Firstyear:
Hey @droideck @mreynolds389 @progier389 it looks like in https://github.com/389ds/389-ds-base/pull/5523#issuecomment-1338620409 I cherry picked these down to 1.4.3, but I think I forgot to do similar in this PR. This pr can probably only go down to 2.0 since I think that's where the migration tools were added (and that's what version we need these in for SUSE).

Any issue with me doing that cherry-pick? Sorry about forgetting in the first place!

Firstyear:
@droideck @mreynolds389 ping on this, good for me to backport down to 2.0?

Issue #5532, opened 2022-11-16
Database compaction TOD needs ot be more robust


mreynolds389:
**Issue Description**

The current implementation of the database compaction TOD setting doesn't work correctly if the TOD happens before the interval is exceeded.  This leads to the compaction happens at really odd times.

mreynolds389:
2b1bfcea0..7b9c833f6  389-ds-base-2.2 -> 389-ds-base-2.2
9d3f23383..49458953a  389-ds-base-2.1 -> 389-ds-base-2.1
a0ad4b58f..6a983f4aa  389-ds-base-2.0 -> 389-ds-base-2.0
afa006893..de5724fe4  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5533, opened 2022-11-16
Issue 5532 - Make db compaction TOD more robust.

mreynolds389:
Bug Description:

The time of day compaction setting does promise that the compaction will happen as configured.  This is becuase the compaction interval starts when the server is started.  Once it wakes up and we are "past" the TOD setting then we compact, but it can happen at any time once the TOD has passed.

Fix Description:

Once the compaction interval is hit we create an "event" with the exact time the compaction should start.

relates: https://github.com/389ds/389-ds-base/issues/5532

### DIFF
--- a/dirsrvtests/tests/suites/config/compact_test.py
+++ b/dirsrvtests/tests/suites/config/compact_test.py
@@ -1,7 +1,16 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2022 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
 import logging
 import pytest
 import os
 import time
+import datetime
 from lib389.utils import get_default_db_lib
 from lib389.tasks import DBCompactTask
 from lib389.backend import DatabaseConfig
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/config/compact_test.py
```diff
@@ -72,11 +73,31 @@ def test_compaction_interval_and_time(topo):
     """
 
     inst = topo.ms["supplier1"]
+
+    # Calculate the compaction time (1 minute from now)
```
mreynolds389, line 85: This should say 2 minutes, I'll fix it later.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -3946,59 +4001,21 @@ bdb_checkpoint_threadmain(void *param)
...
-                rc = bdb_db_compact_one_db(db, inst);
-                if (rc) {
-                    slapi_log_err(SLAPI_LOG_ERR, "bdb_checkpoint_threadmain",
-                                  "compactdb: failed to compact changelog for %s; db error - %d %s\n",
-                                   inst->inst_name, rc, db_strerror(rc));
-                    break;
-                }
-            }
+            /* Start compaction event */
+            compacting = PR_TRUE;
```
tbordaz: This flag rather means that compacting is schedule than compacting is running. Could we change the name ? like compacting_scheduled = PR_TRUE ?

tbordaz (commented):
The code looks good (except a minor comment) but there is something I do not understand. Assuming a monthly compaction is configured (compact_interval) and that the start of the interval is set with DS startup time. What happens if we restart RHDS each 25days ? what part of code handle this ?

mreynolds389:
> The code looks good (except a minor comment) but there is something I do not understand. Assuming a monthly compaction is configured (compact_interval) and that the start of the interval is set with DS startup time. What happens if we restart RHDS each 25days ? what part of code handle this ?

Yup, compaction will never occur in this case.  This is how its always been.

mreynolds389:
> > The code looks good (except a minor comment) but there is something I do not understand. Assuming a monthly compaction is configured (compact_interval) and that the start of the interval is set with DS startup time. What happens if we restart RHDS each 25days ? what part of code handle this ?
> 
> Yup, compaction will never occur in this case. This is how its always been.

Now that we've been compacting at midnight by default, perhaps we should lower the default interval from 30 days to something smaller?

tbordaz:
I think those options are in the hands of admin. Only admin knows the rate update and when there is calm periods. IMHO the default value looks good.
An option would be to run db_stat to evaluate the number of empty pages and to suggest the benefit of compaction. It could be done in healthcheck ? (but in a different ticket)

tbordaz (approved):
The fix looks good to me. ACK

dirsrvtests/tests/suites/config/compact_test.py
droideck: Could you, please, also, add copyright to this file?

droideck (approved):
Otherwise, looks good to me too!

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: > This flag rather means that compacting is schedule than compacting is running. Could we change the name ? like compacting_scheduled = PR_TRUE ?

Fixed

PR #5538, opened 2022-11-18
Issue 3729 - (cont) RFE Extend log of operations statistics in access…

tbordaz:
… log

Bug description:
	This is a continuation of the #3729
	The previous fix did not manage internal SRCH, so
	statistics of internal SRCH were not logged

Fix description:
	For internal operation log_op_stat uses
	connid/op_id/op_internal_id/op_nested_count that have been
	computed log_result

	For direct operation log_op_stat uses info from the
	operation itself (o_connid and o_opid)

	log_op_stat relies on operation_type rather than
	o_tag that is not available for internal operation

relates: #3729

Reviewed by:


progier389 (approved):
LGTM

PR #5546, opened 2022-11-25
Issue 5545 - A random crash in import over lmdb

### LINKED ISSUE #5545: A random crash may occurs while importing an ldif fdile in lmdb database.
**Issue Description**

 - Platform: Fedora

**Steps to Reproduce**
1. Generate a 10 000 000 users ldif file
2. Set up a server using lmdb
3. Import the ldif file
4. See error: (import fails and core is dumped)

**Expected results**
Import should complete without crashing

**Additional context**

A random crash may occurs while importing an ldif fdile in lmdb database.
Note: on 3 run while trying to import 10M entries I got the crash after importing 5M, 9.5M, 0.1M entries (so very random but not frequent)

Core was generated by `/root/pierre_do_not_remove/dsroot/sbin/ns-slapd ldif2db -D /root/pierre_do_not_'.
#0  0x00007f69847c8528 in dbmdb_import_prepare_worker_entry (wqelmnt=0x111e330) at ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c:1309
1309    ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c: No such file or directory.
[Current thread is 1 (Thread 0x7f5cfd8b0700 (LWP 29973))]
(gdb) where
#0  0x00007f69847c8528 in dbmdb_import_prepare_worker_entry (wqelmnt=0x111e330) at ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c:1309
#1  0x00007f69847ca902 in dbmdb_import_worker (param=<optimized out>) at ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c:3155
#2  0x00007f698d2c3968 in _pt_root (arg=0x1116d90) at ../../.././nspr/pr/src/pthreads/ptthread.c:201
#3  0x00007f698cc5e1cf in start_thread (arg=<optimized out>) at pthread_create.c:479
#4  0x00007f698f685dd3 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) x /i 0x00007f69847c8528
=> 0x7f69847c8528 <dbmdb_import_prepare_worker_entry+168>:  repz cmpsb %es:(%rdi),%ds:(%rsi)
(gdb) print $es
$1 = 0
(gdb) print $rdi
$2 = 140091171154120
(gdb) print $ds
$3 = 0
(gdb) print %rsi
A syntax error in expression, near `%rsi'.
(gdb) print $rsi
$4 = 0
(gdb) x /s 140091171154120
0x7f69847e28c8: "version:"
(gdb) p wqelmnt->data
$5 = (void *) 0x7f5cf45db070

mdb_import_threads.c:1309 is:
    if (0 == ctx->wgc.version_found && 0 == strncmp(estr, "version:", 8)) {

so the crash occurs because estr is null (
 apparently a race condition issue because estr is wqelmnt->data which is
 not NULL at the crash time )

progier389:
Random crash due to an accelerator that bypass lock while dequeueing worker thread entry but that cause synchronization issue around the hardware memory cache)

Solution:  lock systematically to perform a membar that ensure proper synchronization.
 (It does not impact the performances because  the provider -> worker queue is not the performance bottleneck.
  (that is the writer thread database operation that limits the throughput) 
 
 Also added 2 improvements:
  - Use MDB_NOSYNC flags during off-line import (Anyway if the process is interrupted the import must be rerun)
  - Log regularly some  statistics about import writer thread (to help determining if the import bottleneck is because 
    the thread is waiting for input data or waiting that the lmdb operation complete

Issue: [5545](https://github.com/389ds/389-ds-base/issues/5545)

Reviewed by:  @tbordaz , @Firstyear

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
@@ -621,14 +621,6 @@ dbmdb_import_monitor_threads(ImportJob *job, int *status)
             sprintf(p, "recent rate %.1f/sec, ",
                          job->recent_progress_rate);
             import_log_notice(job, SLAPI_LOG_INFO, "dbmdb_import_monitor_threads", "%s", buffer);
-            for (current_worker = job->worker_list; current_worker != NULL; current_worker = current_worker->next) {
-                char buf[200];
-                char *summary = mdb_stat_summarize(MDB_STAT_PT(current_worker), buf, sizeof buf);
-                if (summary) {
-                    import_log_notice(job, SLAPI_LOG_INFO, "dbmdb_import_monitor_threads",
-                                      "Thread %s: %s", current_worker->name, summary);
-                }
-            }
         }
 
         count++;
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
@@ -621,6 +621,14 @@ dbmdb_import_monitor_threads(ImportJob *job, int *status)
             sprintf(p, "recent rate %.1f/sec, ",
                          job->recent_progress_rate);
             import_log_notice(job, SLAPI_LOG_INFO, "dbmdb_import_monitor_threads", "%s", buffer);
+            for (current_worker = job->worker_list; current_worker != NULL; current_worker = current_worker->next) {
+                char buf[200];
+                char *summary = mdb_stat_summarize(MDB_STAT_PT(current_worker), buf, sizeof buf);
+                if (summary) {
+                    import_log_notice(job, SLAPI_LOG_INFO, "dbmdb_import_monitor_threads",
+                                      "Thread %s: %s", current_worker->name, summary);
+                }
+            }
         }
 
         count++;
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
@@ -26,6 +26,8 @@
 #include <assert.h>
 #include "mdb_import.h"
 #include "../vlv_srch.h"
+#include <sys/time.h>
+#include <time.h>
 
 #define CV_TIMEOUT    10000000  /* 10 milli seconds timeout */
 
@@ -166,9 +168,9 @@ void
 wait_for_starting(ImportWorkerInfo *info)
 {
     PRIntervalTime sleeptime;
-    sleeptime = PR_MillisecondsToInterval(import_sleep_time);
     /* pause until we're told to run */
     while ((info->command == PAUSE) && !info_is_finished(info)) {
+        sleeptime = PR_MillisecondsToInterval(import_sleep_time);
         info->state = WAITING;
         DS_Sleep(sleeptime);
     }
@@ -3139,13 +3141,11 @@ dbmdb_import_worker(void *param)
         wait_for_starting(info);
 
         /* Wait until data get queued */
-        if (wqelmnt->wait_id == 0) {
-            pthread_mutex_lock(&ctx->workerq.mutex);
-            while (wqelmnt->wait_id == 0 && !info_is_finished(info) && ctx->producer.state != FINISHED) {
-                safe_cond_wait(&ctx->workerq.cv, &ctx->workerq.mutex);
-            }
-            pthread_mutex_unlock(&ctx->workerq.mutex);
+        pthread_mutex_lock(&ctx->workerq.mutex);
+        while (wqelmnt->wait_id == 0 && !info_is_finished(info) && ctx->producer.state != FINISHED) {
+            safe_cond_wait(&ctx->workerq.cv, &ctx->workerq.mutex);
         }
+        pthread_mutex_unlock(&ctx->workerq.mutex);
         if (wqelmnt->wait_id == 0) {
              break;
         }
@@ -3712,9 +3712,13 @@ dbmdb_import_writer(void*param)
     int count = 0;
     int rc = 0;
 
+    MDB_STAT_INIT(info);
     while (!rc && !info_is_finished(info)) {
+        MDB_STAT_STEP(info, MDB_STAT_PAUSE);
         wait_for_starting(info);
+        MDB_STAT_STEP(info, MDB_STAT_READ);
         slot = dbmdb_import_q_getall(&ctx->writerq);
+        MDB_STAT_STEP(info, MDB_STAT_RUN);
         if (info_is_finished(info)) {
             dbmdb_import_q_flush(&ctx->writerq);
             break;
@@ -3725,11 +3729,14 @@ dbmdb_import_writer(void*param)
 
         for (; slot; slot = nextslot) {
             if (!txn) {
+                MDB_STAT_STEP(info, MDB_STAT_TXNSTART);
                 rc = TXN_BEGIN(ctx->ctx->env, NULL, 0, &txn);
             }
             if (!rc) {
+                MDB_STAT_STEP(info, MDB_STAT_WRITE);
                 rc = MDB_PUT(txn, slot->dbi->dbi, &slot->key, &slot->data, 0);
             }
+            MDB_STAT_STEP(info, MDB_STAT_RUN);
             nextslot = slot->next;
             slapi_ch_free((void**)&slot);
         }
@@ -3737,7 +3744,9 @@ dbmdb_import_writer(void*param)
             break;
         }
         if  (count++ >= WRITER_MAX_OPS_IN_TXN) {
+            MDB_STAT_STEP(info, MDB_STAT_TXNSTOP);
             rc = TXN_COMMIT(txn);
+            MDB_STAT_STEP(info, MDB_STAT_RUN);
             if (rc) {
                 break;
             }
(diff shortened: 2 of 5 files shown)


progier389:
Adding an extract of error log to shows how it looks like:
```
[24/Nov/2022:19:00:09.367145216 +0100] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 859688 entries -- average rate 989.3/sec, recent rate 695.5/sec,
[24/Nov/2022:19:00:09.373848877 +0100] - INFO - dbmdb_import_monitor_threads - import userroot: Thread writer: run: 5.56% read: 6.17% write: 87.61% pause: 0.00% txnbegin: 0.00% txncommit: 0.65%
[24/Nov/2022:19:00:29.407574862 +0100] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 874958 entries -- average rate 984.2/sec, recent rate 676.1/sec,
[24/Nov/2022:19:00:29.414285549 +0100] - INFO - dbmdb_import_monitor_threads - import userroot: Thread writer: run: 5.54% read: 6.18% write: 87.62% pause: 0.00% txnbegin: 0.00% txncommit: 0.67%

```

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -4092,9 +4106,16 @@ dbmdb_import_init_writer(ImportJob *job, ImportRole_t role)
 void
 dbmdb_free_import_ctx(ImportJob *job)
 {
+    WorkerQueueData_t *s = NULL;
+    int i=0;
```
Firstyear: offsets to arrays are size_t. declare inside for loop as this allows further compiler optimisation.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -4108,3 +4129,87 @@ dbmdb_free_import_ctx(ImportJob *job)
...
+    char tmp[50];
+    int pos = 0;
+    int len = 0;
+    int i = 0;
+
+    if (!sinfo) {
+        return NULL;
+    }
+
+    for (i=0; i<MDB_STAT_LAST_STEP; i++) {
```
Firstyear: size_t. define i inline. Don't re-use iterator offsets.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -621,6 +621,14 @@ dbmdb_import_monitor_threads(ImportJob *job, int *status)
             sprintf(p, "recent rate %.1f/sec, ",
                          job->recent_progress_rate);
             import_log_notice(job, SLAPI_LOG_INFO, "dbmdb_import_monitor_threads", "%s", buffer);
+            for (current_worker = job->worker_list; current_worker != NULL; current_worker = current_worker->next) {
```
Firstyear: Each of these workers are seperate threads updating their status yes? There are no locks here, and certainly no atomics. So all of these workers will report incorrect/conflicting information. I'm concerned about this, because this will further add issues.

Firstyear (changes requested):

tbordaz (approved):
LGTM

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
progier389: Although you are right, the logged statistics may sometime not be exact, I am not  concerned about it:
As we are aggregating data on a quite long period of time (about 3 hours on a 10M entries import),
in most case (i.e last change missed or mismatch between second and nanoseconds) the error will not be perceptible ( a few seconds against a few thousands of seconds ) and in all causes it will not change the general trend.

Furthermore that is already the case with the way both bdb and lmdb monitoring threads
 computes the import progress statistics (no lock/atomic operation are used). 
I am much more concerned about adding lock ( or even atomic operation)  in the import writer thread that is already the one that is limiting the performance. 

That said, I will add a comment around the progress rate and mdb_stat_summarize log_notice telling that the statistics may be a bit off because of the lack of lock (so that we do not waste time if someone complains)

Firstyear: I'm not going to accept this as an answer. Just because the existing code does this, does not make it "correct". It means we have an opportunity to fix up past mistakes. 

Additionally, we are now seeing non-ordered arches like aarch64 become more prevalent and this code will have many more issues on those architectures.

Either user atomics to set and read the values, use a mutex with a try_lock on the reporter to prevent blocking the importer, or don't have stats at all.

This code would not pass tsan, it would probably have issues with asan if you push it. I don't think it meets the high standards we should have as a team for a high reliability system.

progier389: I am afraid that I disagree with your arguments:
   - "High quality standard" and "correct code" means adapting the solution to the needs and avoiding using a sledgehammer to break a nuts.
   - The non-ordered arches argument is irrelevant: Whatever the order is, a read will either returns the
     right value or the old one (and old is not so old compared to 1 second unless the hardware architecture is seriously broken).
     And we just do not care whether any of these value is the current or the old one.
   - ASAN will not report any issue as the memory is allocated and initialized.
   - IMHO We have a lot of work to do before getting a clean tsan report (a bit like the cleanup I have recently done 
      about covscan report) but it is easy to handle the assumed races (by adding a suppress comment) 
      (Anyway I think it is work we will have to do when working of getting a clean tsan status)

And none of the solutions you are proposing looks really appealing:
  - try_lock will not help. The issue is not about blocking (the monitoring thread was only trying to read the data every 20 seconds so 
    there will almost never be any lock contention) but we are speaking about locking or unlocking 300000 time per seconds
    and adding all these fences have an impact.
  - I do not like either the idea about not providing theses statistics:
    I felt the need to add these statistics to try to understand why in my test with 10M entries
    lmdb import is 3 time slower than bdb one (and the statistics helped to determine that ldbm db was the bottleneck)
    People will complain that the import is not fast enough (It always happens, it is just a recurring fact.)
    and I think we should have a way to determine whether the issue is the write performance or because for
    some reason data does not get provided fast enough.
    so IMHO it is better to have the info out of the box than working with the pressure of a customer.

That said as a conciliation effort:
 - I moved the calloc out of the writer thread and do it when initializing the thread info struct (i.e before creating the threads)
 - statistics are now logged (and read from the thread info struct) only once:
   after the threads have finished their job (so the data are no more moving).

I hope these two changes does help to ease your concerns

Firstyear: I'm not going to click approve on multi threaded code, that doesn't use basic thread synchronisation primitives. It's as simple as that. There are plenty of ways to make this correct, simple and fast. 

If you aim for fast above all else, you will forever sacrifice correctness and simplicity. Just because you *believe* there are no issues here today, does not make that true. There are multi-threaded code with atomics and locks that do 100,000's operations per second without problem. It can be done. 

Either add atomics, or use a lock with try_lock to prevent blocking on the workers/reader. 

I will not approve code that does not adhere to basic memory and thread safety principles.

progier389: This just does not make sense.
There is nothing to synchronize as reader thread now waits than the writer thread is dead before reading the data !!!

tbordaz: @progier389 , diving into that very technical discussion, I have a dummy question. My understanding is that the workers manage their own statistics. Then when all of them, have completed their jobs and are dead, another thread (provider, writer.. ?) reads their statistics, do a summary of them then log them. Am I correct ?

progier389: Exactly: 
Each worker_list threads (including the provider thread, the worker threads and the writer thread) update its own set of statistics.
and the main thread (i.e the import main thread) loops 
   to compute and log the import progression rate (that is the legacy code that I have not changed)
until all threads are completed 
then read the new statistic,  compute the ratio between the different statistics values then log them

tbordaz: Thanks @progier389 for your explanations, so it was a bit more complex than I initially thought.

My understanding is that there are two logs:  import statistics, import rate. 
I agree with @Firstyear that without sync between writers (workers/provider/writer) and reader (main import) there is a risk of invalid import statistics and potential others nasty things. I have no strong opinion between the two options: locking (with a risk of perf impact) and final logging without locking (only one status). In case final logging appear to be unsuffisient we could open a new ticket to implement intermediate logging, at the moment it looks an easy way to address both concerns: exact stat and no perf impact.

Regarding import rate, it is not modified by the PR. @firstyear, is it this part that you would like to improve ?

Firstyear: The problem here is that you can not have multi-threaded coded, synchronising data without using atomics or mutexes. You are risking introducing undefined behaviour, race conditions, crashes. 

I know you are a really exceptional engineer @progier389 but that doesn't mean we can flaunt the rules. When we work on a project like 389-ds we have to bring our very best to ensure the best results for our users and that means following "correct, simple, fast". Today the lack of synchronisation primitives violates "correct". 

Please just add mutexes here. You can use try-lock to avoid blocking. Performance here isn't a problem as already proven by my use of try-lock in the connection table previously.

progier389: Even if I still disagree with your arguments, I realized that since I moved the log of the statistics
 at the end of the import, there is no more any good reason to log them in the monitoring thread.
So I moved the log into the writing thread and now the statistics counters are pure local variables
 and we can stop that fruitless argument that is only delaying the fix of a reproducible crash !

Please, can you review the last changes ?

Firstyear (approved):
Thanks, since it's all in the one thread I'm happy here.

Issue #5547, opened 2022-11-27
automember plugin improvements


mreynolds389:
**Issue Description**

Previously the automember rebuild task you first go through all the automember rules and remove all the memberships, then it would rebuild them from scratch.  This was very expensive, and in the case where several be_txn plugins were enabled it would be even worse.

There was no restriction on how many fixup tasks can be run in parallel - there should only be one allowed at a time.

In the case of memberOf, or other be_txn plguins, being enabled, this could trigger automember plugin to be called more than once even though the user/group was already processed in the same thread.

tbordaz:
@mreynolds389 , any reason to keep the ticket opened. It looks the PR was pushed upstream a long time ago

mreynolds389:
> @mreynolds389 , any reason to keep the ticket opened. It looks the PR was pushed upstream a long time ago

It might have been left open for backporting, but I tihnk thats already happened.  I'll close this ticket for now.

PR #5548, opened 2022-11-28
Issue 5547 - automember plugin improvements

mreynolds389:
Description:

Rebuild task has the following improvements:

- Only one task allowed at a time
- Do not cleanup previous members by default.  Add new CLI option to intentionally cleanup memberships before rebuilding from scratch.
- Add better task logging to show fixup progress

To prevent automember from being called in a nested be_txn loop thread storage is used to check and skip these loops.  Seeing a 10% perf increase when memberof plugin is also enabled

relates: https://github.com/389ds/389-ds-base/issues/5547

### DIFF
--- a/ldap/servers/plugins/automember/automember.c
+++ b/ldap/servers/plugins/automember/automember.c
@@ -1,5 +1,5 @@
 /** BEGIN COPYRIGHT BLOCK
- * Copyright (C) 2011 Red Hat, Inc.
+ * Copyright (C) 2022 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -14,7 +14,7 @@
  * Auto Membership Plug-in
  */
 #include "automember.h"
-
+#include <pthread.h>
 
 /*
  * Plug-in globals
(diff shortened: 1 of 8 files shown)


mreynolds389:
Looks like there is a regression in basic_test.py, investigating...

mreynolds389:
> Looks like there is a regression in basic_test.py, investigating...

Issues fixed, please review...

jchapma (approved):
LGTM

ldap/servers/plugins/automember/automember.c
```diff
@@ -2285,9 +2351,22 @@ automember_task_add(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *eAfter __attr
...
+    PRBool cleanup = PR_FALSE;
 
     *returncode = LDAP_SUCCESS;
 
+    PR_Lock(fixup_lock);
+    if (fixup_running) {
+        PR_Unlock(fixup_lock);
+        *returncode = LDAP_UNWILLING_TO_PERFORM;
+        slapi_log_err(SLAPI_LOG_ERR, AUTOMEMBER_PLUGIN_SUBSYSTEM,
+                "automember_task_add - there is already a fixup task running\n");
```
droideck, line 2364: Could you please add a small test (or just a check to the existing test) - for this addition?
Something like - trying to add a few tasks at the same time, and checking if the error is displayed.

droideck (approved):
Besides the minor addition, looks good to me!

ldap/servers/plugins/automember/automember.c
mreynolds389: Done

Issue #5550, opened 2022-11-28
dsconf monitor dbmon crashes with "Error: math domain error"


Melkor333:
**Issue Description**
When trying to get monitoring info from dsconf, I get the following error:
```
dsconf -D ... -y ... ldap://localhost monitor dbmon
Error: math domain error
```
I traced the error back  to https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/cli_conf/monitor.py#L137 where I get a negative number for `dbcachefree`.
Reason for that negative number may be that `nsslapd-db-page-size` is not set and thus `pagesize = 8 * 1024` is assumed, which for some reason is wrong?
With pdb I printed out `ldbm_mon` which looks like the following, but I don't know how to debug this any further (e.g. get the actual pagesize, which isn't stored in the database):
```
 {'currentnormalizeddncachecount': ['579949'],
 'currentnormalizeddncachesize': ['97431432'],
 'dbcachehitratio': ['99'],
 'dbcachehits': ['89632254'],
 'dbcachepagein': ['347916'],
 'dbcachepageout': ['1010516'],
 'dbcacheroevict': ['104029'],
 'dbcacherwevict': ['748'],
 'dbcachetries': ['89980170'],
 'maxnormalizeddncachesize': ['104857536'],
 'normalizeddncacheevictions': ['0'],
 'normalizeddncachehitratio': ['98'],
 'normalizeddncachehits': ['42945211'],
 'normalizeddncachemisses': ['580440'],
 'normalizeddncachethreadsize': ['0'],
 'normalizeddncachethreadslots': ['579949'],
 'normalizeddncachetries': ['43525651'],
 'nsslapd-db-abort-rate': ['6917'],
 'nsslapd-db-active-txns': ['0'],
 'nsslapd-db-cache-hit': ['89632833'],
 'nsslapd-db-cache-region-wait-rate': ['32'],
 'nsslapd-db-cache-size-bytes': ['1610612736'],
 'nsslapd-db-cache-try': ['89980751'],
 'nsslapd-db-clean-pages': ['245901'],
 'nsslapd-db-commit-rate': ['3936166'],
 'nsslapd-db-configured-locks': ['10000'],
 'nsslapd-db-current-lock-objects': ['13'],
 'nsslapd-db-current-locks': ['12'],
 'nsslapd-db-deadlock-rate': ['0'],
 'nsslapd-db-dirty-pages': ['16'],
[...]

PR #5553, opened 2022-12-07
Issue 5550 - dsconf monitor crashes with Error math domain error

progier389:
Two issues around bdb pagesize handling:
1.  changelog database use the default page size instead of the configured page size because a
2. Exception in dsconf monitor dbconf is the dbcache is used at more than 50% because the "free space" is wrongly computed:  Using the configured page size while bdb mempool always used the file system preferred block size as page size.  
The Solution is:
1.  systematically set the page size after creating the db file handling before an db->open with  DB_CREATE flags
2. Add the mempool pagesize amongs the other monitoring statistics and use this value to compute dbmon dbcache statistics.  Otherwise use the file system preferred block size  

Issue: [5550](https://github.com/389ds/389-ds-base/issues/5550)

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
@@ -1426,6 +1426,37 @@ bdb_check_and_set_import_cache(struct ldbminfo *li)
     return 0;
 }
 
+/*
+ * Creates the db handler with the proper pagesize.
+ * This function should be systematically used instead of db_create when the handler
+ * is used to open a database.
+ * ( db_create may still be used when removing or renaming a database )
+ */
+int
+dbbdb_create_db_for_open(backend *be, const char *funcname, int open_flags, DB **dbp, DB_ENV *dbenv)
+{
+    struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    bdb_config *conf = (bdb_config *)li->li_dblayer_config;
+    int rval;
+
+    rval = db_create(dbp, dbenv, 0);
+    if (rval) {
+        slapi_log_err(SLAPI_LOG_ERR, funcname, "Unable to create db handler! %d\n", rval);
+        return rval;
+    }
+
+    if (open_flags & DB_CREATE) {
+        unsigned pagesize = (conf->bdb_page_size == 0) ? DBLAYER_PAGESIZE : conf->bdb_page_size;
+        rval = (*dbp)->set_pagesize(*dbp, pagesize);
+        if (rval) {
+            slapi_log_err(SLAPI_LOG_ERR,
+                          funcname, "dbp->set_pagesize(%" PRIu32 ") failed %d\n", pagesize, rval);
+            return rval;
+        }
+    }
+    return rval;
+}
+
 
 /* mode is one of
  * DBLAYER_NORMAL_MODE,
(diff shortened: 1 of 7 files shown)


ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -1914,18 +1924,8 @@ bdb_get_aux_id2entry_ext(backend *be, DB **ppDB, DB_ENV **ppEnv, char **path, in
         }
         *ppEnv = mypEnv->bdb_DB_ENV;
     }
-    rval = db_create(&dbp, *ppEnv, 0);
+    rval = dbbdb_create_db_for_open(be, "dblayer_get_aux_id2entry_ext", dbflags, &dbp, *ppEnv);
```
droideck, line 1927: Is it a typo in funcname argument? You have "dblayer_get_aux_id2entry_ext" but the function name it's called from is bdb_get_aux_id2entry_ext.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -2381,7 +2381,7 @@ bdb_get_db(backend *be, char *indexname, int open_flag, struct attrinfo *ai, dbi
 
     if (!ppDB)
         goto out;
-    return_value = db_create((DB**)ppDB, pENV->bdb_DB_ENV, 0);
+    return_value = dbbdb_create_db_for_open(be, "dblayer_open_file", open_flags, (DB**)ppDB, pENV->bdb_DB_ENV);
```
droideck, line 2384: Same here

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -2418,7 +2418,7 @@ bdb_get_db(backend *be, char *indexname, int open_flag, struct attrinfo *ai, dbi
                 dbp, NULL /* txnid */, abs_file_name, subname, DB_BTREE,
                 open_flags, priv->dblayer_file_mode, return_value);
         dbp->close(dbp, 0);
-        return_value = db_create((DB**)ppDB, pENV->bdb_DB_ENV, 0);
+        return_value = dbbdb_create_db_for_open(be, "dblayer_open_file", open_flags,(DB**)ppDB, pENV->bdb_DB_ENV);
```
droideck, line 2421: And here

src/lib389/lib389/cli_conf/monitor.py
```diff
@@ -125,16 +126,20 @@ def db_monitor(inst, basedn, log, args):
     report_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
     ldbm_mon = ldbm_monitor.get_status()
     dbcachesize = int(ldbm_mon['nsslapd-db-cache-size-bytes'][0])
-    if 'nsslapd-db-page-size' in ldbm_mon:
-        pagesize = int(ldbm_mon['nsslapd-db-page-size'][0])
+    if 'nsslapd-db-mp-pagesize' in ldbm_mon:
```
droideck, line 134: Just for clarity, could you please add a comment with an explanation here in the CLI code that mp = mempool?

droideck (approved):
Besides the very minor comments, looks great!

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
progier389: Not really a typo: for consistency I used the name that is used in all slapi_log_xxx calls in this function.
( It is an historical thing: when Ludwig split the dblayer code, he has chosen to keep the original function names in the log. I just used the same logic and reuse the same  old name.
IMHO we should keep the same name than the other logs within same function.
Now whether we should keep the old name or use the current name everywhere could be argued !

src/lib389/lib389/cli_conf/monitor.py
progier389: Done.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
droideck: > ( It is an historical thing: when Ludwig split the dblayer code, he has chosen to keep the original function names in the log. I just used the same logic and reuse the same old name.

Oh, I didn't know there's a historical reasoning there (and I missed that other slapi_log functions do the same).

> IMHO we should keep the same name than the other logs within same function.

For sure!

> Now whether we should keep the old name or use the current name everywhere could be argued !

I think it'll be nice to bring more consistency there because we have:

- `dblayer_xxx` functions with `dblayer_xxx`  log function strings
- `bdb_xxx` functions with `bdb_xxx`
- `bdb_xxx` functions with `dblayer_xxx` 
- `bdb_xxx` functions without `bdb` prefix in the log function string (i.e. `bdb_check_and_set_import_cache` with `check_and_set_import_cache`)

Maybe more; I haven't checked the whole file.
I think the inconsistency is a bit confusing, but it's clearly a task for a separate PR. :)

You have my ack for this one! Thanks!

Issue #5236, opened 2022-03-24
RFE - Add a specific modify form for groups, roles and CoS


droideck:
There're three part of this issue:

1. We should have a searchable/sortable table for group memberships when editing the group entry (objectclass groupOfName/groupsOfUniqueNames).
2. We should have a searchable/sortable table for nested roles when editing the Role entry.
3. And we should add a template, CoS attributes, CoS specifier functionality to CoS definition editing.

mreynolds389:
578dfdd98..fa25edc39  389-ds-base-2.2 -> 389-ds-base-2.2
ec7ce04ea..021417a90  389-ds-base-2.1 -> 389-ds-base-2.1
05347a99e..ba859c701  389-ds-base-2.0 -> 389-ds-base-2.0
7fd039713..3c49e8e74  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5556, opened 2022-12-10
Issue 5236 - UI - add specialized group edit modal

mreynolds389:
Description:

Added a modal for handling groups: viewing, adding and removing members

Revised overall project:

- "Search Base" handling using a label button was not intuitive.  Changed it to a more recognizable href.

- Made table "trash Can" icons visibly clickable

- Edit/add entry wizard, the big red "Empty Value!" label is no longer displayed while you edit the value

relates: https://github.com/389ds/389-ds-base/issues/5236

### DIFF
--- a/src/cockpit/389-console/README.md
+++ b/src/cockpit/389-console/README.md
@@ -162,7 +162,7 @@ To test changes to the 389-console plugin, you can set up links from your worksp
 
 - Create cockpit directory under user's home directory
 
-    mkdir ~/.local/share/cockpit/389-console
+    mkdir ~/.local/share/cockpit
 
 - Link your workspace directory
 
--- a/src/cockpit/389-console/src/css/ds.css
+++ b/src/cockpit/389-console/src/css/ds.css
@@ -88,6 +88,10 @@ td {
     color: #f0ab00 !important;
 }
 
+.ds-pf-green-color {
+    color: green !important;
+}
+
 .ds-editor-tree {
     max-height: 70vh;
     overflow: auto;
@@ -110,6 +114,10 @@ td {
 
 .ds-right-align {
     text-align: right;
+}
+
+.ds-left-align {
+    text-align: left !important;
 }
 
 .ds-label {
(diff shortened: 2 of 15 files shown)


droideck (changes requested):
When I try to edit the group with no users, I can click on 'No Members' checkbox and then click on Remove Selected members. And if I do so, all the hell breaks loose:

image

mreynolds389:
> When I try to edit the group with no users, I can click on 'No Members' checkbox and then click on Remove Selected members. And if I do so, all the hell breaks loose:

Haha, fixed!

droideck:
Oddly, I still experience the same issue when I try to remove the existing member from the group:

image

mreynolds389:
@droideck  sorry I can not reproduce this crash.  I can delete members from groups (all of them) with any issue.

mreynolds389:
>  @droideck sorry I can not reproduce this crash. I can delete members from groups (all of them) with any issue.

I still can not reproduce it, but I found what looked like a bug in that code.  So I fixed that,  but like I said I couldn't reproduce it.  Browser issue?  Anyway give it a shot and see how it works for you.

droideck:
> I still can not reproduce it, but I found what looked like a bug in that code. So I fixed that, but like I said I couldn't reproduce it. Browser issue? Anyway give it a shot and see how it works for you.

That was it! The issue is fixed!

I found another one after I got unblocked (probably a minor clean up issue).
The reproduction steps:

1. Click on Group - then Edit...
2. Add a member using the new interface
3. Click on 'Switch to Generic Interface'
4. Add some new attribute (like description)
5. Then strange things start to happen... (double fields)
image
7. And the result LDIF doesn't work:
image

mreynolds389:
> I found another one after I got unblocked (probably a minor clean up issue). The reproduction steps:

Should be fixed now...

droideck (approved):
All looks great now! Ack.

Happy Holidays, Mark!

PR #5559, opened 2022-12-12
Issue 5558 - non-root instance fails to start on creation

### LINKED ISSUE #5558: non-root instance fails to start on creation
**Issue Description**
[ Cloned from [RHBZ 2151864](https://bugzilla.redhat.com/show_bug.cgi?id=2151864)] 

Instance created after using dscreate ds-root does not start:
`Failed to start ns-slapd: "[08/Dec/2022:06:29:47.991585733 -0500] - ERR - slapd_bootstrap_config - The default password storage scheme could not be read or was not found in the file /home/user/test/etc/dirsrv/slapd-localhost/dse.ldif. It is mandatory.
`
That is because the libpwdchan-plugin password storage are erroneously disabled.

 - Platform: RedHat
 - Package and version: 389-ds-base-2.1.5-4.module+el9dsrv+16995+8a75ed16.x86_64

**Steps to Reproduce**

1. mkdir -p $HOME/bin
2. PATH=$HOME/bin:$PATH
3. dscreate ds-root $HOME/dsroot $HOME/bin
4. cat /tmp/inst.inf 
```
[general]
selinux = False
[slapd]
instance_name = localhost
root_password = password
port = 1389
secure_port = 1636
[backend-userroot]
create_suffix_entry = True
sample_entries = yes
suffix = dc=example,dc=com
```
5. hash -r
6. $ dscreate from-file /tmp/inst.inf

**Expected results**
Instance should be created successfully:
`Completed installation for instance: slapd-localhost`

progier389:
issue: non root installation fails to start after the default storage scheme change.
The solution is to avoid removing the RUST storage scheme from dse.ldif templates while preparing the non root user installation.

Issue:  [5558](https://github.com/progier389/389-ds-base/issues/5558)

Reviewed by: @mreynolds389

### DIFF
--- a/src/lib389/lib389/cli_ctl/instance.py
+++ b/src/lib389/lib389/cli_ctl/instance.py
@@ -227,9 +227,7 @@ def prepare_ds_root(inst, log, args):
         log.debug(f'Copying {bin} into {destbin}')
         os.makedirs(os.path.dirname(destbin), 0o755, True)
         copyfile(bin, destbin)
-    # And finally, update the template files
-    for dse in ('/usr/share/dirsrv/data/template-dse-minimal.ldif', '/usr/share/dirsrv/data/template-dse.ldif'):
-        copy_and_skip_entry(dse, ('libpwdchan-plugin',))
+    # And finally, update the template file
     # Use PO and PF to escape { } in formatted strings
     PO = '{'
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
LGTM

PR #5563, opened 2022-12-13
Issue 5413 - Allow mutliple MemberOf fixup tasks with different bases and filters

mreynolds389:
Description:

A change was made to only allow a single fixup task at a time, but there are cases where you would want to run mutliple tasks but on different branches/filters.

Now we maintain a linked list of bases/filters of the current running tasks to monitor this.

relates: https://github.com/389ds/389-ds-base/issues/5413

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -52,7 +52,6 @@ static Slapi_DN* _pluginDN = NULL;
 MemberOfConfig *qsortConfig = 0;
 static int usetxn = 0;
 static int premodfn = 0;
-static PRBool fixup_running = PR_FALSE;
 static PRLock *fixup_lock = NULL;
 static int32_t fixup_progress_count = 0;
 static int64_t fixup_progress_elapsed = 0;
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/memberof/memberof.c
```diff
@@ -2894,14 +2909,38 @@ memberof_fixup_task_thread(void *arg)
...
+    mo_fixup_ll *prev = NULL;
+    for (mo_fixup_ll *curr = fixup_list; curr; curr = curr->next) {
+        mo_fixup_ll *next = curr->next;
+        if (slapi_sdn_compare(curr->sdn, sdn) == 0 &&
+            strcasecmp(curr->filter_str, td->filter_str) == 0)
+        {
+            if (prev == NULL) {
+                /* first node */
+                fixup_list = next;
+                slapi_sdn_free(&curr->sdn);
```
tbordaz: I think those 3 lines are common, whether prev==NULL or not. You may move them after the test.

tbordaz (approved):
ACK

ldap/servers/plugins/memberof/memberof.c
mreynolds389: I disagree, but could be wrong.  If prev is NULL, then this is the first node, and we need to reset the global head pointer (fixup_list).  We would not want to do that is "prev" has been set.

mreynolds389: Oh wait, I think this was a github issue.  I think you mean the three lines below this (the ones that free curr).  Yes those can go above.  Revising fix...

PR #5569, opened 2022-12-15
Update specfile and rust crates

mreynolds389:
Update the local specfile and cargo crates to make internal test builds a little easier


droideck (approved):
LGTM!

PR #5570, opened 2022-12-15
Issue 3615 - CLI - prevent virtual attribute indexing

mreynolds389:
Description:

Do not allow virtual attributes to be indexed because it breaks search results

relates: https://github.com/389ds/389-ds-base/issues/3615


droideck (approved):
Looks good!

Issue #5278, opened 2022-05-04
`dsidm` asks for the old password on password reset


vashirov:
**Issue Description**

```
[root@server ds]# dsidm localhost --basedn dc=example,dc=com account change_password uid=ldap_user,ou=people,dc=example,dc=com
Enter current password for uid=ldap_user,ou=people,dc=example,dc=com : <IF YOU ENTER HERE ANYTHING, IT FAILS>
Enter new password for uid=ldap_user,ou=people,dc=example,dc=com :
CONFIRM - Enter new password for uid=ldap_user,ou=people,dc=example,dc=com :
Error: 120 - 4 - -1 - Can't contact LDAP server - [] - Invalid oldPasswd value.

[root@server ds]# dsidm localhost --basedn dc=example,dc=com account change_password uid=ldap_user,ou=people,dc=example,dc=com
Enter current password for uid=ldap_user,ou=people,dc=example,dc=com : <IF YOU PRESS ENTER, IT WORKS>
Enter new password for uid=ldap_user,ou=people,dc=example,dc=com :
CONFIRM - Enter new password for uid=ldap_user,ou=people,dc=example,dc=com :
changed password for uid=ldap_user,ou=people,dc=example,dc=com

```

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: 389-ds-base-2.0.15-1.fc35.x86_64

**Steps to Reproduce**
1. Add a user
```
dsidm localhost --basedn dc=example,dc=com user create --uid ldap_user --cn ldap_user --displayName ldap_user --uidNumber 1001 --gidNumber 1001 --homeDirectory /home/ldap_user
```
2. Try to change their password:
```
dsidm localhost --basedn dc=example,dc=com account change_password 
```

**Expected results**
When password is changed by DM, old password should not be prompted.

mreynolds389:
Actually you are supposed to use "reset_password" in this case, but "change_password" should be more robust anyway:

f2a4664b9..578dfdd98  389-ds-base-2.2 -> 389-ds-base-2.2
52de69bf4..ec7ce04ea  389-ds-base-2.1 -> 389-ds-base-2.1
31c413089..05347a99e  389-ds-base-2.0 -> 389-ds-base-2.0
3764f44bf..7fd039713  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5575, opened 2022-12-16
Issue 5278 - CLI - dsidm asks for the old password on password reset

mreynolds389:
Description:  If we are chaning a password as Root DN we don't need the old password

relates: https://github.com/389ds/389-ds-base/issues/5278

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -65,7 +65,8 @@ from lib389.utils import (
     format_cmd_list,
     get_default_db_lib,
     selinux_present,
-    selinux_label_port)
+    selinux_label_port,
+    get_user_is_root)
 from lib389.paths import Paths
 from lib389.nss_ssl import NssSsl
 from lib389.tasks import BackupTask, RestoreTask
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -3484,3 +3485,17 @@ def restore_online(self, archive, db_type=None):
         task.create(properties=task_properties)
 
         return task
+
+    def is_rootdn_bound(self):
+        # Return True if root DN is authenticated for this DirSrv instance
+        if self.binddn is None and get_user_is_root():
```
progier389, line 3494: This code is fine in dsidm case but I think we should improve it to avoid future trouble if someone reuse the function:
   only the fact that is instance is connected ensure that "bind dn is None
      " means that "ldapi was successfully used".
    so I would add a test to return False if  self.state != DIRSRV_STATE_ONLINE
    otherwise root user specifying no bind dn to remote instance may be perceived as root_dn bound.
Except that minor point, LGTM

progier389 (approved):
LGTM except a minor point

src/lib389/lib389/__init__.py
mreynolds389: fixed

Issue #5560, opened 2022-12-12
dscreate run by non superuser set defaults requiring superuser privilege.


progier389:
**Issue Description**
[Clone of [RHBZ 2151948](https://bugzilla.redhat.com/show_bug.cgi?id=2151948)]
When dscreate is run by  a non superuser (typically after using dscreate ds-root) , some of the default value set by dscreate are still requiring superuser privileged. 
The default for selinux, port and secureport options should be changed.

"selinux" option should be False (meaning that if selinux  is globally enforced on the machine then the instance will use the user default selinux context (i.e: without any special privileges)    
port and secureport should be port openable by the user.

**Package Version and Platform:**
 - Platform: Redhat
 - Package and version: 389-ds-base-2.1.5-4.module+el9dsrv+16995+8a75ed16.x86_64

**Steps to Reproduce**

1. mkdir -p $HOME/bin
2. PATH=$HOME/bin:$PATH
3. dscreate ds-root $HOME/dsroot $HOME/bin
4. cat /tmp/inst.inf
```
[slapd]
instance_name = localhost
root_password = password
[backend-userroot]
create_suffix_entry = True
sample_entries = yes
suffix = dc=example,dc=com
```
5. hash -r
6. dscreate from-file /tmp/inst.inf

**Expected results**
Instance should be created successfully:
Completed installation for instance: slapd-localhost

PR #5579, opened 2022-12-20
Issue 5560 - non root dscreate set defaults requiring superuser privilege

progier389:
Problem:
 when run by a non root user some of dscreate default values of the template options are requiring superuser privileges.
Solution: 
 Changing the default value for port, secure_port, selinux, systemd when uid is not 0

Issue: [5560](https://github.com/389ds/389-ds-base/issues/5560)

Reviewed by: @droideck


droideck (approved):
LGTM!

PR #5584, opened 2023-01-03
Issue 3604 - Create a private key/CSR with dsconf/Cockpit

jchapma:
Description: There's no way to create a private key and a CSR with dsconf/Cockpit. However, features for importing a certificate exists, but DS also requires the private key in the NSS database to use the certificate

Fix Description: Modify dsconf/UI to allow creation of a CSR.

relates: https://github.com/389ds/389-ds-base/issues/3604

Reviewed by: ??

### DIFF
--- a/src/cockpit/389-console/src/lib/security/certificateManagement.jsx
+++ b/src/cockpit/389-console/src/lib/security/certificateManagement.jsx
@@ -21,6 +21,7 @@ import {
     SecurityAddCertModal,
     SecurityAddCACertModal,
     SecurityAddCSRModal,
+    SecurityViewCSRModal,
 } from "./securityModals.jsx";
 import PropTypes from "prop-types";
 import { log_cmd } from "../../lib/tools.jsx";
@@ -38,14 +39,23 @@ export class CertificateManagement extends React.Component {
             showEditModal: false,
             showAddModal: false,
             showAddCSRModal: false,
+            showViewCSRModal: false,
             modalSpinning: false,
             showConfirmDelete: false,
             showCSRConfirmDelete: false,
             showKeyConfirmDelete: false,
             certName: "",
             certFile: "",
-            csrSubject: "",
+            csrContent: "",
             csrName: "",
+            csrSubject: "",
+            csrSubjectCommonName: "",
+            csrSubjectOrg: "",
+            csrSubjectOrgUnit: "",
+            csrSubjectLocality: "",
+            csrSubjectState: "",
+            csrSubjectCountry: "",
+            csrSubjectEmail: "",
             keyID: "",
             flags: "",
             _flags: "",
@@ -71,6 +81,7 @@ export class CertificateManagement extends React.Component {
 
         this.addCACert = this.addCACert.bind(this);
         this.handleChange = this.handleChange.bind(this);
+        this.handleCSRChange = this.handleCSRChange.bind(this);
         this.addCert = this.addCert.bind(this);
         this.showAddModal = this.showAddModal.bind(this);
         this.closeAddModal = this.closeAddModal.bind(this);
@@ -78,6 +89,8 @@ export class CertificateManagement extends React.Component {
         this.closeAddCAModal = this.closeAddCAModal.bind(this);
         this.showAddCSRModal = this.showAddCSRModal.bind(this);
         this.closeAddCSRModal = this.closeAddCSRModal.bind(this);
+        this.showViewCSRModal = this.showViewCSRModal.bind(this);
+        this.closeViewCSRModal = this.closeViewCSRModal.bind(this);
         this.showEditModal = this.showEditModal.bind(this);
         this.closeEditModal = this.closeEditModal.bind(this);
         this.showEditCAModal = this.showEditCAModal.bind(this);
@@ -91,6 +104,7 @@ export class CertificateManagement extends React.Component {
         this.delCert = this.delCert.bind(this);
         this.addCSR = this.addCSR.bind(this);
         this.delCSR = this.delCSR.bind(this);
+        this.showCSR = this.showCSR.bind(this);
         this.delKey = this.delKey.bind(this);
         this.closeConfirmDelete = this.closeConfirmDelete.bind(this);
         this.closeCSRConfirmDelete = this.closeCSRConfirmDelete.bind(this);
@@ -99,6 +113,7 @@ export class CertificateManagement extends React.Component {
         this.reloadCACerts = this.reloadCACerts.bind(this);
         this.reloadCSRs = this.reloadCSRs.bind(this);
         this.reloadOrphanKeys = this.reloadOrphanKeys.bind(this);
+        this.buildSubject = this.buildSubject.bind(this);
     }
 
     showAddModal () {
@@ -134,7 +149,8 @@ export class CertificateManagement extends React.Component {
     showAddCSRModal () {
         this.setState({
             showAddCSRModal: true,
-            errObj: { csrName: true, csrSubject: true }
+            csrSubject: "",
+            errObj: { csrName: true, csrSubjectCommonName: true},
         });
     }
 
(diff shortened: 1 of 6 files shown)


mreynolds389:
Security CSR modal had a duplicate header.  This fixes it:

```
diff --git a/src/cockpit/389-console/src/lib/security/securityModals.jsx b/src/cockpit/389-console/src/lib/security/securityModals.jsx
index cac880008..7fc93a32f 100644
--- a/src/cockpit/389-console/src/lib/security/securityModals.jsx
+++ b/src/cockpit/389-console/src/lib/security/securityModals.jsx
@@ -243,12 +243,6 @@ export class SecurityAddCSRModal extends React.Component {
                 ]}
             >
                 <Form isHorizontal autoComplete="off">
-                    <TextContent>
-                        <Text component={TextVariants.h4}>
-                            Create a Certificate Signing Request
-                        </Text>
-                    </TextContent>
-                    <hr />
                     <Grid title="Enter Certificate Signing Request name">
                         <GridItem className="ds-label" span={3}>
                             CSR Name
```

You should add "titles" to each field describing what the field is and what the accepted values are.  

Also, the form seems very incomplete.  Usually there are fields for:  country, locality, server hostname(cn), etc.  Please look into "Certificate signing requests", and make sure the form follows this layout.  For example:  https://www.globalsign.com/en/blog/what-is-a-certificate-signing-request-csr  Then the form should construct the subject DN for you.

Also after creating a request I can not view it.  Or copy its contents.  

Finally I see after a fresh install I already have an "orphaned" key.  Is this expected?  The orphaned tab should describe what an orphaned key is.  Its not clear if it should be deleted or not, etc.

jchapma:
> You should add "titles" to each field describing what the field is and what the accepted values are.
> 
> Also, the form seems very incomplete. Usually there are fields for: country, locality, server hostname(cn), etc. Please look into "Certificate signing requests", and make sure the form follows this layout. For example: https://www.globalsign.com/en/blog/what-is-a-certificate-signing-request-csr Then the form should construct the subject DN for you.
> 
> Also after creating a request I can not view it. Or copy its contents.

These are fixed.

jchapma:
> Finally I see after a fresh install I already have an "orphaned" key. Is this expected? The orphaned tab should describe what an orphaned key is. Its not clear if it should be deleted or not, etc.

After a fresh install there should be no orphan keys, i just tried a fresh install and it seems ok. I added explanation on what an orphan key is.

Thanks

mreynolds389:
FYI I got that error above form an instance created by the CI tests.  So it's probably is an odd state, but the UI should be able to handle that situation without getting stuck.

Then I would remove the info/success alert when displaying the CSR.   
Then under the CSR tab just get ride of that info message (it looks funny anyway)
Under the orphan key tab change that info message to a tiptool or someting similiar.  See Monitor Tab -> Replication  there are two different ways to do it.  See the "Synchronization report" where the info is like a drop down, or Conflict Entries where its a hoverable info message.

Then we have an issue creating a CSR.  The modal "save" button is spinning when you open it, and you can not click it.  Make sure if one of the required fields is blank that hte save button is disabled as well.

Screenshot from 2023-01-11 14-43-51

mreynolds389:
Also bonus points for displaying the generated DN in the modal _as_ you fill in the fields.  See the ACI editor as an example, but I would put the dynamically generated DN at the bottom of the modal.

jchapma:
> FYI I got that error above form an instance created by the CI tests. So it's probably is an odd state, but the UI should be able to handle that situation without getting stuck.

I developed it on a ssca instance so there were always keys in the NSS DB, but yes we should handle that.

jchapma:
> Also bonus points for displaying the generated DN in the modal _as_ you fill in the fields. See the ACI editor as an example, but I would put the dynamically generated DN at the bottom of the mod

Yes I noticed that in a referral modal i think, nice.

jchapma:
> Make sure if one of the required fields is blank that the save button is disabled as well.

Name and CN are the only required fields, i tested all this last night before i pushed and it worked fine, let me revisit.

mreynolds389:
> > Make sure if one of the required fields is blank that the save button is disabled as well.
> 
> Name and CN are the only required fields, i tested all this last night before i pushed and it worked fine, let me revisit.

The problem is that save button is already spinning/saving when you open the modal.  So it's impossible to click it.  It should be disabled until all required fields are entered.

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -165,6 +241,105 @@ export class CertificateManagement extends React.Component {
...
+        log_cmd("addCSR", "Creating CSR", cmd);
+        cockpit
+                .spawn(cmd, { superuser: true, err: "message" })
+                .done(() => {
+                    this.reloadCSRs();
+                    this.setState({
+                        showAddCSRModal: false,
+                        csrSubject: '',
+                        csrName: '',
+                        modalSpinning: false
```
mreynolds389: Should loading be set to false here?  Not sure

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -165,6 +241,105 @@ export class CertificateManagement extends React.Component {
...
+        ];
+
+        log_cmd("showCSR", "Displaying CSR", cmd);
+        cockpit
+                .spawn(cmd, { superuser: true, err: "message"})
+                .done(content => {
+                    this.setState({
+                        csrContent: content,
+                        showViewCSRModal: true,
+                        modalSpinning: true,
```
mreynolds389: modalSpinning needs to be set to false here.  After viewing a CSR the modal is still set as spinning.  This is why the create csr modal is spinning when I open it

mreynolds389: Actually modalSpinning is not even used in CSR viewing, so it should just be removed as it serves no purpose.

mreynolds389: same with "loading", you don't need to set it.

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
jchapma: The behaviour is the same when set to true/false, to be consistent with "addCert" I will leave it as is.

jchapma:
When the country code field of a CSR subject is populated, it has to comply with ISO 3166-1 alpha-2, if not certutil complains and returns an error. Currently we catch this error and inform the user and let them correct the issue.

A more fancy solution is to provide a drop-down with all the country codes, but this means importing another npm package (iso-3166-1-alpha-2). WDYT ?

mreynolds389 (approved):
LGTM but I know @droideck wanted to review it too...

droideck:
When I add more than one CSR, the result subject combines old and new fields.
You need to do a clean-up each time the new form is opened.

image

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -165,6 +241,98 @@ export class CertificateManagement extends React.Component {
...
+                        "success",
+                        `Successfully created CSR`
+                    );
+                })
+                .fail(err => {
+                    const errMsg = JSON.parse(err);
+                    if (errMsg.desc.includes('certutil -s: improperly formatted name:')) {
+                        this.props.addNotification(
+                            "error",
+                            `Error Incorrect County Code`
```
droideck: The message is incorrect because the wrong format is not always about the Country Code:

image

droideck (approved):
A couple of easy-fix comments, but the rest looks great!

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
jchapma: > The message is incorrect because the wrong format is not always about the Country Code:

From a certutil perspective, the only subject field it imposes a format on is the country code. So all the other subject fields accept anything, except a few special characters I didn't test with. Good catch.

PR #5587, opened 2023-01-04
Issue 5585 - lib389 password policy DN handling is incorrect

### LINKED ISSUE #5585: lib389 password policy DN handling is incorrect
**Issue Description**
After a migration between major DS versions, it can happen that already existing password policies will have 'cn' that contains a valid DN in double quote "<DN>".  We need to strip the quotes before processing the DN with python-ldap.

droideck:
Description: After a migration between major DS versions, it can happen that already existing password policies will have 'cn' that contains a valid DN in double quote "". We need to strip the quotes before processing the DN with python-ldap.

Fixes: https://github.com/389ds/389-ds-base/issues/5585

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/password/password_policy_test.py
+++ b/dirsrvtests/tests/suites/password/password_policy_test.py
@@ -13,23 +13,28 @@ This test script will test password policy.
 import os
 import pytest
 import time
+from lib389.config import Config
 from lib389.topologies import topology_st as topo
+from lib389.topologies import topology_m1
 from lib389.idm.domain import Domain
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccounts, UserAccount
 from lib389._constants import DEFAULT_SUFFIX
-from lib389.pwpolicy import PwPolicyManager
+from lib389.pwpolicy import PwPolicyManager, PwPolicyEntries
+from lib389.idm.account import Account
+from lib389.idm.nscontainer import nsContainers
+from lib389.cos import CosPointerDefinitions, CosTemplates
 import ldap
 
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/password/password_policy_test.py
```diff
@@ -1378,6 +1438,74 @@ def test_additional_corner_cases(topo, _policy_setup, _fixture_for_additional_ca
...
+                  f'uid=joe,ou=People,{DEFAULT_SUFFIX}'):
+        pwp.create_user_policy(dn_dn, policy_props)
+
+    # The function creates PwPolicyEntry with cn: "<DN>" value instead of <DN>
+    create_subtree_policy_custom(inst, f'ou=People,{DEFAULT_SUFFIX}', policy_props)
+
+    # And now, try to get the unusual subtree pwpolicy
+    people = pwp.get_pwpolicy_entry(f'ou=people,{DEFAULT_SUFFIX}')
+    people.replace('passwordhistory', 'off')
+    assert people.get_attr_val_utf8('passwordhistory') == 'off'
```
tbordaz, line 1520: Would it be possible to delete users/pw_policy in teardown ?

tbordaz (approved):
Fix looks good to me (just a minor comment)

dirsrvtests/tests/suites/password/password_policy_test.py
droideck: Done, I've separated part of the function into the fixture and have added a subtree removal as a tearDown there.

mreynolds389 (approved):
Ack

Issue #5588, opened 2023-01-05
CI - fix tests


mreynolds389:
**Issue Description**

 Several ACL test are failing because the tests  expect an INDEX exception, but now the server returns an empty list instead of the exception

mreynolds389:
> Looking at test_deny_group_member_all_rights_to_group_members failure:
> 
> ```
>     user = UserAccount(conn, 'uid=test_user_1000,ou=ACLGroup,dc=example,dc=com')
>     with pytest.raises(IndexError):
>         user.get_attr_val_utf8('uid')
> ```
> 
> This seems a bit weird: according to [RFC 4517](https://www.rfc-editor.org/rfc/rfc4517#section-3.3.6), value with DirectoryString syntax should have at least 1 character, so the test seems OK. (uid should not be empty) I think that we should double check if the server is really returning an attribute with an empty value (which would then be a server bug) or if it is a lib389 issue around UserAccount handling.

I think lib389 is just handling the error differently now.  It's just testing if an aci is denying access to read "uid"

mreynolds389:
51b0e501f..7baa5a15c  389-ds-base-2.2 -> 389-ds-base-2.2
17ac5d2f5..d8ddb4ac3  389-ds-base-2.1 -> 389-ds-base-2.1

mreynolds389:
23111609a..27df56c42  389-ds-base-2.0 -> 389-ds-base-2.0
5a5630d8d..015c9abdf  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5589, opened 2023-01-05
Issue 5588 - Fix CI tests

mreynolds389:
Description:

Fix ACL tests that no longer return IndexError but instead return empty list Fix db_home tests when in container and regular db dir is used instead Fix repl monitor test where args_instance was not declared before treating it as dict

relates: https://github.com/389ds/389-ds-base/issues/5588


droideck (approved):
Looks good!
Only one left (some flaky replication issue, as it looks from first glance) and we'll have full green :)

Issue #5593, opened 2023-01-06
 dsidm account subtree-status fails with TypeError 


mreynolds389:
**Issue Description**

Running "dsidm account subtree-status" command fails with TypeError.

**Steps to Reproduce**

    #  dsidm -D "cn=Directory Manager" ldap://localhost -b "dc=example,dc=com" account subtree-status "ou=People,dc=example,dc=com" -f "(uid=*)"

Actual results:

DEBUG: expected list of strings or None
Traceback (most recent call last):
  File "/usr/sbin/dsidm", line 140, in <module>
    result = args.func(inst, dsrc_inst['basedn'], log, args)
  File "/usr/lib/python3.9/site-packages/lib389/cli_idm/account.py", line 87, in subtree_status
    account_list = Accounts(inst, basedn).filter(filter, scope)
  File "/usr/lib/python3.9/site-packages/lib389/_mapped_object.py", line 1264, in filter
    results = self._instance.search_ext_s(
  File "/usr/lib/python3.9/site-packages/lib389/__init__.py", line 174, in inner
    return f(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 863, in search_ext_s
    msgid = self.search_ext(base,scope,filterstr,attrlist,attrsonly,serverctrls,clientctrls,timeout,sizelimit)
  File "/usr/lib/python3.9/site-packages/lib389/__init__.py", line 174, in inner
    return f(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 853, in search_ext
    return self._ldap_call(
  File "/usr/lib/python3.9/site-packages/lib389/__init__.py", line 174, in inner
    return f(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 324, in _ldap_call
    result = func(*args,**kwargs)
TypeError: expected list of strings or None
ERROR: Error: expected list of strings or None

mreynolds389:
8fa65330f..c81ee3486  389-ds-base-2.2 -> 389-ds-base-2.2
bc03cb9bd..9398b8656  389-ds-base-2.1 -> 389-ds-base-2.1
a63e22627..ff6dbd909  389-ds-base-2.0 -> 389-ds-base-2.0
be8e4189c..74ebe9873  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5594, opened 2023-01-06
Issue 5593 - CLI - dsidm account subtree-status fails with TypeError

mreynolds389:
Description: The was a problem with the function parameters being passed to the filter method, and we were missing a parameter for the print entry function.

relates: https://github.com/389ds/389-ds-base/issues/5593


droideck (approved):
LGTM!

Issue #5581, opened 2022-12-21
Dark mode support


vashirov:
**Issue Description**
Cockpit supports dark mode since v269.
When enabled, our application has issues with the some of the components, see screenshots.

**Steps to Reproduce**
1. Go to Session (top right corner) and switch style to Dark, if not set automatically.

**Expected results**
Main banner and tree view for Server, Database, Replication, Monitoring tabs look consistent with the rest of the UI.

**Screenshots**
<img width="1717" alt="SCR-20221221-los-2" src="https://user-images.githubusercontent.com/1745018/208931002-c1d8cd41-d279-4604-b335-8fa7d8e4c721.png">

PR #5601, opened 2023-01-10
Issue 5581 - UI - Support cockpit dark theme

mreynolds389:
Description:  We were previously overriding the background color which caused
              issues with the dark theme.

relates: https://github.com/389ds/389-ds-base/issues/5581

### DIFF
--- a/src/cockpit/389-console/src/css/ds.css
+++ b/src/cockpit/389-console/src/css/ds.css
@@ -1,6 +1,5 @@
 .ds-logo {
     margin: 5px;
-    background-color: white !important;
 }
 
 .ds-logo-style {
@@ -62,10 +61,8 @@ td {
 }
 
 .ds-tree {
-    font-size: 0.5;
-    background-color: #f8f8f8;
     border: 1px solid #909090;
-    color: black;
+    background-color: #f8f8f8;
     width: 300px;
     min-width: 300px;
     min-height: 350px;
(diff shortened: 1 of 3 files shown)


droideck:
Looks great and clean!
I have one minor concern only - the red text colour (for example, during instance creation)
I think it'll look much cleaner if we use the lighter one (it's already used in other places in Cockpit with Dark mode):

Basically, we use `--pf-global--danger-color--100` - `#fe5142`. And the rest of the cockpit uses `--pf-global--danger-color--200` - `#ff7468`
image

Ideally, we need to change the colour of the `(!)` icon too (it's also `#ff7468` color in the rest of the cockpit)

Current colour:
image

Suggested colour:
image

mreynolds389:
@droideck - oh ok you want to override PF classes.  I tried inspecting that element and I couldn't find what sets the color.  How are you overriding it?

mreynolds389:
> @droideck - oh ok you want to override PF classes. I tried inspecting that element and I couldn't find what sets the color. How are you overriding it?

Nevermind I found it

mreynolds389:
@droideck - its all fixed.  Also found a bug when editing groups that would crash the browser and also updates Database tree (replaced the "Create Suffix" button with a  tree action item icon).

droideck (approved):
Looks great! Ack

Issue #5602, opened 2023-01-12
[Cockpit 389ds plugin] Create a new custom Entry bug.


cold01water:
**Issue Description**
An error occurs when selecting an object class in "Create a new custom Entry" from the LDAP browser.

**Package Version and Platform:**
 - Platform: Almalinux9
 - Package and version: github master branch (2023/1/12)
 - Browser: Edge

**Steps to Reproduce**
the above

**Expected results**
You should be able to select object classes and add entries.

**Additional context**

The problem is that the following code is trying to assign to a "const" variable.
If you change it to "let", it will work fine.

* 389-ds-base/src/cockpit/389-console/src/lib/ldap_editor/wizards/operations/addLdapEntry.jsx

```
444             for (const attr of required) {
445                 attr = attr.trim().toLowerCase();

(snip)

496             for (const attr of optional) {
497                 attr = attr.trim();
```

mreynolds389:
Yes we found a similar bug recently with editLdapEntry.  This code used to work (incorrectly) but now browsers have apparently gotten stricter and are now exposing these bugs.  We'll get this fixed asap.

PR #5603, opened 2023-01-12
Issue 5602 - UI - browser crash when trying to modify read-only variable

mreynolds389:
Description:  

Existing code that used to work (incorrectly) is now causing issues.  Need to use "let" instead of "const".

relates: https://github.com/389ds/389-ds-base/issues/5602


droideck (approved):
LGTM!

Issue #5598, opened 2023-01-09
In 2.x, SRCH throughput drops by 10% because of handling of referral


tbordaz:
**Issue Description**
The fix #5170 (filter optimization), [adds](https://github.com/389ds/389-ds-base/commit/90d847426bf64cb5f2f2fd957f928be01d0500f8#diff-320edfdbeeb2154cac972f7a43736ad108c2f2a8ce91926ee4b20e50a46d1e2cR1027) '(objectclass=referral)' to the original filter. This forces to reevaluate the filter on each candidate.
During performance campaign, this reduces the throughput by 10% 2.1 vs 1.4.3 

**Package Version and Platform:**
- since 2.1

**Steps to Reproduce**
hammer the server subtree search with an equality search filter like (uid=user1234).

**Expected results**
The benefit of named reference (referral) is quite limited (although standard) and should not impact the performance

tbordaz:
This bug is revealed by fix #5170  that is only present since 2.1. So target milestone is 2.1.
Need to check that 1.4.3 is not impacted by #5170

tbordaz:
create_subtree_filter adds (objectclass=referral) to the original filter. It is not clear to me which RFC requires that the server returns referrals. For example I failed to find in this [RFC 3296](https://www.ietf.org/rfc/rfc3296.txt) this requirement.

Anyway I think that the problem needs to be fixed in can_skip_filter_test/grok_filter. For example skip filter evaluation (OR) if  all components are equality.

progier389:
About  can_skip_filter_test/grok_filter, I think that we should skip filter evaluation or OR and AND if 
 we can skip all subfilters 
 
 About [RFC 3296](https://www.ietf.org/rfc/rfc3296.txt):
 As I interpret [Section 2.1](https://datatracker.ietf.org/doc/html/rfc3296#section-2.1),
 if ManageDsaIT control is set (i.e SLAPI_MANAGEDSAIT set in the pblock) then referral entries should be handled as regular entries, but if it is not the case,  all the referrals entries matching the search scope should be returned as search referrals ( without trying to apply the search filter on them) 
 So internally to generate the idlist, the OR '(objectclass=referral)'  make sense if  ManageDsaIT control is not set
  When searched entries are added in the result, they are added as referral or as plain entries 
 But if the control is not set both searched entries and referral will be returned as plain entries
   so having the subfilter in that case is a bug.
 
 A last point:
[...]

PR #5604, opened 2023-01-13
Issue 5598 - In 2.x, SRCH throughput drops by 10% because of handling…

### LINKED ISSUE #5170: Filter optimiser

tbordaz:
… of referral

Bug description:
	A part of the fix #5170 append '(objectclass=referral)' to
	the original filter (in subtree scope) in order to conform
        smart referral support https://www.ietf.org/rfc/rfc3296.txt
	This triggers a drop on SRCH throughput (10%).
	#5598 limits the case when '(objectclass=referral)' is added
         - Most of the time a server does not contain smart referral.
           So most of the time it is useless to  add that subfilter
         - It should not be added for internal searches

Fix description:
	A mechanism periodically (each 30s) checks if there are
	smart referral entries (referral_check).
	When a first smart referral is detected or no more smart
	referral is detected it logs a information message.
	During a direct subtree search, 'objectclass=referral' is
	append at the condition it exists at least a referral

relates:  #5598

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/backend.c
+++ b/ldap/servers/slapd/backend.c
@@ -17,6 +17,7 @@
 #include "nspr.h"
 
 static PRMonitor *global_backend_mutex = NULL;
+static Slapi_Eq_Context referral_check_ctx = NULL;
 
 void
 be_init(Slapi_Backend *be, const char *type, const char *name, int isprivate, int logchanges, int sizelimit, int timelimit)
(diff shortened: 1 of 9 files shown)


ldap/servers/slapd/backend.c
```diff
@@ -204,6 +206,108 @@ slapi_be_gettype(Slapi_Backend *be)
...
+{
+    if (referral_check_ctx) {
+        slapi_eq_cancel_rel(referral_check_ctx);
+        referral_check_ctx = NULL;
+    }
+}
+int
+slapi_exist_referrals(void)
+{
+    return exist_referrals;
```
mreynolds389: Do we need a lock here?   I also recall that "exist_referrals" should be a static uint64_t in this case (something to do safety but I don't remember the details anymore)

tbordaz: Thanks for the review Mark. I think it does not need a lock. The update is periodic so if the caller reads an outdated value it is not a big deal.
Regarding the size, it looks a good point. Indeed the cache line will be in competition between workers, I will do test with cache line alignment.

progier389:
I suspect that we also have a regression on search rate result, so why optimizing only the internal search ?

progier389:
Maybe we should  rather have one flag per backend: 
 It is not more costly and it avoid having a referral one one backend impacting all the other backend

mreynolds389:
Several tests are crashing the server.  Not sure if it's related to this PR or not.

progier389:
> Maybe we should rather have one flag per backend:
> It is not more costly and it avoid having a referral one one backend impacting all the other backend

About your valid remark about nested suffix: IMHO we should keep things simple in the search request and just look at the current backend flags but in the task setting the flags if we set the flags to true we should also walk the parent suffixes and also set their flag.

ldap/servers/slapd/backend.c
tbordaz: Performance wise, there is no benefit when making it uint64_t and aligning it on cache size

ldap/servers/slapd/backend.c
```diff
@@ -204,6 +206,108 @@ slapi_be_gettype(Slapi_Backend *be)
...
+    char *filter;
+    char *cookie;
+    int rc;
+    int exist_referral = 0;
+    LDAPControl **server_ctrls;
+
+    /* loop over the backends to check if a it exists a "smart" referral */
+    filter = "(objectclass=referral)";
+    be = slapi_get_first_backend(&cookie);
+    while (!exist_referral && be) {
```
Firstyear: The problem with this logic is that if a referral is added then removed, then the check won't run and we'll continue to assume referrals exist.

ldap/servers/slapd/backend.c
Firstyear: It's probably not needing a lock, but an atomic to set it with RELEASE and to check the exist with ACQUIRE is correct.

ldap/servers/slapd/backend.c
```diff
@@ -204,6 +206,108 @@ slapi_be_gettype(Slapi_Backend *be)
...
+}
+
+/* schedule periodic call to referral_check */
+void
+slapi_referral_check_init(void)
+{
+    referral_check_ctx = slapi_eq_repeat_rel(referral_check,
+                                             NULL, (time_t)0,
+                                             /* 30 sec in milliseconds */
+                                             30 * 1000);
```
Firstyear: I'd even just make this 60 seconds or 5 minutes, I don't think this needs to be checked so frequently.

Firstyear (changes requested):

ldap/servers/slapd/backend.c
progier389: Although using lock or atomic operation as described by @Firstyear is the standard way to set and check flags in different threads, in this specific case I agree with Thierry: it is an overkill.
We do not really care of the timing (as @Firstyear said in another comment we could check if the flag  once every few minutes (so if the change stay in the cpu pipe line until next context switch, does not really matter)

ldap/servers/slapd/backend.c
tbordaz: Not sure to understand your point.
ATM the fix set the flag per server, not per backend. So if a referral exists somewhere, then it stops iterating over the others backends and set the flag.
Now I will rework the fix to make it per backend as @progier389 suggested.

progier389: The logic is ok, but I suspect that @Firstyear was confused by the way you named the variables:
- exist_referral for the local variable
- exist_referrals for the global one
And I also found this choice a bit confusing, IMHO you should have used more different names

tbordaz: Right, it was confusing. I reworked the patch with a per suffix flag. So we no longer have a global 'exist_referrals.
Note that the new patch also suffers from non protected access to the backend flag. I think it is not a big deal even if we miss that flag setting as it will be set again few sec after.

tbordaz:
@Firstyear , I changed the logic of referral_check to be per backend. If a backend contains a smart referral, it sets a flag. That flag is tested on subtree search (slapi_exist_referrals). Is it addressing your requested changes ?

progier389:
I wonder if we should not move the check of existing referral for a backend in a separate function 
 and also call it just after enabling a backend. ( So we will get a more predictive behavior after an import for example )

mreynolds389:
@Firstyear - Thierry make some changes for you, could you please review them?  Thanks!!

Firstyear (approved):

Firstyear:
Sorry about the review delay :)

tbordaz:
> I wonder if we should not move the check of existing referral for a backend in a separate function and also call it just after enabling a backend. ( So we will get a more predictive behavior after an import for example )

@progier389 , the check of the referral is in function 'slapi_exist_referrals(be)'. Is it what you asked for ?

tbordaz:
I made the change (internal search+flags) in slapi_exist_referral.

PR #5606, opened 2023-01-13
Issue 5605 - Adding a slapi_log_backtrace function in libslapd

### LINKED ISSUE #5605: Adding a slapi_log_backtrace function in libslapd
Sometime (for temporary debugging build or when some critical error are detected)
   it is interesting to log the current thread stack in the error log to help the troubleshooting.

In fact such a function already exists in db-mdb code , and the idea is simply to move this function in libslapd and rename it as slapi_log_backtrace

**Additional context**
I already needed to duplicate that function several time while troubleshooting some issue, 
so I think it should better be within libslapd so every module would be able to use it.
Furthermore later on we could improve some of the #ifdef debug code by logging the stack when something is wrong  (for example in the entry cache debug code when entry verify fails)

progier389:
Moving log_stack out of db-mdb code to libslapd and renaming it as slapi_log_backtrace.

Issue: [5605](https://github.com/389ds/389-ds-base/issues/5605)

Reviewed by: @mreynolds389

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_debug.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_debug.c
@@ -12,7 +12,6 @@
 #endif
 #include "mdb_layer.h"
 
-#include <execinfo.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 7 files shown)


mreynolds389 (approved):
LGTM

Issue #5608, opened 2023-01-17
[Cockpit 389ds plugin] Edit entry bug


cold01water:
**Issue Description**
An error occurs when trying to assign a value to a constant in editLdapEntry.jsx.

**Package Version and Platform:**
 - Platform: Almalinux
 - Package and version: master branch
 - Browser Edige

**Steps to Reproduce**
Edit ldap entry.

**Additional context**
Changing the following myVal variable to let works.

* editLdapEntry.jsx

```
 796         for (const datum of updateArray) {
 797             const myAttr = datum.attr;
 798             const myVal = datum.val; //here
 799             const isUserPwd = myAttr.toLowerCase() === "userpassword";
 800             isFilePath = false;

```

mreynolds389:
9b858c608..b15a20598  389-ds-base-2.2 -> 389-ds-base-2.2
51e66263e..6195ce9da  389-ds-base-2.1 -> 389-ds-base-2.1
1e36b3407..12a100f18  389-ds-base-2.0 -> 389-ds-base-2.0
2e8fbde0c..497f721f7  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

Ok think I've fixed all the const vs let issues that have been popping up recently

PR #5612, opened 2023-01-19
Issue 5608 - UI - need to replace some "const" with "let"

mreynolds389:
Description:  Browsers are becoming for strict about jvascript and there were places where a const varaible was being modified.  This is now crashing the server.  So to be overly cautious a lot of consts were changed just to be safe.

relates: https://github.com/389ds/389-ds-base/issues/5608


droideck (approved):
LGTM!

Issue #5565, opened 2022-12-13
Need to change default password storage schema for backwards compatibility


mreynolds389:
**Issue Description**

389-ds-base-1.3.x does not support the Rust password hashers.  If you replicate between 389-ds-1.4.x (or higher) and 1.3.x then all authentication will start failing on 1.3.x.

Until a solution is found for 1.3.x we need to change the default password storage scheme back to the C version:  PBKDF2_SHA256  (vs current Rust version PBKDF2-SHA512)

mreynolds389:
This needs to be done on 1.4.3 right now!  There is a good chance we need to do it for 2.x, but we can wait before making that decision.  

@Firstyear we can discuss in more detail offline if you want.  We do have some things we are looking into (including getting Rust to work on RHEL 7 - not as easy as it sounds but it's the least disruptive of all of the options).

Firstyear:
Is there a reason the customer/user in this case can't just manually configure their hashes on the 1.4 and greater servers to match 1.3?

mreynolds389:
> Is there a reason the customer/user in this case can't just manually configure their hashes on the 1.4 and greater servers to match 1.3?

Of course they can, but the problem is that the default behavior will unexpectedly break customers who have replication set up with older versions (and it won't be easy to fix).

droideck:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2170224

PR #5616, opened 2023-01-20
 Issue 5565 - Add Rust password storage PBKDF2 schemes

droideck:
Description: 389-ds-base-1.3.x does not support the Rust password hashers. If you replicate between 389-ds-1.4.x (or higher) and 1.3.x then all authentication will start failing on 1.3.x.
Backport Rust PBKDF2 implementation to 1.3.10.

Related: https://github.com/389ds/389-ds-base/issues/5565

Reviwed by: ?


droideck:
@Firstyear check it out. Rust in 1.3.10!

Firstyear:
> @Firstyear check it out. Rust in 1.3.10!

Scary 😳

Firstyear (approved):
Looks really good :) well done!

tbordaz:
The support of PBKDF2_SHA512 in 1.3.10 is a requirement to enable it by default in 1.4.3 (RHEL8). However is it a requirement to make it the default hash in 1.3.10 ?
I agree that eventually it will be the default hash scheme (in all versions) but from a deployment pov I see no benefit to force a CU to update a full 1.3.10 topology. We can imagine an admin that applies this patch only on a subset of servers.

droideck:
> The support of PBKDF2_SHA512 in 1.3.10 is a requirement to enable it by default in 1.4.3 (RHEL8). However is it a requirement to make it the default hash in 1.3.10 ? I agree that eventually it will be the default hash scheme (in all versions) but from a deployment pov I see no benefit to force a CU to update a full 1.3.10 topology. We can imagine an admin that applies this patch only on a subset of servers.

Couple of things to consider:

- The initial decision to update happened because of security/performance reasons, more covered here: https://github.com/389ds/389-ds-base/issues/5356
- We don't have the "password hash upgrade on bind" feature in 1.3.10, so it won't be a full 1.3.10 topology upgrade, as only the new users' passwords will be hashed with PBKDF2-SHA512. 

So I think it's really more of a management decision of the direction we want it present to our users:

1. Leave SSHA512 in 1.3.10 but switch to PBKDF2-SHA512 in 1.4.3 and 2.x. And recommend they set PBKDF2-SHA512 in their 1.3.10 topologies.
2. Or move every branch to PBKDF2-SHA512 as a default.

@mreynolds389, what do you think?

mreynolds389:
> So I think it's really more of a management decision of the direction we want it present to our users:
> 
>     1. Leave SSHA512 in 1.3.10 but switch to PBKDF2-SHA512 in 1.4.3 and 2.x. And recommend they set PBKDF2-SHA512 in their 1.3.10 topologies.
> 
>     2. Or move every branch to PBKDF2-SHA512 as a default.
> 
> 
> @mreynolds389, what do you think?

Do not make PBDFK2 the default in 1.3.x.  It just needs to be supported.  You should never change default behavior this late into a release (especially PBKDF2 as it significantly slows down a bind and this would catch a lot of customers off guard after just doing an update).

Thanks!

droideck:
Sure thing!
The changes are made. Please, review. (And I'm running tests once again, just in case)

tbordaz (approved):

droideck:
I accidentally merged the old commit message `Issue 5565 - Change default password storage scheme to PBKDF2-SHA512 (#5616)`. The actual code doesn't change the default storage scheme.
@mreynolds389 if you'll be doing the downstream build, please, be aware of that... Sorry for the inconvenience. :(

Issue #5497, opened 2022-10-20
Boolean LDAP Attribute Syntax too scrict


apateczechmeister:
**Issue Description**
[RFC 4517 #3.3.3](https://www.rfc-editor.org/rfc/rfc4517#section-3.3.3) defines the `Boolean` Attribute with OID 1.3.6.1.4.1.1466.115.121.1.7 in the ABNF

    Boolean = "TRUE" / "FALSE"

[RFC 4517 #3.3.3](https://www.rfc-editor.org/rfc/rfc4517#section-3.3.3) refers to [RFC 4512 #1.4](https://www.rfc-editor.org/rfc/rfc4512#section-1.4) for the syntax definition, which further refers to [RFC 4234](https://www.rfc-editor.org/rfc/rfc4234) which defines ABNF as an *case insensitive* syntax - as does [RFC 7405 #1](https://datatracker.ietf.org/doc/html/rfc7405#section-1) which enhances RFC 4234 with a case sensitive description.

**Package Version and Platform:**
 - RHEL 7.9
 - Package and version: 389-ds-base-1.3.10.2-16.el7_9.x86_64
 - Browser firefox

**Steps to Reproduce**
1. To a default installation add the following schema definition
````
[user@host ~]$ sudo -- cat /etc/dirsrv/slapd-ldap1-dev/schema/99user.ldif
# user schema
################################################################################
dn: cn=schema
#
################################################################################
#
attributetypes: ( 1.3.6.1.4.1.59369.1.1
  NAME 'emailVerified'
  DESC 'User has successfully verified his/her current email'
  SYNTAX 1.3.6.1.4.1.1466.115.121.1.7
  SINGLE-VALUE
  USAGE userApplications
  X-ORIGIN ('APA-IT defined' 'user defined') )
#
################################################################################
#
objectclasses: ( 1.3.6.1.4.1.59369.2.1
  NAME 'apaitIamPerson'
  SUP inetOrgPerson
  AUXILIARY
  MAY (emailVerified)
  X-ORIGIN ('APA-IT defined' 'user defined') )
[user@host ~]$
````
2. Create at some path an `apaitIamPerson` object without our new attribute
4. Try to run `ldapmodify` with the following input
````
[user@host ~]$ cat meins.ldif 
dn: uid=user,ou=users,dc=top
changetype: modify
replace: emailVerified
emailVerified: false
-
[user@host ~]$ 
````
5. ds-389 will respond with
````
[...]

PR #5622, opened 2023-01-26
Issue 5497 - boolean attributes should be case insensitive

mreynolds389:
Description:  
Boolean values are supposed to be case insensitive, but in our code it is sensitive even though the code is in the "cis" file.

relates: https://github.com/389ds/389-ds-base/issues/5497


droideck (approved):
LGTM!

Issue #5624, opened 2023-02-01
RFE - UI - export certificates, and import text base64 encoded certificates 


mreynolds389:
Description of problem:

It would be good to have the following options in certificate management:

1. Import of a certificate using copy-paste of base64 encoded cert or uploading from the file system. Currently it only allows to import from the existing file on a server filesystem.
2. Export of CA or Server cert. Currently there is no way to export them from the Console.

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=1751264

mreynolds389:
7b8652862..11d366aaa  389-ds-base-2.2 -> 389-ds-base-2.2
d4fcfa84e..7d8d7d8b9  389-ds-base-2.1 -> 389-ds-base-2.1
ad8ab413c..0a77fae73  389-ds-base-2.0 -> 389-ds-base-2.0
bb7209634..00f42cfc8  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5625, opened 2023-02-01
Issue 5624 - RFE - UI - export certificates, and import text base64 encioded certificates

mreynolds389:
Description:

All exporting CA and server certificates
Allow importing a certificate by pasting the base64 encoded certificate text into a form

relates: https://github.com/389ds/389-ds-base/issues/5624

### DIFF
--- a/src/cockpit/389-console/src/lib/security/certificateManagement.jsx
+++ b/src/cockpit/389-console/src/lib/security/certificateManagement.jsx
@@ -19,9 +19,9 @@ import {
 import {
     EditCertModal,
     SecurityAddCertModal,
-    SecurityAddCACertModal,
     SecurityAddCSRModal,
     SecurityViewCSRModal,
+    ExportCertModal,
 } from "./securityModals.jsx";
 import PropTypes from "prop-types";
 import { log_cmd } from "../../lib/tools.jsx";
@@ -38,14 +38,17 @@ export class CertificateManagement extends React.Component {
             tableKey: 0,
             showEditModal: false,
             showAddModal: false,
+            showAddCAModal: false,
             showAddCSRModal: false,
             showViewCSRModal: false,
             modalSpinning: false,
             showConfirmDelete: false,
             showCSRConfirmDelete: false,
             showKeyConfirmDelete: false,
+            showExportModal: false,
             certName: "",
             certFile: "",
+            certText: "",
             csrContent: "",
             csrName: "",
             csrSubject: "",
@@ -65,6 +68,55 @@ export class CertificateManagement extends React.Component {
             loading: false,
             modalChecked: false,
             disableSaveBtn: true,
+            exportNickname: "",
+            exportDERFormat: false,
+            exportFileName: "",
+            certRadioFile: false,
+            certRadioText: false,
+            certRadioSelect: true,
+            certRadioUpload: false,
+            availCertNames: [],
+            selectCertName: "",
+            isSelectCertOpen: false,
+            // Upload PEM file
+            uploadValue: "",
+            uploadFileName: "",
+            uploadIsLoading: false,
+            uploadIsRejected: false,
+        };
+
+        // File Upload functions
+        this.handleFileInputChange = (e, file) => {
+            this.setState({
+                uploadFile: file.name
+            });
+        };
+        this.handleTextOrDataChange = (value) => {
+            this.setState({
+                uploadValue: value
+            });
+        };
+        this.handleFileReadStarted = () => {
+            this.setState({
+                uploadIsLoading: true
+            });
+        };
+        this.handleFileReadFinished = () => {
+            this.setState({
+                uploadIsLoading: false
+            });
+        };
+        this.handleClear = () => {
+            this.setState({
+                certText: "",
+                uploadFile: "",
+                uploadIsRejected: false,
+            });
+        };
+        this.handleFileRejected = () => {
+            this.setState({
+                uploadIsRejected: true,
+            });
         };
 
         this.handleNavSelect = (event, tabIndex) => {
@@ -79,13 +131,18 @@ export class CertificateManagement extends React.Component {
             return sorted_flags.join('');
         };
 
-        this.addCACert = this.addCACert.bind(this);
+        this.handleCertSelect = (value) => {
+            this.setState({
+                selectCertName: value,
+            });
+        };
+
         this.handleChange = this.handleChange.bind(this);
         this.handleCSRChange = this.handleCSRChange.bind(this);
         this.addCert = this.addCert.bind(this);
         this.showAddModal = this.showAddModal.bind(this);
-        this.closeAddModal = this.closeAddModal.bind(this);
         this.showAddCAModal = this.showAddCAModal.bind(this);
+        this.closeAddModal = this.closeAddModal.bind(this);
         this.closeAddCAModal = this.closeAddCAModal.bind(this);
         this.showAddCSRModal = this.showAddCSRModal.bind(this);
         this.closeAddCSRModal = this.closeAddCSRModal.bind(this);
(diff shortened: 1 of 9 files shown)


src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -192,44 +230,45 @@ export class CertificateManagement extends React.Component {
...
-            loading: true,
         });
         const cmd = [
             "dsconf", "-j", "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket",
-            "security", "certificate", "add", "--name=" + this.state.certName, "--file=" + this.state.certFile
+            "security", "export-cert", this.state.exportNickname, "--output-file=" + this.state.exportFileName
         ];
-        log_cmd("addCert", "Adding server cert", cmd);
+        if (this.state.exportDERFormat) {
+            cmd += "--binary-format";
```
droideck: It fails here because `cmd` is const. And I wonder if we'll have more `const` issues in other places as I see a lot of `const`s (not added by this PR but in general)

src/cockpit/389-console/src/lib/security/securityModals.jsx
```diff
@@ -10,35 +10,44 @@ import {
...
+        let exportBtnName = "Export Certificate";
         const extraPrimaryProps = {};
         if (spinning) {
-            saveBtnName = "Adding Certificate ...";
-            extraPrimaryProps.spinnerAriaValueText = "Saving";
+            exportBtnName = "Exporting Certificate ...";
+            extraPrimaryProps.spinnerAriaValueText = "Exporting";
         }
 
+        const title = <>Export Certificate: &nbsp;&nbsp;<i>{nickName}</i></>;
```
droideck, line 47: When we do 'Export Certificate', it's not clear to me where it exports and how a user can find it later.
Maybe we should add something like 'Backup List'? Or, at least, report the cert path where the user can find the file.

And when we import the certificate, should we add something like 'file browser' button? I don't have a strong opinion here though...

droideck (changes requested):

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
mreynolds389: Nice catch, its not failing in my browser!!!  And I think I fixed al lthe const issues in recent commits, but yes there could be more I suspose.

droideck: Chrome Version 109.0.5414.74 (Official Build) (64-bit) here :)

src/cockpit/389-console/src/lib/security/securityModals.jsx
mreynolds389: > When we do 'Export Certificate', it's not clear to me where it exports and how a user can find it later. Maybe we should add something like 'Backup List'? Or, at least, report the cert path where the user can find the file.

Let me think about this, but yes it needs to be improved.

> 
> And when we import the certificate, should we add something like 'file browser' button? I don't have a strong opinion here though...

Isn't the file browser only for the client machine?  The import needs to be a file local to the server.  I could list certs that are already present in the nssalpd-certdir (/etc/dirsrv/slapd-inst).

droideck: > Isn't the file browser only for the client machine? The import needs to be a file local to the server. I could list certs that are already present in the nssalpd-certdir (/etc/dirsrv/slapd-inst).

Ideally, I think we need a way to import the certs from nssalpd-certdir and from the client machine.
The second is possible but only with the cert copy past. And I think it'll be nice to offer an option with a file browser. Probably, something like this (FileUpload component):
https://github.com/389ds/389-ds-base/blob/main/src/cockpit/389-console/src/lib/ldap_editor/lib/editableTable.jsx#L824

mreynolds389: @droideck - all request features implemented, please review...

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -192,44 +280,45 @@ export class CertificateManagement extends React.Component {
...
+        let cmd = [
             "dsconf", "-j", "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket",
-            "security", "certificate", "add", "--name=" + this.state.certName, "--file=" + this.state.certFile
+            "security", "export-cert", this.state.exportNickname, "--output-file=" + this.state.exportFileName
         ];
-        log_cmd("addCert", "Adding server cert", cmd);
+        if (this.state.exportDERFormat) {
+            cmd += "--binary-format";
+        }
+        log_cmd("exportCert", "Exporting certificate", cmd);
```
droideck, line 343: Some odd things are going on here... Browser crashes with the next error:
image

src/cockpit/389-console/src/lib/security/securityModals.jsx
```diff
@@ -191,10 +237,116 @@ export class SecurityAddCertModal extends React.Component {
...
+                                <Select
+                                    variant={SelectVariant.single}
+                                    aria-label="Select Input"
+                                    onToggle={handleCertToggle}
+                                    onSelect={handleCertSelect}
+                                    selections={selectCertName}
+                                    isOpen={isSelectCertOpen}
+                                    aria-labelledby="certname"
+                                >
+                                    {certNames.map((cert, index) => (
```
droideck: The 'Choose Cerificate From Server' list is not updated when we export the certificate (so now it's required to update the page manually with F5).

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -192,44 +280,45 @@ export class CertificateManagement extends React.Component {
...
+    exportCert () {
         this.setState({
             modalSpinning: true,
-            loading: true,
         });
-        const cmd = [
+        let cmd = [
             "dsconf", "-j", "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket",
-            "security", "certificate", "add", "--name=" + this.state.certName, "--file=" + this.state.certFile
+            "security", "export-cert", this.state.exportNickname, "--output-file=" + this.state.exportFileName
```
droideck: If I specify `exportFileName` as a "test_cert" - it will name the file exactly like that. Hence, making it undiscoverable by the UI later.

Should we additionally add `.crt` extension there if the user hasn't specified it?

droideck (changes requested):
And one more thing I've noticed is that `Choose Cerificate From Server` is not selectable. See the video:
[Screencast from 02-06-2023 10:01:58 AM.webm](https://user-images.githubusercontent.com/5634892/217050184-2ccf5729-5d62-4796-834d-29d3d81a584d.webm)

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
mreynolds389: Fixed

src/cockpit/389-console/src/lib/security/securityModals.jsx
mreynolds389: Fixed, please review...

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -192,44 +302,57 @@ export class CertificateManagement extends React.Component {
...
                 .done(() => {
                     this.reloadCACerts();
                     this.setState({
-                        showAddModal: false,
-                        certFile: '',
-                        certName: '',
-                        modalSpinning: false
+                        showExportModal: false,
+                        exportNickname: '',
+                        modalSpinning: false,
```
droideck, line 352: I think we need to call `this.getCertFiles();` somewhere here, so the cert import selects are updated after the export.

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -239,24 +362,199 @@ export class CertificateManagement extends React.Component {
...
+
+            log_cmd("addCert", "creating tmp cert file for importing: ", create_cert_cmd);
+            cockpit
+                    .spawn(create_cert_cmd, { superuser: true, err: "message" })
+                    .done(() => {
+                        const cmd = [
+                            "dsconf", "-j", "ldapi://%2fvar%2frun%2fslapd-" + this.props.serverId + ".socket",
+                            "security", certType, "add", "--name=" + this.state.certName, "--file=" + certFile
+                        ];
+                        log_cmd("addCert", "Adding cert (tmp): ", cmd);
```
droideck, line 463: Another thing, it doesn't set appropriate flags when I click `Add CA Certificate`. So it's always added to Server Cert table.
See the video (I've also shown the import select issue I mentioned in another comment):
[Screencast from 02-06-2023 01:31:52 PM.webm](https://user-images.githubusercontent.com/5634892/217093292-19a2da75-b1e1-49d0-8380-02e3e05ef603.webm)

droideck (approved):
A couple of minor issues, but the rest looks awesome! I love the amount of options we can add the certs now. :)

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
mreynolds389: > I think we need to call `this.getCertFiles();` somewhere here, so the cert import selects are updated after the export.

Fixed

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
mreynolds389: Ah I keep introducing regressions.  It opened the server cert modal, that's why it did not have the right flags.

Fixed

src/cockpit/389-console/src/lib/security/certificateManagement.jsx
```diff
@@ -239,24 +363,211 @@ export class CertificateManagement extends React.Component {
...
+
+    addCert (isCACert) {
+        this.setState({
+            modalSpinning: true,
+            loading: true,
+        });
+
+        let certType = "certificate";
+        if (isCACert) {
+            certType = "ca-certificate";
```
droideck, line 443: Okay, sorry, it's once again me with another error I found...

When I export in binary->delete old->add the exported = it fails for me with the next error:
`Danger alert:
Error adding certificate - 'utf-8' codec can't decode byte 0x82 in position 1: invalid start byte`

The video:
[Screencast from 02-06-2023 03:12:04 PM.webm](https://user-images.githubusercontent.com/5634892/217108760-24065802-3b02-4fe6-8b9c-33d2d9725722.webm)

The same is true for Server Cert.

mreynolds389: Yeah perhaps certutil is not what you use for .crt files?  pk12util?   Maybe different flags are needed?  Not sure I'll look into it tomorrow.

mreynolds389: Ok the problem is that lib389 expects that we only work with PEM files.   A minor tweak to lib389 fixed this.  Please review one last time (hopefully)  :-)

droideck: For some reason, it still doesn't work with CA certs... (Server certs are fine now!)

droideck: Looking at the code, may be it's because `add_ca_cert_bundle` works with PEM (hence, failing with binary import):
https://github.com/389ds/389-ds-base/blob/b8c90a5a757099f466326c1ffc68cb89c74c0b14/src/lib389/lib389/nss_ssl.py#L1233-L1240

mreynolds389: Ah yes, fixed now...

droideck (approved):
LGTM! Ack!

PR #5627, opened 2023-02-01
Issue 4577 - Add LMDB pytest github action

progier389:
As @vashirov and @bsimonova discovered in PR [5357,](https://github.com/389ds/389-ds-base/pulls/5357),
 adding a step in the pytest workflow is not working because test are not run 
if a bdb test fails. So I propose to test another approach:
 Duplicate the pytest workflow and it seems to work nicely (The github limit of 20 concurrent jobs prevents that too many tests run at the same time)
 Furthermore as the bdb and lmdb  tests run on different containers it ensure a proper test separations.

Issue [4577](https://github.com/389ds/389-ds-base/issues/4577)

Reviewed by @vashirov, @bsimonova , Thanks !

### DIFF
--- a/.github/workflows/lmdbpytest.yml
+++ b/.github/workflows/lmdbpytest.yml
@@ -1,6 +1,10 @@
 name: LMDB Test
 
 on:
+  push:
+  pull_request:
+  schedule:
+    - cron:  '0 0 * * *'
   workflow_dispatch:
     inputs:
       pytest_tests:
@@ -11,10 +15,6 @@ on:
         description: 'Set to "true" to enable debugging with tmate (https://github.com/marketplace/actions/debugging-with-tmate)'     
         required: false
         default: false
-  workflow_run:
-    workflows: [Test]
-    types:
-      - completed
 
 jobs:
   build:
--- a/.github/workflows/lmdbpytest.yml
+++ b/.github/workflows/lmdbpytest.yml
@@ -1,0 +1,115 @@
+name: LMDB Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      pytest_tests:
+        description: 'Run only specified suites or test modules delimited by space, for example "basic/basic_test.py replication"'
+        required: false
+        default: false
+      debug_enabled:
+        description: 'Set to "true" to enable debugging with tmate (https://github.com/marketplace/actions/debugging-with-tmate)'     
+        required: false
+        default: false
+  workflow_run:
+    workflows: [Test]
+    types:
+      - completed
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-22.04
+    container:
+      image: quay.io/389ds/ci-images:test
+    outputs:
+        matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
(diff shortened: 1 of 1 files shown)


progier389:
A few tests were killed by the 6 hour timeout which is strange (I wonder if there is not a deadlock somewhere).
Unfortunately there is no much detail in such cases.
Created [RFE issue 5628 ](https://github.com/389ds/389-ds-base/5627)to improve the troubleshooting in case of timeout

bsimonova:
Good approach of the issue. 
I am a bit sad that I didn't have an opportunity to get back to my original PR, given all the tasks we had to handle.
Thank you Pierre for looking into this :)

So far the changes look good to me but I'll wait for @vashirov as he has best experience with github actions.

vashirov:
Given that the only difference with the pytest action is setting NSSLAPD_DB_LIB=mdb variable, I'd suggest to look into composite actions:
https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

Having said that, I won't prevent you from merging this PR, as this can be refactored later.

vashirov (approved):

bsimonova (approved):

progier389:
> Given that the only difference with the pytest action is setting NSSLAPD_DB_LIB=mdb variable, I'd suggest to look into composite actions: https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/ https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
> 
> Having said that, I won't prevent you from merging this PR, as this can be refactored later.

Hi @vashirov, 
FYI: I wondered the same before generating the PR and my feeling was that it is not worth to refactor that code.:
    it would probably be cheaper to handle duplicate action code than to refactor it:
One of the reason is that the bdb code and test will get removed in a few years - so the duplication is temporary.
A second reason is that the action code is not modified very often so duplicating the change is not expensive
The third reason is that during my work on [ansible_ds repository actions](https://github.com/389ds/ansible-ds/pull/24) 
I discovered that changing the action code is always a bit costly (error messages are obscure and quite often misleading, and blind coding with error and retry method is quite slow (and frustrating! -;) )) 
 I spent a full work day when I decided to use composite action to build the ansible_ds collection and IMHO it will even be much slower in this repository because the time to build ds packages (several minutes) is much longer than the one to build ansible_ds collection (a few seconds))

So even if I agree that using composite action is nicer and theoretically better, I do not think it is pragmatic in this case.

Anyway, I agree with you that if we finally decide to refactor the code, we could do it later on 
and merging this PR will let us get the lmdb test results in the mean time.

progier389:
Duplicate the pytest workflow and it seems to work nicely (The github limit of 20 concurrent jobs prevents that too many tests run at the same time)
Furthermore as the bdb and lmdb tests run on different containers it ensure a proper test separations.

Issue 4577

Reviewed by @vashirov, @bsimonova , Thanks !
Note: There is a discussion whether we should re factorize or not these actions but if we ever decide to do it, we could do it later one.

PR #5633, opened 2023-02-07
Issue 5588 - CI - fix tests

bsimonova:
Description:
Small fixes for 1.4.3 branch.
setup_ds_test had wrong brackets for parametrize and instead of 'false' value, it was testing 'f', 'a' etc. values. 
ssl_version_test - adjusting the ds version as TLS 1.0 and TLS 1.1 are on LEGACY policy on RHEL 8.8

Relates: https://github.com/389ds/389-ds-base/issues/5588

Reviewed by: @aadhikar (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/tls/ssl_version_test.py
+++ b/dirsrvtests/tests/suites/tls/ssl_version_test.py
@@ -76,7 +76,8 @@ def test_ssl_version_range(topo):
         pass
 
     if not skip:
-        if ds_is_older('1.4.4'):
+        # TLS 1.0 and TLS 1.1 were moved to LEGACY on 1.4.3, adjusting the sanity check
+        if ds_is_older('1.4.3'):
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/tls/ssl_version_test.py
```diff
@@ -76,7 +76,8 @@ def test_ssl_version_range(topo):
         pass
 
     if not skip:
-        if ds_is_older('1.4.4'):
+        # TLS 1.0 and TLS 1.1 were moved to LEGACY on 1.4.3, adjusting the sanity check
+        if ds_is_older('1.4.3.28'):
```
aadhikar: I think it should be 1.4.3

aadhikar: rest LGTM!

PR #5639, opened 2023-02-08
Issue 2435 - RFE - Raise IDL Scan Limit to INT_MAX

### LINKED ISSUE #2435: Raise idlistscanlimit by default
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/49376
- Created at 2017-09-08 08:25:45 by [firstyear](https://pagure.io/user/firstyear) (@Firstyear)

---

#### Issue Description
With https://pagure.io/389-ds-base/issue/49372 we try to avoid hitting large IDLs by fixing the filters.

There are some cases left we may access IDLs and in those cases, we probably want to return an IDL anyway.

One case is if we hit the idlistscanlimit in an OR query. In this case we return ALLIDS to the parent filter, which likely inturn raises this back to the parent. This means that we do a very expensive id2entry scan. Had we of return an IDL, even if the IDL had say 100,000 items, that's still better than doing a full id2entry on a db with 200,000. 

Another case is broad AND queries. If we do a subtree search of objectClass=group under say ou=Groups, if this returns ALLIDS, we have just wasted a lot of time again falling back to a full db search when we could have loaded the IDL and intersected it with parentid.

I think that we should examine the current idlistscanlimit and either raise it, or consider removing it all together.

Firstyear:
Bug Description: The IDL scan limit existed as a naive attempt at query optimisation to force longer IDLs to be skipped in favour of short IDLs. However, since we now have a true query optimiser that can handle this correctl, the IDL scan limit is redundant and not functional. The only condition the IDL scan limit can now impact is making queries slower by rejecting the use of longer IDLs when shortcut conditions are not possible.

Fix Description: Raise the IDL Scan Limit to INT_MAX to observe the results. In the future we can remove the code entirely.

fixes: https://github.com/389ds/389-ds-base/issues/2435

Review by: ???


progier389 (approved):
LGTM

Issue #5600, opened 2023-01-09
Crash in sync-repl


mreynolds389:
**Issue Description**

 github CI found another random crash in sync repl plugin during the dynamic plugins test:

```
=================================== FAILURES ===================================
________________________ test_sync_repl_dynamic_plugin _________________________

topology = <lib389.topologies.TopologyMain object at 0x7f36f06dc510>
request = <FixtureRequest for <Function test_sync_repl_dynamic_plugin>>

    def test_sync_repl_dynamic_plugin(topology, request):
        """Test sync_repl with dynamic plugin
    
        :id: d4f84913-c18a-459f-8525-110f610ca9e6
        :setup: install a standalone instance
        :steps:
            1. reset instance to standard (no retroCL, no sync_repl, no dynamic plugin)
            2. Enable dynamic plugin
            3. Enable retroCL/content_sync
            4. Establish a sync_repl req
        :expectedresults:
            1. Should succeeds
            2. Should succeeds
            3. Should succeeds
            4. Should succeeds
        """
    
        # Reset the instance in a default config
        # Disable content sync plugin
        topology.standalone.plugins.disable(name=PLUGIN_REPL_SYNC)
    
        # Disable retro changelog
        topology.standalone.plugins.disable(name=PLUGIN_RETRO_CHANGELOG)
    
        # Disable dynamic plugins
        topology.standalone.modify_s(DN_CONFIG, [(ldap.MOD_REPLACE, 'nsslapd-dynamic-plugins', b'off')])
        topology.standalone.restart()
    
        # Now start the test
        # Enable dynamic plugins
        try:
            topology.standalone.modify_s(DN_CONFIG, [(ldap.MOD_REPLACE, 'nsslapd-dynamic-plugins', b'on')])
        except ldap.LDAPError as e:
            log.error('Failed to enable dynamic plugin! {}'.format(e.args[0]['desc']))
            assert False
    
        # Enable retro changelog
        topology.standalone.plugins.enable(name=PLUGIN_RETRO_CHANGELOG)
    
        # Enbale content sync plugin
        topology.standalone.plugins.enable(name=PLUGIN_REPL_SYNC)
    
        # create a sync repl client and wait 5 seconds to be sure it is running
>       sync_repl = Sync_persist(topology.standalone)

dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py:638: 
[...]

PR #5649, opened 2023-02-10
Issue 5600 - buffer overflow when enabling sync repl plugin when dyna…

mreynolds389:
…mic plugins is enabled

Description:

Our factory extension code was not designed to work with dynamic plugins.  It assumes all extensions are regiestered at startup.  If extensions are added after startup (when dynamic plgun is is enabled) then this breaks.  The fix is to keep track in each "object" how many extensions were allocated, instead of relying on the global extension count.

relates: https://github.com/389ds/389-ds-base/issues/5600

### DIFF
--- a/ldap/servers/slapd/factory.c
+++ b/ldap/servers/slapd/factory.c
@@ -24,9 +24,9 @@
  *
  * --- An interface is made available to the core server.
  *
- *   int factory_register_type(const char *name, size_t offset)
- *   void *factory_create_extension(int type,void *object,void *parent)
- *   void factory_destroy_extension(int type,void *object,void *parent,void **extension)
+ *   int factory_register_type(const char *name, size_t offset, size_t count_offset)
+ *   void *factory_create_extension(int type,void *object,void *parent, int32_t *ext_count)
+ *   void factory_destroy_extension(int type,void *object,void *parent, void **extension)
  *
  * An object that wishes to make itself available for extension must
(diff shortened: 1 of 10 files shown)


ldap/servers/slapd/factory.c
```diff
@@ -25,8 +25,8 @@
  * --- An interface is made available to the core server.
  *
  *   int factory_register_type(const char *name, size_t offset)
```
droideck: I think this comment also needs to be changed according to the code change:
`factory_register_type(const char *name, size_t offset, size_t count_offset)`

droideck (commented):
Looks good to me, but I'm not very proficient with this part of code... So @progier389, @tbordaz  should check too, I guess. :)

Another thing, will it be possible to add a test for this case?

mreynolds389:
> Another thing, will it be possible to add a test for this case?

:-)  Yes this existing test case taht casues a crash is the test case:  

dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py::test_sync_repl_dynamic_plugin

progier389 (approved):
I would have rather stored the index in the extension pointer
i.e
 ```
struct factory_extension {
    size_t extension_count_offset;
    struct old_factory_extension *extensions[MAX_EXTENSIONS];
 } *extensions;
```
(to avoid impacting the API prototypes and limiting the changes to the factory code)
but I am fine with your solution either.

PR #5650, opened 2023-02-10
issue 5647 - covscan: memory leak in audit log when adding entries

### LINKED ISSUE #5647: covscan: memory leak in audit log when adding entries
**Issue Description**
Covscan reports:
```
1. Defect type: RESOURCE_LEAK
19. 389-ds-base-2.2.4/ldap/servers/slapd/auditlog.c:231: leaked_storage: Variable "vals" going out of scope leaks the storage it points to.
```
And code review confirms that a charray_free is missing. 
Impact: if nsslapd-auditlog-display-attrs is not *  all added entries attributes values that are logged in audit log are leaking. 

 - Platform: RedHat
 - Package and version: 389-ds-base-2.2.4

**Steps to Reproduce**
1. Generate source package 
2. Have a covscan run using these sources
3. Look at the report about the"Resources leaks" section

**Expected results**
No leaks should be reported.

progier389:
covscan reported an issue about "vals" variable in auditlog.c:231 and indeed a charray_free is missing.

Issue: [5647](https://github.com/389ds/389-ds-base/issues/5647)

Reviewed by: @mreynolds389, @droideck

### DIFF
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -228,6 +228,7 @@ add_entry_attrs(Slapi_Entry *entry, lenstr *l)
                 addlenstr(l, log_val);
                 addlenstr(l, "\n");
             }
+            charray_free(vals);
         }
     } else {
         /* Return all attributes */
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -178,6 +178,40 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
 }
 
 /*
+ * Write the attribute values to the audit log as "comments"
+ *
+ *   Slapi_Attr *entry - the attribute begin logged.
+ *   char *attrname - the attribute name.
+ *   lenstr *l - the audit log buffer
+ *
+ *   Resulting output in the log:
+ *
+ *       #ATTR: VALUE
+ *       #ATTR: VALUE
+ */
+static void
+log_entry_attr(Slapi_Attr *entry_attr, char *attrname, lenstr *l)
+{
+    Slapi_Value **vals = attr_get_present_values(entry_attr);
+    for(size_t i = 0; vals && vals[i]; i++) {
+        char log_val[256] = "";
+        const struct berval *bv = slapi_value_get_berval(vals[i]);
+        if (bv->bv_len >= 256) {
+            strncpy(log_val, bv->bv_val, 252);
+            strcpy(log_val+252, "...");
+        } else {
+            strncpy(log_val, bv->bv_val, bv->bv_len);
+            log_val[bv->bv_len] = 0;
+        }
+        addlenstr(l, "#");
+        addlenstr(l, attrname);
+        addlenstr(l, ": ");
+        addlenstr(l, log_val);
+        addlenstr(l, "\n");
+    }
+}
+
+/*
  * Write "requested" attributes from the entry to the audit log as "comments"
  *
  *   Slapi_Entry *entry - the entry being updated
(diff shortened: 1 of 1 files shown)


progier389:
JIRA ticket is [IDMDS-2386]

ldap/servers/slapd/auditlog.c
```diff
@@ -228,6 +228,7 @@ add_entry_attrs(Slapi_Entry *entry, lenstr *l)
                 addlenstr(l, log_val);
                 addlenstr(l, "\n");
             }
+            charray_free(vals);
```
mreynolds389: This might be overkill, but I really  think we could benefit from looping over the entry attr array by reference, rather than copying/freeing the values.  I did something similar with slapi_entry_attr_get_ref(), and we could create a similar function for the attr array.   

If we can we should always avoid mallcoing/freeing if possible.  It will help with memory usage/fragmentation.

Thoughts?

progier389: Good point and it will also avoid lock contention to access the heap.
   but if we want to avoid any malloc IMHO we have better to iterate directly on the value berval's and then also avoid the strlen call. It should looks like:  
  ```
        slapi_entry_attr_find(e, type, &attr);
        int i = slapi_attr_first_value(attr, &sval);
        while (!r && i != -1) {
            const struct berval *val = slapi_value_get_berval(sval);
```

Comment (mreynolds389) [reply ldap/servers/slapd/auditlog.c]: Sounds great, and perhaps this approach can be used else where in the code (future ticket).  

The GCC team says how DS uses memory is "unusual" and it causes a really large memory footprint.   This is why we use jemalloc today instead of gcc.  Part of this is due to all these small and frequent allocations the server does.  It's been an ongoing silent effort to slowly remove these mallocs/frees where we can.

File: ldap/servers/slapd/auditlog.c [hunk 2]
```diff
@@ -177,6 +177,40 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+ *       #ATTR: VALUE
+ *       #ATTR: VALUE
+ */
+static void
+log_entry_attr(Slapi_Attr *entry_attr, char *attrname, lenstr *l)
+{
+    Slapi_Value **vals = attr_get_present_values(entry_attr);
+    for(size_t i = 0; vals && vals[i]; i++) {
+        char log_val[256] = "";
+        struct berval *bv = slapi_value_get_berval(vals[i]);
```
droideck: Should that be `const struct berval *bv = slapi_value_get_berval(vals[i]);`?

ldap/servers/slapd/auditlog.c
```diff
@@ -177,6 +177,40 @@ write_auditfail_log_entry(Slapi_PBlock *pb)
...
+        if (bv->bv_len > 252) {
```
droideck: In the original code, there was `> 256` which is aligned with log_val size.
You changed it to `> 252`.  Is it a typo or was it intentional?

I think 252 will work too tough (as we cut to 252 anyway). Still, double-checking :)

droideck (approved):
The rest looks good!
But I think Mark will say the final ack :)

ldap/servers/slapd/auditlog.c
progier389: Indeed, both you and the compiler are right ! -;)

ldap/servers/slapd/auditlog.c
progier389: The original code was buggy (buffer overflow if strlen(val) == 256)
Thinking about it I think that the right test is  >= 256

progier389:
Fixed the two remarks from @droideck

mreynolds389 (approved):
LGTM!

PR #5657, opened 2023-02-13
Issue 5628 - Handle graceful timeout in CI tests

### LINKED ISSUE #5628: Handle graceful timeout in CI tests
While working on [PR 5627](https://github.com/389ds/389-ds-base/pull/5627), I noticed that github killed some of the tests after 6 hours. But the only provided information was that it was running pytest 
 
I am thinking about setting a 5 hour alarm and adding a SIG_ALRM handler in TopologyMain.__init__
 The handler will try to stop gracefully any running instances
 then 1 minute later to kill them with SIG_ALRM (to generate a core)
then raise a TimeOut Exception that should loudly abort the testcase.

Notes: 
- the alarm is reset in the request finalizer function. 
- May want to check the PYTEST_CURRENT_TEST to only set the alarm in "tests/suites/*" cases
- May have a way to override the timeout globally with environment variable or on a test basis (adding a method in TopologyMain)

That way: We should have core file for deadlocked instances and as the job should still have enough time to
complete (and collect the assets) 
 

I thought about using pytest-timeout module in signal mode but the drawback is that it will not provides the cores
 of the stuck instances. At least in TopologyMain the instance list is known

progier389:
Issue: Sometime CI test fails because github workflow timeout of 6 hour is triggered and in this case there
is no data to help troubleshooting except that timeout occurred.
Solution: Implement a 5 hour timeout in topologies fixture so that test result artefacts get collected before
 github timeout. when the timeout occurs: the pytest test fails because of a TimeoutError exception and ns-slapd get killed first with SIGTERM then with SIGQUIT ( to get the core file in case of deadlock ) 

Specific Test modules may configure their own timeout by adding an autouse fixture
  (see  dirsrvtests/tests/suites/lib389/timeout_test.py) 

Issue: [5628](https://github.com/389ds/389-ds-base/issues/5628)

Reviewed by:  @mreynolds389

### DIFF
--- a/src/lib389/lib389/topologies.py
+++ b/src/lib389/lib389/topologies.py
@@ -9,6 +9,10 @@
 import os
 import logging
 import socket  # For hostname detection for GSSAPI tests
+import time
+import subprocess
+from signal import SIGALRM, alarm, signal
+from datetime import timedelta
 import pytest
 from lib389 import DirSrv
 from lib389.utils import generate_ds_params, is_fips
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/topologies.py
```diff
@@ -28,6 +32,33 @@
 log = logging.getLogger(__name__)
 
 
+# Github kills workflow after 6 hours so lets keep 1 hour margin
+# For workflow and test initialization, and artefact collection
```
mreynolds389: typo, should be artifacts (not artefacts).  Rest looks good...

mreynolds389 (approved):
LGTM

src/lib389/lib389/topologies.py
progier389: Fixed.

Issue #5658, opened 2023-02-13
CLI - unable to add attribute with matching rule


mreynolds389:
**Issue Description**

When trying to add a matching rule, the command fails.
This works fine when the matching rule option is not used.

```
# dsconf ldapi://%2fvar%2frun%2fslapd-<INSTANCE>.socket schema attributetypes add newAttr --syntax 1.3.6.1.4.1.1466.115.121.1.36 --substr numericstringsubstringsmatch --single-value --user-mod --usage userApplications --desc Test
...
vendorVersion: 389-Directory/1.4.3.31 B2023.012.1725

]
DEBUG: cn=schema get_attr_vals('attributeTypes')
DEBUG: value has to be of str, was 'numericstringsubstringsmatch'
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 138, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.6/site-packages/lib389/cli_conf/schema.py", line 137, in add_attributetype
    schema.add_attributetype(parameters)
  File "/usr/lib/python3.6/site-packages/lib389/schema.py", line 344, in add_attributetype
    return self._add_schema_object(parameters, AttributeType)
  File "/usr/lib/python3.6/site-packages/lib389/schema.py", line 225, in _add_schema_object
    return self.add(attr_name, str(schema_object))
  File "/usr/lib64/python3.6/site-packages/ldap/schema/models.py", line 305, in __str__
    result.append(self.key_attr('SUBSTR',self.substr))
  File "/usr/lib64/python3.6/site-packages/ldap/schema/models.py", line 69, in key_attr
    assert value is None or type(value)==str,TypeError("value has to be of str, was %r" % value)
AssertionError: value has to be of str, was 'numericstringsubstringsmatch'
ERROR: Error: value has to be of str, was 'numericstringsubstringsmatch'
```

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2165091

mreynolds389:
19ad55485..27bf3a348  389-ds-base-2.2 -> 389-ds-base-2.2
a2fbedb55..8af9eb068  389-ds-base-2.1 -> 389-ds-base-2.1
b5220e3a1..74c731a02  389-ds-base-2.0 -> 389-ds-base-2.0
ec64b43f5..9b26ccebc  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5660, opened 2023-02-13
Issue 5658 - CLI - unable to add attribute with matching rule

mreynolds389:
Description:  dsconf incorrectly allows for multiple matching rules of the same type (equality, substr, etc).  This causes python-ldap to get upset and error out.  Change arguments to only allow one matching rule per type.

relates: https://github.com/389ds/389-ds-base/issues/5658


progier389 (approved):
LGTM

PR #5673, opened 2023-02-20
Issue 5267 - CI - Fix issues with nsslapd-return-original-entrydn

mreynolds389:
Description:  

Fix CI test to properly set the nsslapd-return-original-entrydn and to restart the server after changing the config setting.

relates: https://github.com/389ds/389-ds-base/issues/5267

### DIFF
--- a/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
+++ b/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
@@ -73,17 +73,18 @@ def test_dsentrydn(topo):
     assert user.get_attr_val_utf8('dsentrydn') == NEW_USER_DN
 
     # Check DN returned to client matches "dsEntryDN"
-    users = UserAccounts(inst, SUFFIX).list()
+    users = UserAccounts(inst, SUFFIX, rdn="ou=humans").list()
     for user in users:
-        if user.dn.startswith("tUser"):
+        if user.rdn.startswith("tUser"):
             assert user.dn == NEW_USER_DN
             break
 
-    # Disable 'nsslapd-return-original-entrydn' andcheck DN is normalized
-    inst.config.replace('nsslapd-return-original-entrydn', 'on')
-    users = UserAccounts(inst, SUFFIX).list()
+    # Disable 'nsslapd-return-original-entrydn' and check DN is normalized
+    inst.config.replace('nsslapd-return-original-entrydn', 'off')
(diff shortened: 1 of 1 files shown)


vashirov (changes requested):
As I mentioned in the bugzilla, this test needs some adjustments.
1. After `ou=people` is renamed to `ou=humans`, `UserAccounts` needs to have `rdn` parameter pointing to `ou=humans`. Otherwise `users` list is empty and there is nothing to iterate through.
```
(Pdb) users = UserAccounts(inst, SUFFIX).list()
(Pdb) len(users)
0
(Pdb) users = UserAccounts(inst, SUFFIX, rdn="ou=humans").list()
(Pdb) len(users)
1
```
2. In the comparison
```python
if user.dn.startswith("tUser"):
```
`dn` starts with `uid=tUser`, so this comparison fails:
```
(Pdb) user.dn
'uid=tUser,ou=humans,dc=example,dc=com'
(Pdb) user.dn.startswith("tUser")
False
```
We should either compare `user.rdn` or change the argument to use `contains()` method.

mreynolds389:
> As I mentioned in the bugzilla, this test needs some adjustments.
> 
>     1. After `ou=people` is renamed to `ou=humans`, `UserAccounts` needs to have `rdn` parameter pointing to `ou=humans`. Otherwise `users` list is empty and there is nothing to iterate through.
> 
> 
> ```
> (Pdb) users = UserAccounts(inst, SUFFIX).list()
> (Pdb) len(users)
> 0
> (Pdb) users = UserAccounts(inst, SUFFIX, rdn="ou=humans").list()
> (Pdb) len(users)
> 1
> ```
> 
>     2. In the comparison
> 
> 
> ```python
> if user.dn.startswith("tUser"):
> ```
> 
> `dn` starts with `uid=tUser`, so this comparison fails:
> 
> ```
> (Pdb) user.dn
> 'uid=tUser,ou=humans,dc=example,dc=com'
> (Pdb) user.dn.startswith("tUser")
> False
> ```
> 
> We should either compare `user.rdn` or change the argument to use `contains()` method.

Fixed

vashirov:
And the issue with `UserAccounts`? :)

mreynolds389:
> And the issue with `UserAccounts`? :)

Err trying to do too much on a Monday morning :-)  Fixed

vashirov (approved):
LGTM, thank you!

PR #5675, opened 2023-02-20
Issue 5600 - buffer overflow when enabling sync repl plugin when dynamic plugins is enabled

mreynolds389:
Description:

Our factory extension code was not designed to work with dynamic plugins.  It assumes all extensions are registered at startup.  If extensions are added after startup (when dynamic plugins is enabled) then this breaks.  The fix is to keep track of how many extensions were allocated per object, instead of relying on the global extension count.

Patch written by Pierre Rogier - thanks!

relates: https://github.com/389ds/389-ds-base/issues/5600


jchapma (approved):
Ran the CI test and reviewed your changes. LGTM

PR #5676, opened 2023-02-20
Issue 5661 - LMDB hangs while Rebuilding the replication changelog RUV

### LINKED ISSUE #5661: LMDB hangs while Rebuilding the replication changelog RUV
**Issue Description**
While running github CI test on [PR 5657](https://github.com/389ds/389-ds-base/pull/5657) : 
`LMDB test_memberof_with_changelog_reset` fails because of the 5 hours timeout.
The hang occurs while starting the server:
```
   def test_memberof_with_changelog_reset(topo_m2):
    ...
>       m1.start()
...
>       raise TimeoutError(f"Test timed out after {str(timedelta(seconds=_test_timeout))}")
E       TimeoutError: Test timed out after 5:00:00
```
and error logs is:
```
389-Directory/2.3.2.202302101853gitb777520 B2023.041.1854
server.example.com:39001 (/etc/dirsrv/slapd-supplier1)
10/Feb/2023:19:16:36.234799574 +0000] - INFO - main - 389-Directory/2.3.2.202302101853gitb777520 B2023.041.1854 starting up
...
[10/Feb/2023:19:17:41.358041567 +0000] - NOTICE - NSMMReplicationPlugin - changelog program - _cl5ConstructRUV - Rebuilding the replication changelog RUV, this may take several minutes...
```
So apparently the server was stuck 5 hours while rebuilding the changelog RUV.

No core was generated so SIGTERM should have been able to stop the server

 - Platform:  Fedora
 - Package and version: main branch.

**Steps to Reproduce**
1. Looks to github actions and check for LMDB Test(memberof_plugin) results

**Expected results**
No timeout errors

progier389:
**Problem**:
Tests that:
1. rebuilding the changelog RUV
2. perform changelog trimming (i.e any test that are long enough)
3. purge a Replica ID in the changelog (i.e demote a supplier/ cleanruv)
Are hanging on lmdb until killed by the timeout because the changelog iterator loops on first entry.

**Reason**:
It is due to a difference in the way bdb and lmdb cusor get works when it is the first cursor operation and
no key is provided. (bdb returns first key while lmdb fails). Because of that db-mdb layers use MDB_FIRST when no key is provided.

**Solution**: 
  Add a new dbimp function that iterates over a cursor calling a callback with key and value for all the records
   (until either the end of the database or the callback says to stop) 
   This is more efficient than current code that walk the cursor through dbimpl API at the replication plugin level 
    because it avoids the dbimpl API  overhead (especially in lmdb case) 
     furthermore it allows to easily handle the dbs specificity.
- Fixed the changelog RUVs rebuild to walk the changelog (using the new dbimpl function) only once instead of twice (collecting min and max for all replica ids then building both standard and purge ruv from these data
- Fixed the changelog trimming code to use the new changelog iterator, also fixed the way the purge ruv get updated (using now the first csn in the changelog (for the replicas whose csn get removed)  instead of the last removed csn )
- Fixed the changelog rid purge code to use the new changelog iterator
- Note: The changelog replication replay iterator is not impacted by this bug and is not changed.
   
- Also fixed a regression related to the addition of timeout in the test topology 
 ( m1h1c1 topology is slightly different from the older version and is having a hub->supplier agreement which is
   making some test fail. Fixed that topology to remove that agreement.)

Issue: [5661](https://github.com/389ds/389-ds-base/issues/5661
 
Reviewed by: @tbordaz ( Thanks! )

### DIFF
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -1841,11 +1841,23 @@ _cl5Iterate(cldb_Handle *cldb, dbi_iterate_cb_t *action_cb, DBLCI_CTX *dblcictx,
             dblcictx->startcsn = *startcsn;
             startcsn = &dblcictx->startcsn;
         }
-        /* DB txn lock accessed pages until the end of the transaction. */
-        rc = TXN_BEGIN(cldb, NULL, &txnid, readonly);
-        if (rc != 0) {
-            rc = _cl5Dberror(cldb, rc, "_cl5Iterate - Failed to begin transaction");
-            continue;
+        if (!readonly || dblayer_is_lmdb(cldb->be)) {
+            /*
+             * if write operation are performed or if lmdb is used, a transaction is needed.
+             * In bdb case DB txn lock accessed pages until the end of the transaction.
+             * in lmdb case the read or write lock is global and also held until the end
+             * of the transaction.
+             */
+            rc = TXN_BEGIN(cldb, NULL, &txnid, readonly);
+            if (rc != 0) {
+                rc = _cl5Dberror(cldb, rc, "_cl5Iterate - Failed to begin transaction");
+                continue;
+            }
+        } else {
+            /* read-only opertion on bdb are transactionless, so no reason to abort txn 
+             * after having seen some number of records 
+             */
+            dblcictx->seen.nbmax = 0;
         }
         /* create cursor */
         rc = dblayer_new_cursor(cldb->be, cldb->db, txnid, &dblcictx->cursor);
@@ -1887,17 +1899,18 @@ _cl5Iterate(cldb_Handle *cldb, dbi_iterate_cb_t *action_cb, DBLCI_CTX *dblcictx,
         dblcictx->changed.tot += dblcictx->changed.nb;
         dblcictx->seen.tot += dblcictx->seen.nb;
 
-        if (dblcictx->changed.nb) {
-            int rc2 = TXN_COMMIT(cldb, txnid);
-            if (rc2 != DBI_RC_SUCCESS) {
-                rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to commit transaction");
-            }
-        } else {
-            int rc2 = TXN_ABORT(cldb, txnid);
-            if (rc2 != DBI_RC_SUCCESS) {
-                rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to abort transaction");
-                if (rc == CL5_SUCCESS || rc == CL5_NOTFOUND) {
-                    rc = rc2;
+        if (txnid) {
+            if (dblcictx->changed.nb) {
+                int rc2 = TXN_COMMIT(cldb, txnid);
+                if (rc2 != DBI_RC_SUCCESS) {
+                    rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to commit transaction");
+                }
+            } else {
+                int rc2 = TXN_ABORT(cldb, txnid);
+                if (rc2 != DBI_RC_SUCCESS) {
+                    if (rc == CL5_SUCCESS || rc == CL5_NOTFOUND) {
+                        rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to abort transaction");
+                    }
                 }
             }
         }
--- a/ldap/servers/plugins/replication/cl5_api.c
+++ b/ldap/servers/plugins/replication/cl5_api.c
@@ -26,6 +26,7 @@
 #if defined(linux)
 #include <sys/vfs.h>
 #endif
+#include "slap.h"
 
 
 #include "cl5.h"
@@ -86,6 +87,12 @@
 #define NO_DISK_SPACE 1024
 #define MIN_DISK_SPACE 10485760 /* 10 MB */
 
+#define MAX_RETRIES 10 /* Maximum number of retry in case of db retryable error */
+#define CL5_TRIM_MAX_PER_TRANSACTION 100
+#define CL5_TRIM_MAX_LOOKUP_PER_TRANSACTION 10000
+#define CL5_PURGE_MAX_PER_TRANSACTION 1000
+#define CL5_PURGE_MAX_LOOKUP_PER_TRANSACTION 10000
+
 /***** Data Definitions *****/
 
 /* possible changelog open modes */
@@ -112,7 +119,7 @@ struct cl5DBFileHandle
 /* info about the changelog file in the main database environment */
 /* usage as CL5DBFile, but for new implementation use a new struct
  * can be replaced later
- */ 
+ */
 {
     dbi_db_t *db;           /* db handle to the changelog file */
     dbi_env_t *dbEnv;       /* db environment shared by all db files */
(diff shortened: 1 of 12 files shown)


ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -86,6 +87,12 @@
 #define NO_DISK_SPACE 1024
 #define MIN_DISK_SPACE 10485760 /* 10 MB */
 
+#define MAX_RETRIES 10 /* Maximum number of retry in case of db retryable error */
+#define CL5_TRIM_MAX_PER_TRANSACTION 100
+#define CL5_TRIM_MAX_LOOKUP_PER_TRANSACTION 10000
```
tbordaz, line 92: This value seems high compare to the default db-lock. I was thinking that db can consume up to one lock per lookup page. Meaning that a single TRIM/PURGE loop can consume all the db-lock. Am I missing something ?

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -1705,6 +1739,213 @@ cl5DBData2Entry(const char *data, PRUint32 len __attribute__((unused)), CL5Entry
...
+        }
+        if (dblcictx->seen.nbmax > 10 && rc == CL5_DB_LOCK_ERROR) {
+            /*
+             * Ran out of locks, need to restart the transaction.
+             * Reduce the the batch count and reset the key to
+             * the starting point.
+             * PR: Not sure weither this code is still working because dbimpl API layer
+             *  may remap bdb error 12 to something else.
+             */
+            dblcictx->seen.nbmax -= 10;
```
tbordaz, line 1886: Were you able to verify that this handling of DB_LOCK_ERROR works ?
If it does that is nice as I was thinking than when the server was getting out of lock, it was almost fatal.
If it works, shouldn't we reduce #seen more significantly ? -10 is just 0.1% step compare to MAX_LOOKUP_PER_TRANSACTION ?

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -4294,63 +4172,66 @@ _cl5CheckMissingCSN(const CSN *csn, const RUV *supplierRuv, cldb_Handle *cldb)
...
-    } else {
-        rc = CL5_SUCCESS;
+    if (slapi_is_loglevel_set(SLAPI_LOG_REPL)) {
+        ruv_dump(cldb->purgeRUV, "clpurgeruv", prFile);
+        ruv_dump(cldb->maxRUV, "clmaxruv", prFile);
     }
+    slapi_write_buffer(prFile, "\n", strlen("\n"));
 
-    return rc;
+    dblcictx.seen.nbmax = CL5_PURGE_MAX_LOOKUP_PER_TRANSACTION;
```
tbordaz, line 4250: This is an export, is it done under a TXN ?

tbordaz (commented):
I like the use of _cl5Iterate as a common iterator. Overall patch looks good but it is difficult to do a precise review with such big change.
Few comments/questions.

ldap/servers/plugins/replication/cl5_api.c
progier389: 1. That was the value in the old code    `2494             if (trimmed == 10000 || (batch_count && trimmed == batch_count)) {`
2. db_stat -h $DB -Cp shows that lock table is real size is nearly 16K (while 10K are configured) so 10K pages should be fine.)

ldap/servers/plugins/replication/cl5_api.c
progier389: Once again 10 was the value in the old code
And IMHO reducing by 10 is fine, because we are not decreasing MAX_LOOKUP_PER_TRANSACTION 
 but the number of operation we were already able to processed (so -1 should be ok).
The only case it should fail a second time is if the other threads are using more txn than it was the case in the first time.
And no, I have not tested, IMHO this feature is broken because bdb_map_error does not handle 12
Should add a new DBI_RC return code to handle cleanly that case.
 and use it instead of the CL5_DB_LOCK_ERROR that we should remove. 
 I am creating a new github issue for that one.

ldap/servers/plugins/replication/cl5_api.c
progier389: Interesting point:
   Apparently the old code used read-only txn on lmdb (lmdb API requires a txn to create a cursor) 
   and none with bdb, but the new code is  always using a read-only txn (cf _cl5Iterate)
I have mixed feeling about that:
  I could always change _cl5Iterate to avoid creating a read only txn in bdb case but I wonder if the old code
  could not lead to weird inconsistency because of the lack of txn ...

ldap/servers/plugins/replication/cl5_api.c
tbordaz: Okay it makes code cleaner and preserve the original behavior. That is fine.

ldap/servers/plugins/replication/cl5_api.c
progier389: Anyway I think that this code is probably obsolete because  the lock monitoring thread should now stop the server before getting out of locks

tbordaz: Thanks for the explanations

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -1705,6 +1739,227 @@ cl5DBData2Entry(const char *data, PRUint32 len __attribute__((unused)), CL5Entry
...
+                int rc2 = TXN_COMMIT(cldb, txnid);
+                if (rc2 != DBI_RC_SUCCESS) {
+                    rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to commit transaction");
+                }
+            } else {
+                int rc2 = TXN_ABORT(cldb, txnid);
+                if (rc2 != DBI_RC_SUCCESS) {
+                    rc = _cl5Dberror(cldb, rc2, "_cl5Iterate - Failed to abort transaction");
+                    if (rc == CL5_SUCCESS || rc == CL5_NOTFOUND) {
+                        rc = rc2;
```
tbordaz: I wonder if this line is correct (I know it is cut/paste old code). Here _cl5Dberror translate a DBI error (rc2) into a CL5 error (rc), but then this translation is lost when rc (CL5) is overwriten by rc2 (DBI).

tbordaz (commented):
Except a single question the patch looks good to me. Just waiting for your feedback then I will ack.

ldap/servers/plugins/replication/cl5_api.c
progier389: You are right: I wanted to avoid overwriting rc if there was a previous error but I mixed up things!
the 1911 line should be moved to replace 1913 one ...

progier389: Now fixed.

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -1705,6 +1739,226 @@ cl5DBData2Entry(const char *data, PRUint32 len __attribute__((unused)), CL5Entry
...
+        if (startcsn) {
+            char csnstr[CSN_STRSIZE] = "";
+            dbi_val_t startingkey = {0};
+            csn_as_string(startcsn, 0, csnstr);
+            dblayer_value_set(cldb->be, &startingkey, csnstr, CSN_STRSIZE);
+            rc = dblayer_cursor_iterate(&dblcictx->cursor, action_cb, &startingkey, dblcictx);
+        } else {
+            rc = dblayer_cursor_iterate(&dblcictx->cursor, action_cb, NULL, dblcictx);
+        }
+        if (dblcictx->seen.nbmax > 10 && rc == CL5_DB_LOCK_ERROR) {
```
tbordaz, line 1878: Sorry to be precise but I think rc is in DBI_ family it can not be compared to CL5_. My understanding is that DB_Lock error is quite BDB related and we can not test it. An option would be to test (rc != DBI_RC_SUCCESS)

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -2291,142 +2545,81 @@ _cl5DoPurging(cleanruv_purge_data *purge_data)
...
+    /* Verify that we have a valid csn */
+    if (key->size != CSN_STRSIZE) {
+        return DBI_RC_SUCCESS;
+    }
+    /* Update last csn */
+    csn_init_by_string(&dblcictx->csn, data->data);
+    if (_cl5CIEventCheckTxnEnd(&dblcictx->seen) ||
+        _cl5CIEventCheckTxnEnd(&dblcictx->changed)) {
+        /* Stop this txn */
+        return DBI_RC_NOTFOUND;
```
tbordaz, line 2578: As the txn overpass a limit (like CL5_PURGE_MAX_LOOKUP_PER_TRANSACTION or CL5_PURGE_MAX_PER_TRANSACTION) shouldn't we return DBI_RC_RETRY ?

tbordaz (commented):
Fix looks good. Two others questions.

ldap/servers/plugins/replication/cl5_api.c
progier389: FYI: I agree with your first statement. That is why some time ago I open  [Issue 5679](https://github.com/389ds/389-ds-base/issues/5679) 
Anyway IMHO this code is now obsolete because the db lock monitoring thread should stop the server before bdb returns this error.

progier389: But I disagree with your suggestion: We must not retry after decreasing the max number of seen records  
 for any database error

ldap/servers/plugins/replication/cl5_api.c
progier389: No: because in retry case, the txn is aborted and iteration is restarted from the starting point.
but with DBI_RC_NOTFOUND the dblayer_iterate function  stops with a SUCCESS rc 
then the txn is commited and the cl5_iterate function restarts a new iteration on last seen record

tbordaz: Excellent. Thanks for these explanations. Do you mind to add this comment (DBI_RC_NOTFOUND->dblayer_iterate succeeds -> txn comming..). It could help in the future.

tbordaz (approved):
Thanks for all your explanations. 
LGTM

ldap/servers/plugins/replication/cl5_api.c
progier389: OK. I added a comment

Issue #5338, opened 2022-06-08
Work queue contention


jchapma:
**Description**
Contention has been identified on the connection work queue, this issue aims to resolve that.

**Additional context**
See the design discussion on this topic here: https://github.com/mreynolds389/389wiki/pull/77

jchapma:
This is a draft implementation of a lockless work queue to address the contention observed when the server is processing connections. However, the design of the lockless queue means it is unsafe to free a node after it is dequeued as another thread could be using it. I have noticed this behaviour intermittently, where a thread tries to access a node that has been freed, breaking the linked list. 

Possible solutions to this problem: 
Use a double word compare and swap, but this is not available on commonly used hardware.
Use node pointer modification counters to only free a node when no thread is using it.

I am currently working on this, hence the WIP label.

jchapma:
Rebase

tbordaz:
When #4812 is fixed and this one also, expected gain are big (>+50% throughput). Our tests will give tuning recommendations that we can implement (another ticket) and describe in the doc.

Firstyear:
Please don't roll your own lockless queue, there are so many subtleties to it that can break things. It would be better to find an existing lockfree queue library to use that has extensive testing and analysis performed, or to roll up the crossbeam rust lock free queue with a C ffi api. 

PS: DWCAS is common, it's on every modern cpu except a single generation of extremely early AMD 64 bit cpus.

PR #5681, opened 2023-02-23
Issue 4812 - Listener thread does not scale with a high num of established connections

jchapma:
Description: Adding num listeners config option, some tidy up

relates: https://github.com/389ds/389-ds-base/issues/4812

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -129,20 +129,21 @@ connection_table_new(int table_size)
     int free_idx = 0;
     ber_len_t maxbersize = config_get_maxbersize();
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
-    ct->size = table_size;
-    ct->list_num = SLAPD_DEFAULT_NUM_LISTENERS;
-    ct->list_size = (table_size+ct->list_num-1)/ct->list_num; /* to avoid rounding issue */
     ct->num_active = (int *)slapi_ch_calloc(1, ct->list_num * sizeof(int));
-    ct->c = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection *));
+    ct->list_num = config_get_num_listeners();
+    ct->size = table_size - (table_size % ct->list_num);
+    ct->list_size = ct->size/ct->list_num;
+    ct->num_active = (int *)slapi_ch_calloc(1, ct->list_num * sizeof(int));
+    ct->c = (Connection **)slapi_ch_calloc(1, ct->size * sizeof(Connection *));
     ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, ct->list_num * sizeof(struct POLL_STRUCT*));
     ct->table_mutex = PR_NewLock();
     /* Allocate the freelist */
-    ct->c_freelist = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection *));
+    ct->c_freelist = (Connection **)slapi_ch_calloc(1, ct->size * sizeof(Connection *));
     /* NEVER use slot 0, this is a list pointer */
     ct->conn_next_offset = 1;
     ct->conn_free_offset = 1;
 
-    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "conntablesize:%d\n", ct->size);
+    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "list_num:%d - list_size:%d\n",  ct->list_num, ct->list_size);
 
     pthread_mutexattr_t monitor_attr = {0};
     pthread_mutexattr_init(&monitor_attr);
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -143,7 +143,8 @@ connection_table_new(int table_size)
     ct->conn_next_offset = 1;
     ct->conn_free_offset = 1;
 
-    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "list_num:%d - list_size:%d\n",  ct->list_num, ct->list_size);
+    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "Number of connection sub-tables %d, each containing %d slots.\n",
+        ct->list_num, ct->list_size);
 
     pthread_mutexattr_t monitor_attr = {0};
     pthread_mutexattr_init(&monitor_attr);
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -2429,13 +2429,11 @@ netaddr2string(const PRNetAddr *addr, char *addrbuf, size_t addrbuflen)
 static int
 createsignalpipe(void)
 {
-    int i;
-
     /* Now we know how many listeners there are, setup signalpipes */
     signalpipes = (signal_pipe*) slapi_ch_calloc(1, (sizeof(signal_pipe) * the_connection_table->list_num));
 
     /* there is a signal pipe for each ct list/thread mapping */
-    for (i = 0; i < the_connection_table->list_num; i++) {
+    for (size_t i = 0; i < the_connection_table->list_num; i++) {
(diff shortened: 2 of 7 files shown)


ldap/servers/slapd/daemon.c
```diff
@@ -2421,11 +2423,15 @@ netaddr2string(const PRNetAddr *addr, char *addrbuf, size_t addrbuflen)
...
 
-
+/* Create a signal pipe for each listener thread */
 static int
 createsignalpipe(void)
 {
     int i;
+
+    /* Now we know how many listeners there are, setup signalpipes */
+    signalpipes = (signal_pipe*) malloc(sizeof(signal_pipe) * the_connection_table->list_num);
```
tbordaz: should use slapi_ch_calloc rather than malloc

ldap/servers/slapd/daemon.c
```diff
@@ -2452,6 +2458,19 @@ createsignalpipe(void)
...
+static int
+destroysignalpipe(void)
+{
+    int i;
+
+    for (i = 0; i < the_connection_table->list_num; i++) {
+        (void) PR_Close(signalpipes[i].signalpipe[0]);
+        (void) PR_Close(signalpipes[i].signalpipe[1]);
+    }
+    free(signalpipes);
```
tbordaz: slapi_ch_free rather than free

ldap/servers/slapd/libglobs.c
```diff
@@ -4997,6 +5002,38 @@ config_set_reservedescriptors(const char *attrname, char *value, char *errorbuf,
...
+
+    if (config_value_is_null(attrname, value, errorbuf, 0)) {
+        return LDAP_OPERATIONS_ERROR;
+    }
+
+    errno = 0;
+    nValue = strtol(value, &endp, 0);
+    if (*endp != '\0' || errno == ERANGE || nValue < minVal || nValue > maxVal) {
+        nValue = (nValue < minVal) ? minVal : maxVal;
+        slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE,
```
tbordaz, line 5029: message should rather mention liseners than conntable split

tbordaz (commented):
The patch looks good but I do not understand the mix of connection table size with number of listeners.
I was expecting the two patches to be separated. Increase default conntable size in one patch and tuning of the listeners number in another patch

ldap/servers/slapd/libglobs.c
progier389: IMHO you should rather reference explicitly the attribute name
 (i.e "%s: invalid value \"%s\", " CONFIG_NUM_LISTENERS_ATTRIBUTE 
        " must range from %d to %d. ""%s ",  ...

ldap/servers/slapd/daemon.c
Firstyear: Yep, fully agreed with tbordaz here. Always use slapi_ch_calloc here, becuase it's actually *faster* than trying to zero something ourselve, and having already zerod blocks of memory is a significant improvement for asan detection but also security.

ldap/servers/slapd/daemon.c
```diff
@@ -2452,6 +2458,19 @@ createsignalpipe(void)
     return (0);
 }
 
+static int
+destroysignalpipe(void)
+{
+    int i;
```
Firstyear: size_t i, and also inline with "for ( size_t i = 0; ...". This allows compiler optimisation over the iterator because the variable is constrained.

Firstyear (changes requested):

jchapma:
> The patch looks good but I do not understand the mix of connection table size with number of listeners. I was expecting the two patches to be separated. Increase default conntable size in one patch and tuning of the listeners number in another patch

Users were having issues with the default connection table size, so a while ago we made some changes to allow the server calculate the connection table size dynamically. (d4682046d570c26a19285b4946d1af80057f26e4) 

The references you see to conntable size in this patch is for the removal of legacy code that wasnt used. It should have been removed with the commit mentioned above but was not.

tbordaz:
@jchapma do you mind to rebase on main branch, I fail to apply it.

ldap/servers/slapd/conntable.c
```diff
@@ -129,20 +129,21 @@ connection_table_new(int table_size)
...
     ct->table_mutex = PR_NewLock();
     /* Allocate the freelist */
-    ct->c_freelist = (Connection **)slapi_ch_calloc(1, table_size * sizeof(Connection *));
+    ct->c_freelist = (Connection **)slapi_ch_calloc(1, ct->size * sizeof(Connection *));
     /* NEVER use slot 0, this is a list pointer */
     ct->conn_next_offset = 1;
     ct->conn_free_offset = 1;
 
-    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "conntablesize:%d\n", ct->size);
+    slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "list_num:%d - list_size:%d\n",  ct->list_num, ct->list_size);
```
tbordaz: This message is logged systematically. It looks a bit  too internals for the admin. The admin set up N listeners, you may log that connection table is split in N sub-tables each containing M slots.

tbordaz (approved):
The patch looks good to me. Just  a minor change regarding logged message.

tbordaz (approved):

jchapma:
Enabled modification of nsslapd-numlisteners via the webUI. Maybe @droideck @mreynolds389 could take a quick look, thanks.

ldap/servers/slapd/daemon.c
```diff
@@ -2423,11 +2425,15 @@ netaddr2string(const PRNetAddr *addr, char *addrbuf, size_t addrbuflen)
...
 static int
 createsignalpipe(void)
 {
     int i;
+
+    /* Now we know how many listeners there are, setup signalpipes */
+    signalpipes = (signal_pipe*) slapi_ch_calloc(1, (sizeof(signal_pipe) * the_connection_table->list_num));
+
     /* there is a signal pipe for each ct list/thread mapping */
     for (i = 0; i < the_connection_table->list_num; i++) {
```
mreynolds389: Use "size_t i = 0" here

mreynolds389:
I am not sure we should add this to the UI at this time.  Since it's sort of still experimental at this time.  @tbordaz @Firstyear what do you guys think?

droideck (commented):
It looks a bit odd on the bottom of PR:
image

Probably, the rebase is still required.

But besides that, even though I don't see any issues with the connection code, I'll leave other folks to check it as I still need more experience with these parts. :)

tbordaz:
> I am not sure we should add this to the UI at this time. Since it's sort of still experimental. @tbordaz @Firstyear what do you guys think?

I agree with your point, no need to add it now in UI.
This ticket fixes a performance bottleneck but it hits the next one (#5338).
It significantly reduces the CPU intensive load of a single listener polling all established connections and spread the load over several listeners. In our tests, it does not significantly increase throughput/reduce response time because of #5338. However we can imagine that in others scenario, with high number of conn, it can help.
ATM I think we should document this tuning as a possible relief when listener eats all CPU. But not expose it as a magic tuning button in the UI.

ldap/servers/slapd/libglobs.c
```diff
@@ -6811,6 +6848,19 @@ config_get_referral_mode(void)
...
 }
 
+int
+config_get_num_listeners(void)
+{
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+    int retVal;
+
+    CFG_LOCK_READ(slapdFrontendConfig);
+    retVal = slapdFrontendConfig->num_listeners;
```
Firstyear, line 6858: This needs to check for num_listeners > 0

Firstyear (changes requested):

ldap/servers/slapd/libglobs.c
progier389: @Firstyear, as the default value is > 0 and the setter checks that the value is between 1 or 4, 
I do not think we really need to add a check here, especially when none of the other getter functions have such tests

Firstyear (approved):

Issue #5683, opened 2023-02-23
Update specfile to skip checks for ASAN builds


mreynolds389:
**Issue Description**

Every time we do an ASAN build we need to modify the specfille to comment out a rpm check which is very annoying.

mreynolds389:
5b70454a6..852857b55  389-ds-base-2.2 -> 389-ds-base-2.2
174099d4f..1985ce673  389-ds-base-2.1 -> 389-ds-base-2.1
c819c480d..25912d7df  389-ds-base-2.0 -> 389-ds-base-2.0
cdf5ca598..6e1b7f05d  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5684, opened 2023-02-23
Issue 4583 - Update specfile to skip checks of ASAN builds

mreynolds389:
Description:  Need to skip check for ASAN builds otherwise builds never complete

relates: https://github.com/389ds/389-ds-base/issues/5683


droideck (approved):
LGTM!

Issue #5554, opened 2022-12-09
Add more tests to security_basic_test suite


droideck:
**Issue Description**
We need to cover ANONYMOUS_BIND and TLSCLIENTAUTH cases (including CERT_MAP_FAILED)

The main feature issue:
https://github.com/389ds/389-ds-base/issues/5335

PR #5688, opened 2023-03-01
Issue 5554 - Add more tests to security_basic_test suite

bsimonova:
Description:
Add test for TCP_ERROR maxbersize B2 scenario.

Relates: https://github.com/389ds/389-ds-base/issues/5554

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/logging/security_basic_test.py
+++ b/dirsrvtests/tests/suites/logging/security_basic_test.py
@@ -92,6 +92,15 @@ def check_log(inst, event_id, msg, dn=None, bind_method=None):
     assert False
 
 
+@pytest.fixture(scope="module")
+def big_file():
+    TEMP_BIG_FILE = ''
+    for x in range(1048576):
+        TEMP_BIG_FILE += '+'
+
+    return TEMP_BIG_FILE
+
+
 def test_invalid_binds(topo, setup_test):
     """Test the various bind scenarios that should be logged in the security log
 
@@ -168,7 +177,7 @@ def test_authorization(topo, setup_test):
     """
     inst = topo.standalone
 
-    # Delete the previous securty logs
+    # Delete the previous security logs
     inst.deleteSecurityLogs()
 
     # Bind as a user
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
LGTM

dirsrvtests/tests/suites/logging/security_basic_test.py
```diff
@@ -258,6 +267,48 @@ def test_tcp_events(topo, setup_test):
...
 
+def test_tcp_events_maxbersize(topo, setup_test, big_file):
+    """Trigger a TCP_ERROR event B2 that should be logged in the security log
+
+    :id: 85e5ac23-4288-4e55-b8c3-1f8f39e95c2b
+    :setup: Standalone Instance
+    :testype: Non-functional
+    :subtype1: Security
+    :subtype2: Penetration
+    :subsystemteam: sst_idm_ds
```
droideck, line 278: I think we need to update https://github.com/389ds/389-ds-base/blob/main/dirsrvtests/testimony.yaml as the validate test fails.

droideck (approved):
Except for the minor issue, looks good!

dirsrvtests/tests/suites/logging/security_basic_test.py
bsimonova: Ah, I see, didn't realize the validation.
There is a requirement to mark tests that fall into Security category in Polarion.
However they are still making changes to it.

I updated the testimony file, @vashirov do you think it's ok like this?
Or should I revert and wait for the final conclusion to the security fields?

Thanks!

vashirov: Yeah, it's fine. `testimony.yml` defines which fields can be present and if they are required or not. So if these fields will be unused, it won't fail the check. And once there will be an agreement on which fields are required, we can update this file.

bsimonova: Thanks!

PR #5691, opened 2023-03-07
Issue 5598 - (2nd) In 2.x, SRCH throughput drops by 10% because of ha…

tbordaz:
…ndling of referral

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_search.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_search.c
@@ -1044,7 +1044,7 @@ build_candidate_list(Slapi_PBlock *pb, backend *be, struct backentry *e, const c
              *  - there is no referral on the server
              *  - this is an internal SRCH
              */
-            filter_exec = slapi_filter_dup(filter);
+            filter_exec = filter;
         } else {
             /* make (|(originalfilter)(objectclass=referral)) */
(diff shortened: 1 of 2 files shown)


progier389:
You should remove the Cargo.lock from that commit and also update the PR description field. 
Except that, LGTM even if I still do not understand why the leak was only seen with your changes.

tbordaz:
> You should remove the Cargo.lock from that commit and also update the PR description field. Except that, LGTM even if I still do not understand why the leak was only seen with your changes.

Thanks for the review, I did the changes. I am also surprised, I have the feeling that the first fix revealed an underlying leak.

progier389 (approved):
LGTM

mreynolds389 (approved):
LGTM - basic and filter tests pass

tbordaz:
Changed the patch because I misunderstood the contain of SLAPI_SEARCH_FILTER_INTENDED and SLAPI_SEARCH_FILTER. The filter optimizer makes SLAPI_SEARCH_FILTER_INTENDED the original filter and craft a new filter in SLAPI_SEARCH_FILTER. But the new filter includes the original one (it justs add '(objectclass=referral)'.
I made a new patch that is fixing the leak (without double free or use-after-free) but need to wait of the full CI run.

mreynolds389 (approved):
Much better!  Ack

PR #5692, opened 2023-03-09
Issue 5598 - (3rd) In 2.x, SRCH throughput drops by 10% because of ha…

tbordaz:
…ndling of referral

Bug description:
        The first fix 5598 logs a single and useless message (INFO)
	when configuring a backend/mapping tree.
	"INFO - slapd_daemon - New referral entries are detected under
        dc=example,dc=com (returned to SRCH req)"
	The reason is that it checks referral (internal search)
	at the backend level. This is called at startup and config.
	Upon config it should not be called because backend/
	mapping tree are not ready for internal search

Fix description:
	Moving the test from ldbm_instance_start(backend) to
	startup ldbm_instance_startall (after slapi_mtn_be_started)

relates:  #5598

Reviewed by: Mark Reynolds

### DIFF
--- a/ldap/servers/slapd/back-ldbm/instance.c
+++ b/ldap/servers/slapd/back-ldbm/instance.c
@@ -259,11 +259,6 @@ ldbm_instance_start(backend *be)
     }
 
     rc = dblayer_instance_start(be, DBLAYER_NORMAL_MODE);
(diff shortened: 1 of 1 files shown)


mreynolds389 (approved):
ACK

Issue #4598, opened 2021-02-05
Performance modify rate: contention on syntax (oid/name) RWlocks


tbordaz:
**Issue Description**
/opt/ldapbench/bin/modrate --hostname instance --port 389 --bindDN 'cn=Directory Manager' --bindPassword xxx --entryDN id=user.[1-10000],ou=People,dc=example,dc=com" --attribute street --attribute description --valueLength 12 --numThreads 5

updates two non unindexed attributes. No need to put more threads as updates are serialized it only increases response time.

By default the syntax names/oids tables are protected by a lock (each one). These locks are the one on higher contention (21%). Attribute syntax are lookup for normalization/comparison.
<pre>
#        ----------- Cacheline ----------    Total      Tot  ----- LLC Load Hitm -----  ---- Store Reference ----
# Index             Address  Node  PA cnt  records     Hitm    Total      Lcl      Rmt    Total    L1Hit   L1Miss
# .....  ..................  ....  ......  .......  .......  .......  .......  .......  .......  .......  .......
#
      0      0x7fd799028500     0     290     3204   21.07%       59       59        0     1435     1435        0
      1      0x7fd79902a980     0      95      139    3.57%       10       10        0       41       41        0
      2      0x7fd799028340     0      52      105    2.50%        7        7        0       38       38        0
      3      0x7fd7868e1000     0     140      272    2.14%        6        6        0      168      162        6
...

</pre>

When turning off the protection 'nsslapd-schemamod: off' it gives a 5% gain of throughput (similar resp time): 3555/s vs 3370/s.
This issue is to evaluate possible improvements.

**Package Version and Platform:**
All plateforms/versions

**Steps to Reproduce**
provision a 10K entries and apply updates MOD_REPLACE on non indexed attributes

**Expected results**
Reduce the contention.

mreynolds389:
Something I had a patch for and lost :-(, was having each worker thread have its own copy of these tables.  The only issue was how to deal with refreshing those per thread copies without it being too costly.  Anyway, I just wanted to present this idea for this issue.

Firstyear:
@mreynolds389 If I make a C ffi wrapper to https://docs.rs/concread/0.2.7/concread/cowcell/struct.CowCell.html this would be perfect. It allows rwlock semantics but with no blocking on writes.

tbordaz:
[...]

PR #5693, opened 2023-03-09
Issue 4598 - Copy attribute syntax hashtables to each worker thread

mreynolds389:
Description:

There is a lot of contention around the rwlocks for the attribute syntax hashtables.  Since syntaxes rarely change we can just keep a copy in each thread and avoid locking.

Then the worker can check for changes on each loop and rebuild the hashtables as needed.

Did some code cleanup in schema.c

relates: https://github.com/389ds/389-ds-base/issues/4598


progier389:
Still reviewing but I have a  general comment:
  - I would have put the code that check version change and duplicate the hash table (in connection.c) in a separate function (and in libslapd rather than in ns-slapd)
  And same thing for the cleanup code.
  The rational is that we will likely want to reuse that code later on in other threads (import threads and maybe some plugins threads)

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1501,10 +1547,68 @@ attr_syntax_init(void)
...
+        /* we have a duplicate, skip over it so we don't leak memory */
+        return HT_ENUMERATE_NEXT;
+    }
+
+    a_copy = attr_syntax_dup(a);
+    a_copy->asi_ht_copy = PR_TRUE;
+
+    /* Name table */
+    PL_HashTableAdd(tables->name, a_copy->asi_name, a_copy);
+    for (size_t i = 0; a_copy->asi_aliases && a_copy->asi_aliases[i] != NULL; ++i) {
```
progier389: should not name the variable i as a parameter is already named i

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1501,10 +1547,68 @@ attr_syntax_init(void)
     return 0;
 }
 
+static PRIntn
+attr_syntax_copy_insert(PLHashEntry *he, PRIntn i, void *arg)
```
progier389: should flag i as unused

ldap/servers/slapd/attrsyntax.c
mreynolds389: Fixed, but I'm surprised this didn't raise a compiler warning.  Nice catch :-)

mreynolds389:
> Still reviewing but I have a general comment:

This is technically still a WIP.  Need to do some performance testing with it.  I just wanted to get the working code into a PR :-)

> 
>     * I would have put the code that check version change and duplicate the hash table (in connection.c) in a separate function (and in libslapd rather than in ns-slapd)
>       And same thing for the cleanup code.
>       The rational is that we will likely want to reuse that code later on in other threads (import threads and maybe some plugins threads)
Sounds good, I'll work on that next...

progier389 (approved):
Except a few minor points, its looks fine. 
I hope we will see the impact in performance test.

P.S: Avoid doing massive code cleanup in a PR impacting sensitive code:
  It really makes the review much harder:
   For example I am not absolutely sure about what was really changed in schema.c
     because the cleanup just hide the functional changes ! -:(
    ( I think it is all around the DSE_SCHEMA_USE_GLOBAL flags but I am not absolutely sure.)

mreynolds389:
> Except a few minor points, its looks fine. I hope we will see the impact in performance test.
> 
> P.S: Avoid doing massive code cleanup in a PR impacting sensitive code: It really makes the review much harder: For example I am not absolutely sure about what was really changed in schema.c because the cleanup just hide the functional changes ! -:( ( I think it is all around the DSE_SCHEMA_USE_GLOBAL flags but I am not absolutely sure.)

I know I know, the cleanup started off simple and small, and then it turned into a can of worms.  But yes in schema.c the main change was to add a flag so we always update/access the global hash tables and not the local thread hash tables.

mreynolds389:
All changes made.  I hope to get performance testing done next week.

progier389 (approved):
LGTM

Firstyear:
> There is a lot of contention around the rwlocks for the attribute syntax hashtables. Since syntaxes rarely change we can just keep a copy in each thread and avoid locking.

Wouldn't contention on an rwlock imply that there are attempts to frequently write to these values? 

Rwlocks with .read() are effectively free since it becomes an atomic ref count. 

How did you determine there was contention here?

mreynolds389:
> > There is a lot of contention around the rwlocks for the attribute syntax hashtables. Since syntaxes rarely change we can just keep a copy in each thread and avoid locking.
> 
> Wouldn't contention on an rwlock imply that there are attempts to frequently write to these values?
> 
> Rwlocks with .read() are effectively free since it becomes an atomic ref count.
> 
> How did you determine there was contention here?

Thierry did this investigation in the parent Issue:

https://github.com/389ds/389-ds-base/issues/4598

Using modrate, and removing these RWlock gave a 5% performance increase.  I'm not sure what tool he used to measure the contention but that report is also in the parent issue.

Firstyear:
> I do not think that we are speaking of the usual lock contention here (as the lock is never held in write mode) but rather memory cache contention FYI: I also noticed the same thing when I run the performance test about lmdb import. In the final tests once dn cache was disabled, , according to '/usr/bin/perf c2c', the first four top contentions were in rw lock below attr_syntax_get_by_name_locking_optional (__pthread_rwlock_unlock and ___pthread_rwlock_rdlock)

Memory cache contention doesn't make sense though, because the pages will be in the shared state in the various cpu caches. By splitting this we'll be putting more pressure on the cpu caches because we are no longer sharing the pages. 

So I'm still really curious about this and the assumption about the cause that is occuring.

progier389:
As I told the you the perf c2c tool reports these as a the top cache contention so there are contention even if we do not understand why it happens ! -;)
That said I can understand that you would like to know exactly why it is the case.  I am a bit curious too (but not srprised as is not the first time that I have seen that a mostly read rw lock was surprisingly costly). 
For my part, I learned to trust the "/usr/bin/perf" tool because it helped me to understand why in my first tests the lmdb online import was ten time slower than the off line one (while the import algorithm is exactly the same) 
Anyway I guess we will see the result when we will be able to run the performance test on the new code.

Firstyear:
I would certainly trust perf here, but I think understanding why something happens is going to give a better solution than just splitting up a structure and copying it (which might reduce the contention but it's going to cause data-duplication in the cache meaning we get worse cache sharing and overall worse performance).

So understanding the problem seems like a better first step :)

mreynolds389:
### Initial Performance Results
-------------------------------------

This was all done on a single system (client and server)

#### Import

1-2% performance increase (only tested bdb)

#### Modrate

No improvement

#### Search Rate

##### Single Filter:  (uid=VALUE)

19% improvement

##### Compound Filter:  (&(uid=VALUE)(objectclass=posixaccount))

15% improvement 

#### Conclusion

The original issue said "modrate" got an improvement, but I don't see that.  However the search improvement is significant!

mreynolds389:
Also...  For the import test I also tried added hashtable copies to bdb_import_foreman(), bdb_import_producer(), and bdb_public_bdb_import_main(), but it did not seem to help import performance (with bdb).

mreynolds389:
Well I was playing around with the test cases and when I reran the same tests I ran above the performance improvement was very small.  Sigh.  I ran those tests multiple times with each version and the results were very consistent (15-19% improvement), now they are not.  I think I need to wait until we get those dedicated lab machines to run these tests correctly (separating client from server).

mreynolds389:
OK testing on large lab machines shows no noticeable improvement with this fix :-(   That doesn't mean this should be closed, as we are most likely hitting another bottle neck, so this fix "could" improve things down the road when other issues are addressed

tbordaz:
@mreynolds389 do you know why this PR was stopped ? the idea looked promissing but was not merged any reason for that ?

progier389:
@tbordaz, If I remember rightly, it was not committed because there was no noticeable improvement with these changes ...

tbordaz:
> @tbordaz, If I remember rightly, it was not committed because there was no noticeable improvement with these changes ...

I agree it was also my remembering. However I am surprised, syntax are protected and even if they are always free it should create some pressure on the lock moving to all CPUs. May be we should revisit the testcase before closing this ticket.

mreynolds389:
> > @tbordaz, If I remember rightly, it was not committed because there was no noticeable improvement with these changes ...
> 
> I agree it was also my remembering. However I am surprised, syntax are protected and even if they are always free it should create some pressure on the lock moving to all CPUs. May be we should revisit the testcase before closing this ticket.

@tbordaz @progier389 

Looking back I just did modrate & searchrate testing and at first I saw a huge improvement, but then I didn't.  I have a feeling I was using inconsistent environments. So I would love to see this go through performance testing again.

mreynolds389:
Ok, I still did not see any improvement with modrate or searchrate using 10 threads.  But... when I bumped it up to 32 threads (on a 22 core system), I saw almost a 10% improvement with searchrate, although modrate did not improve (using LMDB).

So this is still worth pursuing.  I plan to do more tests and I will put together a summary report.

tbordaz:
I believe those additional paths would get benefit of the patch
- ldif import
- checking the filter against the schema
- preparation of idl
- virtual attributes (cos)

You may define long filters. I am not sure that the number of matching entries is crucial.
Define one virtual attributes. I am not sure that it requires to add it in the filter.
Having a second backend that you may import should disturb the searchrate on the first backend

PR #5698, opened 2023-03-15
Issue 5697 - Obsolete nsslapd-ldapimaprootdn attribute

### LINKED ISSUE #5697: Obsolete nsslapd-ldapimaprootdn attribute
Description of problem:
LDAPI code uses nsslapd-ldapimaprootdn to map Unix root entry to a Root DN entry. It usually has the same value as nsslapd-rootdn. Changing one attribute but not changing the other leads to a non-functional autobind configuration that breaks dsconf and WebUI.

LDAPI code should use nsslapd-rootdn value instead of having two separate entries that should be kept in sync. This should make changing Root DN simpler and without fear that it will break dsconf or WebUI access.

See https://github.com/389ds/389-ds-base/issues/5453 for context.

droideck:
Description: LDAPI code uses nsslapd-ldapimaprootdn to map Unix root entry to a Root DN entry. It usually has the same value as nsslapd-rootdn. Changing one attribute but not changing the other leads to a non-functional autobind configuration that breaks dsconf and WebUI.

LDAPI code should use nsslapd-rootdn value instead of having two separate entries that should be kept in sync. This should make changing Root DN simpler and without fear that it will break dsconf or WebUI access.

Fixes: https://github.com/389ds/389-ds-base/issues/5697

Reviewed by: ?

### DIFF
--- a/ldap/ldif/template-dse-minimal.ldif.in
+++ b/ldap/ldif/template-dse-minimal.ldif.in
@@ -23,7 +23,6 @@ nsslapd-rootpw: %ds_passwd%
 nsslapd-ldapilisten: %ldapi_enabled%
 nsslapd-ldapifilepath: %ldapi%
 nsslapd-ldapiautobind: %ldapi_autobind%
-nsslapd-ldapimaprootdn: %rootdn%
 
 dn: cn=auto_bind,cn=config
 objectclass: top
--- a/ldap/ldif/template-dse.ldif.in
+++ b/ldap/ldif/template-dse.ldif.in
@@ -24,7 +24,6 @@ nsslapd-rootpw: %ds_passwd%
 nsslapd-ldapilisten: %ldapi_enabled%
 nsslapd-ldapifilepath: %ldapi%
 nsslapd-ldapiautobind: %ldapi_autobind%
-nsslapd-ldapimaprootdn: %rootdn%
 
 dn: cn=encryption,cn=config
 objectClass: top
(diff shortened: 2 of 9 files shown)


droideck:
The core part of the fix is in - `ldap/servers/slapd/ldapi.c`.

```
-            char *root_dn = config_get_ldapi_root_dn();
+            char *root_dn = config_get_rootdn();
```

For the rest of the fix, I've put warnings so when admin sets nsslapd-ldapimaprootdn attribute, he knows it's obsolete.

As an alternative, I was considering the complete `nsslapd-ldapimaprootdn` removal but I don't think we can do it in the middle of RHEL 9 cycle.
 
@mreynolds389 what do you think? Should we fully remove the attribute or just use the warnings (as done in my PR)?

Firstyear:
I think we can't remove the attribute at all because if we remove it we'll trigger an "unknown" attribute error on startup? I seem to recall that being an issue

droideck:
> I think we can't remove the attribute at all because if we remove it we'll trigger an "unknown" attribute error on startup? I seem to recall that being an issue

We've removed one recently: https://github.com/389ds/389-ds-base/commit/6bc98164d3b4dc76f3d0de391292bc98ac9b041d
I think we can remove it, but I'm not sure if we should :)

Firstyear:
In that case, then "yolo" and remove it :)

mreynolds389 (approved):
If ipa install succeeds this gets my ack

droideck:
> In that case, then "yolo" and remove it :)

I think it'll be safer to make it gradually so the administrators can adopt their envs in a predictable manner. It just feels a bit "harsh" to remove the attribute out of nowhere without any predictability for our users.

> If ipa install succeeds this gets my ack

Done. ipa-server-install works fine! Merging...

Issue #5700, opened 2023-03-16
SRCH performance drop 2.0 vs 1.4


tbordaz:
**Issue Description**
A testcase reveal a significant (~30times) performance drop that was introduced in #5170 [commit](https://github.com/389ds/389-ds-base/commit/40563667e12da3042ea5a0e3771c1e1db1dafa06) 
Sometime an indexed search filter can lead to a candidate list that only contains entries matching the filter. In such case, it is useless to reevaluate the filter against each returned entries. It is called bypass filter (SR_FLAG_MUST_APPLY_FILTER_TEST==0).
The above commit force the evaluation of the filter in all cases.

The problem is that filter evaluation is expensive by itself and become very expensive if the returned entry contains attributes with large number of values. For example a uniquemember of a big group.

**Package Version and Platform:**
2.x

**Steps to Reproduce**
testcase will be provided later

**Expected results**
2.x should be as fast as 1.4

Firstyear:
The filter test has to be forced though for the reasons discussed repeatedly through that issue - previously we actually had potential for query corruption in our search code. So I think any changes here need to be extremely carefully considered.

PS: We pretty much always force filter tests because grok filter is busted and should be removed.

tbordaz:
@Firstyear , I fully agree with you this subject was discussed. However the pointed commit has a severe performance hit. It was pushed with a testcase and reworking the patch the testcase is still pass without the perf hit.
I will send a PR (hopefully today) to check if the orginal patch or original testcase need to be improved

tbordaz:
It is a leftover that I forgot to update. @Firstyear is correct the testcase is not valid and the suggested PR was wrong. I am closing this ticket

PR #5703, opened 2023-03-17
Issue 5700 - SRCH performance drop 2.0 vs 1.4

tbordaz:
Bug description:
	The commit 40563667e12da3042ea5a0e3771c1e1db1dafa06
	force the evaluation of the filter before returning
	entries. This evaluation can be skipped in case
	of indexed search and filter syntax without OR
	(sr_flags ~SR_FLAG_MUST_APPLY_FILTER_TEST).
	The consequence of the fix is a severe drop of
	search performance.
	The testcase provided with the commit are PASS
	without part of the original fix

Fix description:
	Revert parts of the fix.
	This patch intend to identify new testcase
	that confirm the need to force the evaluation
	of the filter

relates: #5700

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_search.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_search.c
@@ -34,7 +34,7 @@ static IDList *base_candidates(Slapi_PBlock *pb, struct backentry *e);
 static IDList *onelevel_candidates(Slapi_PBlock *pb, backend *be, const char *base, Slapi_Filter *filter, int *lookup_returned_allidsp, int *err);
 static back_search_result_set *new_search_result_set(IDList *idl, int vlv, int lookthroughlimit);
 static void delete_search_result_set(Slapi_PBlock *pb, back_search_result_set **sr);
-static int can_skip_filter_test(Slapi_PBlock *pb, struct slapi_filter *f, int scope, IDList *idl);
+static int can_skip_filter_test(Slapi_PBlock *pb, struct slapi_filter *f, int scope, IDList *idl, int sr_flags);
 
 /* This is for performance testing, allows us to disable ACL checking altogether */
 #if defined(DISABLE_ACL_CHECK)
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/filterentry.c
```diff
@@ -770,12 +770,6 @@ slapi_vattr_filter_test_ext(
     int rc = 0; /* a no op request succeeds */
     int access_check_done = 0;
 
-    if (only_check_access != 0) {
-        slapi_log_err(SLAPI_LOG_ERR, "slapi_vattr_filter_test_ext",
-            "⚠️  DANGER ⚠️  - only_check_access mode is BROKEN!!! YOU MUST CHECK ACCESS WITH FILTER MATCHING");
```
Firstyear, line 775: This comment was not incorrect. This is broken, and can lead to access control bypasses.

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1753,15 +1753,15 @@ ldbm_back_next_search_entry(Slapi_PBlock *pb)
                         slapi_log_err(SLAPI_LOG_FILTER, "ldbm_back_next_search_entry",
                                       "Bypassing filter test for %s\n", slapi_entry_get_dn_const(e->ep_entry));
                         if (ACL_CHECK_FLAG) {
-                            filter_test = slapi_vattr_filter_test(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG);
+                            filter_test = slapi_vattr_filter_test_ext(pb, e->ep_entry, filter_intent, ACL_CHECK_FLAG, 1);
```
Firstyear, line 1757: we can't set this. This allows access control bypasses and is a CVE.

Firstyear (changes requested):
I know you want to try to fix this to remove the extra filter tests, but I had to re-add them because they are correct. There were access control bypasses and query corruptions here.

I know it sucks that it affects performance, but setting this flags just undoes all that work I did to fix these to be correct and secure. 

The existing code is *already* the correct set of filter tests and access controls to prevent these bypasses. 

I can't approve this, because it would introduce vulnerabilities. 

If you want to actually improve this, then *remove grok_filter*.

Grok filter is *fundamentally broken* and always forces filter tests even if they are not needed!

Firstyear:
@tbordaz The change you are proposing here is a straight up security bypass. I can't in good consciousness approve this. 

The comments I have left in filterentry.c are correct - deleting them doesn't change anything. The only_check_access mode is broken because in only_check_access mode it doesn't actually properly apply access controls. 

I don't know why you are so hesitant to listen to me and *just remove grok filter*. 

Look at grok filter. If it sees an OR condition it will always force filter tests. If it sees anything that isn't an EQ filter it forces the test. Look at it https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/ldbm_search.c#L1307 Anything that returns '0' here forces a filter test. look at how many conditions will have the test forced. 

This one function pretty much prevents us from being able to skip filter tests. Which leaves us having to double filter test - one for the "filter as executed" due to referrals and our optimisation and re-arrangement. And one for access control.

mreynolds389:
:-)  In @tbordaz defense, I think this was a WIP PR.  I think Thierry is just making commits as he tests new things.  I don't think he's ignoring your comments.  I think he's just exploring the "space".  Keep in mind he is also working on other things at the same time so he might not be able to respond very quickly.

Firstyear:
We can't trade performance for security though. That's my major concern here because previously we were exposed to aci bypass as well as query corruption.

I think you really will benefit greatly from looking to drop grok filter.

PR #5705, opened 2023-03-20
Issue 1081 - Stop schema replication from overwriting x-origin

mreynolds389:
Bug Description:

During schema replication all attributes/objectclasses were rewritten as "user defined" on the consumer.  This was happening because we treated all schema updates, regardless of the origin, as new "custom" schema.

Fix Description:

If a schema update is a replicated operation do not check/adjust x_origin value.

relates: https://github.com/389ds/389-ds-base/issues/1081


progier389 (approved):
LGTM

PR #5706, opened 2023-03-20
Issue 4812 - Listener thread does not scale with a high num of established connections

jchapma:
Bug description: Latest commit introduced zero alloc bug.

Fix description: Memory allocation of a buffer that tracks the number of 
                            active connections per list is attempted before we retrieve 
                            the number of lists value from config. Copy and paste error 
                            introduced during rework. 

relates:  https://github.com/389ds/389-ds-base/issues/4812

Reviewed by:


mreynolds389 (approved):
LGTM, thanks!

PR #5708, opened 2023-03-24
Issue 5701 - CLI - Fix referral mode setting

### LINKED ISSUE #5701: Referral mode is not working
Description of problem:

Referral mode not working and failing with error: ERROR: Error: 103 - 10 - 53 - Server is unwilling to perform - [] -
need to set nsslapd-referral

How reproducible:
Every time

Steps to Reproduce:
1. Create two instances

2. Stop the instance on the host you want to redirect:

```
# systemctl stop dirsrv@instance_name
```

3. Run ns-slapd in referral mode:

```
# ns-slapd refer -D /etc/dirsrv/slapd-instance_name [-p port] -r referral_url
```

4.Even though nsslapd-referral is set:

```
cn=data_on_1,cn=ldbm database,cn=plugins,cn=config
...
nsslapd-referral: ldap://localhost:38902/ou=people,dc=example,dc=com
```

Updating the state doesn't work:

```# dsconf -v instance1 backend suffix set --state referral
ou=data_on_instance1,dc=example,dc=com
...
ldap.UNWILLING_TO_PERFORM: {'msgtype': 103, 'msgid': 10, 'result': 53,
'desc': 'Server is unwilling to perform', 'ctrls': [], 'info': 'need
to set
 nsslapd-referral before moving to referral state\n'}
```

droideck:
Bug Description: Referral mode not working and failing with error: ERROR: Error: 103 - 10 - 53 - Server is unwilling to perform - [] - need to set nsslapd-referral It happens because in CLI, we set nsslapd-referral to the backend when it should be set to "cn=mapping tree".

Fix description: Set the attribute to the correct object. Add get_state and set_state custom functions to BackendSuffixView. Fix minor typos.

Fixes: https://github.com/389ds/389-ds-base/issues/5701

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -1165,13 +1165,36 @@ class BackendSuffixView(CompositeDSLdapObject):
             'nsslapd-cachesize',
             'nsslapd-dncachememsize',
             'nsslapd-readonly',
-            'nsslapd-referral',
             'nsslapd-require-index',
+            'nsslapd-suffix'
         ]
         mt_args = [
             'orphan',
             'nsslapd-state',
+            'nsslapd-referral',
         ]
         mt = be._mts.get(be.get_suffix())
         self.add_component(be, be_args)
         self.add_component(mt, mt_args)
+
+    def get_state(self):
+        return self.get_attr_val_utf8('nsslapd-state')
+
+    def set_state(self, new_state):
+        new_state = new_state.lower()
+        suffix = self.get_attr_val_utf8('nsslapd-suffix')
+
+        if new_state not in ['backend', 'disabled',  'referral',  'referral on update']:
+            raise ValueError(f"Invalid backend state {new_state}, value must be one of the following: 'backend', 'disabled',  'referral',  'referral on update'")
+
+        # Can not change state of replicated backend
+        replicas = Replicas(self._instance)
+        try:
+            # Check if replication is enabled
+            replicas.get(suffix)
+            raise ValueError("Can not change the backend state of a replicated suffix")
+        except ldap.NO_SUCH_OBJECT:
+            pass
+
+        # Ok, change the state
+        self.set('nsslapd-state', new_state)
--- a/src/cockpit/389-console/src/lib/database/referrals.jsx
+++ b/src/cockpit/389-console/src/lib/database/referrals.jsx
@@ -272,7 +272,7 @@ export class SuffixReferrals extends React.Component {
                     spinning={this.state.modalSpinning}
                     item={this.state.removeRef}
                     checked={this.state.modalChecked}
-                    mTitle="Delete Refferal"
+                    mTitle="Delete Referral"
                     mMsg="Are you sure you want to delete this referral?"
                     mSpinningMsg="Deleting ..."
                     mBtnName="Delete"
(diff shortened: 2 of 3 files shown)


src/lib389/lib389/backend.py
```diff
@@ -1165,13 +1165,36 @@ def __init__(self, instance, be):
...
+
+        if new_state not in ['backend', 'disabled',  'referral',  'referral on update']:
+            raise ValueError(f"Invalid backend state {new_state}, value must be one of the following: 'backend', 'disabled',  'referral',  'referral on update'")
+
+        # Can not change state of replicated backend
+        replicas = Replicas(self._instance)
+        try:
+            # Check if replication is enabled
+            replicas.get(suffix)
+            raise ValueError("Can not change the backend state of a replicated suffix")
```
tbordaz, line 1195: I am unsure to understand this restriction. If I have a consumer/hub, its mapping tree will eventually be 'referral on update'. With this test, will it be possible to set a consumer/hub mapping tree on 'referral on update' ?

tbordaz (commented):
The overall patch looks good, just a question regarding replicated mapping tree.

src/lib389/lib389/backend.py
droideck: It was added some time ago here by Mark:
https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/backend.py#L895
I got it from there to preserve the existing behaviour. :)

@mreynolds389 do you recall the reason for the check?
Is it adjusted automatically for consumer/hub mapping tree?

mreynolds389: I recall it being a requirement.  Does it cause any problems?

tbordaz: @mreynolds389 , if it is called for a replicated backend with mt.set_state('referral on update') then it will raise the exception "Can not change the backend state of a replicated suffix". IIRC for a consumer/hub nsslapd-state needs to be 'referral on update' will it be done ?

progier389: I think that the restriction is fine. There is no hard requirement but the replication plugin regularly resets that state trough repl_set_mtn_state_and_referrals, so allowing an user to set the mapping tree state on a replica is very tricky:
An administrator may think he set  safely the backend in read-only mode 
  but replication may reset it to rw mode behind the scene ...

tbordaz: Good point. Indeed the server itself may reset the state, mainly as referral if REPLICA_TYPE_READONLY and also during total update.
However I still have a doubt as this patch is about a CLI enhancement to allow an administrator to switch a backend/suffix into a referral (see '[Creating suffix referrals](https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/administration_guide/configuring_directory_databases-using_referrals#Using_Referrals-Creating_Suffix_Referrals)'). The is no documented restriction regarding replication.

tbordaz: Opps... my mistake. I realize I missed your point @progier389 . Setting referral/backend on a replicated suffix is useless because it will be reset. You are correct.
The procedure in admin guide should state that it does not apply on replicated suffix.
The patch is good. Ack

tbordaz (approved):
Thanks for the explanations. ACK

Issue #5710, opened 2023-03-27
subtree search statistics for index lookup does not report ancestorid/entryrdn lookups


tbordaz:
**Issue Description**
The RFE #3729 allows to collect index lookups per search operation. For subtree search it does not record/log the statistics regarding the indexes ancestorid/entryrdn

**Package Version and Platform:**
Since 2.2 (#3729)

**Steps to Reproduce**
Enables statistics (dsconf instance config replace nsslapd-statlog-level=1), do a subtree search with base=suffix.
In access logs ancestorid lookups are not logged

**Expected results**
ancestorid id should be logged

tbordaz:
7915e85a5..fca27c3d0  389-ds-base-1.4.3

PR #5711, opened 2023-03-27
Issue 5710 - subtree search statistics for index lookup does not repo…

tbordaz:
…rt ancestorid/entryrdn lookups

Bug description:
	The RFE #3729 allows to collect index lookups per search
	operation. For subtree searches the server lookup ancestorid
	and those lookup are not recorded

Fix description:
	if statistics are enabled, record ancestorid lookup

relates: #5710

Reviewed by:


mreynolds389 (approved):
LGTM

Issue #5707, opened 2023-03-23
RFE - Add config parameter to close client connections on failed BIND


jchapma:
**Problem statement**
Malformed applications that ignore BIND return code can load the server with unnecessary requests.

**Solution**
Add a config option that will allow the closure of a client connection from server side when a BIND is failing.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=1987471

mreynolds389:
Bugzilla is set for RHEL 9.3, please cherry pick this to the 2.3 branch.  There should probably be a UI update for this as well, but it can wait for now...

jchapma:
> Bugzilla is set for RHEL 9.3, please cherry pick this to the 2.3 branch. There should probably be a UI update for this as well, but it can wait for now...

 fc34eec58..7743b3862  389-ds-base-2.3 -> 389-ds-base-2.3
UI patch ready

PR #5712, opened 2023-03-27
Issue 5705 - Add config parameter to close client conns on failed bind

jchapma:
Description: Malformed applications that ignore BIND return code can load the server with unnecessary requests

Fix description: Add a config option that will allow the closure of a client connection from server side when a BIND is failing.

relates: https://github.com/389ds/389-ds-base/issues/5707

Reviewed by: ??

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -51,6 +51,24 @@ ROOTDSE_DEF_ATTR_LIST = ('namingContexts',
                          'vendorVersion')
 
 
+@pytest.fixture(scope="function")
+def _reset_attr(request, topology_st):
+    """ Reset nsslapd-close-on-failed-bind attr to the default (off) """
+
+    def fin():
+        dm = DirectoryManager(topology_st.standalone)
+        try:
+            dm_conn = dm.bind()
+            dm_conn.config.replace('nsslapd-close-on-failed-bind', 'off')
+            assert (dm_conn.config.get_attr_val_utf8('nsslapd-close-on-failed-bind')) == 'off'
+        except ldap.LDAPError as e:
+            log.error('Failure reseting attr')
+            assert False
+        topology_st.standalone.restart()
+
+    request.addfinalizer(fin)
+
+
 @pytest.fixture(scope="module")
 def import_example_ldif(topology_st):
     """Import the Example LDIF for the tests in this suite"""
@@ -1508,7 +1526,7 @@ def test_suffix_case(topology_st):
     assert domain.dn == TEST_SUFFIX
 
 
-def test_bind_disconnect_invalid_entry(topology_st):
+def test_bind_disconnect_invalid_entry(topology_st, _reset_attr):
     """Test close connection on failed bind with invalid entry
 
     :id: b378543e-32dc-432a-9756-ce318d6d654b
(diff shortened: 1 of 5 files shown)


droideck (commented):
Could you please add a simple test case for the setting?

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1506,7 +1507,287 @@ def test_suffix_case(topology_st):
...
+def test_bind_disconnect_cert_map_failed(topology_st):
+    """Test close connection on failed bind with a failed cert mapping
+
+    :id: 0ac60f76-1fd9-4080-a82b-21807e6bc292
+    :setup: Standalone Instance
+    :steps:
+        1. enable TLS
+        2. create/get a user
+        3. get details of ssca key and cert
+        4: create 2 user certificates, one good, one bad
```
droideck: Please, fix the Validate Tests CI issues

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1506,7 +1507,287 @@ def test_suffix_case(topology_st):
...
+    dm = DirectoryManager(inst)
+
+    # create/get user
+    users = UserAccounts(inst, DEFAULT_SUFFIX)
+    try:
+        user = users.create_test_user()
+        user.set("userPassword", PW_DM)
+    except ldap.ALREADY_EXISTS:
+        user = users.get('test_user_1000')
+        pass
```
droideck: The `pass` is not needed here and in other places in the tests, because the `pass` statement is used as a placeholder for future code.

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1506,7 +1507,287 @@ def test_suffix_case(topology_st):
...
+
+    # verify user can bind and search
+    try:
+        inst.simple_bind_s(user.dn, PW_DM)
+    except ldap.LDAPError as e:
+        log.info('exception description: ' + e.args[0]['desc'])
```
droideck: Please, avoid the "silent" exceptions in the test code.
The test should always fail if something unwanted happens.
And, I think, you can safely use `log.error()` here instead of `log.info()`.

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1506,7 +1507,287 @@ def test_suffix_case(topology_st):
...
+    with pytest.raises(ldap.SERVER_DOWN):
+        inst.search_s(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE, '(objectclass=top)', ['dn'])
+    try:
+        dm_conn = dm.bind()
+    except ldap.LDAPError:
+        log.info('dm bind failed')
+
+    # cleanup
+    try:
+        dm_conn.config.replace('nsslapd-close-on-failed-bind', 'off')
```
droideck: If the test will fail before that point, the cleanup won't happen.
So please, use either a pytest fixture (for setup-cleanup), or either `finally:` block with `try:` block through the whole test.

droideck (changes requested):
A few minor issues, the rest looks good!

droideck (approved):
Please, fix Validate Test CI.
image
Besides that, looks good! Ack

Issue #5704, opened 2023-03-18
crash in sync_refresh_initial_content


abbra:
When running FreeIPA on Rawhide in Azure CI I occasionally get crashes in the directory server plugins. This time it is ldap/servers/plugins/sync/sync_refresh.c:413.

```
Core was generated by `/usr/sbin/ns-slapd -D /etc/dirsrv/slapd-IPA-TEST -i /run/dirsrv/slapd-IPA-TEST.'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  sync_refresh_initial_content (pb=pb@entry=0x7f6e37a00000,
    sync_persist=sync_persist@entry=0, tid=tid@entry=0x0, sc=0x0)
    at ldap/servers/plugins/sync/sync_refresh.c:413
413         if (sc->openldap_compat == PR_TRUE) {
```
[20831.stacktrace.txt](https://github.com/389ds/389-ds-base/files/11008274/20831.stacktrace.txt)

abbra:
Packages:
```
389-ds-base-2.3.2-2.fc39.x86_64
389-ds-base-libs-2.3.2-2.fc39.x86_64
```

abbra:
Coredumpt collector logs are below. We had two crashes, one in `ns-slapd`, one in `agetty`, but the other one is not related (it looks like an issue in glibc with architecture-specific optimisations not working well).
```
2023-03-17T13:48:15.1761305Z + mkdir /freeipa/ipa_envs/coredumps
2023-03-17T13:48:15.1777027Z + cd /freeipa/ipa_envs/coredumps
2023-03-17T13:48:15.1779270Z ++ coredumpctl --no-pager --directory=/var/log/host_journal '--since=2023-03-17 13:22:08' -F COREDUMP_PID
2023-03-17T13:48:15.1830534Z + pids='20831
2023-03-17T13:48:15.1831279Z 23279'
2023-03-17T13:48:15.1831685Z + for pid in $pids
2023-03-17T13:48:15.1832620Z + coredumpctl --no-pager '--since=2023-03-17 13:22:08' --directory=/var/log/host_journal -o 20831.core dump 20831
2023-03-17T13:48:15.1900108Z            PID: [0;1;39m20831[0m (ns-slapd)
2023-03-17T13:48:15.1900797Z            UID: 389 (dirsrv)
2023-03-17T13:48:15.1901222Z            GID: 389 (dirsrv)
[...]

PR #5720, opened 2023-04-04
Issue 5704 - crash in sync_refresh_initial_content

tbordaz:
Bug description:
	If the last record of the changelog is not accessible
	then the session record is NULL. It crashes the server
	when it is dereferenced.
	I failed to reproduce it, including disabling/removing
	'cn=changelog' backend/mapping tree. So I guess it
	happens during rare dynamic.

Fix description:
	Return a failure when the session cookie is not
	initialized

relates: #5704

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/sync/sync_refresh.c
+++ b/ldap/servers/plugins/sync/sync_refresh.c
@@ -100,7 +100,22 @@ sync_srch_refresh_pre_search(Slapi_PBlock *pb)
              * consumer.
              */
             session_cookie = sync_cookie_create(pb, client_cookie);
-            PR_ASSERT(session_cookie);
+            if (session_cookie == NULL) {
+                /* In some rare case access to the retroCL fails.
+                 * It can happen when retroCL is just created and
+                 * does not contain any record.
+                 * As we are not able to retrieve the last changenumber
+                 * just return a failure.
+                 * Another option would be to set cookie_change_info=0
+                 * if we can not retrieve any record in retroCL
+                 * (in sync_cookie_create)
+                 */
+                slapi_log_err(SLAPI_LOG_ERR, SYNC_PLUGIN_SUBSYSTEM,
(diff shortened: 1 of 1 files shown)


abbra:
I keep seeing it crashing in IPA upgrade tests on Azure CI. `sync_cookie_get_change_info()` does an internal search for `cn=changelog` and it does not return an entry at all. This causes an error in generating a session cookie.

tbordaz:
> I keep seeing it crashing in IPA upgrade tests on Azure CI. `sync_cookie_get_change_info()` does an internal search for `cn=changelog` and it does not return an entry at all. This causes an error in generating a session cookie.

Yes that is the scenario that this patch tries to eliminate. Is it still crashing with that patch ?

abbra:
I haven't tried. Do you have a rawhide repo somewhere?

tbordaz:
> I haven't tried. Do you have a rawhide repo somewhere?

Here is a build main+fix on [copr](https://download.copr.fedorainfracloud.org/results/tbordaz/F20buildDS389/fedora-rawhide-x86_64/05745109-389-ds-base/) repo

abbra:
Thanks, so far I have not been able to reproduce a crash with this COPR build. But I haven't been able to reproduce it before locally as well...

ldap/servers/plugins/sync/sync_refresh.c
```diff
@@ -101,6 +101,13 @@ sync_srch_refresh_pre_search(Slapi_PBlock *pb)
              */
             session_cookie = sync_cookie_create(pb, client_cookie);
             PR_ASSERT(session_cookie);
```
mreynolds389: Should probably remove this assert since session_cookie does have a valid reason for being NULL.

tbordaz: Correct, I removed it.

mreynolds389 (approved):
LGTM

Issue #5722, opened 2023-04-07
RFE When a filter contains 'nsrole', improve response time by rewriting the filter.


tbordaz:
**Issue Description**
'nsrole' is a virtual attribute and is not indexed. With a poorly selective filter like below the search may be not indexed 
<pre>
"(&(nsrole=cn=managed_role,cn=suffix)(objectclass=posixAccount)))"
</pre>

The fix is to rewrite filter components containing 'nsrole' type. ATM the fix would be limited to managed/filtered roles

**Package Version and Platform:**
all versions

**Steps to Reproduce**
provide a test case

**Expected results**
Getting faster SRCH when filter uses 'nsrole'

tbordaz:
a55fa3087..bcbad8745  main

mreynolds389:
@tbordaz  please backport to all relevant branches.  This is expected to be in 2.3 branch (and probably a good idea to put it in earlier branches as well).

tbordaz:
cde047740..f2f6c2b9e  389-ds-base-2.3

tbordaz:
Improving the testcase with [PR](https://github.com/389ds/389-ds-base/pull/5904)

475ee01dc..63fd271fa main
11a1ddc8f..1df340797  389-ds-base-2.3
ababe094c..f1eb2f1d6  389-ds-base-2.2
96012932f..0747f7694  389-ds-base-2.1
849cc42b2..5881562a9  389-ds-base-2.0
e721c2f13..c70be2e97  389-ds-base-1.4.4
c5344e8c6..837b47d68  389-ds-base-1.4.3

PR #5723, opened 2023-04-07
Issue 5722 - RFE When a filter contains 'nsrole', improve response ti…

tbordaz:
…me by rewriting the filter

Bug description:
	'nsrole' is a virtual attribute and is not indexed.
         With a poorly selective filter like below the search may be not indexed
	"(&(nsrole=cn=managed_role,cn=suffix)(objectclass=posixAccount)))"

	The RFE is to rewrite the filter component contains 'nsrole'
        attribute type.
	Rewritten component can then been indexed

Fix description:
	For managed role, it replaces 'nsrole' with 'nsroleDN'
        attribute type

	For filtered roled, it replace the 'nsrole' component
	with the nsRoleFilter value

relates: #5722

Reviewed by:


ldap/servers/plugins/roles/roles_cache.c
```diff
@@ -2131,3 +2131,139 @@ roles_cache_dump(caddr_t data, caddr_t arg __attribute__((unused)))
 
     return 0;
 }
+
+
+/* This is an example callback to substitute
```
progier389, line 2136: Cut and paste error:  
that comment are wrong (this code is no more an example and I would like to see described somewhere what filter ava is changed and by what it is replaced ...)
i.e: nsrole=some_dn is replaced by:
    nsRoleDN=some_dn if some_dn is a simple or managed role
    the filter in nsRoleFilter attribute value of some_dn entry if some_dn is a filtered role

Note: the same remark applies for the other functions comments too ...

ldap/servers/plugins/roles/roles_cache.c
```diff
@@ -2131,3 +2131,139 @@ roles_cache_dump(caddr_t data, caddr_t arg __attribute__((unused)))
...
+
+    /* Substitution is only valid if original attribute is NSROLEATTR */
+    if ((substitute_arg == NULL) ||
+        (substitute_arg->attrtype_from == NULL) ||
+        (substitute_arg->attrtype_to == NULL)) {
+        return;
+    }
+    if (strcasecmp(substitute_arg->attrtype_from, NSROLEATTR)) {
+        return;
+    }
```
progier389, line 2169: IMPORTANT ! Should also check that it is an equality filter

ldap/servers/plugins/roles/roles_cache.c
```diff
@@ -2131,3 +2131,139 @@ roles_cache_dump(caddr_t data, caddr_t arg __attribute__((unused)))
...
+
+    /* Check that assertion does not refer to a filter/nested role */
+    if (slapi_filter_get_ava(f, &type, &bval)) {
+        return;
+    }
+    sdn = slapi_sdn_new_dn_byref(bval->bv_val);
+    if (slapi_search_internal_get_entry(sdn, attrs, &nsrole_entry, roles_get_plugin_identity())) {
+        goto bail;
```
progier389, line 2201: Could probably do something much clever in this case:
if the role does not exists then the filter does not match anything. 
 (goto bail means having an unindexed search)

ldap/servers/slapd/filter.c
```diff
@@ -1032,6 +1032,52 @@ slapi_filter_get_subfilt(
     return (0);
 }
 
+/* copy/merged from ldmb_attr.c:free_the_filter_bits/replace_filter
+ * it replaces the bits of the Slapi_Filter with the ones taken from strfilter
+ */
+int
+slapi_filter_replace_strfilter(Slapi_Filter *f, char *strfilter)
```
progier389: should probably move free_the_filter_bits and replace_filter out of ldbm_attr.c here as slapi_ function
 and use them in ldbm_attr ( with a flag: slapi_filter_replace_strfilter(Slapi_Filter *f, char *strfilter, int free_bits)

progier389 (changes requested):
an important issue that you should absolutely take in account in the hotfix you plan to deliver (you cannot replace the filter component if it is not an equality filter because in that case you do not know which role applies ...)
and a few minor remarks that are not mandatory in the hotfix but should be taken in account in upstream release

progier389 (approved):
OK for the hotfix as the major issue is now fixed.

ldap/servers/plugins/roles/roles_cache.c
tbordaz: You are right. If the role does not exist then this component should return no ID. It could be replaced with indexed attribute type and not valid value like '(entryid=-2)'

tbordaz: I did the change so that a component with a not existing role will be replaced by an indexed component with IDL=0. "(nsuniqueid=-1)"

ldap/servers/slapd/filter.c
```diff
@@ -1032,6 +1032,72 @@ slapi_filter_get_subfilt(
...
+/*
+ * it replaces the bits of the Slapi_Filter with the ones taken from strfilter
+ */
+int
+slapi_filter_replace_strfilter(Slapi_Filter *f, char *strfilter)
+{
+    slapi_filter_free_bits(f);
+    return (slapi_filter_replace_ex(f, strfilter));
+}
+
```
tbordaz, line 1100: Those slapi functions are from cut/paste from ldbm_attr.c because they are also needed to craft the filter in roles

ldap/servers/slapd/filter.c
tbordaz: I did move those functions from ldbm_attr to filter.c

tbordaz:
@progier389 , do you mind having a second look at recent changes ? (testcases, handling of filter component with non existing role, move functions to slapi_filter_..

progier389:
> OK for the hotfix as the major issue is now fixed.

LGTM !

Issue #5646, opened 2023-02-10
Various memory leaks


mreynolds389:
**Issue Description**

```
Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c708ba097 in calloc (/lib64/libasan.so.8+0xba097)
    #1 0x7f5c7006ac25 in slapi_ch_calloc (/usr/lib64/dirsrv/libslapd.so.0+0x26ac25)
    #2 0x7f5c7006b7a1 in get_ldapmessage_controls_ext (/usr/lib64/dirsrv/libslapd.so.0+0x26b7a1)
    #3 0x55639e2aadee in do_search ldap/servers/slapd/search.c:349
    #4 0x55639e2813fe in connection_dispatch_operation ldap/servers/slapd/connection.c:677
    #5 0x55639e2813fe in connection_threadmain ldap/servers/slapd/connection.c:1805
    #6 0x7f5c70f39412 in _pt_root (/lib64/libnspr4.so+0x2c412)
    #7 0x7f5c6fa8cdec in start_thread (/lib64/libc.so.6+0x8cdec)
    #8 0x7f5c6fb1236f in clone3 (/lib64/libc.so.6+0x11236f)


Indirect leak of 25 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c708ba6af in __interceptor_malloc (/lib64/libasan.so.8+0xba6af)
    #1 0x7f5c70fc6ed4 in ber_memalloc_x (/lib64/liblber.so.2+0x6ed4)
    #2 0x7f5c70fc6f98 in ber_get_stringbv (/lib64/liblber.so.2+0x6f98)
    #3 0x7f5c70fc734d in ber_get_stringa (/lib64/liblber.so.2+0x734d)
    #4 0x7f5c70fc9820 in ber_scanf (/lib64/liblber.so.2+0x9820)
    #5 0x7f5c7006b813 in get_ldapmessage_controls_ext (/usr/lib64/dirsrv/libslapd.so.0+0x26b813)
    #6 0x55639e2aadee in do_search ldap/servers/slapd/search.c:349
    #7 0x55639e2813fe in connection_dispatch_operation ldap/servers/slapd/connection.c:677
    #8 0x55639e2813fe in connection_threadmain ldap/servers/slapd/connection.c:1805
    #9 0x7f5c70f39412 in _pt_root (/lib64/libnspr4.so+0x2c412)
[...]

PR #5725, opened 2023-04-12
Issue 5646 - Various memory leaks

jchapma:
Bug description: A memory leak occurs when a sync repl search is run in refreshPersist mode. The connection from sync repl consumer is closed without freeing up the ldap req ctrls.

Fix description: When the connection to the client is closed or on shutdown free the request control structure if it exists.

relates: https://github.com/389ds/389-ds-base/issues/5646

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -925,12 +925,12 @@ sync_send_results(void *arg)
     int conn_acq_flag = 0;
     Slapi_Connection *conn = NULL;
     Slapi_Operation *op = req->req_orig_op;
+    LDAPControl **ctrls = NULL;
     int rc;
     PRUint64 connid;
     int opid;
     char **attrs_dup;
     char *strFilter;
-    LDAPControl **ctrls = NULL;
 
     slapi_pblock_get(req->req_pblock, SLAPI_CONN_ID, &connid);
     slapi_pblock_get(req->req_pblock, SLAPI_OPERATION_ID, &opid);
@@ -1073,7 +1073,7 @@ done:
     slapi_ch_free((void **)&strFilter);
     slapi_pblock_set(req->req_pblock, SLAPI_SEARCH_STRFILTER, NULL);
 
-    slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);
+    slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, &ctrls);
     if (ctrls) {
         ldap_controls_free(ctrls);
         slapi_pblock_set(req->req_pblock, SLAPI_REQCONTROLS, NULL);
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -930,6 +930,7 @@ sync_send_results(void *arg)
     int opid;
     char **attrs_dup;
     char *strFilter;
+    LDAPControl **ctrls = NULL;
 
     slapi_pblock_get(req->req_pblock, SLAPI_CONN_ID, &connid);
     slapi_pblock_get(req->req_pblock, SLAPI_OPERATION_ID, &opid);
@@ -1071,6 +1072,11 @@ done:
     slapi_pblock_get(req->req_pblock, SLAPI_SEARCH_STRFILTER, &strFilter);
     slapi_ch_free((void **)&strFilter);
     slapi_pblock_set(req->req_pblock, SLAPI_SEARCH_STRFILTER, NULL);
+
+    slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);
+    if (ctrls) {
+        ldap_controls_free(ctrls);
+    }
 
     slapi_pblock_destroy(req->req_pblock);
     req->req_pblock = NULL;
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -1076,6 +1076,7 @@ done:
     slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);
     if (ctrls) {
         ldap_controls_free(ctrls);
+        slapi_pblock_set(req->req_pblock, SLAPI_REQCONTROLS, NULL);
     }
 
     slapi_pblock_destroy(req->req_pblock);
--- a/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
+++ b/dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py
@@ -231,7 +231,7 @@ class Sync_persist(threading.Thread, ReconnectLDAPObject, SyncreplConsumer):
             print('syncrepl_poll: LDAP error (%s)', e)
         self.result = ldap_connection.get_cookies()
         log.info("ZZZ result = %s" % self.result)
-        self.conn.unbind()
+        ldap_connection.unbind()
 
 def test_sync_repl_mep(topology, request):
     """Test sync repl with MEP plugin that triggers several
@@ -404,12 +404,10 @@ def test_sync_repl_cookie_add_del(topology, init_sync_repl_plugins, request):
       6.: succeeds
     """
     inst = topology[0]
-
     # create a sync repl client and wait 5 seconds to be sure it is running
     sync_repl = Sync_persist(inst)
     sync_repl.start()
     time.sleep(5)
-
     # create users, that automember/memberof will generate nested updates
     users = UserAccounts(inst, DEFAULT_SUFFIX)
     users_set = []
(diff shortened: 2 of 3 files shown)


progier389 (approved):
LGTM

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -1072,6 +1073,11 @@ sync_send_results(void *arg)
     slapi_ch_free((void **)&strFilter);
     slapi_pblock_set(req->req_pblock, SLAPI_SEARCH_STRFILTER, NULL);
 
+    slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);
+    if (ctrls) {
+        ldap_controls_free(ctrls);
```
droideck, line 1078: I think we should also set NULL there in the `if`:

```
slapi_pblock_set(req->req_pblock, SLAPI_REQCONTROLS, NULL);
```

droideck (changes requested):

ldap/servers/plugins/sync/sync_persist.c
jchapma: Hi @droideck 
The pblock is destroyed directly after this call so I felt there was no need to reset the request controls memory.

Firstyear: Better to set to NULL just incase a new use arises where this assumption no longer holds true.

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -1072,6 +1073,12 @@ sync_send_results(void *arg)
     slapi_ch_free((void **)&strFilter);
     slapi_pblock_set(req->req_pblock, SLAPI_SEARCH_STRFILTER, NULL);
 
+    slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);
```
tbordaz: Sorry for my late answer. There is something I do not understand here: where this control in req_block comes from ?

We have a client sync_persist request with a control (sync_persist). This control is in the original SLAPI_REQCONTROLS. I initially thought it was erroneously copied in sync_pblock_copy but it looks it is not. So the only step where it could be set is in slapi_send_ldap_search_entry/send_ldap_search_entry/sync_send_results but here it is set in SLAPI_SEARCH_CTRLS not in SLAPI_REQCONTROLS.

In short I fail to find where this control comes from. I think the initial control (sync) should not be copied in req_block because it is only useful to select if it is a refresh/persist and useless in sync_send_results. But anyway I do not see where it is copy
If the coming from the sent results (set in SLAPI_SEARCH_CTRLS), how can it be retrieved in SLAPI_REQCONTROLS

droideck (approved):
If the fix resolves the memory leak, - ack from me.
But I join Thierry's concern, and we should figure our why it fixes the issue (if it fixes it).
Thanks!

ldap/servers/plugins/sync/sync_persist.c
jchapma: > Sorry for my late answer. There is something I do not understand here: where this control in req_block comes from ?

Here is what I think happens.


**Allocation**
When a client binds with ldapv3 we check for ldap request controls: https://github.com/389ds/389-ds-base/blob/56939cc3017d9527a30d4816ed75c654b096c371/ldap/servers/slapd/bind.c#L227

If they exist we allocate memory for the req ctrl, iterate through the ber elements populating the req ctrls: https://github.com/389ds/389-ds-base/blob/56939cc3017d9527a30d4816ed75c654b096c371/ldap/servers/slapd/control.c#L259

Then set the req ctrl in the pblock: https://github.com/389ds/389-ds-base/blob/56939cc3017d9527a30d4816ed75c654b096c371/ldap/servers/slapd/control.c#L357

**Use**
The sync repl plugin registers a pre search function: https://github.com/389ds/389-ds-base/blob/56939cc3017d9527a30d4816ed75c654b096c371/ldap/servers/plugins/sync/sync_init.c#L125

If a client performs a syn repl search in refreshPersist mode then a thread is kicked off to send results back to the client:
https://github.com/389ds/389-ds-base/blob/56939cc3017d9527a30d4816ed75c654b096c371/ldap/servers/plugins/sync/sync_refresh.c#L116

**Leak**
The thread function (sync_send_results) does its magic and terminates. But when the sync repl client closes the connection the req ctrls memory is never freed up. Hence the mem leak highlighted by asan.

tbordaz: Do you have the backtrace where the leaking control is allocated ?

I am surprise that we could retrieve in sync SRCH a control allocated during the BIND. I feel like only SRCH control can be stored in the req_pblock.

jchapma:
```
Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c708ba097 in calloc (/lib64/libasan.so.8+0xba097)
    #1 0x7f5c7006ac25 in slapi_ch_calloc (/usr/lib64/dirsrv/libslapd.so.0+0x26ac25)
    #2 0x7f5c7006b7a1 in get_ldapmessage_controls_ext (/usr/lib64/dirsrv/libslapd.so.0+0x26b7a1)
    #3 0x55639e2aadee in do_search ldap/servers/slapd/search.c:349
    #4 0x55639e2813fe in connection_dispatch_operation ldap/servers/slapd/connection.c:677
    #5 0x55639e2813fe in connection_threadmain ldap/servers/slapd/connection.c:1805
    #6 0x7f5c70f39412 in _pt_root (/lib64/libnspr4.so+0x2c412)
    #7 0x7f5c6fa8cdec in start_thread (/lib64/libc.so.6+0x8cdec)
    #8 0x7f5c6fb1236f in clone3 (/lib64/libc.so.6+0x11236f)


Indirect leak of 25 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c708ba6af in __interceptor_malloc (/lib64/libasan.so.8+0xba6af)
    #1 0x7f5c70fc6ed4 in ber_memalloc_x (/lib64/liblber.so.2+0x6ed4)
    #2 0x7f5c70fc6f98 in ber_get_stringbv (/lib64/liblber.so.2+0x6f98)
    #3 0x7f5c70fc734d in ber_get_stringa (/lib64/liblber.so.2+0x734d)
    #4 0x7f5c70fc9820 in ber_scanf (/lib64/liblber.so.2+0x9820)
    #5 0x7f5c7006b813 in get_ldapmessage_controls_ext (/usr/lib64/dirsrv/libslapd.so.0+0x26b813)
    #6 0x55639e2aadee in do_search ldap/servers/slapd/search.c:349
    #7 0x55639e2813fe in connection_dispatch_operation ldap/servers/slapd/connection.c:677
    #8 0x55639e2813fe in connection_threadmain ldap/servers/slapd/connection.c:1805
    #9 0x7f5c70f39412 in _pt_root (/lib64/libnspr4.so+0x2c412)
    #10 0x7f5c6fa8cdec in start_thread (/lib64/libc.so.6+0x8cdec)
    #11 0x7f5c6fb1236f in clone3 (/lib64/libc.so.6+0x11236f)
```

tbordaz: Excellent.
so this is the control within the original SRCH req that is set in SLAPI_REQCONTROLS. My understanding it is the sync control (along with the type of sync refresh/persist and the cookie). I think this control should not go into req_pblock that is used to send back result and do check the control. I wonder if the control is not erroneously moved to req_pblock for example during sync_persist_add.

jchapma: I ran the CI test with asan before and after this change, it does remove the memory leak.

tbordaz: So the original operation (where the controls are) is in req->req_orig_op. I fail to understand how it got referenced in req_pblock but clearly it is as your fix fixes the leak.
In normal processing this is operation_done that clear the control but here I doubt we can use it.

The last thing I miss is that instead of 'slapi_pblock_get(req->req_pblock, SLAPI_REQCONTROLS, ctrls);' I would expect &ctrls.

Except that the fix looks valid. ACK (and sorry for the noise)

tbordaz (approved):

jchapma:
@tbordaz 

Apologies for taking so long getting back to you, I completely forgot about this issue. Also, I am eating a huge piece of humble pie, you were right and I was wrong.

In the case of a sync repl search, run in refreshPersist mode, the request controls are not passed to the search results thread at all, I had "assumed" the req_ctrls were copied into the req in [sync_pblock_copy()](https://github.com/389ds/389-ds-base/blob/7c90259f4ff2bbc6bf579dca1cc52cf6ffa41bf8/ldap/servers/plugins/sync/sync_persist.c#L642), but they were not.

What really confuses me is when we free the req_ctrls memory in the results thread ASAN doesn't report a leak, even though there has been no memory allocated ??

So, the options are to copy the req_ctrls into the req that the search thread uses and free it in the thread or free it in 
[sync_persist_add()](https://github.com/389ds/389-ds-base/blob/7c90259f4ff2bbc6bf579dca1cc52cf6ffa41bf8/ldap/servers/plugins/sync/sync_persist.c#L647)

WDYT ?

Issue #5726, opened 2023-04-12
ns-slapd crashing in ldbm_back_upgradednformat


tbordaz:
**Issue Description**
During backend redesign [phase 1](https://pagure.io/389-ds-base/issue/49476) ldbm functions (archive2db, ldif2db,... upgradednformat) were modified to isolate BDB dependencies. In case of online call, the initialization of the ldbminfo structure needs to be done by ldbm functions. The initialization (call of dblayer_setup) is missing in upgradednformat so it crashes the server when accessing the ldbminfo structure

**Package Version and Platform:**
all version since 1.4.2

**Steps to Reproduce**
launch
upgradednformat -N -n userroot  -a /var/lib/dirsrv/slapd-<INSTANCE>/db/userroot -Z <INSTANCE>

**Expected results**
It should not crash

tbordaz:
56939cc30..6ae14bc8f  main
85ab1bc46..f0c804b0a  389-ds-base-2.3
d38c1667a..c13fe3d9e  389-ds-base-2.2
b6ea53f1c..9f896ea73  389-ds-base-2.1
9a1f2b758..269499440  389-ds-base-2.0
e882746f2..21ef813f2  389-ds-base-1.4.4
4b79bc350..528227d65  389-ds-base-1.4.3

PR #5727, opened 2023-04-12
Issue 5726 - ns-slapd crashing in ldbm_back_upgradednformat

tbordaz:
Bug description:
	With LDBM / BDB separation, LDBM functions like
	upgradednformat need to initialize ldbminfo

Fix description:
	call dblayer_setup in upgradednformat

relates: #5726

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldif2ldbm.c
+++ b/ldap/servers/slapd/back-ldbm/ldif2ldbm.c
@@ -360,7 +360,17 @@ int
 ldbm_back_upgradednformat(Slapi_PBlock *pb)
 {
     struct ldbminfo *li = NULL;
+    int task_flags;
     slapi_pblock_get(pb, SLAPI_PLUGIN_PRIVATE, &li);
+    slapi_pblock_get(pb, SLAPI_TASK_FLAGS, &task_flags);
+
+    if (task_flags & SLAPI_TASK_RUNNING_FROM_COMMANDLINE) {
+        if (dblayer_setup(li)) {
+            slapi_log_err(SLAPI_LOG_CRIT, "ldbm_back_upgradednformat", "dblayer_setup failed\n");
+            return -1;
+        }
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/back-ldbm/ldif2ldbm.c
```diff
@@ -360,7 +360,14 @@ int
 ldbm_back_upgradednformat(Slapi_PBlock *pb)
 {
     struct ldbminfo *li = NULL;
+    int task_flags;
     slapi_pblock_get(pb, SLAPI_PLUGIN_PRIVATE, &li);
+    slapi_pblock_get(pb, SLAPI_TASK_FLAGS, &task_flags);
+
+    if (task_flags & SLAPI_TASK_RUNNING_FROM_COMMANDLINE) {
+        dblayer_setup(li);
```
droideck: I wonder if it is worth putting a check for the return code here so that if it fails, it won't fail silently.
Something like:

```
        if (dblayer_setup(li)) {
            slapi_log_err(SLAPI_LOG_CRIT, "ldbm_back_upgradednformat", "dblayer_setup failed\n");
            return -1;
        }
```

droideck (changes requested):
The test is not run in CI because it doesn't have '_test' in its name.
Hence, please, rename it to something like `upgradednformat_test.py`

ldap/servers/slapd/back-ldbm/ldif2ldbm.c
tbordaz: absolutely right. Thanks

droideck (approved):
Looks great! Ack.

P.S. For the sake of transparency, could please you also mention in the commit message that you've updated [src/Cargo.lock](https://github.com/389ds/389-ds-base/pull/5727/files#diff-9ab3c1002aa8dc4cb786416ce598f1ffe63a2a341ffe764b07dc1b653a47dd9d)? Or you can discard the changes from the PR as it's not related to the change.

Issue #5156, opened 2022-02-07
RFE create a slapi-private function that returns the groups that a target entry is member of


tbordaz:
**Is your feature request related to a problem? Please describe.**
It exists in memberof plugin, a function static **memberof_get_groups** that retrieves the direct and indirect groups that a given entry is memberof. In addition this function updates the target entry with the **memberof** attribute.

This function could be reused in others plugins and could be provided as slapi-private function

**Describe the solution you'd like**
TBD

tbordaz:
This ticket needs take into account requirements of others plugins/service: memberof plugin, referential integrity plugin, acl code has a group cache, with ticket 4667 also replication will deal with groups of binddns (see https://github.com/389ds/389-ds-base/issues/1097)

tbordaz:
The design of this RFE is [here](https://www.port389.org/docs/389ds/design/slapi_memberof.html)

tbordaz:
RFE was pushed upstream with tests that recompute memberof when slapi_memberof is called
dd7d487ab..9a05477f6 main

tbordaz:
18ef874d8..881cade75 main (build breakage)
2d1e14555..ea0ed1f51  main (build warning)

tbordaz:
RFE available upstream

tbordaz:
990455004..eaabdd2a3 main tests slapi_memberof reusing memberof values

tbordaz:
f2f6c2b9e..642760b73  389-ds-base-2.3

PR #5745, opened 2023-04-27
Issue 5156 - fix build breakage from slapi-memberof commit

mreynolds389:
Description:  Function prototypes were not declared correctly and this breaks the builds on new compilers.

relates: #5156

### DIFF
--- a/ldap/servers/slapd/slapi-private.h
+++ b/ldap/servers/slapd/slapi-private.h
@@ -1260,8 +1260,8 @@ void dup_ldif_line(struct berval *copy, const char *line, const char *endline);
 
 /* slapi-memberof.c */
 int slapi_memberof(Slapi_MemberOfConfig *config, Slapi_DN *member_sdn, Slapi_MemberOfResult *result);
-void slapi_memberof_free_memberof_plugin_config();
-int slapi_memberof_load_memberof_plugin_config();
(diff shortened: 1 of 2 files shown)


droideck (approved):
LGTM! Besides the minor issue.

ldap/servers/slapd/slapi-private.h
```diff
@@ -1260,8 +1260,8 @@ void dup_ldif_line(struct berval *copy, const char *line, const char *endline);
 
 /* slapi-memberof.c */
 int slapi_memberof(Slapi_MemberOfConfig *config, Slapi_DN *member_sdn, Slapi_MemberOfResult *result);
-void slapi_memberof_free_memberof_plugin_config();
-int slapi_memberof_load_memberof_plugin_config();
+void slapi_memberof_free_memberof_plugin_config(void);
```
droideck, line 1263: Should we also change the signature here?
https://github.com/389ds/389-ds-base/blob/18ef874d839036e5f025acf511c03dbc57ccef70/ldap/servers/slapd/slapi-memberof.c#L583
(and the same for other function)

mreynolds389: you only need "void" set in the header. I think we get an error if you set it in the c file

PR #5748, opened 2023-05-01
Issue 152 - RFE - Add support for LDAP alias entries

mreynolds389:
Description:  

Per RFC rfc4512#section-2.6 add support for Alias Entries.
 Currently this is only designed to work with "base" searches.

Thanks for @anilech for the code contribution!!!

relates: https://github.com/389ds/389-ds-base/issues/152

### DIFF
--- a/dirsrvtests/tests/suites/plugins/alias_entries_test.py
+++ b/dirsrvtests/tests/suites/plugins/alias_entries_test.py
@@ -1,0 +1,111 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2023 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+import ldap
+import time
+from lib389.utils import ensure_str
+from lib389.plugins import AliasEntriesPlugin
+from lib389.idm.user import UserAccount
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.topologies import topology_st as topo
+
+log = logging.getLogger(__name__)
+
+TEST_ENTRY_NAME = "entry"
+TEST_ENTRY_DN = "cn=entry," + DEFAULT_SUFFIX
+TEST_ALIAS_NAME = "alias entry"
+TEST_ALIAS_DN = "cn=alias_entry," + DEFAULT_SUFFIX
+TEST_ALIAS_DN_WRONG = "cn=alias_entry_not_there," + DEFAULT_SUFFIX
+EXPECTED_UIDNUM = "1000"
+
+
+def test_entry_alias(topo):
+    """Test that deref search for alias entry works
+
+    :id: 454e85af-0e20-4a36-9b3a-02562b1db53d
+    :setup: Standalone Instance
+    :steps:
+        1. Enable alias entry plugin
+        2. Create entry and alias entry
+        3. Set deref option and do a base search
+        4. Test non-base scope ssearch returns error
+        5. Test invalid alias DN returns error
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+    """
+    inst = topo.standalone
+
+    # Enable Alias Entries plugin
+    alias_plugin = AliasEntriesPlugin(inst)
+    alias_plugin.enable()
+    inst.restart()
+
+    # Add entry
+    test_user = UserAccount(inst, TEST_ENTRY_DN)
+    test_user.create(properties={
+        'uid': TEST_ENTRY_NAME,
+        'cn': TEST_ENTRY_NAME,
+        'sn': TEST_ENTRY_NAME,
+        'userPassword': TEST_ENTRY_NAME,
+        'uidNumber': '1000',
+        'gidNumber': '2000',
+        'homeDirectory': '/home/alias_test',
+    })
+
+    # Add entry that has an alias set to the first entry
+    test_alias = UserAccount(inst, TEST_ALIAS_DN)
+    test_alias.create(properties={
+        'uid': TEST_ALIAS_NAME,
+        'cn': TEST_ALIAS_NAME,
+        'sn': TEST_ALIAS_NAME,
+        'userPassword': TEST_ALIAS_NAME,
+        'uidNumber': '1001',
+        'gidNumber': '2001',
+        'homeDirectory': '/home/alias_test',
+        'objectclass': ['alias', 'extensibleObject'],
+        'aliasedObjectName': TEST_ENTRY_DN,
+    })
+
+    # Set the deref "finding" option
+    inst.set_option(ldap.OPT_DEREF, ldap.DEREF_FINDING)
+
+    # Do base search which could map entry to the aliased one
+    log.info("Test alias")
+    deref_user = UserAccount(inst, TEST_ALIAS_DN)
+    result = deref_user.search(scope="base")
+    assert result[0].dn == TEST_ENTRY_DN
+    assert ensure_str(result[0].getValue('uidNumber')) == EXPECTED_UIDNUM
+
+    # Do non-base search which could raise an error
+    log.info("Test unsupported search scope")
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        deref_user.search(scope="subtree")
+
+    # Reset the aliasObjectname to a DN that does not exist, and try again
+    log.info("Test invalid alias")
+    test_alias.replace('aliasedObjectName', TEST_ALIAS_DN_WRONG)
+    try:
+        deref_user.search(scope="base")
+        assert False
+    except ldap.LDAPError as e:
+        msg = e.args[0]['info']
+        assert msg.startswith("Failed to dereference alias object")
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
--- a/ldap/servers/plugins/alias_entries/alias-entries.c
+++ b/ldap/servers/plugins/alias_entries/alias-entries.c
@@ -1,0 +1,145 @@
+/** BEGIN COPYRIGHT BLOCK
+ * Copyright (C) 2023 anilech
+ * Copyright (C) 2023 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * License: GPL (version 3 or any later version).
+ * See LICENSE for details.
+ * END COPYRIGHT BLOCK **/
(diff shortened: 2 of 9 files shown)


ldap/servers/plugins/alias_entries/alias-entries.c
```diff
@@ -0,0 +1,123 @@
...
+    /* deref must be enabled */
+    if (slapi_pblock_get(pb, SLAPI_SEARCH_DEREF, &deref_enabled) != 0) {
+        return 0;
+    }
+    if (deref_enabled != LDAP_DEREF_FINDING && deref_enabled != LDAP_DEREF_ALWAYS) {
+        return 0;
+    }
+
+    /* Only base search is supported */
+    if (slapi_pblock_get(pb, SLAPI_SEARCH_SCOPE, &scope) != 0 || scope != LDAP_SCOPE_BASE) {
```
tbordaz: We could log a warning message that the dereferenceAlias indicator is not supported for sub/one level scope.

ldap/servers/plugins/alias_entries/alias-entries.c
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /* Base dn should be provided */
+    if (slapi_pblock_get(pb, SLAPI_SEARCH_TARGET_SDN, &search_target) != 0 || search_target == NULL) {
+        return 0;
+    }
+
+    /* Skip our own requests */
+    if (slapi_pblock_get(pb, SLAPI_SEARCH_STRFILTER, &str_filter) != 0 ||
+        strcmp(str_filter, ALIASFILTER) == 0)
```
tbordaz: I am unsure that 'strfilter' is always set. Could we test 'str_filter' before strcmp.

ldap/servers/plugins/alias_entries/alias-entries.c
```diff
@@ -0,0 +1,123 @@
...
+        return 0;
+    }
+
+    /* Follow the alias chain */
+    dn2 = search_target;
+    do {
+        dn1 = dn2;
+        dn2 = alias_get_next_dn(dn1);
+        if (i > 0 && dn2 != NULL) {
+            slapi_sdn_free(&dn1);
```
tbordaz, line 114: I think we may have double free, if E1->E2. A search with base=E1 frees the 'dn1', but the sdn is still referenced in 'search_target' and pblock.

ldap/servers/plugins/alias_entries/alias-entries.c
```diff
@@ -0,0 +1,123 @@
...
+    } while (dn2 != NULL && i++ < MAXALIASCHAIN);
+
+    if (dn1 == search_target) {
+        /* Source dn is not an alias */
+        return 0;
+    }
+
+    if (dn2 == NULL) {
+        /* alias resolved, set new base */
+        slapi_sdn_free(&search_target);
```
tbordaz, line 137: Here we free search_target that was already freed as 'dn1'

ldap/servers/plugins/alias_entries/alias-entries.c
```diff
@@ -0,0 +1,123 @@
...
+        slapi_pblock_set(pb, SLAPI_SEARCH_TARGET_SDN, dn1);
```
tbordaz, line 138: That looks fine as the loop should have freed the target_sdn.

ldap/servers/plugins/alias_entries/alias-entries.c
progier389: @tbordaz, not sure what you were thinking exactly, but I do not think logging a message in error log is a good idea. (or at least not something logged at the default log level) because of the risk of flooding the log file. Adding a note in the access log result, or in the ldap result message may be useful.

ldap/servers/plugins/alias_entries/alias-entries.c
progier389: @tbordaz: No because of i>0  test: (nothing is freed at first iteration)

ldap/servers/plugins/alias_entries/alias-entries.c
progier389: No, search_target is not already freed.

progier389 (approved):
LGTM, except that I agree with thierry about testing that str_filter is not NULL

ldap/servers/plugins/alias_entries/alias-entries.c
tbordaz: I agree there is a risk of flooding error logs, unless it is logged only once. Ideally the ldap_result would be the most helpful for client but the plugin does not return result.

ldap/servers/plugins/alias_entries/alias-entries.c
tbordaz: ah nice !
Sorry for the noise

ldap/servers/plugins/alias_entries/alias-entries.c
tbordaz: Yes that is correct, the search_target was not freed.

tbordaz (approved):
after a second reading, the patch looks good.

dirsrvtests/tests/suites/plugins/alias_entries_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_entry_alias(topo):
+    """Test that deref search for alias entry works
+
+    :id: 454e85af-0e20-4a36-9b3a-02562b1db53d
+    :setup: Standalone Instance
+    :steps:
+        1. Enable alias entry plugin
+        2. Create entry and alias entry
+        2. Set deref option and do a base search
```
tbordaz: numbering is incorrect.

mreynolds389: > numbering is incorrect.

I'll fix this, and I did run this PR through ASAN and did not find any issues.

mreynolds389:
> LGTM, except that I agree with thierry about testing that str_filter is not NULL

Yup I'll work on that...

ldap/servers/plugins/alias_entries/alias-entries.c
progier389: But the search operation does, and the plugin could set the SLAPI_PB_RESULT_TEXT in pblock ...

ldap/servers/plugins/alias_entries/alias-entries.c
droideck: I have a doubt here.

If the function alias_get_next_dn returns NULL, dn1 is not freed. And alias_get_next_dn can return NULL if (slapi_search_internal_pb(pb) != 0).

droideck: So I wonder if we should change it to something like that:

```
    if (i > 0 && dn1 != search_target) {
        slapi_sdn_free(&dn1);
    }
```

So dn1 is always freed unless it is the original search_target.

progier389: No the code is good: We cannot not free dn1 as it is used afterward and stored in the search pblock.
  and checking for search_target is redundant with i>0 test ( because dn1 ==  search_target <==> i == 0 )
      (I agree that checking for dn1 != search_target instead of i > 0 would be clearer but code is OK. )
  dn1 is the current dereferenced target (the loop handles the case of an alias pointing on another alias case)
  so if there is a next item (i.e dn2), we should free the previous dn1 before overwriting it in next iteration.
  And the loop aborted case (because there are too many hops) is fine too.
    dn1 has been freed if not search_target by the (i > 0 && dn2 != NULL) and dn2 is freed 
     in the  dn2 == NULL ...  else code

droideck: My main concern is that the  `slapi_search_internal_pb(pb)` call from `alias_get_next_dn` function  may fail (for any unexpected reason).
The way `alias_get_next_dn` function is written, it will return NULL.

If it'll happen in the middle of the do-while loop here, we won't free `dn1` and it won't be `search_target` either.
So what will happen next - we'll assign `dn1` to `SLAPI_SEARCH_TARGET_SDN` and, unless I miss something, it won't be right.

But yeah, my solution is not complete for this case...

ldap/servers/plugins/alias_entries/alias-entries.c
mreynolds389: > I am unsure that 'strfilter' is always set. Could we test 'str_filter' before strcmp.

Fixed

ldap/servers/plugins/alias_entries/alias-entries.c
mreynolds389: True, if the internal search failed (e.g. nonexistent DN in aliasedObjectName) then the parent in the chain would be returned instead.  So this can be hardened for a such a case.  

Do we log an warning?  
Do we return an error to the client?
Do we just return the parent?
Or return no result at all (err=0 nentries=0)?

Probably should just return an error to client: operations error?  Unwilling to perform?

droideck: My vote is for the operations error. I think we should notify the user that something is wrong and point out that dereference is broken in his requested scope.

If the admin decided to use the plugin - probably there's a good reason for that and a warning, parent's return or just no results is not enough in my opinion.

@mreynolds389 @tbordaz @progier389 thoughts? :)

mreynolds389: The RFC does not provide any requirement for such a scenario.  I'm torn between not adjusting the search target at all, or returning an error (err=1 works for me).

progier389: We can argue about it, but IMHO its depends of the failure cause.
  if it is NOSUCHOBJECT the current code behavior is not so bad: returning the last entry that we were able to dereference. Anyway by looking at its objectclass, the client application should easily determine that it is an alias and that something is wrong. (returning the original entry is possible too)
  If it is another error, I would return the internal search error code and a result text like: "'%s' while dereferencing alias %s" internal_search_result_text, dn1

mreynolds389: Ok, since this a preop plugin I can not actually cause the operation to fail, but I can attach a result text.  I also can not get python-ldap to detect this result text in the CI, but I can confirm with ldapsearch that it is there.

ldap/servers/plugins/alias_entries/alias-entries.c
mreynolds389: > We could log a warning message that the dereferenceAlias indicator is not supported for sub/one level scope.

I decided to make it a full blown error if you don't use a base-level search.  Any concerns here?

Please review...

progier389: Looks good !

tbordaz: Looks good to me as well

droideck (approved):
LGTM!

Issue #5749, opened 2023-05-02
RFE - allow account policy plugin to handle both inactivity and password expiration


mreynolds389:
**Issue Description**

https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/administration_guide/account-policy-plugin

This describes how you can use inactivity **or** account expiration, but not both at the same time.  This RFE is allow both scenarios to be handled instead of just one or the other.

mreynolds389:
Downstream bug:

https://bugzilla.redhat.com/show_bug.cgi?id=2174161

mreynolds389:
Design doc:

https://www.port389.org/docs/389ds/design/account-policy-inactivity-and-expiration-design.html

PR #5750, opened 2023-05-02
Issue 5749 - RFE - Allow Account Policy Plugin to handle inactivity and expiration at the same time

mreynolds389:
Description:

Currently Account Policy Plugin as a state attribute and alternate state attribute. If the main state attribute is NOT present in the entry then it fails back to the alternate state attribute.

This RFE adds a new setting that tells the plugin to check both state attributes. The purpose of this is for expiration and inactivity, so this is meant to be used when the alternate state attribute is 'passwordExpirationtime'.  So if the main state attribute is OK, it will then check the alternate state attribute for inactivity.

relates: https://github.com/389ds/389-ds-base/issues/5749

### DIFF
--- a/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
+++ b/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
@@ -1,0 +1,119 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2023 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import ldap
+import logging
+import pytest
+import os
+import time
+from lib389.topologies import topology_st as topo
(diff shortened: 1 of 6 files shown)


tbordaz (approved):
Nice. LGTM

dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    :id: 704310de-a2eb-4ee7-baf3-9770c0fbf07c
+    :setup: Standalone Instance
+    :steps:
+        1. Fill in test case steps here
+        2. And indent them like this (RST format requirement)
+    :expectedresults:
+        1. Fill in the result that is expected
+        2. For each test step
+    """
```
droideck: Could be improved :)

droideck (changes requested):
When I try to create an AccPol config entry, the browser crushes for me:

image

dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
mreynolds389: Fixed

mreynolds389:
> When I try to create an AccPol config entry, the browser crushes for me

All fixed, please review.

droideck (approved):
Looks great! Ack

PR #5762, opened 2023-05-11
Issue 3527 - Add PROXY protocol support

### LINKED ISSUE #3527: [RFE] Add PROXY protocol support to 389-ds-base via confiuration item - similar to Postfix.
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/50470
- Created at 2019-06-25 16:16:55 by [mreynolds](https://pagure.io/user/mreynolds) (@mreynolds389)
- Assigned to [mhonek](https://pagure.io/user/mhonek) (@kenoh)
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=1382123

---

Ticket was cloned from Red Hat Bugzilla: [Bug 1382123](https://bugzilla.redhat.com/show_bug.cgi?id=1382123)

_Please note that this Bug is private and may not be accessible as it contains confidential Red Hat customer information._

```text
Description of problem:

Please add support to 389-base for the PROXY protocol for ACI evaluation and
also for logging client queries. The proxy protocol is described here:

http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt

Background:
As a network engineer, I can say that having a load balancer in path in your
network is a bad idea. It is bad because it becomes part of the network and it
becomes the weakest link. It limits the capacity of the network and becomes
additional points of failure in the network. The ideal place for a load
balancer is on the side, with the client traffic being network address
translated to address ranges from SNAT pools, where the server recieving the
traffic never directly sees the IP address of the client.

Loadbalancing out of path traffic to a group of ldap servers presents a
semi-unique problem when ACIs must be evaluated against client IP address and
also for client logging. The PROXY protocol provides provides this information
to the backend servers via an additional TCP header so that the ACIs can be
correctly evaluated and client traffic can be logged.

A great example of non-http software that is capable of using the additional
tcp header is the Postfix MTA. There is an announcement here:

http://permalink.gmane.org/gmane.comp.web.haproxy/8881

Thank you for your consideration.
```

droideck:
Description: Add support to 389-base for the PROXY protocol for ACI evaluation and also for logging client queries.

The proxy protocol is described here:
http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt

Fixes: https://github.com/389ds/389-ds-base/issues/3527

Reviewed by: ?


droideck:
Unit tests pass, and Asan is clean:
```
cmocka_unit_test(test_libslapd_haproxy_v1),
cmocka_unit_test(test_libslapd_haproxy_v2_valid),
cmocka_unit_test(test_libslapd_haproxy_v2_valid_local),
cmocka_unit_test(test_libslapd_haproxy_v2_invalid),
```

But at the last moment, I had some issues with DS connection management, investigating...

In the meantime, you can go ahead and review haproxy.c parsing part as it's complete.
The design is on review here:
https://github.com/389ds/389wiki/pull/89/files

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,376 @@
...
+ * @return: Returns 0 on successful operation, -1 on error.
+ */
+int haproxy_receive(int fd, int *proxy_connection, PRNetAddr *pr_netaddr_from, PRNetAddr *pr_netaddr_dest)
+{
+    /* Buffer to store the header received from the HAProxy server */
+    char hdr[HAPROXY_HEADER_MAX_LEN + 1];
+    size_t hdr_len;
+
+    /* Attempt to receive the header from the HAProxy server */
+    if ((hdr_len = recv(fd, hdr, sizeof(hdr) - 1, MSG_PEEK)) <= 0) {
```
Firstyear: I think there is a denial of service here. Since this is done in the accept thread, if a provider was to send an initial data of 1 byte, this is smaller than size of hdr. recv will block until len bytes are available, and since this is in the accept thread then this will cause the accept thread to stall completely and not timeout. 

As a result I think two important changes are needed.

First, the reading of this data can't and shouldn't be in the accept thread - the accept thread has to be as light as possible, and has to accept quickly, so this will potentially add latency and performance issues.

Second, that the reading of this proxy header must occur in a way that a short message won't stall the accept thread, or that a timeout occurs.

Firstyear (changes requested):

ldap/servers/slapd/haproxy.c
droideck: Sure, it totally makes sense (I was struggling for some time trying to figure out the best place to put it, as I'm pretty new to our connection code).

So,,, Will this be the better place, then?
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/connection.c#L284

connection_reset call, right before we get the client's address. And if it's a proxy connection, we can skip this altogether:
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/connection.c#L285-L346

I'm still trying to figure out what to do with Destination IP address from HAproxy when it's present. Maybe log it to CONNS log level just for diagnosis purposes?

droideck: @Firstyear okay, I've moved the `haproxy_recieve` call inside of `connection_reset`. The reason is that we need to set the client address correctly, and  `connection_reset` already logs the IP to the access log (so it'll be confusing to parse it later as the proxy IP would not be a correct IP printed in this case).

> First, the reading of this data can't and shouldn't be in the accept thread - the accept thread has to be as light as possible, and has to accept quickly, so this will potentially add latency and performance issues.

If I understand correctly, `connection_reset` is still in the `accept_thread`. What I can think of (if you think it will make it considerably slow) - we can add a config setting - something like `nsslapd-parse-haproxy: [on, off]` so the admin will use it only when they have HAProxy in their topology. This way, it will harm the performance only when they have HAProxy (but they will win in other ways). What do you think?

> Second, that the reading of this proxy header must occur in a way that a short message won't stall the accept thread, or that a timeout occurs.

`recv` won't wait on 1 byte as per specs. But it will block on 0 bytes. I've added a 'nonblocking switch' to that area. Do you think it will work for our case?

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,452 @@
...
+    pr_netaddr->inet.family = PR_AF_INET;
+    pr_netaddr->inet.port = in_port;
+    pr_netaddr->inet.ip = in_addr;
+
+    /* Print the address in a human-readable format */
+    char addr[INET_ADDRSTRLEN];
+    if (inet_ntop(AF_INET, &(pr_netaddr->inet.ip), addr, INET_ADDRSTRLEN) == NULL) {
+        slapi_log_err(SLAPI_LOG_CONNS, "haproxy_parse_v2_addr_v4", "Failed to print address.\n");
+    } else {
+        slapi_log_err(SLAPI_LOG_TRACE, "haproxy_parse_v2_addr_v4", "Address: %s\n", addr);
```
mreynolds389: Could this also be SLAPI_LOG_CONNS?  TRACE is just bloated and almost unusable

ldap/servers/slapd/connection.c
```diff
@@ -271,12 +271,18 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
...
+    int proxy_connection = 0;
+    PRNetAddr pr_netaddr_dest = {{0}};
 
     slapi_log_err(SLAPI_LOG_CONNS, "connection_reset", "new %sconnection on %d\n", pTmp, conn->c_sd);
 
     /* bump our count of connections and update SNMP stats */
     conn->c_connid = slapi_counter_increment(num_conns);
 
+    if (haproxy_receive(conn->c_sd, &proxy_connection, from, &pr_netaddr_dest) != 0) {
+       slapi_log_err(SLAPI_LOG_CONNS, "connection_reset", "Failed to recieve HAProxy information\n");
```
mreynolds389: indentation :-)

mreynolds389:
Minor issues here...   While I know it's now OK to declare variables anywhere in a function, it would be nice to keep them at the top of the function, and all the "code" below it.  For example you have some trace logging calls (which should not be TRACE IMHO) at the top of the function then the variable declarations.  Also I see "*proxy_connection = 0;" at the top of some functions.  I just want to be consistent across the code base.

droideck:
Done, please review...

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,457 @@
...
+    size_t hdr_len;
+    int was_blocking;
+
+    if (set_nonblocking_if_needed(fd, &was_blocking) == -1) {
+        return -1;
+    }
+    /* Attempt to receive the header from the HAProxy server */
+    recv_result = recv(fd, hdr, sizeof(hdr) - 1, MSG_PEEK);
+
+    if (set_blocking_if_needed(fd, was_blocking) == -1) {
```
progier389: Is there a reason to put back into blocking mode at this point?
  If not, we should only do this at the very end of the function
    (I have the feeling that switch the mode back before parsing the header is useless (because unless mistaken there is no I/O performed in the parsing phase)

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,457 @@
...
+    /* Null-terminate the header string */
+    if (hdr_len < sizeof(hdr)) {
+        hdr[hdr_len] = 0;
+    } else {
+        slapi_log_err(SLAPI_LOG_CONNS, "haproxy_receive", "Recieved header is too long or an error is returned: %d\n", hdr_len);
+        return -1;
+    }
+
+    /* Try to parse the header as a version 1 header. If that fails, try as a version 2 header. */
+    if (haproxy_parse_v1_hdr(hdr, &hdr_len, proxy_connection, pr_netaddr_from, pr_netaddr_dest) != 0) {
```
progier389: IMHO something is wrong is the logic.
   Error cases should be handled differently than "not a header" cases:
     i.e the parsing function should return one of 3 values:
        HEADER_PARSED, NOT_AN_HEADER, ERROR
          (and should close connection after an error)
        - for v1:  if data does not start with PROXY it is NOT_AN_HEADER otherwise it is an error
        - for v2: if data len is too small or signature is invalid it is NOT_AN_HEADER otherwise it is an error

ldap/servers/slapd/connection.c
```diff
@@ -271,12 +271,18 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
...
 
     slapi_log_err(SLAPI_LOG_CONNS, "connection_reset", "new %sconnection on %d\n", pTmp, conn->c_sd);
 
     /* bump our count of connections and update SNMP stats */
     conn->c_connid = slapi_counter_increment(num_conns);
 
+    if (haproxy_receive(conn->c_sd, &proxy_connection, from, &pr_netaddr_dest) != 0) {
+        slapi_log_err(SLAPI_LOG_CONNS, "connection_reset", "Failed to recieve HAProxy information\n");
+    }
+
```
progier389: Several things are wrong here:
  - if haproxy_receive returns that an header is found.
      - IMPORTANT FOR SECURITY: Should check that the socket sender address (i.e the proxy host) is allowed to 
        send a proxy request (i.e  a new config attribute is needed)
      The pr_netaddr_dest (i.e the proxyed host) should be stored in the connection (so that aci/ host check - check the original host  rather than the proxy host )

progier389 (changes requested):

progier389:
A last point, we should check how this work for SSL/TLS (NSPR/NSS layers may requires to set some callbacks to get the proxied host instead of using the sender socket address ...

ldap/servers/slapd/connection.c
droideck: What kind of config attribute do you have in mind?
Do you mean a multivalued attribute for a list of trusted HAProxy servers' IP addresses?

If so, I was thinking this should be taken care of at the network level using a firewall, but I guess another level of security wouldn't hurt...

ldap/servers/slapd/connection.c
```diff
@@ -300,7 +306,8 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
             }
```
tbordaz: 'from' is craft by the client so it can contain wrong data. I wonder if in case of proxy_connection we should fail if "from->raw.family == PR_AF_LOCAL".

ldap/servers/slapd/connection.c
tbordaz: I do not know if this is the same concern as @progier389 said, but in case haproxy_receive can be broken by a craft message we should have a config mechanism to disable the call to haproxy_receive.

ldap/servers/slapd/connection.c
```diff
@@ -271,12 +271,18 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
...
     int in_referral_mode = config_check_referral_mode();
+    int proxy_connection = 0;
+    PRNetAddr pr_netaddr_dest = {{0}};
 
     slapi_log_err(SLAPI_LOG_CONNS, "connection_reset", "new %sconnection on %d\n", pTmp, conn->c_sd);
 
     /* bump our count of connections and update SNMP stats */
     conn->c_connid = slapi_counter_increment(num_conns);
 
+    if (haproxy_receive(conn->c_sd, &proxy_connection, from, &pr_netaddr_dest) != 0) {
```
tbordaz: Just for curiosity, what happens if LDAP client is not HAproxy ? A regular client sending a LDAP bind, can it break its first LDAP request ?

ldap/servers/slapd/connection.c
progier389: Yes I was thinking about a multivalued list of trusted IP addresses. I do not think that we can afford not to do an access control at this level: 
 Firewall cannot help the deployment that does not use HAProxy because in that case any client that have access to the server could then fake its address by crafting an HAProxy header
 
Furthermore the HAProxy reference document explicitly warn against that in its security consideration section:
> Implementers should be very careful about not trying to automatically detect
> whether they have to decode the header or not, but rather they must only rely
> on a configuration parameter. Indeed, if the opportunity is left to a normal
> client to use the protocol, he will be able to hide his activities or make them
> appear as coming from someone else. However, accepting the header only from a
> number of known sources should be safe.

 I do not think that we should  wholly disable the call to haproxy_receive as @tbordaz suggest I think that decoding the haproxy record and rejecting it if it is not allowed is more clear in term of diagnostic 
  (You could explicitly see in access log that there was an HAProxy attempt and that it was rejected)
 
 Another point: IMHO we should log a message in access log that the connection is proxyed (mentionning the proxied  address) (Because otherwise we may have problem with troubleshooting ... )
 I was thinking about:
  [date] conn=N HAPROXY version=1 new_address=.... OK/denied

ldap/servers/slapd/connection.c
progier389: Unless mistaken checking that incoming address is trusted should take care of that case without having to explicitly test the familly.

ldap/servers/slapd/haproxy.c
progier389: Thinking more about it: I think that resetting the mode at this point is ok, but we should not bother to switch again the mode around the second recv (We know that the data are present because we have already read them so the second recv will not block whatever the mode is)

ldap/servers/slapd/connection.c
progier389: About 
> Just for curiosity, what happens if LDAP client is not HAproxy ? A regular client sending a LDAP bind, can it break its first LDAP request ?
haproxy_receive function is quite straightforward:
    the first PDU is read with MSG_PEEK flag (so it stay in the socket queue)
      if it is not a haproxy v1 or an haproxy v2 that is the end of the function.
        (So the answer to @tbordaz question is that the first PDU is still unchanged at that point)
      if it is an haproxy v1 exactly header_len bytes are read from the socket with 0 flag (so the haproxy header is removed from the socket queue)

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,457 @@
...
+    char hdr[HAPROXY_HEADER_MAX_LEN + 1];
+    ssize_t recv_result = 0;
+    size_t hdr_len;
+    int was_blocking;
+
+    if (set_nonblocking_if_needed(fd, &was_blocking) == -1) {
+        return -1;
+    }
+    /* Attempt to receive the header from the HAProxy server */
+    recv_result = recv(fd, hdr, sizeof(hdr) - 1, MSG_PEEK);
```
jchapma: Could we check the socket for data using just 1 recv call rather than 2 ?

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,457 @@
...
+ */
+int haproxy_receive(int fd, int *proxy_connection, PRNetAddr *pr_netaddr_from, PRNetAddr *pr_netaddr_dest)
+{
+    /* Buffer to store the header received from the HAProxy server */
+    char hdr[HAPROXY_HEADER_MAX_LEN + 1];
+    ssize_t recv_result = 0;
+    size_t hdr_len;
+    int was_blocking;
+
+    if (set_nonblocking_if_needed(fd, &was_blocking) == -1) {
```
jchapma: Would using the MSG_DONTWAIT flag with recv mean that we dont care if the socket is blocking or not. I havent used it before but the man page says it enables a non blocking operation.

ldap/servers/slapd/connection.c
jchapma: Maybe a log message to the security logs also ?

ldap/servers/slapd/haproxy.c
Firstyear: Blocking in the accept thread isn't really something we can do, nor can we have the accept thread stuck waiting on the haproxy info. This must be in the client thread, so if that means we have to change how we log the IP address then that's a *good* thing. For example, it actually probably is good to log the connection IP as coming from the HAPROXY server, and then have a special log message that states "connection from HAPROXY from IP X:Y originates from client A:B"

ldap/servers/slapd/connection.c
Firstyear: I agreed with @progier389 here, we need a way to say that we accept and trust the HAPROXY info we are being sent. If that isn't set, then we should reject connections that attempt to tamper with this.

ldap/servers/slapd/haproxy.c
droideck: But that's exactly what I do, I unblock the socket so `recv` won't wait on anything and fails right away if there's 0 byte. 

Please, check haproxy.c and its `set_nonblocking_if_needed` and `set_blocking_if_needed` around `recv` call...

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,457 @@
...
+ * @param proxy_connection: A pointer to an integer to store the proxy connection status (0 or 1).
+ * @param pr_netaddr_from: A pointer to a PRNetAddr structure to store the source address info.
+ * @param pr_netaddr_dest: A pointer to a PRNetAddr structure to store the destination address info.
+ * 
+ * @return: Returns 0 on successful operation, -1 on error.
+ */
+int haproxy_receive(int fd, int *proxy_connection, PRNetAddr *pr_netaddr_from, PRNetAddr *pr_netaddr_dest)
+{
+    /* Buffer to store the header received from the HAProxy server */
+    char hdr[HAPROXY_HEADER_MAX_LEN + 1];
```
Firstyear: Needs to be zerod with `= {0}`

ldap/servers/slapd/haproxy.c
Firstyear: It's still happening in the accept thread though, which is kind of my point here that it will add latency to the accept thread - there are enough issues with people trying to improve the performance of this, so I'm worried that we'll end up some issues here. 

For example, imagine if for some reason a valid haproxy header was sent but then fragments over the recv call? We would then close the client connection immediately. Or if the connection opens and 0 bytes are sent, then will we skip the haproxy header entirely and then move to the worker which will cause the initial message to be invalid BER?

I think there are a lot of ways this can go wrong in it's current design, and I'm worried about how that will impact clients in subtle ways. I think the right thing is to check for the haproxy header in the worker threads before the first message is received, and if *then* it's present and is a valid haproxy header we pivot the address of the conn. If there is no header we proceed as usual.

droideck: Thanks! Sounds good, and I surely don't mind. :)
As I mentioned, my 389 DS connection management knowledge is sparse.
Will this be a good place to put the `haproxy_receive()` call and then continue with logging? (line 1171)
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/connection.c#L1171

Is it what you meant?

ldap/servers/slapd/haproxy.c
droideck: Nope, one `recv` is for peeking only. And the second one removes the HAProxy header when it's parsed successfully.

ldap/servers/slapd/haproxy.c
droideck: Awesome find! Thanks!

progier389 (approved):
LGTM

test/test_slapd.h
```diff
@@ -55,6 +55,12 @@ void test_libslapd_counters_atomic_overflow(void **state);
 void test_libslapd_pal_meminfo(void **state);
 void test_libslapd_util_cachesane(void **state);
 
+/* libslapd-haproxy */
+void test_libslapd_haproxy_v1(void **state);
```
Firstyear, line 59: deeply appreciate you using cmocka here <3

ldap/servers/slapd/haproxy.c
```diff
@@ -0,0 +1,392 @@
...
+ * Function to receive and parse HAProxy headers, supporting both v1 and v2 of the protocol.
+ * 
+ * @param fd: The file descriptor of the socket from which to read.
+ * @param proxy_connection: A pointer to an integer to store the proxy connection status (0 or 1).
+ * @param pr_netaddr_from: A pointer to a PRNetAddr structure to store the source address info.
+ * @param pr_netaddr_dest: A pointer to a PRNetAddr structure to store the destination address info.
+ * 
+ * @return: Returns 0 on successful operation, -1 on error.
+ */
+int haproxy_receive(int fd, int *proxy_connection, PRNetAddr *pr_netaddr_from, PRNetAddr *pr_netaddr_dest)
```
Firstyear, line 350: A possible case here is that you have the DS server behind haproxy, but an admin bypasses haproxy to ldap to the server. That would cause this to error because the BER doesn't decode to a haproxy header leading to a disconnect. I think that this needs to handle the haproxy header being *optional* rather than the current setup of it being required (I think ... if I'm reading it properly).

Firstyear (changes requested):
This looks way better, I really appreciate you moving the parser into the connection read operation area. That will really help with the performance.

ldap/servers/slapd/haproxy.c
progier389: As I read it, if the first PDU does not come through an haproxy, the first PDU match neither haproxy V1 nor V2 signature so haproxy_receive returns HAPROXY_NOT_A_HEADER then connection_read_operation will just go on processing the request happily.
So the header is already optionnal. BTW if it was not the case, most of the CI tests would have failed ! -;)
That said, there may be a loophole somewhere because some of the CI tests suspiciously fail with errors related to connection closure but I have not seen P3 or P4 disconnection code in the artifacts so it may not be related either ...

Firstyear: Ahh I misread the logic in the if statement, yes, it only checks for the _ERROR case. 

CI would also be passing since haproxy parsing isn't on by default. So I think that in the haproxy = true case, there needs to be some tests for when we don't have haproxy incoming.

droideck: A simple test for the attribute and binds is added. We can think about something more complex later. This PR already becomes too big to handle. :)

progier389: Sounds fair but I wonder if we should not also add an environment variable that will be taken in account to set the attribute when creating the instances (i.e in _create_instances in topologies.py)
That way we should be able to easily run most of the existing tests with the haproxy enabled.

droideck: Okay, I added the environment variable, and I've hardcoded it so our CI can run with the nsslapd-haproxy-trusted-ip attribute set (so we check for HAProxy headers). And all looks good.
The existing failures are not related, I think.

@Firstyear, please, review :)

Firstyear (approved):

ldap/servers/slapd/libglobs.c
```diff
@@ -2678,6 +2683,59 @@ config_set_listenhost(const char *attrname __attribute__((unused)), char *value,
...
+        PL_strncasecmp((char *)value[0]->bv_val, HAPROXY_TRUSTED_IP_REMOVE_CMD, value[0]->bv_len) != 0) {
+        for (size_t i = 0; value[i] != NULL; i++) {
+            end = strspn(value[i]->bv_val, "0123456789:ABCDEFabcdef.*");
+            /* 
+            * If no valid characters are found, or if there are characters after the valid ones,
+            * then print an error message and exit with LDAP_OPERATIONS_ERROR.
+            */
+            if (!end || value[i]->bv_val[end] != '\0') {
+                slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE, "IP address contains invalid characters (%s), skipping\n",
+                                value[i]->bv_val);
```
mreynolds389, line 2716: indent, same below for the same function call.

mreynolds389 (approved):
Nice work Simon!  Ack

Issue #5765, opened 2023-05-11
Improve installer selinux handling


mreynolds389:
**Issue Description**

When labeling a file sometimes the OS is "busy" and returns a resource unavailable error.  When labeling ports we will retry, but not when we label files/directories.  We should use the same level of robustness for both ports and files.

```
Validate installation settings ...
    Create file system structures ...
    Perform SELinux labeling ...
    Setting label dirsrv_var_lib_t in seLinux file context /var/lib/dirsrv/slapd-RHEL87-HMSIDM-TEST/bak.
    ERROR CMD: semanage fcontext -a -t dirsrv_var_lib_t /var/lib/dirsrv/slapd-RHEL87-HMSIDM-TEST/bak ; STDOUT:  ; STDERR: libsemanage.semanage_get_lock: Could not get direct transaction lock at /var/lib/selinux/targeted/semanage.trans.LOCK. (Resource temporarily unavailable).
    BlockingIOError: [Errno 11] Resource temporarily unavailable
  
    Instance creation failed!  Failed to set SElinux label dirsrv_var_lib_t on /var/lib/dirsrv/slapd-RHEL87-HMSIDM-TEST/bak: Command '['semanage', 'fcontext', '-a', '-t', 'dirsrv_var_lib_t', '/var/lib/dirsrv/slapd-RHEL87-HMSIDM-TEST/bak']' returned non-zero exit status 1
```

mreynolds389:
2a27121f1..dfa4e810d  389-ds-base-2.3 -> 389-ds-base-2.3
09236fb2d..5a96da059  389-ds-base-2.2 -> 389-ds-base-2.2
cd3d6241e..c95231bb9  389-ds-base-2.1 -> 389-ds-base-2.1
d88083b0f..77cb1716a  389-ds-base-2.0 -> 389-ds-base-2.0

PR #5766, opened 2023-05-11
Issue 5765 - Improve installer selinux handling

mreynolds389:
Description:  

When labeling ports we retry on error, and we should do the same when labeling files

relates: https://github.com/389ds/389-ds-base/issues/5765

### DIFF
--- a/src/lib389/lib389/utils.py
+++ b/src/lib389/lib389/utils.py
@@ -270,7 +270,7 @@ def _get_selinux_fcontext_info():
 
 
 def resolve_selinux_path(path):
-    '''Return the path as expected by semanage fcontext'''
+    """Return the path as expected by semanage fcontext"""
     path = str(Path(path).resolve())
     if selinux_present():
         _get_selinux_fcontext_info()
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/utils.py
```diff
@@ -309,20 +309,34 @@ def selinux_label_file(path, label):
...
-            log.debug(f"Setting label {label} in SELinux file context {path}.")
-            result = subprocess.run(["semanage", "fcontext", "-a", "-t", label, path],
-                                     stdout=subprocess.PIPE,
-                                     stderr=subprocess.PIPE)
-            args = ' '.join(ensure_list_str(result.args))
-            stdout = ensure_str(result.stdout)
-            stderr = ensure_str(result.stderr)
-            if result.returncode != 0:
-                log.error(f"ERROR CMD: {args} ; STDOUT: {stdout} ; STDERR: {stderr}")
+        rc = 0
```
droideck, line 312: While you are there, could you please fix the docstring typo I've just noticed (wrong type label):

```
    :param label: The label to set (None to unset a label)
    :type path: str
```

droideck (approved):
Looks good!

Issue #5768, opened 2023-05-12
Broken import of CA certs


lolllpop:
**Issue Description**
Importing CA certs from /data/tls/ca/*.crt into docker container results in error message, container does not start.

**Package Version and Platform:**
 - Platform: docker / podman 4.4.1
 - Package and version: docker.io/389ds/dirsrv:2.4

**Steps to Reproduce**
1. setup folder containing the certs, all in PEM format
```
ssl/
├── ca/
│   ├── ca-inter.crt
│   └── ca-root.crt
├── server.crt
└── server.key
```

2. start docker container 
```
podman run --replace --name 389ds \
  --volume ./ssl:/data/tls \
  --env DS_DM_PASSWORD=xxx123 \
  --env DS_SUFFIX_NAME=dc=my,dc=org \
  docker.io/389ds/dirsrv:2.4
```

3. Error message importing a CA cert
```
INFO: Checking for PEM TLS files ...
INFO: Found -> ['ca-root.crt', 'ca-inter.crt']
INFO: Have /data/tls/server.key -> True
INFO: Have /data/tls/server.crt -> True
INFO: Have /data/tls/ca -> True
INFO: Have /data/config/pwdfile.txt -> True
INFO: TLS PEM requirements met - configuring NSSDB ...
DEBUG: Allocate local instance <class 'lib389.DirSrv'> with None
DEBUG: del_cert cmd: /usr/bin/certutil -D -d /data/config -n Server-Cert -f /data/config/pwdfile.txt
INFO: Enrolling -> /data/tls/ca/ca-root.crt
Traceback (most recent call last):
  File "/usr/lib/dirsrv/dscontainer", line 467, in <module>
    begin_magic()
  File "/usr/lib/dirsrv/dscontainer", line 300, in begin_magic
    _begin_setup_pem_tls()
  File "/usr/lib/dirsrv/dscontainer", line 141, in _begin_setup_pem_tls
    tls.add_cert(nickname=ca_path, input_file=ca_path)
  File "/usr/lib/python3.10/site-packages/lib389/nss_ssl.py", line 1187, in add_cert
    raise ValueError(f"Certificate ({nickname}) is not a server certificate")
ValueError: Certificate (/data/tls/ca/ca-root.crt) is not a server certificate
```

**Expected results**
CA certs get imported and used, e.g. for replication via LDAPs later on.

**Additional context**
[...]

PR #5769, opened 2023-05-16
Issue 5768 - CLI/UI - cert checks are too strict, and other issues

mreynolds389:
Description:

THe certificate type checks for CA/server break if there are no certificate extensions set (use openssl in that case to gather the info instead). dscontainter needed to be updated for new cert checks, and UI adding certs improvements.

relates: https://github.com/389ds/389-ds-base/issues/5768


droideck (approved):
LGTM!

Issue #5751, opened 2023-05-03
Cleanallruv task crashes on consumer


tbordaz:
**Issue Description**
When a consumer processes a cleanallRuv task it tries to clean its CL. A consumer has not CL and so dereference a NULL pointer

**Package Version and Platform:**
It is a side effect of CL refactoring that landed in RHEL9/RHDS12.

**Steps to Reproduce**
Create a M2_C2 topology. Make sure replication works. then launch
dsconf supplier1 repl-tasks cleanallruv --suffix 'dc=example,dc=com' --replica-id 2 

**Expected results**
Consumer should not crash

tbordaz:
Linked to BZ 2192827

PR #5775, opened 2023-05-22
Issue 5751 - Cleanallruv task crashes on consumer

tbordaz:
Bug description:
	During CL refactoring (changelog DB was integrated into the main DB #2621)
	several parts of code (removed DB, export/import CL,
	cleanallRUV,..) calls replica_get_cl_info to retrieve the
	changelog of a replica. If the replica does not contain a
	changelog (consumer) the returned pointer is NULL.
	Some code assume the pointer is not NULL and derefence it.

Fix description:
	For all calls to replica_get_cl_info, check the pointer
	before referencing it

relates: #5751

Reviewed by:


mreynolds389 (approved):
Ack

Issue #5778, opened 2023-05-25
UI - error message in Monitor tab if .dsrc file does not exist


mreynolds389:
**Issue Description**

The UI shows an error when clicking on the Monitor tab if the .dsrc file does not exist.  This error should be ignored since the .dsrc file is not required.

mreynolds389:
d3fe48a61..6e04b6fcb  389-ds-base-2.3 -> 389-ds-base-2.3
28adbec75..22c174ad0  389-ds-base-2.2 -> 389-ds-base-2.2

PR #5779, opened 2023-05-25
Issue 5778 - UI - Remove error message if .dsrc is missing

mreynolds389:
Description:  

Having a .dsrc file is not required, so the UI should not report an error if it's not present

relates: https://github.com/389ds/389-ds-base/issues/5778


droideck (approved):

progier389 (approved):
LGTM

PR #5787, opened 2023-06-05
Issue 5786 - Set minimal permissions on GitHub Workflows

### LINKED ISSUE #5786: CI: set minimal permissions on GitHub Workflows
Similarly to this issue: https://github.com/python-ldap/python-ldap/issues/523
We need to set minimal permissions on our GitHub Workflows.

Defining minimal permissions secures you against erroneous or malicious behaviour from external jobs you call from your workflow. It's especially important in case they get compromised.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

droideck:
Set minimal permissions on our GitHub Workflows.
Defining minimal permissions secures you against erroneous or malicious behavior from external jobs you call from your workflow. It's especially important in case they get compromised.

Fixes: https://github.com/389ds/389-ds-base/issues/5786

Reviewed by: ?

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -3,6 +3,11 @@ on:
   - pull_request
   - push
 
+permissions: 
+  actions: read
+  packages: read
+  contents: read
+
 jobs:
   compile:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


mreynolds389 (approved):
LGTM

Issue #5789, opened 2023-06-05
Improve error handling when replica has not been initialized


mreynolds389:
**Issue Description**

When a replica has not been initialized we get odd/vague errors in the logs and by ds-replcheck, making it hard to understand what is wrong.  We can improve the logging and CLI tools to better handle these situations

See also:

- https://bugzilla.redhat.com/show_bug.cgi?id=2208104
- https://bugzilla.redhat.com/show_bug.cgi?id=2211690 (closed not a bug)

PR #5790, opened 2023-06-06
Issue 5789 - Improve ds-replcheck error handling

mreynolds389:
Description:  

When replication is not fully configured the tool outputs vaguemessages.  These should be cleaned up to indicate that replication was not initialized

Relates: https://github.com/389ds/389-ds-base/issues/5789

### DIFF
--- a/src/lib389/lib389/replica.py
+++ b/src/lib389/lib389/replica.py
@@ -45,7 +45,7 @@ from lib389.idm.services import ServiceAccounts
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.conflicts import ConflictEntries
 from lib389.lint import (DSREPLLE0001, DSREPLLE0002, DSREPLLE0003, DSREPLLE0004,
-                         DSREPLLE0005, DSCLLE0001)
+                         DSREPLLE0005, DSREPLLE0006, DSCLLE0001)
 
 
 class ReplicaLegacy(object):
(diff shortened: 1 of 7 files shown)


src/lib389/lib389/replica.py
```diff
@@ -1650,6 +1650,9 @@ def get_ruv(self):
                 escapehatch='i am sure')[0]
             data = ensure_list_str(ent.getValues('nsds50ruv'))
         except IndexError:
+            # There are no ruv elements, it's okay
+            pass
+        except ldap.NO_SUCH_OBJECT:
```
droideck: Minor nitpick, you can unite the exceptions with `except (IndexError, ldap.NO_SUCH_OBJECT):` line as the results are the same anyway.

droideck (approved):
The rest looks good!

src/lib389/lib389/replica.py
```diff
@@ -1650,6 +1650,9 @@ def get_ruv(self):
...
             # There is no ruv entry, it's okay
             pass
```
tbordaz: The BZ (https://bugzilla.redhat.com/show_bug.cgi?id=2208104#c8) is related to a invalid replication setup. A replica containing a RUV entry but no replicated suffix entry. So the RUV can not be found with a get_ruv search (err=32).

Can this patch be used for such invalid setup ?

progier389: I think that trapping the exception would prevent the exception but people will still be puzzled to see an empty ruv.
I would have expected that a message was also added in get_ruv (li_conf/replication.py) if there are no ruvs to warn the admin that the db is likely uninitialized.

mreynolds389: But I am reporting that the replica needs to be initialized.  Am I missing something?

At least for ds-replcheck I am, for get_ruv() I am not, but I don't think that's the issue here since that is just a helper function.  In the code we were ignoring the missing RUV if we got an index error, so I was just continuing with that same logic.

progier389: at least wwo commands are impacted: 
-    **ds-replcheck**
-    **dsconf instance replication get-ruv --suffix suffix**

In ds-replcheck you do log proper message but not in dsconf get-ruv case
BTW: IMHO you should also check that **dsctl ps healthcheck** give a clear report

mreynolds389:
Changes made, please review...

tbordaz (approved):
LGTM

progier389 (approved):
LGTM

Issue #5785, opened 2023-06-01
CLI - arg completion is broken


mreynolds389:
**Issue Description**

In Fedora 38, maybe other releases, CLI arg completion broke.  Maybe something changed in python, not sure.  Needs investigation

mreynolds389:
Needed to run:  

activate-global-python-argcomplete

It's now better but not 100% working.  Still investigating

mreynolds389:
Ok everything is working, but I wonder if we should call the  **activate-global-python-argcomplete** script during instance creation? 

We really need argcompletion working out of the box, or else the CLI is very annoying to use.

vashirov:
Fedora bug for python-argcomplete: https://bugzilla.redhat.com/show_bug.cgi?id=2211862

mreynolds389:
This diff seems to be sufficient:

```
diff --git a/rpm/389-ds-base.spec.in b/rpm/389-ds-base.spec.in
index 87af05a59..662a737c9 100644
--- a/rpm/389-ds-base.spec.in
+++ b/rpm/389-ds-base.spec.in
@@ -404,6 +404,12 @@ autoreconf -fiv
   sed -r -i '/^(Protect(Home|Hostname|KernelLogs)|PrivateMounts)=/d' %{_builddir}/%{name}-%{version}%{?prerel}/wrappers/*.service.in
 %endif
 
+for clitool in dsconf dsctl dsidm dscreate
+do
+  register-python-argcomplete "${clitool}" > "/usr/share/bash-completion/completions/${clitool}"
+done
+
 %if 0%{?rhel} > 7 || 0%{?fedora}
 # lib389
 pushd ./src/lib389
```

mreynolds389:
6e04b6fcb..5f9b4d6fe  389-ds-base-2.3 -> 389-ds-base-2.3

adelton:
The change caused regression https://bugzilla.redhat.com/show_bug.cgi?id=2225231.

Please note that the test builds that Miro showed https://bugzilla.redhat.com/show_bug.cgi?id=2211862 used %files properly, not %post.

PR #5792, opened 2023-06-08
Issue 5785 - move bash completion to post section of specfile

mreynolds389:
Description:  Need to move bash completion setup to %post section of specfile. Previously it was done during the build process which is incorrect and breaks builds.

relates: https://github.com/389ds/389-ds-base/issues/5785

Reviewed by: ?


droideck (approved):
LGTM

PR #5641, opened 2023-02-08
Issue 5632 - CLI - improve error handling with db2ldif

mreynolds389:
Description: 

Have the CLI check if the ldif location exists.  This also prevents the database from getting trashed by skipping the export attempt.

relates: https://github.com/389ds/389-ds-base/issues/5632

PR #5795, opened 2023-06-09
Allow empty export path for db2ldif

multipleofzero:
Until recently, `db2ldif` did not require an export path to be specified and would use a specified default location to create a timestamped file.

See [the `else` statement in
`__init__.py`](https://github.com/389ds/389-ds-base/blob/a30a9e0d6f74774e3c5a0bbe11347d13dc37ddb4/src/lib389/lib389/__init__.py#L2759)

Commit cee276d which was merged in #5641 introduced
[a check](https://github.com/389ds/389-ds-base/commit/cee276d3983aafe6459b4b6545b4d01200a9b892#diff-0df3612e00b94edc2072af211e8c3bacaec694cdfa0f84b0b42a84066f119957R60) to ensure the targeted export file's parent actually exists. This check is fine if a target filename *is* provided, but it fails when no filename is provided; a use case that was supported before.

Therefore, the check should only be done *iff* a filename for the export is provided.

### DIFF
--- a/dirsrvtests/tests/suites/export/export_test.py
+++ b/dirsrvtests/tests/suites/export/export_test.py
@@ -55,7 +55,7 @@ def test_dbtasks_db2ldif_with_non_accessible_ldif_file_path(topo):
         1. Operation successful
         2. Operation properly fails, without crashing
         3. An error code different from 139 (segmentation fault) should be reported
-        4. "The LDIF file location does not exist" is returned
+        4. "location does not exist" is returned
     """
     export_ldif = '/tmp/nonexistent/export.ldif'
 
@@ -63,10 +63,10 @@ def test_dbtasks_db2ldif_with_non_accessible_ldif_file_path(topo):
     topo.standalone.stop()
 
     log.info("Performing an offline export to a non accessible ldif file path - should fail properly")
-    expected_output="The LDIF file location does not exist"
+    expected_output="location does not exist"
     with pytest.raises(ValueError) as e:
         run_db2ldif_and_clear_logs(topo, topo.standalone, DEFAULT_BENAME, export_ldif, expected_output)
-    assert "The LDIF file location does not exist" in str(e.value)
+    assert "location does not exist" in str(e.value)
 
     log.info("Restarting the instance...")
     topo.standalone.start()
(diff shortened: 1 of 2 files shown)


mreynolds389:
Thanks for the contribution.  This PR is breaking an existing test.  Can you please apply this diff to your PR?

```
diff --git a/dirsrvtests/tests/suites/export/export_test.py b/dirsrvtests/tests/suites/export/export_test.py
index cb6badbe8..7d9039778 100644
--- a/dirsrvtests/tests/suites/export/export_test.py
+++ b/dirsrvtests/tests/suites/export/export_test.py
@@ -55,7 +55,7 @@ def test_dbtasks_db2ldif_with_non_accessible_ldif_file_path(topo):
         1. Operation successful
         2. Operation properly fails, without crashing
         3. An error code different from 139 (segmentation fault) should be reported
-        4. "The LDIF file location does not exist" is returned
+        4. "location does not exist" is returned
     """
     export_ldif = '/tmp/nonexistent/export.ldif'
 
@@ -63,10 +63,10 @@ def test_dbtasks_db2ldif_with_non_accessible_ldif_file_path(topo):
     topo.standalone.stop()
 
     log.info("Performing an offline export to a non accessible ldif file path - should fail properly")
-    expected_output="The LDIF file location does not exist"
+    expected_output="location does not exist"
     with pytest.raises(ValueError) as e:
         run_db2ldif_and_clear_logs(topo, topo.standalone, DEFAULT_BENAME, export_ldif, expected_output)
-    assert "The LDIF file location does not exist" in str(e.value)
+    assert "location does not exist" in str(e.value)
 
     log.info("Restarting the instance...")
     topo.standalone.start()
```

multipleofzero:
Applied the diff, thank you!

mreynolds389:
> Applied the diff, thank you @mreynolds389!

Running the checks again, if all is well we can merge this!

droideck (approved):
I run these tests: `py.test -s -v ds/dirsrvtests/tests/suites/clu/ ds/dirsrvtests/tests/suites/export/ ds/dirsrvtests/tests/suites/import/`

And results are all good. Ack from me!

```
===== 73 passed, 2 xfailed, 1 xpassed, 157 warnings in 2766.52s (0:46:06) ======
```

Please squash the commits and update the commit message via our guidelines:

Thanks!

multipleofzero:
Created RFE issue (#5822) and squashed the two commits into one.

PR #5796, opened 2023-06-12
Issue 5761 - Worker thread dynamic management

### LINKED ISSUE #5761: Dynamically configure number of worker threads
We have seen several cases where the issue was a tuning of the number of worker threads that handles the connection I/O and the operations. The idea is to make the administrator life easier by auto adapting dynamically that number.

A solution that:
1.   Try to adapt the configuration to the current load x
2.   Allows to dynamically change the maximum/minimum number of threads without having to restart the server
3.   Do not overreact on sudden peak of load 
4.   Implement some safeguards to protect 

The solution that I am thinking about implements:
   for [1]: add/remove a few threads based on average number of threads versus average work queue size
   for [2]: use a sorted list instead of an array to keep the snmp index
   for [3]: keep last number of thread change time and implement a delay before doing next such change
   for [4]: have a minimum and maximum configurable thread number
                minimum is the current nsslapd-threadsnumber (to ease up upgrade)
                maximum default value is configured according to the number of cores (and maybe the available memory)   

**Additional context**
Issue 5159](https://github.com/389ds/389-ds-base/issues/5159) discussion is partially related to this RFE
But the current feature does not take in account the threads somehow waiting for locks while handling an operation)
although I agree with @Firstyear that is an important point, (because we will likely be yet again in the server unresponsive situation once the maximum configurable thread number
I think we should process incrementally and handled this point later on in another issue.
especially since the answer is probably not related to the number of worker threads but how the server behaves when the load is higher than it can handle 
(IMHO at some point we should return LDAP_BUSY for new operation coming on the locked backend to warn the client that it is useless to retry too soon and lefting worker threads for other backend like config change or monitoring)

progier389:
Objectives: 
- Allow to configure the number of worker threads without having to restart the server
- Decrease the worker thread global mutex contention but removing the associated condition variable
  ==> Increase the "searchrate" performance 
  
Solution: See https://github.com/389ds/389ds.github.io/blob/main/docs/389ds/design/worker-threads.md

Issue: [5761](https://github.com/389ds/389-ds-base/issues/5761)

Reviewed by:  @tbordaz    (Thanks!)

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -480,11 +480,13 @@ op_thread_tinfo_init(int thread_idx)
 void
 op_thread_set_threads_number(int threadsnumber)
 {
+    static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
     int oldnbthreads = 0;
     pc_tinfo_t **oldthreads = NULL;
     int len = 0;
     int rc = 0;
 
+    pthread_mutex_lock(&mutex);
     /* Adjust snmp variables */
     pthread_mutex_lock(&g_pc.snmp.mutex);
     oldnbthreads = g_pc.snmp.nbthreads;
@@ -547,6 +549,7 @@ op_thread_set_threads_number(int threadsnumber)
     g_pc.waiting_jobs.hwm = 0;
     g_pc.jobs_free_list.hwm = 0;
     pthread_mutex_unlock(&g_pc.snmp.mutex);
+    pthread_mutex_unlock(&mutex);
     slapi_ch_free((void**)&oldthreads);
 }
 
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -23,12 +23,14 @@
 #include "prlog.h" /* for PR_ASSERT */
 #include "fe.h"
 #include <sasl/sasl.h>
+#include "wthreads.h"
+#include "snmp_collator.h"
 #if defined(LINUX)
 #include <netinet/tcp.h> /* for TCP_CORK */
 #endif
 
 typedef Connection work_q_item;
-static void connection_threadmain(void *arg);
+static void *connection_threadmain(void *arg);
 static void connection_add_operation(Connection *conn, Operation *op);
 static void connection_free_private_buffer(Connection *conn);
 static void op_copy_identity(Connection *conn, Operation *op);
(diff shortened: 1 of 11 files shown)


progier389:
Creating the PR to be able to easily share the prototype.
So far the firsts tests using ldclt -e bindeach,esearch with default config on my laptop and was a bit surprised by the result and had to double check in the access log shows that the bind and search are really done ! -;)
Looking forward for the real performance tests ...

progier389:
Performance tests shows that there are no performance regression during stress tests but no much improvement either
Work queue lock contention time decreased but once again it does not seems to impact the stress performance 
The main interest is that the number of thread can now be changed dynamically.

ldap/servers/slapd/connection.c
```diff
@@ -425,71 +419,191 @@ connection_reset(Connection *conn, int ns, PRNetAddr *from, int fromLen __attrib
...
+    return NULL;
+}
+
+/*
+ * Increase or decrease the number of worker threads
+ */
+void
+op_thread_set_threads_number(int threadsnumber)
+{
+    int oldnbthreads = 0;
```
tbordaz, line 484: If op_thread_set_thread_number is called in parallel, should not serialize the call ?

ldap/servers/slapd/wthreads.h
```diff
@@ -0,0 +1,89 @@
...
+    } snmp;
+} pc_t;
+
+/* Connection status values returned by
+    connection_wait_for_new_work(), connection_read_operation(), etc. */
+typedef enum {
+    CONN_FOUND_WORK_TO_DO,
+    CONN_SHUTDOWN,
+    CONN_NOWORK,
+    CONN_DONE,
```
tbordaz, line 80: Is it still needed ?

ldap/servers/slapd/connection.c
```diff
@@ -963,57 +1069,81 @@ connection_free_private_buffer(Connection *conn)
...
-     * in connection_threadmain, and just clear it out
+    /*
+     * Init the pblock with conn and a fresh operation
+     * Operation no more realloced or got from a stack but
+     * it is stored in worker thread context and reused afterwards
+     * Caller should held conn->c_mutex
      */
-    /* *ppb = (Slapi_PBlock *) slapi_ch_calloc( 1, sizeof(Slapi_PBlock) ); */
-    /* *ppb = slapi_pblock_new(); */
+    Slapi_Operation *op = tinfo->op;
```
tbordaz, line 1083: tinfo->op is set when g_pc_do_not_reuse_operation() is called. It looks surprising to call operation_init without testing 'op'

tbordaz (commented):
The code looks very nice. Just few minor comments/questions

ldap/servers/slapd/connection.c
progier389: Good catch. I should add a specific mutex to serialize this function

ldap/servers/slapd/wthreads.h
progier389: Yes, these values are still used by the worker thread (but anyway a merge with the turbo mode removal is needed)

ldap/servers/slapd/connection.c
progier389: No problem here because operation_init checks for NULL operation

ldap/servers/slapd/wthreads.h
tbordaz: I was more specifically talking about CONN_DONE (IIRC)

progier389: grep -c CONN_DONE connection.c 
18

ldap/servers/slapd/connection.c
progier389: It is now fixed by latest commit

tbordaz (approved):
Sorry for the delay. LGTM

PR #5799, opened 2023-06-14
Issue 5798 - CLI - Add multi-valued support to dsconf config

### LINKED ISSUE #5798: CLI - dsconf inst config should support multi-valued operations
**Issue Description**
We have two editable multi-valued attributes in cn=config: nsslapd-haproxy-trusted-ip and nsslapd-referral.

Our current cn=config implementation doesn't support bunch ADD operations.
And while it'll be too much work to refactor our cn=config (and I'm not sure we need that at all), we still should
make our CLI tools more robust so it can handle multi-valued attributes correctly.

droideck:
Description: Currently, we have two editable multi-valued attributes in cn=config: nsslapd-haproxy-trusted-ip and nsslapd-referral.

Our current cn=config implementation doesn't support bunch ADD operations. Make our CLI tools more robust so it can handle multi-valued attributes correctly.

Add add_many method to DSLdapObject.

Fixes: https://github.com/389ds/389-ds-base/issues/5798

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_conf/config.py
+++ b/src/lib389/lib389/cli_conf/config.py
@@ -60,7 +60,7 @@ def _config_get_existing_attrs(conf, args, op_type):
                 if op_type == OpType.delete:
                     conf.remove_all(attr)
                 else:
-                    raise ValueError("You must specify a value to add for the attribute ({})".format(attr_name))
+                    raise ValueError(f"You must specify a value to add for the attribute ({attr_name})")
         return existing_attrs
     else:
         # Missing value
@@ -100,7 +100,7 @@ def config_add_attr(inst, basedn, log, args):
                 except ldap.TYPE_OR_VALUE_EXISTS:
                     pass
             else:
-                raise ValueError("You must specify a value to add for the attribute ({})".format(attr_name))
+                raise ValueError(f"You must specify a value to add for the attribute ({attr_name})")
     else:
         # Missing value
         raise ValueError("Missing attribute to add")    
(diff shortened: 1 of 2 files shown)


tbordaz (commented):
I know it is late to suggest that change but would it make sense to set the 'nsslapd-haproxy-trusted-ip' values in a config subentry like cn=haproxy,cn=config ?

progier389:
> I know it is late to suggest that change but would it make sense to set the 'nsslapd-haproxy-trusted-ip' values in a config subentry like cn=haproxy,cn=config ?

IMHO the answer is no, it would not make sense:
- What would be the benefit of such a change ?  Unless mistaken, it just add complexity.
- Furthermore 'nsslapd-haproxy-trusted-ip' is not the only multi valued attribute in cn=config so @droideck solution also solve the referrals handling issue.

src/lib389/lib389/cli_conf/config.py
```diff
@@ -25,6 +27,46 @@ def _config_display_ldapimaprootdn_warning(log, args):
...
+                        if existing_attrs[attr_name] == []:
+                            del existing_attrs[attr_name]
+
+                    if op_type == OpType.delete:
+                        conf.remove_all(attr_name)
+            else:
+                if op_type == OpType.delete:
+                    conf.remove_all(attr)
+                else:
+                    raise ValueError("You must specify a value to add for the attribute ({})".format(attr_name))
```
mreynolds389: Very minor issue, and you don't have to do this, but I'd like to use the "f" prefix instead of "format()".  Again there is nothing wrong here, so I'll leave it up to you.

In the meantime, Ack!

mreynolds389 (approved):

Issue #5793, opened 2023-06-08
UI - switch from webpack to esbuild bundler


mreynolds389:
**Issue Description**

To stay consistent and current with Cockpit (via https://github.com/cockpit-project/starter-kit) we need to move from webpack to esbuild

mreynolds389:
cb83c8f12..0272d9914  389-ds-base-2.3 -> 389-ds-base-2.3

droideck:
6b9a86cc1..fe348d80a  389-ds-base-2.3 -> 389-ds-base-2.3

mreynolds389:
Fix LDIF export suffix selection and other crashes related to webpack/rebuild migration:

fe348d80a..8f797b9b7  389-ds-base-2.3 -> 389-ds-base-2.3

Fix just import selection:

94d69c7d0..8a7e696f8  389-ds-base-2.2 -> 389-ds-base-2.2
e377c9af1..cccc9ef4a  389-ds-base-2.1 -> 389-ds-base-2.1
09e984337..e07780678  389-ds-base-2.0 -> 389-ds-base-2.0
dc4130da5..a4561f4c1  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

droideck:
a8f062ef9..9e6979e31  389-ds-base-2.3 -> 389-ds-base-2.3
   48834f141..7fbf7bbbd  389-ds-base-2.2 -> 389-ds-base-2.2
   cccc9ef4a..1cdd011eb  389-ds-base-2.1 -> 389-ds-base-2.1
   e07780678..65bcca75a  389-ds-base-2.0 -> 389-ds-base-2.0
   a4561f4c1..e568180d1  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5800, opened 2023-06-15
Issue 5793 - UI - move from webpack to esbuild bundler

mreynolds389:
Description:

To stay consistent and current with Cockpit (via https://github.com/cockpit-project/starter-kit) we need to move from webpack to esbuild

relates: https://github.com/389ds/389-ds-base/issues/5793


Makefile.am
```diff
@@ -2099,10 +2099,10 @@ WEBPACK_TEST = src/cockpit/389-console/cockpit_dist/index.html
 # Cockpit UI plugin - we install the dependancies and build the JS sources
 # and then we use install-data-hook for copying the results on 'make install'
 $(NODE_MODULES_TEST):
```
droideck, line 2101: I think we also need to change to `NODE_MODULES_TEST = src/cockpit/389-console/package-lock.json` here, in Makefile.am

src/cockpit/389-console/node_modules.mk
```diff
@@ -1,15 +0,0 @@
-install: package.json
-	npm ci
```
droideck, line 2: I think we still need to include `npm ci` and some kind of `npm run audit-ci` (but it's probably done differently with build.js way) if it's possible.
Now I see in the build log that we don't check for vulnerabilities at all.

rpm.mk
```diff
@@ -47,31 +47,32 @@ bundle-rust:
...
+	NODE_ENV=production ./build.js; \
+	cp -r dist cockpit_dist
 endif
 
 dist-bz2: install-node-modules download-cargo-dependencies
 ifeq ($(COCKPIT_ON), 1)
 	cd src/cockpit/389-console; \
 	rm -rf cockpit_dist; \
-	NODE_ENV=production make -f node_modules.mk build-cockpit-plugin; \
-	mv node_modules node_modules.release; \
```
droideck, line 65: These lines were for the case that we don't need all of the npm packages in our release tarball.

And I think, Cockpit has the same idea:
https://github.com/cockpit-project/cockpit/blob/main/package.json#LL3C61-L3C61

For some reason, they still have something in their tarball but it's minimal:
~/src/cockpit/cockpit-294.1/node_modules on rawhide ?1 ❯ ll                                                                                                                     at 09:02:39 AM
total 0
drwxr-xr-x. 1 spichugi spichugi 170 Jun 15 04:09 chrome-remote-interface
drwxr-xr-x. 1 spichugi spichugi  92 Jun 15 04:09 commander
drwxr-xr-x. 1 spichugi spichugi  94 Jun 15 04:09 sizzle
drwxr-xr-x. 1 spichugi spichugi  98 Jun 15 04:09 ws

droideck (changes requested):
I think we should check that Fedora and RHEL can be built with the tarball made with these changes (including running `make -f rpm.mk dist-bz2` and `make -f rpm.mk rpms`)

And that UI functions correctly after that.

I've tried to build rpm with `mock` on a fresh VM and yep, we have unneeded node_modules there. And we have issues with UI (but I'm not sure if it's related to this PR or something else...): DN validation fails for correct DNs.

image

Makefile.am
mreynolds389: > I think we also need to change to `NODE_MODULES_TEST = src/cockpit/389-console/package-lock.json` here, in Makefile.am

I previously tried this and it caused build failures, but I'll look into it again

rpm.mk
mreynolds389: These lines also caused build failures because the following "tar" command expects node_modules to exist.

droideck: For the tar command (and eventually for the build time later on) we had `touch -r src/cockpit/389-console/package.json $(NODE_MODULES_TEST)` which essentially created an empty file that's sole purpose was to "trick" npm into desired usage.

mreynolds389: But there is no node_modules directory, so the tar command fails

droideck: Okay, my bad: to be presize, - previously, this helped `tar` command to succeed (and be correctly "tricked" into the desired state as I previously mentioned):

```
NODE_MODULES_TEST = src/cockpit/389-console/node_modules/webpack
mkdir -p $(NODE_MODULES_TEST)
touch -r src/cockpit/389-console/package.json $(NODE_MODULES_TEST)
```

rpm.mk
```diff
@@ -47,32 +47,35 @@ bundle-rust:
...
+	rm -rf node_modules; \
+	rm -rf node_modules.release; \
+	npm ci > /dev/null
 endif
 
 build-cockpit: install-node-modules
 ifeq ($(COCKPIT_ON), 1)
+ifndef SKIP_AUDIT_CI
+	cd src/cockpit/389-console; \
+	npm audit
```
droideck: If I understand correctly, `npm audit` will just check, but it won't stop the execution. We need `npm audit-ci` for complete control here. More info: https://www.npmjs.com/package/audit-ci
So if we have any critical vulnerability issues, our CI will fail immediately.

droideck (commented):
I've tested the fedora rawhide build with the change, and all looks good!
Also, I see that CI test failures are probably not related to the PR. I've rerun it a few times and always got a bit different results with all the same errors related to the fact that the component was not found (probably, the page didn't have enough time to load).

So besides the minor issue with audit-ci below - looks good!

I have one question, though. Why did you copy the cockpit-components here? I thought if we run as part of their package, we can use their component without issues... But I guess there's some issue with that...

rpm.mk
mreynolds389: There is no "npm audit-ci" anymore, but there is a "npx- audit-ci".  It seems to work correctly.

mreynolds389:
@droideck - I think everything has been addressed now....

rpm.mk
```diff
@@ -47,32 +47,35 @@ bundle-rust:
...
-	NODE_ENV=production make -f node_modules.mk build-cockpit-plugin
+	NODE_ENV=production ./build.js; \
+	cp -r dist cockpit_dist
 endif
 
 dist-bz2: install-node-modules download-cargo-dependencies
 ifeq ($(COCKPIT_ON), 1)
 	cd src/cockpit/389-console; \
 	rm -rf cockpit_dist; \
-	NODE_ENV=production make -f node_modules.mk build-cockpit-plugin; \
```
droideck, line 64: Previously, we were running audit-ci as part of `NODE_ENV=production make -f node_modules.mk build-cockpit-plugin;` here at `dist-bz2:`. But now when I execute `TAG=389-ds-base-2.4.4 make -f rpm.mk dist-bz2` - audit-ci is not run at all.

Should we also add it here?

src/cockpit/389-console/package.json
```diff
@@ -10,47 +11,41 @@
   "license": "GPL-3.0-or-later",
   "main": "index.js",
   "scripts": {
-    "build": "webpack",
-    "watch": "webpack --watch --progress",
-    "audit-ci": "audit-ci --config audit-ci.json",
```
droideck, line 15: When I try to run audit-ci, it asks me for the dependency.

```
npx audit-ci
Need to install the following packages:
  audit-ci
Ok to proceed? (y) y
```

I think we should add it back to src/cockpit/389-console/package.json...

droideck (approved):
A couple of minor issues - but the rest looks great! Ack

src/cockpit/389-console/package.json
mreynolds389: > When I try to run audit-ci, it asks me for the dependency.
> 
> ```
> npx audit-ci
> Need to install the following packages:
>   audit-ci
> Ok to proceed? (y) y
> ```
> 
> I think we should add it back to src/cockpit/389-console/package.json...

Done

PR #5801, opened 2023-06-15
Issue 3527 - Fix HAProxy x390x compatibility and compiler warnings

droideck:
Description: We need to support both big-endian (x390x) and little-endian (x86) architectures, it's better to dynamically adjust the byte order in our test cases based on the architecture of the system executing the tests. Define the values depending on the architecture.

Fix minor compiler warnings.

Related: https://github.com/389ds/389-ds-base/issues/3527

Reviewed by: ?


droideck:
I've tried to put together a function for the format insurance (with a combination of htonl and htons), but in the result, everything looked just too complex for a simple set of unit tests, so I removed everything and made it a simple arch check.

mreynolds389 (approved):

Issue #5755, opened 2023-05-04
The Massive memory leaking on update operations


i-ivantsov:
The Massive memory leaking on update operations are seen.

Package Version and Platform:

Platform: Debian GNU/Linux 10(buster)
Package and version: 389-ds-2.4.0~git0.87efeb2-163.2.x86_64 in docker container 389-Directory/2.4.0 B2023.116.0000

1) parallel update operations in 25 connections (read, create and delete operations are performed too);
2) ds389 server had used all available memory.

There is a diagram with memory usage collected with nmon and nmonchart: 
![memory leaks](https://user-images.githubusercontent.com/123725498/236256647-dcb8fcd4-f5a1-4452-8911-64eb8e37a7c4.jpg)

ATokma:
In my practice of using 389ds, I also observed a massive memory leak on versions 2.2, 2.3, 2.4 with creation, modification and deletion operations. Approximately 1.000,000 operations in 6 hours process 389ds allocate 32 Gb (all memory compunter) and crash proccess (out off  memory).

OGSmv:
I have the same problem with 389ds. During modification operations the memory grows up intensively and never be freed in future.

vashirov:
In Fedora/RHEL we ship jemalloc and LD_PRELOAD it in systemd unit for dirsrv. In a container this is not happening:
1. Image from Docker Hub ships a different rpm that doesn't bundle jemalloc, because they have a separate jemalloc rpm, but it's not installed in that image.
2. In order to use it you also have to explicitly pass LD_PRELOAD environment variable (which I think we should do by default as we do it in Fedora/RHEL).

@ATokma and @OGSmv, could you please confirm that you see this in a container too? Or what distro/package do you use?

@i-ivantsov, could you please make a new image based on docker.io/389ds/dirsrv that also includes jemalloc package and pass `LD_PRELOAD=/usr/lib64/libjemalloc.so.2` variable to the container? Or try an image from quay.io/389ds/dirsrv and pass `LD_PRELOAD=/usr/lib64/dirsrv/lib/libjemalloc.so.2` variable?

Thanks.

OGSmv:
> In Fedora/RHEL we ship jemalloc and LD_PRELOAD it in systemd unit for dirsrv. In a container this is not happening:
> 
> 1. Image from Docker Hub ships a different rpm that doesn't bundle jemalloc, because they have a separate jemalloc rpm, but it's not installed in that image.
[...]

PR #5803, opened 2023-06-16
Issue 5755 - The Massive memory leaking on update operations

jchapma:
Bug description: Memory leak with creation, modification and deletion operations.

Fix description: When multiple search filters are used we use set manipulation to construct the final idl results. In this corner case we set the idl_set->compliment_head pointer which is never freed.

relates: https://github.com/389ds/389-ds-base/issues/5755

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/back-ldbm/idl_set.c
+++ b/ldap/servers/slapd/back-ldbm/idl_set.c
@@ -168,6 +168,9 @@ idl_set_free_idls(IDListSet *idl_set)
 void
 idl_set_destroy(IDListSet *idl_set)
 {
+    if (idl_set->complement_head) {
+        idl_free(&idl_set->complement_head);
+    }
     slapi_ch_free((void **)&(idl_set));
 }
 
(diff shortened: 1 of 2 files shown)


vashirov:
Hey @jchapma,
this passes my [test](https://github.com/vashirov/389-ds-base/blob/afa2945a0419cb07ce20d0716678792b2e8096b5/dirsrvtests/tests/suites/memory_leaks/allids_search_test.py), feel free to include it as part of this PR.
Thanks!

ldap/servers/slapd/back-ldbm/idl_set.c
```diff
@@ -168,6 +168,9 @@ idl_set_free_idls(IDListSet *idl_set)
 void
 idl_set_destroy(IDListSet *idl_set)
 {
+    if (idl_set->complement_head) {
```
mreynolds389: You don't need to check for NULL here, it is already done in idl_free()

mreynolds389: Besides that ack

mreynolds389 (approved):

Issue #5804, opened 2023-06-18
dtablesize being set to soft maxfiledescriptor limit


jchapma:
**Issue Description**
389ds is not setting dtablesize properly based when systemd is setting the maxfiledescriptors with it's default settings.
     dtablesize stays 1024 which causes massive slowdown once you hit around 950 connection.

**Package Version and Platform:**
 - Platform: Rhel 8
 - Package and version: 389-ds-base-1.4.3

**Steps to Reproduce**
1. Confirm maxfiledescriptors value is set to 1024
2. Create ~950 connections
3. Notice queries are very slow. 
4. Failure to go above  ~970 concurrent connections.

**Expected results**
Support up to 10k connections

**Additional context**
Backport of https://github.com/389ds/389-ds-base/issues/5469

jchapma:
Merged to 389-ds-base-1.4.3

tbordaz:
Hi @jchapma ,  this fix was pushed in 1.4.3

a6f0981de..231f92a55  389-ds-base-1.4.3

Does it apply only in 1.4.3 or all branches up to main ?

jchapma:
The root cause of this issue is when the default connection_table_size is set too low (1024), limiting the number of connections the server can simultaneously support. 

The number of listeners feature https://github.com/389ds/389-ds-base/pull/5090 added extra connection table  monitoring threads along with increasing the default connection table size. This feature was merged into the 389-ds-base-2.3 branch. 

A subset of this feature, code that increases the default connection table size was backported to the below branches:
389-ds-base-1.4.3
389-ds-base-1.4.4
389-ds-base-2.0
389-ds-base-2.1
389-ds-base-2.2

tbordaz:
@jchapma , the ticket looks fixed in all impacted branch. Is it okay to close it ?

jchapma:
@tbordaz, yes i believe so, thanks

PR #5806, opened 2023-06-19
Issue 5804 - dtablesize being set to soft maxfiledescriptor limit

jchapma:
Bug Description: 389ds is not setting dtablesize properly based when systemd is setting the maxfiledescriptors with it's default settings. dtablesize stays 1024 which causes massive slowdown once you hit around 950 connection

Fix Description: dtablesize is set to the connection table size, this commit sets the connection table size/dtablesize to the system max file descriptor number less the reserve file descriptors.

relates: https://github.com/389ds/389-ds-base/issues/5804

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -82,7 +82,7 @@ PRFileDesc *signalpipe[2];
 static int writesignalpipe = SLAPD_INVALID_SOCKET;
 static int readsignalpipe = SLAPD_INVALID_SOCKET;
 #define FDS_SIGNAL_PIPE 0
-#define FDS_PROCESS_MAX 64000
+#define MAX_LDAP_CONNS 64000
 
 static PRThread *accept_thread_p = NULL;
 static PRThread *disk_thread_p = NULL;
@@ -1065,6 +1065,10 @@ slapd_daemon(daemon_ports_t *ports)
     uint64_t threads;
     int in_referral_mode = config_check_referral_mode();
     int connection_table_size = get_connection_table_size();
+    if (!connection_table_size) {
+        slapi_log_err(SLAPI_LOG_ERR, "slapd_daemon", "Not enough available file descriuptors");
+        exit(1);
+    }
     the_connection_table = connection_table_new(connection_table_size);
 
     /*
@@ -2858,11 +2862,13 @@ get_connection_table_size(void)
     resrvdesc = config_get_reservedescriptors();
     if (maxdesc > resrvdesc) {
          size = (maxdesc - resrvdesc);
+    } else {
+        return 0;
     }
 
-    /* Verify size does not exceed process max fds */
-    if (size > FDS_PROCESS_MAX) {
-        size = (FDS_PROCESS_MAX - resrvdesc);
+    /* Verify size does not exceed max num of conns */
+    if (size > MAX_LDAP_CONNS) {
+        size = (MAX_LDAP_CONNS - resrvdesc);
     }
 
     slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -2868,7 +2868,7 @@ get_connection_table_size(void)
 
     /* Verify size does not exceed max num of conns */
     if (size > MAX_LDAP_CONNS) {
-        size = (MAX_LDAP_CONNS - resrvdesc);
+        size = MAX_LDAP_CONNS;
     }
 
     slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -2865,6 +2865,11 @@ get_connection_table_size(void)
         size = (FDS_PROCESS_MAX - resrvdesc);
     }
 
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+    CFG_LOCK_WRITE(slapdFrontendConfig);
+    slapdFrontendConfig->conntablesize = size;
+    CFG_UNLOCK_WRITE(slapdFrontendConfig);
+
     return size;
 }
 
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/daemon.c
```diff
@@ -2845,16 +2846,23 @@ catch_signals()
...
-     */
-    if (maxdesc >= 0 && size > maxdesc) {
-        size = maxdesc;
+    /* Validate configured reserve descriptors */
+    validate_num_config_reservedescriptors();
+
+    resrvdesc = config_get_reservedescriptors();
+    if (maxdesc > resrvdesc) {
+         size = (maxdesc - resrvdesc);
+    }
```
tbordaz, line 2867: What if maxdesc <= resrvdesc ?

tbordaz (commented):
The change looks good with a minor comment.
A remark, I would appreciate that the fix description states that 'conntablesize' field disappear from the config entry (both cn=config and internal struct). Also do you know the explanation why there is a big slowdown when the nb of established connection is close to the limit ?

jchapma:
> The change looks good with a minor comment. A remark, I would appreciate that the fix description states that 'conntablesize' field disappear from the config entry (both cn=config and internal struct). Also do you know the explanation why there is a big slowdown when the nb of established connection is close to the limit ?

This commit is a backport of a subset of the functionality of  https://github.com/389ds/389-ds-base/pull/5472 that addresses an issue where the default connectiontable size was set too low by default (1024)  and the server failed to accept connections when the connection count was ~950. The bits that were backported were the reserve descriptor calculation based on the servers needs and the auto calculation of the connectiontable size. (conntablesize = max_descriptors - reserve descriptors). I just copied the commit description from the https://bugzilla.redhat.com/show_bug.cgi?id=2210491, but I can amend this commits description if you like.

ldap/servers/slapd/daemon.c
jchapma: > What if maxdesc <= resrvdesc ?

maxdesc is the max number of allowed open file descriptors on the system and resrvdesc is the number of reserve file descriptors for server use (SSL, backends, repl agmts, etc). IMHO maxdesc would never be > resrvdesc, but if you feel a check is required I can add one.

jchapma:
> Also do you know the explanation why there is a big slowdown when the nb of established connection is close to the limit ?

Im guessing that when the server is getting low on file descriptors there might be some kind of system latency causing this, maybe...

rhjyoung:
I filed a Bugzilla with Red Hat on this issue:
                https://bugzilla.redhat.com/show_bug.cgi?id=2210491

Basically once you hit around 950 connections the entire system just slows completely down. So any large enviroment this causes 389ds to slow way to the point of taking up to 1 minute to return a query.

By using work around to that sets the softlimit to the limit. Things work fine.

jchapma:
nsslapd-conntablesize attribute has been added back, setting of this attr is disabled and a warning message returned to the user. The user can still "get" the value of this attr, so rather than return a attr value of 0, I set it in the when the connection table is being created.

ldap/servers/slapd/daemon.c
progier389: That would means that there is a serious configuration issue that requires administrator action
  ( Probably need to  increase the system file descriptors limit, in this case 
      You may do a test log an error as fatal then exit

ldap/servers/slapd/daemon.c
```diff
@@ -2845,18 +2846,30 @@ catch_signals()
...
+    /* Validate configured reserve descriptors */
+    validate_num_config_reservedescriptors();
+
+    resrvdesc = config_get_reservedescriptors();
+    if (maxdesc > resrvdesc) {
+         size = (maxdesc - resrvdesc);
     }
 
+    /* Verify size does not exceed process max fds */
+    if (size > FDS_PROCESS_MAX) {
```
progier389: That seems a bit suspicious as maxdesc is the process max fds.
There is a second limit related to the number of tcpip port that can be open (in therory 65536 -2) in practice I would say 65000 (too keep ssh and other tools connectivity), but it is an hard limit (resrvdesc does not matter at this point)
i.e: it should be if (size > MAX_TCP_PORTS) { size = MAX_TCP_PORTS; }

ldap/servers/slapd/libglobs.c
```diff
@@ -4830,43 +4822,11 @@ config_set_maxdescriptors(const char *attrname, char *value, char *errorbuf, int
...
-                              "Server will use a setting of %d.",
-                              attrname, value, maxVal, maxVal);
-        if (nValue > maxVal) {
-            nValue = maxVal;
-            retVal = LDAP_UNWILLING_TO_PERFORM;
-        } else {
-            retVal = LDAP_OPERATIONS_ERROR;
-        }
-    }
+    slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE,
```
progier389, line 4827: should probably print the value in the warning message

ldap/servers/slapd/daemon.c
jchapma: We used FDS_PROCESS_MAX (64000) in the initial PR, but if you like I can change it to your proposal.

progier389: IMHO you should change the name: FDS_PROCESS_MAX is misleading as you assumed it was file descriptors but
At this point we are no more dealing with the number of file descriptors as it has already been taken in account by  
 maxdesc, we are dealing with the maximum number of connections that may be open, (IMHO the right name is probably MAX_LDAP_CONNS)  and it is an hard limit. Now 64000 or 65000 may be argued, so let use the existing value.

ldap/servers/slapd/daemon.c
```diff
@@ -2845,18 +2850,32 @@ catch_signals()
...
+    resrvdesc = config_get_reservedescriptors();
+    if (maxdesc > resrvdesc) {
+         size = (maxdesc - resrvdesc);
+    } else {
+        return 0;
+    }
+
+    /* Verify size does not exceed max num of conns */
+    if (size > MAX_LDAP_CONNS) {
+        size = (MAX_LDAP_CONNS - resrvdesc);
```
progier389: Should be size = MAX_LDAP_CONNS;   (reserved file descriptors has already been taken in account in previous test)
Everything else seems ok.

progier389 (approved):
With [174c75c] commit (rework2),  it is OK.

Issue #5752, opened 2023-05-03
RFE -  Provide a history for 'LastLoginTime' 


jchapma:
**Description of problem**
  When a user did a successfully bind, the "LastLoginTime" attribute is updated. We have now a request from our security department to display the users last successful bind before the current one. When we just read out this attribute the value is already updated, so that the user did not see his real last successful
login, in fact he sees the current login date and time.

**Describe the solution you'd like**
A history of previous login time stamps

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=1680061

jchapma:
@droideck 
Added support for the disabling of lastloginhistory with a size of 0 and added checks for non positive int argument to lastloginhistory CLI.

Can I push to this issue or should I create a new one ?

droideck:
Extended tests:

   6981d0842..ab9dfcbdb  389-ds-base-2.3 -> 389-ds-base-2.3
   835f12bb7..380b15bd7  389-ds-base-2.2 -> 389-ds-base-2.2

aivanov389:
There is a new side effect of this additional functionality of Account Policy Plugin: it changes **modifyTimestamp** of the user entry after this user successfully binds. The Account Policy Plugin has another parameter (**alwaysrecordlogin**) that does not affect modifyTimestamp when user logs in (that is how it functioned until this RFE). 

We use modifyTimestamp for our synchronization services and after this commit we have a large number of entries that seem changed all the time. After some investigation we have narrowed the problem to this new functionality.

Disabling this new functionality by setting lastLoginHistorySize: 0 still changes the modifyTimestamp of the entry after each successful BIND of the user.

jchapma:
@aivanov389 

I will create a new issue for this and will have a fix tomorrow

jchapma:
5cc25029e..91978a1fd  389-ds-base-2.2 -> 389-ds-base-2.2
ed0093d02..ea04a0454  389-ds-base-2.3 -> 389-ds-base-2.3

PR #5807, opened 2023-06-21
Issue 5752 - RFE - Provide a history for LastLoginTime

jchapma:
Bug Description: For the lastloginhistory feature the user can set the number of login histories that are saved by modifing the lastloginhistorysize attribute. The CLI currently allows setting this attribute value to 0 or a non positive int.

Fix Description: Add support for a lastloginhistorysize of 0 which would disable the feature. Add CLI support for restricting non positive int values.

relates: https://github.com/389ds/389-ds-base/issues/5752

Reviewed by: ?

### DIFF
--- a/ldap/servers/plugins/acctpolicy/acct_config.c
+++ b/ldap/servers/plugins/acctpolicy/acct_config.c
@@ -161,13 +161,13 @@ acct_policy_entry2config(Slapi_Entry *e, acctPluginCfg *newcfg)
                 value = strtoul(hist_size_str, 0, 0);
                 if (value >= 0) {
                     newcfg->login_history_size = value;
-                    slapi_ch_free_string(&hist_size_str);
                 } else {
                     slapi_log_err(SLAPI_LOG_WARNING, PLUGIN_NAME,
                                   "acct_policy_entry2config - Invalid value for login-history-size: %d, "
                                   "Using default value of %d\n", value, DEFAULT_LASTLOGIN_HISTORY_SIZE);
                     newcfg->login_history_size = DEFAULT_LASTLOGIN_HISTORY_SIZE;
                 }
+                slapi_ch_free_string(&hist_size_str);
             }
         } else {
             newcfg->login_history_size = DEFAULT_LASTLOGIN_HISTORY_SIZE;
--- a/ldap/servers/plugins/acctpolicy/acct_config.c
+++ b/ldap/servers/plugins/acctpolicy/acct_config.c
@@ -154,6 +154,7 @@ acct_policy_entry2config(Slapi_Entry *e, acctPluginCfg *newcfg)
         newcfg->login_history_attr = slapi_ch_strdup(LASTLOGIN_HISTORY_ATTR);
         if (has_attr(e, LASTLOGIN_HISTORY_SIZE_ATTR, &hist_size)) {
             newcfg->login_history_size = atoi(hist_size);
+            slapi_ch_free_string(&hist_size);
         } else {
             newcfg->login_history_size = DEFAULT_LASTLOGIN_HISTORY_SIZE;
         }
--- a/ldap/servers/plugins/acctpolicy/acct_config.c
+++ b/ldap/servers/plugins/acctpolicy/acct_config.c
@@ -74,6 +74,7 @@ static int
 acct_policy_entry2config(Slapi_Entry *e, acctPluginCfg *newcfg)
 {
     char *config_val;
+    int value = 0;
     int rc = 0;
 
     if (newcfg == NULL) {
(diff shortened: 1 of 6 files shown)


droideck:
Could you please rebase the PR's branch onto main? (so the newly added tests are included in the CI run)

droideck:
Two things:
1. I think you still miss the validation. Currently it's possible to set `lastLoginHistorySize: five` and it will accept it (and when I do `ldapsearch` it displays `lastLoginHistorySize: five` back).
2. The test I've added is a bit wrong. There's a diff that should fix it (feel free to apply to your PR):  https://drive.google.com/file/d/1X7AM6FGQAALY8J0FaXVjGSuQCQIOEGKm/view?usp=sharing

jchapma:
> 2\. The test I've added is a bit wrong. There's a diff that should fix it (feel free to apply to your PR):  https://drive.google.com/file/d/1X7AM6FGQAALY8J0FaXVjGSuQCQIOEGKm/view?usp=sharing

I added your changes to the test, but I see you are expecting an INVALID_SYNTAX error when using a negative or non int value. I used argparse to validate the args passed to the CLI so no exception is raised, the user is informed that the value is invalid.

droideck:
> > 2. The test I've added is a bit wrong. There's a diff that should fix it (feel free to apply to your PR):  https://drive.google.com/file/d/1X7AM6FGQAALY8J0FaXVjGSuQCQIOEGKm/view?usp=sharing
> 
> I added your changes to the test, but I see you are expecting an INVALID_SYNTAX error when using a negative or non int value. I used argparse to validate the args passed to the CLI so no exception is raised, the user is informed that the value is invalid.

But users do not always use our CLI. They can modify values with ldapmodify or with another way to invoke LDAP protocol call (the way our tests do)

jchapma:
> But users do not always use our CLI. They can modify values with ldapmodify or with another way to invoke LDAP protocol call (the way our tests do)

Thats fair enough, so I have added the int_validate attr syntax check in the mod pre op for this attribute. Regarding the negative int argument, if the accpol detects a negative value it sets the default instead. What do you think ?

droideck:
> Thats fair enough, so I have added the int_validate attr syntax check for this attribute. Regarding the negative int argument, if the accpol plugin detects a negative value it sets the default instead. What do you think ?

In my opinion, it'll be better to stick to the valid range - something like "0 (the feature is disabled) to the maximum 32-bit integer value (2147483647)" and reject other values.

It'll help to have clear and explicitly defined behaviour which follows the same logic as the rest of our config attributes.

But I don't have a very strong opinion here, and I'm open to the discussion if you think otherwise :)

mreynolds389:
> > Thats fair enough, so I have added the int_validate attr syntax check for this attribute. Regarding the negative int argument, if the accpol plugin detects a negative value it sets the default instead. What do you think ?
> 
> In my opinion, it'll be better to stick to the valid range - something like "0 (the feature is disabled) to the maximum 32-bit integer value (2147483647)" and reject other values.
> 
> It'll help to have clear and explicitly defined behaviour which follows the same logic as the rest of our config attributes.
> 
> But I don't have a very strong opinion here, and I'm open to the discussion if you think otherwise :)

Something to keep in mind is that while it's good the client (CLI) checks for bad values/ranges, the server needs to do it too.   You should never rely on the client doing the right thing.  So in acct_policy_load_config_startup() we should do all these checks/adjustments.

jchapma:
> So in acct_policy_load_config_startup() we should do all these checks/adjustments.

The int_validate syntax check takes care of the value type and in the function you mention, if we read a negative int from the config we use the default value instead. This is similar to how the dnaplugin handles int attributes.

droideck:
> The int_validate syntax check takes care of the value type and in the function you mention, if we read a negative int from the config we use the default value instead. This is similar to how the dnaplugin handles int attributes.

I think it's more complicated there as the integers interact with each other a lot. So it's done to avoid unwanted conflicts and crashes. (i.e. https://github.com/389ds/389-ds-base/commit/8f3baf3c8f8168d9fdd387eb3a7a774dc05b41db avoids zero division)

For lastloginhistorysize, I'm not sure we need that. And I think we can safely restrict the values and all be okay.

jchapma:
I think there might be some confusion with the setting of the account policy plugin lastloginhistorysize attribute. In particular how validation of this int attr is managed.

I propose to add support to the server that will return invalid syntax if a non integer value is used, if a negative integer is used, the server will log a warning message to the error logs and use the default value. This is similar to how some of the other plugins manage integer attributes. The link you shared in your last post is an example of this.

I understand you want the server to return invalid syntax for a non integer and negative integer, inferred from the initial tests you added.

droideck:
I still think we should be more restrictive for lastloginhistorysize as opposed to dnaplugin handling, as lastloginhistory is more about security, and it should be explicitly defined by the administrator, in my opinion.

Still, setting the default on a negative value is not so bad, even though I like it less. If other developers agree with your approach, we can proceed with your proposal, as I don't have a very hard opinion in this case.

mreynolds389:
> I still think we should be more restrictive for lastloginhistorysize as opposed to dnaplugin handling, as lastloginhistory is more about security, and it should be explicitly defined by the administrator, in my opinion.
> 
> Still, setting the default on a negative value is not so bad, even though I like it less. If other developers agree with your approach, we can proceed with your proposal, as I don't have a very hard opinion in this case.

I think what we need is have validation like what is used in libglobs.c (e.g. config_set_num_listeners()).  Where we make sure it's a "number", the number is not too large or small, and fallback to some default value is it is invalid.

ldap/servers/plugins/acctpolicy/acct_config.c
```diff
@@ -150,10 +151,24 @@ acct_policy_entry2config(Slapi_Entry *e, acctPluginCfg *newcfg)
...
+            if (hist_size_str) {
+                value = strtoul(hist_size_str, 0, 0);
+                if (value >= 0) {
+                    newcfg->login_history_size = value;
+                    slapi_ch_free_string(&hist_size_str);
+                } else {
+                    slapi_log_err(SLAPI_LOG_WARNING, PLUGIN_NAME,
+                                  "acct_policy_entry2config - Invalid value for login-history-size: %d, "
+                                  "Using default value of %d\n", value, DEFAULT_LASTLOGIN_HISTORY_SIZE);
+                    newcfg->login_history_size = DEFAULT_LASTLOGIN_HISTORY_SIZE;
```
droideck, line 168: I think we should also free `hist_size_str` here.

droideck (commented):
Okay, besides the minor issue noted below, and as long as another dev will set 'approved', - ack from me! :)

mreynolds389 (approved):

mreynolds389:
@jchapma - can you please merge this I need to do an upstream build, thanks!

Issue #5729, opened 2023-04-14
Memory leak in factory_create_extension


jchapma:
**Issue Description**
The functionality to extend an operation object introduces a memory leak

**Package Version and Platform:**
 - Package and version:  389-Directory/2.3.2 (main branch)

**Steps to Reproduce**
1. Run the sync repl cookie add del CI tests - LSAN_OPTIONS=exitcode=0:log_threads=1:verbosity=1 PYINSTALL=1 py.test -sv dirsrvtests/tests/suites/syncrepl_plugin/basic_test.py::test_sync_repl_cookie_add_del
2. Check asan output
```
Indirect leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x7f3c1aa8f7f7 in __interceptor_calloc (/lib64/libasan.so.6+0xb07f7)
    #1 0x7f3c1a26d29d in slapi_ch_calloc ldap/servers/slapd/ch_malloc.c:175
    #2 0x7f3c1a2c1716 in factory_create_extension ldap/servers/slapd/factory.c:320
    #3 0x437c09 in connection_add_operation ldap/servers/slapd/connection.c:2086
    #4 0x4371ae in connection_activity ldap/servers/slapd/connection.c:1941
    #5 0x444b93 in handle_pr_read_ready ldap/servers/slapd/daemon.c:1621
    #6 0x442ede in ct_list_thread ldap/servers/slapd/daemon.c:1306
    #7 0x7f3c19442553 in _pt_root ../../.././nspr/pr/src/pthreads/ptthread.c:201
    #8 0x7f3c193d4431 in start_thread /usr/src/debug/glibc-2.31-74-gd0c84d22b6/nptl/pthread_create.c:477
    #9 0x7f3c19cae6d2 in __clone (/lib64/libc.so.6+0x1016d2)

Indirect leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x7f3c1aa8f7f7 in __interceptor_calloc (/lib64/libasan.so.6+0xb07f7)
    #1 0x7f3c1a26d29d in slapi_ch_calloc ldap/servers/slapd/ch_malloc.c:175
    #2 0x7f3c1a34f43f in op_stat_constructor ldap/servers/slapd/operation.c:690
    #3 0x7f3c1a2c18ac in factory_create_extension ldap/servers/slapd/factory.c:328
    #4 0x437c09 in connection_add_operation ldap/servers/slapd/connection.c:2086
[...]

PR #5814, opened 2023-06-27
Issue 5729 - Memory leak in factory_create_extension

jchapma:
Bug description: Mem leak in sync repl operation extension code. In syn persist when we release the connection, the operation extension is not free'd.

Fix description: In syn persist when we release the connection, free the operation extension if it exists.

relates: https://github.com/389ds/389-ds-base/issues/5729

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -1084,7 +1084,7 @@ done:
 
     slapi_ch_free((void **)&req->req_orig_base);
     slapi_filter_free(req->req_filter, 1);
-    sync_cookie_free(&req->req_cookie);
+
     for (qnode = req->ps_eq_head; qnode; qnode = qnodenext) {
         qnodenext = qnode->sync_next;
         sync_node_free(&qnode);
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -1084,7 +1084,7 @@ sync_send_results(void *arg)
 
     slapi_ch_free((void **)&req->req_orig_base);
     slapi_filter_free(req->req_filter, 1);
-    sync_cookie_free(&req->req_cookie);
```
droideck, line 1087: I'm not very familiar with this area of code... Could you please explain why we don't need to `sync_cookie_free` here?

progier389: As I understand it is because its leads to a double free as the cookie is created and freed by the caller function.

jchapma: Yes, thats exactly it. The destructor for the sync operation extension frees the cookie so we get a double free error from asan.

progier389 (approved):
LGTM

PR #5815, opened 2023-06-28
Issue 1081 - CI - Add more tests for overwriting x-origin issue

droideck:
Description: Add test suite that tests an attributetype and its x-origin values in a replicated environment s1c1 and s1h1c1. Also, make sure that custom x-origin is correctly overwritten in the replication event as 'user defined'.

Related: https://github.com/389ds/389-ds-base/issues/1081

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/schema/schema_replication_origin_test.py
+++ b/dirsrvtests/tests/suites/schema/schema_replication_origin_test.py
@@ -14,7 +14,7 @@ import ldap
 import pytest
 from lib389._constants import SUFFIX, ReplicaRole, DEFAULT_SUFFIX
 from lib389.topologies import create_topology
-from lib389.replica import Agreements
+from lib389.replica import Agreements, ReplicationManager
 from lib389.schema import Schema
 from lib389.idm.user import UserAccounts
 from lib389.cli_base import LogCapture
@@ -177,6 +177,7 @@ def test_schema_xorigin_repl(topology, schema_replication_init, xorigin):
         7. The startup and final state of the schema replication process should be as expected
     """
 
+    repl = ReplicationManager(DEFAULT_SUFFIX)
     user = schema_replication_init
     hub = None
     supplier = topology['topology'].ms["supplier1"]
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/schema/schema_replication_origin_test.py
```diff
@@ -0,0 +1,237 @@
...
+def trigger_update(topology, user_rdn, num):
+    """It triggers an update on the supplier. This will start a replication
+    session and a schema push
+    """
+
+    users_s = UserAccounts(topology.ms["supplier1"], DEFAULT_SUFFIX)
+    user = users_s.get(user_rdn)
+    user.replace('telephonenumber', str(num))
+
+    # wait 10 seconds that the update is replicated
```
progier389: I would change the comment to: wait until the update is replicated (until up to x seconds)
IMHO using  "for loop in range(30):"   is more pythonic
And I would rather wait up to 30s because we had to increase the replication timeout in a lot of testcases.
 10s is very often working fine but from time to time the test vm is really crawling and we get timeout while nothing is broken ...

dirsrvtests/tests/suites/schema/schema_replication_origin_test.py
```diff
@@ -0,0 +1,234 @@
...
+    add_attributetype(supplier, 1, 'testAttribute', xorigin)
+
+    # Search for attribute with JSON option
+    schema = Schema(supplier)
+    attr_result = schema.query_attributetype('testAttribute', json=True)
+    # Verify the x-origin value is correct
+    assert attr_result['at']['x_origin'][0] == "user defined"
+
+    trigger_schema_push(topology, user.rdn, 1)
+    time.sleep(5)
```
progier389: I am always wary about time.sleep in CI test (I have noticed that sometime the VM are frozen for up to 20 seconds).  IMHO in this case you should wait for replication between supplier and consumer

dirsrvtests/tests/suites/schema/schema_replication_origin_test.py
```diff
@@ -0,0 +1,234 @@
...
+        assert False
+
+    # add a new OC on the supplier so that its nsSchemaCSN is larger than the consumer (wait 2s)
+    add_attributetype(consumer, 2, 'testAttributeCA', xorigin)
+    time.sleep(2)
+    add_attributetype(supplier, 3, 'testAttributeSA', xorigin)
+
+    # now push the scheam
+    trigger_schema_push(topology, user.rdn, 2)
+    time.sleep(5)
```
progier389: Same comment than above

progier389 (approved):
LGTM

Issue #4719, opened 2021-04-08
dsconf fails to add a PTA URL due to an invalid check.


mreynolds389:
When trying to configure the PTA plugin by adding an URL, the dsconf command fails.

Version-Release number of selected component (if applicable):
# cat /etc/redhat-release
Red Hat Enterprise Linux release 8.3 (Ootpa)
#
# rpm -qa | grep 389-ds-base-1
389-ds-base-1.4.3.13-1.module+el8dsrv+8334+69a46a2e.x86_64
#

How reproducible:
Always.

Steps to Reproduce:
Launch the command in verbose mode. It fails since it considers the DN as invalid:

# dsconf --verbose -D "cn=Directory Manager" ldap://localhost:10389 plugin pass-through-auth url add "ldap://localhost:7389/o=redhat"

...

DEBUG: Subtree is an invalid DN
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 134, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.6/site-packages/lib389/cli_conf/plugins/passthroughauth.py", line 88, in pta_add
    new_url_l = _validate_url(args.URL.lower())
  File "/usr/lib/python3.6/site-packages/lib389/cli_conf/plugins/passthroughauth.py", line 58, in _validate_url
    raise ValueError("Subtree is an invalid DN")
ValueError: Subtree is an invalid DN
ERROR: Error: Subtree is an invalid DN

https://bugzilla.redhat.com/show_bug.cgi?id=1947469

QUASARFREAK:
I'm having this exact same issue:

# cat /etc/redhat-release
Red Hat Enterprise Linux release 8.1 (Ootpa)

# rpm -qa | grep 389-ds-base-1
389-ds-base-1.4.3.21-3.module+el8dsrv+10401+3d549418.x86_64

# dsconf --verbose -D "cn=Directory Manager" ldap://localhost:10389 plugin pass-through-auth url add ldap://ldaporig:10389/ou=People,c=arg,O=BBVA
DEBUG: The 389 Directory Server Configuration Tool
DEBUG: Inspired by works of: ITS, The University of Adelaide
DEBUG: dsrc path: /root/.dsrc
DEBUG: dsrc container path: /data/config/container.inf
DEBUG: dsrc instances: []
DEBUG: dsrc no such section: slapd-ldap://localhost:10389
[...]

PR #5823, opened 2023-07-03
Issue 4719 - CI - Add dsconf add a PTA URL test

v-cech:
Description: This test checks that you are able to add a PTA URL through dsconf.

Relates: https://github.com/389ds/389-ds-base/issues/4719

Reviewed by: ???

### DIFF
--- a/src/lib389/lib389/cli_conf/plugins/ldappassthrough.py
+++ b/src/lib389/lib389/cli_conf/plugins/ldappassthrough.py
@@ -88,6 +88,7 @@ def pta_add(inst, basedn, log, args):
     if new_url_l in urls:
         raise ldap.ALREADY_EXISTS("Entry %s already exists" % args.URL)
     plugin.add("nsslapd-pluginarg%s" % next_num, args.URL)
+    log.info("Successfully added URL")
 
 
 def pta_edit(inst, basedn, log, args):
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/cli_conf/plugins/ldappassthrough.py
```diff
@@ -88,6 +88,7 @@ def pta_add(inst, basedn, log, args):
     if new_url_l in urls:
         raise ldap.ALREADY_EXISTS("Entry %s already exists" % args.URL)
     plugin.add("nsslapd-pluginarg%s" % next_num, args.URL)
+    log.info("Command successful.")
```
mreynolds389: I think this should log something like this: "Successfully added URL".  This is how most of the other "success" messages are logged  "Command successful" is not used any where else, so we should try and be consistent with other CLI  success messages.

But otherwise looks good!  Ack

mreynolds389 (approved):

droideck (approved):
I think the commit subject is a bit misleading, and it should be something like this:

`Issue 4719 - CI - Add dsconf add a PTA URL test` (and in the commit body, we can mention the result message change.)

Otherwise, looks good!

bsimonova (approved):
LGTM

Issue #5825, opened 2023-07-05
healthcheck - password storage scheme warning is lacking information


mreynolds389:
**Issue Description**
 
When password storage scheme for root DN, or users, is not secure we do not specify which setting is not secure, and we do not list what the insecure scheme is.

mreynolds389:
https://bugzilla.redhat.com/show_bug.cgi?id=2220890

mreynolds389:
e76bfe1d0..6b9a86cc1  389-ds-base-2.3 -> 389-ds-base-2.3
b674879e9..94d69c7d0  389-ds-base-2.2 -> 389-ds-base-2.2
4b25163e8..e377c9af1  389-ds-base-2.1 -> 389-ds-base-2.1
b2784617a..09e984337  389-ds-base-2.0 -> 389-ds-base-2.0
71544acee..dc4130da5  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #5826, opened 2023-07-05
Issue 5825 - healthcheck - password storage scheme warning needs more info

mreynolds389:
Description:  

Add the current/insecure scheme to the report, and state which config setting is insecure.

relates: https://github.com/389ds/389-ds-base/issues/5825


jchapma (approved):
Tried out this PR and works fine for me, ack

src/lib389/lib389/config.py
```diff
@@ -217,7 +217,18 @@ def _lint_passwordscheme(self):
         u_root_scheme = self.get_attr_val_utf8('nsslapd-rootpwstoragescheme')
         if u_root_scheme not in allowed_schemes or u_password_scheme not in allowed_schemes:
```
mreynolds389: This needs to be fixed.  We should only checking u_root here.  Update coming...

droideck (approved):
LGTM!

PR #5827, opened 2023-07-06
Issue 5793 - UI - Fix minor crashes

droideck:
Description: After a massive move from webpack to esbuild bundler rework, fix two minor crashes which happened because of minor copy-paste errors.

Related: https://github.com/389ds/389-ds-base/issues/5793

Reviewed by: ?


mreynolds389 (approved):
ACK!

PR #5829, opened 2023-07-07
Bump version to 2.4.2

mreynolds389:


droideck (approved):
Ack!

One thing, though. In case you plan to do a build, be aware that the fix for this regression https://github.com/389ds/389-ds-base/pull/5807 is not merged yet (it needs an ack from another dev - just to confirm that we want to go with the suggested approach)

mreynolds389:
> Ack!
> 
> One thing, though. In case you plan to do a build, be aware that the fix for this regression #5807 is not merged yet (it needs an ack from another dev - just to confirm that we want to go with the suggested approach)

Ugh, ok, I'll wait

PR #5838, opened 2023-07-13
Issue 4551 - Paged search impacts performance

### LINKED ISSUE #4551: Paged search impacts performance
**Issue Description**
When a server is under search load, a paged search from a regular user impacts the performance. 
Same paged search requested by DM has much smaller impact. 

 - Platform: Fedora 33
 - Package and version: 389-ds-base-1.4.4.9-1.fc33.x86_64. 

**Steps to Reproduce**
1. Create a db with 40000 users
2. Run search load using  `ldclt -D "uid=test,ou=people,dc=example,dc=com" -w test -e bindeach,esearch -b "ou=people,dc=example,dc=com" -f "uid=00001"`
3. While the search load is running, run paged search in a loop that requests all ids: 
`while : ; do ldapsearch -D "uid=test,ou=people,dc=example,dc=com" -w test -b dc=example,dc=com 'uid=*'  -E pr=100/noprompt; done `

**Actual results**
```
ldclt[1353]: Average rate: 2415.70/thr  (2415.70/sec), total:  24157 -- only ldclt is running 
ldclt[1353]: Average rate: 2342.50/thr  (2342.50/sec), total:  23425
ldclt[1353]: Average rate: 1048.90/thr  (1048.90/sec), total:  10489 -\
ldclt[1353]: Average rate:  413.10/thr  ( 413.10/sec), total:   4131   | paged search from a regular user
ldclt[1353]: Average rate:  461.00/thr  ( 461.00/sec), total:   4610   |
ldclt[1353]: Average rate: 1759.30/thr  (1759.30/sec), total:  17593 -/
ldclt[1353]: Average rate: 2374.70/thr  (2374.70/sec), total:  23747
ldclt[1353]: Average rate: 1952.70/thr  (1952.70/sec), total:  19527 -\
ldclt[1353]: Average rate: 1783.00/thr  (1783.00/sec), total:  17830   | paged search from DM
ldclt[1353]: Average rate: 1749.70/thr  (1749.70/sec), total:  17497 -/
ldclt[1353]: Average rate: 2319.80/thr  (2319.80/sec), total:  23198
ldclt[1353]: Average rate: 2378.80/thr  (2378.80/sec), total:  23788
```

**Expected results**
There shouldn't be a significant drop in performance.

**Additional information**
@wisebaldone who reported the issue, also mentioned that 1.2.11.15-48 doesn't have the issue and 1.2.11.15-97 does have.

progier389:
Problem: 
    Having a script looping doing a search with paged result impact greatly the performance of other clients 
      (for example ldclt bind+search rate decreased by 80% in the test case)

Cause:
    Page result field in connection were protected by the connection mutex that is also used by the listener thread, in some cases this cause contention that delays the handling of new operations

Solution:
     Do not rely on the connection mutex to protect the page result context but on a dedicated array of locks. 
 
 Issue: [4551](https://github.com/389ds/389-ds-base/issues/4551)

 Reviewed by: @Firstyear  Thanks !

### DIFF
--- a/ldap/servers/slapd/opshared.c
+++ b/ldap/servers/slapd/opshared.c
@@ -696,7 +696,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
              * In async paged result case, the search result might be released
              * by other theads.  We need to double check it in the locked region.
              */
-            pthread_mutex_lock(&(pb_conn->c_mutex));
+            pthread_mutex_lock(pageresult_lock_get_addr(pb_conn));
             pr_search_result = pagedresults_get_search_result(pb_conn, operation, 1 /*locked*/, pr_idx);
             if (pr_search_result) {
                 if (pagedresults_is_abandoned_or_notavailable(pb_conn, 1 /*locked*/, pr_idx)) {
@@ -704,7 +704,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                     /* Previous operation was abandoned and the simplepaged object is not in use. */
                     send_ldap_result(pb, 0, NULL, "Simple Paged Results Search abandoned", 0, NULL);
                     rc = LDAP_SUCCESS;
-                    pthread_mutex_unlock(&(pb_conn->c_mutex));
+                    pthread_mutex_unlock(pageresult_lock_get_addr(pb_conn));
                     goto free_and_return;
                 } else {
                     slapi_pblock_set(pb, SLAPI_SEARCH_RESULT_SET, pr_search_result);
@@ -718,7 +718,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                 pr_stat = PAGEDRESULTS_SEARCH_END;
                 rc = LDAP_SUCCESS;
             }
-            pthread_mutex_unlock(&(pb_conn->c_mutex));
+            pthread_mutex_unlock(pageresult_lock_get_addr(pb_conn));
             pagedresults_unlock(pb_conn, pr_idx);
 
             if ((PAGEDRESULTS_SEARCH_END == pr_stat) || (0 == pnentries)) {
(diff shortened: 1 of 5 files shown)


ldap/servers/slapd/opshared.c
```diff
@@ -696,15 +696,15 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
              * In async paged result case, the search result might be released
              * by other theads.  We need to double check it in the locked region.
              */
-            pthread_mutex_lock(&(pb_conn->c_mutex));
+            pthread_mutex_lock(pageresult_lock_get_addr(pb_conn));
```
Firstyear: micro-optimisation - keep a reference to the lock pointer from this point to re-use lower to avoid the extra dereferences.

Firstyear (commented):
This seems like a really good improvement, I wonder if this affects SSSD as well since that is a frequent user of paged searches.

progier389:
> This seems like a really good improvement, I wonder if this affects SSSD as well since that is a frequent user of paged searches.

It probably does. FYI: Thierry has also recently seen this issue in a real production environment blocking the whole server until nsslapd-ioblocktimeout expired.

ldap/servers/slapd/opshared.c
progier389: Done.

Firstyear (approved):

Firstyear:
> > This seems like a really good improvement, I wonder if this affects SSSD as well since that is a frequent user of paged searches.
> 
> It probably does. FYI: Thierry has also recently seen this issue in a real production environment blocking the whole server until nsslapd-ioblocktimeout expired.

Okay, given I see this is targeted to 1.4.3 I guess we plan to backport it all the way down?

progier389:
> Okay, given I see this is targeted to 1.4.3 I guess we plan to backport it all the way down?
Indeed I plan to cherry-pick it down to 1.4.3

PR #5847, opened 2023-07-20
Issue 5825 - healthcheck - update allowed password schemes for fips mode (1.4.3 only)

mreynolds389:
Description:

In 1.4.3 in fips mode should allow SSHA512 as not all 1.4.3 released versions support the new Rust password hashers.

relates: https://github.com/389ds/389-ds-base/issues/5825


droideck (approved):
LGTM

Issue #5848, opened 2023-07-21
dsconf should prevent setting the replicaID for hub and consumer roles


kimettog:
**Issue Description**
dsconf accepts the "replica-id" option when setting a hub or a consumer.
The replica configuration entry is correctly created ( replicaID is set to 65535 ).

It would nice to prevent users setting the replicaID for hub and consumer roles because the value is set automatically anyway.
Customers could try creating hubs and consumers using a replicaID different of 65535.

**Package Version and Platform:**
 - Platform: RHEL
 - Package and version: 389-ds-base-1.4.3.22-1.module+el8dsrv+10501+8ce33e95.x86_64

**Steps to Reproduce**
1. Try to create a consumer using a replicaID different of 65535:
$ dsconf -v -D "cn=Directory Manager" ldap://localhost:1389 replication enable --suffix="dc=test,dc=com" --role="consumer"  --replica-id=1

2. Check the actual configured value:
$ dsconf -D "cn=Directory Manager" ldap://localhost:1389 replication get --suffix="dc=test,dc=com" | grep -i replicaIDEnter password for cn=Directory Manager on ldap://localhost:1389:
nsDS5ReplicaId: 65535
Actual results:
One can pass the parameter "replica-id" to dsconf when creating hubs or consumers.

**Expected results**
Don't allow to set the "replica-id" manually for hubs and consumers.

**Additional context**
$ dsconf -v -D "cn=Directory Manager" ldap://localhost:1389 replication enable --suffix="dc=test,dc=com" --role="consumer"  --replica-id=1
DEBUG: The 389 Directory Server Configuration Tool
...
DEBUG: Checking "None" under None : {'cn': 'changelog5', 'nsslapd-changelogdir': '/var/lib/dirsrv/slapd-alps9/changelogdb'}
DEBUG: Validated dn cn=changelog5,cn=config
DEBUG: Creating cn=changelog5,cn=config
DEBUG: updating dn: cn=changelog5,cn=config
DEBUG: updated dn: cn=changelog5,cn=config with {'objectclass': [b'top', b'nsChangelogConfig']}
DEBUG: updating dn: cn=changelog5,cn=config
DEBUG: updated dn: cn=changelog5,cn=config with {'cn': [b'changelog5'], 'nsslapd-changelogdir': [b'/var/lib/dirsrv/slapd-alps9/changelogdb']}
[...]

PR #5849, opened 2023-07-21
Issue 5848 - dsconf should prevent setting the replicaID for hub and consumer roles

kimettog:
Bug Description:

dsconf accepts the "replica-id" option when setting a hub or a consumer. The replica configuration entry is correctly created ( replicaID is set to 65535 ). we should  prevent users setting the replicaID for hub and consumer roles because the value is set automatically anyway.

Fix Description:
Check if role is "consumer" or "hub" and if so deny option to set the ReplicaID

relates: https://github.com/389ds/389-ds-base/issues/5848

Author: Gilbert Kimetto

Reviewed by: ?? Simon Pichugi

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsconf_test.py
+++ b/dirsrvtests/tests/suites/clu/dsconf_test.py
@@ -8,17 +8,20 @@
 #
 """Test dsconf CLI with LDAPS"""
 
-import ldap
-import logging
-import pytest
 import subprocess
+import logging
 import os
-from lib389._constants import DEFAULT_SUFFIX, DN_DM
-from lib389.topologies import topology_st
+from lib389.cli_base import LogCapture
+import pytest
+import ldap
+from lib389._constants import DEFAULT_SUFFIX, DN_DM, ReplicaRole
+from lib389.topologies import create_topology
+
 
 pytestmark = pytest.mark.tier1
 
 log = logging.getLogger(__name__)
+
 
 @pytest.fixture(scope="function")
 def enable_config(request, topology_st, config_type):
@@ -43,6 +46,15 @@ def enable_config(request, topology_st, config_type):
 
     request.addfinalizer(fin)
 
+
+@pytest.fixture(scope="function")
+def topology_st(request):
+    """Create DS standalone instance"""
+
+    topology = create_topology({ReplicaRole.STANDALONE: 1})
+
+    topology.logcap = LogCapture()
+    return topology
 
 def test_backend_referral(topology_st):
     """Test setting and deleting referral in backend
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -167,3 +167,36 @@ def test_dsconf_with_ldaps(topology_st, enable_config, config_type):
...
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Setting the "replica-id" manually for consumers not allowed.
+        4. Success
+        5. Setting the "replica-id" manually for hubs is not allowed.
+    """
+    print("DN_DM {}".format(DN_DM))
+    cmdline=['/usr/sbin/dsconf', 'standalone1', '-D',
+                 DN_DM, '-w', 'password', 'replication', 'enable', '--suffix',DEFAULT_SUFFIX,'--role',instance_role, '--replica-id=1' ]
```
mreynolds389: Please fix indentation and spacing around commas.  Perhaps run pep8 or flake to check the code syntax.

src/lib389/lib389/cli_conf/replication.py
```diff
@@ -183,6 +183,13 @@ def enable_replication(inst, basedn, log, args):
         # rid is good add it to the props
         repl_properties['nsDS5ReplicaId'] = args.replica_id
 
+    # Validate consumer and hub settings
+    elif role == "consumer" or role == "hub":
+        # Check Replica ID
+        if args.replica_id is not None or args.replica_id != 65535:
+            #Error, Replica ID cannot be specified for consumer and hub roles
```
mreynolds389: Should have a space after the "#"

dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -167,3 +167,36 @@ def test_dsconf_with_ldaps(topology_st, enable_config, config_type):
...
     assert proc.returncode == 0
+
+@pytest.mark.parametrize('instance_role', ('consumer','hub'))
+def test_check_replica_id_rejected (topology_st, instance_role):
+    """Test dsconf CLI does not accept replica-id parameter for comsumer and hubs
+
+    :id: 274b47f8-111a-11ee-8321-98fa9ba19b65
+    :parametrized: yes
+    :customerscenario: True
+    :setup: 3 instance(s)
```
droideck: Typo, as it's only 1 instance now:)

dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -167,3 +167,36 @@ def test_dsconf_with_ldaps(topology_st, enable_config, config_type):
...
+    :steps:
+        1. Create ldap instance with 3 instances
```
droideck: Same here

dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -167,3 +167,36 @@ def test_dsconf_with_ldaps(topology_st, enable_config, config_type):
     msg = proc.communicate()
     log.info(f'output message : {msg[0]}')
     assert proc.returncode == 0
+
+@pytest.mark.parametrize('instance_role', ('consumer','hub'))
```
droideck: You need to add an override topology function - `@pytest.fixture(scope="function")` - `topology_st` to this test suite for the parametrization to work correctly.

```
def topology_st(request):
    """Create DS standalone instance"""

    topology = create_topology({ReplicaRole.STANDALONE: 1}, request=request)

    topology.logcap = LogCapture()
    return topology
 ```

**Review by droideck — CHANGES_REQUESTED** [review]

File: dirsrvtests/tests/suites/clu/dsconf_test.py (hunk 1 above)
Comment (kimettog) [reply dirsrvtests/tests/suites/clu/dsconf_test.py]: Thank you! I fixed the indentation and I will also check any other syntax/formatting I could have missed.

File: src/lib389/lib389/cli_conf/replication.py (hunk 2 above)
Comment (kimettog) [reply src/lib389/lib389/cli_conf/replication.py]: Added a space and checked for other formatting thank you!

File: dirsrvtests/tests/suites/clu/dsconf_test.py (hunk 3 above)
Comment (kimettog) [reply dirsrvtests/tests/suites/clu/dsconf_test.py]: Thank you fixed the comment to show creating once instance

File: dirsrvtests/tests/suites/clu/dsconf_test.py (hunk 4 above)
Comment (kimettog) [reply dirsrvtests/tests/suites/clu/dsconf_test.py]: Thank you fixed the comment to show creating once instance

File: dirsrvtests/tests/suites/clu/dsconf_test.py (hunk 5 above)
Comment (kimettog) [reply dirsrvtests/tests/suites/clu/dsconf_test.py]: added the new fixture/function and tested ...also imported LogCapture Thanks for the tip!

**Review by droideck — APPROVED** [review]
Awesome! LGTM!

File: src/lib389/lib389/cli_conf/replication.py [hunk 6]
```diff
@@ -183,6 +183,13 @@ def enable_replication(inst, basedn, log, args):
         # rid is good add it to the props
         repl_properties['nsDS5ReplicaId'] = args.replica_id
 
+    # Validate consumer and hub settings
+    elif role == "consumer" or role == "hub":
+        # Check Replica ID
+        if args.replica_id is not None or args.replica_id != 65535:
```
vashirov, line 189: This introduced a regression: I can no longer create consumer or hub, it fails with:
```
/usr/sbin/dsconf -j ldapi://%2fvar%2frun%2fslapd-M2.socket replication enable --suffix=dc=example,dc=com --role=consumer --bind-dn="cn=replication manager,cn=config"
{
    "desc": "Replica ID cannot be specified for consumer and hub roles"
}
```

I think `or` should be replaced with `and`, because if `args.replica_id` is `None`, the second part of the condition will evaluate to `True`:
```
(Pdb) print(args.replica_id)
None
(Pdb) print(args.replica_id != 65535)
True
(Pdb) print(args.replica_id is not None or args.replica_id != 65535)
True
```

tbordaz: I agree, the 'or' should be replaced with a 'and'. The only allowed value is 65535.

Issue #5850, opened 2023-07-24
Error: name 'args_instance' is not defined" when dsconf <instance> replication monitor


definite:
**Issue Description**
I ran:

~~~
dsconf -v -D 'cn=directory manager' ldap://localhost replication monitor
~~~

And input credentials like the following:

Enter password for cn=Directory Manager on ldap://localhost

Enter a bind DN for replica.example.com:389: cn=directory manager
Enter a password for cn=directory manager on replica.example.com:389: 

Then
~~~
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 138, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.9/site-packages/lib389/cli_conf/replication.py", line 422, in get_repl_monitor_info
    report_dict = repl_monitor.generate_report(get_credentials, args.json)
  File "/usr/lib/python3.9/site-packages/lib389/replica.py", line 2708, in generate_report
    args_instance[SER_HOST] = supplier_hostname
NameError: name 'args_instance' is not defined
ERROR: Error: name 'args_instance' is not defined
~~~

**Package Version and Platform:**
 - Platform: RHEL9
 - Package and version: 389-ds-base-2.2.4-3.el9.x86_64

**Steps to Reproduce**
0. Setup environment with 389-ds replicas
1. dsconf -v -D 'cn=directory manager' ldap://localhost replication manager
2. Input credentials

Following error shown:
~~~
Traceback (most recent call last):
  File "/usr/sbin/dsconf", line 138, in <module>
    result = args.func(inst, None, log, args)
  File "/usr/lib/python3.9/site-packages/lib389/cli_conf/replication.py", line 422, in get_repl_monitor_info
    report_dict = repl_monitor.generate_report(get_credentials, args.json)
  File "/usr/lib/python3.9/site-packages/lib389/replica.py", line 2708, in generate_report
    args_instance[SER_HOST] = supplier_hostname
NameError: name 'args_instance' is not defined
ERROR: Error: name 'args_instance' is not defined
~~~

**Expected results**
No error

**Additional context**
Workaround: edit replica.py, and add args_instance to lib389._constants import like:

~~~
from lib389._constants import CONSUMER_REPLICAID, REPLICA_RDWR_TYPE, REPLICA_FLAGS_WRITE, REPLICA_RDONLY_TYPE, \
[...]

PR #5851, opened 2023-07-24
fix(issue5850): add args_instance to lib389._constance import

definite:
Fixes issue #5850 by adding args_instance to lib389._constance import


droideck:
Should we close it? As I see the discussion at https://github.com/389ds/389-ds-base/issues/5850 stated that the issue is already fixed.

Issue #5853, opened 2023-07-24
The tracking issue for cargo update commits


droideck:
**Issue Description**
New vulnerabilities, compile warnings, etc. can arise from time to time in Rust packages.
They should be addressed by running `cargo update --manifest-path=./src/Cargo.toml`. Sometimes it can require manual intrusion.

The PRs can be linked to this issue.

droideck:
3b43a7b23..bd3140f24  389-ds-base-2.3 -> 389-ds-base-2.3
   5e4956428..8d74381b0  389-ds-base-2.2 -> 389-ds-base-2.2
   98b5df6ed..86c736baa  389-ds-base-2.1 -> 389-ds-base-2.1
   3470b8e38..744cb41df  389-ds-base-2.0 -> 389-ds-base-2.0
   7e07c9510..e0ac93b75  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

vashirov:
06803f90a..4900884f4  389-ds-base-1.4.3 -> 389-ds-base-1.4.3
   d1102569e..75407006a  389-ds-base-2.0 -> 389-ds-base-2.0
   f5d92ef39..085d4faa7  389-ds-base-2.1 -> 389-ds-base-2.1
   a9ae3f9df..70a8f3fcb  389-ds-base-2.2 -> 389-ds-base-2.2
   e7f711922..830900f0a  389-ds-base-2.3 -> 389-ds-base-2.3
   2bb120ee5..8617d2f6b  389-ds-base-2.4 -> 389-ds-base-2.4
   de5a7dfdc..152fbe5ea  389-ds-base-2.5 -> 389-ds-base-2.5
   5c17aca02..7f6266dc1  389-ds-base-3.0 -> 389-ds-base-3.0

PR #5854, opened 2023-07-24
Issue 5853 - Update Cargo.lock and fix minor warning

### LINKED ISSUE #5861: FTBFS with rust-1.71.0
**Issue Description**

Reported in https://bugzilla.redhat.com/show_bug.cgi?id=2225532

```
error: using `.borrow()` on a double reference, which returns `&concread::cowcell::CowCellReadTxn<CacheStats>` instead of borrowing the inner type
  --> librslapd/src/cache.rs:60:28
   |
60 |     let stats = stat_rguard.borrow();
   |                            ^^^^^^^^^
   |
note: the lint level is defined here
  --> librslapd/src/lib.rs:1:9
   |
1  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(suspicious_double_ref_op)]` implied by `#[deny(warnings)]`
error: could not compile `librslapd` (lib) due to previous error

```

```
error: variable does not need to be mutable
   --> slapi_r_plugin/src/value.rs:185:13
    |
185 |         let mut v = unsafe { slapi_value_new() };
    |             ----^
    |             |
    |             help: remove this `mut`
    |
note: the lint level is defined here
   --> slapi_r_plugin/src/lib.rs:1:9
    |
1   | #![deny(warnings)]
    |         ^^^^^^^^
    = note: `#[deny(unused_mut)]` implied by `#[deny(warnings)]`
error: could not compile `slapi_r_plugin` (lib) due to previous error
```

droideck:
Description: Run cargo update --manifest-path=./src/Cargo.toml.
Add minimum supported rust version field to the manifests.
Fix minor 'variable does not need to be mutable' error.

Another error:
error: using `.borrow()` on a double reference, which returns
`&concread::cowcell::CowCellReadTxn<CacheStats>` instead of borrowing the inner type
    
We're getting the error about borrowing a double reference because
we're trying to borrow a type that is already a reference.
Fix - use the type directly.
    
Related: https://github.com/389ds/389-ds-base/issues/5853
Fixes: https://github.com/389ds/389-ds-base/issues/5861
    
Reviewed by: ?

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "ahash"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -28,7 +43,7 @@ version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
- "hermit-abi 0.1.19",
+ "hermit-abi",
  "libc",
  "winapi",
 ]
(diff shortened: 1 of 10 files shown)


vashirov (approved):
LGTM!

vashirov:
I think we need this changes backported to at least 2.2 (F37)

droideck:
> I think we need this changes backported to at least 2.2 (F37)

Agree.

@Firstyear could you please check it out in case I missed something with this new Rust version (and compact with old) :)

Firstyear:
You should set the MSRV in Cargo.toml, that way the compiler will self-detect when it's too old and warn for distributors/platforms. See https://doc.rust-lang.org/beta/cargo/reference/manifest.html#the-rust-version-field

droideck:
> You should set the MSRV in Cargo.toml, that way the compiler will self-detect when it's too old and warn for distributors/platforms. See https://doc.rust-lang.org/beta/cargo/reference/manifest.html#the-rust-version-field

Good idea! I've updated the manifests.

vashirov:
rust-1.71 is now in stable Fedora repos, recent PRs started to fail during build time. But our downstreams do not have it yet (like z-stream). And `Cargo.lock` changes will likely go there at some point to address vulnerable dependencies. So I'd be very careful with setting `rust-version`.

droideck:
Okay, I've set it to 1.70 (1.71 was still failing on `Test / Build`).
It should be safe and robust enough.

@Firstyear, will you be okay with that?

droideck:
Okay, we can modify it later if needed.
As of now, we need to deal with the CI failures so other work is not blocked by that.
Merging as I've got Viktor's ack...

PR #5866, opened 2023-07-28
Issue 5834 - AccountPolicyPlugin erroring for some users

### LINKED ISSUE #5834: AccountPolicyPlugin erroring for some users
Hey,

 Since i updated to the lastest Freeipa version (IPA, version: 4.10.1), i started noticing some error in the 389 error.log.

```bash
> [11/Jul/2023:13:50:20.591388308 -0300] - ERR - acct_update_login_history - Modify error 20 on entry 'uid=xyz,cn=users,cn=accounts,dc=mydomain'
> [11/Jul/2023:13:51:52.993254666 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230710114037Z) failed.
> [11/Jul/2023:13:51:52.995079255 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230710114037Z) failed.
> [11/Jul/2023:13:51:53.014918310 -0300] - ERR - acct_update_login_history - Modify error 20 on entry 'uid=abc,cn=users,cn=accounts,dc=mydomain'
> [11/Jul/2023:13:52:12.898523693 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230711165203Z) failed.
> [11/Jul/2023:13:52:12.901073197 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230711165203Z) failed.
> [11/Jul/2023:13:52:12.914936792 -0300] - ERR - acct_update_login_history - Modify error 20 on entry 'uid=abc,cn=users,cn=accounts,dc=mydomain'
> [11/Jul/2023:13:52:21.897598474 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230711165144Z) failed.
> [11/Jul/2023:13:52:21.899560503 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230711165144Z) failed.
> [11/Jul/2023:13:52:21.913459796 -0300] - ERR - acct_update_login_history - Modify error 20 on entry 'uid=xyz,cn=users,cn=accounts,dc=mydomain'
> [11/Jul/2023:13:52:42.939623733 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230710221057Z) failed.
> [11/Jul/2023:13:52:42.942050331 -0300] - ERR - attrlist_replace - attr_replace (lastLoginHistory, 20230710221057Z) failed.
> [11/Jul/2023:13:52:42.955004117 -0300] - ERR - acct_update_login_history - Modify error 20 on entry 'fqdn=hostname.unila.intranet,cn=computers,cn=accounts,dc=mydomain'
```

I'm not sure why this is happening and how can i fix this, unfortunetly i'm using the accountpolicy_plugin to update the last auth date via LDAP, but the field lastLoginHistory seems to be something added in the lastest releases.

For additional details:
- we have 3 replicas but the accountpolicy_plugin is enabled only in 1 server were we concentrate the ldap auth. 
- I can see that the users has the multi-value attribute lastLoginHistory filled, so at some point this is updated.
- It seems to be happening with both user accounts and hosts.

image
image

Any ideas on how to fix this? Thanks.

jchapma:
Bug Description: With the account policy plugin enabled and lastloginhistory size set to non 0 an issue occurs during simultaneous binds of the same user. In this case the timestamp to be stored in the lastloginHistory attribute already exists from a previous bind, and generates an error message.

Fix Description: Before a timestamp is added to the lastloginHistory attribute a check is performed to make sure it doesnt already exist.

Bug Description: A side effect of lastloginHistory feature is that the modifytimestamp value is updated after a successful bind, even when the feature is disabled.

Fix Description: Ensure the entry is not modified when this feature is disabled.

Fixes:	https://github.com/389ds/389-ds-base/issues/5834 
Relates:https://github.com/389ds/389-ds-base/issues/5752

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/acctpolicy/acct_plugin.c
+++ b/ldap/servers/plugins/acctpolicy/acct_plugin.c
@@ -198,6 +198,7 @@ acct_update_login_history(const char *dn, char *timestr)
     void *plugin_id = NULL;
     int rc = -1;
     int num_entries = 0;
+    int mod_required = 0;
     size_t i = 0;
     char **login_hist = NULL;
     Slapi_PBlock *entry_pb = NULL;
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/acctpolicy/acct_plugin.c
```diff
@@ -208,75 +209,96 @@ acct_update_login_history(const char *dn, char *timestr)
...
-            /* remap array*/
-            for (i = 0; i < (cfg->login_history_size - 1); i++) {
-                login_hist[i] = login_hist[(diff + 1) + i];
+            /* remap array if it exists */
+            for (i = 0; i < cfg->login_history_size; i++) {
+                login_hist[i] = login_hist[diff + (i + 1)];
+            }
+            if (cfg->login_history_size != 0) {
+                /* realloc and append latest value */
+                login_hist = (char **)slapi_ch_realloc((char *)login_hist, sizeof(char *) * (cfg->login_history_size + 2));
```
progier389: realloc size is wrong ( should be login_history_size+1) and BTW I do not think that you  
should even realloc  in this case (the buffer is large enough to store the values and will be freed soon (within the pblock_destroy))

progier389 (approved):
LGTM except a minor thing

ldap/servers/plugins/acctpolicy/acct_plugin.c
```diff
@@ -208,75 +209,96 @@ acct_update_login_history(const char *dn, char *timestr)
...
 
+    plugin_id = get_identity();
+    sdn = slapi_sdn_new_normdn_byref(dn);
+    slapi_search_get_entry(&entry_pb, sdn, NULL, &e, plugin_id);
+    slapi_sdn_free(&sdn);
+
     config_rd_lock();
     cfg = get_config();
 
+    /* does this value already exist in the entry */
```
tbordaz, line 230: You may check that 'e' is not NULL.

ldap/servers/plugins/acctpolicy/acct_plugin.c
```diff
@@ -208,75 +209,96 @@ acct_update_login_history(const char *dn, char *timestr)
...
-    list_of_mods[0] = &attribute;
-    list_of_mods[1] = NULL;
-
-    mod_pb = slapi_pblock_new();
-    slapi_modify_internal_set_pb(mod_pb, dn, list_of_mods, NULL, NULL, plugin_id, 0);
-    slapi_modify_internal_pb(mod_pb);
-    slapi_pblock_get(mod_pb, SLAPI_PLUGIN_INTOP_RESULT, &rc);
-    if (rc != LDAP_SUCCESS) {
-        slapi_log_err(SLAPI_LOG_ERR, "acct_update_login_history", "Modify error %d on entry '%s'\n", rc, dn);
+    if (mod_required) {
```
tbordaz, line 283: I have a doubt, if login_history_size==0 what will be the result ?
An attribute 'lastloginhistory' present with no value, or no 'lastloginhistory' attribute ?
I think it would be better to have no 'lastloginhistory' attribute.

progier389: No problems here: 
   in a modify replace a NULL or an empty modifier list means that the attribute is removed. 
BTW in ldap there is no such thing as an attribute without value. 
 ( Attribute always have a value (that value may be empty but in this case you have to specify explicitly that a value is present and the value size is 0)

ldap/servers/plugins/acctpolicy/acct_plugin.c
jchapma: > realloc size is wrong ( should be login_history_size+1) and BTW I do not think that you should even realloc in this case (the buffer is large enough to store the values and will be freed soon (within the pblock_destroy))

Yes you are correct, in this case there is no need for a realloc.

ldap/servers/plugins/acctpolicy/acct_plugin.c
jchapma: > I have a doubt, if login_history_size==0 what will be the result ?

If login_history_size == 0 then the feature is disabled and we do nothing unless the entries array has been trimmed.

Issue #5859, opened 2023-07-25
dbscan fails with AttributeError: 'list' object has no attribute 'extends'


vashirov:
**Issue Description**
Test `dirsrvtests/tests/suites/password/regression_test.py::test_unhashed_pw_switch` fails with an error in dbscan: 
```
>           cmd.extends(['-f', indexfile])
E           AttributeError: 'list' object has no attribute 'extends'

src/lib389/lib389/__init__.py:3057: AttributeError
```

vashirov:
2983d762c..7f0af8770  389-ds-base-2.1 -> 389-ds-base-2.1
   c11e86f6b..9c21538ba  389-ds-base-2.2 -> 389-ds-base-2.2
   7ce99c28a..ce9f523be  389-ds-base-2.3 -> 389-ds-base-2.3

PR #5874, opened 2023-08-01
Issue 5872 - `dbscan()` in lib389 can return bytes

### LINKED ISSUE #5872: `dbscan()` in lib389 can return bytes
**Issue Description**
`dbscan()` in lib389 extracts information from the database file. Most of the time the information returned by `dbscan` executable is strings. But when attribute encryption or changelog encryption is enabled, the database can contain values that can't be parsed as a string in Python.

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dirsrvtests/tests/suites/replication/encryption_cl5_test.py:65: in _check_unhashed_userpw_encrypted
    dbscanOut = inst.dbscan(DEFAULT_BENAME, 'replication_changelog')
/usr/local/lib/python3.9/site-packages/lib389/__init__.py:3072: in dbscan
    result = subprocess.run(cmd, text=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
/usr/lib64/python3.9/subprocess.py:507: in run
    stdout, stderr = process.communicate(input, timeout=timeout)
/usr/lib64/python3.9/subprocess.py:1121: in communicate
    stdout = self.stdout.read()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <encodings.utf_8.IncrementalDecoder object at 0x7f510ffabb80>
input = b"\ndbid: 0000006f000000000000\n\tentry count: 11\n\ndbid: 000000de000000000000\n\tpurge ruv:\n\t\t{replicageneration}...94\xf7\x9f\xa5\xf4\xfb\xd5\xb49\x87W\n\t\tunhashed#user#password: \xa1\x98\xcf\xea\xb8F\xa8\xc9FHe\x8f\x0b\\\xfa\xd7\n"
final = True

    def decode(self, input, final=False):
        # decode input (taking the buffer into account)
        data = self.buffer + input
>       (result, consumed) = self._buffer_decode(data, self.errors, final)
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 6019: invalid start byte

/usr/lib64/python3.9/codecs.py:322: UnicodeDecodeError
```

By default subprocess output is considered bytes:
https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.stdout
> stdout
> Captured stdout from the child process. A bytes sequence, or a string if [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) was called with an encoding, errors, or text=True. None if stdout was not captured.

But we explicitly use `text=True` to indicate that it is supposed to be a string:
https://github.com/389ds/389-ds-base/blob/96959cf7b67be8b544efa25b6ad813c0034841b7/src/lib389/lib389/__init__.py#L3072

I think we should change `dbscan()` to always return bytes.

vashirov:
Bug Description:
When attribute encryption or changelog encryption is used, `dbscan()` can return bytes instead of a string.

Fix Description:
* Update subprocess call to expect bytes instead of string.
* Revert changes to the tests done in 8bf7829ce3e3a8990fccd2fdbe7ae15ca1c8f0e7.
* Update entryrdn_test to expect output from dbscan as bytes.

Fixes: https://github.com/389ds/389-ds-base/issues/5872
Relates: https://github.com/389ds/389-ds-base/issues/5859

Reviewed-by:

### DIFF
--- a/dirsrvtests/tests/suites/indexes/entryrdn_test.py
+++ b/dirsrvtests/tests/suites/indexes/entryrdn_test.py
@@ -17,7 +17,7 @@ from lib389.idm.user import UserAccounts
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.topologies import topology_m2 as topo_m2
 from lib389.agreement import Agreements
-from lib389.utils import ds_is_older
+from lib389.utils import ds_is_older, ensure_bytes
 from lib389.tasks import Tasks,ExportTask, ImportTask
 from lib389.topologies import topology_st as topo
 
@@ -50,7 +50,7 @@ log = logging.getLogger(__name__)
 def checkdbscancount(inst, pattern, expected_count):
     inst.restart()
     dbscanOut = inst.dbscan(args=['-f', f'{inst.dbdir}/{DEFAULT_BENAME}/entryrdn.db', '-A'], stopping=False)
-    count = dbscanOut.count(pattern)
+    count = dbscanOut.count(ensure_bytes(pattern))
     if count != expected_count:
         log.info(f"dbscan output is: {dbscanOut}")
     assert count == expected_count
--- a/dirsrvtests/tests/suites/password/regression_test.py
+++ b/dirsrvtests/tests/suites/password/regression_test.py
@@ -13,7 +13,7 @@ from lib389._constants import SUFFIX, PASSWORD, DN_DM, DN_CONFIG, PLUGIN_RETRO_C
 from lib389 import Entry
 from lib389.topologies import topology_m1 as topo_supplier
 from lib389.idm.user import UserAccounts
-from lib389.utils import ldap, os, logging, ds_is_newer, ds_supports_new_changelog
+from lib389.utils import ldap, os, logging, ensure_bytes, ds_is_newer, ds_supports_new_changelog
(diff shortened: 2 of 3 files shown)


progier389 (approved):
LGTM

droideck (approved):
Ack from me too!

Issue #5883, opened 2023-08-03
Remove connection mutex contention risk on autobind


progier389:
**Issue Description**
Code review show that c_mutex is held for a quite long period (in bind.c) while performing autobind:
When doing anonymous bind on a connection using SASL/EXTERNAL mechanism 
  the mutex is held during the whole lookup that maps the SSL certificate to the bind dn
 but the page search issue showed that holding that c_mutex has a terrible impact on overall performance

**Package Version and Platform:**
 - Platform:  Fedora, RHEL
 - Package and version: main branch 

**Steps to Reproduce**
The steps are quite similar to the paged search result performance issue. 
 The trick is to enable the internal operation flags in access log and disable the standard operation, it makes the fact that the lock is held during autobind internal search very visible.

1. Setup an instance (named standalone1) and having suffix dc=example,dc=com with default tests entries generated by dscreate"
2. Run: `dsconf standalone1 config replace  nsslapd-accesslog-level=4`
3. Run:   `ldclt -p 389  -e esearch -b "dc=example,dc=com" -f "cn=demo"`
4. In another windows run a loop on:
`ldapsearch -Q -LLL -Y EXTERNAL -H ldapi://%2fvar%2frun%2fslapd-standalone1.socket -b dc=example,dc=com cn=demo`
5. Then you should see the ldclt performance dropping significantly:
```
ldclt[416127]: Average rate: 14371.20/thr  (14371.20/sec), total: 143712
ldclt[416127]: Average rate: 14527.00/thr  (14527.00/sec), total: 145270
ldclt[416127]: Average rate: 8274.10/thr  (8274.10/sec), total:  82741
ldclt[416127]: Average rate: 4551.20/thr  (4551.20/sec), total:  45512
ldclt[416127]: Average rate: 4191.30/thr  (4191.30/sec), total:  41913

```

Should not have a significant performance drop.

**Additional context**
I thought about several solutions:
 - computing the mapping out of the lock and only lock to get/data into the connection (but it only solve the autobind code)
 - split the lock between the data needed for accept and listen threads (but it impacts lots of places and so there are risks ... )
 - using try lock to acquire the lock in the listening thread and just skip the connection if the lock is busy. 
   That may leads to CPU load increase (because the listening thread loops until the mutex get released but
   at least events on other connections are handled).

progier389:
Jira ticket: IDMDS-3501

progier389:
[...]

PR #5886, opened 2023-08-04
Issue 5883 - Remove connection mutex contention risk on autobind

progier389:
Problem: A contention on the connection c_mutex is blocking the listener thread when autobind is performed. 

Solution: Let the listener thread skip the connection if the mutex is held by another thread

Issue: [5883](https://github.com/389ds/389-ds-base/issues/5883)

Reviewed by: @mreynolds389 , @droideck Thanks


droideck (approved):
Looks good!

mreynolds389 (approved):

PR #5892, opened 2023-08-10
Issue 5082 - slugify: ModuleNotFoundError when running test cases

### LINKED ISSUE #5082: slugify: ModuleNotFoundError when running test cases.
**Issue Description**
slugify: ModuleNotFoundError when running test cases.

```
ImportError while loading conftest '/workspace/ds/dirsrvtests/conftest.py'.
dirsrvtests/conftest.py:12: in <module>
    from slugify import slugify
E   ModuleNotFoundError: No module named 'slugify'
```

- Platform: RHEL 9 (will check for fedora)

**Steps to Reproduce**
1. Run any test case

**Expected results**
Test cases should run without any error.

vashirov:
Bug Description:
slugify module is used in WebUI tests for creating filenames for screenshots. But it's often not installed by default, since it's not required by lib389. WebUI tests are executed only when a WEBUI environment variable is present, so we should import it under the same condition.

Fix Description:
Import slugify module only when WEBUI environment variable is present and WebUI tests are executed.

Reviewed-by:

Fixes: https://github.com/389ds/389-ds-base/issues/5082

### DIFF
--- a/dirsrvtests/conftest.py
+++ b/dirsrvtests/conftest.py
@@ -18,8 +18,10 @@ import gzip
 from .report import getReport
 from lib389.paths import Paths
 from enum import Enum
-from slugify import slugify
-from pathlib import Path
+
+if "WEBUI" in os.environ:
+    from slugify import slugify
+    from pathlib import Path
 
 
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

Issue #5852, opened 2023-07-24
Remove turbo mode from the code base


jchapma:
**Description.**
Turbo mode is a feature that improves the search performance for a low number of highly active connections. Since its introduction the deployment model has changed and this feature is no longer required. Turbo mode adds some complexity to the code so removal will simplify connection code.

**Solution**
Following investigation into the pros/cons of turbo mode it has been decided that this feature should be removed from the codebase.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2225016

Firstyear:
I think we agreed we could remove it if we had epoll rather than select over the fd's since the point of turbo is to avoid the costly call back to select and it setting up the fd arrays. So provided we have epoll working, I'm all for ripping out turbo mode.

jchapma:
HI @Firstyear 
From what I understand there was a performance degradation with the epoll implementation: https://github.com/389ds/389-ds-base/pull/5182#issuecomment-1073948670

Firstyear:
> HI @Firstyear From what I understand there was a performance degradation with the epoll implementation: [#5182 (comment)](https://github.com/389ds/389-ds-base/pull/5182#issuecomment-1073948670)

Yes, but that's because *our* code was creating the performance issues, rather than epoll. We tend to have pretty awful locking through out the code base, so we should be looking to fix epoll here in the long run since it will be a better approach.

PR #5893, opened 2023-08-10
Issue 5852 = Remove turbo mode from the code base

jchapma:
Description: Turbo mode is a feature that improves the search performance for a low number of highly active connections. Since its introduction the deployment model has changed and this feature is no longer required. Turbo mode adds some complexity to the code so removal will simplify connection code.

Fix description: Following investigation into the pros/cons of turbo mode it has been decided that this feature should be removed from the codebase.

Relates: https://github.com/389ds/389-ds-base/issues/5852

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -41,6 +41,8 @@ static void log_ber_too_big_error(const Connection *conn,
 static PRStack *op_stack;     /* stack of Slapi_Operation * objects so we don't have to malloc/free every time */
 static PRInt32 op_stack_size; /* size of op_stack */
 
+#define CONN_TIMEOUT_INTERVAL 100
+
 struct Slapi_op_stack
 {
     PRStackElem stackelem; /* must be first in struct for PRStack to work */
@@ -1168,7 +1170,7 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
 
     /* If we still haven't seen a complete PDU, read from the network */
     while (*tag == LBER_DEFAULT) {
-        int32_t ioblocktimeout_waits = conn->c_ioblocktimeout;
+        int32_t ioblocktimeout_waits = conn->c_ioblocktimeout/CONN_TIMEOUT_INTERVAL;
         /* We should never get here with data remaining in the buffer */
         PR_ASSERT(!new_operation || !conn_buffered_data_avail_nolock(conn, &conn_closed));
         /* We make a non-blocking read call */
@@ -1262,7 +1264,7 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
             if (SLAPD_PR_WOULD_BLOCK_ERROR(err) ||
                 SLAPD_SYSTEM_WOULD_BLOCK_ERROR(syserr)) {
                 struct PRPollDesc pr_pd;
-                PRIntervalTime timeout = PR_MillisecondsToInterval(100);
+                PRIntervalTime timeout = PR_MillisecondsToInterval(CONN_TIMEOUT_INTERVAL);
                 pr_pd.fd = (PRFileDesc *)conn->c_prfd;
                 pr_pd.in_flags = PR_POLL_READ;
                 pr_pd.out_flags = 0;
@@ -1418,6 +1420,7 @@ connection_threadmain(void *arg)
     int32_t *snmp_vars_idx = (int32_t *) arg;
     /* wait forever for new pb until one is available or shutdown */
     int32_t interval = 0; /* used be  10 seconds */
+    time_t curtime = slapi_current_rel_time_t();
     Connection *conn = NULL;
     Operation *op;
     ber_tag_t tag = 0;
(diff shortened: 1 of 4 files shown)


progier389 (approved):
LGTM

ldap/servers/slapd/connection.c
```diff
@@ -1311,7 +1262,7 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
             if (SLAPD_PR_WOULD_BLOCK_ERROR(err) ||
                 SLAPD_SYSTEM_WOULD_BLOCK_ERROR(syserr)) {
                 struct PRPollDesc pr_pd;
-                PRIntervalTime timeout = PR_MillisecondsToInterval(CONN_TURBO_TIMEOUT_INTERVAL);
+                PRIntervalTime timeout = PR_MillisecondsToInterval(100);
```
Firstyear: Probably should be a constant for future tweaking.

Firstyear (changes requested):
I love PR's that delete lots of code like this :)

vashirov:
I had turbo mode turned off in dse.ldif. After I updated packages, my instance refused to start:
```
Aug 16 09:31:59 rhel93 ns-slapd[5820]: [16/Aug/2023:09:31:59.487554556 -0400] - ERR - dse_read_one_file - The entry cn=config in file /etc/dirsrv/slapd-localhost/dse.ldif (lineno: 10) is invalid, error code 1 (Operations error) - Setting of nsslapd-enable-turbo-mode attribute is currently disabled, it will be removed in the future
```
I think we should not make this an error, but rather WARN or INFO, and not prevent DS from starting.

ldap/servers/slapd/connection.c
```diff
@@ -1217,7 +1168,7 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
 
     /* If we still haven't seen a complete PDU, read from the network */
     while (*tag == LBER_DEFAULT) {
-        int32_t ioblocktimeout_waits = conn->c_ioblocktimeout / CONN_TURBO_TIMEOUT_INTERVAL;
+        int32_t ioblocktimeout_waits = conn->c_ioblocktimeout;
```
tbordaz: My understanding is that ioblocktimeout_waits is a count of the number of wait, not a duration. So we should set it to the number of time we 'waits_done'.

This is why it divides ioblocktimeout by CONN_TURBO_TIMEOUT_INTERVAL and sleep for CONN_TURBO_TIMEOUT_INTERVAL.

ldap/servers/slapd/connection.c
```diff
@@ -1731,52 +1531,11 @@ connection_threadmain(void *arg)
         maxthreads = conn->c_max_threads_per_conn;
         more_data = 0;
         ret = connection_read_operation(conn, op, &tag, &more_data);
-        if ((ret == CONN_DONE) || (ret == CONN_TIMEDOUT)) {
```
tbordaz: This block is just logging debug info regarding the connection. It logs info about turbo mode, that should be removed,  but the rest of the message looks good to me.

ldap/servers/slapd/connection.c
```diff
@@ -1839,12 +1585,6 @@ connection_threadmain(void *arg)
                 pthread_mutex_lock(&(conn->c_mutex));
                 /* don't do this if it would put us over the max threads per conn */
                 if (conn->c_threadnumber < maxthreads) {
-                    /* for turbo, c_idlesince is set above - for !turbo and
-                     * !more_data, we put the conn back in the poll loop and
-                     * c_idlesince is set in handle_pr_read_ready - since we
-                     * are bypassing both of those, we set idlesince here
-                     */
-                    conn->c_idlesince = curtime;
```
tbordaz: Why removing update of idlesince ? There are more data on the socket so it looks to me we can update idlesince.

tbordaz (commented):
Nice code. Few comments/questions

jchapma:
The nsslapd-enable-turbo-mode attribute will remain in the code for a short while, the user will be able to get the attr value and will receive an info message when trying to replace its value.

In the initial commit, I return error code 1 (operations error) when the user tries to replace the attr value, which will issue an info message to the user. But, in the corner case where the user has upgraded DS, instance restart explodes while loading dse.ldif (https://github.com/389ds/389-ds-base/pull/5893#issuecomment-1680617793)

So instead, I return success when the user tries to replace the attr value and I have added the attr type to reject_attr_type() in configdse.c. This way I can issue a message to the user when a replace is attempted and maintain dse compatibility. 

What do you think ?

tbordaz:
A performance issue was found with the replication total init testcase (see https://github.com/389ds/389-ds-base/issues/5761). Additional tests during that investigation, focus on turbo mode impact. It is looking turbo mode was beneficial.
A possibility is that a randomly selected worker, was picking up the first entry of the total update and because of the constant flow of new entries, its turns into a listener/worker (thanks to turbo mode) and finally applies many/all the operations of the total init.

It would be interesting to verify turbo-mode removal on total init

jchapma:
> A performance issue was found with the replication total init testcase (see #5761). Additional tests during that investigation, focus on turbo mode impact. It is looking turbo mode was beneficial. A possibility is that a randomly selected worker, was picking up the first entry of the total update and because of the constant flow of new entries, its turns into a listener/worker (thanks to turbo mode) and finally applies many/all the operations of the total init.
> 
> It would be interesting to verify turbo-mode removal on total init

I agree, following our discussion yesterday it seems like turbo mode could play a part, well for a total init anyway. I will open a ticket to investigate this.

progier389:
> I agree, following our discussion yesterday it seems like turbo mode could play a part, well for a total init anyway. I will open a ticket to investigate this.
FYI: In yesterday tests, I have seen that after disabling the turbo mode, during total update the average operation optime and wtime increased (but not the time between start and end of the total update session.   
  That said the database was quite small (around 550 entries) but it will probably worth to redo the test with 100K or 1M entries  )

tbordaz:
Another thing to clarify is that bind/accept slowness when turbo mode is 'on'. (see solution 4292701). ATM it is not clear to me why it is slower, possibly the consequence a long computation to switch on/off turbo mode.
So AFAIK, this PR requires verification of 'total init' (https://github.com/389ds/389-ds-base/pull/5893#issuecomment-1779199586) and long accept/bind

PR #5897, opened 2023-08-11
issue 5890 part 2 - Need a tester for testing multiple listening thread feature

### LINKED ISSUE #5890: Need a tester for testing multiple listening thread feature
Stress test are not useful to test the impact of  having multiple listener threads feature. (No meaningful impact)
IMHO we need a tool that measure the operation latency with a moderate load (that does not make the server crawling)

A tools that open n connection and generate from time to time a search operation on a randomly open connection.
Then aggregate the operation time and display it regularly

progier389:
**Problem:** after latest commit --amend: elapsed time is now written every operation
       instead of getting aggregated every seconds

**Cause:** I forgot that time.time() is returning a float instead of an int as in C

**Fix:** Convert time.time() result to int

**Issue:** [5890](https://github.com/389ds/389-ds-base/issues/5890)

**Reviewed by:** ?

### DIFF
--- a/dirsrvtests/tests/perf/ltest.py
+++ b/dirsrvtests/tests/perf/ltest.py
@@ -45,7 +45,7 @@ for i in range(args.nbconn):
         raise ex
 print (f'{args.nbconn} connections are open. Starting the latency test using {args} as parameters')
 
-now = time.time()
+now = int(time.time())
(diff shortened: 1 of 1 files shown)


vashirov (approved):
LGTM!

Issue #5906, opened 2023-08-22
Under high pressure on entry cache, a bind on an entry managing retry count may trigger entry cache flush


tbordaz:
**Issue Description**
The bug looks corner case but it is not and may occur quite frequently. The symptoms are quite large (error msg, high CPU/etime/wtime, sometime replication lag,...). The most immediate one is such messages in errors logs
<pre>
... ERR - find_entry_internal_dn - Retry count exceeded (entry_dn)
... WARN - pw_apply_mods - Modify error 32 on entry 'entry_dn'
</pre>

**Package Version and Platform:**
all versions

**Steps to Reproduce**
ATM no reproducer but could likely be reproduced

**Expected results**
should not see those messages. no high cpu/etime/wtime on BIND failure.

tbordaz:
Closing this ticket as the described issue is wrong (see PR)

PR #5907, opened 2023-08-22
Issue 5906 - Under high pressure on entry cache, a bind on an entry m…

tbordaz:
…anaging retry count may trigger entry cache flush

Bug description:
	Under high pressure on the entry cache, there is a good chance
	that an unlocked entry get removed from the cache.
	During a bind, if the password policy of the target entry
	requires an update of the entry (for example
	'passwordRetryCount', the target entry may get out of the
	cache.
	In such case it will trigger entry cache flush (revert_cache)
	that is CPU intensive and that blocks others operations
	accessing the cache.

Fix description:
	If the target entry has the lockout policy, then lock
	the entry in the cache

relates: #5906

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_bind.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_bind.c
@@ -30,6 +30,7 @@ ldbm_back_bind(Slapi_PBlock *pb)
     back_txn txn = {NULL};
     int rc = SLAPI_BIND_SUCCESS;
     int result_sent = 0;
+    int entrylocked = 0;
 
     /* get parameters */
     slapi_pblock_get(pb, SLAPI_BACKEND, &be);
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/back-ldbm/ldbm_bind.c
```diff
@@ -63,13 +64,33 @@ ldbm_back_bind(Slapi_PBlock *pb)
...
+        passwdPolicy *pwpolicy = NULL;
+
+        /* in case of lockout password policy we may need to update
+         * the target entry (when sending the result via update_pw_retry).
+         * In such case we need lock it in the cache
+         */
+        slapi_pblock_get(pb, SLAPI_TARGET_SDN, &sdn);
+        dn = slapi_sdn_get_dn(sdn);
+        pwpolicy = new_passwdPolicy(pb, dn);
+        if (pwpolicy && (pwpolicy->pw_lockout == 1)) {
```
Firstyear, line 79: Doesn't this mean that these entries will be permanently pinned in the cache? That could cause the cache to thrash pretty hard in some setups.

Firstyear (changes requested):

ldap/servers/slapd/back-ldbm/ldbm_bind.c
tbordaz: You are correct, find_entry2modify pins the entry into the cache. This is the reason why the fix calls cache_unlock_entry if entrylocked is set.

Firstyear: Yes, but my point is there is no condition to *unpin* an entry. This means if the cache has 4000 slots, there are 3500 users, then only 500 cache slots will be usable for anything else. This problem is worse if the number of users exceeds the cache size since then they won't be cached at all I would expect.

progier389: Sorry but I do not understand your point:
  If I understand rightly this fix, the entry is pinned when getting it then unpinned (with cache_unlock_entry) once ldap_send_result (which may use the entry) is completed. So the maximum number of entries pinned by this fix is limited to the maximum number of worker thread (and not the number of users)

progier389:
But I also have a concern: send_ldap_result may block until iotimeout in case of problem 
and I am afraid that the user will not being able to bind or modify its entry during that time ...

tbordaz:
> But I also have a concern: send_ldap_result may block until iotimeout in case of problem and I am afraid that the user will not being able to bind or modify its entry during that time ...

Do you mean, there are two binds in parallel on the same entry with lockout policy. The first bind acquires the lock on the entry, fails (for any reason like wrong password), sends Invalid_credential that hangs until ioblocktimeout. Then the second bind can not proceed until ioblocktimeout ends. Am I getting your point ?
I fully agree with that scenario. So the fix should unlock the entry (if locked) before sending result. What do you think ?

progier389:
Yes, that is the scenario that I was thinking of, 
  Your solution will fix that concern. The test should be within send_ldap_result_ext
    (after the password policy handling but before sending the first ber)

tbordaz:
Well I have to apologize on this one because I think the reported issue does not exist.
Indeed the backend callback upload the entry in the entry cache and then when returning the result it check the password policy of the entry and may update it. Checking is done via internal search and update by internal update. If the entry gets out of the cache before checking, it is just reloaded by internal search. If the entry gets out of the cache between checking and update, it is just reloaded by internal modify.
In short I do not see how internal modify could fail (Retry count exceeded) because the entry being unlockable. Each steps backend bind, password  policy check and update have its own mechanism to get the entry and lock it if necessary.

The only benefit of this fix would be to lock the entry since the beginning of the bind (benefit of performance) but the original ticket was for robustness and IIUC I was wrong.

I propose to close that ticket. What do you think ?

progier389:
Feel free to close this PR:
The proposed fix would need a big rework and IMHO it is no more a priority as it will not solve the customer issue that we are seeing.
And as you said the issue description is indeed wrong:
( `the target entry may get out of the cache. In such case it will trigger entry cache flush ` is  wrong )

tbordaz:
Thanks @progier389 , @Firstyear for the reviews. I am closing this PR/ticket as the claimed issue is wrong

PR #5908, opened 2023-08-22
Issue 5853 - Revert MSRV check

droideck:
Description: We should be careful with the rust-version manifest field on older 389-ds-base versions as it's harder to predict at which point in time Rust was updated in some older environments.

Related: https://github.com/389ds/389-ds-base/issues/5861

Reviewed by: ?

### DIFF
--- a/src/librnsslapd/Cargo.toml
+++ b/src/librnsslapd/Cargo.toml
@@ -2,7 +2,6 @@
 name = "librnsslapd"
 version = "0.1.0"
 authors = ["William Brown <william@blackhats.net.au>"]
-rust-version = "1.70"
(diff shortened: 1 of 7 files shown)


droideck:
We need to revert it on 2.2, 2.1, 2.0 branches as some distros may include pretty old Rust packages, and it'll be better to start from 2.3 and forth because:

- The new 1.70 Rust version broke our code on `main` and required an update. Hence we'll benefit from using MSRV there;
- It's easier and more predictable to move forward instead of trying to readjust the past (which works as it is). Hence reverting on older branches.

Running GitHub CI to make sure that everything is okay.

PR #5910, opened 2023-08-28
Issue 5909 - Multi listener hang with 20k connections

jchapma:
Bug Description: When the server is configured with multiple listeners, the connection table is divided into multiple sub tables. These sub tables are then mapped to a single freelist to enable efficient allocation of new connections. Each sub table is a double linked list with element 0 used as the head of the list. During the mapping of sub tables to freelist the head of each sub table is incorrectly mapped to the freelist, creating a "hole" in the freelist.

Fix Description: Skip element 0 of each sub table when mapping to the single freelist.

Fixes: https//github.com/389ds/389-ds-base/issues/5909

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -154,7 +154,7 @@ connection_table_new(int table_size)
         /* We rely on the fact that we called calloc, which zeros the block, so we don't
         * init any structure element unless a zero value is troublesome later
         */
-        for (i = 0; i < ct->list_size; i++) {
+        for (i = 1; i < ct->list_size; i++) {
             /*
             * Technically this is a no-op due to calloc, but we should always be
             * careful with things like this ....


vashirov (approved):
This fixes a hang with 20k connections and 4 listeners for me, ack.

ldap/servers/slapd/conntable.c
```diff
@@ -154,7 +154,7 @@ connection_table_new(int table_size)
         /* We rely on the fact that we called calloc, which zeros the block, so we don't
         * init any structure element unless a zero value is troublesome later
         */
-        for (i = 0; i < ct->list_size; i++) {
+        for (i = 1; i < ct->list_size; i++) {
```
Firstyear, line 157: Isn't the first element of the conntable  a linked list header/pointer? I think that's why this was set to 1 here because the 0'th slot was reserved for LL chasing.

jchapma: @Firstyear 
It is, I cant believe I missed that one :(

Firstyear: It's really not obvious, even I overlooked it when I first worked on that part of the code.

So sadly you'll need to revert this change :(

jchapma: This commit broke some of the CI tests, it has been fixed in https://github.com/389ds/389-ds-base/pull/5917

PR #5925, opened 2023-09-08
Issue 843 - Add a warning to slapi_valueset_add_value_ext

droideck:
Description: The combination of SLAPI_VALUE_FLAG_DUPCHECK and SLAPI_VALUE_FLAG_PASSIN flags is not recommended for slapi_valueset_add_value_ext.

Using this combination could result in undefined behaviour related to memory management. If you need both flags, please use the slapi_valueset_add_attr_value_ext function instead and ensure proper cleanup if there's an error.

We don't use the function with the above flag combination, but someone in the community might (even though it's highly unlikely). Hence, as of now, it isn't worth investing more time into this, and the documentation should be updated with this change.

Related: https://github.com/389ds/389-ds-base/issues/843

Reviewed by: ?


progier389 (approved):
LGTM

Issue #5924, opened 2023-09-07
ASAN server build crash when looping opening/closing connections 


progier389:
**Issue Description**
Get a crash with following stack when doing an ldapmodify loop  targeted to ldapi port

```
 Crash signature is:
==1861306==ERROR: AddressSanitizer: SEGV on unknown address 0x001f00096fe2 (pc 0x7f00639213c7 bp 0x7effb6f70920 sp 0x7effb6f706a0 T31)
==1861306==The signal is caused by a READ memory access.
#0 0x7f00639213c7 in _pr_poll_with_poll (/lib64/libnspr4.so+0x283c7) (BuildId: 274218d7d2f6b1532ca85ca6dcfca4a47312cc54)
#1 0x43de56 in accept_thread /home/progier/sb/389/tst/source/389-ds-base/ldap/servers/slapd/daemon.c:843
#2 0x7f0063926738 in _pt_root (/lib64/libnspr4.so+0x2d738) (BuildId: 274218d7d2f6b1532ca85ca6dcfca4a47312cc54)
#3 0x7f0063dd1906 in start_thread (/lib64/libc.so.6+0x8c906) (BuildId: 245240a31888ad5c11bbc55b18e02d87388f59a9)
#4 0x7f0063e5786f in __clone3 (/lib64/libc.so.6+0x11286f) (BuildId: 245240a31888ad5c11bbc55b18e02d87388f59a9)
```

One of the cause is because the listening file descriptor is closed instead of the accepted file descriptor 
 reading the code, I see that the variable names are misleading and there was a confusion.
The other cause is why are we unable to get a connection slot while only a single connection is open ?

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: 
       Isuue reproduce on:
           -  389-ds-base-2.2
           - main branches
        But not on: 
           -  389-ds-base-2.1
           -  389-ds-base-4.3.1

**Steps to Reproduce**

Use the following script:
```
#!/bin/bash

INAME=scrdl
P="$PREFIX/var/run/slapd-$INAME.socket"
URI="ldapi://$(echo $P | sed 's,/,%2f,g')"

#Limit the connection table size to speed up the test
ulimit  -n 256
ulimit -H -n 256

# remove the previous instance if running again the test
$PREFIX/usr/sbin/dsctl $INAME remove --do-it >/dev/null 2>&1

# Create ds instance (make sure to disable systemctl
#  to insure that ns-slapd is a children of this shell script)
echo "
[...]

PR #5926, opened 2023-09-08
issue 5924 - ASAN server build crash when looping opening/closing connections

progier389:
Issue: Got a crash due to:
      1. Failure to get a connection slot because connection  freelist is misshandled.
      2. A confusion between listening and acceptedfd descriptor leaded to 
          close the listening descriptor while handing the error.
          
Solution: 
      Rename clearly the file descriptor variables
      Close the accepted file descriptor in error handler
      Rewrite the freelist management so that first connection chosen is the last released one.
       (Code is simpler, this fix the end of list issue, and it avoid to spread the open connection over too much memory)

Issue:   #5924 

Reviewed by:  @Firstyear, @vashirov, @droideck,  @tbordaz  (Thanks !)

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -2194,90 +2194,48 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
             assert True
 
 
-@pytest.fixture(scope="module")
-def dscreate_instance(request):
-    template_file = "/tmp/dssetup.inf"
-    serverid = "test-instance"
-    template_text = f"""[general]
-config_version = 2
-# This invalid hostname ...
-full_machine_name = localhost.localdomain
-# Means we absolutely require this.
-strict_host_checking = False
-# In tests, we can be run in containers, NEVER trust
-# that systemd is there, or functional in any capacity
-systemd = False
-
-[slapd]
-instance_name = {serverid}
-root_dn = cn=directory manager
-root_password = {PW_DM}
-# We do not have access to high ports in containers,
-# so default to something higher.
-port = 38999
-secure_port = 63699
-
-
-[backend-userroot]
-suffix = {DEFAULT_SUFFIX}
-sample_entries = yes
-"""
-
-    with open(template_file, "w") as template_fd:
-        template_fd.write(template_text)
-
-    def fin():
-        if DEBUGGING:
-            return
-        os.remove(template_file)
-        try:
-            subprocess.check_call(['dsctl', serverid, 'remove', '--do-it'], stderr=subprocess.STDOUT)
-        except subprocess.CalledProcessError as e:
-            log.fatal("Failed to remove test instance  Error ({}) {}".format(e.returncode, e.output))
-            raise e from None
-
-    request.addfinalizer(fin)
-
-    # Lets remove the instance if it exists.
-    try:
-        subprocess.check_call(['dsctl', serverid, 'remove', '--do-it'], stderr=subprocess.STDOUT)
-    except subprocess.CalledProcessError as e:
-        pass
-
-    # Then creates the instance
-    # Unset PYTHONPATH to avoid mixing old CLI tools and new lib389
-    tmp_env = os.environ
-    if "PYTHONPATH" in tmp_env:
-        del tmp_env["PYTHONPATH"]
+def run_dsctl(serverid, command, *args):
     try:
         subprocess.check_call([
-            'dscreate',
-            'from-file',
-            template_file
-        ], env=tmp_env, stderr=subprocess.STDOUT)
+            'dsctl',
+            serverid,
+            command,
+            *args
+        ], stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as e:
         log.fatal("dscreate failed!  Error ({}) {}".format(e.returncode, e.output))
         raise e from None
 
-    inst = DirSrv(verbose=True, external_log=log)
-    dse_ldif = DSEldif(inst, serverid=serverid)
 
-    socket_path = dse_ldif.get("cn=config", "nsslapd-ldapifilepath")
-    inst.local_simple_allocate(
-       serverid=serverid,
-       ldapuri=f"ldapi://{socket_path[0].replace('/', '%2f')}",
-       password=PW_DM
-    )
-    inst.ldapi_enabled = 'on'
-    inst.ldapi_socket = socket_path
-    inst.ldapi_autobind = 'off'
-    try:
-        inst.open()
-    except Exception as e:
-        log.fatal("Failed to connect via ldapi to %s instance" % serverid)
-        raise e from None
+@pytest.fixture(scope="module")
+def dscreate_instance(request, topology_st):
 
-    return inst
+    inst = topology_st.standalone
+    # Disable systemd by turning off with_system in .inf file
+    old_path = Paths()._get_defaults_loc(DEFAULTS_PATH)
+    new_path = f'{old_path}.orig'
+    run_dsctl(inst.serverid, 'stop')
+    os.rename(old_path, new_path)
+
+    def fin():
+        try:
+            run_dsctl(inst.serverid, 'stop')
+        finally:
+            os.rename(new_path, old_path)
+            inst.start()
+
+    request.addfinalizer(fin)
+
+    with open(new_path, 'rt') as fin:
+       with open(old_path, 'wt') as fout:
+            for line in fin:
+                if line.startswith('with_systemd'):
+                    fout.write('with_systemd = 0\n')
+                else:
+                    fout.write(line)
+    run_dsctl(inst.serverid, 'start')
+
+    return topology_st.standalone
 
 
 @pytest.fixture(scope="module", params=set(range(1,5)))
@@ -2287,7 +2245,7 @@ def dscreate_with_numlistener(request, dscreate_instance):
     inst = dscreate_instance
     inst_name = inst.serverid
 
-    wrapper_file = f"{os.getenv("HOME")}/dswrapper.sh"
+    wrapper_file = "/tmp/dswrapper.sh"
     wrapper_text = f"""#!/usr/bin/bash
 ulimit -n {maxfds}
 ulimit -H -n {maxfds}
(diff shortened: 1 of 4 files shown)


dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -2186,6 +2194,194 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
             assert True
 
 
+@pytest.fixture(scope="module")
+def dscreate_instance(request):
```
vashirov: I see that we have couple of similar tests in basic test suite that use this way of creating the instance. What is the reason for this? Can't we use the ones from topologies.py or create a new topology with the required settings?

progier389: One of the reason is that theses tests use some non default values typically systemd = False (to disable systemctl) 
The sad thing is that our test topology framework lack the ability to specify arbitrary template parameters ( or dse.ldif config parameters that requires restart ) but anyway I think that such a change should be done as part of issue #3082 (the removal of pytest dependency on lib38)

Firstyear: > One of the reason is that theses tests use some non default values typically systemd = False (to disable systemctl) The sad thing is that our test topology framework lack the ability to specify arbitrary template parameters ( or dse.ldif config parameters that requires restart ) but anyway I think that such a change should be done as part of issue #3082 (the removal of pytest dependency on lib38)

That was an extremely intentional design choice. It was to force growing lib389 to support the full surface area of our server config, so that it could become the cli tools that we have today. It also meant that the surface area of the admin tool internals is being tested incidentally during these integration tests. And it shows the "process" an admin would go through to make these changes, and how that would happen in reality. 

The other part was that dse.ldif was getting a bit "out of control" template wise. I wanted to get things to work in containers and that meant we had to support in place migrations and upgrades, so any value that existed in dse.ldif could be "removed" and a default from libglobs used. It could never be perfect, but it's significantly helped reduce complexity in that area for users.

progier389: - I think you answered to my second point (why not having a phase allowing to tweak the dse.ldif
    before the initial startup). Ok this argument makes sense even if sometime it also makes the test case very complex (for example having to export the db, remove the db files and reimport when changing some db parameter like the page size) in some case.

  - But my main point is that right now it is cumbersome to use lib389 to create an instance 
   without explicitly using dscreate subprocesses.
    Creating an instance that is named "foo" is simple with dscreate, it should also be simple with direct call to lib389 too.  IMHO the upper level function that mimics dscreate (but providing the data from a dict instead of the dscreate template file) is just missing.

ldap/servers/slapd/conntable.c
```diff
@@ -48,69 +48,61 @@
...
  * to O(1) worst case as we always recieve and empty slot *or* ct full. It also reduces lock/atomic
  * contention on the CPU to improve things.
+ * Lastly a change was done: Instead of having a sliding windows that tend to get a never used
+ * slot for each new connection, it nows reuse last freed one. That has several benefits:
+ *  - Fix a bug because the last free list slaots may not be alloced.
+ *  - Avoid to grow the memory footprint when there is no much load
+ *  - Simplify the code (as only a single is needed. )
  *
- * The freelist is a ringbuffer of pointers to the connection table. On a small scale it looks like:
+ * The freelist is a ringbuffer of pointers to the connection table.
```
Firstyear: It's no longer a ring buffer, it's a stack. :)

ldap/servers/slapd/conntable.c
```diff
@@ -48,69 +48,61 @@
...
- *  | _NULL  | _NULL  | _ ptr  | _ ptr  | _ ptr  |
+ *  | slot 0 | slot 1 | slot 2 | slot 3 | slot 4 |
+ *  | _NULL  | _NULL  | _ ptr  | _ ptr  | NULL   |
  *  |--------------------------------------------|
- *     ^                  ^- conn_next
- *     |
- *     \-- conn_free
+ *                        ^- conn_next
  *
  * When a connection is "freed" we return it to conn_free, which is then also slid up.
```
Firstyear: conn_free removed, so comment needs update.

Firstyear: Also mention that conn_next is moving back to the left here for the freed conn ptr to occup the slot to the left.

ldap/servers/slapd/conntable.c
```diff
@@ -137,10 +129,8 @@ connection_table_new(int table_size)
     ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, ct->list_num * sizeof(struct POLL_STRUCT*));
     ct->table_mutex = PR_NewLock();
     /* Allocate the freelist */
-    ct->c_freelist = (Connection **)slapi_ch_calloc(1, ct->size * sizeof(Connection *));
-    /* NEVER use slot 0, this is a list pointer */
-    ct->conn_next_offset = 1;
-    ct->conn_free_offset = 1;
+    ct->c_freelist = (Connection **)slapi_ch_calloc(1, (ct->size+1) * sizeof(Connection *));
```
Firstyear, line 130: Needs a comment that size here needs the +1 to be greater than the number of connections avaliable due to null terminate of the freelist.

ldap/servers/slapd/conntable.c
```diff
@@ -273,22 +263,21 @@ connection_table_get_connection(Connection_Table *ct, int sd)
...
     if (c != NULL) {
         /* We allocated it, so now NULL the slot and move forward. */
-        ct->c_freelist[ct->conn_next_offset] = NULL;
-        /* Handle overflow. */
-        ct->conn_next_offset = (ct->conn_next_offset + 1) % ct->size;
-        if (ct->conn_next_offset == 0) {
-            /* Never use slot 0 */
-            ct->conn_next_offset += 1;
-        }
+        ct->c_freelist[ct->conn_next_offset++] = NULL;
```
Firstyear, line 268: PR_ASSERT here that conn_next_offset must be less than < size + 1.

ldap/servers/slapd/conntable.c
```diff
@@ -48,69 +48,61 @@
  * under the connection table lock. This should move the allocation algorithm from O(n) worst case
  * to O(1) worst case as we always recieve and empty slot *or* ct full. It also reduces lock/atomic
  * contention on the CPU to improve things.
+ * Lastly a change was done: Instead of having a sliding windows that tend to get a never used
+ * slot for each new connection, it nows reuse last freed one. That has several benefits:
+ *  - Fix a bug because the last free list slaots may not be alloced.
```
Firstyear: Why were the last slots not allocated?

Firstyear (changes requested):

ldap/servers/slapd/conntable.c
progier389: Mostly because I saw it was the case when gdb triggered the watch point ! -;) 
More seriously: 
  I initially thought it was because of rounding issue (i.e: table_size != ct->list_size*ct->numlist)
but the way initialization is done, table_size >= ct->size*ct->numlist so there should not be any empty slot.
So I think that the reason is that we skip first slot for each listener table when filling the free list  
(I suspect the slot it is reserved (maybe for the signal pipe ? )

ldap/servers/slapd/conntable.c
progier389: Fixed.

ldap/servers/slapd/conntable.c
progier389: Fixed

Firstyear (approved):
The C side looks good to me, I think @droideck should check the python though :)

ldap/servers/slapd/conntable.c
```diff
@@ -48,69 +48,59 @@
...
- *  | _ ptr  | _ ptr  | _ ptr  | _ ptr  | _ ptr  |
+ *  | slot 0 | slot 1 | slot 2 | slot 3 | slot 4 |
+ *  | _ ptr  | _ ptr  | _ ptr  | _ ptr  | NULL   |
  *  |--------------------------------------------|
- *     ^  ^- conn_next
- *     |
- *     \-- conn_free
+ *        ^- conn_next
  *
  * As we allocate, we shift conn_next through the list, yielding the ptr that was stored (and
```
tbordaz: stack not list

ldap/servers/slapd/conntable.c
```diff
@@ -461,14 +449,10 @@ connection_table_move_connection_out_of_active_list(Connection_Table *ct, Connec
...
     PR_ASSERT(c->c_refcnt == 0);
-    PR_ASSERT(ct->conn_free_offset != 0);
-    PR_ASSERT(ct->c_freelist[ct->conn_free_offset] == NULL);
-    ct->c_freelist[ct->conn_free_offset] = c;
-    ct->conn_free_offset = (ct->conn_free_offset + 1) % ct->size;
-    if (ct->conn_free_offset == 0) {
-        /* Never use slot 0 */
-        ct->conn_free_offset += 1;
-    }
+    PR_ASSERT(ct->conn_next_offset != 0);
```
tbordaz, line 452: Could you confirm ? I understood that the range [0..ct_size-1] are valid values in the freelist.

tbordaz (approved):
The code looks good to me. Just minor comments

ldap/servers/slapd/conntable.c
progier389: The PR_ASSERT is correct.
AFAIK the free list range is [0..ct_size].
But the various PR_ASSERT checks some specific conditions depending of the assert location.
In connection_table_move_connection_out_of_active_list code,  the connection that is moved should still be open (even if it may be in the closing state) so the free list cannot be full (i.e ct->conn_next_offset > 0)

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -2186,6 +2379,102 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
...
+
+
+@pytest.mark.skipif(ds_is_older('2.2.0.0'),
+                    reason="This test is only required with multiple listener support.")
+def test_conn_limits(dscreate_with_numlistener):
+    """Check the connections limits for various number of listeners.
+
+    :id: 7be2eb5c-4d8f-11ee-ae3d-482ae39447e5
+    :parametrized: yes
+    :setup: None
```
droideck: I think you should mention the steps you did in `dscreate_with_numlistener`.
So I think you should move the first step here as it's actually more about `setup` than the test step.

droideck (approved):
Besides the minor issue, both C and Python code look good to me!

PR #5929, opened 2023-09-14
Issue 1870 - Add a CI test

droideck:
Description: If two instances of the plugin are created with the same origin scope,
only the request of one of the plugins will be satisfied.
(i.e. template and instance A and B, that both make entries in ou A and B,
from ou=People, only A will work)

Related: https://github.com/389ds/389-ds-base/issues/1870

Author: wibrown
Reviser: spichugi

Reviewed by: ?


progier389 (approved):
LGTM

PR #5930, opened 2023-09-18
Update replication.py

strzh:
is this wrong with args?

### DIFF
--- a/man/man1/ldclt.1
+++ b/man/man1/ldclt.1
@@ -217,6 +217,49 @@ Execution parameters:
 .br
 \fBrandomauthidhigh=value\fR high value for random SASL Authid.
 .PP
+.SH EXAMPLES
+To populate a database with 1000 user objects create a file template.ldif containing:
+.PP
+.nf
+.RS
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/cli_conf/replication.py
```diff
@@ -633,7 +633,7 @@ def create_repl_manager(inst, basedn, log, args):
     if args.passwd is not None:
         repl_manager_password = args.passwd
     elif args.passwd_file is not None:
```
tbordaz: Shouldn't it be args.bind_passwd_file ?

progier389: IMHO @tbordaz is right: 
bind-password-file is used in all sub-parsers except in repl_add_manager_parser  
So the right way to solve the issue is probably to change the parser 
(but keeping the --password-file as alias to avoid compatibility issue)
i.e: repl_add_manager_parser.add_argument('--bind-password-file', '--passwd-file', help="File containing the password")

PR #5934, opened 2023-09-21
update replication.py

### LINKED ISSUE #5935: bug with --passwd-file when create-manager
**Issue Description**
could not use --passwd-file when execute below commands:
dsconf instance replication create-manager --passwd-file /file/name
error message:
Error: 'Namespace' object has no attribute 'bind_passwd_file'

 - Platform: [SLES]
 - Package and version: [e.g. lib389-2.2.8~git37.fdb3bae-150500.3.11.1]
 - Browser [NA]

**Steps to Reproduce**
1. execute command: dsconf instance replication create-manager --passwd-file /file/name
2. See error: Error: 'Namespace' object has no attribute 'bind_passwd_file'

**Expected results**
Shoud without issues.

**Screenshots**
NA

**Additional context**
Solved issue with https://github.com/389ds/389-ds-base/pull/5934

strzh:
Fix a dsconf fails because a naming mismatch between argparse parameters and args attribute) 
Solved by using consistent naming while keeping old name for compatibility

Issue: #5935 

Reviewer: @tbordaz @progier389 @droideck

### DIFF
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -633,6 +633,8 @@ def create_repl_manager(inst, basedn, log, args):
     if args.passwd is not None:
         repl_manager_password = args.passwd
     elif args.passwd_file is not None:
+        repl_manager_password = get_passwd_from_file(args.passwd_file)
+    elif args.bind_passwd_file is not None:
         repl_manager_password = get_passwd_from_file(args.bind_passwd_file)
     elif repl_manager_password == "":
         repl_manager_password = _get_arg(None, msg=f"Enter replication manager password for \"{manager_dn}\"",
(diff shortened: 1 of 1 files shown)


progier389 (approved):
The changes look good. 
Thanks for the contribution !
But you should also create the issue associated with PR 
   (In your case, the important part is the one that describes step by step how you reproduce the bug) 
(FYI: The issue will be helpful later on for:
   - the quality engineer that test the bug is fixed in a given release
   - the bug documentation
   -  tracking the eventual cherry-pick in other branches
 )

strzh:
> The changes look good. Thanks for the contribution ! But you should also create the issue associated with PR (In your case, the important part is the one that describes step by step how you reproduce the bug) (FYI: The issue will be helpful later on for:
> 
> * the quality engineer that test the bug is fixed in a given release
> * the bug documentation
> * tracking the eventual cherry-pick in other branches
>   )

Hi, created related issue https://github.com/389ds/389-ds-base/issues/5935

droideck (changes requested):
I have a doubt about the change...
Do we really need to introduce another parameter for compatibility when it's just a simple bug?

I think we can leave `--passwd-file` as it is and use its value for `repl_manager_password`.
What do you think?

progier389:
HI @droideck, In fact that was the original idea in the first PR but both Thierry and me were thinking it is better to have a consistent interface (when looking at the other argparse option in  replica.py it is always named bind-password-file.)
I would simply have added an alias but IMHO the way it is done here as the merit of being more clear.

droideck (approved):
Okay, if no one objects beside me, I agree with the majority. Ack!

PR #5936, opened 2023-09-22
Issue 1925 - Add a CI test

droideck:
Description: Verify that the issue is not present. Cover the scenario when
we remove existing VLVs, create new VLVs (with the same name) and
then we do online re-indexing.

Related: https://github.com/389ds/389-ds-base/issues/1925

Reviewed by: ?


progier389 (approved):
LGTM

PR #5940, opened 2023-09-28
Issue 5938 - Attribute Names changed to lowercase after adding the Attributes

### LINKED ISSUE #5938: Attribute Names changed to lowercase after adding the Attributes.
Description of problem:
We are working with the web console to edit the attributes, which have both capital and lowercase letters in their names. The capital letters within the names of an attribute changes whenever the attributes are added to an object class. 

389-ds-base-1.4.3

Steps to Reproduce:
1. Create  an attribute called TestAttribute

2. On the shell:
[root@dirsrv2 ~]# grep -r TestAttribute /etc/dirsrv/slapd-localhost/* 
/etc/dirsrv/slapd-localhost/schema/99user.ldif:attributeTypes: ( TestAttribute-oid NAME 'TestAttribute' DESC 'Just a Test' SY

3.Edit the description of attribute from web ui

On the shell:
[root@dirsrv2 ~]# grep -i -r TestAttribute /etc/dirsrv/slapd-localhost/* 
/etc/dirsrv/slapd-localhost/schema/99user.ldif:attributeTypes: ( testattribute-oid NAME 'testattribute' DESC 'Just another T

4. The Attributes name was changed to lower case.

Expected results:

> It should be in the same syntax as added.

> CamelCase not reflected when editing an attribute

droideck:
Bug Description: When working with the web console to edit the attributes
with capital and lowercase letters in their names. The capital letters within
the names of an attribute change whenever the attributes are added
to an object class.

Fix Description: Presevrer the case in both UI and CLI when doing
edit/add/list operations.

Fixes: https://github.com/389ds/389-ds-base/issues/5938

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/lib/schema/schemaTables.jsx
+++ b/src/cockpit/389-console/src/lib/schema/schemaTables.jsx
@@ -760,7 +760,7 @@ class MatchingRulesTable extends React.Component {
             let name = "";
             // Check for matches of all the parts
             if (row.names.length > 0) {
-                name = row.name[0].toLowerCase();
+                name = row.name[0];
             }
             if (val !== "" && name.indexOf(val) === -1 &&
                 row.oid[0].indexOf(val) === -1 &&
--- a/src/lib389/lib389/schema.py
+++ b/src/lib389/lib389/schema.py
@@ -128,10 +128,10 @@ class Schema(DSLdapObject):
             for obj in results:
                 obj_i = vars(object_model(obj))
                 if len(obj_i["names"]) == 1:
-                    obj_i['name'] = obj_i['names'][0].lower()
+                    obj_i['name'] = obj_i['names'][0]
                     obj_i['aliases'] = None
                 elif len(obj_i["names"]) > 1:
-                    obj_i['name'] = obj_i['names'][0].lower()
+                    obj_i['name'] = obj_i['names'][0]
                     obj_i['aliases'] = obj_i['names'][1:]
                 else:
                     obj_i['name'] = ""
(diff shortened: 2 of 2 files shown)


mreynolds389:
My concern here is that I think lib389 expects schema to be lowercase in all cases.  I would make sure all basic schema operations work in a mixed case environment

droideck:
> My concern here is that I think lib389 expects schema to be lowercase in all cases. I would make sure all basic schema operations work in a mixed case environment

Works for me!
Still, feel free to test it, folks, in case I missed something. :)

progier389 (approved):
Looks good. 
 Apparently the various schema items lookup are still done using lowercase, so it should work.
Anyway, if we discover that we still  problem with case insensitive comparison in lib389, my feeling is that rather that adding lower() everywhere we should then take the time to design a CaseInsensitiveString (str subclass) that  overwrites __eq__, __lt__, __le__ and __hash__ and use this class for all attribute types and most attribute values (according to the schema)

droideck:
> Looks good. Apparently the various schema items lookup are still done using lowercase, so it should work. Anyway, if we discover that we still problem with case insensitive comparison in lib389, my feeling is that rather that adding lower() everywhere we should then take the time to design a CaseInsensitiveString (str subclass) that overwrites **eq**, **lt**, **le** and **hash** and use this class for all attribute types and most attribute values (according to the schema)

Hmhm, a CaseInsensitiveString subclass sounds good! It'll bring consistency and clarity in how case insensitivity is handled throughout the system.

PR #5943, opened 2023-10-04
Issue 5843 - dsconf / dscreate should be able to handle lmdb parameters

### LINKED ISSUE #5843: dsconf / dscreate should be able to handle lmdb parameters
**Issue Description**
lmdb specific parameters should be visible in dsconf

 - Platform: Fedora
 - Package and version: main

**Steps to Reproduce**
1. Configure an instance with lmdb
2. Run dsconf supplier1 backend config set --help

**Expected results**
Should see the lmdb specific parameters like:
- nsslapd-mdb-max-size
- nsslapd-mdb-max-readers
- nsslapd-mdb-max-dbs 

**Additional context**
dscreate create-template should also propose the nsslapd-mdb-max-size as it is a critical parameter
dscreate interactive should ask for 
 - the bdb/mdb mode
 - the mdb max size if in mdb mode

progier389:
Description:

  **`dscreate` changes:**
    - make db_lib a standard option instead of an advanced one
    - add mdb_max_size as standard option 
  **`dsconf instance backend config set` changes:** 
     - add --mdb_max_size option associated with nsslapd-mdb-max-size
     - add --mdb_max_readers option associated with nsslapd-mdb_max_readers
     - add --mdb_max_dbs option associated with nsslapd-mdb_max_dbs

Issue: #5843 

Reviewed by: @Firstyear and @droideck  ( Thanks ! )

### DIFF
--- a/src/lib389/lib389/cli_conf/backend.py
+++ b/src/lib389/lib389/cli_conf/backend.py
@@ -65,8 +65,7 @@ arg_to_attr = {
         'deadlock_policy': 'nsslapd-db-deadlock-policy',
         'db_home_directory': 'nsslapd-db-home-directory',
         'db_lib': 'nsslapd-backend-implement',
-        'lmdb_size': 'nsslapd-mdb-max-size',    # For dscreate (unit is Gb)
-        'mdb_max_size': 'nsslapd-mdb-max-size', # For dsconf (unit is bytes)
+        'mdb_max_size': 'nsslapd-mdb-max-size',
         'mdb_max_readers': 'nsslapd-mdb-max-readers',
         'mdb_max_dbs': 'nsslapd-mdb-max-dbs',
         # VLV attributes
--- a/src/lib389/lib389/cli_conf/backend.py
+++ b/src/lib389/lib389/cli_conf/backend.py
@@ -65,6 +65,10 @@ arg_to_attr = {
         'deadlock_policy': 'nsslapd-db-deadlock-policy',
         'db_home_directory': 'nsslapd-db-home-directory',
         'db_lib': 'nsslapd-backend-implement',
+        'lmdb_size': 'nsslapd-mdb-max-size',    # For dscreate (unit is Gb)
+        'mdb_max_size': 'nsslapd-mdb-max-size', # For dsconf (unit is bytes)
+        'mdb_max_readers': 'nsslapd-mdb-max-readers',
+        'mdb_max_dbs': 'nsslapd-mdb-max-dbs',
         # VLV attributes
         'search_base': 'vlvbase',
         'search_scope': 'vlvscope',
@@ -1081,6 +1085,10 @@ def create_parser(subparsers):
     set_db_config_parser.add_argument('--deadlock-policy', help='Adjusts the backend database deadlock policy (Advanced setting)')
     set_db_config_parser.add_argument('--db-home-directory', help='Sets the directory for the database mmapped files (Advanced setting)')
     set_db_config_parser.add_argument('--db-lib', help='Sets which db lib is used. Valid values are: bdb or mdb')
+    set_db_config_parser.add_argument('--mdb-max-size', help='Sets the lmdb database maximum size (in bytes).')
+    set_db_config_parser.add_argument('--mdb-max-readers', help='Sets the lmdb database maximum number of readers (Advanced setting)')
+    set_db_config_parser.add_argument('--mdb-max-dbs', help='Sets the lmdb database maximum number of sub databases (Advanced setting)')
+
 
     #######################################################
     # Database & Suffix Monitor
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -15,7 +15,8 @@ import re
 import glob
 import shutil
 from lib389.dseldif import DSEldif
-from lib389._constants import DEFAULT_LMDB_SIZE, GIGABYTE
+from lib389._constants import DEFAULT_LMDB_SIZE
+from lib389.utils import parse_size
 import subprocess
 
 
@@ -236,7 +237,7 @@ def dblib_bdb2mdb(inst, log, args):
         total_dbi += be['dbi']
 
     # Round up dbmap size
-    dbmap_size = DEFAULT_LMDB_SIZE * GIGABYTE
+    dbmap_size = parse_size(DEFAULT_LMDB_SIZE)
     while (total_dbsize * DBSIZE_MARGIN > dbmap_size):
         dbmap_size *= 1.25
 
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -15,11 +15,11 @@ import re
 import glob
 import shutil
 from lib389.dseldif import DSEldif
+from lib389._constants import DEFAULT_LMDB_SIZE, GIGABYTE
 import subprocess
 
 
 DBLIB_LDIF_PREFIX = "__dblib-"
-DEFAULT_DBMAP_SIZE = 2 * 1024 * 1024 * 1024
 DBSIZE_MARGIN = 1.2
 DBI_MARGIN = 60
 
@@ -236,7 +236,7 @@ def dblib_bdb2mdb(inst, log, args):
         total_dbi += be['dbi']
 
     # Round up dbmap size
-    dbmap_size = DEFAULT_DBMAP_SIZE
+    dbmap_size = DEFAULT_LMDB_SIZE * GIGABYTE
     while (total_dbsize * DBSIZE_MARGIN > dbmap_size):
         dbmap_size *= 1.25
 
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -16,8 +16,9 @@ import glob
 import shutil
 from lib389.dseldif import DSEldif
 from lib389._constants import DEFAULT_LMDB_SIZE
-from lib389.utils import parse_size
+from lib389.utils import parse_size, format_size
 import subprocess
+from errno import ENOSPC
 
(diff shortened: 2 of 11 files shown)


src/lib389/lib389/cli_conf/backend.py
```diff
@@ -65,6 +65,10 @@
         'deadlock_policy': 'nsslapd-db-deadlock-policy',
         'db_home_directory': 'nsslapd-db-home-directory',
         'db_lib': 'nsslapd-backend-implement',
+        'lmdb_size': 'nsslapd-mdb-max-size',    # For dscreate (unit is Gb)
```
Firstyear: Why have this as two separate parameters? Just because of the different units? Why not do something else to parse the units or similar?

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -236,7 +236,7 @@ def dblib_bdb2mdb(inst, log, args):
         total_dbi += be['dbi']
 
     # Round up dbmap size
-    dbmap_size = DEFAULT_DBMAP_SIZE
+    dbmap_size = DEFAULT_LMDB_SIZE * GIGABYTE
     while (total_dbsize * DBSIZE_MARGIN > dbmap_size):
         dbmap_size *= 1.25
```
Firstyear: I'm not sure what this is trying to achieve? maybe some better comments about this algo and why.

src/lib389/lib389/instance/options.py
```diff
@@ -114,6 +115,8 @@ def parse_inf_config(self, config):
                     v = config.getint(self._section, k)
                 elif self._type[k] == bool:
                     v = config.getboolean(self._section, k)
+                elif self._type[k] == float:
```
Firstyear: This should be an int not a float. But better would be a way to parse the units.

src/lib389/lib389/instance/setup.py
```diff
@@ -446,6 +449,36 @@ def create_from_cli(self):
             slapd['root_password'] = rootpw1
             break
 
+        # Database implementation (db_lib)
+        while 1:
```
Firstyear: while True.

Firstyear (changes requested):

src/lib389/lib389/cli_conf/backend.py
progier389: I hesitated about that one, but IMHO you are right: parsing human friendly size would be more consistent and that is not hard do. So I will change that.

src/lib389/lib389/cli_ctl/dblib.py
progier389: When looking at the whole function, it is clear that this code computes a size for lmdb database that is large enough to store the existing data.
IMHO the puzzling question is not what it tries to achieve but why it is done this way 
   (starting from the default size and increasing the size by steps until it is large enough) 
    rather than using directly the computed size plus a margin ?
 There was probably a reason but I just do not remember it and whatever it was it is likely arguable
 But that piece of code was written 2 years ago and the only thing that changed is the way to get the default value. So my feeling is that we are getting out of the scope of this change and if you think we should really change things in this area, we should better open a new issue.

src/lib389/lib389/instance/setup.py
progier389: I looked at other occurrences as I cut and pasted that piece of code from previous option handling
and indeed the file is plagued by that one.
grep -c 'while 1' instance/setup.py
11
I will global substitute them ...

src/lib389/lib389/utils.py
```diff
@@ -1701,10 +1708,36 @@ def is_valid_hostname(hostname):
     return all(allowed.match(x) for x in hostname.split("."))
 
 
+def parse_size(size):
```
droideck, line 1711: Could you please add docstrings for the added functions, describing what they do. Like here, and `def get_default_mdb_max_size(paths):`.

droideck (changes requested):
Beside that, I see that the dblib test fails (I'm not sure if it's related):

```
    def test_dblib_migration(topo_m2, init_user):
        """
        Verify dsctl dblib xxxxxxx sub commands (migration between bdb and lmdb)
    
        :id: 5d327c34-e77a-46e5-a8aa-0a552f9bbdef
        :setup: Two suppliers Instance
        :steps:
            1. Determine current database
            2. Switch to the other database
            3. Check that replication works
        :expectedresults:
            1. Success
            2. Success
            3. Success
        """
        s1 = topo_m2.ms["supplier1"]
        s2 = topo_m2.ms["supplier2"]
        db_lib = s1.get_db_lib()
        repl = ReplicationManager(DEFAULT_SUFFIX)
        users = UserAccounts(s1, DEFAULT_SUFFIX)
        assert users.get('test entry')
        args = FakeArgs({'tmpdir': None})
        if db_lib == 'bdb':
            dblib_bdb2mdb(s1, log, args)
            dblib_cleanup(s1, log, args)
            _check_db(s1, log, 'mdb')
            repl.test_replication_topology([s1, s2])
            dblib_mdb2bdb(s1, log, args)
            dblib_cleanup(s1, log, args)
            _check_db(s1, log, 'bdb')
            repl.test_replication_topology([s1, s2])
        else:
            dblib_mdb2bdb(s1, log, args)
            dblib_cleanup(s1, log, args)
            _check_db(s1, log, 'bdb')
            repl.test_replication_topology([s1, s2])
            dblib_bdb2mdb(s1, log, args)
            dblib_cleanup(s1, log, args)
>           _check_db(s1, log, 'mdb')

dirsrvtests/tests/suites/clu/dsctl_dblib_test.py:118: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
dirsrvtests/tests/suites/clu/dsctl_dblib_test.py:56: in _check_db
    assert DatabaseConfig(inst).get_db_lib() == impl
/usr/lib/python3.11/site-packages/lib389/backend.py:1107: in __init__
    self._db_lib = self.get_attr_val_utf8_l('nsslapd-backend-implement')
/usr/lib/python3.11/site-packages/lib389/_mapped_object.py:800: in get_attr_val_utf8_l
    x = self.get_attr_val(key)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <lib389.backend.DatabaseConfig object at 0x7f566e5fd710>
key = 'nsslapd-backend-implement', use_json = False

    def get_attr_val(self, key, use_json=False):
        self._log.debug("%s getVal(%r)" % (self._dn, key))
        # We might need to add a state check for NONE dn.
        if self._instance.state != DIRSRV_STATE_ONLINE:
>           raise ValueError("Invalid state. Cannot get properties on instance that is not ONLINE")
E           ValueError: Invalid state. Cannot get properties on instance that is not ONLINE
```

src/lib389/lib389/cli_ctl/dblib.py
progier389: Anyway I have to review that code to fix a regression due to the change of default db map size from 2Gb to 20Gb 
So I spend the time to understand what I have done 2 years ago ! -;)  
Then I added a few comments, especially on the tests done after the size computation.

src/lib389/lib389/instance/options.py
progier389: Fixed (it is now a string)

src/lib389/lib389/utils.py
progier389: Done

progier389:
I fixed the review comments and the failing clu test case (migration tools tests)
Tests are now OK.
There is still a failure in vlv test in a very recently added test case (about testing VLV reindexing)
And I am pretty sure it is a genuine bug (vlv index is a real nightmare with lmdb) but unrelated to the current PR 
 I will open an issue about that one.

src/lib389/lib389/utils.py
```diff
@@ -1701,10 +1708,66 @@ def is_valid_hostname(hostname):
...
+    # Make sure that there is enough available disk space
+    # otherwise decrease the value
+    try:
+        statvfs = os.statvfs(paths.db_dir)
+        avail = statvfs.f_frsize * statvfs.f_bavail
+        avail *= 0.8 # Reserve 20% as margin
+        if size > avail:
+            mdb_max_size = str(avail)
+    except Exception as e:
+        pass
```
droideck: I think we should at least log the error to `debug` log level... But ideally, we would skip only the expected error types (not all of  them).

droideck (approved):
Besides the minor issue, looks good to me!

src/lib389/lib389/utils.py
progier389: I agree: it should have been OSError instead of Exception ...
If we choose to log a message, we also need to change the test. 
 (Because FileNotFound is expected when using the default Paths 
    (instance is not known and the path is  /var/lib/dirsrv/slapd-{instance_name}/db  ) 
I guess we have better retry with the parent path.
Working on it ...

progier389: It is fixed. 
Also added a small change to display properly the DSLdapObject subclass objects:
Will now display:
`E       AssertionError:  assert not [<class 'lib389.idm.account.Account'>(instance_name="standalone1", dn="uid=demo_user,ou=people,dc=example,dc=com"), <class 'lib389.idm.account.Account'>(instance_name="standalone1", dn="cn=demo_group,ou=Groups,dc=example,dc=com")]
`
Instead of:
`E       AssertionError: assert not [<lib389.idm.account.Account object at 0x7fcc849ea790>, <lib389.idm.account.Account object at 0x7fcc849eb150>]`

Firstyear (approved):

src/lib389/lib389/utils.py
droideck: Looks good! Thanks!

PR #5950, opened 2023-10-13
Issue 5843 - Fix dscreate create-template issue

progier389:
Regression with dscreate from-instance before creating first instance:
  the target directory does not exists and a warning is inserted in the template file
  which is then broken.

Fix is pretty trivial:
  Select an existing directory to compute the file system free space.

Issue #5943 

Reviewed by: @droideck (Thanks !)


droideck (approved):
LGTM

PR #5953, opened 2023-10-15
Issue 6061 - Certificate lifetime displayed as NaN

Adadov:
Bug Description:
HOST_TIME_GMT content depends on user locale. On french version (maybe others too) date calculation doesn't work 

Fix Description:
Add "LANG=C" before date comand to ensure format will be correct.

Author: Adadov
Relates: #6061

### DIFF
--- a/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
+++ b/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
@@ -998,7 +998,7 @@ export function showCertificate (certificate, showCertCallback) {
         'echo ' + certificate +
     ' | base64 --decode | openssl x509 -inform DER -noout -subject -issuer -dates -serial ;' +
     // ' echo HOST_TIME_GMT=`TZ=GMT date "+%Y-%m-%d %H:%M:%S %Z"` ' // Issue with Firefox and Safari.
-    ' echo HOST_TIME_GMT=`TZ=GMT date`'
+    ' echo HOST_TIME_GMT=`date -Iminutes`'
     ];
 
     let result = {};


src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
```diff
@@ -998,7 +998,7 @@ export function showCertificate (certificate, showCertCallback) {
         'echo ' + certificate +
     ' | base64 --decode | openssl x509 -inform DER -noout -subject -issuer -dates -serial ;' +
     // ' echo HOST_TIME_GMT=`TZ=GMT date "+%Y-%m-%d %H:%M:%S %Z"` ' // Issue with Firefox and Safari.
-    ' echo HOST_TIME_GMT=`TZ=GMT date`'
+    ' LANG=C; echo HOST_TIME_GMT=`TZ=GMT date`'
```
vashirov:
```suggestion
    ' echo HOST_TIME_GMT=`LANG=C TZ=GMT date`'
```
I think this would be more clear since we already pass a variable for `date` to use GMT timezone.

vashirov (changes requested):
Hi @Adadov,
sorry for the late review. Could you please open an issue for tracking purposes and link it in the commit message and PR, similar to other commit messages?
Thanks.

src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
progier389: Generally speaking, it is better to avoid `VAR=value; cmd` pattern because VAR may not be exported (and therefore may not be set in cmd environment). `export VAR=value; cmd` or `VAR=value cmd` should be used instead.
It is not really an issue in this specific case because if LANG is not set, C is still be used as a default value, but  @vashirov proposal avoid the reader to wonder  whether the code is correct or not! :wink:

vashirov (approved):

vashirov:
@Adadov, thank you for your contribution!

Issue #5956, opened 2023-10-17
After an upgrade the LDAP server won't start if nsslapd-conntablesize is present in the dse.ldif file


tmihinto:
**Issue Description**
After an upgrade of the 389-ds packages, the server won't start and it complains about *nsslapd-conntablesize* being present in the dse.ldif file.

**Package Version and Platform:**
 - Platform: RHEL 8.8
 - Package and version: 389-ds-base-1.4.3.35-2.module+el8.8.0+19693+b24f535c.x86_64

**Steps to Reproduce**
1. Make sure nsslapd-conntablesize is present in the dse.ldif
2. Upgrade the 389-ds packages ( here I'm using a system with IPA installed )
3. Attempts to restart the LDAP server will fail.
For instance:
`ipactl start
Starting Directory Service
Failed to start Directory Service: CalledProcessError(Command ['/bin/systemctl', 'start', 'dirsrv@<INSTANCE>.service'] returned non-zero exit status 1)
`
4. LDAP errors log:
`[17/Oct/2023:14:45:29.422170234 +0200] - ERR - dse_read_one_file - The entry cn=config in file /etc/dirsrv/slapd-<INSTANCE>/dse.ldif (lineno: 17) is invalid, error code 1 (Operations error) - User setting of nsslapd-conntablesize attribute is disabled, server has auto calculated its value to 0.`
`[17/Oct/2023:14:45:29.493095110 +0200] - ERR - init_dse_file - Could not load config file [dse.ldif]`
`[17/Oct/2023:14:45:29.494808167 +0200] - ERR - setup_internal_backends - Please edit the file to correct the reported problems and then restart the server.`

5. After removing the nsslapd-conntablesize line, the server can be started.

**Expected results**
Working LDAP server.

**Additional context**
This can also be reproduced by editing directly the dse.ldif and adding the nsslapd-conntablesize parameter.

jchapma:
19db79958..88616ccc6  389-ds-base-1.4.4 -> 389-ds-base-1.4.4
0fbb32b5d..71aeab39f  389-ds-base-2.0 -> 389-ds-base-2.0
a2dbf9c8d..8ac08e978  389-ds-base-2.1 -> 389-ds-base-2.1
f62be2526..8884fcf37  389-ds-base-2.2 -> 389-ds-base-2.2

tbordaz:
[...]

PR #5963, opened 2023-10-18
 Issue 5956 - After an upgrade the server won't start - nsslapd-connta…  …blesize

jchapma:
Description: Add a test case for issue 5956

Related: https://github.com/389ds/389-ds-base/issues/5956

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -13,7 +13,7 @@ from lib389.idm.user import UserAccount, UserAccounts
 import pytest
 from lib389.tasks import *
 from lib389.utils import *
-from lib389.topologies import topology_st
+from lib389.topologies import topology_st as topo
 from lib389.dbgen import dbgen_users
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389._constants import DN_DM, PASSWORD, PW_DM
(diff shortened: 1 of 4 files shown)


dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1713,6 +1713,38 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
...
+    assert inst.status() == True
+
+    inst.stop()
+    assert inst.status() == False
+    assert not dse_ldif.get(DN_CONFIG, attr)
+    dse_ldif.add(DN_CONFIG, attr, attr_value)
+    assert dse_ldif.get(DN_CONFIG, attr)
+
+    inst.start()
+    assert inst.status() == True
```
tbordaz: You may check that the message "User setting of nsslapd-conntablesize is disabled,..." is logged in the error and after startup the server can answer simple request (like search base "")

tbordaz (approved):

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -1713,6 +1713,38 @@ def test_dscreate_with_different_rdn(dscreate_test_rdn_value):
         else:
             assert True
 
+def test_conntablesize_attr_dse(topo):
```
tbordaz: This test may be skipped on main and after 2.3

droideck:
I think we usually create the PR for `main` first and then cherry-pick it to older branches. Or do you want this PR in 1.4.3 only? If so, then nevermind.

dirsrvtests/tests/suites/basic/basic_test.py
jchapma: Done

dirsrvtests/tests/suites/basic/basic_test.py
jchapma: The test runs on branches 1.4.3 -> 2.2

jchapma:
> I think we usually create the PR for `main` first and then cherry-pick it to older branches. Or do you want this PR in 1.4.3 only? If so, then nevermind.

In this case the issue is just on 1.4.3, we dont have this issue on main branch.

PR #5964, opened 2023-10-18
Issue 5954 - Disable Transparent Huge Pages v2

### LINKED ISSUE #5954: Disable Transparent Huge Pages
**Description of problem:**
As described in [bug 2068120](https://bugzilla.redhat.com/show_bug.cgi?id=2068120), THP can have negative effects on RHDS performance.
This seems to happen with large caches.
Some customers are getting uncontrolled memory usage while others are seeing high CPU consumption.
It would make sense to either disable THP for RHDS or at least add some messages in the errors logs.
RHDS healthcheck tool should also report when THP is enabled on the system.

**Version-Release number of selected component (if applicable):**
This is impacting all RHDS 11 and 12 versions.

**How reproducible:**
At customer sites.

**Steps to Reproduce:**
1. Configure large caches ( > 50 GB )
2. Check memory and CPU usage when RHDS is under heavy load
3. Disable THP ==> RHDS uses less memory and CPU

**Actual results:**
High memory and CPU usage.

**Expected results:**
Controlled memory and CPU usage.

**Additional info:**
    https://access.redhat.com/solutions/3799821

vashirov:
Bug Description:
THP can have negative effects on DS performance when large caches are
used.

Fix Description:
* Add a new variable for `ns-slapd` THP_DISABLE.
  When THP_DISABLE is set to 1, THP is disabled for `ns-slapd` process
  via `prctl(2)`. With any other value, THP settings are untouched.

Before:
```
$ grep THP /proc/$(pidof ns-slapd)/status
THP_enabled:    1
```

After
```
$ grep THP /proc/$(pidof ns-slapd)/status
THP_enabled:    0
```

* Add a new healthcheck linter, that checks if THP is disabled system-wide
  or per instance. In case THP is enabled for both the system and the
  process, it prints recommendations how to disable THP.

Fixes: https://github.com/389ds/389-ds-base/issues/5954

Reviewed-by: ???

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_tunables_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_tunables_test.py
@@ -1,0 +1,148 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2023 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+
+import subprocess
+import pytest
+import re
+import os
+from lib389.utils import *
+from lib389.cli_base import FakeArgs
+from lib389.topologies import topology_st
+from lib389.cli_ctl.health import health_check_run
+from lib389.paths import Paths
+
+CMD_OUTPUT = 'No issues found.'
+JSON_OUTPUT = '[]'
+RET_CODE = 'DSTHPLE0001'
+
+log = logging.getLogger(__name__)
+p = Paths()
+
+
+def run_healthcheck_and_flush_log(topology, instance, searched_code=None, json=False, searched_code2=None,
+                                  list_checks=False, list_errors=False, check=None, searched_list=None):
+    args = FakeArgs()
+    args.instance = instance.serverid
+    args.verbose = instance.verbose
+    args.list_errors = list_errors
+    args.list_checks = list_checks
+    args.check = check
+    args.dry_run = False
+    args.json = json
+
+    log.info('Use healthcheck with --json == {} option'.format(json))
+    health_check_run(instance, topology.logcap.log, args)
+
+    if searched_list is not None:
+        for item in searched_list:
+            assert topology.logcap.contains(item)
+            log.info('Healthcheck returned searched item: %s' % item)
+    else:
+        assert topology.logcap.contains(searched_code)
+        log.info('Healthcheck returned searched code: %s' % searched_code)
+
+    if searched_code2 is not None:
+        assert topology.logcap.contains(searched_code2)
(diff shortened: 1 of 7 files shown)


tbordaz (approved):
Thanks LGTM
Let's see which option is preferred

Firstyear:
My concern here is we are addressing the symptom not the problem. I think if we disable THP then we'll likely never turn it back on again, and we could be losing out on benefits. Shouldn't we pursue a fix to the underlying issue?

vashirov:
I understand your concern. I think that the underlying issue is that THP is not applicable to certain workloads, where data access patterns are not contiguous, with a lot of small allocations and deallocations.
Some examples:
https://www.pingcap.com/blog/transparent-huge-pages-why-we-disable-it-for-databases/
https://www.mongodb.com/docs/manual/tutorial/transparent-huge-pages/
https://redis.io/docs/management/optimization/latency/#latency-induced-by-transparent-huge-pages
https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-HUGE-PAGES
https://docs.oracle.com/en/database/oracle/oracle-database/19/ladbi/disabling-transparent-hugepages.html

Every optimization is a tradeoff and comes with a cost. THP can reduce page faults, but in our case it also causes memory fragmentation, CPU spikes (`khugepaged` has a non-zero overhead) and latency spikes. If we can't take advantage of this optimization due to our workload and usage patterns, I see no reason to keep it enabled.

Firstyear:
> Every optimization is a tradeoff and comes with a cost. THP can reduce page faults, but in our case it also causes memory fragmentation, CPU spikes (`khugepaged` has a non-zero overhead) and latency spikes. If we can't take advantage of this optimization due to our workload and usage patterns, I see no reason to keep it enabled.

Okay, lets disable it then.

Firstyear (approved):

vashirov:
@Firstyear, thanks!

@tbordaz, any thoughts on the healthcheck part? Do you think it's still needed?

tbordaz:
> @Firstyear, thanks!
> 
> @tbordaz, any thoughts on the healthcheck part? Do you think it's still needed?

IIUC, with the patch the server requests to disable THP for itself. So the benefit of healthcheck would be to double check that the request was successful. I think it worthes doing that check if it is not too complex to implement.

droideck (approved):

vashirov:
I've added healthcheck linter and tests for it, please review.
Thanks!

vashirov:
I fixed the test, now it passes. @tbordaz, please let me know if you're happy with the healthcheck changes, thanks.

tbordaz:
> I fixed the test, now it passes. @tbordaz, please let me know if you're happy with the healthcheck changes, thanks.

Yes that looks perfect to me

PR #5974, opened 2023-10-30
Issue 5973 - Fix fedora cop RawHide builds

### LINKED ISSUE #5973: Fix fedora cop RawHide builds in @389ds/389-ds-base-nightly
**Issue Description**

rawHide builds are failing after a upgrade of gcc
(see  https://copr.fedorainfracloud.org/coprs/g/389ds/389-ds-base-nightly/build/6579720/ )

 - Platform: fedora-rawhide-s390x and fedora-rawhide-x86_64
 - Package and version:  main branch - nightly builds 

**Steps to Reproduce**
1. Looks at https://copr.fedorainfracloud.org/coprs/g/389ds/389-ds-base-nightly/builds/ results
2. Click on latest build
3. See fedora-rawhide-s390x and fedora-rawhide-x86_64 status: `Failed`

**Expected results**
fedora-rawhide-s390x and fedora-rawhide-x86_64 status should be `succeded`

**Additional context**
Two issues here: 
   - One looks like a plain cpp bug (regression in #include handling)
   - Second one is that large file functions (i.e open64) are no more defined with the current compilation flags. 
 Both issue can be easily worked around:
   - first one by moving `#include <sys/statvfs.h>` before `#include of "ldbm-backend.h"`
   - second one by not using large file function in __LP64__ architecture 
      (in this case, large files are handled properly by the standard functions anyway)

progier389:
Problem: @389ds/389-ds-base-nightly copr nigthly builds faild on
 fedora-rawhide-s390x and fedora-rawhide-x86_64

Solution:
   [1] Work around a gcc cpp bug by moving stavfs.h include line
       before ldbm-backend.h include line
   [2] Do not use large file API on LP64 architecture

Issue:  @5973

Reviewed by:  @tbordaz , @droideck (Thanks !)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -18,7 +18,10 @@
 
 #if defined(HPUX11) || defined(OS_solaris) || defined(linux)
 /* built-in 64-bit file I/O support */
+#if ! defined(__LP64__)
+/* But not on 64-bit arch: It is needless and build fails since gcc 13.2.1-4 */
 #define DB_USE_64LFS
+#endif
 #endif
 
 /* needed by at least HPUX and Solaris, to define off64_t */
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
@@ -7,14 +7,14 @@
  * END COPYRIGHT BLOCK **/
 
 
+#include <sys/types.h>
+#include <sys/statvfs.h>
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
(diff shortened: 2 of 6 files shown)


tbordaz (approved):
LGTM

droideck (approved):
Looks good.

A couple of minor notes:
- do you know if the GCC CPP bug was filed already?
- In the PR description - `Issue: closing @5973` - it won't work AFAIK. Probably, it should be something like `Fixes: https://github.com/389ds/389-ds-base/issues/5973` (which then will be resolved correctly to something like `Fixes: #5973` but in the actual repo it will stay as a raw link)

progier389:
I do not think the bug was reported (I did not found anything about it but I am not sure I was looking at the right place)
About the description: yes I also noticed that the "closing" keywords does not work as advertised so I set the "development" link afterwards as usual!

PR #5982, opened 2023-11-15
Bump version to 2.5.0

jchapma:


VERSION.sh
```diff
@@ -10,7 +10,7 @@ vendor="389 Project"
 # PACKAGE_VERSION is constructed from these
 VERSION_MAJOR=2
 VERSION_MINOR=4
```
vashirov: Shouldn't this be 2.5.0, now that 2.4 is branched?

jchapma: It should :(

vashirov (approved):

PR #5985, opened 2023-11-16
Issue 5984 - Crash when paged result search are abandoned

### LINKED ISSUE #4551: Paged search impacts performance

### LINKED ISSUE #5984: Crash when paged result search are abandoned
**Issue Description**
The fact that the paged result search is now using a lock different than the connection lock introduced a race condition 
  with the ABANDON operation leading to a double free that usually leads to an heap corruption and  a crash (SIGSEGV)
Note:  a signature of that crash is that:
 strings -a `path_of_core_file`  | grep 'ABANDON targetop=Simple Paged Results' 
displays some results (if access log is not disabled)

 - Platform: Any
 - Package and version: Unfortunately all branches released after July 2023 are impacted 

**Steps to Reproduce**
Will provide a CI test but the idea is to have a relatively long paged result search and abandoning it while it is sending entries.

**Expected results**
No crash !

progier389:
Problem:
  Fix #4551 has changed the lock that protects the paged result data
  within a connection. But the abandon operation attempts to free
  the paged search result with the connection lock.
  This leads to race condition and double free causing an heap
  corruption and a SIGSEGV.

  Solution:
   - Get a copy of the operation data that needs to be logged.
   - Unlock the connection mutex (to avoid deadlock risk)
   - Free the paged result while holding the paged result lock.

Issue: 5984

### DIFF
--- a/ldap/servers/slapd/abandon.c
+++ b/ldap/servers/slapd/abandon.c
@@ -38,6 +38,12 @@ do_abandon(Slapi_PBlock *pb)
     Connection *pb_conn = NULL;
     Operation *pb_op = NULL;
     Operation *o;
+    /* Keep a copy of some data because o may vanish once conn is unlocked */
+    struct {
+        struct timespec hr_time_end;
+        int nentries;
+        int opid;
+    } o_copy; 
 
     slapi_pblock_get(pb, SLAPI_OPERATION, &pb_op);
     slapi_pblock_get(pb, SLAPI_CONNECTION, &pb_conn);
@@ -90,8 +96,12 @@ do_abandon(Slapi_PBlock *pb)
 
     pthread_mutex_lock(&(pb_conn->c_mutex));
     for (o = pb_conn->c_ops; o != NULL; o = o->o_next) {
-        if (o->o_msgid == id && o != pb_op)
+        if (o->o_msgid == id && o != pb_op) {
+            slapi_operation_time_elapsed(o, &o_copy.hr_time_end);
+            o_copy.nentries = o->o_results.r.r_search.nentries;
+            o_copy.opid = o->o_opid;
             break;
+        }
     }
 
     if (o != NULL) {
@@ -130,7 +140,8 @@ do_abandon(Slapi_PBlock *pb)
         slapi_log_err(SLAPI_LOG_TRACE, "do_abandon", "op not found\n");
     }
 
-    if (0 == pagedresults_free_one_msgid_nolock(pb_conn, id)) {
+    pthread_mutex_unlock(&(pb_conn->c_mutex));
+    if (0 == pagedresults_free_one_msgid(pb_conn, id, pageresult_lock_get_addr(pb_conn))) {
         slapi_log_access(LDAP_DEBUG_STATS, "conn=%" PRIu64
                                            " op=%d ABANDON targetop=Simple Paged Results msgid=%d\n",
                          pb_conn->c_connid, pb_op->o_opid, id);
@@ -143,15 +154,11 @@ do_abandon(Slapi_PBlock *pb)
                                            " targetop=SUPPRESSED-BY-PLUGIN msgid=%d\n",
                          pb_conn->c_connid, pb_op->o_opid, id);
     } else {
-        struct timespec o_hr_time_end;
-        slapi_operation_time_elapsed(o, &o_hr_time_end);
         slapi_log_access(LDAP_DEBUG_STATS, "conn=%" PRIu64 " op=%d ABANDON"
                                            " targetop=%d msgid=%d nentries=%d etime=%" PRId64 ".%010" PRId64 "\n",
-                         pb_conn->c_connid, pb_op->o_opid, o->o_opid, id,
-                         o->o_results.r.r_search.nentries, (int64_t)o_hr_time_end.tv_sec, (int64_t)o_hr_time_end.tv_nsec);
+                         pb_conn->c_connid, pb_op->o_opid, o_copy.opid, id,
+                         o_copy.nentries, (int64_t)o_copy.hr_time_end.tv_sec, (int64_t)o_copy.hr_time_end.tv_nsec);
     }
-
-    pthread_mutex_unlock(&(pb_conn->c_mutex));
     /*
      * Wake up the persistent searches, so they
      * can notice if they've been abandoned.
--- a/ldap/servers/slapd/opshared.c
+++ b/ldap/servers/slapd/opshared.c
@@ -921,9 +921,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                     next_be = NULL; /* to break the loop */
                     if (operation->o_status & SLAPI_OP_STATUS_ABANDONED) {
                         /* It turned out this search was abandoned. */
-                        pthread_mutex_lock(pagedresults_mutex);
-                        pagedresults_free_one_msgid_nolock(pb_conn, operation->o_msgid);
-                        pthread_mutex_unlock(pagedresults_mutex);
+                        pagedresults_free_one_msgid(pb_conn, operation->o_msgid, pagedresults_mutex);
                         /* paged-results-request was abandoned; making an empty cookie. */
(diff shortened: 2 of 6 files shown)


ldap/servers/slapd/opshared.c
```diff
@@ -921,9 +921,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                     next_be = NULL; /* to break the loop */
                     if (operation->o_status & SLAPI_OP_STATUS_ABANDONED) {
                         /* It turned out this search was abandoned. */
-                        pthread_mutex_lock(pagedresults_mutex);
-                        pagedresults_free_one_msgid_nolock(pb_conn, operation->o_msgid);
-                        pthread_mutex_unlock(pagedresults_mutex);
+                        pagedresults_free_one_msgid(pb_conn, operation->o_msgid, pagedresults_mutex);
```
tbordaz, line 924: IIUC it looks to me that pagedresults_mutex is set if it is an paged result search.
In case of abandon (not paged result) how is it set ?

ldap/servers/slapd/pagedresults.c
```diff
@@ -350,7 +354,7 @@ pagedresults_free_one(Connection *conn, Operation *op, int index)
  * Used for abandoning - pageresult_lock_get_addr(conn) is already locked in do_abandone.
  */
 int
-pagedresults_free_one_msgid_nolock(Connection *conn, ber_int_t msgid)
+pagedresults_free_one_msgid(Connection *conn, ber_int_t msgid, pthread_mutex_t *mutex)
```
tbordaz, line 357: Why not retrieving the mutex in the function itself with pageresult_lock_get_addr(conn) ?

ldap/servers/slapd/pagedresults.c
```diff
@@ -361,6 +365,7 @@ pagedresults_free_one_msgid_nolock(Connection *conn, ber_int_t msgid)
         } else {
             slapi_log_err(SLAPI_LOG_TRACE,
                           "pagedresults_free_one_msgid_nolock", "=> msgid=%d\n", msgid);
+            pthread_mutex_lock(mutex);
```
tbordaz, line 368: A dummy question not related to that specific line.
The "random" mutex is taken from a hash table based on the 'conn' value (pageresult_lock_get_addr). How can we be sure there is no collision ?
For example, two different connections using the same mutex.

Another question, I believe it is possible to have several paged result searches on the same connection. Any risk they are stepping on each over toes ? I mean lock/unlock the same mutex

ldap/servers/slapd/abandon.c
```diff
@@ -130,7 +140,8 @@ do_abandon(Slapi_PBlock *pb)
         slapi_log_err(SLAPI_LOG_TRACE, "do_abandon", "op not found\n");
     }
 
-    if (0 == pagedresults_free_one_msgid_nolock(pb_conn, id)) {
+    pthread_mutex_unlock(&(pb_conn->c_mutex));
```
tbordaz, line 143: Does that mean that one of the possible symptom was a deadlock ?

ldap/servers/slapd/abandon.c
```diff
@@ -130,7 +140,8 @@ do_abandon(Slapi_PBlock *pb)
         slapi_log_err(SLAPI_LOG_TRACE, "do_abandon", "op not found\n");
     }
 
-    if (0 == pagedresults_free_one_msgid_nolock(pb_conn, id)) {
```
tbordaz, line 133: IIUC here was the culprit of the crashes. We freed using c_mutex instead of paged_result_mutex. Correct ?

tbordaz (commented):
The overall fix looks good to me.
I added few comments/questions to clarify my understanding.

ldap/servers/slapd/pagedresults.c
progier389: Indeed in both these scenarios different threads will use the same mutex and there will be some lock contention.
But that is the finality of a mutex and the contention is smaller with an array of mutexes than with a single one !

ldap/servers/slapd/pagedresults.c
progier389: Why recomputing the hash in the function when the caller already have computed it ( as in opshared.c case ) ?

ldap/servers/slapd/opshared.c
progier389: Abandon code does not use `pagedresults_mutex` but  `pageresult_lock_get_addr(pb_conn)`
which will return one of the mutex of the hash array initialized during the server startup
(whether there are paged search result associated with the connection or not ... )

ldap/servers/slapd/abandon.c
progier389: Well everything is possible after an heap corruption but deadlock is really very improbable, because the previous code was not locking at all (causing the race condition and the heap corruption) 
But my first idea was to simply protect the pagedresults_free_one_msgid_nolock with the paged mutex lock and I realized that this could generate deadlock if the connection mutex was not freed before.

ldap/servers/slapd/abandon.c
progier389: Yes

ldap/servers/slapd/pagedresults.c
tbordaz: Yes that make sense. The target was to reduce the contention. thanks

ldap/servers/slapd/pagedresults.c
tbordaz: exact

ldap/servers/slapd/opshared.c
tbordaz: Sorry, my question was confusing. The call (pagedresults_free_one_msgid) uses pagedresults_mutex and I thought that pagedresults_mutex was not initialized. But I was wrong, we are in a 'op_is_pagedresults' block where pagedresults_mutex is initialized.
Sorry for the noise

tbordaz (approved):
The C code looks good to me

Issue #5980, opened 2023-11-14
RFE: Improve instance startup failure handling


progier389:
**Is your feature request related to a problem? Please describe.**
`dsctl instance starts` only reports that startup failed without providing any hints.
This make diagnostic much painful. (Especially in the case of CI test reports)

**Describe the solution you'd like**
IMHO in case of failure,  before logging the general "Failed to start" message, 
`lib389 DirSrv.start()` should log the content of the error log file 
 after the last "starting up" message skipping any line with one of the following keywords:
   DEBUG, INFO, NOTICE, WARN
 
 That usually provides enough information to understand the problem.

progier389:
This one was handled when fixing lmdb CI test issue as I needed it to understand why systemctl was failing.
in PR [#5996](https://github.com/389ds/389-ds-base/pull/5996)

PR #5987, opened 2023-11-20
Issue 5984 - Crash when paged result search are abandoned - fix2

### LINKED ISSUE #5984: Crash when paged result search are abandoned

progier389:
Chasing several rabbits at the same time is a bad idea !
and I mixed branches and unwillingly pushed one commit for #5980 in #5984 
  just before the PR #5985 merge ! -:(
Hopefully it does not break anything but just logs some useless crap if instance fails to starts.
Anyway This commit reverts the change about __init.py
 and also do a minor code cleanup (removed a trailing space) in abandon.c 

Issue #5984 

Reviewed by: @tbordaz  Thanks !

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -1048,24 +1048,6 @@ class DirSrv(SimpleLDAPObject, object):
 
         self.state = DIRSRV_STATE_OFFLINE
 
-    def dump_errorlog(self):
-        '''
-            Its logs all errors messages within the error log that occured 
-            after the last startup.
-        '''
-        if os.path.isfile(self.errlog):
-            lines = []
-            with open(self.errlog, 'r') as file:
-                for line in file:
-                    if "starting up" in line:
-                        lines = []
-                    for key in ( 'DEBUG', 'INFO', 'NOTICE', 'WARN' ):
-                        if key in line:
-                            lines.append(line)
-                            break
-            for line in lines:
-                self.log.error(line)
-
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -1048,24 +1048,6 @@ def close(self):
 
         self.state = DIRSRV_STATE_OFFLINE
 
-    def dump_errorlog(self):
```
tbordaz, line 1051: This helper function looks valid Are you sure you want to revert that change ?

tbordaz (approved):
LGTM

src/lib389/lib389/__init__.py
progier389: Yes, The fix is incomplete (a test is inverted) and had a side effect on CI tests 
 (because it change an exception type in some cases )
 I plan to create a clean PR for this improvement.

PR #5991, opened 2023-11-21
Issue 5980 - Improve instance startup failure handling

progier389:
Displays the important error log messages (those that are not: INFO/DEBUG/WARNING) when the server fails to start
to provide the root cause of the failure and help to diagnose some CI tests failures.

Issue: #5990

Reviewed by: @tbordaz  Thanks!

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -1048,6 +1048,38 @@ class DirSrv(SimpleLDAPObject, object):
 
         self.state = DIRSRV_STATE_OFFLINE
 
+    def dump_errorlog(self):
+        '''
+            Its logs all errors messages within the error log that occured 
+            after the last startup.
+        '''
+        errlog = self.ds_paths.error_log
+        if os.path.isfile(errlog):
+            lines = []
+            nbskipped = 0
+            with open(errlog, 'r') as file:
+                for line in file:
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -1071,7 +1104,13 @@ def start(self, timeout=120, post_open=True):
         if self.with_systemd():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
-            subprocess.check_output(["systemctl", "start", "dirsrv@%s" % self.serverid], stderr=subprocess.STDOUT)
+            try:
+                subprocess.check_output(["systemctl", "start", "dirsrv@%s" % self.serverid], stderr=subprocess.STDOUT)
+            except subprocess.CalledProcessError as e:
+                self.dump_errorlog()
+                logger.info('Failed to start dirsrv@%s: "%s"' % (self.serverid, e.output.decode()))
```
progier389, line 1111: FYI: Using logger.info to logs data here because self.log.info and logger.errors fails in `dsctl instance start`
          [1] self.log is not yet initialized
          [2] error log level is not enabled.  (only INFO (and DEBUG if -v option is used) are set)  and nothing is logged.

tbordaz (approved):
LGTM

Issue #5944, opened 2023-10-05
Reversion of the entry cache should be limited to BETXN plugin failures


tbordaz:
**Issue Description**
During an update if an BETXN plugin fails the full TXN is aborted and the DB returns to the previous state. However potential internal updates, done by BETXN plugins, are also applied on the entry cache. Some entries in the entry cache are left in a state that does not reflect the DB state. To prevent this mismatch, upon BETXN failure,  the fix https://pagure.io/389-ds-base/issue/50260 reverts some entries in the entry cache .

The computation of the entries to revert is expensive and impacts performance. The problem is that reversion is not limited to the cases of BETXN failures that was the initial goal.

**Package Version and Platform:**
since 1.3.10

**Steps to Reproduce**
will be provided in the PR

tbordaz:
cfc0d7576..855e6f29b main

tbordaz:
Sorry I was too fast, I think it needs backport down to 1.4.3.
It is a common issue and have some impact in production

tbordaz:
cd7fc221e..f9d09a907  389-ds-base-2.4
dd03a509c..445a881d9  389-ds-base-2.3
5eb779bc6..e2ed3561d  389-ds-base-2.2
40081442f..27904223d  389-ds-base-2.1
7edfc7add..37868719e  389-ds-base-2.0
8c917ee5e..ab5e8e484  389-ds-base-1.4.4
fcb04abe7..77bfdfe4c  389-ds-base-1.4.3

PR #5994, opened 2023-11-28
Issue 5944 - Reversion of the entry cache should be limited to BETXN …

tbordaz:
…plugin failures

Bug description:
	During an update if an BETXN plugin fails the full TXN is aborted and
	the DB returns to the previous state. However potential internal
	updates, done by BETXN plugins, are also applied on the entry cache.
	Some entries in the entry cache are left in a state that does not
	reflect the DB state. To prevent this mismatch, upon BETXN failure,
	the fix https://pagure.io/389-ds-base/issue/50260 reverts some entries
	in the entry cache .

	The problem is that reversion is not limited to the cases of BETXN
	failures that was the initial goal. So a "regular" error like schema
	violation could trigger revert_cache

Fix description:
	The fix flags if the failure is due to BETXN failures and
	trigger revert_cache only in that case

relates: #5944

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -556,7 +556,12 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
 {
     Hashtable *ht = cache->c_idtable; /* start with the ID table as it's in both ENTRY and DN caches */
     void *e, *laste = NULL;
+    char flush_etime[ETIME_BUFSIZ] = {0};
+    struct timespec duration;
+    struct timespec flush_start;
+    struct timespec flush_end;
 
+    clock_gettime(CLOCK_MONOTONIC, &flush_start);
     cache_lock(cache);
 
     for (size_t i = 0; i < ht->size; i++) {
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1400,9 +1406,10 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
                     slapi_pblock_set(pb, SLAPI_PLUGIN_OPRETURN, ldap_result_code ? &ldap_result_code : &retval);
                 }
                 slapi_pblock_get(pb, SLAPI_PB_RESULT_TEXT, &ldap_result_message);
+                betxn_callback_fails = 1;
```
progier389: Seems like an overkill:
  Why setting a variable to True then testing it just after rather than keeping the old code ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -638,6 +643,11 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
     }
 
     cache_unlock(cache);
+
+    clock_gettime(CLOCK_MONOTONIC, &flush_end);
+    slapi_timespec_diff(&flush_end, &flush_start, &duration);
+    snprintf(flush_etime, ETIME_BUFSIZ, "%" PRId64 ".%.09" PRId64 "", (int64_t)duration.tv_sec, (int64_t)duration.tv_nsec);
+    slapi_log_err(SLAPI_LOG_WARNING, "flush_hash", "Upon BETXN callback failure, entry cache is flushed during %s\n", flush_etime);
```
progier389, line 650: I do not know how frequent this message is displayed in a real deployment 
 and I wonder if WARNING is not a bit too frightening (Maybe INFO would be better )

tbordaz: I had the same concern. Failure of a plugin is not a big deal but revert_cache is so slow (especially with large entry cache) and almost freeze the server (all requests accessing to the entry cache are stuck) that I was thinking the first thing that an admin would do is to check the error log. Making it a warning may match what LDAP clients reports.

progier389 (approved):
LGTM

Issue #5975, opened 2023-10-30
test_replication_with_mod_delete_and_modrdn_operations CI test sometime fails


progier389:
**Issue Description**
test_replication_with_mod_delete_and_modrdn_operations sometime fails
>       test2 = users_s2.get(f"test_user_{USER2_UID}")
E           ldap.NO_SUCH_OBJECT: No object exists given the filter criteria: test_user_2 (&(&(objectclass=account)(objectclass=posixaccount)(objectclass=inetOrgPerson)(objectclass=organizationalPerson))(|(uid=test_user_2)))

https://github.com/389ds/389-ds-base/actions/runs/6696127133/job/18193767783#step:7:500

Problem is that the test case does not rightly wait that the change is replicated
  *it waits that changes from S2 are replicated to S1 instead of the opposite:
    `repl.wait_for_replication(S2, S1)`  should be `repl.wait_for_replication(S1, S2) ` 

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main branch

**Steps to Reproduce**
1. looks at the CI tests results 

**Expected results**
Test should not fails

**Additional context**
Trivial fix

PR #5996, opened 2023-11-29
Issue 5993 - Fix several race condition around CI tests

### LINKED ISSUE #5993: Fix several race condition around CI tests
**Issue Description**
some CI Tests are randomly failing.

 - Platform: Fedora
 - Package and version: main branch (nightly tests)

**Steps to Reproduce**
1. Survey the CI test results for some time.
2. See failure:

**Expected results**
 CI Tests results should be stable.

**Additional context**
 Several different root causes where found and in fact all of them seems related to a change of the tests dynamic (maybe due to having faster test VM):
    - issue #5975 ==> Should 
    - server sometime fails to start/restart (Problem is that systemd default restart rate limit was reached by the LMDB tests) ==> Need to increase the burst threshold.
    - automember plugin failure seems that the task is aborted before its thread started. ==> Need a better synchronization mechanism.
    - test_etime_order_of_magnitude sometime fails. (IMHO this is normal as nothing prevent the thread to be preempted at
        the "wrong" time). ==> Marking the test as flappy
    - Improve failure diagnostic:
        - Do not abort the report if error log is not available.
        - Log "systemctl status" if start fails and no significant error is found in error log.

progier389:
Some CI tests are randomly failing:

Several different root causes where found and in fact all of them seems related to a change of the tests dynamic (maybe due to having faster test VM):
- issue https://github.com/389ds/389-ds-base/issues/5975 ==> Should wait properly for the replication
- server sometime fails to start/restart (Problem is that systemd default restart rate limit was reached by the LMDB tests) ==> Need to increase the burst threshold.
- automember plugin failure (rebuild task sometime finished too fast (before the tested command is run) ==> Retry the test until task is not finished or too many attempt have been done.
- test_etime_order_of_magnitude sometime fails. (IMHO this is normal as nothing prevent the thread to be preempted at
the "wrong" time). ==> Marking the test as flappy
- Improve failure diagnostic:
    - Do not abort the report if error log is not available.
    - Log "systemctl status" if start fails and no significant error is found in error log.

Issue: #5993 

### DIFF
--- a/dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
+++ b/dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
@@ -91,6 +91,7 @@ def test_abort(automember_fixture, topo):
     task.wait()
 
     # Check errors log for abort message
+    log.info('AUTOMEMBER REBUILD TASK: dn=%s %s' % (task.dn, task.get_all_attrs_utf8()))
     assert topo.standalone.searchErrorsLog("task was intentionally aborted")
 
 
--- a/dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
+++ b/dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
@@ -18,6 +18,8 @@ from lib389.topologies import topology_st as topo
 
 log = logging.getLogger(__name__)
 
+# maximum number of retries when rebuild task is too fast.
+MAX_TRIES = 10
 
 @pytest.fixture(scope="module")
 def automember_fixture(topo, request):
@@ -27,7 +29,7 @@ def automember_fixture(topo, request):
 
     # Create users
     users = UserAccounts(topo.standalone, DEFAULT_SUFFIX, rdn=None)
-    NUM_USERS = 1000
+    NUM_USERS = 5000
     for num in range(NUM_USERS):
         num_ran = int(round(num))
         USER_NAME = 'test%05d' % num_ran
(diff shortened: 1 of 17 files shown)


progier389:
ds_logs test_inactivty_and_expiration is still often failing. I do not know why ...

dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
```diff
@@ -81,16 +83,25 @@ def test_abort(automember_fixture, topo):
...
+    task_exit_code = '0'
+    nbtries = 0
+    # Loops if previously run task has completed successfully
+    while task_exit_code == '0':
+        # ensure there is not too many loops.
+        assert nbtries < MAX_TRIES
+        nbtries += 1
+        # Start rebuild task
+        task = automemberplugin.fixup(DEFAULT_SUFFIX, "objectclass=top")
+        time.sleep(1)
```
tbordaz, line 95: 1sec sleep looks good, we could also do an active loop on task entry and abort it immediately.

tbordaz (approved):
LGTM

dirsrvtests/tests/suites/automember_plugin/automember_abort_test.py
progier389: As previously the test was waiting 1 second, I prefer to keep it as it. 
But I agree that we could loop on the task state to ensure it is no more in the setup phase

Issue #5998, opened 2023-11-30
plugins/accpol_check_all_state_attrs_test.py::test_inactivty_and_expiration CI test fails


progier389:
**Issue Description**
Once timing issue #5997 is fixed. the test fails systematically. 
So if the test is really to test that bind is rejected if it occurs after password has expired for longer than the inactivity limit
then the test case is wrong (config issue ?) or there is a bug in the account policy plugin. 

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main + #5997 fix.

**Steps to Reproduce**
1. See nightly CI tests error.

**Expected results**
Test should pass

droideck:
This PR fixes the test: https://github.com/389ds/389-ds-base/pull/6323

PR #5999, opened 2023-11-30
Issue 5997 - test_inactivty_and_expiration CI testcase is wrong

### LINKED ISSUE #5997: plugins/accpol_check_all_state_attrs_test.py::test_inactivty_and_expiration CI testcase is wrong
**Issue Description**
Two issues here:
 [1] plugins/accpol_check_all_state_attrs_test.py::test_inactivty_and_expiration test is sometime failing and it is not testing 
       what it is supposed to test:
          (The instance restart time is often greater than the account inactivity limit so the test is in fact only 
           testing the inactivity) 

 - Platform: Fedora

**Steps to Reproduce**
1. See the random error in nightly CI tests results.

**Expected results**
Test should always pass

**Screenshots**

**Additional context**

progier389:
Problem: test case is not doing what it is supposed to do
because the inactivity limit is often smaller than the server restart time
so in most case the test only checks the account inactivity limit.
But once timing issue are fixed, there is a second issue #5998 (looks like the tested feature does not work as intended!) 
Solution: 
  Increase the inactivity limit to 1 minute
  Make sure we wait enough time to trigger the inactivity limit since last password change but not since last bind.
Mark the test as xfail because of  issue #5998  that is not fixed by this PR 

Issue #5997 

reviewed by: @droideck  (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
+++ b/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
@@ -12,10 +12,19 @@ import pytest
 import os
 import time
 from lib389.topologies import topology_st as topo
-from lib389._constants import DN_CONFIG, DEFAULT_SUFFIX, PLUGIN_ACCT_POLICY, DN_PLUGIN, PASSWORD
+from lib389._constants import (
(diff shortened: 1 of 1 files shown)


droideck (approved):
LGTM!

PR #6005, opened 2023-12-06
Issue 6004 - idletimeout may be ignored

### LINKED ISSUE #6004: idletimeout may be ignored.
**Issue Description**
See https://issues.redhat.com/browse/RHEL-17511

 - Platform: redHat
 - Package and version: 9.4
 

**Steps to Reproduce**
1. Have an application that:
     - open a connection
     - bind as an non directory manager user
     - wait longer than idletimeout
2. run: dsconf instance monitor server 
3. See error: number of open connections is 2 instead of 1 

**Expected results**
Connection should be closed

**Additional context**
looks like the timeout is only handled when there are some activity on the other connections.

progier389:
Problem: idletimeout is still not handled when binding as non root (unless there are some activity
    on another connection)
Fix:
    Add a slapi_eq_repeat_rel handler that walks all active connection every seconds and check if the timeout is expired.
Note about CI test:
    Notice that idletimeout is never enforced for connections bound as root (i.e cn=directory manager). 

Issue #6004 

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -921,6 +921,17 @@ slapd_sockets_ports_free(daemon_ports_t *ports_info)
 }
 
 /*
+ * Tells if idle timeout has expired
+ */
+static inline int __attribute__((always_inline))
+has_idletimeout_expired(Connection *c, time_t curtime)
+{
+    return (c->c_state != CONN_STATE_FREE && !c->c_gettingber &&
+            c->c_idletimeout > 0 && NULL == c->c_ops &&
+            curtime - c->c_idlesince >= c->c_idletimeout);
+}
+
+/*
  * slapi_eq_repeat_rel callback that checks that idletimeout has not expired.
  */
 void
@@ -932,9 +943,7 @@ check_idletimeout(time_t when __attribute__((unused)), void *arg __attribute__((
     for (int list_num = 0; list_num < ct->list_num; list_num++) {
         for (Connection *c = connection_table_get_first_active_connection(ct, list_num);
              c != NULL; c = connection_table_get_next_active_connection(ct, c)) {
-            if (c->c_state == CONN_STATE_FREE || c->c_gettingber ||
-                c->c_ops != NULL || c->c_idletimeout <= 0 ||
-                curtime - c->c_idlesince < c->c_idletimeout) {
+            if (!has_idletimeout_expired(c, curtime)) {
                 continue;
             }
             /* Looks like idletimeout has expired, lets acquire the lock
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/daemon.c
```diff
@@ -1123,6 +1160,9 @@ slapd_daemon(daemon_ports_t *ports)
                   "MAINPID=%lu",
                   (unsigned long)getpid());
 #endif
+    slapi_eq_repeat_rel(check_idletimeout, NULL,
+                        slapi_current_rel_time_t(),
+                        MILLISECONDS_PER_SECOND);
```
droideck, line 1172: I wonder if it's not too often... (scanning all of the connections every second)

Probably, it makes sense to see that there is no performance drop in situations with many user connections.

ldap/servers/slapd/daemon.c
```diff
@@ -918,6 +920,41 @@ slapd_sockets_ports_free(daemon_ports_t *ports_info)
...
+                curtime - c->c_idlesince < c->c_idletimeout) {
+                continue;
+            }
+            /* Looks like idletimeout has expired, lets acquire the lock
+             * and double check.
+             */
+            if (pthread_mutex_trylock(&(c->c_mutex)) == EBUSY) {
+                continue;
+            }
+            if (c->c_state != CONN_STATE_FREE && !c->c_gettingber &&
```
tbordaz: This test is looking the opposite (!) of the test above (that skip this connection). It would be good to have a single macro/function.

ldap/servers/slapd/daemon.c
```diff
@@ -918,6 +920,41 @@ slapd_sockets_ports_free(daemon_ports_t *ports_info)
 #endif
 }
 
+/*
+ * slapi_eq_repeat_rel callback that checks that idletimeout has not expired.
+ */
+void
```
tbordaz, line 937: I like this approach but I wonder why the previous approach is no longer working.
In slapd_deamon we are looping to call setup_pr_read_pds with a timeout (1s) on poll. 
Is idletime not longer tested because this function is no longer used ?

tbordaz (approved):
Patch looks good.
few comments

ldap/servers/slapd/daemon.c
progier389: I doubt we will notice anything: 
   During stress test the listener threads are walking all the active connections every tenth of ms (or even faster)  to build and check the polling table so walking it every second should have an impact below 0.01%

ldap/servers/slapd/daemon.c
progier389: setup_pr_read_pds is still called (by ct_list_thread) at least every seconds but it does not test for idle timeout !
  If you grep c_idletimeout in previous code, you see that the only place it is tested is in  handle_pr_read_ready 
    and ct_list_thread does not call handle_pr_read_ready if the poll timed out

ldap/servers/slapd/daemon.c
progier389: Added an inline function

Issue #5939, opened 2023-09-28
During an update, if the target entry is reverted in the entry cache, the server should not retry to lock it.


tbordaz:
**Issue Description**
When an entry is reverted in the entry cache, (reverted means a previous update flagged it as INVALID), it can stay there for some time until refcnt is 0. During that period, an update targeting that entry will systematically fails.
Instead of looping for an CPU intensive retry attempts. The update should fail immediately because it contributes to keep the entry in the entry cache (refcnt).

The signature of this issue is a message in the error log: "Retry count exceeded"

**Package Version and Platform:**
Since 1.4.3 (https://pagure.io/389-ds-base/issue/50260)

**Steps to Reproduce**
To be provided

**Expected results**
Message "Retry count exceeded" should be extremely rare.

tbordaz:
b9726faae..1ab0a0924  main
f9d09a907..75b026b4a  389-ds-base-2.4
445a881d9..5ca046516  389-ds-base-2.3
e2ed3561d..500e70717  389-ds-base-2.2
27904223d..da5e78ec9  389-ds-base-2.1
37868719e..fbfdb4ed9  389-ds-base-2.0
ab5e8e484..001ad7978  389-ds-base-1.4.4
77bfdfe4c..0fae6cae9  389-ds-base-1.4.3

PR #6007, opened 2023-12-11
Issue 5939 - During an update, if the target entry is reverted in the…

tbordaz:
… entry cache, the server should not retry to lock it

Bug description:
	During an update if an BETXN plugin fails the full TXN is aborted and the DB
	returns to the previous state.
	However potential internal updates, done by BETXN plugins, are also applied
	on the entry cache.
	Even if the TXN is aborted some entries in the entry cache are left in a state
	that does not reflect the DB state.
	The fix https://pagure.io/389-ds-base/issue/50260 "reverts" those
	entries, setting their state to INVALID.

	A problem is that reverted entries stay in the entry cache, until refcnt is 0.
	During that period, an update targeting that entry fails to retrieve the
	entry from the entry cache and fails to add it again as it already exist
	the entry.
	The update iterates 1000 times, trying to read the entry and to fetch it
	from DB.
	This is a pure waste as the reverted entry stays too long.

	The signature of this issue is a message in the error log: "Retry count exceeded"

Fix description:
	The fix consiste in the loops (fetch on DN or NSUNIQUEID) to test if the
        entry state is INVALID.
	In such case it aborts the loop and return a failure.

relates: #5939

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/findentry.c
+++ b/ldap/servers/slapd/back-ldbm/findentry.c
@@ -98,6 +98,7 @@ find_entry_internal_dn(
     size_t tries = 0;
     int isroot = 0;
     int op_type;
+    int reverted_entry = 0;
 
     /* get the managedsait ldap message control */
     slapi_pblock_get(pb, SLAPI_MANAGEDSAIT, &managedsait);
@@ -141,11 +142,19 @@ find_entry_internal_dn(
          */
         slapi_log_err(SLAPI_LOG_ARGS, "find_entry_internal_dn",
                       "   Retrying (%s)\n", slapi_sdn_get_dn(sdn));
+        if (cache_is_reverted_entry(&inst->inst_cache, e)) {
+            reverted_entry = 1;
+            break;
+        }
         CACHE_RETURN(&inst->inst_cache, &e);
         tries++;
     }
     if (tries >= LDBM_CACHE_RETRY_COUNT) {
         slapi_log_err(SLAPI_LOG_ERR, "find_entry_internal_dn", "Retry count exceeded (%s)\n", slapi_sdn_get_dn(sdn));
+    }
+    if (reverted_entry) {
+        slapi_send_ldap_result(pb, LDAP_BUSY, NULL, "target entry busy because of a canceled operation", 0, NULL);
+        return (NULL);
     }
     /*
      * there is no such entry in this server. see how far we
@@ -262,6 +271,7 @@ find_entry_internal_uniqueid(
     struct backentry *e;
     int err;
     size_t tries = 0;
+    int reverted_entry = 0;
 
     while ((tries < LDBM_CACHE_RETRY_COUNT) &&
            (e = uniqueid2entry(be, uniqueid, txn, &err)) != NULL) {
@@ -283,6 +293,10 @@ find_entry_internal_uniqueid(
          */
         slapi_log_err(SLAPI_LOG_ARGS,
                       "   find_entry_internal_uniqueid", "Retrying; uniqueid = (%s)\n", uniqueid);
+        if (cache_is_reverted_entry(&inst->inst_cache, e)) {
+            reverted_entry = 1;
+            break;
+        }
         CACHE_RETURN(&inst->inst_cache, &e);
         tries++;
     }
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -262,6 +271,7 @@ find_entry_internal_uniqueid(
     struct backentry *e;
     int err;
     size_t tries = 0;
+    int reverted_entry = 0;
 
     while ((tries < LDBM_CACHE_RETRY_COUNT) &&
```
progier389, line 276: Should we not reset reverted_entry in the while loop ?

ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -292,9 +306,14 @@ find_entry_internal_uniqueid(
                       uniqueid);
     }
 
-    /* entry not found */
-    slapi_send_ldap_result(pb, (0 == err || DBI_RC_NOTFOUND == err) ? LDAP_NO_SUCH_OBJECT : LDAP_OPERATIONS_ERROR, NULL /* matched */, NULL,
-                           0, NULL);
+    if (reverted_entry) {
+        slapi_send_ldap_result(pb, LDAP_BUSY, NULL, NULL, 0, NULL);
```
progier389: IMHO We should add a rexr message to explain why LDAP_BUSY was returned. 
 (i.e:  slapi_send_ldap_result(pb, LDAP_BUSY, "target entry busy because of a cancelled operation.", NULL, 0, NULL);

ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -98,6 +98,7 @@ find_entry_internal_dn(
     size_t tries = 0;
     int isroot = 0;
     int op_type;
+    int reverted_entry = 0;
 
     /* get the managedsait ldap message control */
     slapi_pblock_get(pb, SLAPI_MANAGEDSAIT, &managedsait);
```
progier389, line 104: Should we not reset the reverted_entry flag in the while loop ?

ldap/servers/slapd/back-ldbm/findentry.c
```diff
@@ -141,12 +142,20 @@ find_entry_internal_dn(
...
+            break;
+        }
         CACHE_RETURN(&inst->inst_cache, &e);
         tries++;
     }
     if (tries >= LDBM_CACHE_RETRY_COUNT) {
         slapi_log_err(SLAPI_LOG_ERR, "find_entry_internal_dn", "Retry count exceeded (%s)\n", slapi_sdn_get_dn(sdn));
     }
+    if (reverted_entry) {
+        slapi_send_ldap_result(pb, LDAP_BUSY, NULL, NULL, 0, NULL);
```
progier389: Should add text to explain we return BUSY

ldap/servers/slapd/back-ldbm/findentry.c
tbordaz: A reverted entry may be the target of several requests that started before it got reverted. Those req increased the refcnt and we have to wait that all requests call cache_return. The last cache_return will reset and remove the entry from the cache.

ldap/servers/slapd/back-ldbm/findentry.c
tbordaz: yes good point

ldap/servers/slapd/back-ldbm/findentry.c
tbordaz: see response above. In short it is the job of cache_return to reset/free the entry when refcnt==0

progier389 (approved):
LGTM

droideck (approved):
The code looks good, but I think we need to make it flaky... (it has already failed and I've rerun it once)

```
>       assert topology_st.standalone.ds_error_log.match('.*cache_is_reverted_entry - Entry reverted.*')
E       AssertionError: assert []
E        +  where [] = <bound method DirsrvLog.match of <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>>('.*cache_is_reverted_entry - Entry reverted.*')
E        +    where <bound method DirsrvLog.match of <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>> = <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>.match
E        +      where <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0> = <lib389.DirSrv object at 0x7f7dce59f920>.ds_error_log
E        +        where <lib389.DirSrv object at 0x7f7dce59f920> = <lib389.topologies.TopologyMain object at 0x7f7dce59f950>.standalone
```

tbordaz:
> The code looks good, but I think we need to make it flaky... (it has already failed and I've rerun it once)
> 
> ```
> >       assert topology_st.standalone.ds_error_log.match('.*cache_is_reverted_entry - Entry reverted.*')
> E       AssertionError: assert []
> E        +  where [] = <bound method DirsrvLog.match of <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>>('.*cache_is_reverted_entry - Entry reverted.*')
> E        +    where <bound method DirsrvLog.match of <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>> = <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0>.match
> E        +      where <lib389.dirsrv_log.DirsrvErrorLog object at 0x7f7dce46b2f0> = <lib389.DirSrv object at 0x7f7dce59f920>.ds_error_log
> E        +        where <lib389.DirSrv object at 0x7f7dce59f920> = <lib389.topologies.TopologyMain object at 0x7f7dce59f950>.standalone
> ```

Indeed the test relies on the dynamic of a entry state and it is not a surprise that it sometime fails. I change the test to flaky and in addition did more parallel (failing) updates that made the test almost always successful on my laptop.

Issue #6010, opened 2023-12-12
389 ds ignores nsslapd-maxdescriptors


resposit:
**Issue Description**
389 ds 2.3.6 seems to ignore nsslapd-maxdescriptors=32768

**Package Version and Platform:**
 - Platform: RockyLinux 9.3
 - Package and version: 389-ds-base-2.3.6-3.el9_3.x86_64

**Steps to Reproduce**
In /etc/systemd/system/dirsrv@cloud20.service.d/override.conf I have:
```
[Service]
LimitNOFILE=32768
LimitNPROC=16384
```

In dse.ldif I have: 
```
nsslapd-maxdescriptors: 32768
nsslapd-reservedescriptors: 16384
```
When start directory server:
` systemctl start dirsrv@cloud20.service`

 I get the error:
`[12/Dec/2023:13:00:09.358667991 +0100] - ERR - slapd_daemon - Not enough descriptors to accept any connections. This may be because the maxdescriptors configuration directive is too small, the hard limit on descriptors is too small (see limit(1)), or the reservedescriptors configuration directive is too large. Try increasing the number of descriptors available to the slapd process. The current value is 16384. 16384 descriptors are currently reserved for internal slapd use, so the total number of descriptors available to the process must be greater than 16384.
`
It seems 389 ds ignores the setting nsslapd-maxdescriptors=32768 and considers nsslapd-maxdescriptors=16384. In fact I change in dse.ldif:
```
nsslapd-maxdescriptors: 32768
nsslapd-reservedescriptors: 8192
```
Then the server starts properly. Strangely it then reports the correct nsslapd-maxdescriptors value:
```
[root@cloud20 etc]# dsconf -D "cn=Directory Manager" ldap://localhost config get nsslapd-maxdescriptors
Enter password for cn=Directory Manager on ldap://localhost:
nsslapd-maxdescriptors: 32768
```

**Additional context**
I have a twin server (same configuration) running 389-ds-base-2.2.9-2.el8.x86_64 and I have no such issue

progier389:
Indeed nsslapd-maxdescriptors is no more used ( that is the nofile hard limit that matters) 
In fact I suspect that your problem is not about the nsslapd-maxdescriptors but that the nsslapd-reservedescriptors meaning has unexpectedly changed.
    In older version it was the number of file descriptors used for the db, log files, import/export file, ... but NOT for the connections sockets while now it seems to also include the sockets 
[...]

PR #6019, opened 2023-12-21
Issue 6010 - 389 ds ignores nsslapd-maxdescriptors 

jchapma:
Bug description: During server startup the connection table size is assumed
to be lower than or equal to the number of configured reserve file descriptors.
This prevents the server from starting whem the number of reserve descriptors
is high.

Fix description: Change the check to make sure the connection table size is
not greater than (max descriptors - reserve descriptors).

Also, the number of reserve descriptors is used to determine if the server can
accept a new connection. This has been changed to compare the connection table
size against the current number of connections.

Relates: https://github.com/389ds/389-ds-base/issues/6010


progier389 (approved):
LGTM

jchapma:
Apologies @progier389 

Im not sure why I didn't create this PR to merge into main and not 2.3. I will create another PR and cherry pick to relevant branches.

PR #6023, opened 2024-01-02
str2entry_dupcheck not all errors are logged #2182

kimettog:
Issue ##### - <SUMMARY>

Bug Description:

str2entry_dupcheck not all errors are logged #2182

Fix Description:

A number of error cases in str2entry_dupcheck are only logged at LOG_TRACE, when they should be logged at an error level (WARNING, ERR). This can cause silent failures of imports.

relates: #2182 

Author: Gilbert Kimetto

Reviewed by: Simon Pichugi


dirsrvtests/tests/suites/import/import_warning_error_test.py
```diff
@@ -0,0 +1,134 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2021 Red Hat, Inc.
```
droideck: The date is wrong

dirsrvtests/tests/suites/import/import_warning_error_test.py
```diff
@@ -0,0 +1,134 @@
...
+    backend_import(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+
+    log.info('Check logs for a warning message')
+    assert topology_st.logcap.contains(message)
+    str2_dup_err_warn_log_lst = ([warn_log for warn_log in (standalone.ds_error_log.readlines()) if "str2entry_dupcheck" in warn_log])
+    stripped_err_warn_lst = set ([s[69:] for s in str2_dup_err_warn_log_lst])
+    _lst = [exp_msg_dup_dn,exp_msg_dup_uid,exp_msg_dup_ou_dn]
+    for msg in _lst:
+        log.info(f"Checking for {msg} warning in Logs")
+        assert msg in stripped_err_warn_lst,(f"Test Failed - Unable to find {msg}.")
```
droideck, line 178: In the GitHub CI, the test fails here:

```
>           assert msg in stripped_err_warn_lst,(f"Test Failed - Unable to find {msg}.")
E           AssertionError: Test Failed - Unable to find Entry has multiple dns "uid=demo,ou=People,dc=example,dc=com" and "uid=demo,ou=People,dc=example,dc=com" (second ignored)
E             .
E           assert 'Entry has multiple dns "uid=demo,ou=People,dc=example,dc=com" and "uid=demo,ou=People,dc=example,dc=com" (second ignored)\n' in {'Entry has multiple dns "dc=example,dc=com" and "dc=example,dc=com" (second ignored)\n', 'Entry has multiple uniqueids e5c4172a-97aa11eb-aaa8e47e-b1e12808 and e5c4172a-97aa11eb-aaa8e47e-b1e12808 (second ignored)\n'}
```

ldap/servers/slapd/entry.c
```diff
@@ -775,15 +775,13 @@ str2entry_dupcheck(const char *rawdn, const char *s, int flags, int read_statein
     char *normdn = NULL;
     int strict = 0;
     struct berval bval = {0};
-
+
```
droideck: Typo

ldap/servers/slapd/entry.c
```diff
@@ -1197,7 +1195,7 @@ str2entry_dupcheck(const char *rawdn, const char *s, int flags, int read_statein
     if (e->e_flags & SLAPI_ENTRY_FLAG_TOMBSTONE) {
         /* tombstone */
         if (_entry_set_tombstone_rdn(e, slapi_entry_get_dn_const(e))) {
-            slapi_log_err(SLAPI_LOG_TRACE, "str2entry_dupcheck",
+            slapi_log_err(SLAPI_LOG_ERR, "str2entry_dupcheck",
```
droideck, line 1204: I have a doubt that we need to change it for tombstones as it can be less useful for the administrator in the end.
@Firstyear @tbordaz @progier389 @jchapma what do you think?

droideck (changes requested):

ldap/servers/slapd/entry.c
progier389: I think that logging the tombstone error at ERR level is OK because:
  - There should be no error unless something is seriously amiss (like a tampered import file) so increasing the log level does not generate noise.
  - Failing to import tombstone may lead to serious replication problems (broken replication or divergent entries among the replicas) so I do not think that such error should be silent.

dirsrvtests/tests/suites/import/import_warning_error_test.py
kimettog: Ack will fix. thank you! :+1:

ldap/servers/slapd/entry.c
kimettog: Ack. Fixed

dirsrvtests/tests/suites/import/import_warning_error_test.py
kimettog: Ah I see the problem. Thanks! Fixed

droideck (changes requested):
Hmm, I've tried to import perfectly fine LDIF, and it gave me a lot of false warnings...
Steps to reproduce:

1. Install an instance with sample entries
2. Export a suffix from UI (Database - Suffixes - Suffix Tasks - Export Suffix) - Name the ldif.
3. Import the suffix from UI (Database - Suffixes - Suffix Tasks - Initialise Suffix) - Choose the ldif.
4. Check the logs for WARNs.

Could you please add another test that will check for clean export-import?
And we need to ensure that the logs are not filled with false warnings.

droideck:
We'll continue this work once we get back to #2182. The code from this PR will probably be reused, with credit to @kimettog.

PR #6026, opened 2024-01-05
Issue 6022 - lmdb inconsistency between vlv index and vlv cache names

### LINKED ISSUE #6022: lmdb inconsistency between vlv index sub-database name and vlv cache sub-database name
**Issue Description**

Theoretically when vlv are used on lmdb, there should be one vlv index cache sub-database per vlv index sub-database.
But after configuring a freeipa server and two replica, sometime there are two caches sub-db (one with the bename and one without). The dbname  should always be present. 

 - Platform: Fedora

**Steps to Reproduce**
1. Install 3 virtual machine running with Fedora 39 and freeipa and 389ds 3.0. 
2. Ensure that hostname and /etc/hosts are set properly (so that all hosts  
3. Setup 1 host as freeipa master and two as replicas
4. Run dbscan -L _dbhomedir_path_ > output
5. grep -c  "vlv#" output
6. grep -c recno output
7. Compare steps [5.] and [6.] results
8. See error

**Expected results**
steps [5.] result should be the double of steps [6.] 

vlv index sub db should looks like: /var/lib/dirsrv/slapd-IPA-TEST/db//ipaca/vlv#allcertspkitomcatindex.db
vlv index cache sub-db should looks like:  
  /var/lib/dirsrv/slapd-IPA-TEST/db//~recno-cache/ipaca/vlv#allcertspkitomcatindex.db
but for 2 of the 35 vlv indexes I also see something like:
  /var/lib/dirsrv/slapd-IPA-TEST/db//~recno-cache//vlv#allcertspkitomcatindex.db flags: 
(with an empty bename)

So its looks like that sometime the bename is not filled.

**Additional context**
I will have to dig  to understand when the cache without bename get created. 
  I suspect it is after an on-line or off-line reindex or import (or maybe bulk-import ?)

progier389:
Problem: dbstat -L shows two vlv cache db for a single vlv index db. 
There should only have a single one. 

Fix: 
 Added a CI Test
 Using a single dbmdb_recno_cache_get_dbname function to get the cache db name.
 Fix dbmdb_build_dbname to also append the backend name if the name is a vlv cache

Also fixed some issue found while creating the CI test:
  Fixed an error message that puzzled me to make it clearer.
  Fixed a race condition in lmdb bulk import that logged crappy data in error logs and crashed the CI tests.
  
Issue: #6022 

Reviewed by: @droideck (Thanks !)

### DIFF
--- a/dirsrvtests/tests/suites/vlv/regression_test.py
+++ b/dirsrvtests/tests/suites/vlv/regression_test.py
@@ -16,7 +16,7 @@ from lib389.properties import TASK_WAIT
 from lib389.index import *
 from lib389.mappingTree import *
 from lib389.backend import *
-from lib389.idm.user import UserAccounts
+from lib389.idm.user import UserAccounts, UserAccount
 from ldap.controls.vlv import VLVRequestControl
 from ldap.controls.sss import SSSRequestControl
 from lib389.replica import ReplicationManager
@@ -28,6 +28,15 @@ pytestmark = pytest.mark.tier1
 
 logging.getLogger(__name__).setLevel(logging.DEBUG)
 log = logging.getLogger(__name__)
+
+STARTING_UID_INDEX = 1000
+# VLV search always returns first the entry before the target,
+# and the entry uid=demo_user,ou=people,dc=example,dc=com
+# is before the uid=testuserNNNN,dc=example,dc=com entries
+# so VLV_SEARCH_OFFSET (The difference between the vlv index
+# and the NNNN value) is:
+VLV_SEARCH_OFFSET = STARTING_UID_INDEX - 2
+NUMUSERS_TEST_VLV_REINDEX = 5000
 
 
 def open_new_ldapi_conn(dsinstance):
@@ -58,8 +67,8 @@ def check_vlv_search(conn, **kwargs):
         filterstr='(uid=*)',
         serverctrls=[vlv_control, sss_control]
     )
-    imin = offset + 998 - before_count
-    imax = offset + 998 + after_count
+    imin = offset + VLV_SEARCH_OFFSET - before_count
+    imax = offset + VLV_SEARCH_OFFSET + after_count
 
     for i, (dn, entry) in enumerate(result, start=imin):
         assert i <= imax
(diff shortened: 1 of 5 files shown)


progier389:
Got a nasty surprise: Also the test is passing as a single test but test fails when all the vlv tests are run. 
   First issue was that a simple cleanup was needed to remove data from previous testcase 
      but once that fixed the test is still failing: now some of the entries returned by the vlv are not the expected one. 
==> review is on hold until that issue is solved.

progier389:
I discovered that the vlv inconstancy I described in previews  comments also exists with bdb so it is not related to the current issue. Created issue #6028 to track it. Restart works around that issue.
So it is ok to review.

dirsrvtests/tests/suites/vlv/regression_test.py
```diff
@@ -246,6 +259,103 @@ def test_vlv_recreation_reindex(topology_st):
...
+    entries = [ UserAccount(M1, dn=f'uid=testuser{uid},DEFAULT_SUFFIX') for uid in
+        range(STARTING_UID_INDEX, STARTING_UID_INDEX+NUMUSERS_TEST_VLV_REINDEX) ]
+    entries.append(VLVIndex(M2, dn='cn=vlvIdx,cn=vlvSrch,cn=userRoot,cn=ldbm database,cn=plugins,cn=config'))
+    entries.append(VLVSearch(M2, dn='cn=vlvSrch,cn=userRoot,cn=ldbm database,cn=plugins,cn=config'))
+    for entry in entries:
+        if entry.exists():
+            entry.delete()
+    repl.wait_for_replication(M1, M2)
+    # Restart the instance to workaround github issue #6028
+    M2.restart()
```
droideck: If I understood correctly, the workaround is required only for BDB. Should we restart only for BDB tests, then? (by adding some if here or something)

Also, kind of nitpick, I think it'll be useful to have a full link here (`https://github.com/389ds/389-ds-base/issues/6028`) in case we'll change the bug tracker somewhere in the future.

droideck (approved):
Besides the minor issue, looks good!

dirsrvtests/tests/suites/vlv/regression_test.py
progier389: The new testcase is still failing, and I realized that I made a mistake when creating issue 6028 
  (used wrong test case in the reproducer) 
Redoing the test I noticed that the inconsistency occurs only in mdb and the hang only in bdb 
So I fixed 6028 description for mdb inconsistency and created #6029 for the bdb hang.
I changed the code to do as you suggested.
  use full link to  https://github.com/389ds/389-ds-base/issues/6029 and restart M2 only in bdb case
  
Note the new test case is still failing with mdb but fixing  6028 should solve it.

PR #6027, opened 2024-01-07
Issue 6010 - 389 ds ignores nsslapd-maxdescriptors

jchapma:
Bug description: During server startup the connection table size is assumed to be lower than or equal to the number of configured reserve file descriptors. This prevents the server from starting whem the number of reserve descriptors is high.

Fix description: Change the check to make sure the connection table size is not greater than (max descriptors - reserve descriptors).

Also, the number of reserve descriptors is used to determine if the server can accept a new connection. This has been changed to compare the connection table size against the current number of connections.

Relates: https://github.com/389ds/389-ds-base/issues/6010

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
+++ b/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
@@ -136,9 +136,9 @@ def test_reserve_descriptors_high(topology_st):
     :expectedresults:
         1. Success
         2. Value of SYSTEMD_LIMIT is returned
-        2. Success
-        3. Values of SYSTEMD_LIMIT -2 is returned
-        4. Instance starts correctly
+        3. Success
+        4. Values of SYSTEMD_LIMIT -2 is returned
+        5. Instance starts correctly
     """
 
     # Set nsslapd-maxdescriptors to a custom value
--- a/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
+++ b/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
@@ -83,7 +83,7 @@ def test_fd_limits(topology_st):
 def test_reserve_descriptor_validation(topology_st):
     """Test the reserve descriptor self check
 
-    :id: TODO
+    :id: 9bacdbcc-7754-4955-8a56-1d8c82bce274
     :setup: Standalone Instance
     :steps:
         1. Set attr nsslapd-reservedescriptors to a low value of RESRV_DESC_VAL (10)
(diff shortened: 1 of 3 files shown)


progier389 (approved):
LGTM

dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
```diff
@@ -121,6 +121,41 @@ def test_reserve_descriptor_validation(topology_st):
...
+    :steps:
+        1. Set attr nsslapd-maxdescriptors to systemd LimitNOFILE value
+        2. Verify value has been set
+        3. Set attr nsslapd-reservedescriptors to 2 less than nsslapd-maxdescriptors
+        4. Verify value has been set
+        5. Restart instance
+    :expectedresults:
+        1. Success
+        2. Value of SYSTEMD_LIMIT is returned
+        2. Success
```
droideck: Validation GitHub CI test fails here (the repeated number).

droideck (approved):

droideck (changes requested):
Hmm, I see that basic test - test_conn_limits fails now. I think it can be related to this PR.

progier389:
Good catch @droideck !  test_conn_limits sets a small value for the max number of file descriptors and try to fill up the connection table, so it is quite probable that it need to be adapted ...

jchapma:
I had a patch for the two issues mentioned here, but didn't get a chance to push till now.

ldap/servers/slapd/conntable.c
```diff
@@ -122,7 +122,8 @@ connection_table_new(int table_size)
     ct->list_num = config_get_num_listeners();
     ct->num_active = (int *)slapi_ch_calloc(1, ct->list_num * sizeof(int));
     ct->size = table_size - (table_size % ct->list_num);
-    ct->list_size = ct->size/ct->list_num;
+    /* Account for first slot of each list being used as head (c_next).  */
+    ct->list_size = (ct->size/ct->list_num)+1;
```
jchapma, line 126: There was a bug when we were mapping the connection table lists to the freelist, as slot 0 of each ct list is used as head of the linked list it was skipped,  resulting in each list being 1 less than it should be. 

This became apparent when the freelist offset is used to determine if the server has room for a new connection.

ldap/servers/slapd/daemon.c
```diff
@@ -831,7 +830,7 @@ accept_thread(void *vports)
 
     while (!g_get_shutdown()) {
         /* Do we need to accept new connections? */
-        int accept_new_connections = ((ct->size - g_get_current_conn_count()) > slapdFrontendConfig->reservedescriptors);
+        int accept_new_connections = (ct->size > ct->conn_next_offset);
```
jchapma, line 833: IMHO a better way for the accept_thread to know if there is "room" for a new connection is to use the freelist offset. A downside of using the freelist offset is that it is protected by a lock, but  I didn't want to lock and unlock in the accept_thread loop. 

Apart from the accept_thread, the only other user of the freelist offset are the connection table list threads, where they move connections back onto the freelist, decrementing the offset. So in a worse case scenario the offset can be reduced by a max of 1, meaning a single connection slot could be unused, IMHO this is better than adding the overhead of locking/unlocking here. WDYT

ldap/servers/slapd/daemon.c
```diff
@@ -2083,7 +2082,7 @@ unfurl_banners(Connection_Table *ct, daemon_ports_t *ports, PRFileDesc **n_tcps,
     char addrbuf[256];
     int isfirsttime = 1;
 
-    if (ct->size <= slapdFrontendConfig->reservedescriptors) {
+    if (ct->size > (slapdFrontendConfig->maxdescriptors - slapdFrontendConfig->reservedescriptors)) {
```
jchapma, line 2085: This code is hit on startup, so we can only use the static descriptor values

droideck (approved):
Looks good, and the tests pass!

If @progier389 also satisfied with the new changes, I think we are good to go!

ldap/servers/slapd/daemon.c
tbordaz: IIUC I agree with your proposal. However a side effect when connection table is close to be full is that we may erroneously believe the CT is full AND wait for 250ms before accepting a new connection.
In case we "believe" there is no room, (ct->size <= ct->conn_next_offset), we may double check the value of conn_next_offset (holding the lock) before sleeping. Most of the time we will not have to double check, so it will benefit from the fix and the lock payload is only when CT is almost full.

PR #6034, opened 2024-01-16
Flag that replica_id is int not str (Fixes #6033)

### LINKED ISSUE #6033: dscreate create-template claims replica_id is (str)
**Issue Description**
Generating a sample template using `dscreate create-template` gives the impression that replica_id can be a string

```
> dscreate  create-template | grep -A3 "^# replica_id"
# replica_id (str)
# Description: Set the unique replication identifier for this replica's database (suppliers only)
# Default value: 1
;replica_id = 1
```

It should be "replica_id (int)"

 - Platform: Opensuse Leap 15.5
 - Package and version: 389-ds-2.2.8~git51.3688d68-150500.3.14.1.x86_64

**Steps to Reproduce**
dscreate  create-template 

**Expected results**
```
# replica_id (int)
# Description: Set the unique replication identifier for this replica's database (suppliers only)
# Default value: 1
;replica_id = 1
```

**Additional context**
setting this to a string (yes, I know the default is 1) crashes the dscreate later when importing template
```
    rid_num = int(rid)
ValueError: invalid literal for int() with base 10: 'polaris'
```

Elwell:


vashirov:
@Elwell, could you please update the commit message according to https://www.port389.org/docs/389ds/contributing.html? I.e. bug description and fix description, even though it's trivial. It would greatly help, thanks!

vashirov (approved):

PR #6035, opened 2024-01-17
Issue 6032 - Replication is broken after backup restore

### LINKED ISSUE #6032: Replication broken after backup restore with freeipa configuration
**Issue Description**
Replication never synchronize after DS restore from a previous backup

 - Platform:  Fedora
 - Package and version: 3.0.0

**Steps to Reproduce**
1. setup a freeipa server and a freeipa replica
2. (on server) dsconf IPA-TEST backup create
3. (on server) kinit admin
4. (on server) ipa user-add test1_supplier --first test --last user
5. (on replica) kinit admin
6. (on replica) ipa user-add test1_replica --first test --last user
7. Wait until replication is in sync
8. dsconf IPA-TEST backup restore /var/lib/dirsrv/slapd-IPA-TEST/bak/IPA-TEST-*
9. Wait a few minutes to let the time for the replica to replay its changes toward the server
10. dsctl IPA-TEST  healthcheck
11. See Error: The replication agreement (metoipareplica.ipa.test) under "dc=ipa,dc=test" is not in synchronization.

**Expected results**
Replication should be able to synchronize

**Additional context**
dsconf IPA-TEST repl get-ruv --suffix dc=ipa,dc=test
Show that the server maxcsn is higher than the replica one.

progier389:
Replication is broken after doing an offline backup then later on an online or offline restore
Note: with online backup changelog is discarded at restore time (because it has no purge RUV)
In fact there are multiple cause:
  [1] _cl5CICbInit is building wrongly the changelog RUVs so changelog is recreated
  [2] Changelog is not cleared when it is "Recreated because of wrong test in dbmdb_back_ctrl
  [3] Replication keep alive get created before the replica get back in sync. This creates missing csn.
 Solution:
    [1] Fix _cl5CICbInit to get the csn from the changelog record key and store properly the min and max in the context.
    [2] Replace invalid test by a proper one.
    [3] Change keep alive update starting delay from 2 seconds to 10 minutes (i.e twice the maximum backoff timeout)
         To let a chance for the other supplier to replay the missing changes.
  Also added/modified some more data when replication log are enabled
  Note: this is a partial fix as a proper "resync after db reload" is not handled so this left issues (typically because
    of the plugin internal operations like memberof plugin or if there are lots of changes to replay) but at least is is enough for the CI test ...
   
Issue: #6032 

### DIFF
--- a/dirsrvtests/tests/suites/backups/backup_test.py
+++ b/dirsrvtests/tests/suites/backups/backup_test.py
@@ -9,15 +9,17 @@
 import logging
 import pytest
 import os
+import shutil
 from datetime import datetime
 from lib389._constants import DEFAULT_SUFFIX, INSTALL_LATEST_CONFIG
 from lib389.properties import BACKEND_SAMPLE_ENTRIES, TASK_WAIT
-from lib389.topologies import topology_st as topo
+from lib389.topologies import topology_st as topo, topology_m2 as topo_m2
 from lib389.backend import Backend
 from lib389.tasks import BackupTask, RestoreTask
 from lib389.config import BDB_LDBMConfig
 from lib389 import DSEldif
 from lib389.utils import ds_is_older, get_default_db_lib
+from lib389.replica import ReplicationManager
 import tempfile
 
 pytestmark = pytest.mark.tier1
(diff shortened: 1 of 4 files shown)


droideck (approved):
Looks good!
`2*PROTOCOL_BACKOFF_MAXIMUM,` part feels a bit rough but, IIUC, you said you work on a fix that will deal with the "resync after restore" and this will be reverted later.
And anyway, the `2*PROTOCOL_BACKOFF_MAXIMUM,`  is still more stable than nothing :)

Ack from me.

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -2589,7 +2589,7 @@ _cl5CICbInit(dbi_val_t *key, dbi_val_t *data, DBLCI_CTX *dblcictx)
         return DBI_RC_SUCCESS;
     }
     /* Update last csn */
-    csn_init_by_string(&dblcictx->csn, data->data);
+    csn_init_by_string(&dblcictx->csn, key->data);
```
tbordaz, line 2592: I understand that fix but I can not understand how it have been working before.
_cl5Iterate is used for Trimming, cleanRuV,.... We know it works. But IIUC it was setting the CL iterator with crappy csn (data instead of key). It looks it worked to access the first CL entry but on next entries,  'startcsn' in _cl5Iterate should have been completely random !
I am missing something here.

Anyway this fix is valid.

ldap/servers/plugins/replication/cl5_api.c
```diff
@@ -3030,9 +3029,9 @@ _cl5GenRUVInfo(dbi_val_t *key, dbi_val_t *data, void *ctx)
     ridinfo = _cl5GetRidInfo(dblcictx, rid, PR_TRUE);
     if (ridinfo->new == 1) {
         ridinfo->new = 0;
-        ridinfo->mincsn = csn;
+        ridinfo->mincsn = dblcictx->csn;
     }
-    ridinfo->maxcsn = csn;
+    ridinfo->maxcsn = dblcictx->csn;
```
tbordaz, line 3034: Sorry again same comment. The fix is valid but how did it work before ?
The csn was NULL and not modified. Does that mean the _cl5GenRUVInfo generated ridinfo with NULL min/max csn ?

tbordaz (commented):
The fixes are looking so good... that I wonder how it has been working before ;)

ldap/servers/plugins/replication/cl5_api.c
progier389: Yes, it is broken since  5a755180c192c79d65cc13e75ec6c473d7f11883 on Mar  2023 
 Issue 5661 - LMDB hangs while Rebuilding the replication changelog RUV (#5676)
So it is quite recent, 2.4,  I think and this code is only called in case of disorderly shutdown or restore or import
  (i.e: When purge RUV and upperbound Ruv are not found in changelog) 
 so it is not that frequent, and obviously The CI tests were not testing it until now ...

ldap/servers/plugins/replication/cl5_api.c
progier389: Exactltly

ldap/servers/plugins/replication/cl5_api.c
tbordaz: Ah !! Okay I thought it was a old (always existing) bug :)
Thanks for the detailed explanation

dirsrvtests/tests/suites/backups/backup_test.py
```diff
@@ -106,6 +108,65 @@ def test_db_home_dir_online_backup(topo):
...
+        # Note: wait_for_replication perform changes
+        repl.wait_for_replication(S1, S2)
+        repl.wait_for_replication(S2, S1)
+        # Step 4: Stop S1
+        S1.stop()
+        # Step 5: Destroy S1 database
+        if get_default_db_lib() == "mdb":
+            os.remove(f'{S1.ds_paths.db_dir}/data.mdb')
+        else:
+            shutil.rmtree(f'{S1.ds_paths.db_dir}/userRoot')
```
tbordaz, line 157: A backup should restore all backends. Why only removing userRoot ?
Also the removal of the database is not described in the [doc](https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/administration_guide/restoring_directory_server#doc-wrapper) . Is it necessary so the doc should describe it ?

tbordaz (commented):
Minor comment. ACK

progier389:
> Looks good! `2*PROTOCOL_BACKOFF_MAXIMUM,` part feels a bit rough but, IIUC, you said you work on a fix that will deal with the "resync after restore" and this will be reverted later. And anyway, the `2*PROTOCOL_BACKOFF_MAXIMUM,` is still more stable than nothing :)

FYI: Initially I had the same feeling when I decided to change the starting timeout but I realized that waiting 10 minutes is not that long for a timer that fires once per hour ! The keep alive entry just need to be updated regularly to limit the number of changes to look at before finding the first change to replay but precise timing is not required.

dirsrvtests/tests/suites/backups/backup_test.py
progier389: In production the removal is not needed: the restore task is supposed to delete the whole db (so the doc is OK), But in this test case, I wanted to be absolutely sure that the db get destroyed to verify that  the "restore" task restores it properly. The test use only a single backend (i.e: userRoot) and in Python it is easier to remove it than to remove everything in the db directory.

PR #6045, opened 2024-01-20
Issue 6043, 6044 - Enhance Rust and JS bundling and add SPDX licenses for both

### LINKED ISSUE #6043: Add the bundled JS libraries in the spec file
**Issue Description**
The primary objective for adding the bundled JavaScript libraries in the spec file of 389-ds-base revolves around improving security vulnerability tracking. 

Enhance the existing generation script.

### LINKED ISSUE #6044: Generate Combined JS and Cargo Dependencies SPDX License
**Issue Description**
We need to generate a combined JavaScript (npm) and Cargo dependencies license using SPDX format.
The use of standardized identifiers for licenses will add an efficient and reliable identification of the effective 389-ds-base license.

droideck:
Description: Update the generation script in 'rpm.mk' and 'bundle-rust-downstream.py' to include SPDX license information for combined JavaScript (npm) and Cargo dependencies.

Fixes: https://github.com/389ds/389-ds-base/issues/6043
Fixes: https://github.com/389ds/389-ds-base/issues/6044

Reviewed by: ?


rpm/bundle-rust-npm.py
```diff
@@ -0,0 +1,251 @@
...
+    if not os.path.isdir(args.npm_path):
+        log.error(f"Path {args.npm_path} does not exist or is not a directory")
+        sys.exit(1)
+
+    cleaned = clean_specfile(args.spec_file)
+
+    rust_output = run_cmd(["cargo", "license", "--json", "--current-dir", args.cargo_path])
+    npm_output = run_cmd(["license-checker", "--json", "--start", args.npm_path])
+
+    if rust_output is None or npm_output is None:
```
vashirov, line 251: For me it fails before it reaches this code:
```
DS_SPECFILE=$HOME/src/fedora/389-ds-base/389-ds-base.spec make -f rpm.mk bundle-rust-npm
python3 rpm/bundle-rust-npm.py src/ src/cockpit/389-console/ /home/vashirov/src/fedora/389-ds-base/389-ds-base.spec --backup-specfile
Backing up file /home/vashirov/src/fedora/389-ds-base/389-ds-base.spec to /home/vashirov/src/fedora/389-ds-base/389-ds-base.spec.20240124_221456
Remove 'Bundled cargo crates list' content from /home/vashirov/src/fedora/389-ds-base/389-ds-base.spec
  File "/home/vashirov/src/ds/rpm/bundle-rust-npm.py", line 233, in <module>
    npm_output = run_cmd(["license-checker", "--json", "--start", args.npm_path])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vashirov/src/ds/rpm/bundle-rust-npm.py", line 75, in run_cmd
    result = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.12/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'license-checker'
make: *** [rpm.mk:48: bundle-rust-npm] Error 1
```

Perhaps we should check if `cargo-license` and `license-checker` are in $PATH before running them?

rpm/bundle-rust-npm.py
```diff
@@ -0,0 +1,251 @@
...
+
+IGNORED_RUST_PACKAGES: Set[str] = {"librslapd", "librnsslapd", "slapd", "slapi_r_plugin", "entryuuid", "entryuuid_syntax", "pwdchan"}
+IGNORED_NPM_PACKAGES: Set[str] = {"389-console"}
+PACKAGE_REGEX = re.compile(r"(.*)@(.*)")
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description="""Add 'Provides: bundled(crate(foo)) = version' 'Provides: bundled(npm(bar)) = version' to a Fedora based specfile.
+Additionally, add a helper comment with a comulated License metainfo which is based on Cargo.lock and Package-lock.json files content.
+You need to have 'cargo install cargo-license' and 'npm install -g license-checker' to be able to run this script.""")
```
vashirov: I'd prefer not install NPM packages globally. I usually run these through `npx`.

rpm/bundle-rust-npm.py
```diff
@@ -0,0 +1,251 @@
...
+        log.error(f"Path {args.cargo_path} does not exist or is not a directory")
+        sys.exit(1)
+    if not os.path.isdir(args.npm_path):
+        log.error(f"Path {args.npm_path} does not exist or is not a directory")
+        sys.exit(1)
+
+    cleaned = clean_specfile(args.spec_file)
+
+    rust_output = run_cmd(["cargo", "license", "--json", "--current-dir", args.cargo_path])
+    npm_output = run_cmd(["license-checker", "--json", "--start", args.npm_path])
```
vashirov: See my comment above about installing packages globally. 
```suggestion
    npm_output = run_cmd(["npx", "license-checker", "--json", "--start", args.npm_path])
```

vashirov (changes requested):
I have `Apache2` and `Python-2.0` strings in the License field, which are not recognized by Fedora's `license-validate` tool. I think the first one should be converted to `Apache-2.0`, but the second one I couldn't find in https://gitlab.com/fedora/legal/fedora-license-data (is it PSF-2.0?). Is this the one from argparse?

droideck:
> I have `Apache2` and `Python-2.0` strings in the License field, which are not recognized by Fedora's `license-validate` tool. I think the first one should be converted to `Apache-2.0`, but the second one I couldn't find in https://gitlab.com/fedora/legal/fedora-license-data (is it PSF-2.0?). Is this the one from argparse?

For Apache2, yep, it's Apache-2.0. They updated the value here - https://github.com/dominictarr/pause-stream/commit/de6cb688ad8832c86b5c2807a0a6fcbeffab54cd - 7 years ago, but I'm not sure when/if it will be released in npm. I've added a workaround.

And for argparse, I created an issue here https://gitlab.com/fedora/legal/fedora-license-data/-/issues/470 and I contacted the maintainer... For now, I think we would need to wait for the certainty that it's PSF-2.0.

P.S. a bit of interesting history shared by Richard Fontana:

"For background, CPython is notionally released under a "stack" of licenses consisting basically of the license corresponding to SPDX `PSF-2.0` plus some legacy licenses associated with several past employers of Guido van Rossum. Such a stack of licenses was submitted years ago for approval by the Open Source Initiative, but it was apparently the *wrong* license (it's not clear it was ever used in precisely that form). Fedora discovered this problem shortly after the switch to SPDX identifiers in August 2022. So now SPDX has 'Python-2.0' (the OSI version of the CPython stack) and 'Python-2.0.1' (the *real* CPython license stack)."

vashirov:
> For Apache2, yep, it's Apache-2.0. They updated the value here - [dominictarr/pause-stream@de6cb68](https://github.com/dominictarr/pause-stream/commit/de6cb688ad8832c86b5c2807a0a6fcbeffab54cd) - 7 years ago, but I'm not sure when/if it will be released in npm. I've added a workaround.

Likely never :) The repo is archived, and the whole package was replaced with a re-export of `through` https://github.com/dominictarr/pause-stream/commit/4a6fe3dc2c11091b1efbfde912e0473719ed9cc0 

`pause-stream` is required by `event-stream`, but it's also archived. It comes in our dependency tree from `audit-ci` and they have an open ticket to replace `event-stream` with something else https://github.com/IBM/audit-ci/issues/315 

`audit-ci` is used only at the build time, we don't ship it itself or any of the dependencies that it pulls. It doesn't modify any code, just stops the build pipeline. Do we still need to publish licenses for it and all of its dependencies in our spec file?
 
> And for argparse, I created an issue here https://gitlab.com/fedora/legal/fedora-license-data/-/issues/470 and I contacted the maintainer... For now, I think we would need to wait for the certainty that it's PSF-2.0.

Thanks, I added myself to the watchers list.

droideck:
> > For Apache2, yep, it's Apache-2.0. They updated the value here - [dominictarr/pause-stream@de6cb68](https://github.com/dominictarr/pause-stream/commit/de6cb688ad8832c86b5c2807a0a6fcbeffab54cd) - 7 years ago, but I'm not sure when/if it will be released in npm. I've added a workaround.
> 
> Likely never :) The repo is archived, and the whole package was replaced with a re-export of `through` [dominictarr/pause-stream@4a6fe3d](https://github.com/dominictarr/pause-stream/commit/4a6fe3dc2c11091b1efbfde912e0473719ed9cc0)
> 
> `pause-stream` is required by `event-stream`, but it's also archived. It comes in our dependency tree from `audit-ci` and they have an open ticket to replace `event-stream` with something else [IBM/audit-ci#315](https://github.com/IBM/audit-ci/issues/315)
> 
> `audit-ci` is used only at the build time, we don't ship it itself or any of the dependencies that it pulls. It doesn't modify any code, just stops the build pipeline. Do we still need to publish licenses for it and all of its dependencies in our spec file?

Actually, yes, we don't need it at all, as we run the `audit-ci` command using `npx`. I've created a separate PR for that - https://github.com/389ds/389-ds-base/pull/6056

And I also checked how the Cockpit project does its bundles, and they actually run it only for production packages, so I've added the `--production` flag to my script.

Please review.

vashirov (approved):
LGTM!

PR #6050, opened 2024-01-24
Issue 6049 - lmdb - changelog is wrongly recreated by reindex task

### LINKED ISSUE #6049: using lmdb the changelog is wrongly recreated by reindex task.
**Issue Description**
freeipa test case fails because replicationis broken (missing csn in changelog), and error logs shows that changelog get recreated after a reindex task. 

Freeipa ipatests.test_integration.test_simple_replication.TestSimpleReplication test fails in https://github.com/freeipa-pr-ci2/freeipa/pull/3326

The replica error logs shows:

```
[23/Jan/2024:08:16:27.387781400 +0000] - DEBUG - NSMMReplicationPlugin - replay_update - agmt="cn=meTomaster.ipa.test" (master:389): Sending modify operation (dn="cn=master.ipa.test-to-replica0.ipa.test,cn=domain,cn=topology,cn=ipa,cn=etc,dc=ipa,dc=test" csn=65af75da000200030000)
...
                                                                     
[23/Jan/2024:08:16:28.261824676 +0000] - INFO - dbmdb_db2index - userRoot: Indexing attribute: ipaidpconfiglink
...
[23/Jan/2024:08:16:28.572130451 +0000] - INFO - dbmdb_public_dbmdb_import_main - reindex userRoot: Reindexing complete.  Processed 554 entries in 1 seconds. (554.00 entries/sec)

...

[23/Jan/2024:08:16:28.579981020 +0000] - DEBUG - NSMMReplicationPlugin - replica_reload_ruv - changelog contains changes that are not in the databae.
[23/Jan/2024:08:16:28.580384608 +0000] - WARN - NSMMReplicationPlugin - replica_reload_ruv - New data for replica dc=ipa,dc=test does not match the data in the changelog.
 Recreating the changelog file. This could affect replication with replica's consumers in which case the consumers should be reinitialized.

...

[23/Jan/2024:08:16:29.776478091 +0000] - ERR - agmt="cn=meTomaster.ipa.test" (master:389) - clcache_load_buffer - Can't locate CSN 65af75da000200030000 in the changelog (DB rc=-12797). If replication stops, the consumer may need to be reinitialized.
```
==> Looks like the RUV was not properly written in the db when the backend got disabled by the reindex task so the changelog get discarded when backend is reenabled after the reindex

 - Platform: Fedora

**Steps to Reproduce**
1. run freeipa tests.
2. See test result:  **389ds-fedora/simple_replication** fails
3. freeipa test_simple_replication

**Expected results**
1. freeipa **389ds-fedora/simple_replication ** should pass
4. changelog should never be recreated after a reindex.

**Additional context**

progier389:
dbmdb_import_all_done called at the end of import, bulk import and reindex is reenabling the backend
 which trigger the replication plugin to check if data were not reloaded, but in the reindex case, the backend was not disabled (so the db ruv is not up to date) and changelog is then discarded .
 The solution is to set back the backend in not busy mode when doing a reindex.

Issue:  #6049 

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
@@ -736,7 +736,11 @@ dbmdb_import_all_done(ImportJob *job, int ret)
             ldbm_set_last_usn(inst->inst_be);
 
             /* bring backend online again */
-            slapi_mtn_be_enable(inst->inst_be);
+            if (job->flags & FLAG_REINDEXING) {
+                instance_set_not_busy(inst);
+            } else {
+                slapi_mtn_be_enable(inst->inst_be);
+            }
             slapi_log_err(SLAPI_LOG_INFO, "dbmdb_import_all_done",
                           "Backend %s is now online.\n",
                           slapi_be_get_name(inst->inst_be));
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
@@ -735,7 +735,17 @@ dbmdb_import_all_done(ImportJob *job, int ret)
             /* Reset USN slapi_counter with the last key of the entryUSN index */
             ldbm_set_last_usn(inst->inst_be);
 
-            /* bring backend online again */
+            /* Bring backend online again:
+             * In lmdb case, the import framework is also used for reindexing
+             * while in bdb case reindexing uses its own code.
+             * So dbmdb_import_all_done is called either after 
+             * dbmdb_ldif2db or after dbmdb_db2index while
+             * bdb_import_all_done is only called after bdb_ldif2db.
+             *
+             * dbmdb_db2index uses instance_set_busy_and_readonly 
+             * while dbmdb_ldif2db uses slapi_mtn_be_disable
+             * and these functions have to be reverted accordingly.
+             */
             if (job->flags & FLAG_REINDEXING) {
                 instance_set_not_busy(inst);
             } else {
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -29,6 +29,7 @@ from lib389.replica import Replicas, ReplicationManager, ReplicaRole
 from lib389.agreement import Agreements
 from lib389 import pid_from_file
 from lib389.dseldif import *
+from lib389.tasks import Tasks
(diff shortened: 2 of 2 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -736,7 +736,11 @@ dbmdb_import_all_done(ImportJob *job, int ret)
             ldbm_set_last_usn(inst->inst_be);
 
             /* bring backend online again */
-            slapi_mtn_be_enable(inst->inst_be);
+            if (job->flags & FLAG_REINDEXING) {
```
tbordaz, line 749: Looking at bdb_import_all_done we have similar lines (enable backend even in case of reindex). Should it be also fixed in bdb_import_all_done ?

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -736,7 +736,11 @@ dbmdb_import_all_done(ImportJob *job, int ret)
...
+                instance_set_not_busy(inst);
```
tbordaz, line 750: Fix looks good. A bit painful to retrieve where it was set, I found bdb_db2index and dbmdb_db2index. Do you mind to put a comment from where the BUSY flag was set.

tbordaz (commented):
Fix looks good. A minor question in the review.

Also something I do not understand, do you know why this problem raise only now. It should have failed similarly with BDB.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
progier389: No: Unlike in lmdb, bdb_import_all_done is not called after bdb_db2index (And that is why there is no problem with bdb)
That is due to a difference in the way xxx_db2index are coded:
 - bdb: The function is quite large (864 lines) because it prepares the reindex - performs the reindex - sets back backend in non_busy mode. BTW that is why bdb reindex is slower than import
 - mdb:  The function is simpler (85 lines) because it simply prepare the reindex then reuse the import framework pipeline (in lmdb, the reindex processing is quite similar to the import except than data are read from entryid instead of ldif)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
progier389: Ok, I will add a comment

progier389: Done.

tbordaz (approved):
Thanks for the explanations.
LGTM

PR #6053, opened 2024-01-24
Issue 6052 - Paged results test sets hostname to `localhost` on test …

### LINKED ISSUE #6052: Paged results test sets hostname to `localhost` on test collection
**Issue Description**
Paged results test module has some code outside of the test functions and fixtures:
https://github.com/389ds/389-ds-base/blob/7a158c753a8e62dac7464dad15838aa864ea917c/dirsrvtests/tests/suites/paged_results/paged_results_test.py#L49-L54
It gets interpreted by pytest *on test collection*. These tests might be even skipped, but the code to change the hostname would still be executed. This leads to a situation, where certain test cases fail with:
```
E         ldap.SERVER_DOWN: {'result': -1, 'desc': "Can't contact LDAP server", 'ctrls': [], 'info': 'TLS: hostname does not match subjectAltName in peer certificate'}
```

Code like this should be in a fixture. But in this case I don't think we need to change the hostname at all.

vashirov:
…collection

Bug Description:
Paged results test module has some code outside of the test functions and fixtures. It gets interpreted by pytest on test collection. These tests might be even skipped, but the code to change the hostname would still be executed. This leads to a situation, where certain test cases fail with:
```
E         ldap.SERVER_DOWN: {'result': -1, 'desc': "Can't contact LDAP server", 'ctrls': [], 'info': 'TLS: hostname does not match subjectAltName in peer certificate'}
```

Fix Description:
Remove the code that changes hostname, since the test no longer does the checks based on the hostname, only on IP address.

Fixes: https://github.com/389ds/389-ds-base/issues/6052

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/paged_results/paged_results_test.py
+++ b/dirsrvtests/tests/suites/paged_results/paged_results_test.py
@@ -46,14 +46,6 @@ NEW_SUFFIX_2 = 'ou={},{}'.format(NEW_SUFFIX_2_NAME, NEW_SUFFIX_1)
 NEW_BACKEND_1 = 'parent_base'
 NEW_BACKEND_2 = 'child_base'
 
-OLD_HOSTNAME = socket.gethostname()
-if os.getuid() == 0:
-    socket.sethostname('localhost')
-HOSTNAME = socket.gethostname()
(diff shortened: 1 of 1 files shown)


tbordaz (approved):
LGTM

bsimonova (approved):

PR #6054, opened 2024-01-24
Issue 6051 - Drop unused pytest markers

### LINKED ISSUE #6051: Drop unused pytest markers
**Issue Description**
We have pytest markers such as `bz12345` or `ds1234`. After migration from Pagure to GitHub `dsNNNN` markers no longer point to the right GitHub ticket. I wanted to update them and include in `pytest.ini`, but realized that we no longer actually use these markers. In fact, I use `git log` and `git blame` far more often to identify the related issues.
I discussed this with @bsimonova and we decided to drop these markers altogether. This should also reduce the warnings noise.

vashirov:
Bug Description:
We have pytest markers such as `bz12345` or `ds1234`, but they are not registered in `pytest.ini` and generate warnings. We no longer use them to executed tests, and `git log` and `git blame` can be used for repo archeology.

Fix Description:
Delete unused pytest markers.

Fixes: https://github.com/389ds/389-ds-base/issues/6051

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/acl/enhanced_aci_modrnd_test.py
+++ b/dirsrvtests/tests/suites/acl/enhanced_aci_modrnd_test.py
@@ -64,7 +64,6 @@ def env_setup(topology_st):
     topology_st.standalone.modify_s(CONTAINER_2, mod)
 
 
-@pytest.mark.ds47553
 def test_enhanced_aci_modrnd(topology_st, env_setup):
     """Tests, that MODRDN operation is allowed,
     if user has ACI right '(all)' under superior entries,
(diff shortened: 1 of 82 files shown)


progier389 (approved):
LGTM

droideck (approved):
While we're at unused pytest markers topic. Should we, may be, clarify the lib389 docs a bit?

https://github.com/389ds/389-ds-base/blob/main/src/lib389/doc/source/guidelines.rst?plain=1#L278-L310

It mentions `pytest.mark.ssl`, even though we clearly don't use it for that (we use tier0, tier1 actively though).
But I'm not sure if we even build the docs anymore, so feel free to discard my suggestions.

Ack from me anyway.

vashirov:
This documentation doesn't contradict the current usage. What can be added there is the need to specify the new marker in `pytest.ini`. I will update the docs.

vashirov:
@droideck, I've added an additional step to lib389 docs, please review.

droideck (approved):
Looks good!

P.S. I was thinking about the point that we don't use anything else except  `tier0, tier1, etc.` so it makes sense to mention them as an example in the docs instead of `tls/ssl`.
But, I think, it's also good as it is. So it's up to you. :)
Ack from me.

PR #6056, opened 2024-01-27
Issue 3555 - Remove audit-ci from dependencies

droideck:
Description: We use npx for audit-ci runs. Hence we don't need the package installed at all.
Remove audit-ci from package.json and a new generate package-lock.json.

Related: https://github.com/389ds/389-ds-base/issues/3555

Reviewed by: ?

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -22,4 +22,4 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Run NPM Audit CI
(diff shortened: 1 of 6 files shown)


vashirov (changes requested):
On a system where `audit-ci` was not previously installed, `make -f rpms` stops with
```
npx audit-ci
Need to install the following packages:
audit-ci@6.6.1
Ok to proceed? (y)
```

We should update `Makefile.am` and `rpm.mk` to use `npx --yes audit-ci` instead.

vashirov (approved):

PR #6062, opened 2024-01-31
Issue 6068 - Add dscontainer stop function

### LINKED ISSUE #6068: Add dscontainer stop function
There currently is not a stop function in dscontainer.  It would be nice to have for use cases such as testing/debugging, plus custom container setups run during the Docker build in which dscontainer is started to do some custom configs, then later a stop function would be nice to gracefully stop dscontainer.   Discussed in  https://github.com/389ds/389-ds-base/discussions/6058.

A simple stop() function added to dscontainer that gracefully stops the ns-slapd process.

We discussed doing more elaborate things such as introducing the `psutil` lib and doing more robust signaling such as enumerating child processes.  It was determined that ns-slapd process doesn't fork children processes, and instead relies only on threads so signaling code can be simplifed.   An alternative to a PID file approach is to use process name matching, though names aren't unqiue so PID file is preferred.  Name lookup could be used as a fallback, but is likely overkill,

**Additional context**
A PR has already been submitted: https://github.com/389ds/389-ds-base/pull/6062

slominskir:
Fixes https://github.com/389ds/389-ds-base/issues/6068.  See https://github.com/389ds/389-ds-base/discussions/6058

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -34,8 +34,7 @@ import subprocess
 import argparse, argcomplete
 from argparse import RawTextHelpFormatter
 
-
-from lib389 import DirSrv
+from lib389 import DirSrv, pid_exists, pid_from_file
 from lib389.cli_base import setup_script_logger
 from lib389.instance.setup import SetupDs
 from lib389.instance.options import General2Base, Slapd2Base
@@ -57,6 +56,9 @@ from lib389.idm.directorymanager import DirectoryManager
 # We setup the logger in verbose mode to make sure debug info
 # is always available!
 log = setup_script_logger("container-init", True)
+
+# PID FILE
+PID_FILE = "/data/run/slapd-localhost.pid"
 
 
 # Handle any dead child process signals we receive. Wait for them to terminate, or
@@ -337,9 +339,7 @@ binddn = cn=Directory Manager
     ds_proc = subprocess.Popen([
         "%s/ns-slapd" % paths.sbin_dir,
         "-D", paths.config_dir,
-        # This container version doesn't actually use or need the pidfile to track
-        # the process.
-        # "-i", "/data/run/slapd-localhost.pid",
+        "-i", PID_FILE,
         "-d", loglevel,
         ], stdout=None, stderr=None, env=os.environ.copy())
 
@@ -425,6 +425,18 @@ def begin_healthcheck(ds_proc, log_exception):
     return (True, False)
 
 
+def stop():
+    stop_timeout = os.getenv("DS_STOP_TIMEOUT", 60)
+    count = int(stop_timeout)
+    pid = pid_from_file(PID_FILE)
+    os.kill(pid, signal.SIGTERM)
+    while pid_exists(pid) and count > 0:
+        time.sleep(1)
+        count -= 1
+    if pid_exists(pid):
+        os.kill(pid, signal.SIGKILL)
+
+
 if __name__ == '__main__':
     # Before all else, we are INIT so setup sigchild
     signal.signal(signal.SIGCHLD, _sigchild_handler)
@@ -462,6 +474,9 @@ container host.
     parser.add_argument('-r', '--runit',
                         help="Actually run the instance! You understand what that means ...",
                         action='store_true', default=False, dest='runit')
+    parser.add_argument('-s', '--stop',
+                        help="Stop the instance",
+                        action='store_true', default=False, dest='stop')
     parser.add_argument('-H', '--healthcheck',
                         help="Start a healthcheck inside of the container for an instance. You should understand what this means ...",
                         action='store_true', default=False, dest='healthcheck')
@@ -472,6 +487,8 @@ container host.
 
     if args.runit:
         begin_magic()
+    elif args.stop:
+        stop()
     elif args.healthcheck:
         if begin_healthcheck(None, False) == (False, True):
             sys.exit(0)


src/lib389/cli/dscontainer
```diff
@@ -425,6 +427,19 @@ def begin_healthcheck(ds_proc, log_exception):
...
 
+def stop():
+    pid=int(open(PID_FILE).read())
+    parent = psutil.Process(pid)
+    procs = parent.children(recursive=True)
+    procs.append(parent)
+    for p in procs:
+        p.terminate()
+
+    gone, alive = psutil.wait_procs(procs, timeout=3)
```
Firstyear: I think the timeout should be a bit longer than 3 seconds here.

src/lib389/cli/dscontainer
```diff
@@ -462,6 +477,9 @@ container host.
     parser.add_argument('-r', '--runit',
                         help="Actually run the instance! You understand what that means ...",
                         action='store_true', default=False, dest='runit')
+    parser.add_argument('-s', '--stop',
```
Firstyear, line 477: How do you intend to call "stop"?

Firstyear (commented):
Mostly looks good, just a few small details.

src/lib389/cli/dscontainer
```diff
@@ -24,6 +24,7 @@
 #
 
 import grp
+import psutil
```
vashirov: I'd prefer not to add additional dependencies to external packages. `os.kill` can be used to send signals.

src/lib389/cli/dscontainer
```diff
@@ -58,6 +59,9 @@ from lib389.idm.directorymanager import DirectoryManager
 # is always available!
 log = setup_script_logger("container-init", True)
 
+# PID FILE
+PID_FILE=/data/run/slapd-localhost.pid
```
vashirov: lib389 has Paths module that returns constants like this. This module is already used by `dscontainer` for getting DS version. To get PID file location you can use something like this:
```
paths = Paths(serverid='localhost')
PID_FILE = paths.pid_file
```
And then you can also use `pid_from_file` from lib389 to read the file.

src/lib389/cli/dscontainer
```diff
@@ -425,6 +427,19 @@ def begin_healthcheck(ds_proc, log_exception):
     return (True, False)
 
 
+def stop():
+    pid=int(open(PID_FILE).read())
+    parent = psutil.Process(pid)
+    procs = parent.children(recursive=True)
+    procs.append(parent)
+    for p in procs:
+        p.terminate()
```
vashirov: If you terminate the parent process, all child processes would be terminated too. Is there a reason to go after each child process separately?

src/lib389/cli/dscontainer
slominskir: Good idea.  Will fix.

src/lib389/cli/dscontainer
slominskir: I think a `stop()` function is mostly useful for running dscontainer during a container build, then shutting it down gracefully (though container debugging could be another use case).  Specifically there are some initialization things that happen only on first run (but must happen).  Plus, it's possible someone would want to extend the image and build-in all of their configuration up-front during the build.   This wouldn't be a generally applicable image at that point, but some people may want a custom image for personal use.  If so, being able to gracefully stop would be nice.  Here is where I would use it: https://github.com/JeffersonLab/dirsrv/blob/77f3ead9d19bb990ce0814b3f7960c8d8c7674eb/scripts/once.sh#L29

slominskir: Also worth mentioning it really only makes sense to have `stop()` if dscontainer is NOT running as init (PID 1).  In my case I run a bash script [docker-entrypoint.sh](https://github.com/JeffersonLab/dirsrv/blob/main/scripts/docker-entrypoint.sh) as init to give me the most flexibility and I think this is a common case for development and testing.  I think bash does reap zombies, but may not forward signals though.   I was pondering how best to handle PID 1 here: https://github.com/389ds/389-ds-base/discussions/6058#discussioncomment-8323600

src/lib389/cli/dscontainer
slominskir: I'm mainly using psutil because it returns a [Process given a PID](https://github.com/slominskir/389-ds-base/blob/65a113e9cdff8af918032a0d21beb04a78852461/src/lib389/cli/dscontainer#L432).   Is there a good alternative?

src/lib389/cli/dscontainer
slominskir: Great.  Will update.  As mentioned here https://github.com/389ds/389-ds-base/discussions/6058#discussioncomment-8326466 some developer API docs would be good (or a pointer to them if I'm overlooking them).  Combing through all Python files trying to understand the API isn't great for contributors.

vashirov: We used to have lib389.readthedocs.io, but now it seems absent. @droideck, do you know why we don't have it anymore?

src/lib389/cli/dscontainer
slominskir: I think each process is supposed to cleanly shutdown all child processes, but if they don't then they become orphans (parent goes away first) or zombies (parent signals child, but does not wait for exit status) and in both cases they are re-parented to PID 1, which is supposed to do the clean up for you so in many cases people don't notice.   This may be one of those sneaky complications with UNIX, though I might be over-thinking this.   In our case dscontainer may or may not be PID 1 and dscontainer may or may not be handling cleanup properly (hard to tell).

slominskir: Also worth mentioning that the stop code is complicated by timing and checking as it first asks the child politely to stop (SIGTERM), then waits a while, then if child didn't exit gracefully in specified time then issues SIGKILL.   I think this whole thing is rather complicated if you do it gracefully.  Which is why having a built-in stop is valuable so users don't have to think about all these details or hack together a less robust kill command as I did originally.

src/lib389/cli/dscontainer
vashirov: `ns-slapd` doesn't start new processes, only new threads. So if you send SIGTERM to it, it should stop the threads. 

```
podman top --latest user, pid,  args
USER        PID         COMMAND
root        1           /usr/bin/python3 /usr/libexec/dirsrv/dscontainer -r
root        17          /usr/sbin/ns-slapd -D /etc/dirsrv/slapd-localhost -i /data/run/slapd-localhost.pid
```

I think `os.kill()` should be sufficient, is it not?

src/lib389/cli/dscontainer
vashirov: See my comment above https://github.com/389ds/389-ds-base/pull/6062#discussion_r1474810237

src/lib389/cli/dscontainer
slominskir: I guess so, if dscontainer will only ever be 1 process with no children.    I was wondering if I was over-thinking this.   Should we still attempt graceful shutdown SIGTERM, wait, followed by SIGKILL if necessary?  Or is that unnecessary?

vashirov: Existing code in `dscontainer` catches SIGTERM, SIGINT, SIGHUP:
https://github.com/389ds/389-ds-base/blob/1f95b57fb8b03f72c7d2a2ce5b14eed5ba320407/src/lib389/cli/dscontainer#L431-L434

and handles them in `kill_ds()`:
https://github.com/389ds/389-ds-base/blob/1f95b57fb8b03f72c7d2a2ce5b14eed5ba320407/src/lib389/cli/dscontainer#L346-L363 

(`terminate` is a SIGTERM)

So I don't think we need to complicate things here. Container engines do what you describe already:
https://docs.podman.io/en/latest/markdown/podman-stop.1.html
https://docs.docker.com/engine/reference/commandline/container_stop/

slominskir: OK, sounds like we can simplify.  Will update code to send SIGTERM to ns-slapd via os.kill().

slominskir:
I see the [contributing guide](https://www.port389.org/docs/389ds/contributing.html) says to rebase-squash my commits into one.  Is this up-to-date (correct)?   GitHub has supported merge-time squash-rebase for awhile now and if I do force squash-rebase you may find it hard to follow updates per our discussion above.

vashirov:
Previously this project was hosted on pagure.io, so some recommendations might not be up to date. Either way is fine, this is not a large change.

src/lib389/cli/dscontainer
slominskir: I'm not sure the `paths.pid_file` approach is going to work since it appears dscontainer actually avoids the typical slapd start and uses subprocess command instead (supplying pidfile there): https://github.com/slominskir/389-ds-base/blob/65a113e9cdff8af918032a0d21beb04a78852461/src/lib389/cli/dscontainer#L338-L341

Even if `paths.pid_file` did have the correct value, it appears unnecessary to look it up as it's assigned to subprocess command in dscontainer so we already know it because we assigned it in dscontainer.

src/lib389/cli/dscontainer
slominskir: Actually, as I dug into the code to make the simplification I realized that one major issue with removing psutil and a Process handle is there appears to be no robust way to wait on the process we're trying to kill.   It might be "good enough", but ideally we do signal then wait.  Might be alternative way to obtain process handle to wait on?   Simply doing a sleep and hoping for the best is really no better than the ps, grep, cut, kill hack I have now.

slominskir: Also, it might be better to error on the side of overly cautious, especially since dscontainer may actually be running as PID 1 in some cases and unfortunately that brings an extra burden of carefully cleaning up processes.   Specifically we had discussed [possibly adding a docker-entrypoint-initdb](https://github.com/389ds/389-ds-base/discussions/6058) functionality to dscontainer.  At that point dscontainer would be running user scripts, which could be issuing execs (spawning child processes).

slominskir: Could use https://docs.python.org/3/library/os.html#os.waitpid I suppose.  Is this good enough?  Are we concerned about init (PID 1) use-cases?   Might be separate issue I guess.

slominskir:
I'm hoping to get CI approval because I'm on Windows so building this project locally looks like quite the chore.  Would like to see how build goes: https://github.com/389ds/389-ds-base/actions/runs/7746769086

src/lib389/cli/dscontainer
Firstyear: I think we already have some waitpid code in the dscontainer file you can use?

src/lib389/cli/dscontainer
```diff
@@ -425,6 +425,12 @@ def begin_healthcheck(ds_proc, log_exception):
     return (True, False)
 
 
+def stop():
+    pid = pid_from_file(PID_FILE)
+    os.kill(pid, signal.SIGTERM)
+    os.waitpid(pid)
```
vashirov: When I run `/usr/libexec/dirsrv/dscontainer --stop`, it stops the DS, but throws an error and exits immediately instead of waiting:
```
podman exec -til /bin/bash
bash-5.2# /usr/libexec/dirsrv/dscontainer --stop
  File "/usr/libexec/dirsrv/dscontainer", line 485, in <module>
    stop()
  File "/usr/libexec/dirsrv/dscontainer", line 431, in stop
    os.waitpid(pid)
TypeError: waitpid() takes exactly 2 arguments (1 given)
```
`os.waitpid` can't be used here because it's only applicable for the child processes. It's used in another place in `dscontainer` but only because `dscontainer` actually starts that process that we wait for.

Instead we can check if PID exists by using lib389 function [`pid_exists`](https://github.com/389ds/389-ds-base/blob/1f95b57fb8b03f72c7d2a2ce5b14eed5ba320407/src/lib389/lib389/__init__.py#L181):

```suggestion
    count = 10 # Instead of hardcoding maybe we can add a similar env variable like DS_STARTUP_TIMEOUT
    while pid_exists(pid) and count > 0:
        time.sleep(1)
        count -= 1
    if pid_exists(pid):
        os.kill(pid, signal.SIGKILL)
```

vashirov (changes requested):

src/lib389/cli/dscontainer
slominskir: Sounds good.

vashirov:
@slominskir, I pushed another commit to address a missing import and use env variable `DS_STOP_TIMEOUT`. I tested in my env and it works. Please test it in your container build and check if the stop command works for you and if it's exactly what you want.
Also, please create a separate issue (for our tracking purposes) and mention it in the commit message (see git log for examples). Thanks!

vashirov (approved):

vashirov:
PR merged, thank you for your contribution, @slominskir!

Issue #6046, opened 2024-01-23
Make dscreate to work during kickstart installations


moshevds:
**Is your feature request related to a problem? Please describe.**

I want to run dscreate during a kickstart installation (F39).

**Describe the solution you'd like**

The only problem I encountered is that systemctl is not available in a chroot environment. Since lib389 already both supports systems with and without systemd, I was able to slightly alter the systemd checks so that the instance is able to start and install correctly. The two points that I changed are `DirSrv.status` and `DirSrv.start`. Unfortunately, I'm not familiar enough with lib389 to know what impact this may have elsewhere and/or what other places should be changed as well. Therefore, I'm not posting this as a PR just yet.

One additional issue I noticed is that "systemctl is-active" will return exit status 0 in the chroot environment, which lib389 interprets as if the instance is already running. (Arguably this could be seen as a systemctl bug.)

**Additional context**

This is the patch I have included in my kickstart file and it works for my purpose:
```diff
--- /usr/lib/python3.12/site-packages/lib389/__init__.py
+++ /usr/lib/python3.12/site-packages/lib389/__init__.py
@@ -1068,7 +1068,7 @@
         if self.status() is True:
             return

-        if self.with_systemd():
+        if self.with_systemd_running():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
             subprocess.check_output(["systemctl", "start", "dirsrv@%s" % self.serverid], stderr=subprocess.STDOUT)
@@ -1168,7 +1168,7 @@

         Will update the self.state parameter.
         """
-        if self.with_systemd():
+        if self.with_systemd_running():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
             rc = subprocess.call(["systemctl",
@@ -1685,6 +1685,16 @@
             return self.systemd_override
         return self.ds_paths.with_systemd
 
+    def with_systemd_running(self):
+        if not self.with_systemd():
+            return False
+        rc = subprocess.call(["systemd-detect-virt", "--chroot",
+                              "--quiet"])
+        if rc == 0:
+            rc = subprocess.call(["systemctl", "is-system-running",
+                                  "--quiet"])
+        return rc == 0
+
     def pid_file(self):
         if self._containerised:
             return "/data/run/slapd-localhost.pid"
```
What additional work is needed to merge this feature into dscreate and/or lib389? I may be able to spend a bit more time on this to create a PR if no big changes are needed. If another approach is needed, then please consider this as the feature request for that.

[...]

PR #6066, opened 2024-02-01
Issue 6046 - Make dscreate to work during kickstart installations

moshevds:
Like discussed in #6046 this makes dscreate for chroot environments. I changed a couple things from my proof of concept:
 - checking `systemctl is-system-running` is enough to make things work correctly for chroots, and I think it makes more sense this way. In general, I don't think it makes sense to depend on systemctl invocations when systemd is not operational.
 - The check is also added to the stop method. This was also not working correctly for chroots.
 - I added a comment about the exit code. Although is this case exit code 0 matches the normally expected "success" result, I got confused a bit because systemctl isn't completely predictable with its exit codes.

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -1105,7 +1105,7 @@ class DirSrv(SimpleLDAPObject, object):
         if self.status() is True:
             return
 
-        if self.with_systemd():
+        if self.with_systemd_running():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
             try:
@@ -1186,7 +1186,7 @@ class DirSrv(SimpleLDAPObject, object):
         if self.status() is False:
             return
 
-        if self.with_systemd():
+        if self.with_systemd_running():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
             subprocess.check_output(["systemctl", "stop", "dirsrv@%s" % self.serverid], stderr=subprocess.STDOUT)
@@ -1213,7 +1213,7 @@ class DirSrv(SimpleLDAPObject, object):
 
         Will update the self.state parameter.
         """
-        if self.with_systemd():
+        if self.with_systemd_running():
             self.log.debug("systemd status -> True")
             # Do systemd things here ...
             rc = subprocess.call(["systemctl",
@@ -1729,6 +1729,17 @@ class DirSrv(SimpleLDAPObject, object):
         if self.systemd_override is not None:
             return self.systemd_override
         return self.ds_paths.with_systemd
+
+    def with_systemd_running(self):
+        if not self.with_systemd():
+            return False
+        cp = subprocess.run(["systemctl", "is-system-running"],
+                            universal_newlines=True, capture_output=True)
+        # is-system-running can detect the 7 modes (initializing, starting,
+        # running, degraded, maintenance, stopping, offline) or "unknown".
+        # To keep things simple, we assume that anything other than "offline"
+        # means that systemd is usable.
+        return cp.stdout.strip() != 'offline'
 
     def pid_file(self):
         if self._containerised:


src/lib389/lib389/__init__.py
```diff
@@ -1730,6 +1730,16 @@ def with_systemd(self):
...
 
+    def with_systemd_running(self):
+        if not self.with_systemd():
+            return False
+        rc = subprocess.call(["systemctl", "is-system-running", "--quiet"])
+        # is-system-running can detect the 7 modes (initializing, starting,
+        # running, degraded, maintenance, stopping, offline) or "unknown".
+        # Exit code 0 is given for the "running" mode, which is the only one
+        # the systemd documentation describes as fully operational.
+        return rc == 0
```
vashirov: I have doubts about this check.
On my system I have one unit failed, and `systemctl is-system-running` reports `degraded`. 
So when I start the server using `dsctl localhost start`, it thinks that systemd is not running and starts the server outside of systemd:

```
sudo dsctl -v localhost start
DEBUG: The 389 Directory Server Administration Tool
DEBUG: Inspired by works of: ITS, The University of Adelaide
DEBUG: Called with: Namespace(verbose=True, instance='localhost', json=False, list=False, remove_all=False, func=<function instance_start at 0xffffa127f7e0>)
DEBUG: Allocate local instance <class 'lib389.DirSrv'> with None
DEBUG: Instance allocated
DEBUG: systemd status -> False
DEBUG: pid file /run/dirsrv/slapd-localhost.pid -> 74271
DEBUG: Pid of 74271 is not running for localhost
DEBUG: systemd status -> False
DEBUG: DEBUG: starting with ['/usr/sbin/ns-slapd', '-D', '/etc/dirsrv/slapd-localhost', '-i', '/run/dirsrv/slapd-localhost.pid']
INFO: Instance "localhost" has been started
```

```
systemctl status dirsrv@localhost | grep -i active
     Active: inactive (dead)
```

vashirov (changes requested):

src/lib389/lib389/__init__.py
moshevds: The one other option that I considered was to check for the "offline" mode specifically. Or re-introduce the check for chroot environments with `systemd-detect-virt`.

Also, `is-system-running` has a `--wait` option that I think will prevent it from returning initializing, starting or stopping. That leaves the maintenance for which I'm not sure `systemctl start` will work or not.

Do you have any opinion about those choices? The rationale for checking for `offline` specifically is that we know for certain that `systemctl start` wont work there, and all other modes will keep the same behavior that they currently have.

vashirov: Yes, I think checking for `offline` mode specifically (as an indicator that systemd is not running as PID 1) is a better option. I checked kickstart chroot and container environments, where systemd is present but not running, they report `offline`.

src/lib389/lib389/__init__.py
```diff
@@ -1730,6 +1730,17 @@ def with_systemd(self):
             return self.systemd_override
         return self.ds_paths.with_systemd
 
+    def with_systemd_running(self):
+        if not self.with_systemd():
+            return False
+        cp = subprocess.run(["systemctl", "is-system-running"],
+                            capture_output=True)
```
vashirov:
```suggestion
                            universal_newlines=True,capture_output=True)
```

src/lib389/lib389/__init__.py
```diff
@@ -1730,6 +1730,17 @@ def with_systemd(self):
...
+        # is-system-running can detect the 7 modes (initializing, starting,
+        # running, degraded, maintenance, stopping, offline) or "unknown".
+        # To keep things simple, we assume that anything other than "offline"
+        # means that systemd is usable.
+        return cp.stdout != b'offline\n'
```
vashirov:
```suggestion
        return cp.stdout != 'offline'
```

moshevds: Thanks for the suggestion. I used `text=True` because the docs say that `universal_newlines` is deprecated.
The newline is still there even with `text=True`, but is that a problem? Systemd doesn't work on any non-unix system, so the newline will always be `\n`. I think `text=True` may also convert the newline if used on OSes where newline is not `\n`, but it doesn't impact us at all. Right?

vashirov: `text` was added in Python 3.7, but we still have to support RHEL 8 with Python 3.6. So even though `universal_newlines` is marked as deprecated, it's still available and not removed yet.
We can use `cp.stdout.strip()` to compare without new lines, this is more about code readability than platform support.

moshevds: That makes sense. I updated the PR again.

vashirov (approved):

vashirov:
@moshevds, thank you for your contribution!

PR #6070, opened 2024-02-05
Issue 6069 - Avoid race condition

### LINKED ISSUE #6069: Avoid race condition in dscontainer setting Directory Manager password
The dscontainer currently signals that the container is healthy (via the `dscontainer -H` healthcheck) before the Directory Manager password is actually set.   For users/consumers of dscontainer the healthcheck is generally the signal that the container is ready and that configuration can commence (including configuration using the Directory Manager account).   The workaround at the moment is for users/consumers to issue a sleep command and hope to order the setup commands after the Directory Manager account password is set to the provided environment variable value, else resort to local-only ldapi setup commands in a root run container.

One solution would be to update the healthcheck such that it monitors an additional condition: that it is only "healthy" once the existing check says so and a marker file has been set indicating that core initialization is complete (including Directory Manager password handling).

One alternative is to only set the password the first time the server is initialized.   This makes the container much less useful as it's often desirable to initialize the container and run setup ahead of time during a container build phase, yet this would "hard-code" the password.   The option to dynamically set it at runtime via environment variable is desirable.   

There is some otherwise non-specific initialization that can be run during the docker build except for the Directory Manager password.  It has [been suggested](https://github.com/389ds/389-ds-base/discussions/6058#discussioncomment-8368775) that it is possible to set the Directory Manager password in such a way as to make it unusable for clients to bind with, yet satisfies the initialization requirement.  This could be ideal.   Otherwise, setting a fixed one hard-coded into the pre-built image could be justified if either strong warnings are given OR a check such that a dynamic password to override is no longer optional, but required else the server won't start.

**Additional context**
Discussed in https://github.com/389ds/389-ds-base/discussions/6058

slominskir:
Bug Description:
The dscontainer currently signals that the container is healthy (via the dscontainer -H healthcheck) before the Directory Manager password is actually set.

Fix Description:
The startup code now sets a marker file once the runtime supplied password is set and the command line invoked healthcheck returns healthy only if the existing checks pass and the marker file exists

Fixes https://github.com/389ds/389-ds-base/issues/6069


droideck:
Thanks for the work on this, @slominskir.
The underlying issue in #6069 still looks valid, but this PR has been inactive for a while and the implementation may need a fresh look, particularly around the readiness check.
@vashirov, would you be okay with closing this PR while keeping #6069 open for future work?

PR #6078, opened 2024-02-06
Issue 6077 - Provide a Development Container

### LINKED ISSUE #6077: Provide a Development Container
It would be nice if we had a quick, easy, repeatable, and portable way to build and run this project - specifically in a container.  This would make it much easier for contributors as they may be on various platforms such as Mac, Windows, or Linux and with varying versions of tools/compilers installed.  See discussion https://github.com/389ds/389-ds-base/discussions/6072.

A container specifically intended for getting contributors up and running quickly from anywhere so they can test their PRs.

Simply relying on the existing GitHub Actions is a quick and dirty method.  Not quite as nice as running code locally with quicker iteration and more control such as running anything directly.  There are also more official looking ways like the ["Development Containers"](https://containers.dev/) spec, but that doesn't look very mature at the moment, plus having a container sans spec metadata to start is a prerequisite if it is later decided to pursue the spec.

**Additional context**
PR incoming.

slominskir:
Bug Description:
It would be nice if we had a quick, easy, repeatable, and portable way to build and run this project - specifically in a container.

Fix Description:
A container specifically intended for getting contributors up and running quickly from anywhere so they can test their PRs.

Fixes: #6077

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-  "dockerComposeFile": "compose.yaml",
-  "service": "dirsrv",
-  "workspaceFolder": "/389-ds-base"
+  "image": "quay.io/389ds/devcontainer",
+  "runArgs": ["--env-file",".devcontainer/overrides.env"],
+  "capAdd": ["SYS_PTRACE"]
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "dockerComposeFile": "compose.yaml",
+  "service": "dirsrv",
+  "workspaceFolder": "/389-ds-base"
+}
--- a/.devcontainer/.devcontainer-compose.json
+++ b/.devcontainer/.devcontainer-compose.json
@@ -1,0 +1,5 @@
+{
+  "dockerComposeFile": "compose.yaml",
+  "service": "dirsrv",
+  "workspaceFolder": "/389-ds-base"
+}
--- a/.devcontainer/.devcontainer-image.json
+++ b/.devcontainer/.devcontainer-image.json
@@ -1,5 +1,0 @@
-{
-  "image": "quay.io/389ds/devcontainer",
-  "runArgs": ["--env-file",".devcontainer/overrides.env"],
-  "capAdd": ["SYS_PTRACE"]
-}
--- a/.devcontainer/.devcontainer-image.json
+++ b/.devcontainer/.devcontainer-image.json
@@ -1,0 +1,5 @@
+{
+  "image": "quay.io/389ds/devcontainer",
+  "runArgs": ["--env-file",".devcontainer/overrides.env"],
+  "capAdd": ["SYS_PTRACE"]
+}
--- a/.devcontainer/Containerfile
+++ b/.devcontainer/Containerfile
@@ -1,0 +1,17 @@
+FROM quay.io/389ds/devcontainer
+
+COPY ./container-entrypoint.sh /container-entrypoint.sh
+
+RUN dnf install -y tini
+
+# DEFAULTS
+ENV CC=gcc \
+    CXX=g++ \
+    CFLAGS="-O2 -g" \
+    CXXFLAGS="" \
+    LDFLAGS="" \
+    DS_DM_PASSWORD=password \
+    DS_SUFFIX_NAME=dc=example,dc=com
+
+ENTRYPOINT ["/usr/bin/tini", "--", "/container-entrypoint.sh"]
+
--- a/.devcontainer/DEV_CONTAINER.md
+++ b/.devcontainer/DEV_CONTAINER.md
@@ -1,0 +1,43 @@
+# 389 Directory Server Development Container
+
+## What is this for?
+The Development Container simplifies building and running the 389 Directory Server from anywhere including Windows, Mac, and Linux by leveraging a container. The environment is therefore already setup with the correct tools and tool versions to build and run.
+
+## How does it work
+You need to have a local container runtime such as [Docker Desktop](https://www.docker.com/products/docker-desktop/), [Podman](https://podman.io/), or [Kubernetes](https://kubernetes.io/). [Docker Compose](https://docs.docker.com/compose/) is recommended as well to simplify commands, and avoid long repetitive command lines, but not strictly required. If you don't have Docker Compose you will need to examine the `dev.yaml` and translate to container run commands. Container run commands with Kubernetes also [requires translation](https://kubernetes.io/docs/reference/kubectl/docker-cli-to-kubectl/).
+
+By default, the project directory is bind mounted into the container such that changes can be made on the local workstation using your favorite IDE.  Other modes of operation are possible such as editing in-container or using remote development tooling such as [VS Code](https://code.visualstudio.com/docs/remote/remote-overview) or [IntelliJ](https://www.jetbrains.com/remote-development/).
+
+## Get and start development container:
+```
+git clone https://github.com/389ds/389-ds-base
+cd 389-ds-base/.devcontainer
+docker compose up
+```
+
+## Build and Run inside the container:
+```
+docker exec -it bash dirsrv
+cd 389-ds-base
+./build.sh
+/usr/libexec/dirsrv/dscontainer -r &
+# Or run dsctl, etc.
+```
+
+## Configure:
+You can configure the development container using environment variables.  Create a [.devcontainer/overrides.env](https://docs.docker.com/compose/environment-variables/set-environment-variables/) file in the project root.
+
+Container Runtime Environment Variables:
+
+| Name               | Description                                                                                                                                                       | Default  | 
+|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|
+| CC                 | C compiler                                                                                                                                                        | gcc      |
+| CXX                | C++ compiler                                                                                                                                                      | g++      |
+| CFLAGS             | C compiler flags                                                                                                                                                  | "-O2 -g" |
+| CXXFLAGS           | C++ compiler flags                                                                                                                                                |          |
+| LDFLAGS            | Linker flags                                                                                                                                                      |          |
+| DS_BACKEND_NAME    | Database backend                                                                                                                                                  | example  |
+| DS_DM_PASSWORD     | Directory Manager password                                                                                                                                        | password |
+| DS_START_TIMEOUT   | Number of seconds to start dscontainer before timeout                                                                                                             | 60       |
+| DS_STOP_TIMEOUT    | Number of seconds to stop dscontainer before timeout                                                                                                              | 60       |
+| CUSTOM_CRT_URL     | Optional URL to site-specific (self-signed) corporate intercepting TLS proxy CA file; sometimes necessary to allow build to download artifacts from the Internet. |          |
--- a/.devcontainer/compose.yaml
+++ b/.devcontainer/compose.yaml
@@ -1,0 +1,17 @@
+services:
+  dirsrv:
+    build:
+      context: .
+      dockerfile: Containerfile
+    hostname: dirsrv
+    container_name: dirsrv
+    ports:
+      - "3389:3389"
+      - "3636:3636"
+    cap_add:
+      - SYS_PTRACE
+    env_file:
+      - path: ./overrides.env
+        required: false
+    volumes:
+      - ../:/389-ds-base
--- a/.devcontainer/container-entrypoint.sh
+++ b/.devcontainer/container-entrypoint.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+##
+# Check if a site-specific (self-signed) CA cert is set in the environment variable CUSTOM_CRT_URL,
+# and if it is then update trust anchors.  This is necessary for users behind a corporate TLS
+# intercepting proxy as otherwise the build will fail to download artifacts from the Internet.
+#
+# Note: check only runs if hasn't run yet.  However, if container isn't running as root then the
+# location of marker files needs to be updated.
+##
+if [ ! -f /cert-check-complete ]; then
+  if [ -z "$CUSTOM_CRT_URL" ] ; then echo "No custom cert specified in CUSTOM_CRT_URL env; skipping"; else \
+    curl -o /etc/pki/ca-trust/source/anchors/customcert.crt $CUSTOM_CRT_URL
+    update-ca-trust
+  fi
+  touch /cert-check-complete
+else
+  echo "Cert check already complete; skipping"
+fi
+
+echo "Container ready!"
+
+##
+# Note: PID 1 should actually be tini or similar.  If users decide to override the entrypoint
+# it is recommended to continue to use tini or similar.
+##
+sleep infinity
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -240,3 +240,4 @@ vendor.tar.gz
 /rust-nsslapd-private.h
 /rust-slapi-private.h
 /src/lib389/setup.py
+/.devcontainer/overrides.env
--- a/README.md
+++ b/README.md
@@ -33,7 +33,9 @@ Building
 
 Note: **--enable-asan** is optional, and it should only be used for debugging/development purposes.
 
-See also full [building guide](https://www.port389.org/docs/389ds/development/building.html).
+See also:
+- Full [Building Guide](https://www.port389.org/docs/389ds/development/building.html)
+- Portable [Development Container guide](https://github.com/389ds/389-ds-base/blob/main/.devcontainer/DEV_CONTAINER.md)
 
 Testing
 -------
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -e
+
+autoreconf -fiv
+./configure --enable-debug --with-openldap --enable-cmocka --enable-asan
+make
+make lib389
+make install
+make lib389-install


Containerfile-dev
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/389ds/ci-images:fedora
```
vashirov: This image is x86_64 only. On Apple silicon Macs it would be better to use aarch64 images. The [Containerfile](https://github.com/389ds/ci-images/blob/main/Containerfile.fedora) for this image is rather small, so I think it would be better to incorporate it in its entirety here. Fedora image is multiarch.

Containerfile-dev
```diff
@@ -0,0 +1,9 @@
...
+ARG CUSTOM_CRT_URL
+
+RUN if [ -z "$CUSTOM_CRT_URL" ] ; then echo "No custom cert needed"; else \
```
vashirov: This makes sense for the deployment, but how is this necessary for the development?

vashirov (changes requested):
Hi @slominskir,

I thought that the devcontainer definitions should live under `.devcontainer`, at least according to its specification: https://containers.dev/implementors/spec/

See also: 
https://code.visualstudio.com/docs/devcontainers/create-dev-container
and
https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers

This would allow a seamless integration of this container in VSCode and github.dev.

Could you please update your PR according to the spec?
Thanks.

slominskir:
> Could you please update your PR according to the spec?

Will do.  I found the spec to be mostly icing as 90% of the value of a development container is just use a plain container.  The spec adds metadata that would allow running directly in a web browser on GitHub.com though so I guess that's something.   IntelliJ and VSCode will also integrate with it more readily.  It may be the case I'm using Docker Compose at the moment to handle some features handled by spec compliant IDEs.

Containerfile-dev
slominskir: Will do.   Also, I was wondering if the container build should be included in the default run instructions.  For the most flexibility I suppose it should.  For faster start-up it may be best to make it optional and default to using a pre-built image on DockerHub or quay.io (389ds account?).   Also to be considered is how to run the container build.  I generally have used GitHub action on [publish new release](https://github.com/JeffersonLab/dirsrv/blob/main/.github/workflows/docker-publish.yml) trigger.   I have generally made the default compose behavior be to use an image and provide an optional build (via compose file precedence) which then requires [merging](https://docs.docker.com/compose/multiple-compose-files/merge/) two of three files: compose.yaml, compose.override.yaml, and build.yaml ([example](https://github.com/JeffersonLab/dirsrv/blob/main/docker-compose.override.yml), long winded [explanation](https://gist.github.com/slominskir/a7da801e8259f5974c978f9c3091d52c#how-does-this-work)).  Not clear how you want this handled.

Containerfile-dev
slominskir: The build can fail when run behind a corporate intercepting TLS proxy because the build fetches files from the Internet.  This is a common issue for me at least.  I often create Dockerfiles at home on my Windows PC (or travel Macbook) that are fine, but then don't work as soon I try to run them onsite at work on RHEL.  There may be a cleaner way to handle this, but for the most flexibility allowing  a user to pass in via environment variable the path to their site's local PKI cert is nice.   So for me, the build fails at work unless I have a .env file containing: 

```
CUSTOM_CRT_URL=https://pki.jlab.org/JLabCA.crt
```

I suppose we could just add a bash script that users could execute at run-time to set trusted certs (or otherwise just provide tips in the README on how to update the trust anchors).  Trying to make the image easy to use for all.  Let me know how you'd prefer to proceed.

slominskir: To clarify, the dirsrv build definately will fail behind our corporate firewall.  I think the Docker build is fine though at least at our site, not sure if someone else's site proxy intercepts dnf file fetches in a breaking way.  I have had other docker builds fail behind the proxy due to yum/dnf/apt get, whatever.

vashirov: I haven't considered an intercepting TLS proxy, so thank you for the explanation. I would remove `dnf install -y openssl nss-tools libasan` though, because it's not related to this condition and `openssl` and `nss-tools` are already runtime dependencies that are pulled in by the previous image.

Containerfile-dev
vashirov: I've [added](https://github.com/389ds/ds-container/tree/main/devcontainer) `quay.io/389ds/devcontainer` image based on rawhide, it has 2 arches (amd64 and arm64) and is rebuilt nightly. This is enough to open a workspace in a devcontainer in VSCode. `devcontainer.json` covers some aspects of the compose (ports, mounts, etc), so not sure if it's really needed for this purpose.

dev.yaml
```diff
@@ -0,0 +1,21 @@
+services:
```
vashirov: We'd also need `SYS_PTRACE` capability to run debugger or ASAN builds inside the container. This can be added in `devcontainer.json` too.

vashirov (commented):
>I found the spec to be mostly icing as 90% of the value of a development container is just use a plain container. 

I'd rather not reinvent the wheel, given the devcontainer spec is widely supported https://containers.dev/supporting

> The spec adds metadata that would allow running directly in a web browser on GitHub.com though so I guess that's something. IntelliJ and VSCode will also integrate with it more readily. It may be the case I'm using Docker Compose at the moment to handle some features handled by spec compliant IDEs.

Yes, `devcontainer.json` covers aspects of Compose, so a separate YAML file for it might not be needed.

dev.yaml
slominskir: Where is the Dockerfile for https://quay.io/repository/389ds/devcontainer?  Should it be stored in an accessible place?   If I remove "dnf install libasan" the build fails with:

```
/usr/bin/ld: cannot find /usr/lib64/libasan.so.8.0.0: No such file or directory
make[1]: *** [Makefile:5801: libsvrcore.la] Error 1
make[1]: Leaving directory '/389-ds-base'
make: *** [Makefile:3946: all] Error 2
```

Seems like it isn't really optional if [programmatically running the build](https://github.com/slominskir/389-ds-base/blob/e5b1dbb055b1e1e821830d64f2bece0181160c22/build.sh#L2).

Similarly, the "dscontainer -r" command fails without "nss-tools" with:

```
  File "/usr/lib64/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.12/subprocess.py", line 1953, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/certutil'
```

Similar for "openssl"

slominskir: Ah, sorry, it's here:  https://github.com/389ds/ds-container/tree/main/devcontainer.  Prob should add libasan, openssl, and nss-tools there?

slominskir:
In response to:
> I'd rather not reinvent the wheel
...
> Yes, `devcontainer.json` covers aspects of Compose, so a separate YAML file for it might not be needed.

Actually, devcontainer.json is the one re-inventing the wheel.  Docker Compose has been around a lot longer and devcontainer spec just copies much of the same information into a different format.  This is why my initial PR ignored the spec.  It looks to me like Microsoft applying a thin wrapper around existing Docker APIs, and isn't really that well supported either (mostly Microsoft VS Code and GitHub).  For example if I try to create a DevContainer in IntelliJ I see it is labeled as a Beta feature.  When developing with containers it often involves orchestrating multiple containers at once, which means it generally is best to use Docker Compose.   Fortunately the devspec does allow just deferring to Docker Compose, which is what I recommend:

__devcontainer.json__
```
{
  "dockerComposeFile": "dev.yaml",
  "service": "dirsrv",
  "workspaceFolder": "/389-ds-base"
}
```

It's also possible to embed this thin wrapper metadata into the Docker image itself using Docker image labels: https://containers.dev/implementors/spec/#image-metadata

dev.yaml
slominskir: Also, IntelliJ won't launch the devcontainer unlesss ps is installed.  Then it will try to copy a backend app into the container.  Very buggy at the moment though.  So probably need:

`dnf install procps` in the Containerfile as well.

slominskir: Bind mounting the project files from the container into the local filesystem for the IDE to edit works reasonably well so I'm not sure I'd use the remote editing IDE feature at the moment.   Currently with IntelliJ it opens a new window and kinda works but complains that Java isn't installed inside the container:

image

slominskir: Also probably bad to do both at the same time.  Either need to remove bind mount else don't remote edit.   Not sure if that means devcontainer author must choose up front as if volume is defined in compose it gets used.

slominskir: Testing running the .devcontainer in GitHub codespaces fails at the moment.  At first glance it appears the version of Compose inside codespaces is too old (< 2.24.0) because the dev.yaml compose file appears to choke it with log output:

```
validating /var/lib/docker/codespacemount/workspace/389-ds-base/dev.yaml: services.dirsrv.env_file.0 must be a string


2024-02-15 18:21:02.402Z: Exit code 15
2024-02-15 18:21:02.473Z: {"outcome":"error","message":"Command failed: docker-compose -f /var/lib/docker/codespacemount/workspace/389-ds-base/dev.yaml -f /var/lib/docker/codespacemount/.persistedshare/docker-compose.codespaces.yml --profile * config","description":"An error occurred retrieving the Docker Compose configuration."}
2024-02-15 18:21:02.477Z: Error: Command failed: docker-compose -f /var/lib/docker/codespacemount/workspace/389-ds-base/dev.yaml -f /var/lib/docker/codespacemount/.persistedshare/docker-compose.codespaces.yml --profile * config
2024-02-15 18:21:02.481Z:     at _g (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:462:889)
2024-02-15 18:21:02.486Z:     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2024-02-15 18:21:02.489Z:     at async J$ (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:429:1484)
2024-02-15 18:21:02.491Z:     at async x$ (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:409:3165)
2024-02-15 18:21:02.493Z:     at async gAA (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:481:3833)
2024-02-15 18:21:02.496Z:     at async BC (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:481:4775)
2024-02-15 18:21:02.498Z:     at async xeA (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:614:11265)
2024-02-15 18:21:02.500Z:     at async UeA (/.codespaces/agent/bin/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:614:11006)
2024-02-15 18:21:02.502Z: devcontainer process exited with exit code 1
``` 

Could be something else, but running that compose file via Docker Compose works locally with latest Docker Desktop on Windows.

slominskir: Also worth mentioning codespaces is more of a gee wizz cool thing, but may not be of practical use anyways.  Every GitHub user has a very small amount of free codespace time, but as some have observed, once you run out and actually have to pay for it it may end up being [cheaper to buy a Macbook laptop](https://notes.alexkehayias.com/github-codespaces-is-too-slow-and-expensive-to-replace-local-development/) and run locally.  I'd rather develop locally anyways.   Again, why my initial PR ignored the spec.  Just being able to run a container locally via compose that is pre-loaded with the correct versions of tools to build and run an app is the real win.

Also worth mentioning our preloaded [devcontainer](https://github.com/389ds/ds-container/blob/8e37868e93047206eafba8590ba977476289c332/devcontainer/Containerfile) probably should use tini or similar.  [Example](https://github.com/JeffersonLab/dirsrv/blob/ed4b8d21988cfc42004f4902ddea9c3768219585/Dockerfile#L20-L23).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dockerComposeFile": "../dev.yaml",
```
vashirov: Preferred name for Compose file is `compose.yaml`: https://docs.docker.com/compose/compose-application-model/#the-compose-file
Please also move it under `.devcontainer`.

Containerfile-dev
```diff
@@ -0,0 +1,12 @@
+FROM quay.io/389ds/devcontainer
+ARG CUSTOM_CRT_URL
+
+ARG CUSTOM_CRT_URL
```
vashirov: ARG is added twice, typo?

Containerfile-dev
```diff
@@ -0,0 +1,12 @@
...
+
+RUN if [ -z "$CUSTOM_CRT_URL" ] ; then echo "No custom cert needed"; else \
+          curl -o /etc/pki/ca-trust/source/anchors/customcert.crt $CUSTOM_CRT_URL \
+          && update-ca-trust \
+          ; fi \
+          && dnf install -y openssl nss-tools libasan procps
```
vashirov: I've updated devcontainer image, it now contains these packages.

DEV_CONTAINER.md
vashirov, line 1: Please move it under `.devcontainer/README.md`.

DEV_CONTAINER.md
```diff
@@ -0,0 +1,45 @@
...
+Container Runtime Environment Variables:
+
+| Name             | Description                                           | Default           | 
+|------------------|-------------------------------------------------------|-------------------|
+| CC               | C compiler                                            | gcc               |
+| CXX              | C++ compiler                                          | g++               |
+| CFLAGS           | C compiler flags                                      | "-O2 -g"          |
+| CXXFLAGS         | C++ compiler flags                                    |                   |
+| LDFLAGS          | Linker flags                                          |                   |
+| DS_BACKEND_NAME  | Development database backend                          | example           |
```
vashirov: I don't see this env variable used anywhere in the source.

DEV_CONTAINER.md
```diff
@@ -0,0 +1,45 @@
...
+| DS_DM_PASSWORD   | Development Directory Manager password                | password          |
```
vashirov: I think we can omit "Development" in the descriptions.

DEV_CONTAINER.md
```diff
@@ -0,0 +1,45 @@
+# 389 Directory Server Development Container
+
+## What is this for?
+The Development Container simplifies building and running the 389 Directory Server from anywhere including Windows, Mac, and Linux by leveraging a container.  The environment is therefore already setup with the correct tools and tool versions to build and run.
```
vashirov: Nit: please avoid double/triple spaces after dot.

dev-defaults.env
```diff
@@ -0,0 +1,8 @@
+CC=gcc
+CXX=g++
+CFLAGS="-O2 -g"
+CXXFLAGS=""
+LDFLAGS=""
+DS_DM_PASSWORD=password
+DS_SUFFIX_NAME=dc=example,dc=com
+DS_BACKEND_NAME=example
```
vashirov: I don't see this env variable used anywhere in the source.

dev.yaml
```diff
@@ -0,0 +1,23 @@
...
+    container_name: dirsrv
+    ports:
+      - "3389:3389"
+      - "3636:3636"
+    init: true
+    entrypoint: ["sleep", "infinity"]
+    cap_add:
+      - SYS_PTRACE
+    env_file:
+      - path: ./dev-defaults.env
```
vashirov: Please move these env files under `.devcontainer`, and omit `dev-`. I think it's clear that this is a devcontainer, no need to put Dev/Development everywhere.

dev.yaml
vashirov: This week I will get my hands on Windows 11 laptop to test this. I believe primary use case is still local development, not cloud-first, although it's nice if it's supported.

Containerfile-dev
vashirov, line 1: Please move this file under `.devcontainer/Containerfile`.

vashirov (changes requested):

.devcontainer/devcontainer.json
slominskir: Done.    The `compose.yaml` name is the default (doesn't require `-f`) and when it was at root of project I wasn't sure the dev compose was the "default" as you might have a demo of the project as the default compose (that's what I do).  Moving to .devcontainer makes it a non-issue.

Containerfile-dev
slominskir: Typo.  It might make sense to just move everything to upstream container build that can be moved there (tini, default envs) and then move site-specific cert handling to a runtime entrypoint.   Proposed changes are now in place.   Since upstream Containerfile is in separate project I updated the Containerfile here, but this Containerfile can probably go away if we move all this upstream.

Containerfile-dev
slominskir: Great.  Consider setting the environment variable default values in the upstream Containerfile as well.  Plus consider setting the entrypoint to use tini with container-entrypoint.sh that handles site-specific cert handling.

DEV_CONTAINER.md
slominskir: Done.

DEV_CONTAINER.md
slominskir: The backend name is needed as soon as you want to create a database, but otherwise not strictly required.  I use it in the default setup scripts in my dirsrv container here:

https://github.com/JeffersonLab/dirsrv/blob/ed4b8d21988cfc42004f4902ddea9c3768219585/scripts/example/docker-entrypoint-initdb.d/01_schema.sh#L5

dev.yaml
slominskir: Done.

DEV_CONTAINER.md
slominskir: Fixed.  This is interesting.  The number of spaces is somewhat subjective but matching project README (singled space after period in the two sentences that follow another) makes sense (consistency is prob most important).     When I was in school we had to use two spaces (since middle school typing class).    It's changed over the years, but appears to have always been somewhat subjective and guide specific.   I think originally two spaces is a typewriter convention due to monospaced fonts.   When the README is translated to HTML all spaces are collapsed in browser per HTML rules anyways, but source code formatting is important.  It looks like somewhat recently (2019) APA style guide folded to pressure and changed from two spaces to 1 and then in 2020 Microsoft updated Word to ensure single spaces.   I have like 25+ years of typing habit/reflex to undo now I guess.   Also Markdown requires trailing 2 spaces on last sentence in a paragraph if your intention is to create a newline as an actual newline character doesn't cut it (at first I thought your comment was about this). 

- https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/
- https://www.grammarly.com/blog/spaces-after-period/

slominskir:
Many things can be moved to the upstream build or otherwise out of Compose, but environment overrides are tricky.   The Compose optional `env_file` doesn't appear to have an equivalent, but that relatively new feature also doesn't work on GitHub (presumably Compose version is too old).   A devcontainer JSON metadata that just pointed at an image and skipped Compose is almost the same except the override.env file is no longer optional, but required:  https://github.com/389ds/389-ds-base/pull/6078/files#diff-66406abc8095b6e7bc6eb42880d7826f4dfe8af41e5f319b9b617b76adaa5ff1

slominskir:
Also worth mentioning the local bind mount approach doesn't work on GitHub codespaces either.  Only the remote development option works (instruments the dev container with an agent that communicates via ports).   See: https://code.visualstudio.com/remote/advancedcontainers/add-local-file-mount.   So it seems decision to support GitHub codespaces (cloud dev) may dictate how we configure dev container.   Having an alternative .devcontainer metadata file might be an option as some of the decisions such as bind local files vs remote dev and container image vs container build could be made options by having separate metadata files.

Also worth mentioning despite "dev container" word being squatted on by Microsoft here it really doesn't represent all development containers and certainly isn't only way to make them.  I think we're still in the big companies duking it out for control phase.   There are a ton of competing specs / companies (with admittedly less on the nose sounding tech names) such as:
- https://www.gitpod.io/   (GitLab, Amazon)
- https://coder.com/  (Netflix, DropBox, Square) - does actually support .devcontainer json now, but native format is templates
-  https://devpod.sh/ - open source version of codespaces?
- https://github.com/recode-sh/cli - this one probably dying, but is looks like another open source version of codespaces basically

slominskir:
So the empty overrides.env file isn't so great because now Git tracks changes to it even though it's in .gitignore.   It may be ok to remove it and the `runArgs` from the devcontainer.json for the cloud run scenario and instead just have users override the default environment at runtime in-shell if desired (the env vars are mostly for the build commands and dscontainer command).  For testing and dev the defaults probably won't be changed much anyways.

I set the default devcontainer.json to be the cloud use-case (no local files mounted) as GitHub codespaces doesn't appear to give you an option to choose the file.    For local dev (with local file mount) you don't really need the devcontainer wrapper metadata and instead can just issue `docker compose up`.  So maybe both use-cases are supported this way.

Users can avoid spending hours trying to get the environment setup and can instead either issue `docker compose up` else click on the `Code` button in GitHub to run the dev container in the cloud and view it in-browser.  Verified the build works, though GitHub starts to sweat and warns you with:

image

Not sure how long the [free/personal](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) 120 processor core-hours will last at this rate! 

I was also able to run `/usr/libexec/dirsrv/dscontainer -r &` and then `/usr/libexec/dirsrv/dscontainer -s`, so it seems you can run the app too, though the debugging shows some warning that may be related to running in cloud container:

```
INFO: 389-ds-container started.
[20/Feb/2024:20:22:24.896411080 +0000] - ERR - SECU_Strerror - Sequence error in error strings at item -4985 - error -4992 (Paged Search Time Limit Exceeded - T3)
[20/Feb/2024:20:22:24.911722854 +0000] - ERR - SECU_Strerror - Should come after 393 - error -4992 (Plugin - P1)
```

vashirov:
> Also worth mentioning despite "dev container" word being squatted on by Microsoft here it really doesn't represent all development containers and certainly isn't only way to make them. 

It appears that you and I have different interpretations of what "devcontainer" is. Some of your proposed changes are irrelevant to our team workflows and are suitable only for your needs. In this case, I suggest that you create a custom `compose.yaml` file and use your own container images to meet your specific requirements. This will allow you to have more control over your development environment and tailor it to your specific needs.

At this time, I will stop working on this and will be prioritizing the fixing of our broken CI (as you keenly pointed out in #6102).

Thanks for understanding.

slominskir:
Hi,
  I'm not sure where our differences lie in understanding dev containers, please spell them out.  I also don't see anything specific to my needs, please elaborate.

My perspective is that a dev container is simply a container used for development and is important to make it easy for contributors.   I have pointed out that there are at least two very different workflows: 1 in the cloud and 2 locally.   I think the local workflow is more important, but the cloud one is cool. I've come up with a solution that addresses both.  For the local workflow I believe a Compose metadata wrapper/runner/orchestrator is the best case.  For the cloud version Microsoft's .devcontainer spec appears reasonable.
  
I have already created a dev container, and used it to test my PRs.  Ideally one is simply provided by the project.   Asking contributors to spend hours getting their environment right is an unnecessary hurdle nowadays.    I hope you push this across the finish line eventually.

Regards,

Ryan

droideck:
Thanks for putting this proposal together, @slominskir.
Since the intended development-container workflow still needs agreement and this draft has been inactive for some time, it may be better to revisit the feature with a fresh scope.
@vashirov, would you be okay with closing this PR for now?

Issue #6067, opened 2024-02-03
Improve CLI help section for dsidm and general parent arguments


droideck:
**Issue Description**
We have two issues that make a user's life harder when they use 389 DS CLI tools.

1. When you create an entry with dsidm, the OU where this entry is placed must exist. However, this information, and the required OUs are neither mentioned in --help, the man page, nor does the error message contains any hint.

    Suggestion:
    * dsidm should mention the required OUs both in --help and the man page.
    * It would be nice if the error message could show a meaningful error:
      > Error: 105 - 4 - 32 - No such object - [] - dc=example,dc=com
      
      It's really not self-explaining that it means that ou={people,group,...} is missing

2. When we run:
    ```
    dsctl test db2bak -h
    usage: dsctl [instance] db2bak [-h] [archive]
    ```
    There is no [-v] option before instance_name mentioned, so user will not know he can use it unless he runs "dsctl -h". 

    It should mention [-v] in help like this:
    ```
    usage: dsctl [-v] [instance] db2bak [-h] [archive]
    ```

droideck:
BZ#1 - https://bugzilla.redhat.com/show_bug.cgi?id=2164990
BZ#2 - https://bugzilla.redhat.com/show_bug.cgi?id=2019735

droideck:
For the second case, I investigated the issue from different sides, and I think we have at least three options here:

1. Just add `-v` and `-j` parameters to all of the subparsers.

droideck:
For the second case (`There is no [-v] option before instance_name mentioned`), I tried different approaches, and I think we have at least three options here:

1. Just add `-v` and `-j` parameters to all of the subparsers.
   That won't actually work (or at least, I wasn't able to find how to make it work with argparse).
   If we do so, `-v` is accepted at this position `dsctl [-v] [instance] db2bak [-h] [archive]`, but it's always set to the default `False` because `-v` from a subparser overwrites the value.

2. We can specify `-v` and `-j` at the [shared parent parser](https://docs.python.org/3/library/argparse.html#parents)
   It is how argparse assumes we approach the problem. It is simple, consistent and without any hacks.
   But it will bring a problem, that current way of doing things won't work anymore `usage: dsctl [-v] [instance] db2bak [-h] [archive]`.
   Only this will work: `usage: dsctl [instance] db2bak [-v] [-j] [-h] [archive]`. Hence, we'll need to fix all of the WebUI cmd lines and it may break some customer's scripts.

[...]

PR #6079, opened 2024-02-07
Issue 6067 - Improve dsidm CLI No Such Entry handling

droideck:
Add additional error processing to dsidm CLI tool for when basedn or OU subentries are absent.

Related: https://github.com/389ds/389-ds-base/issues/6067

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_idm/__init__.py
+++ b/src/lib389/lib389/cli_idm/__init__.py
@@ -1,15 +1,30 @@
 # --- BEGIN COPYRIGHT BLOCK ---
 # Copyright (C) 2016, William Brown <william at blackhats.net.au>
-# Copyright (C) 2023 Red Hat, Inc.
+# Copyright (C) 2024 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 
+import sys
 import ldap
 from getpass import getpass
 import json
+from lib389._mapped_object import DSLdapObject
+from lib389.cli_base import _get_dn_arg
+from lib389.idm.user import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_USER
+from lib389.idm.group import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_GROUP
+from lib389.idm.posixgroup import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_POSIXGROUP
+from lib389.idm.services import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_SERVICES
+
+# Create a dict where key is module and value is an rpm to search
+BASEDN_RDNS = {
+    'user': DEFAULT_BASEDN_RDN_USER,
+    'group': DEFAULT_BASEDN_RDN_GROUP,
+    'posixgroup': DEFAULT_BASEDN_RDN_POSIXGROUP,
+    'services': DEFAULT_BASEDN_RDN_SERVICES,
+}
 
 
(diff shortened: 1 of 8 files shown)


src/lib389/lib389/cli_idm/__init__.py
```diff
@@ -1,15 +1,30 @@
...
 from getpass import getpass
 import json
+from lib389._mapped_object import DSLdapObject
+from lib389.cli_base import _get_dn_arg
+from lib389.idm.user import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_USER
+from lib389.idm.group import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_GROUP
+from lib389.idm.posixgroup import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_POSIXGROUP
+from lib389.idm.services import DEFAULT_BASEDN_RDN as DEFAULT_BASEDN_RDN_SERVICES
+
+# Create a dict where key is module and value is an rpm to search
```
vashirov: What does this comment mean?

vashirov (approved):
It's still a bit awkward that it allows to enter some details only to fail later:
```
dsidm localhost -b dc=example,dc=com service create
Enter value for cn : test
Enter value for description : test
Error: The base DN "ou=Services,dc=example,dc=com" does not exist
```

But the new error message is an improvement :)
LGTM

src/lib389/lib389/cli_idm/__init__.py
droideck: I guess doing many builds at that time affected my brain somehow (hence rdn->rpm typo). Fixed.

droideck:
> It's still a bit awkward that it allows to enter some details only to fail later:
> 
> ```
> dsidm localhost -b dc=example,dc=com service create
> Enter value for cn : test
> Enter value for description : test
> Error: The base DN "ou=Services,dc=example,dc=com" does not exist
> ```
> 
> But the new error message is an improvement :) LGTM

Okay, it was actually another typo in the same code area (`services` instead of `service`).
Now it should correctly display `Error: The DN "ou=Services,dc=example,dc=com" does not exist. It is required for "service" subcommand. Please create it first.` error.

Firstyear:
Sorry to re-open this one, but shouldn't https://github.com/389ds/389-ds-base/blob/83668192bddcb6daa272eb418a96c10931d3386b/src/lib389/lib389/cli_idm/__init__.py#L55 be reading the basedn from dsrc as well? Else you always need it with -b.

progier389:
Hi @Firstyear .
IMHO, it would change the current behavior:
   If I understood rightly in current code the base dn is interactively asked if -b option is missing (like for other dsidm mandatory options).
   So if we use a default value from dsrc there will be no way to interactively ask for values
   unless we choose to drastically change the interface (for example by adding an interactive mode option)
I have no strong feeling about that (on the plus side: it would make dsidm more like the other dsxxx commands, 
 but on the minus side it will change the interface and may disrupt some customer scripts/habits.)
In any case, if we want to pursue this idea, I think it is unrelated to the original issue and deserves its own ticket.
What do you think ?

Firstyear:
But if the default value is set, it's assumed the admin wants to use that *unless* they override it with -b? 

We had a customer report this as a regression in behaviour, which is why I bring it up :)

droideck:
> But if the default value is set, it's assumed the admin wants to use that _unless_ they override it with -b?
> 
> We had a customer report this as a regression in behaviour, which is why I bring it up :)
> 
> edit: https://bugzilla.suse.com/show_bug.cgi?id=1231462

I agree. I think it should be `[check if -b provided]` -> if not -> `[check if basedn is in .dsrc]` -> if not -> `[ask interactively]`.
I'll prepare PR, it's an easy fix.

Issue #6080, opened 2024-02-07
ns-slapd crash in referint_get_config


tbordaz:
**Issue Description**
Crash during teardown phase of tests/suites/plugins/referint_test.py::test_invalid_referint_log test

**Package Version and Platform:**
 - all version

**Steps to Reproduce**
1. run tests/suites/plugins/referint_test.py::test_invalid_referint_log
2. check coredumpctl list

**Expected results**
It should not crash

tbordaz:
4fe22eccb..72c211b89 main

PR #6081, opened 2024-02-07
Issue 6080 - ns-slapd crash in referint_get_config

tbordaz:
Bug description:
	Referential integrity plugin spawn a thread to run
	integrity check/update in a deferred way. It uses a log
	file to pipe changes to check. The name of the file,
	stored in the config, is read periodically.
	At shutdown, referint plugin close callback notifies
	the thread to stop and free the config.
	The problem is that the thread may check the config
	while it was notify to stop.

Fix description:
	Check the notification (keeprunning) before reading
	the config

relates: #6080

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/referint/referint.c
+++ b/ldap/servers/plugins/referint/referint.c
@@ -75,6 +75,7 @@ static pthread_mutex_t keeprunning_mutex;
 static pthread_cond_t keeprunning_cv;
 
 static int keeprunning = 0;
+static uint64_t batch_thread_running = 0;
 static referint_config *config = NULL;
 static Slapi_DN *_ConfigAreaDN = NULL;
 static Slapi_DN *_pluginDN = NULL;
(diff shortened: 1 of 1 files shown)


progier389 (changes requested):
Sorry, but the fix looks wrong:
[1] Not sure you really want to lose the pending changes at shutdown
[2]  It does not ensure that the thread is really stopped when destroying the config.  
IMHO After resetting the keeprunning flag in referint_postop_close once the keeprunning_mutex is released you have to wait for the thread completion before going on with the clean up

tbordaz:
Thanks for the review. You are correct the first patch was wrong. I changed the mechanism to sync plugin stop fct with the batch thread.

progier389 (approved):
LGTM even I would have preferred the use of a wait_cond but atomic operation and sleep should also do the job !

PR #6087, opened 2024-02-08
Issue 6086 - Ambiguous warning about SELinux in dscreate for non-root user

### LINKED ISSUE #6086: Ambiguous warning about SELinux in dscreate for non-root user
**Description of problem:**
When an instance is created using dscreate under non-root user, there is a scary looking ambiguous warning:

> Selinux support will be disabled, continue? [yes]:

It's not clear to the user what exactly are the implications (system-wide disabling of SELinux?).

We should provide a better wording for this question, for example

> SELinux contexts for instance files will not be applied, continue? [yes]:

Or remove it completely, since there is no other choice and the implications are benign.

**Version-Release number of selected component (if applicable):**
389-ds-base-2.1.5-4.module+el9dsrv+16995+8a75ed16.x86_64

**How reproducible:**
always

**Steps to Reproduce:**
$ mkdir -p bin
$ dscreate ds-root $HOME/test $HOME/bin
$ hash -r
$ dscreate interactive

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=2151944

vashirov:
Bug Description:
When an instance is created using dscreate under non-root user, there is a scary looking ambiguous warning:

> Selinux support will be disabled, continue? [yes]:

It's not clear to the user what exactly are the implications (system-wide disabling of SELinux?).

We should provide a better wording.

Fix Description:
Change the wording and fix spelling of SELinux in the log messages.

Fixes: https://github.com/389ds/389-ds-base/issues/6086

### DIFF
--- a/src/lib389/lib389/config.py
+++ b/src/lib389/lib389/config.py
@@ -70,9 +70,9 @@ class Config(DSLdapObject):
         if selinux_present() and (key.lower() == 'nsslapd-secureport' or key.lower() == 'nsslapd-port'):
             # Get old port and remove label
             old_port = self.get_attr_val_utf8(key)
-            self.log.debug("Removing old port's selinux label...")
+            self.log.debug("Removing old port's SELinux label...")
             selinux_label_port(old_port, remove_label=True)
-            self.log.debug("Setting new port's selinux label...")
+            self.log.debug("Setting new port's SELinux label...")
             selinux_label_port(value)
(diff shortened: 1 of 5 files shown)


progier389 (commented):
I wonder if we should logs all these warning when PREFIX environment variable is set 
Otherwise it is fine for me ...

progier389 (approved):
LGTM

PR #6091, opened 2024-02-13
Issue 6057 - vlv search may result wrong result with lmdb

### LINKED ISSUE #6057: vlv search may result wrong result with lmdb
**Issue Description**
test_backup_and_restore_TestBackupAndRestoreWithReplica freeipa test is failing with 389ds main branch
 (see https://github.com/freeipa-pr-ci2/freeipa/pull/3346. )
screening from freeipa tests suspect that a vlv serach returns the wrong result (See https://pagure.io/freeipa/issue/9523)
( I am not surprising as the VLV handling code is quite complex with lmdb because lmdb does not support RECNO )

 - Platform: Fedora
 - Browser: N/A

**Steps to Reproduce**
See https://pagure.io/freeipa/issue/9523
  (Typically setup a ipa server and an ipa replica with CA then run ipa-healthcheck on the server)

**Additional context**
freeipa team may fix the issue as the VLV searches removal task progress, but we should fix it on 389ds too.

progier389:
Different issue related to vlv index and import/bulk import:

1.   vlv sub database was not open when the backend was started 
2.   vlv index was not cleaned by import/bulk import
3.   vlv index was not rebuilt by import/bulk import
4.   vlv index not rebuilt by explicit vlv reindex.
5.   vlv index not rebuilt by explicit vlv reindex if vlv name contains hyphen.
6. vlv index not rebuilt if basedn is not the suffix. 

 In fact all theses issues had the same cause: the backend vlv search list is empty after the server get restarted.

 Solution:
    [For 1,2 and 3] Fix the test_vlv_cache_subdb_names to ensure that vlv index are properly cleaned 
      and recreated by a bulk import
    Initialize the vlv search list if it is not yet initialized when starting an instance (just before opening 
      all the sub databases associated with the backend) rather than doing it before restarting the instance after the import.
    [For 4] Add a new member for vlv in the import context and handle it properly.
    [For 5] Convert the vlv name as a dbname and store it is a separate list - compare the dbname when checking if vlv is reindexed.
    [for 6] Rebuild the proper entry dn (in case of reindex) to be able to evaluate the vlv scope
             to rebuild the dn I used the entry_info data (stored in a temporary database) that contains the rdn/nrdn/ 
               and ancestors IDs (used to to rebuild the entryrdn index) and now also store the dn which is simply 
                propagated by adding the entry rdn to the parent entry dn.  

Issue: #6057 

Reviewed by: @tbordaz , @droideck  (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/vlv/regression_test.py
+++ b/dirsrvtests/tests/suites/vlv/regression_test.py
@@ -81,6 +81,7 @@ def check_vlv_search(conn, **kwargs):
     before_count = kwargs.get('before_count', 1)
     after_count = kwargs.get('after_count', 3)
     offset = kwargs.get('offset', 3501)
+    basedn = kwargs.get('basedn', DEFAULT_SUFFIX)
 
     vlv_control = VLVRequestControl(criticality=True,
         before_count=before_count,
@@ -92,7 +93,7 @@ def check_vlv_search(conn, **kwargs):
 
     sss_control = SSSRequestControl(criticality=True, ordering_rules=['cn'])
     result = conn.search_ext_s(
-        base='dc=example,dc=com',
+        base=basedn,
         scope=ldap.SCOPE_SUBTREE,
         filterstr='(uid=*)',
         serverctrls=[vlv_control, sss_control]
@@ -127,12 +128,12 @@ def add_users(inst, users_num):
         add_an_user(inst, users, uid)
 
 
-def create_vlv_search_and_index(inst, hyphen=''):
+def create_vlv_search_and_index(inst, hyphen='', basedn=DEFAULT_SUFFIX):
     vlv_searches = VLVSearch(inst)
     vlv_search_properties = {
         "objectclass": ["top", "vlvSearch"],
         "cn": f"vlv{hyphen}Srch",
-        "vlvbase": DEFAULT_SUFFIX,
+        "vlvbase": basedn,
         "vlvfilter": "(uid=*)",
         "vlvscope": "2",
     }
@@ -307,14 +308,30 @@ def test_vlv_recreation_reindex(topology_st):
     check_vlv_search(conn)
 
 
+def get_db_filename(vlvname, prefix='', iscache=False):
+    # Normalize the vlv db name
+    chars = set('abcdefghijklmnopqrstuvwxyz0123456789')
+    vlvname = ''.join(c for c in vlvname.lower() if c in chars)
+    bename = 'userRoot'
+    cache = ''
+    if iscache:
+        cache = '~recno-cache/'
+    if get_default_db_lib() == "mdb":
+        bename = bename.lower()
+    return f'{prefix}{bename}/{cache}vlv#{vlvname}.db'
+
+
 def verify_keys_in_subdb(vlvname, inst, count):
     """
     Verify that vlv index as the expected number of keys.
     """
-    output = run_dbscan(['-f', f'{inst.dbdir}/userroot/vlv#{vlvname}.db'])
+    dblib = get_default_db_lib()
+    dbfile = get_db_filename(vlvname, prefix=f'{inst.dbdir}/')
+    output = run_dbscan(['-D', dblib, '-f', dbfile])
     # Count all lines except the MBD environment ones.
     found = output.count('\n') - output.count('MBD environment')
     if found != count:
+        log.info(f'Running: dbscan -D {dblib} -f "{dbfile}"')
         log.info(f'dbscan output for vlv {vlvname} is: {output}')
     assert found == count
 
@@ -323,25 +340,32 @@ def verify_vlv_subdb_names(vlvname, inst, count=None):
     """
     Verify that vlv index and cache sub database names are conistent on lmdb.
     """
-    # Normalize the vlv db name
-    chars = set('abcdefghijklmnopqrstuvwxyz0123456789')
-    vlvname = ''.join(c for c in vlvname.lower() if c in chars)
-    if get_default_db_lib() == "mdb":
-        output = run_dbscan(['-D', 'mdb', '-L', inst.dbdir])
-        vlvsubdbs = []
-        nbvlvindex = 0
-        for line in output.split("\n"):
-            if not line:
-                continue
-            fname = line.split()[0]
-            if 'vlv#' in fname:
-                vlvsubdbs.append(fname)
-                if '/~recno-cache/' not in fname:
-                    nbvlvindex += 1
-        log.info(f'vlv sub databases are: {vlvsubdbs}')
-        assert f'{inst.dbdir}/userroot/vlv#{vlvname}.db' in vlvsubdbs
-        assert f'{inst.dbdir}/userroot/~recno-cache/vlv#{vlvname}.db' in vlvsubdbs
+    dblib = get_default_db_lib()
+    if dblib == "bdb":
+        # Let the time to sync bdb txn log to the db
+        prefix = ''
+        time.sleep(60)
+    else:
+        prefix = f'{inst.dbdir}/'
+    output = run_dbscan(['-D', dblib, '-L', inst.dbdir])
+    vlvsubdbs = []
+    nbvlvindex = 0
+    for line in output.split("\n"):
+        if not line:
+            continue
+        fname = line.split()[0]
+        if 'vlv#' in fname:
+            vlvsubdbs.append(fname)
+            if '/~recno-cache/' not in fname:
+                nbvlvindex += 1
+    log.info(f'vlv sub databases are: {vlvsubdbs}')
+
+    assert get_db_filename(vlvname, prefix=prefix) in vlvsubdbs
+    if dblib == "mdb":
+        assert get_db_filename(vlvname, prefix=prefix, iscache=True) in vlvsubdbs
         assert len(vlvsubdbs) == 2 * nbvlvindex
+    else:
+        assert len(vlvsubdbs) == nbvlvindex
     if count:
         verify_keys_in_subdb(vlvname, inst, count)
 
@@ -368,7 +392,7 @@ def bootstrap_replication(M1, M2):
     # Assign the entry to the replica
     replica2.replace('nsDS5ReplicaBindDN', brm.dn)
 
-
+@pytest.mark.skipif(get_default_db_lib() == "bdb", reason="Not supported over bdb")
 def test_vlv_cache_subdb_names(topology_m2):
     """
     Testing that vlv index cache sub database name is consistent.
@@ -519,8 +543,10 @@ def test_vlv_start_with_bad_dse(topology_st):
         # Step 2: Restart the server
         inst.start()
 
+@pytest.mark.skipif(get_default_db_lib() == "bdb", reason="bdb may hang because of github #6029")
 @pytest.mark.parametrize("hyphen", ( '', '-' ))
-def test_vlv_reindex(topology_st, hyphen):
+@pytest.mark.parametrize("basedn", ( DEFAULT_SUFFIX,  f'ou=People,{DEFAULT_SUFFIX}' ))
+def test_vlv_reindex(topology_st, hyphen, basedn):
     """Test VLV reindexing.
 
     :id: d5dc0d8e-cbe6-11ee-95b1-482ae39447e5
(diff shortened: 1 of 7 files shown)


progier389:
A bit tedious, but this time, the freeipa test pass ...

progier389:
But unfortunately CI tests show regressions both in LDBM and BDB :disappointed: 
On bdb it is likely an issue in the CI test as only lmdb code is impacted by this fix.

progier389:
Fixed the LMDB regression about missing entries because ancestorid index was corrupted. (I forgot to realign a part of the code that was also using the entry_info record)
About BDB,  one of the failure cause is due to an unrelated regression that impacts the main branch 
 (some weird messages about paged search timeout are logged and the presence of these error messages 
   cause failure in some tests)

progier389:
This time the CI tests look good (FYI: the import tests fail because of [Issue #6103](https://github.com/389ds/389-ds-base/issues/6103))

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -3111,7 +3155,8 @@ process_vlv_index(backentry *ep, ImportWorkerInfo *info)
         struct vlvIndex *vlv_index = ps->vlv_index;
         Slapi_PBlock *pb = slapi_pblock_new();
         slapi_pblock_set(pb, SLAPI_BACKEND, be);
-        if (vlv_index && (ctx->indexAttrs==NULL || attr_in_list(vlv_index->vlv_name, ctx->indexAttrs))) {
+        if (vlv_index && vlv_index->vlv_attrinfo &&
+            is_reindexed_attr(vlv_index->vlv_attrinfo->ai_type , ctx, ctx->indexVlvs)) {
```
tbordaz, line 3159: Any reason why relaying on vlv_index->vlv_attrinfo->ai_type rather than vlv_index->vlv_name

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -900,24 +910,46 @@ dbmdb_import_entry_info_by_param(EntryInfoParam_t *param, WorkerQueueData_t *wqe
...
+            wqelmt->entry_info[INFO_IDX_ANCESTORS+pinfo[INFO_IDX_NB_ANCESTORS]] = pinfo[INFO_IDX_ENTRY_ID];
+        }
+        memcpy(INFO_NRDN(wqelmt->entry_info), nrdn, wqelmt->entry_info[INFO_IDX_NRDN_LEN]);
+        memcpy(INFO_RDN(wqelmt->entry_info), rdn, wqelmt->entry_info[INFO_IDX_RDN_LEN]);
+        if (dnlen>0) {
+            char *dn = INFO_DN(wqelmt->entry_info);
+            memcpy(dn, rdn, rdnlen);
+            if (pinfo[INFO_IDX_DN_LEN]) {
+                dn[rdnlen] = ',';
+                memcpy(dn+rdnlen+1, INFO_DN(pinfo), pinfo[INFO_IDX_DN_LEN]);
```
tbordaz, line 947: Code quite difficult to understand, sorry in advance if my question is confusing.
Few lines above you computed 'dnlen' with +1 and I understood it was for '\0'. Here we are adding ',' between 'rdn' and '<rest_of_dn>'. I am concerned if that added  ',' was already account in the dnlen.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -769,29 +769,12 @@ dbmdb_import_all_done(ImportJob *job, int ret)
...
-        if (strcasecmp(attrname+4, attr) == 0) {
-            return 1;
-        }
-    }
-    return 0;
-}
-
 /* vlv_getindices callback that truncate vlv index (in reindex case) */
 static int
 truncate_index_dbi(struct attrinfo *ai, ImportCtx_t *ctx)
```
tbordaz, line 774: The function truncate_index_dbi looks specific to vlv. Would it make sense to rename it with 'vlv' in its name ?

tbordaz (approved):
The overall fix is difficult to review without a deep dive in vlv indexes. I only looked at cosmetic and local logical without understanding the complete view.
LGTM

droideck (commented):
I see that `dirsrvtests/tests/suites/import/import_test.py::test_entry_with_escaped_characters_fails_to_import_and_index ` fails (but it doesn't fail on `main`). Could it be related?

progier389:
FYI: About the test_entry_with_escaped_characters_fails_to_import_and_index test:  a look at the error log shows that it is  https://github.com/389ds/389-ds-base/issues/6103 That James fixed it yesterday but I have not rebased the branch with that fix.  ==> The problem should be fixed by the merge.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: Yes (BTW: I fall in that pitfall too! (I first used vlvName then got in trouble)  :wink: 
That is because in truncate_index_dbi (that deletes the vlv index that are reindexed) the vlv_index is not available but only the ai.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: Both the space for ',' and final \0 are accounted in  dnlen:
            dnlen = rdnlen + 1 + pinfo[4];  /* dn len (including final \0) */
in other words:   strlen(rdb) + 1 (for the comma) + parent dn len (including the \0)
 
  BTW there is a minor glitch: should better be INFO_IDX_DN_LEN instead of 4:
       dnlen = rdnlen + 1 + pinfo[INFO_IDX_DN_LEN];

dirsrvtests/tests/suites/vlv/regression_test.py
```diff
@@ -106,11 +108,7 @@ def check_vlv_search(conn, **kwargs):
         assert dn.lower() == expected_dn.lower()
 
 
-def add_users(inst, users_num):
-    users = UserAccounts(inst, DEFAULT_SUFFIX)
-    log.info(f'Adding {users_num} users')
-    for i in range(0, users_num):
-        uid = STARTING_UID_INDEX + i
+def add_an_user(inst, users, uid):
```
droideck, line 111: Minor nitpick - the correct grammar will be `add_a_user`. But probably, add_user will be even better for the function name here.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -900,24 +910,46 @@ dbmdb_import_entry_info_by_param(EntryInfoParam_t *param, WorkerQueueData_t *wqe
...
-         */
-        len = strlen(nrdn) + strlen(rdn) + 2 + (5 + pinfo[1]) * sizeof(ID);
+        size_t rdnlen = strlen(rdn);
+        size_t dnlen = 0;
+        if (param->flags & EIP_RDN) {
+            /* In reindex case, dn must be rebuilt to be able to scope properly the vlv index */
+            dnlen = rdnlen + 1 + pinfo[INFO_IDX_DN_LEN];  /* dn len (including final \0) */
+        }
+
+        len = rdnlen + strlen(nrdn) + 2 + dnlen + (INFO_IDX_ANCESTORS + 1 + pinfo[INFO_IDX_NB_ANCESTORS]) * sizeof(ID);
```
droideck, line 920: It's a very minor thing, but why not calculate `strlen(nrdn)` one time in the beginning the same way as it's done for `rdnlen`? :)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -93,21 +93,31 @@ char *mdb_stat_summarize(mdb_stat_info_t *sinfo, char *buf, size_t bufsize);
...
+#define INFO_IDX_ENTRY_ID       0
+#define INFO_IDX_NB_ANCESTORS   1
+#define INFO_IDX_NRDN_LEN       2
+#define INFO_IDX_RDN_LEN        3
+#define INFO_IDX_DN_LEN         4
+#define INFO_IDX_ANCESTORS      5
+#define INFO_NRDN(info)         ((char*)(&((ID*)(info))[INFO_IDX_ANCESTORS+((ID*)(info))[INFO_IDX_NB_ANCESTORS]]))
+#define INFO_RDN(info)          (INFO_NRDN(info)+((ID*)(info))[INFO_IDX_NRDN_LEN])
+#define INFO_DN(info)           (INFO_RDN(info)+((ID*)(info))[INFO_IDX_RDN_LEN])
+#define INFO_RECORD_LEN(info)   ((INFO_DN(info)-(char*)(info))+(info)[INFO_IDX_DN_LEN])
```
droideck, line 120: What is `INFO_RECORD_LEN`? I don't see it used anywhere (or mentioned in the comment above)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
droideck, line 3159: It's not related to this change, but it could be nice to fix/clarify. In this function (just a few lines below) there's error message `slapi_log_err(SLAPI_LOG_ERR, "process_regular_index", "index_addordel_values_ext_sv failed.\n");`.

I think it could be a typo, as it has the wrong function name.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -3306,11 +3351,19 @@ dbmdb_add_import_index(ImportCtx_t *ctx, const char *name, IndexInfo *ii)
             slapi_log_err(SLAPI_LOG_INFO, "dbmdb_db2index",
                           "%s: Indexing %s\n", job->inst->inst_name, mii->name);
         } else {
-            if (job->task) {
-                slapi_task_log_notice(job->task, "%s: Indexing attribute: %s", job->inst->inst_name, mii->name);
+            if (ii->ai->ai_indexmask == INDEX_VLV) {
+                if (job->task) {
+                    slapi_task_log_notice(job->task, "%s: Indexing VLV: %s", job->inst->inst_name, mii->name);
```
droideck, line 3356: Out of curiosity, where `mii` is freed? I see one `if` - 

```
if (a->offset_dbi) {
        *(MdbIndexInfo_t**)(((char*)ctx)+a->offset_dbi) = mii;
    }
```

But if we are not `a->offset_dbi` - what then?

droideck (approved):
Some minor comments and questions, but besides that - looks good! Ack

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
droideck: It looks like it's just a LEN of the data, but it could be nice to describe.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: mii is stored in an avl hashtable so I think it is freed by: avl_free(ctx->indexes, (IFP) free_ii);

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: Indeed! looks like a cut and paste issue! :wink:

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: I agree: it is cleaner - I added rdnlen because I needed it to compute the dn size but I did not need to use nrdnlen for the dn size computation so I just did not think about it ...

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
droideck: But it's put there only in case `if (a->offset_dbi) {`, right?

Anyway, I probably miss something, as I'm not sure I fully understand what really happens here with `ii`, `ctx->job` and how we build the index list in general. I guess I need dive deeper here :)

progier389: No: it is always stored in ctx->indexes map (last line of the function is: `avl_insert(&ctx->indexes, mii, cmp_mii, NULL);` )

PR #5562, opened 2022-12-12
Issue 5561 - Nightly tests are failing

vashirov:
Bug Description:
We use ubuntu-latest as our runner image for testing in containers. Recently there was a switch from 20.04 to 22.04 that caused test failires.

Fix Description:
* Pin runner image to 22.04
* Remove cgroups mount from docker cmd since ubuntu-22.04 now supports cgroupsv2

Fixes: https://github.com/389ds/389-ds-base/issues/5561

Reviewed-by: ???

PR #6095, opened 2024-02-15
Issue 6094 - Add coverity scan workflow

### LINKED ISSUE #6094: Add coverity scan workflow
**Issue Description**
Add a new workflow for SAST (Static application security testing) using Coverity.

vashirov:
Description:
Add a new workflow for SAST (Static application security testing) using Coverity.

Fixes: https://github.com/389ds/389-ds-base/issues/6094

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,24 @@
+name: Coverity Scan
+
+on:
+  schedule:
+    - cron:  '30 9 * * 1'
+
+jobs:
+  coverity:
+    runs-on: ubuntu-22.04
+
+    container:
+      image: quay.io/389ds/ci-images:fedora
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Checkout and configure
+        run: autoreconf -fvi && ./configure
+
+      - uses: vapier/coverity-scan-action@v1
+        with:
+          project: '389ds/389-ds-base'
+          command: make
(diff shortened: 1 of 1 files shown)


progier389:
Do we really want to run covscan for every push ?
IMHO we should have a way to limit the triggering rate ...

progier389:
Once a week seems good.

progier389 (approved):
LGTM

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,24 @@
+name: Coverity Scan
+
+on:
+  schedule:
+    - cron:  '30 9 * * 1'
+
+jobs:
+  coverity:
+    runs-on: ubuntu-22.04
```
droideck, line 9: Minor nitpick, - is there any particular reason to use `22.04`?

I see we use `latest` at https://github.com/389ds/389-ds-base/blob/main/.github/workflows/compile.yml and all runs fine.

And at the action's repo https://github.com/vapier/coverity-scan-action also `latest` is used.

droideck (approved):
Anyway, you have my ack! :)

.github/workflows/coverity.yml
vashirov: Yes, there is: https://github.com/389ds/389-ds-base/pull/5562
I want to avoid similar sudden failures when they will switch `ubuntu-latest` to something else again.

Issue #6096, opened 2024-02-15
Improve connection timeout error logging


jchapma:
**Issue Description**
When a paged result search is run with a time limit, if the time limit is exceed the server closes the connection with closed IO timeout (nsslapd-ioblocktimeout) - T2. This error message is incorrect as the reason the connection has been closed was because the specified time limit on a paged result search has been exceeded.

**Package Version and Platform:**
 - Package and version: 389-ds-base main branch

**Steps to Reproduce**
1. Add 1k entries
2. Run a paged result search with a time limit
```
ldapsearch -H "ldap://localhost:389" -LLL -o ldif-wrap=0 -D "cn=dm" -w password -l 5 -E pr=5 -b "dc=example,dc=com" "(objectclass=inetUser)"  dn
```
4. Check the access logs for io block timeout error
```
[14/Feb/2024:11:47:51.245578281 +0100] conn=17 op=-1 fd=65 Disconnect - IO Block Timeout (nsslapd-ioblocktimeout) - T2
```

**Expected results**
```
[15/Feb/2024:16:12:47.014439458 +0000] conn=5 op=-1 fd=125 Disconnect - Paged Search Limit Exceeded - T3
```

**Additional context**
Proposed by @tbordaz

PR #6097, opened 2024-02-15
Issue 6096 - Improve connection timeout error logging

jchapma:
Bug description: When a paged result search is run with a time limit, if the time limit is exceed the server closes the connection with closed IO timeout (nsslapd-ioblocktimeout) - T2. This error message is incorrect as the reason the connection has been closed was because the specified time limit on a paged result search has been exceeded.

Fix description: Correct error message

Relates: https://github.com/389ds/389-ds-base/issues/6096

Reviewed by:


ldap/admin/src/logconv.pl
```diff
@@ -355,6 +355,7 @@
 $connmsg{"B4"} = "Server failed to flush data (response) back to Client";
 $connmsg{"T1"} = "Idle Timeout Exceeded";
 $connmsg{"T2"} = "IO Block Timeout Exceeded or NTSSL Timeout";
+$connmsg{"T3"} = "Paged Search Limit Exceeded";
```
jchapma: I thought it was best to another another T version, for T2 there are already 2 timeout reasons.

tbordaz: Right good idea. I would replace 'Search Limit' with 'Search Time Limit'

ldap/servers/slapd/disconnect_error_strings.h
```diff
@@ -27,6 +27,7 @@ ER2(SLAPD_DISCONNECT_BER_FLUSH, "Server failed to flush response back to Client
 ER2(SLAPD_DISCONNECT_IDLE_TIMEOUT, "Idle Timeout (nsslapd-idletimeout) - T1")
 ER2(SLAPD_DISCONNECT_REVENTS, "Poll revents - R1")
 ER2(SLAPD_DISCONNECT_IO_TIMEOUT, "IO Block Timeout (nsslapd-ioblocktimeout) - T2")
+ER2(SLAPD_DISCONNECT_PAGED_SEARCH_LIMIT, "Paged Search Limit Exceeded - T3")
```
tbordaz: 'Search Time Limit' would be more precise as there is also size limit that could be confusing

tbordaz (approved):
LGTM

ldap/servers/slapd/disconnect_error_strings.h
jchapma: Thats a good point, I will update

PR #6101, opened 2024-02-19
Issue 6099 - `systemd = False` has no effect

### LINKED ISSUE #6099: dscreate: "start = False" and "systemd = False" have no effect
**Issue Description**
When automating the configuration of 389ds, it is not possible to prevent the server being started via systemd's systemctl start.

This creates a deadlock as dscreate's hidden systemd calls collide with automation systemd.

 - Platform: RHEL9
 - Package and version: 389-ds-base-2.3.6-3.el9.x86_64
 - Browser

**Steps to Reproduce**
1. Add start = False and systemd = False to slapd-test.inf
2. Run "/usr/sbin/dscreate from-file /etc/dirsrv/slapd-test.inf"
3. dscreate hangs, because it just executed "systemctl start dirsrv@test" exactly as it was told not to, inside a systemd service that runs before dirsrv@test.

**Expected results**
Systemd is not touched, 389ds is not started, exactly as the options require.

**Additional context**
This problem has come up a few times:

https://bugzilla.redhat.com/show_bug.cgi?id=1872910
https://github.com/389ds/389-ds-base/issues/5452

The "systemd" option is broken because it's conditional on a variable called self.containerised:

https://github.com/389ds/389-ds-base/blob/d379ac9b0fed2e78b7200054c9d4197b3b2122d6/src/lib389/lib389/instance/setup.py#L950

A search through the code for the "start" option shows that this option is never parsed.

vashirov:
Bug Description:
`systemd = False` doesn't override `with_systemd = 1` from `defaults.inf` when used with `dscreate`. It is only effective when setup is running in a containerized environment (via `dscontainer`). But for some special use cases it's important that DS installation runs without systemd.

Fix Description:
Remove the condition for overriding systemd flag.
`systemd` option is not exposed in the default template, it's listed there only when `--advanced` flag is used, so it should not affect regular installations.

Fixes: https://github.com/389ds/389-ds-base/issues/6099

### DIFF
--- a/src/lib389/lib389/instance/setup.py
+++ b/src/lib389/lib389/instance/setup.py
@@ -947,8 +947,8 @@ class SetupDs(object):
         # Should I move this import? I think this prevents some recursion
         from lib389 import DirSrv
         ds_instance = DirSrv(self.verbose, containerised=self.containerised)
-        if self.containerised:
-            ds_instance.systemd_override = general['systemd']
+
+        ds_instance.systemd_override = general['systemd']
 
         # By default SUSE does something extremely silly - it creates a hostname
         # that CANT be resolved by DNS. As a result this causes all installs to


vashirov:
`dscreate` tests in the basic test suite fail because of the `assert` in the `try..except` block, it prevents finalizer to run to clean up and remove the instance before the next test. Working on the tests fix...

vashirov:
This uncovered another issue: server can't be stopped, if it was created with `systemd = False`. `ns-slapd` keeps running, because lib389 has no idea that the override existed when the instance was created, and tries to stop it using systemd (because `with_systemd = 1` is in the `defaults.inf`). And we don't expose this in `.dsrc`.

mwhahaha:
@vashirov Any update on this?  I'm hitting this trying to bootstrap an instance running in a container that mounts host directories and this change seems to be what I need.

vashirov:
@mwhahaha, sorry, no updates, this fell through the cracks. I will check it next week.

mwhahaha:
No worries. Actually I looked into this further and there are some other minor issues when running in a container. The pid file is hardcoded to `/data/run/slapd-localhost.pid` and some other minor stuff. I ended up taking the existing dscontainer script and creating something based off it to handle our environment instead. From a manual configuration for container standpoint, we need the `start = false` and `systemd = false` to both work as that's what dscontainer does when trying to initialize fresh instances.

progier389 (approved):
LGTM

aadhikar:
@vashirov Hey, can you please rebase and merge it?

vashirov:
@aadhikar, no, it's not ready and not tested properly yet. I converted it to draft, do not merge it yet.

Issue #6103, opened 2024-02-20
New connection timeout error breaks errormap


jchapma:
**Issue Description**
A recent addition to the connection disconnect error messaging, conflicts with how errormap.c maps error strings 

**Package Version and Platform:**
 - Package and version: 389-ds-base main branch

**Steps to Reproduce**
1. Run import test
```
sudo pytest -s -v  ./dirsrvtests/tests/suites/import/import_test.py::test_entry_with_escaped_characters_fails_to_import_and_index
```
3. Tests fail because errormap logs an error to DS error logs, which are nothing to do with the import test

**Expected results**
No "error" in the DS error logs

**Additional context**
Introduced with https://github.com/389ds/389-ds-base/issues/6096

jchapma:
d09c11005..616daff82  389-ds-base-3.0 -> 389-ds-base-3.0
6da79f751..7e0fc2ff0  389-ds-base-2.5 -> 389-ds-base-2.5
c12dad024..332aad7a7  389-ds-base-2.4 -> 389-ds-base-2.4
05eac41a1..0226e8774  389-ds-base-2.3 -> 389-ds-base-2.3
563e42150..900bad76c  389-ds-base-2.2 -> 389-ds-base-2.2
a6c24e97f..88fcb070b  389-ds-base-2.1 -> 389-ds-base-2.1
0dbff28f4..de82df015  389-ds-base-2.0 -> 389-ds-base-2.0
af51ea43f..347f05627  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

vashirov:
c34558d50..b8dbffaf8  389-ds-base-1.4.3   -> 389-ds-base-1.4.3
   bc084078b..96b19fd69  389-ds-base-2.0 -> 389-ds-base-2.0

PR #6104, opened 2024-02-20
Issue 6103 - New connection timeout error breaks errormap

jchapma:
Bug description: A recent addition to the connection disconnect error messaging, conflicts with how errormap.c maps error codes/strings.

Fix description: errormap expects error codes/strings to be in ascending order. Moved the new error code to the bottom of the list.

Relates: https://github.com/389ds/389-ds-base/issues/6103

Reviewed by:


droideck (approved):
LGTM

progier389 (commented):
Changes looks good but IMHO you should add a comment in disconnect_error_strings.h to warn that the definition order matters

progier389 (approved):
LGTM

PR #6107, opened 2024-02-25
Issue 3527 - Support HAProxy and Instance on the same machine configuration

droideck:
Issue 3527 - Support HAProxy and Instance on the same machine configuration
    
Description: Improve how we handle HAProxy connections to work better when
the DS and HAProxy are on the same machine.
Ensure the client and header destination IPs are checked against the trusted IP list.
    
Additionally, this change will also allow configuration having
HAProxy listening on a different subnet than one used to forward the request.
  
Related: https://github.com/389ds/389-ds-base/issues/3527
   
Reviewed by: ?


progier389 (approved):
LGTM:
So instead of checking "ns-slapd connection Client IP" == "HAProxy Header Destination IP" and the address is trusted
The test is now that both
           "ns-slapd connection Client IP" is trusted
           "HAProxy Header Destination IP" is trusted
That seems good. 
BTW in fact this change will also allow configuration having
 HAProxy listening on a different subnet than one used to forward the request.

jchapma (commented):
LGTM

Issue #5842, opened 2023-07-14
Add log buffering to all the log types


mreynolds389:
**Issue Description**

Currently only the access and security logs offer log buffering.  Log buffering is needed for active logs so the server is not doing non-stop disk IO and slowing the server down.

We used to have a script, which has not worked in years, ds-logpipe.py.  This script was typically used when enabling verbose error log levels.  Instead of fixing this script (even if it's fixed it will break healthcheck while it's in use), if we add log buffering as an option for the error log it should prevent this server performance impact.  Same for audit logs.

This should be tested, and if it works well then we should add log buffering as an option for all the DS logs (keep it off by default though).

progier389:
Seems a good idea, but I wonder if we should not rather have explicitly 3 values:
   "on", "off", and "default"
        "default" means that buffering is enabled if log level is not the default one

mreynolds389:
Audit log buffering

f19b93e17..791018702  389-ds-base-3.0 -> 389-ds-base-3.0
6eb3c3b34..d37958b48  389-ds-base-2.5 -> 389-ds-base-2.5
3a6695238..a9814b9b4  389-ds-base-2.4 -> 389-ds-base-2.4
0226e8774..327bedac5  389-ds-base-2.3 -> 389-ds-base-2.3
900bad76c..9180f4aa8  389-ds-base-2.2 -> 389-ds-base-2.2
7d1bc439a..921467392  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

mreynolds389:
Working on error log buffering next...

mreynolds389:
Error log buffering:

021f78c80..4648172a1  389-ds-base-3.0 -> 389-ds-base-3.0
c9cea5621..6ff06ec61  389-ds-base-2.5 -> 389-ds-base-2.5
f825a3846..07915750e  389-ds-base-2.4 -> 389-ds-base-2.4

PR #6109, opened 2024-02-28
Issue 5842 - Add log buffering to audit log

mreynolds389:
Description:

Add log buffering to audit/auditfail logs.  Since these logs are intertwined there is only one config setting for both logs:

    nsslapd-auditlog-logbuffering: on/off

relates: https://github.com/389ds/389-ds-base/issues/5842

### DIFF
--- a/src/lib389/lib389/topologies.py
+++ b/src/lib389/lib389/topologies.py
@@ -124,6 +124,7 @@ def _create_instances(topo_dict, suffix):
             instance.use_ldap_uri()
             instance.open()
             instance.config.set('nsslapd-accesslog-logbuffering','off')
+            instance.config.set('nsslapd-auditlog-logbuffering','off')
             if role == ReplicaRole.STANDALONE:
                 ins[instance.serverid] = instance
                 instances.update(ins)
@@ -528,7 +529,7 @@ def topology_m1h1c1(request):
     topo_roles = {ReplicaRole.SUPPLIER: 1, ReplicaRole.HUB: 1, ReplicaRole.CONSUMER: 1}
     topology = create_topology(topo_roles, request=request)
 
-    # Since topology implements timeout, create_topology supports hub 
+    # Since topology implements timeout, create_topology supports hub
     # but hub and suppliers are fully meshed while historically this topology
     # did not have hub->master agreement.
     # ==> we must remove hub->master agmt that breaks some test (like promote_demote)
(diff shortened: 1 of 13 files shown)


droideck (commented):
Will it be possible to add WebUI, CLI, helathcheck, and some simple checks to ds_logs_test.py and healthcheck tests?

I guess some that can be moved to another PR, but it'll be nice to cover the feature fully on one place as it's all interconnected (to some degree).

mreynolds389:
> Will it be possible to add WebUI, CLI, helathcheck, and some simple checks to ds_logs_test.py and healthcheck tests?

Well in a way it is tested as if we don't disable audit log buffering a lot of tests will fail :-)  But I hear what you are saying.  

Ok I'll work on this PR a bit more, but I don't know what you what type of healthcheck you are referring to.  Seems unrelated as log buffering has nothing to do with the health of the server.

src/lib389/lib389/topologies.py
```diff
@@ -124,6 +124,7 @@ def _create_instances(topo_dict, suffix):
             instance.use_ldap_uri()
             instance.open()
             instance.config.set('nsslapd-accesslog-logbuffering','off')
+            instance.config.set('nsslapd-auditlog-logbuffering','off')
```
droideck, line 127: But I see that you disable it here in `_create_instances` - so it's not tested. Or do I miss something?

droideck (commented):
For the healthcheck, I just saw that we have it for other types of log buffering - like this https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/lint.py#L126-L156

And the tests for the check.

mreynolds389:
> For the healthcheck, I just saw that we have it for other types of log buffering - like this https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/lint.py#L126-L156
> 

True, and if the new default is "enabled" (I'm actually waiting for someone to complain about that) then we should check if it's off.  

Does this also mean we should have a healthcheck if verbose error logging is enabled too?

src/lib389/lib389/topologies.py
mreynolds389: What I was joking about was if it's broken when it's absent, but it works when its present, then that's a form of testing.  But I said I knew what you meant, and I will add a test case...

droideck: Oh, okay! :D Yeah, true)

mreynolds389:
@droideck - all reqauested changes made (test case, healthcheck/lint, and healthcheck test case), please review...

droideck (approved):
Looks great!

PR #6111, opened 2024-02-29
Issue 6110 - Typo in Account Policy plugin message

### LINKED ISSUE #6110: Typo in Account Policy plugin message
**Issue Description**
When saving configuration for Account Policy Plugin, the confirming message says "Config entry cn=config,cn=Account Policy Plugin,cn=plugins,cn=config was successfully seted". It should be 'set'

389-ds-base-2.0.14-7.module+el9dsrv+14845+69b2f526.x86_64
cockpit-389-ds-2.0.14-7.module+el9dsrv+14845+69b2f526.noarch

**Steps to Reproduce**
1. Go to Plugins tab -> Click Account Policy
2. Enable plugin -> restart
3. Create Config
4. Check "Always Record Login"
5. Click Add config

bsimonova:
Description:
Remove the additional letters after success message

Fixes: https://github.com/389ds/389-ds-base/issues/6110

Reviewed by: ???

### DIFF
--- a/src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
+++ b/src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
@@ -490,7 +490,7 @@ class AccountPolicy extends React.Component {
                     console.info("accountPolicyOperation", "Result", content);
                     this.props.addNotification(
                         "success",
-                        cockpit.format(_("Config entry $0 was successfully $1ed"), configDN, action)
+                        cockpit.format(_("Config entry $0 was successfully $1"), configDN, action === "add" ? "added" : "set")
                     );
                     this.props.pluginListHandler();
                     this.handleCloseModal();


src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
```diff
@@ -490,7 +490,7 @@ class AccountPolicy extends React.Component {
                     console.info("accountPolicyOperation", "Result", content);
                     this.props.addNotification(
                         "success",
-                        cockpit.format(_("Config entry $0 was successfully $1ed"), configDN, action)
+                        cockpit.format(_("Config entry $0 was successfully $1"), configDN, action)
```
progier389: The code seems generic 
  I suspect that action may be "add" and in this case it should display "added"
  IMHO you had better test explicitly for action == set

progier389 (changes requested):

src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
bsimonova: Yes, it is generic. 
In my steps working around the config entry always produced 'set' currently for all outcomes.
So I went for the generic modification as well.
But I see it might cause issue for 'add' and 'delete' option :( 
I am open to other approaches as I took this as a learning opportunity around cockpit code.

progier389: FYI: I was thinking about something like:
 if (action === "set") {
      cockpit.format(_("Config entry $0 was successfully set"), configDN, action)
 } else {
      cockpit.format(_("Config entry $0 was successfully $1ed"), configDN, action)
   }
But if the function is called with "set" in all cases, then your code is fine !

progier389 (approved):
LGTM

src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
droideck: I think you can just do something like this:

```
cockpit.format(_("Config entry $0 was successfully $1"), configDN, action === "add" ? "added" : "set")
```

or assign it to a const prior somewhere along the line:

```
  const actionPastTense = action === "add" ? "added" : "set";
  this.props.addNotification(
                        "success",
                        cockpit.format(_("Config entry $0 was successfully $1"), configDN, actionPastTense)
                        );
 ```

droideck (changes requested):

src/cockpit/389-console/src/lib/plugins/accountPolicy.jsx
bsimonova: Thank you both.
I will apply these.

droideck (approved):

PR #6113, opened 2024-03-01
Issue 6112 - RFE - add new operation note for MFA authentications

### LINKED ISSUE #6112: Add a new operation note for MFA authentications
For auditing purposes we need a new operation note to provide more information about a bind, specifically from a OTP/MFA plugin.   So the plugin would use the Slapi API to set the new "note" flag, and we would add a new message to the BIND or RESULT access & security log entry.

mreynolds389:
Add a new operation note to indicate that a MFA plugin performed the BIND.  This implies that the plugin must set the note itself as there is no other way to detect this:

    slapi_pblock_set_flag_operation_notes(pb, SLAPI_OP_NOTE_MFA_AUTH);

The purpose for this is for auditing needs

Fixes: https://github.com/389ds/389-ds-base/issues/6112

### DIFF
--- a/ldap/admin/src/logconv.pl
+++ b/ldap/admin/src/logconv.pl
@@ -2,11 +2,11 @@
 #
 # BEGIN COPYRIGHT BLOCK
 # Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
-# Copyright (C) 2022 Red Hat, Inc.
+# Copyright (C) 2010-2024 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
-# See LICENSE for details. 
+# See LICENSE for details.
 # END COPYRIGHT BLOCK
 #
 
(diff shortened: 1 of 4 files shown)


mreynolds389:
I did not write an CI tests because this feature requires a custom pre-bind plugin to set the new "notes" flag.  It is not something the DS can detect on its own.

mreynolds389:
Design Doc:

https://www.port389.org/docs/389ds/design/mfa-operation-note-design.html

abbra:
Thanks, looks good. I have added support for this note generation to https://github.com/freeipa/freeipa/pull/7200 but it needs to be built against 389-ds with this change.

droideck (approved):
The code looks good! But yeah, I haven't tested it:)

mreynolds389:
> The code looks good! But yeah, I haven't tested it:)

FYI, I sort of did test this, but only by hard coding every bind to have the flag.  And the logging was correct.  Not sure there is much more to do without putting a lot of effort into testing (like adding a new testing plugin, etc) for what I consider to be a very simple change.

mreynolds389:
> > The code looks good! But yeah, I haven't tested it:)
> 
> FYI, I sort of did test this, but only by hard coding every bind to have the flag. And the logging was correct. Not sure there is much more to do without putting a lot of effort into testing (like adding a new testing plugin, etc) for what I consider to be a very simple change.

That being said, it would be nice to have a testing plugin that works with the CI, but lets look into that in a different PR...

PR #6130, opened 2024-03-21
Issue 6105 - lmdb - Cannot create entries with long rdn

### LINKED ISSUE #6105: lmdb - Cannot create entries with long rdn
**Issue Description**
Full description in https://pagure.io/freeipa/issue/9529
Problem root cause:
   An entry may have several children, so Cid key containing the reference of childrens of entry with entrid "id" has several records `. To allow that, entryrdn.db is open with MDB_DUPSORT flag. But this means that len(key) + len(data) is limited 
   with some hardcoded limit (i.e: 511 bytes with the current version of liblmdb)
  And as the entryrdn.db data record contains both the rd and the normalized rdn the limit is reached with rdn lenght greater than about 250 bytes 

 - Platform:  Fedora
 - Package and version: 3.0.1

**Steps to Reproduce**
1. Through freeipa tests (See https://pagure.io/freeipa/issue/9529)
2. You can also reproduce by adding an entry with and rdn using a naming attribute with a value having 250 bytes (or more)

**Expected results**

**Screenshots**

**Additional context**
Jira ticket is: IDMDS-4148

progier389:
The issue is that entryrdn.db does not support data longer than 511 bytes and same issue about the internal db keys used for the import/reindex pipeline. 
Solution for entryrdn. I created a dbi that does not support duplicate keys (but it supports long data)
  and replaced the long rdn_elem  by a "redirect" rdn_elem that contains the key to the long data in @long-entryrdn.db
 As entryrdn code is tricky, I have first done some cleanup by using a context as parameter rather than several parameters
   (because the parameter list was getting quite large for some functions)  and added new functions to factorize the open/close for the context (and the cursor open/close) retry loop)
  I also improved the debug features because I need it ! 
  About the import internal db/ I create redirect key that refer to records whose data contains the real key
  
  Issue #6105 
  
  Reviewed by: @droideck

### DIFF
--- a/dirsrvtests/tests/suites/indexes/entryrdn_test.py
+++ b/dirsrvtests/tests/suites/indexes/entryrdn_test.py
@@ -13,7 +13,7 @@ import ldap
 import logging
 from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
 from lib389.backend import Backends
-from lib389.idm.user import UserAccounts
+from lib389.idm.user import UserAccounts, UserAccount
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.topologies import topology_m2 as topo_m2
 from lib389.agreement import Agreements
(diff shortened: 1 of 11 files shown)


progier389:
FYI: To the reviewer:  I split the changes in two commit:
  The first one is quite straight forward (The use of a context struct instead of several parameters like the backend, the db, the cursor, ..) 
  The second one is the real meat of the change

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -1581,14 +1590,77 @@ dbmdb_privdb_destroy(mdb_privdb_t **db)
...
+    MDB_val data = {0};
+    unsigned long nextid = 0;
+    char *pt = longkey->mv_data;
+    int rc = 0;
+
+    memset(smallkey, 0, sizeof *smallkey);
+    /* Prepare the samllkey used to store the long key with id=0 */
+    strcpy(smallkey->header, "==>");
+    for (pt += longkey->mv_size - 1; pt >= (char*)longkey->mv_data; pt--) {
+        smallkey->h += ((smallkey->h << 3) | (smallkey->h >> ((8 * sizeof(long)) - 3))) ^ (*pt & 0x1f);
```
droideck, line 1608: I wonder if it'll be possible to stumble upon a collision here. If I understood correctly, part of the logic here is that you get  lowest 5 bits from each byte. So if we'll have A LOT of long  very similar rdns in a directory, - I think we can get into a bit of trouble (even though highly unlikely)

Should we use some existing fast hashing function instead? It can be an overkill though (depending on the function)

droideck (approved):
Besides the minor concern, looks good!

dirsrvtests/tests/suites/indexes/entryrdn_test.py
```diff
@@ -141,6 +141,100 @@ def test_tombstone(topo_m2):
     checkdbscancount(s1, 'nsuniqueid', EXPECTED_NB_NSNIQUEID)
 
 
+def test_long_rdn(topo_m2):
+    """
```
droideck, line 163: The test description is empty...

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -1581,14 +1590,77 @@ dbmdb_privdb_destroy(mdb_privdb_t **db)
...
+    }
+    key.mv_data = smallkey;
+    key.mv_size = sizeof (*smallkey);
+    rc = MDB_CURSOR_GET(db->cursor, &key, &data, MDB_SET_RANGE);
+    /* Let walk all the keys cwthjat have the right hash */
```
droideck: 'cwthjat' typo

droideck (approved):

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
progier389: As with any hash, it is indeed possible to get collision (that is why there is also an id to have a discriminant key) and when reading the data there is a loop 
I do not think that the fact that keeping only the 5 lowest bits is really an issue: 
Usually most of the normalized (r)dn characters are in "0123456789abcdefghijklmnopqrstuvwxyz_ ,=+-"
and for this set of characters, the 5 lowest bits are discriminant.
The other part of the hash logic is that the sum is rotated left by 3 bits to spread the hash code value among the 64 bits of the hash code.
BTW in the import temporary db, the key is the entry dn (not the rdn) and the dn are unique.
A last point: IMHO this function spread much better the hash value than the dn_hash function that we use for the entry cache  (in which IMHO the 63-48 bits probably never changes)

progier389:
Fixed the typo and the missing test description.
Also added extra steps to the test case to also check children with parent having a long rdn and moddn

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
droideck: Sounds good! Thank you for the detailed explanation!
Yeah, the chances of collision are really slim.
Okay, we'll see how it performs (I think we better have some tests around that area as performance might really decrease with collisions and very large data sets).

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -1581,14 +1590,77 @@ dbmdb_privdb_destroy(mdb_privdb_t **db)
...
+    /* Prepare the samllkey used to store the long key with id=0 */
+    strcpy(smallkey->header, "==>");
+    for (pt += longkey->mv_size - 1; pt >= (char*)longkey->mv_data; pt--) {
+        smallkey->h += ((smallkey->h << 3) | (smallkey->h >> ((8 * sizeof(long)) - 3))) ^ (*pt & 0x1f);
+    }
+    key.mv_data = smallkey;
+    key.mv_size = sizeof (*smallkey);
+    rc = MDB_CURSOR_GET(db->cursor, &key, &data, MDB_SET_RANGE);
+    /* Let walk all the keys that have the right hash */
+    while (rc == 0 && key.mv_size == sizeof(privdb_small_key_t) &&
```
tbordaz, line 1614: Hi @progier389 , sorry for the late question. I am a bit puzzled, I assume this loop is to handle duplicate (small) keys. Correct ?
My understanding is that a database handling duplicate keys have values with limited size. So how the longkey could fit in a value of such database ?

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -1581,14 +1590,77 @@ dbmdb_privdb_destroy(mdb_privdb_t **db)
...
     if (!rc) {
-        rc = MDB_CURSOR_GET(db->cursor, key, data, MDB_SET_KEY);
+        if (key->mv_size > db->maxkeysize) {
+            privdb_small_key_t small_key = {0};
+            MDB_val key2 = {0};
+            key2.mv_data = &small_key;
+            key2.mv_size = sizeof small_key;
+            rc = dbmdb_privdb_init_small_key(db, key, 0, &small_key);
+            if (!rc) {
+                rc = MDB_CURSOR_GET(db->cursor, &key2, data, MDB_SET_KEY);
```
tbordaz, line 1659: I am surprise it uses the same db->cursor. Depending on the size of the key, is not it using a different database/cursor. Is db->cursor already set to the appropriate DB ?

tbordaz (commented):
LGTM.
minor questions

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
progier389: Note: that there are two different mechanism
  [1] For entryrdn (in this case, AFAIK there are two database)
            Entryrdn:   small-key --> long-data  
                is replaced by:
            Entryrdn:   small-key --> redirect-key:
            @longentryrdn: redirect-key: long-data
  [2] For the import temporary private db (which is the case here) 
             privdb: long-key --> data
                  is replaced by: 
             small-key1 --> data
             small-key2 --> long-key     (smallkey1 and smallkey2 are the same except for the first characters (= versus @) 
So yes there is only 1 db (and 1 single cursor) involved in this case

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
progier389: The small keys looks like  "==>HaveValue:Count" 
The goal of the loop is to get the highest Count for a given hash value (to handle the collisions)

PR #6131, opened 2024-03-25
Issue 5105 - lmdb - Cannot create entries with long rdn - fix covscan

progier389:
A minor code cleanup issue fixing: CID 1540880 CID 1540879 CID 1540878 CID 1540876 CID 1540875
  All these report have the same pattern but on different function. 
  The issue is that ctx == NULL is tested as part of the parameter validity tests (even if it is never NULL)
  then goto bail but the bail code dereference ctx to potentially free some resources.
  So I changed the code from:
     Log Entering in Function
     If (One of parameter is NULL) {
          Log Error message
          goto bail
      } 
   To:
    If (One of parameter is NULL) {
          Log Error message
          return -1;
     }
     Log Entering in Function
CID 1540877 is a real issue about a potential memory leak in case of error (shoud goto bail0 instead of bail to make sure childelem is free)

Issue: #6105      
      
Reviewed by: @droideck  Thanks!


droideck (approved):
Looks good!

PR #6137, opened 2024-03-28
Issue 6136 - failure in freeipa tests

### LINKED ISSUE #6136: failure in freeipa test
**Issue Description**
   When trying to add a CI test that mimic freeipa nightly test that perform backup/restore (using freeipa commands)
   I hit several issues (impacting both bdb and lmdb)
   [1] bdb - offline import fail:
  ` [28/Mar/2024:11:33:07.004652982 +0100] - ERR - bdb_ldif2db - ipaca: Failed to write import file, error -5950: File not found.`
    [2] mdb - vlv index are not generated - looks like it is because because one level scope is handled by 
         setting a parentid=basedn_entryid subfilter but the entryid is not known when the filter is created.
         bdb uses vlv_grok_new_import_entry in the import formean to reset the filter when a vlv basedn entry is added
         and mdb should do the same. 
     [3] mdb- crash while trying to import an entry without parent (i.e a suffix entry) that does not belong to the backend 

 - Platform: Fedora

**Steps to Reproduce**
See the CI test test_vlv_as_freeipa_backup_restore

progier389:
Several issue detected when adding a CI test that mimic one of freeipa nightly test :

bdb - offline import fail when trying to create the guardian file because instance is not yet fully initialized and the generated path is wrong - fixed by using the directory from ldbminfo and the instance names that are defined.

mdb - vlv index are not generated because for one level scoped vlv, the entryid is not properly set.
  should use  vlv_grok_new_import_entry to reset the vlv filter when the base entry is added (as it is done in bdb).
  also added a function to mark the vlv_grok_new_import_entry as uninitialized before the import 

mdb- crash while trying to import an entry without parent (i.e a suffix entry) that does not belong to the backend
  fixed by avoiding the null pointer exception in that case

Issue: #6136 

Reviewed by: @droideck, @jchapma (Thanks!)


progier389:
vlv CI test is hanging on lmdb standalone instance is looping withing dbmdb_rm_db_file,
   looks like there is a problem with ai_dblayer_count  (The surprising thing is that I do not hit it when running on a non root PREFIXed build)

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -5785,9 +5785,8 @@ bdb_restore(struct ldbminfo *li, char *src_dir, Slapi_Task *task)
 static char *
 bdb__import_file_name(ldbm_instance *inst)
 {
-    char *fname = slapi_ch_smprintf("%s/.import_%s",
-                                    inst->inst_parent_dir_name,
-                                    inst->inst_dir_name);
+    struct ldbminfo *li = inst->inst_li;
+    char *fname = slapi_ch_smprintf("%s/.import_%s", li->li_directory, inst->inst_name);
     slapi_log_err(SLAPI_LOG_DEBUG, "bdb__import_file_name", "DBG: fname=%s\n", fname);
```
droideck: It's not related exactly to this PR, but it still caught my eye. Do you know of the double underscore (bdb__) is intentional in this function name or just a typo?

dirsrvtests/tests/suites/vlv/regression_test.py
```diff
@@ -103,53 +112,53 @@ def check_vlv_search(conn, **kwargs):
...
+def add_users(inst, users_num, suffix=DEFAULT_SUFFIX):
+    users = UserAccounts(inst, suffix)
     log.info(f'Adding {users_num} users')
     for i in range(0, users_num):
         uid = STARTING_UID_INDEX + i
         add_user(inst, users, uid)
 
 
-def create_vlv_search_and_index(inst, hyphen='', basedn=DEFAULT_SUFFIX):
+def create_vlv_search_and_index(inst, basedn=DEFAULT_SUFFIX, bename='userRoot', scope=2, prefix="vlv"):
```
droideck: For better readability, I think we should use `ldap.LDAP_SCOPE_SUBTREE` for the scope here, instead of magic number.

dirsrvtests/tests/suites/vlv/regression_test.py
```diff
@@ -170,6 +179,120 @@ def cleanup(inst):
...
+        be = bes.create(properties={
+            'cn': bename,
+            'nsslapd-suffix': suffix,
+        })
+        # Add suffix entry
+        Organization(inst, dn=suffix).create(properties={ 'o': bename, })
+        # Configure vlv
+        vlv_search, vlv_index = create_vlv_search_and_index(
+            inst, basedn=suffix,
+            bename=bename, scope=1,
```
droideck: Same here regarding the scope.

droideck (approved):
Besides the minor issues, looks good!

jchapma:
I reviewed the C files and it more or less makes sense to me. Re the CI test, it must be the attribute info db layer count that is not being decremented, surely...

Ack

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
progier389: There is no name conflict so IMHO it is a typo, I will change it.

dirsrvtests/tests/suites/vlv/regression_test.py
progier389: OK, will use ldap.SCOPE_SUBTREE instead of 2

dirsrvtests/tests/suites/vlv/regression_test.py
progier389: OK, will use ldap.SCOPE_ONELEVEL instead of 1

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -456,6 +456,7 @@ dbmdb_open_all_files(dbmdb_ctx_t *ctx, backend *be)
             TST(add_dbi(&octx, be, special_names[i], ctxflags));
             sn_dbis[i] = octx.dbi;
         }
+        TST(add_dbi(&octx, be, LDBM_LONG_ENTRYRDN_STR, ctxflags));
```
tbordaz, line 459: Any reason why having not added LDBM_LONG_ENTRYRDN_STR in special_names ?

ldap/servers/slapd/back-ldbm/vlv.c
```diff
@@ -362,11 +361,60 @@ vlv_grok_new_import_entry(const struct backentry *e, backend *be)
...
+             */
+            continue;
+        }
+        p->vlv_initialized = 0;
+        if (!slapi_sdn_isempty(p->vlv_base)) {
+            struct backentry *e = NULL;
+            entry_address addr;
+            addr.sdn = p->vlv_base;
+            addr.uniqueid = NULL;
+            e = find_entry(pb, be, &addr, txn, NULL);
```
tbordaz, line 403: If the base entry does not exist at that time, it will stay vlv_initialized=0. That looks fine to me. Does it require documentation ?

ldap/servers/slapd/back-ldbm/vlv_srch.c
```diff
@@ -776,6 +776,7 @@ vlvIndex_checkforindex(struct vlvIndex *p, backend *be)
         /* In lmdb case, always open the dbi */
         if (li->li_flags & LI_LMDB_IMPL) {
             (void) dblayer_get_index_file(be, p->vlv_attrinfo, &db, 0) ;
+            dblayer_release_index_file(be, p->vlv_attrinfo, db);
```
tbordaz, line 779: It looks open/close. What is it for ?

tbordaz (commented):
LGTM
few questions

ldap/servers/slapd/back-ldbm/vlv.c
progier389: Interesting point. IMHO the right thing to do would be to check the uninitialized vlv when adding an entry ...

ldap/servers/slapd/back-ldbm/vlv_srch.c
progier389: dblayer_release_index_file decrease a ref count to tell that the index is no more used.
On bdb it could lead to db instance closure but that is not the case with lmdb  (db instances are not closed until closing the whole db environment (i.e at shutdown time) )
So the open/release partern just create/open the database instance and keep it open (but not busy).
Having it open is important because it cannot be open when needed by a search operation: 
because there is an already open read txn and there is no way we can open a new file within that txn:
We can open the file with a new write txn, but in that case the previously created read txn has no visibility of the new file.
The big rule to remember with LMDB is is that ALL DATABASE INSTANCES MUST BE OPEN when bringing a backend on-line. The only case we could open/create db instance is when changing the config.

PR #6143, opened 2024-04-09
Issue 6142 - [RFE] Add LMDB configuration related checks into Healthcheck tool

### LINKED ISSUE #6142: [RFE] Add LMDB configuration related checks into Healthcheck tool
**Description**
Add a warning in healthcheck if bdb is still used.
Add warnings if there's a mismatch in configuration attributes.

droideck:
Description
Add a warning in healthcheck if bdb is still used. Add warnings if there's a mismatch in configuration attributes.

Fixes: https://github.com/389ds/389-ds-base/issues/6142

Reviewed by: @progier389 (Thanks!)

### DIFF
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -34,7 +34,7 @@ from lib389.encrypted_attributes import EncryptedAttr, EncryptedAttrs
 # This is for sample entry creation.
 from lib389.configurations import get_sample_entries
 
-from lib389.lint import DSBLE0001, DSBLE0002, DSBLE0003, DSVIRTLE0001, DSCLLE0001
+from lib389.lint import DSBLE0001, DSBLE0002, DSBLE0003, DSBLE0004, DSBLE0005, DSBLE0006, DSVIRTLE0001, DSCLLE0001
 
 
 class BackendLegacy(object):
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/backend.py
```diff
@@ -510,6 +510,82 @@ def _lint_mappingtree(self):
...
+        mdb_files.sort()
+        bdb_files.sort()
+
+        if (backend_impl == 'bdb' and not all(file in db_files for file in bdb_files)) or \
+           (backend_impl == 'mdb' and not all(file in db_files for file in mdb_files)):
+            result = DSBLE0004
+            result['items'] = [self.lint_uid()]
+            yield result
+
+    def _lint_backend_config_attributes(self):
```
progier389: I am not sure that we should check the config attribute: I am pretty sure that there are leftover after a migration from bdb to lmdb or the opposite. and the attribute are not invalid, they are ignored. 
If we still decide to keep it as warning, IMHO severity should be Low (nothing is broken but cleanup was not done)

src/lib389/lib389/backend.py
```diff
@@ -510,6 +510,82 @@ def _lint_mappingtree(self):
...
+
+    def _lint_backend_implementation(self):
+        backend_impl = self._instance.get_db_lib()
+        if backend_impl == 'bdb':
+            result = DSBLE0006
+            result['items'] = [self.lint_uid()]
+            yield result
+
+
+    def _lint_backend_implementation_mismatch(self):
```
progier389: I wonder if this should be tested.
FYI: If they are not already existing, the dbfiles are created when the server is started.
So such a warning will be transient  (and seen when performing a manual migration or having explicitly removed the db files)

src/lib389/lib389/backend.py
droideck: To my understanding, it's a thing that may lead to issues, but it's not an issue itself (for example, an administrator will expect certain configurations to work, but it won't because it's "dead attributes" now). So it's better to notify them via health check.

I agree that Low is better. I was harsh to set higher severity. :D

src/lib389/lib389/backend.py
droideck: Same here; it's more of a clean-up, low-severity health check, so the administrator is aware that some things still need to be done for a clean migration finish.

But I don't have a strong opinion about this one. If you think it's really unnecessary, I'm okay with removing it.

src/lib389/lib389/backend.py
progier389: Ok. I am  fine about keeping it and use low severity.

src/lib389/lib389/backend.py
progier389: The more I think about it, the more I think that the interesting warning is not if the db files are missing 
 because simply restarting the server will fix recreate them but rather if the db file for the both implementations are present (meaning that both database `dsctl instance dblib cleanup` has not yet been run)
  and it is still a minor severity (Useless data not removed)

progier389 (approved):
LGTM

droideck:
7657726dc..06fef7342  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6144, opened 2024-04-09
Issue 6141 - freeipa test_topology_TestCASpecificRUVs is failing

### LINKED ISSUE #6141: lmdb - freeipa test_topology_TestCASpecificRUVs is failing
**Issue Description**
https://github.com/freeipa-pr-ci2/freeipa/pull/3558 shows that 389ds-fedora/test_topology_TestCASpecificRUVs test is still failing.

 - Platform: Fedora
 - Package and version: 3.0.1 main branch nightly builds
 - Browser N/A

**Steps to Reproduce**
Loom at the **[testing_master_389ds] Nightly**  bi-weekly tests result on [https://github.com/freeipa-pr-ci2/freeipa/pulls](https://github.com/freeipa-pr-ci2/freeipa/pulls) 

**Expected results**
Test **test_topology_TestCASpecificRUVs** should pass
Only **test_topology_TestCASpecificRUVs** is expected to fail

**Additional context**
Should also add  this test as a pure 389ds CI test

progier389:
On lmdb, vlv search using a value instead of range may fail (set target on first record instead of smallest record whose key is greater of equal to the wanted value).
The reason is that a test is inverted when walking the cursor to find the record position so the loop end after first iteration. 
Also fix a coverity scan warning

Issue: #6141 

Reviewed by: @tbordaz

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
@@ -2338,17 +2338,17 @@ int dbmdb_cursor_get_recno(dbi_cursor_t *cursor, MDB_val *dbmdb_key, MDB_val *db
         rc = MDB_CURSOR_OPEN(mdb_cursor_txn(cursor->cur), mdb_cursor_dbi(cursor->cur), &newcur);
     }
     if (rc == 0) {
-        rc = MDB_CURSOR_GET(newcur, &rce->key, &rce-> data, MDB_SET);
+        rc = MDB_CURSOR_GET(newcur, &rce->key, &rce->data, MDB_SET);
     }
     while (rc == 0) {
         cmpres = dbmdb_cmp_dbi_record(mdb_cursor_dbi(cursor->cur), &curpos_key, &curpos_data, &rce->key, &rce->data);
-        if (cmpres >= 0) {
+        if (cmpres <= 0) {
             break;
         }
         rce->recno++;
         rc = MDB_CURSOR_GET(newcur, &rce->key, &rce->data, MDB_NEXT);
     }
-    if (cmpres > 0) {
+    if (cmpres < 0) {
         rc = MDB_NOTFOUND;
     }
     if (rc == 0) {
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
```diff
@@ -2338,17 +2338,17 @@ int dbmdb_cursor_get_recno(dbi_cursor_t *cursor, MDB_val *dbmdb_key, MDB_val *db
         rc = MDB_CURSOR_OPEN(mdb_cursor_txn(cursor->cur), mdb_cursor_dbi(cursor->cur), &newcur);
     }
     if (rc == 0) {
-        rc = MDB_CURSOR_GET(newcur, &rce->key, &rce-> data, MDB_SET);
+        rc = MDB_CURSOR_GET(newcur, &rce->key, &rce->data, MDB_SET);
     }
     while (rc == 0) {
         cmpres = dbmdb_cmp_dbi_record(mdb_cursor_dbi(cursor->cur), &curpos_key, &curpos_data, &rce->key, &rce->data);
-        if (cmpres >= 0) {
+        if (cmpres <= 0) {
```
tbordaz, line 2345: To be honest I read and reread this code several times and still do not understand it.
I understand that rce starts from the beginning of index and count recno up to curpos_key.
I wound assume that rce.key is lower than cursor_key during the loop and that we stop the loop when it gets higher or equal (the previous code).

The new code is working even if I do not understand it ;).
I would appreciate comments or examples in the code to explain the loop.

tbordaz (approved):

ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
progier389: > I understand that rce starts from the beginning of index and count recno up to curpos_key.
> I wound assume that rce.key is lower than cursor_key during the loop and that we stop the loop when it gets higher or equal

You are right up to that point: the loop should go on while  cursor_key > rce.key
and should stop it when cursor_key <= rce.key

> (the previous code).

That is wrong: cursor_key <= rce.key is cmpres <= 0 (the new code)
( Looks like I am not the only one that inverts the test ! :wink: )

PR #6152, opened 2024-04-19
Issue 6151 - Use `%bcond` macro for conditional builds in the  spec file

### LINKED ISSUE #6151: Use `%bcond` macro for conditional builds in the spec file
**Issue Description**
`rpmbuild` supports conditional package builds with the command line switches `--with` and `--without`:
https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html

This is useful to rebuild an existing src.rpm file without editing the spec file first. Or build in COPR with macros overrides for some options. For example, automatic rebuilds from dist-git in COPR to produced sanitized builds.

We use our custom global variables for different options such as `use_cockpit` or `use_asan`. Instead we should switch to `%bcond` macro.

vashirov:
`rpmbuild` supports conditional package builds with the command line
switches `--with` and `--without`:
https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html

This is useful to rebuild an existing src.rpm file without editing the
spec file first. Or build in COPR with macros overrides for some
options. For example, automatic rebuilds from dist-git in COPR to
produced sanitized builds.

We use our custom global variables for different options such as
`use_cockpit` or `use_asan`. Instead we should switch to `%bcond` macro.

Fixes: https://github.com/389ds/389-ds-base/issues/6151


vashirov:
Build failed in CI with:
```
python3 rpm/bundle-rust-npm.py src/ src/cockpit/389-console/ /__w/389-ds-base/389-ds-base/rpmbuild/SPECS/389-ds-base.spec -f
  File "/__w/389-ds-base/389-ds-base/rpm/bundle-rust-npm.py", line 23, in <module>
    from lib389.cli_base import setup_script_logger
ModuleNotFoundError: No module named 'lib389'
make: *** [rpm.mk:169: rpms] Error 1
```

`rpm/bundle-rust-npm.py` uses functions from `lib389` module, which is not installed. I think our build system should not depend on the `lib389` library. The code is trivial and I think we can just copy it to `rpm/bundle-rust-npm.py`. @droideck WDYT?

rpm/bundle-rust-npm.py
```diff
@@ -263,8 +266,8 @@ def write_provides_bundled(provides_lines: List[str], spec_file: str, cleaned: b
...
     license_string = create_license_line(rust_crates, npm_packages)
     replace_license(args.spec_file, license_string)
-    log.info(f"Specfile {args.spec_file} is successfully modified! Please:\n"
-              "1. Open the specfile with your editor of choice\n"
-              "2. Make sure that Provides with bundled crates are correct\n"
-              "3. Follow the instructions for 'License:' field and remove the helper comments")
+    print(f"Spec file {args.spec_file} is successfully modified! Please:")
+    print("1. Open the spec file with your editor of choice")
+    print("2. Make sure that 'Provides:' with bundled crates are correct")
+    print("3. Follow the instructions for 'License:' field and remove the helper comments")
```
droideck: Minor nitpick - we probably don't need this last line when we do a `fix-it` run.

droideck (approved):
Besides that, all looks good!

But there was a bit of something strange. I saw an issue with cargo-license error:
https://github.com/389ds/389-ds-base/actions/runs/8799564272/job/24148859806

```
python3 rpm/bundle-rust-npm.py src/ src/cockpit/389-console/ /__w/389-ds-base/389-ds-base/rpmbuild/SPECS/389-ds-base.spec -f
ERROR:bundle-rust-npm:cargo-license is not installed. Please install it with 'cargo install cargo-license' and try again.
```

But when I did a rerun, it disappeared (and the run after that). Can it be some weird race condition with containers?

rpm/bundle-rust-npm.py
vashirov: Yes, I agree. I have another series of changes coming up, one of them implements this :)

vashirov:
> But when I did a rerun, it disappeared (and the run after that). Can it be some weird race condition with containers?

The container image didn't have `cargo-license`, I fixed it in https://github.com/389ds/ci-images/commit/c569be383a88bad3c23650fce24edf7afb41cf4e. You probably had the old container image.

rpm/bundle-rust-npm.py
vashirov: @droideck, sorry for the delay, I pushed another series of commits, please take a look.

droideck (approved):
Looks great! Ack

Issue #6159, opened 2024-04-30
Add a test to check URP add and delete conflict


progier389:
The goal is simply to add some test about URP conflict handling in the ADD - DEL because I got some surprise when dealing with conflict entries. 
One of the test is  performing ADD,  SYNC TO REPLICA A, SYNC TO REPLICA B, DEL and test all orders of that sequence between 3 replica
 Note: there are about 5700 orders and the test spend 6 days
A second test tries to clean conflict entry in a weird case (crossed entries that should theoretically not happen but was sometime seen anyway)

Theses tests requires URP_VERY_LONG_TEST environment variable to run.

Note: 
 Initially I though about trying all the order of single actions on the 3 replica but that is more than 1M orders and 3.5 years of test ! -;)

PR #6160, opened 2024-04-30
Issue 6159 - Add a test to check URP add and delete conflict

progier389:
Add URP tests that run if URP_VERY_LONG_TEST environment variable is set One test spends 6 days and check the 5770 different way of running the (Add, sync agmt 1, sync agmt 2, Del) sequence on 3 suppliers and check that when everything is in sync, the entries are the same everywhere Second test generate crossed entries and conflict entries
 (In theory that should not happen but we have sometime seen them)
And tries to remove one of the entry.
Then once everything is back in sync, it check that the entry are the same
 The second test fails - Apparently there is a problem in that corner case

Issue: #6159 

### DIFF
--- a/dirsrvtests/tests/suites/replication/urp.py
+++ b/dirsrvtests/tests/suites/replication/urp.py
@@ -1,0 +1,379 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import pytest, time
+import os
+import glob
+import ldap
+from shutil import copyfile, rmtree
+from itertools import permutations, product
+from contextlib import contextmanager
+from lib389.tasks import *
+from lib389.utils import *
+from lib389.topologies import topology_m3, topology_m2, set_timeout
+from lib389.replica import *
+from lib389._constants import *
+from lib389.properties import TASK_WAIT
+from lib389.index import *
+from lib389.mappingTree import *
+from lib389.backend import *
+from lib389.conflicts import ConflictEntry
+from lib389.idm.user import UserAccounts, UserAccount
+from lib389.idm.organization import Organization
+from lib389.agreement import Agreements
+from lib389.idm.organizationalunit import OrganizationalUnits
+
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+SKIP_THE_TESTS = os.getenv('URP_VERY_LONG_TEST') is None
+
+# Replication synchronization timeout in seconds
+REPL_SYNC_TIMEOUT = 300
+
+# test_urp_delete is a long test spending days
+set_timeout(24*3600*10)
+
+
+def normalize_values(v):
+    # Return lower case sorted values.
+    if isinstance(v, bytes):
+        return v.decode('utf-8').lower()
+    result = [ normalize_values(val) for val in v ]
+    result.sort()
+    return result
+
+
+def get_entry_info(entry):
+    norm_entry = { k.lower():normalize_values(v) for k,v in entry.items() }
+    uuid = norm_entry['nsuniqueid'][0]
+    if 'nstombstone' in norm_entry['objectclass']:
+        return "tombstone:"+uuid
+    elif 'nsds5replconflict' in norm_entry:
+        return "conflict:"+uuid
+    else:
+        return "regular:"+uuid
+
+
+def search_entries(inst, dnfilter):
+    ldapurl, certdir = get_ldapurl_from_serverid(inst.serverid)
+    assert 'ldapi://' in ldapurl
+    conn = ldap.initialize(ldapurl)
+    conn.sasl_interactive_bind_s("", ldap.sasl.external())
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/replication/urp.py
```diff
@@ -0,0 +1,379 @@
...
+            self.insts = [ UrpTesterInstances(self, inst) for inst in topo ]
+            inst = topo[0]
+            self.ldif = f'{inst.get_ldif_dir()}/userroot.ldif'
+            tasks = Tasks(inst)
+            tasks.exportLDIF(DEFAULT_SUFFIX, output_file=self.ldif, args={EXPORT_REPL_INFO: True, TASK_WAIT: True})
+
+        def resync_agmt(self, instfrom, instto):
+            log.info(f"Enabling replication agreement from {instfrom.serverid} to {instto.serverid}")
+            self.repl.enable_to_supplier(instto, [instfrom,])
+            self.repl.wait_for_replication(instfrom, instto, timeout=REPL_SYNC_TIMEOUT)
```
droideck, line 222: It probably depends on the machine (I had a pretty slow VM) but I got at least 3 failures in `test_urp_delete` here with `Replication from ldap://inst3:39003 to ldap://inst2:39002 is NOT working. (too many retries)`

Probably, it's expected here or we can tune it a bit? (or make it flaky with some retries).
Still, the test work and looks good as it is.

dirsrvtests/tests/suites/replication/urp.py
```diff
@@ -0,0 +1,379 @@
...
+    import_ldif1 = f'{s1.get_ldif_dir()}/db1.ldif'
+    import_ldif2 = f'{s1.get_ldif_dir()}/db2.ldif'
+    Tasks(s1).exportLDIF(DEFAULT_SUFFIX, output_file=export_ldif, args={EXPORT_REPL_INFO: True, TASK_WAIT: True})
+    gen_ldif_file(export_ldif, f'{datadir}/db1.ldif', import_ldif1)
+    gen_ldif_file(export_ldif, f'{datadir}/db2.ldif', import_ldif2)
+    # 2. import the ldif files
+    Tasks(s1).importLDIF(DEFAULT_SUFFIX, input_file=import_ldif1, args={TASK_WAIT: True})
+    Tasks(s2).importLDIF(DEFAULT_SUFFIX, input_file=import_ldif2, args={TASK_WAIT: True})
+    # 3. Remove a conflict entry
+    ConflictEntry(s1, 'cn=u22449+nsUniqueId=c6654281-f11b11ee-ad93a02d-7ba2db25,dc=example,dc=com').delete()
```
droideck, line 361: I have a failure here, though:

`E         ldap.OPERATIONS_ERROR: {'msgtype': 107, 'msgid': 88, 'result': 1, 'desc': 'Operations error', 'ctrls': [], 'ldap_request': "delete_ext_s(('cn=u22449+nsUniqueId=c6654281-f11b11ee-ad93a02d-7ba2db25,dc=example,dc=com',),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance supplier1"}`

When I run it standalone:

`[root@inst ds]# URP_VERY_LONG_TEST=yes py.test-3 -s -v dirsrvtests/tests/suites/replication/urp.py::test_urp_with_crossed_entries`

droideck (changes requested):

dirsrvtests/tests/suites/replication/urp.py
progier389: On a relatively fast machine, the test is already needing 6 days so IMHO it is not realistic to 
 perform retries or run on slow machine. But I could add comment that the test should run on a fast machine and require memory (as there is 3 instances running even if the db are tiny )

dirsrvtests/tests/suites/replication/urp.py
progier389: Although I know that this test is failing, your failure is unexpected:
I have a different behavior on my machine:

```
URP_VERY_LONG_TEST=yes py.test-3 -s -v $PWD/urp.py::test_urp_with_crossed_entries
...
>       assert u1 == u2
E       AssertionError: assert ['regular:cd8...02d-7ba2db25'] == ['conflict:cd...02d-7ba2db25']
...
```

Probably worth to try to understand why we get different results

progier389:
Added a new commit to:
1. Add a comment about using reasonably fast machine
2. Assert that import tasks were successful
3. Set ldif file permission so that they are readable by everyone (dirsrv) 

The permission issue was the reason why the tests results are different when using standard package compared to my environment (The import failed and the db was empty)

dirsrvtests/tests/suites/replication/urp.py
droideck: Rechecked, and I got the same error as you now:

```
        u1 = search_entries(s1, 'cn=u22449')
        u2 = search_entries(s2, 'cn=u22449')
>       assert u1 == u2
E       AssertionError: assert ['regular:cd8...02d-7ba2db25'] == ['conflict:cd...02d-7ba2db25']
```

progier389: Ok, this time it is the expected result.   URP does not properly handle this case 
BTW, I should add a @pytest.mark.xfail for this one so it will be explicit.

progier389:
Added the xfail, now the test result is 
urp.py::test_urp_with_crossed_entries XFAIL (URP does not properly handle this case)                             [100%]

droideck (approved):
LGTM

PR #6167, opened 2024-05-13
fix issue6165

### LINKED ISSUE #6166: Server crash when using the referential integrity plugin in the case when processing a line from the logfile
**Issue Description**
Server crash when using the referential integrity control plugin in the case when processing a line from the _logfile_

**Steps to Reproduce**
1) enable the plugin using transactions (betxnpostoperation)
2) set delay > =1
3) remove permission to write to the directory where the log file will be created 
4) move entry from containerscope
5) server crash

**Expected results**
Logging an error message and continuing work

**Additional context**
When processing the next line **GetNextLine(thisline, MAX_LINE, prfd)**
the **ldap_utf8strtok_r** function may return _null_.

When passing _null_ to **strcasecmp**, a crash may occur inside.
The _first_ call **ldap_utf8strtok_r(NULL** and the _ptoken_ assignment are checked for _null_, the _second_ and _third_ are not.

iamarty3z:
Problem: Server crash when using the referential integrity plugin when transactions are used and an error occurs when opening a file for writing.
Cause: The crash is caused by using an uninitialized mutex PR_Unlock(referint_mutex)) after the error message and before calling referint_unlock();
Fix: The line using the uninitialized mutex **PR_Unlock(referint_mutex)** has been _removed_. It opens with an initialization check in the function below - **referint_unlock().**

Issue: #6166 

Reviewed by: @progier389

### DIFF
--- a/ldap/servers/plugins/referint/referint.c
+++ b/ldap/servers/plugins/referint/referint.c
@@ -1627,7 +1627,6 @@ writeintegritylog(Slapi_PBlock *pb, char *logfilename, Slapi_DN *sdn, char *newr
                       "writeintegritylog - Could not write integrity log \"%s\" " SLAPI_COMPONENT_NAME_NSPR " %d (%s)\n",
                       logfilename, PR_GetError(), slapd_pr_strerror(PR_GetError()));
 
-        PR_Unlock(referint_mutex);
         referint_unlock();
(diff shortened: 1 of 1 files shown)


progier389 (approved):
Looks good to me.

Issue #6172, opened 2024-05-15
RFE: improve the performance of evaluation of filter component when tested against a large valueset (like group members)


tbordaz:
**Is your feature request related to a problem? Please describe.**
Before returning an entry (to a SRCH) the server check that the SRCH filter matches the entry. If a component of the filter is testing a large valueset in the candidate entry, it is expensive to check the matching.
A typical issue is with a component like '(uniquemember=foo)' and the candidate entry is a group containing thousands of 'uniquemember'.

1. Create 10000 groups and 10000 users (with a user 'foo')
2. Make all groups containing all users.
3. do search with filter "(&(objectclass=groupofuniquenames)(uniqueMember=foo))"
4. response time is long because of cache priming
5. do search with filter "(&(objectclass=groupofuniquenames)(uniqueMember=foo))" is better but still slow

Once primed the second search should be much faster

**Describe the solution you'd like**
Proposed an option to use the sorted valueset when checking the match (ava test)

**Describe alternatives you've considered**

**Additional context**
I initially thought that the performance issue was related to a filter bypass (on indexed searches) #6030  but I doubt that the bypass is always possible. So I prefer to make the test of the match faster.

tbordaz:
884deb60a..904dc9906 main

PR #6173, opened 2024-05-15
Issue 6172 - RFE: improve the performance of evaluation of filter com…

tbordaz:
…ponent when tested against a large valueset (like group members)

Bug description:
Before returning an entry (to a SRCH) the server checks that the entry matches the SRCH filter. If a filter component (equality) is testing the value (ava) against a large valueset (like uniquemember values), it takes a long time because of large number of value and required normalization of the values. This can be improved taking benefit of sorted valueset. Those sorted valueset were created to improve updates of large valueset (groups) but not used in SRCH path.

Fix description:
if config param 'nsslapd-filter-match-use-sorted-vs = on' then it uses slapi_valueset_find (that tries to use sorted valueset) rather than plugin_call_syntax_filter_ava.
In both case sorted valueset and plugin_call_syntax_filter_ava, ava and values are normalized.
In sorted valueset, the values have been normalized to insert the index in the sorted array and then comparison is done on normalized values. In plugin_call_syntax_filter_ava, all values in valuearray (of valueset) are normalized before comparison.

Likely this optimization should be dropped for extended search

relates: #6172

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/filter/filter_test.py
+++ b/dirsrvtests/tests/suites/filter/filter_test.py
@@ -9,7 +9,11 @@
 import logging
 
 import pytest
+import time
+from lib389.dirsrv_log import DirsrvAccessLog
 from lib389.tasks import *
+from lib389.backend import Backends, Backend
+from lib389.dbgen import dbgen_users, dbgen_groups
 from lib389.topologies import topology_st
 from lib389._constants import PASSWORD, DEFAULT_SUFFIX, DN_DM, SUFFIX
 from lib389.utils import *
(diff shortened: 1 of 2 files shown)


progier389 (changes requested):
The idea of using directly the valueset is good.
but it can only be done if  ftype == LDAP_FILTER_EQUALITY 
and in this case I do not think that using yet another config parameter is really needed (in 389ds, equality is implement by comparing normalized values anyway)
That said, IMHO we should also write some CI tests for that one:
  - to check the scenario you describe 
  - and also a scenario storing data on a case ignore string attribute then  searching  for uppercase then for lowercase  (To double check that the filter values are properly normalized)

progier389 (approved):
Looks good except a minor point: 
I would change the "else" comment to something easier to understand, for example:
When sorted valuearray optimization cannot be used lets filter the value according to its syntax"

droideck (approved):
Looks good

tbordaz:
@progier389 , @droideck , Thanks for the review. I will update the patch accordingly in addition with a testcase I am working on.

tbordaz:
I added a testcase to the patch

PR #6179, opened 2024-05-28
Issue 6155 - ldap-agent fails to start because of permission error

### LINKED ISSUE #6155: ldap-agent fails to start because of permission error
**Issue Description**

 - Platform: Fedora 39
 - Browser N/A

**Steps to Reproduce**
1. dnf install -y 389-ds-base 389-ds-base-snmp net-snmp
2. echo master agentx >> /etc/snmp/snmpd.conf
4. systemctl restart snmpd
5. dscreate interactive  (Create a ds instance as usual)
6. edit  /etc/dirsrv/config/ldap-agent.conf to add the new instance
7. systemctl start dirsrv-snmp
8. See the error (instance fails to start)

**Expected results**
dirsrv-snmp service should be able to start

**Additional context**

/run/dirsrv  is created with the wrong group should be root instead of dirsrv so that both 
 ns-slapd and ldap-agent can read and create files in it.
Fix should be in src/lib389/lib389/instance/setup.py around # Setup tmpfiles_d

progier389:
Issue: dirsrv-snmp service fails to starts when SELinux is enforced because of AVC preventing to open some files
One workaround is to use the dac_override capability but it is a bad practice.
Fix:  Setting proper permissions:
  - Running ldap-agent with uid=root and gid=dirsrv to be able to access both snmp and dirsrv resources.
  - Setting read permission on the group for the dse.ldif file
  - Setting r/w permissions on the group for the snmp semaphore and mmap file
     For that one special care is needed because ns-slapd umask overrides the file creation permission
     as is better to avoid changing the umask (changing umask within the code is not thread safe, 
     and the current 0022 umask value is correct for most of the files) so the safest way is to chmod the snmp file 
     if the needed permission are not set.
     
Issue: #6155

Reviewed by: @droideck , @vashirov (Thanks ! )

### DIFF
--- a/dirsrvtests/tests/suites/snmp/snmp.py
+++ b/dirsrvtests/tests/suites/snmp/snmp.py
@@ -1,0 +1,214 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import os
+import pytest
+import logging
+import subprocess
+import ldap
+from datetime import datetime
+from shutil import copyfile
+from lib389.topologies import topology_m2 as topo_m2
+from lib389.utils import selinux_present
+
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+
+SNMP_USER = 'user_name'
+SNMP_PASSWORD = 'authentication_password'
+SNMP_PRIVATE = 'private_password'
+
+# LDAP OID in MIB
+LDAP_OID = '.1.3.6.1.4.1.2312.6.1.1'
+LDAPCONNECTIONS_OID = f'{LDAP_OID}.21'
+
+
+def run_cmd(cmd, check_returncode=True):
+    """Run a command"""
+
+    log.info(f'Run: {cmd}')
+    result = subprocess.run(cmd, capture_output=True, universal_newlines=True)
+    log.info(f'STDOUT of {cmd} is:\n{result.stdout}')
+    log.info(f'STDERR of {cmd} is:\n{result.stderr}')
+    if check_returncode:
+        result.check_returncode()
+    return result
+
+
+def add_lines(lines, filename):
+    """Add lines that are not already present at the end of a file"""
+
+    log.info(f'add_lines({lines}, {filename})')
+    try:
+        with open(filename, 'r') as fd:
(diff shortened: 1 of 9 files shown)


progier389:
Note: 
 The test also show that two rules are missing in the ldap SELinux policy to allow the use of mmap on the snmp semaphore file and on the snmp mmap file:
     allow dirsrv_snmp_t dirsrv_tmpfs_t:file map;
     allow dirsrv_snmp_t dirsrv_var_run_t:file map;

dirsrvtests/tests/suites/snmp/snmp.py
```diff
@@ -0,0 +1,160 @@
...
+def setup_snmp(topo_m2, request):
+    """Install snmp and configure it"""
+
+    inst1 = topo_m2.ms["supplier1"]
+    if not inst1.with_systemd_running():
+        pytest.skip('This test requires systemd')
+        return None
+
+    # Install snmp
+    run_cmd(['/usr/bin/dnf', 'install', '-y', 'net-snmp', 'net-snmp-utils', '389-ds-base-snmp'])
```
vashirov: We should not install anything during the tests, instead we should check if the required executables are present, and skip the test if they are missing. See for example https://github.com/389ds/389-ds-base/blob/c019af140ca53955e67038e1c9fe18620410638d/dirsrvtests/tests/suites/basic/basic_test.py#L1230-L1231

progier389: Ok.
IMHO, we should also generate a quay.io test image having net-snmp and net-snmp-utils packages
BTW (we do not hit the map AVC when running the test on that image (unlike in F40))

vashirov: Ok, I pushed a new image based on F40 containing net-snmp and net-snmp-utils.
We won't see any AVC denials in GH actions because the tests are running in a container on Ubuntu VM, that doesn't use SELinux.

progier389: Thanks for the new image. 
Indeed I have not thought to verify if SELinux is enforced.
This interresting mistake let me think that I should split the test in two tests (one  about running snmp smoke test) and the second one about checking  AVC (and it will be skipped if SELinux is not enabled). 
At least it will help us to realize that we are not testing all the cases! :wink:

progier389:
Split the test in two and skip AVC test if SELinux is not enabled. 
Fixed a few errors about missing  packages handling
Fixed a problem with the curtime generation 
FYI: Both testcases are PASS on Fedora40 1minutetip VMs after installing required packages  and fixing SELinux policy ldap-agent rule ("2 map permissions are missing"

droideck (approved):
LGTM!

progier389:
3fe566124..eb7e57d77  main -> main
   0fb74967f..c3ff6e8ff  389-ds-base-3.0 -> 389-ds-base-3.0
   cf6cdd05b..7cb34ed42  389-ds-base-2.5 -> 389-ds-base-2.5

Issue #6170, opened 2024-05-14
Audit log buffering is dropping info


mreynolds389:
**Issue Description**

Noticed if I add a large static group that most members in the operation are not present in the log.  This appears to be somehow related to the new buffering that was recently implemented.

mreynolds389:
75d034f08..b6c2d789b  389-ds-base-3.0 -> 389-ds-base-3.0
dd84b1f2c..fad9ff38b  389-ds-base-2.5 -> 389-ds-base-2.5
d3115084c..0e19d4e80  389-ds-base-2.4 -> 389-ds-base-2.4
78bbc02ff..da3cd1a67  389-ds-base-2.3 -> 389-ds-base-2.3
291094d1d..776407cb9  389-ds-base-2.2 -> 389-ds-base-2.2
798d68bc2..3c7813341  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6180, opened 2024-05-28
Issue 6170 - audit log buffering doesn't handle large updates

mreynolds389:
Description:

A large update, like adding 10K memebrs to a group, gets truncated. When the update is larger than the buffer then flush the current buffer, and then directly write the large update to the log file (skipping the buffering)

Relates: https://github.com/389ds/389-ds-base/issues/6170

### DIFF
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -445,9 +445,6 @@ g_log_init()
     loginfo.log_audit_logchain = NULL;
     loginfo.log_audit_buffer = log_create_buffer(LOG_BUFFER_MAXSIZE);
     loginfo.log_audit_compress = cfg->auditlog_compress;
-    if ((loginfo.log_audit_rwlock = slapi_new_rwlock()) == NULL) {
-        exit(-1);
-    }
     if (loginfo.log_audit_buffer == NULL) {
         exit(-1);
     }
@@ -481,9 +478,6 @@ g_log_init()
     loginfo.log_auditfail_buffer = log_create_buffer(LOG_BUFFER_MAXSIZE);
     loginfo.log_backend = LOGGING_BACKEND_INTERNAL;
     loginfo.log_auditfail_compress = cfg->auditfaillog_compress;
-    if ((loginfo.log_auditfail_rwlock = slapi_new_rwlock()) == NULL) {
-        exit(-1);
-    }
     if (loginfo.log_auditfail_buffer == NULL) {
         exit(-1);
     }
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/log.c
```diff
@@ -2417,25 +2417,41 @@ auditfail_log_openf(char *pathname, int locked)
...
 
+    /*
+     * An audit entry can be larger than the buffer, flush the current buffer
+     * and write new audit entry directly to the file
+     */
+    if (msg_len > SLAPI_LOG_BUFSIZ) {
+        PR_Lock(lbi->lock);
+        log_flush_buffer(lbi, SLAPD_AUDIT_LOG, 0 /* do not sync to disk */);
+        PR_Unlock(lbi->lock);
+        /* Now directly write the entire message to disk */
```
progier389: Something looks fuzzy with the lock handling.
There is a potential race condition if another thread writes in the buffer  between  PR_Unlock(lbi->lock); and LOG_AUDIT_LOCK_WRITE(); 
IMHO LOG_WRITE_NOW_NO_ERR should be called while protected by the PR_Lock(lbi->lock); as it is the case when called in log_buffer_flush

progier389 (changes requested):

ldap/servers/slapd/log.c
mreynolds389: Log rotation checks happen in log_flush_buffer() but yes the buffer can start to fill up again, but its ok since we are adding our update directly and should be in the correct order.  I could hold on to the log buffer lock through the entire process, but we have code in log_append_buffer2() that avoids holding this lock for a long time for performance reasons - so this is why I mimicked that code.

As for creating a second buffer, the update could exceed the buffer again and again.   Imagine adding 20K members to a group (this will exceed the buffer maxsize 20 times). It starts to become really complicated.   I'm just trying to keep it simple and just directly write the whole update at one shot.

progier389:
BTW: about the locks something looks fishy in existing code:
  sometime log_flush_buffer is protected by lbi->lock mutex and sometime by LOG_AUDIT_LOCK_WRITE rw lock 
IMHO logs_flush  code is wrong and the lock that should be held is the buffer lock (i.e loginfo.log_audit_buffer->lock) rather than the file lock

progier389:
Don't we have the same issue with vslapd_log_auditfail ?

mreynolds389:
> Don't we have the same issue with vslapd_log_auditfail ?

I do have the same code "fix" for auditfail

mreynolds389:
> BTW: about the locks something looks fishy in existing code: sometime log_flush_buffer is protected by lbi->lock mutex and sometime by LOG_AUDIT_LOCK_WRITE rw lock IMHO logs_flush code is wrong and the lock that should be held is the buffer lock (i.e loginfo.log_audit_buffer->lock) rather than the file lock

Yeah I just noticed that.  logs_flush() doesn't use the buffer lock at all, but the global log lock.  Hmmm yeah things are definitely off around this area.

ldap/servers/slapd/log.c
progier389: Hi agree about log rotation, I realized it just after having written the comment. :wink:

Sorry but both log_append_buffer2 and log_append_audit_buffer hold the lbi->lock buffer when effectively writing in the file (i.e when calling log_flush_buffer )
The only thing that is done out of the lock is to copy the data in the reserved area within the buffer 
 (but there is no buffer in the overlong message case)
==> you must hold lbi->lock buffer while writing directly on the file

Faking LogBufferInfo is  an overhead as log_flush_buffer will mostly calling LOG_WRITE_NOW_NO_ERR anyway
(That is why I also removed that idea)
But you are wrong about "the update could exceed the buffer again and again." :
The idea was to generate a LogBufferInfo using  directly the message as buffer (and not an external BUFSIZ limited buffer):
lbi2.top = log_data;
lbi2.current = log_data+msg_len;
lbl2.maxsize = msg_len;
lbi2.lock = NULL;
lbi2.refcount = 0;

BTW I think that a final \n must also be written in the file.

mreynolds389: > Hi agree about log rotation, I realized it just after having written the comment. 😉
> 
> Sorry but both log_append_buffer2 and log_append_audit_buffer hold the lbi->lock buffer when effectively writing in the file (i.e when calling log_flush_buffer ) The only thing that is done out of the lock is to copy the data in the reserved area within the buffer (but there is no buffer in the overlong message case) ==> you must hold lbi->lock buffer while writing directly on the file

You're right, I'll fix that

> 
> Faking LogBufferInfo is an overhead as log_flush_buffer will mostly calling LOG_WRITE_NOW_NO_ERR anyway (That is why I also removed that idea) But you are wrong about "the update could exceed the buffer again and again." : 

Well... it would if we reused the current structure, but what you are proposing is bypassing this fixed sized buffer.  Sorry I might have been too hasty in my response as I thought you wanted to reuse the existing LBI

> The idea was to generate a LogBufferInfo using directly the message as buffer (and not an external BUFSIZ limited buffer): lbi2.top = log_data; lbi2.current = log_data+msg_len; lbl2.maxsize = msg_len; lbi2.lock = NULL; lbi2.refcount = 0;

Ok let me try this out, but I'm not sure I understand the benefit to faking a LBI when we are just writing it directly to the file anyway.

> 
> BTW I think that a final \n must also be written in the file.

I had to remove that newline from PR_snprint() because it was adding two blank lines in the log (I promise).  The audit entry update from auditlog.c already adds the newline so adding a second newline in PR_snprint() was not necessary.

mreynolds389:
@progier389 ok all changes applied, please review...

progier389 (approved):
LGTM

PR #6185, opened 2024-05-30
Issue 6186 - Change default salt sizes generated in crypt_pwd

jasonborden:
Issue - #6186 - Increase the amount of salt crypt_pwd generates

Bug Description:
Salt currently generated by crypt_pwd is only 12 bits which is rather weak.

Fix Description:
Makes the salt generated the same length as linux shadow:
12bits (2 b64 chars) for CRYPT
48bits (8 b64 chars) for CRYPT-MD5
96bits (16 b64 chars) for CRYPT-SHA256 and CRYPT-SHA512

relates: https://lists.fedorahosted.org/archives/list/freeipa-users@lists.fedorahosted.org/thread/ZJXVFQ6XC2IEROA2LZNBXKQ6YWAJHAIU/ 

Author: Jason Borden

Reviewed by: @progier389, @merlinthp (Thanks!)

### DIFF
--- a/ldap/servers/plugins/pwdstorage/crypt_pwd.c
+++ b/ldap/servers/plugins/pwdstorage/crypt_pwd.c
@@ -39,11 +39,12 @@
 static unsigned char itoa64[] = /* 0 ... 63 => ascii - 64 */
     "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
 
-#define CRYPT_UNIX 0
-#define CRYPT_MD5 1
-#define CRYPT_SHA256 2
-#define CRYPT_SHA512 3
+/* Use the same salt lengths as shadow */
+#define CRYPT_UNIX_SALT_LENGTH 2
+#define CRYPT_MD5_SALT_LENGTH 8
+#define CRYPT_SHA_SALT_LENGTH 16
 
+#define CRYPT_SALT_STRING_MAXLEN CRYPT_SHA_SALT_LENGTH + 1
 
 int
 crypt_pw_cmp(const char *userpwd, const char *dbpwd)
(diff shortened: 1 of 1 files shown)


ldap/servers/plugins/pwdstorage/crypt_pwd.c
```diff
@@ -119,21 +109,21 @@ crypt_pw_enc_by_hash(const char *pwd, int hash_algo)
 char *
 crypt_pw_enc(const char *pwd)
 {
-    return crypt_pw_enc_by_hash(pwd, CRYPT_UNIX);
+    return crypt_pw_enc_by_hash(pwd, CRYPT_UNIX_SALT_LENGTH, NULL);
```
progier389: Globally the change looks good.
A minor remark: 
Instead of using NULL as algo_id and checking for it in crypt_pw_enc_by_hash, you could use the empty string "" so that the slapi_ch_smprintf handles directly all the cases

jasonborden: Oh yeah, good point. I can make that change from NULL to "", but my git skills are terrible. Can I still squash the new commit into the original for the PR?

mreynolds389: > Oh yeah, good point. I can make that change from NULL to "", but my git skills are terrible. Can I still squash the new commit into the original for the PR?

You have a few options, you can squash the commits (see https://www.port389.org/docs/389ds/contributing.html#getting-the-patch-ready)   Or you could make your change, and do "git commit -a --amend" and then do a "git push --force".  This will leave it as a single commit

jasonborden: Ok, that change has been made. Thanks for the help on the git commands!

progier389 (approved):
LGTM

Issue #6188, opened 2024-05-30
Minor cn=config nsslapd-haproxy-trusted-ip issues


droideck:
**Issue Description**
This ticket is for fixing minor issues with the nsslapd-haproxy-trusted-ip attribute.
Issues:

- It is not present in cn=config by default.
- It is not present in cn=schema.
- Deleting a single value from the attribute doesn't work and requires server restart to get rid of the attribute.

We need to investigate the issues and decide how the behaviour should be defined (as our cn=config processing is different from usual LDAP modify operations)

droideck:
Okay, for `It is not present in cn=schema.` - the PR is ready.

For `It is not present in cn=config by default.` - the way cn=config works, we don't display the attribute if it doesn't have a value. So `ldapsearch -xLLL -D "cn=Directory Manager" -w password -H ldap://localhost:389 -b cn=config nsslapd-haproxy-trusted-ip | grep haproxy` will naturally fail as expected (nothing is found).
But when the value is set, the ldapsearch will find the line.

And as for `Deleting a single value from the attribute doesn't work and requires server restart to get rid of the attribute.`, I think it actually happens because of the old issue - https://github.com/389ds/389-ds-base/issues/2078
And because we have DSE callbacks that still do the operations on the dse.ldif.
I think this issue should be fixed as part of the https://github.com/389ds/389-ds-base/issues/2078 ticket to avoid duplicates and cover all cases. It is a low priority I think, though.

droideck:
b6c2d789b..245ffc01a  389-ds-base-3.0 -> 389-ds-base-3.0
   fad9ff38b..6c7047ad7  389-ds-base-2.5 -> 389-ds-base-2.5
   0e19d4e80..2884c2766  389-ds-base-2.4 -> 389-ds-base-2.4
   3c7813341..01607109c  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6201, opened 2024-06-05
Issue 6188 - Add nsslapd-haproxy-trusted-ip to cn=schema

droideck:
Description: Add HAProxy trusted IP address multi-valued attribute to cn=schema in 01core389.ldif

Related: https://github.com/389ds/389-ds-base/issues/6188

Reviewed by: ?


progier389 (approved):
LGTM
FYI: At first glance I was a bit surprised to not see the new attribute added in any objectClasses but none of the other config attributes is in nsslapdConfig. They rely on the fact that cn=config is an extensibleObject.  So this change looks fine

PR #6204, opened 2024-06-06

### LINKED ISSUE #6181: Allow systemd to manage uid/gid changes rather than calling setuid.

vashirov:
Description:
Expand CapabilityBoundingSet to include addittional capabilites.

Fixes: https://github.com/389ds/389-ds-base/issues/6181

### DIFF
--- a/wrappers/systemd.template.service.in
+++ b/wrappers/systemd.template.service.in
@@ -25,7 +25,7 @@ ExecStart=@sbindir@/ns-slapd -D @instconfigdir@/slapd-%i -i /run/@package_name@/
 
 # Allow non-root instances to bind to low ports.
 AmbientCapabilities=CAP_NET_BIND_SERVICE
-CapabilityBoundingSet=CAP_NET_BIND_SERVICE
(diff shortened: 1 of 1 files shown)


vashirov:
There are still failures in CI, marking as draft for now.

progier389 (approved):
LGTM
At first I wondered why the DAC_OVERRIDE capability is needed then I realized that it is because ns-slapd must read the dse.ldif file before doing the setuid()

vashirov:
Added CAP_CHOWN, now it works in a container too.

PR #6205, opened 2024-06-06
Issue 6199 - wrong search query during certificate based authentication

### LINKED ISSUE #6199: LDAP unprotected search query during certificate based authentication
**Issue Description**

During certificate based client authentication, certmap may be configured to use the user cert's subjectDN to look for a match. The SubjectDN is however not escaped and may provides unexpected search result.

 - Platform: Fedora
 - Browser: N/A

**Steps to Reproduce**
See https://issues.redhat.com/browse/IDMDS-3959

progier389:
Problems:
   SubjectDN extracted from the certificate is not escaped when used by certmap.conf
   Other extracted value are wrongly escaped and quoted when added in filter
   
   Solution: Ensure that proper escape function is used in these two cases.
     Values in filter should not be quoted but * should be escaped.
     
   Note: I considered to reuse the ldap_bv2escaped_filter_value function but it needless realloc the returned data
     so I ended up to rewrite something the escape function (which is quite straightforward anyway).

Issue: #6199 

Reviewed by: @droideck

### DIFF
--- a/dirsrvtests/tests/suites/sasl/regression_test.py
+++ b/dirsrvtests/tests/suites/sasl/regression_test.py
@@ -14,12 +14,18 @@ from lib389.utils import *
 from lib389.topologies import topology_m2
 from lib389._constants import *
 from lib389.replica import ReplicationManager
+from lib389.config import CertmapLegacy, Config
+from lib389.idm.user import UserAccounts
+from lib389.nss_ssl import NssSsl, USER_ISSUER
 
 pytestmark = [pytest.mark.tier1,
               pytest.mark.skipif(ds_is_older('1.3.5'), reason="Not implemented")]
 
+DEBUGGING = os.getenv("DEBUGGING", default=False)
 logging.getLogger(__name__).setLevel(logging.DEBUG)
 log = logging.getLogger(__name__)
+
+test_user_uid = 1000
 
 ISSUER = 'cn=CAcert'
 CACERT = 'CAcertificate'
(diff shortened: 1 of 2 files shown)


lib/ldaputil/certmap.c
```diff
@@ -757,23 +829,29 @@ ldapu_cert_searchfn_default(void *cert, LDAP *ld, void *certmap_info_in, const c
...
             return rv;
         }
-        len = strlen(certmap_info->searchAttr) + strlen(subjectDN) +
-              strlen("=") + 1;
-        certFilter = (char *)ldapu_malloc(len * sizeof(char));
+        size_t subjectDN_len = strlen(subjectDN);
+        size_t eqout_len = 3 * subjectDN_len + 1; /* big enough for worst case (and final \0) */
+        size_t searchAttr_len = strlen(certmap_info->searchAttr);
+        size_t len = searchAttr_len + 1 + eqout_len;
+        char *certFilter = (char *)ldapu_malloc(len * sizeof(char));
```
droideck, line 841: It's not about your code, but still, we can fix it here;- shouldn't the check be `if (!(*certFilter)) {`?
As we need to check if the value at the address `certFilter` points to is NULL (i.e., if the allocation failed).

lib/ldaputil/certmap.c
```diff
@@ -757,23 +829,29 @@ ldapu_cert_searchfn_default(void *cert, LDAP *ld, void *certmap_info_in, const c
...
             free(subjectDN);
             return LDAPU_ERR_OUT_OF_MEMORY;
         }
-        sprintf(certFilter, "%s=%s", certmap_info->searchAttr, subjectDN);
+        strcpy(certFilter, certmap_info->searchAttr);
+        certFilter[searchAttr_len] = '=';
+        rv = value2filter(certFilter+searchAttr_len+1, eqout_len, subjectDN, subjectDN_len);
         free(subjectDN);
+        len = strlen(certFilter);
+        if (rv != LDAPU_SUCCESS || len <= 0) {
```
droideck, line 851: I think we need to free `certFilter` if we fail here.

lib/ldaputil/certmap.c
```diff
@@ -964,14 +1042,22 @@ AddAVAToBuf(char *buf, int size, int *len, const char *tagName, CERTAVA *ava)
...
 
-    rv = CERT_RFC1485_EscapeAndQuote(buf + taglen,
-                                     size - taglen,
-                                     (char *)ava->value.data + lenLen,
-                                     ava->value.len - lenLen);
-
-    *len += strlen(buf);
-
-    return (rv == SECSuccess ? LDAPU_SUCCESS : LDAPU_FAILED);
+    if (is_filter != 0) {
```
droideck, line 1046: Not a nitpick but curiosity. :)
Why `if (is_filter != 0) {` instead of `if (is_filter) {`?

dirsrvtests/tests/suites/sasl/regression_test.py
```diff
@@ -172,6 +178,174 @@ def test_openldap_no_nss_crypto(topology_m2):
...
+        raise e
+
+
+def test_bind_certifcate_with_unescaped_subject(topology_m2, request):
+    """Test close connection on failed bind with a failed cert mapping
+
+    :id: 7545ad30-21dc-11ef-a27e-482ae39447e5
+    :setup: Replication with two suppliers
+            (Only one instance is used but it is better to reuse the
+             same topology than the other tests in this module)
```
droideck: Validate CI failed here:

```

test_bind_certifcate_with_unescaped_subject:243
-----------------------------------------------

* Docstring has RST parsing issues. RST parser messages:

  * Unexpected indentation.

      :setup: Replication with two suppliers
              (Only one instance is used but it is better to reuse the
    >          same topology than the other tests in this module)
      :steps:
```

droideck (changes requested):

lib/ldaputil/certmap.c
progier389: HIMO you should look again. The above line declare certFilter as a pointer on char and initialize it
BTW *certFilter is an uninitialized random value as malloc does not initialize the buffer data

lib/ldaputil/certmap.c
progier389: It is explicitly mentioned in the "Secure coding practice" 

https://cyber.gouv.fr/sites/default/files/2022/04/anssi-guide-rules_for_secure_c_language_software_development-v1.4.pdf
10.6 Implicit comparison with 0 prohibited

progier389:
Fixed the extra space that cause "Validate tests" to fail 
Fixed the missing free(certFilter)

lib/ldaputil/certmap.c
droideck: Okay, I think I got a bit confused here... All good!

droideck (approved):
I think it'll be better to wait for CI to be fixed so we can check the PR against tests, but the code looks good to me!

progier389:
I agree about waiting for the CI tests

tbordaz:
Sorry to come late to this thread. Just a question regarding value2filter_sizes.
should not we also escape chars: ',' (comma), '=' (equal), '&' (and), '|' (or), '!'

progier389:
According to https://datatracker.ietf.org/doc/html/rfc4515:
```
The <valueencoding> rule ensures that the entire filter string is a
   valid UTF-8 string and provides that the octets that represent the
   ASCII characters "*" (ASCII 0x2a), "(" (ASCII 0x28), ")" (ASCII
   0x29), "\" (ASCII 0x5c), and NUL (ASCII 0x00) are represented as a
   backslash "\" (ASCII 0x5c) followed by the two hexadecimal digits
   representing the value of the encoded octet.
```
   So the answer is no.

PR #6208, opened 2024-06-08
Issue 6183 - Slow ldif2db import on a newly created BDB backend

### LINKED ISSUE #6183: Slow ldif2db import on a newly created BDB backend
**Issue Description**
After creating a new BDB backend, we autotune the cache only when restarting. So, an administrator will try to import an LDIF before that; she will have a very slow import. The workaround is very simple - restart after the new backend is created.
So we either need to:

- modify the CLI (or the core backend creation code)

OR at least

- modify our CI tests so they restart the instance after the backend is created but before the restore/ldif2db. 

**Steps to Reproduce Slow Import**
Steps to reproduce the behaviour:
1. Install an instance
2. Create a new backend
3. Stop the instance
4. Run ldif2db on the new backend - It's very slow.
5. Start the instance

**Steps to Reproduce Fast Import**
Steps to reproduce the behaviour:
1. Install an instance
2. Create a new backend
3. Restart the instance (Autotuning is performed)
4. Stop the instance
5. Run ldif2db on the new backend - It's fast enough.
6. Start the instance

droideck:
Bug Description: After creating a new BDB backend, we autotune the cache only when restarting. So, an administrator will try to import an LDIF before that; she will have a very slow import.

Fix Description: Do the autotuning during the backend creation. Add a CI test for the scenario.

Fixes: https://github.com/389ds/389-ds-base/issues/6183

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/import/regression_test.py
+++ b/dirsrvtests/tests/suites/import/regression_test.py
@@ -434,7 +434,59 @@ def test_large_ldif2db_ancestorid_index_creation(topo, _set_mdb_map_size):
         # The time for the ancestorid index creation should be less than 10s for an offline import of an ldif file with 100000 entries / 5 entries per node
         # Should be adjusted if these numbers are modified in the test
         assert etime <= 10
-    
+
+
+def create_backend_and_import(instance, ldif_file, suffix, backend):
+    log.info(f'Add suffix:{suffix} and backend: {backend}...')
+    backends = Backends(instance)
+    backends.create(properties={'nsslapd-suffix': suffix, 'name': backend})
+    props = {'numUsers': 10000, 'nodeLimit': 5, 'suffix': suffix}
+
+    log.info(f'Create a large nested ldif file using dbgen : {ldif_file}')
+    dbgen_nested_ldif(instance, ldif_file, props)
+
+    log.info('Stop the server and run offline import...')
+    instance.stop()
+    log.info('Measure the import time for the ldif file...')
+    start = time.time()
+    assert instance.ldif2db(backend, None, None, None, ldif_file)
+    end = time.time()
+    instance.start()
+    return end - start
+
+
+@pytest.mark.xfail(not _check_disk_space(), reason="not enough disk space for lmdb map")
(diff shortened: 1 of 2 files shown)


progier389 (approved):
Looks good, but IMHO now that the CI tests are working again, you should better rebase and double check the CI test result.

dirsrvtests/tests/suites/import/regression_test.py
```diff
@@ -434,7 +434,59 @@ def test_large_ldif2db_ancestorid_index_creation(topo, _set_mdb_map_size):
...
+    instance.stop()
+    log.info('Measure the import time for the ldif file...')
+    start = time.time()
+    assert instance.ldif2db(backend, None, None, None, ldif_file)
+    end = time.time()
+    instance.start()
+    return end - start
+
+
+@pytest.mark.xfail(not _check_disk_space(), reason="not enough disk space for lmdb map")
```
tbordaz: Is this condition necessary. IIUC the problem only exist with bdb.

tbordaz (approved):
LGTM

tbordaz:
@droideck , the patch looks good to me as well but something to clarify: the tuning of the cache is missing, what specific cache (db cache, entry cache, import cache,..) ?

If it is db cache, something that is not clear to me is that when the instance is created, it contains a default backend and I would expect db cache to be set.

dirsrvtests/tests/suites/import/regression_test.py
droideck: Good catch!

droideck:
> @droideck , the patch looks good to me as well but something to clarify: the tuning of the cache is missing, what specific cache (db cache, entry cache, import cache,..) ?

I have suspicions it's the DN cache, but I am not exactly sure.
After I found that the issue was with autotuning, I focused on the solution. And as it's the correct thing to do — perform an autotuning after an instance creation (if autotuning is enabled), I just left with that as I had more stuff to do.

I can dive deeper into that if you want, though. It'll be interesting to dive into. :)

> 
> If it is db cache, something that is not clear to me is that when the instance is created, it contains a default backend and I would expect db cache to be set.

PR #6209, opened 2024-06-08
Issue 6200 - Disable WebUI CI tests

### LINKED ISSUE #6200: Should disable WebUI CI tests
WebUI CI test is systematically failing  for more than a year now. - and it as always been a bit flaky
And we never were able to find the bandwidth to fix it.
The Idea is to disable it because it is one of the last failing test. 

My feeling is to add a RUN_WEBUI_TEST environment variable and skip the test if it is not set 
 (So the test will still be visible and we may re-enable it if we fix it)

Remove "webui" from generate_matrix.py  (like snmp was removed) but IMHO it has two drawbacks:
 - Remove the test visibility 
- Test will still be run by Q/E 

**Additional context**
We need to have a reliable 100% PASS test suite

vashirov:
Description:
Currently WebUI tests fail. There are known issues both in tests and the code. We should re-enable the tests back when we get to fix those issues.

Fixes: https://github.com/389ds/389-ds-base/issues/6200

### DIFF
--- a/.github/scripts/generate_matrix.py
+++ b/.github/scripts/generate_matrix.py
@@ -24,6 +24,9 @@ else:
     # Filter out snmp as it is an empty directory:
     suites.remove('snmp')
 
+    # Filter out webui because of broken tests
+    suites.remove('webui')
+
     # Run each replication test module separately to speed things up
     suites.remove('replication')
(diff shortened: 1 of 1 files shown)


droideck (approved):

progier389:
Funny! I thought that you rather proposed to remove the WEBUI=1 variable from the tests  workflows.
But anyway it is fine for me ...

PR #6211, opened 2024-06-10
Issue 6120 - /usr/lib64/dirsrv/plugins/libback-bdb.so has an invalid-…

### LINKED ISSUE #6210: /usr/lib64/dirsrv/plugins/libback-bdb.so has an invalid-looking DT_RPATH: /usr/lib/dirsrv
**Issue Description**
rpminspect reports an invalid DT_RPATH `/usr/lib/dirsrv`.

It comes from https://github.com/389ds/389-ds-base/blob/216ffc0705f0de04d4a67c05b54acf4645721d25/m4/bundle_libdb.m4#L11

Changing it to `lib64` might break prefixed installs, where `libdir` defaults to `$prefix/lib` instead of `$prefix/lib64` on 64-bit systems, so a `--libdir` parameter needs to be passed for these installs in addition to `--prefix`.

vashirov:
…looking DT_RPATH: /usr/lib/dirsrv

Bug Description:
rpminspect reports an invalid DT_RPATH /usr/lib/dirsrv It's evaluated in m4/bundle_libdb.m4 from

```
-R${prefix}/lib/dirsrv"
```

Fix Description:
Change it to lib64

Fixes: https://github.com/389ds/389-ds-base/issues/6210

### DIFF
--- a/m4/bundle_libdb.m4
+++ b/m4/bundle_libdb.m4
@@ -8,7 +8,7 @@
 
 AC_MSG_CHECKING(Handling bundle_libdb)
 
-db_lib="-L${with_bundle_libdb}/.libs -R${prefix}/lib/dirsrv"
+db_lib="-L${with_bundle_libdb}/.libs -R${prefix}/lib64/dirsrv"
 db_incdir=$with_bundle_libdb
 db_inc="-I $db_incdir"
 db_libver="5.3-389ds"


m4/bundle_libdb.m4
```diff
@@ -8,7 +8,7 @@
 
 AC_MSG_CHECKING(Handling bundle_libdb)
 
-db_lib="-L${with_bundle_libdb}/.libs -R${prefix}/lib/dirsrv"
+db_lib="-L${with_bundle_libdb}/.libs -R${prefix}/lib64/dirsrv"
```
progier389, line 11: I wonder if we really need -R option at all. As it is the only place where it is used.

vashirov: Without -R builds fails.

progier389 (approved):
LGTM

Issue #6222, opened 2024-06-12
CI test acl/ test_timeofday_keyword sometime fails


progier389:
**Issue Description**
The test may fails because of timing issue 
>       org.replace("seeAlso", "cn=1")
E         ldap.INSUFFICIENT_ACCESS: {'msgtype': 103, 'msgid': 3, 'result': 50, 'desc': 'Insufficient access', 'ctrls': [], 'info': "Insufficient 'write' privilege to the 'seeAlso' attribute of entry 'ou=timeofday,ou=keywords,dc=example,dc=com'.\n", 'ldap_request': "modify_ext_s(('ou=Timeofday,ou=Keywords,dc=example,dc=com', [(2, 'seeAlso', [b'cn=1'])]),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance standalone1"}

IMHO this test should be done in a loop and the time should be checked after testing the access to verify it is still in the same minute.

PR #6223, opened 2024-06-12
Issue 6222 - CI test acl/test_timeofday_keyword sometime fails

progier389:
CI test acl/test_timeofday_keyword sometime fails because current time (in minutes) changes during the test
Solution is to run the test in loop and retry if the time has changed.
Also fix:
Similar issue with test_dayofweek_keyword_today_can_access (with time in days)
Skip the tests that sets the hostname if run as non root

Issue: #6222 


progier389:
Fixed 'validate' and 'testimony print' output

droideck (approved):
Nice solution! Ack :)

Issue #6227, opened 2024-06-17
dsconf schema does not show inChain matching rule


tbordaz:
**Issue Description**
The command 'dsconf instance schema matchingrules list' does not show inChain matching rule

**Package Version and Platform:**
 - since 2.4

**Steps to Reproduce**
1. install an instance
2. run 'dsconf standalone1 schema matchingrules list | grep -i inchain'

**Expected results**
The inchain matching rule should be displayed

**Additional context**

tbordaz:
f76503dee..1f41661c7  main
ae8ab2141..0fb74967f  389-ds-base-3.0
0f35be137..cf6cdd05b  389-ds-base-2.5
43057ebc3..cc723ef46  389-ds-base-2.4

PR #6228, opened 2024-06-17
Issue 6227 - dsconf schema does not show inChain matching rule

tbordaz:
Bug description:
	The registered inChain MR does defined any matching rule
	syntax (mr_syntax).
	When dsconf reads the matching rules (read_schema_dse)
	it only reports those which have OID and SYNTAX.
	As a consequence InChain was not reported.

Fix description:
	The syntax defines that assersion syntax that is
	distinguished name. Add this syntax to the register
	struct

relates: #6227

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/filter/inchain_test.py
+++ b/dirsrvtests/tests/suites/filter/inchain_test.py
@@ -15,6 +15,7 @@ from lib389._constants import DEFAULT_SUFFIX, PW_DM, PLUGIN_MEMBER_OF
 from lib389.topologies import topology_st as topo
 from lib389.plugins import MemberOfPlugin
 
+from lib389.schema import Schema
 from lib389.idm.user import UserAccount, UserAccounts
 from lib389.idm.account import Accounts
 from lib389.idm.account import Anonymous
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/filter/inchain_test.py
```diff
@@ -812,6 +812,27 @@ def test_invalid_assertion(topo):
...
+        2. Checks that the output contains inchaineMatch matching rule
+    :expectedresults:
+        1. Success
+        2. Success
+    """
+    dsconf_cmd= ['/usr/sbin/dsconf', topo.standalone.serverid, 'schema', 'matchingrules', 'list']
+    # Set referral
+    log.info("Use dsconf to check defined matching rules")
+    log.info(f'Command used: %{dsconf_cmd}')
+    proc = subprocess.Popen(dsconf_cmd, stdout=subprocess.PIPE)
```
progier389: Hi Thierry, 
IMHO you should avoid to use subprocess as far as possible. 
In this case you can simply check directly the schema:

schema = Schema(inst)
#mrs = [ str(mr) for mr in schema.get_matchingrules() ]
mrs = [ f"{mr.oid} {mr.names[0]}" for mr in schema.get_matchingrules() if len(mr.names) > 0 ]
print(mrs)
assert  '1.2.840.113556.1.4.1941 inchainMatch' in mrs

progier389 (approved):
LGTM

droideck (approved):

Issue #6238, opened 2024-06-27
Provide option to change audit log format to use JSON


mreynolds389:
## Description

Provide option to write audit log in a json format.

See:  https://www.port389.org/docs/389ds/design/audit-json-logging-design.html

mreynolds389:
PR was merged to Main branch.  It should probably be backported to other branches, but I'll wait for that decision from the lead/PO.

mreynolds389:
a548fa4ab..8bcd1b97b  389-ds-base-3.0 -> 389-ds-base-3.0
2b68afcc0..08e704016  389-ds-base-2.5 -> 389-ds-base-2.5

PR #6239, opened 2024-06-27
Issue 6238 - RFE - add option to write audit log in JSON format

mreynolds389:
Description:

Add option to set the format between: default, json, or json-pretty

json-pretty just writes the JSON format in a vertical structure verses one condensed line of text.

You can also adjust the date format using strftime formatting

Relates: https://github.com/389ds/389-ds-base/issues/6238

### DIFF
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -1,6 +1,6 @@
 /** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
- * Copyright (C) 2005 Red Hat, Inc.
+ * Copyright (C) 2005-2024 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -10,7 +10,6 @@
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
-
 
 #include "slap.h"
 
@@ -23,6 +22,7 @@
 #define ATTR_DELETEOLDRDN "deleteoldrdn"
 #define ATTR_NEWSUPERIOR "newsuperior"
 #define ATTR_MODIFIERSNAME "modifiersname"
+#define JBUFSIZE 75
 char *attr_changetype = ATTR_CHANGETYPE;
 char *attr_newrdn = ATTR_NEWRDN;
 char *attr_deleteoldrdn = ATTR_DELETEOLDRDN;
@@ -33,7 +33,9 @@ static int audit_hide_unhashed_pw = 1;
 static int auditfail_hide_unhashed_pw = 1;
 
 /* Forward Declarations */
-static void write_audit_file(Slapi_Entry *entry, int logtype, int optype, const char *dn, void *change, int flag, time_t curtime, int rc, int sourcelog);
+static void write_audit_file(Slapi_PBlock *pb, Slapi_Entry *entry, int logtype,
+                             int optype, const char *dn, void *change,
+                             int flag, time_t curtime, int rc, int sourcelog);
 
 static const char *modrdn_changes[4];
 
@@ -99,7 +101,8 @@ write_audit_log_entry(Slapi_PBlock *pb)
     curtime = slapi_current_utc_time();
     /* log the raw, unnormalized DN */
     dn = slapi_sdn_get_udn(sdn);
-    write_audit_file(entry, SLAPD_AUDIT_LOG, operation_get_type(op), dn, change, flag, curtime, LDAP_SUCCESS, SLAPD_AUDIT_LOG);
+    write_audit_file(pb, entry, SLAPD_AUDIT_LOG, operation_get_type(op), dn,
+                     change, flag, curtime, LDAP_SUCCESS, SLAPD_AUDIT_LOG);
 }
 
 void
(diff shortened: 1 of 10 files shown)


ldap/servers/slapd/auditlog.c
```diff
@@ -194,7 +199,7 @@ log_entry_attr(Slapi_Attr *entry_attr, char *attrname, lenstr *l)
 {
     Slapi_Value **vals = attr_get_present_values(entry_attr);
     for(size_t i = 0; vals && vals[i]; i++) {
-        char log_val[256] = "";
+        char log_val[256] = {0};
```
progier389: Why changing that ?
Setting all 255 bytes to 0 instead of only setting the first one, seems like an useless overhead (especially since it is done for each values on each attribute.)

ldap/servers/slapd/auditlog.c
```diff
@@ -211,6 +216,29 @@ log_entry_attr(Slapi_Attr *entry_attr, char *attrname, lenstr *l)
     }
 }
 
+static void
+log_entry_attr_json(Slapi_Attr *entry_attr, char *attrname, json_object *attr_list)
+{
+    Slapi_Value **vals = attr_get_present_values(entry_attr);
+    for(size_t i = 0; vals && vals[i]; i++) {
+        json_object *attr_obj = json_object_new_object();
+        char log_val[256] = {0};
```
progier389: same remark, should probably better be char log_val[256] = "";

ldap/servers/slapd/auditlog.c
```diff
@@ -266,20 +301,285 @@ add_entry_attrs(Slapi_Entry *entry, lenstr *l)
...
+                target_dn = "unknown";
+            }
+        } else if (optype == SLAPI_OPERATION_DELETE) {
+            slapi_pblock_get(pb, SLAPI_DELETE_TARGET_SDN, &target_dn);
+            if (target_sdn) {
+                target_dn = slapi_sdn_get_ndn(target_sdn);
+            } else {
+                target_dn = "unknown";
+            }
+        } else if (optype == SLAPI_OPERATION_DELETE) {
```
progier389: Duplicate case:  if (optype == SLAPI_OPERATION_DELETE) {
should probably be a SLAPI_OPERATION_MODRDN)

ldap/servers/slapd/auditlog.c
```diff
@@ -266,20 +301,285 @@ add_entry_attrs(Slapi_Entry *entry, lenstr *l)
...
+    char *timestr = NULL;
+    char *tmp = NULL;
+    char *tmpsave = NULL;
+    uint64_t conn_id = 0;
+    int32_t rc = -1;
+    int32_t op_id = 0;
+
+    time_format = config_get_auditlog_time_format();
+    (void)localtime_r(&curtime, &tms);
+    if (strftime(local_time, JBUFSIZE, time_format, &tms) == 0) {
```
progier389, line 378: Should we really convert the timestamp in the json version ?
The local time string format is better for the human, but json logs are intended for some post-processing tools 
and IMHO the raw UTC time as integer is usually better for tools (Because each tool may then easily convert it the way it need it.) 
For example the replication latency tool has to convert back the access log timestamp to integer because the different logs may be in different timezone)
And storing directly the integer is faster.

vashirov: Please check this discussion https://github.com/389ds/389ds.github.io/pull/14#discussion_r1575708169
I asked for ISO 8601 Date string because it's unambiguous, human-readable, and well-supported.

progier389 (changes requested):
The "duplicate if" must be changed,
The other remarks are more minor and discussion is open !

ldap/servers/slapd/auditlog.c
progier389: I agree with ISO 8601 but I do not think we should allow to change the format
 and in any case it should be gmtime and not localtime. 
IMHO allowing to change the format may not be a good idea. 
 Not only it add one more parameter, but there is a big difference between the two formats:
- The ascii output is intended to be directly readable by humans.
- The json output must be post-processed by a tools
And if the timestamp format may change, the tool developper will have to take care of the timestamp format 
If it is only ISO 860, he has only to take care of ISO 8601 format.
So IMHO we should have a single ISO 8601 format and let the tool format the date
Sorry that I have not realized that point in April !

mreynolds389: @vashirov @progier389 The reason I made the time format configurable is because customers/community have been asking for this for years (for DS and IDM).  I see no harm in having it, but if you guys think it's a bad idea then I will remove it (no problem).

ldap/servers/slapd/auditlog.c
mreynolds389: > Duplicate case: if (optype == SLAPI_OPERATION_DELETE) { should probably be a SLAPI_OPERATION_MODRDN)

Nice catch!

ldap/servers/slapd/auditlog.c
mreynolds389: > Why changing that ? Setting all 255 bytes to 0 instead of only setting the first one, seems like an useless overhead (especially since it is done for each values on each attribute.)

I thought it was cleaner, but you have a good point.  I'll change it

ldap/servers/slapd/auditlog.c
mreynolds389: > I agree with ISO 8601 but I do not think we should allow to change the format and in any case it should be gmtime and not localtime. 

FYI, the "time" key, which is also the local time.  This is how the current audit log records the time.  So I kept that for legacy purposes and then introduced the "date" key for the custom time.  Again, this is an open discussion but I just wanted clarify what the current PR is doing.

progier389: You are right: I overlook the **time** key.
 So forgot what i previously wrote: I do not really care about the way the **date** value is formatted ( local time and a configurable format is fine)
 But for "time" you should use ISO 8601 as viktor mentionned. 
  (i.e: using  strftime(isotime, (sizeof isotime), "%FT%TZ", gmtime(&curtime));

rather than format_localTime

progier389: Oups! should rather use gmtime_r instead of gm_time in multi threaded environment ! :wink:

mreynolds389:
Ok all the changes have been made.  I changed the names of the time keys from "date" -> "local_time", and "time" -> "gm_time".  Not sure if those are the best names to use, but it's clearer than before.

progier389 (approved):
LGTM

PR #6242, opened 2024-06-28
Issue 6241 - Add support for CRYPT-YESCRYPT

### LINKED ISSUE #6241: Add support for CRYPT-YESCRYPT
Yescrypt has supplanted CRYPT-SHA512 as the default hashing algorithm in many distros including Arch, Debian, and Fedora. It would be nice to also have support for generating passwords using this hashing algorithm in 389-ds-base

Implement CRYPT-YESCRYPT as a password storage scheme.

jasonborden:
Description:
Implements CRYPT-YESCRYPT as a password storage scheme

Issue: #6241 

Reviewed by: @progier389

### DIFF
--- a/ldap/ldif/template-dse.ldif.in
+++ b/ldap/ldif/template-dse.ldif.in
@@ -195,6 +195,15 @@ nsslapd-plugininitfunc: crypt_sha512_pwd_storage_scheme_init
 nsslapd-plugintype: pwdstoragescheme
 nsslapd-pluginenabled: on
 
+dn: cn=CRYPT-YESCRYPT,cn=Password Storage Schemes,cn=plugins,cn=config
+objectclass: top
+objectclass: nsSlapdPlugin
+cn: CRYPT-YESCRYPT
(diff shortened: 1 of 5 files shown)


jasonborden:
This PR uses the same defaults parameters as the distros that use yescrypt: 24 b64 salt chars and j9T params. To use yescrypt the system will need libxcrypt 4.2 or higher.

progier389 (commented):
Looks good, 
but  I would prefer that we use #define for the standard crypt algorithm id rather than embedding directly the string in the code. 

Maybe a bit picky but this code is sensitive in term of security, 
and it is better make it as easily readable as possible ...

/* Standard Crypt Algorithms identifiers (with usual options when needed) */
#define CRYPT_ALGO_ID_SHA512  "$6$"
#define CRYPT_ALGO_ID_YESCRYPT "$y$j9T$"

jasonborden:
Updated with suggested #defined algorithm IDs and comment.

progier389 (approved):
Looks good.

Issue #6216, opened 2024-06-10
CI test import/test_fast_slow_import sometime fail


progier389:
**Issue Description**

CI test_fast_slow_import sometime fails with:
>       assert total_time1 < total_time2
E       assert 2.2014029026031494 < 2.1553120613098145

We should at least mark it as flaky 
note: The db size (1K users) is now far too small to provide reliable result but I do not think it is worth to try fix the test
  (disable it is enough) as using private memory in import is an obscure feature specific to obsolete bdb database.

PR #6247, opened 2024-07-01
Issue 6216 - CI test_fast_slow_import sometime fail

progier389:
The test comparing times around 2 seconds is pretty instable.
With this fix, the test is still running, checking that import with private memory works
but does not check any more that it is faster than standard import

Issue: #6216 

Reviewed by: @droideck (Thanks!)


droideck (approved):
Looks good!

PR #6257, opened 2024-07-09
Issue 6256 - nsslapd-numlisteners limit is not enforced

### LINKED ISSUE #6256: nsslapd-numlisteners limit is not enforced
**Issue Description**
When a invalid value for the attribute nsslapd-numlisteners is used, config normalises the value but the invalid value is written to dse.ldif.

 - Package and version: 389-ds-base-2.3.4-3.el9.x86_64

**Steps to Reproduce**
1. dsconf localhost config get nsslapd-numlisteners
nsslapd-numlisteners: 1
2. dsconf localhost config replace nsslapd-numlisteners=5
Successfully replaced "nsslapd-numlisteners"
3. dsconf localhost config get nsslapd-numlisteners
nsslapd-numlisteners: 4
4. grep nsslapd-numlisteners /etc/dirsrv/slapd-localhost/dse.ldif
nsslapd-numlisteners: 5
5. systemctl restart dirsrv@localhost
6. dsconf localhost config get nsslapd-numlisteners
nsslapd-numlisteners: 4
7. grep nsslapd-numlisteners /etc/dirsrv/slapd-localhost/dse.ldif
nsslapd-numlisteners: 5

**Expected results**
Invalid values should return an operations error and not be written to dse.ldif

jchapma:
Description: When a invalid value for the attribute nsslapd-numlisteners is used, config normalises the value but the invalid value is written to dse.ldif.

Fix description: Modify config to return operations error if an invalid value is used. Modify configdse to handle an operations error if an invalid value is used.

Fixes: https://github.com/389ds/389-ds-base/issues/6256

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/configdse.c
+++ b/ldap/servers/slapd/configdse.c
@@ -287,8 +287,7 @@ load_config_dse(Slapi_PBlock *pb __attribute__((unused)),
         if (attr_name) {
             retval = config_set(attr_name, values, returntext, 1 /* force apply */);
             if ((strcasecmp(attr_name, CONFIG_MAXDESCRIPTORS_ATTRIBUTE) == 0) ||
-                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0) ||
-                (strcasecmp(attr_name, CONFIG_NUM_LISTENERS_ATTRIBUTE) == 0)) {
+                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0)) {
                 /* We should not treat an LDAP_UNWILLING_TO_PERFORM as fatal for
                  * the these config attributes.  This error is returned when
                  * the value we are trying to set is higher than the current
--- a/ldap/servers/slapd/configdse.c
+++ b/ldap/servers/slapd/configdse.c
@@ -287,7 +287,8 @@ load_config_dse(Slapi_PBlock *pb __attribute__((unused)),
         if (attr_name) {
             retval = config_set(attr_name, values, returntext, 1 /* force apply */);
             if ((strcasecmp(attr_name, CONFIG_MAXDESCRIPTORS_ATTRIBUTE) == 0) ||
-                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0)) {
+                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0) ||
+                (strcasecmp(attr_name, CONFIG_NUM_LISTENERS_ATTRIBUTE) == 0)) {
                 /* We should not treat an LDAP_UNWILLING_TO_PERFORM as fatal for
                  * the these config attributes.  This error is returned when
                  * the value we are trying to set is higher than the current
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/configdse.c
```diff
@@ -287,7 +287,8 @@ load_config_dse(Slapi_PBlock *pb __attribute__((unused)),
         if (attr_name) {
             retval = config_set(attr_name, values, returntext, 1 /* force apply */);
             if ((strcasecmp(attr_name, CONFIG_MAXDESCRIPTORS_ATTRIBUTE) == 0) ||
-                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0)) {
+                (strcasecmp(attr_name, CONFIG_RESERVEDESCRIPTORS_ATTRIBUTE) == 0) ||
+                (strcasecmp(attr_name, CONFIG_NUM_LISTENERS_ATTRIBUTE) == 0)) {
```
droideck: I wonder why to put the attribute to this check if we don't benefit from that?

We should either remove it from here or we should add this (or something like this with LDAP_UNWILLING_TO_PERFORM) to `ldap/servers/slapd/libglobs.c::config_set_num_listeners`:

```C
        if (nValue > maxVal) {
            nValue = maxVal;
            retVal = LDAP_UNWILLING_TO_PERFORM;
        } else {
            retVal = LDAP_OPERATIONS_ERROR;
        }
```

Or do I miss something?

droideck (changes requested):

ldap/servers/slapd/configdse.c
jchapma: The check is here to cover the corner case where a user has changed version, if we don't add the check and the "nsslapd-numlisteners" attr is present in dse.ldif the instance will fail to start. 

IIUC we need to return LDAP_OPERATIONS_ERROR when an invalid value has been used for this attr, otherwise the invalid value gets written to dse.ldif.

jchapma: I had  a look over this again and yes, it is best to return LDAP_UNWILLING_TO_PERFORM when an invalid value for the nsslapd-numlisteners is used.

droideck:
Sorry, I was waiting for GitHub CI to be working again, but maybe we had some fix recently? Could you please rebase onto `main`?

ldap/servers/slapd/configdse.c
droideck: Okay, IIUC, you changed the behaviour now to reject the value instead of normalizing it (as it was before).
Could you please elaborate on why you think it's the current thing to do for `nsslapd-numlisteners`?

You probably need to change the commit message and test the docstring with the explanation.

jchapma: When the dse backend callback is called there are two passes done for each mod, the first pass validates the value and the second pass apply's the change. With the existing behaviour, when an invalid value is input the config code normalises the value during the first pass, then during the second pass the invalid value is written to dse.ldif, which is why this issue was created.

To avoid this we need to reject an invalid value so the second pass doesn't apply the change. To reject the invalid value we need to return either LDAP_OPERATIONS_ERROR like almost all of the other config_set functions or LDAP_UNWILLING_TO_PERFORM.

Yes the plan was to update the commit message during merge, I need to update the CI test so I will update the doc string also.

jchapma:
> Sorry, I was waiting for GitHub CI to be working again, but maybe we had some fix recently? Could you please rebase onto `main`?
> 
> It will also give us the test you and @ssidhaye added.

Rebased with main branch, the CI test I added wasn't used, I just added the with statement to @ssidhaye test.

droideck (approved):
LGTM!

PR #6260, opened 2024-07-12
Issue 4778 - Add COMPACT_CL5 task to dsconf replication

droideck:
Description: In 1.4.3, the changelog is not part of a backend. It can be compacted with nsTask: CAMPACT_CL5 as part of the replication entry.
Add the task as a compact-changelog command under the dsconf replication tool. Add tests for the feature and fix old tests.

Related: https://github.com/389ds/389-ds-base/issues/4778

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -1199,6 +1199,11 @@ def restore_cl_dir(inst, basedn, log, args):
     replicas.restore_changelog(replica_roots=args.REPLICA_ROOTS, log=log)
 
 
+def compact_cl5(inst, basedn, log, args):
+    replicas = Replicas(inst)
+    replicas.compact_changelog(replica_roots=args.REPLICA_ROOTS, log=log)
+
+
 def create_parser(subparsers):
 
     ############################################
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/cli_conf/replication.py
```diff
@@ -1326,6 +1331,11 @@ def create_parser(subparsers):
                               help="Specify one replica root whose changelog you want to restore. "
                                    "The replica root will be consumed from the LDIF file name if the option is omitted.")
 
+    compact_cl = repl_subcommands.add_parser('compact-changelog', help='Compact the changelog database')
+    compact_cl.set_defaults(func=compact_cl5)
+    compact_cl.add_argument('REPLICA_ROOTS', nargs="+",
```
progier389, line 1336: REPLICA_ROOTS terminology is a bit unusual,  
Why not using the usual "suffix" option instead ?

droideck: I just used the same terminology as in `'from-changelogdir'` and `'-r', '--replica-roots', nargs="+",` in `'dump-changelog'`, etc.

And as it's a fix for 1.4.3, I'd rather not change the terminology and keep it consistent. :)

progier389:
More serious issue:
The "Compile" action fails .
 IMHO the issue is that you are trying to merge directly on 389ds:389-ds-base-1.4.3 rather than on main branch

droideck:
> More serious issue: The "Compile" action fails . IMHO the issue is that you are trying to merge directly on 389ds:389-ds-base-1.4.3 rather than on main branch

The fix is exactly for 1.4.3 as COMPACT_CL5  is only in 1.4.3 - https://github.com/389ds/389-ds-base/commit/8537cee0d89672e6e9c73b6708c0d8a61745d43b
You won't find COMPACT_CL5 in `main` branch as we moved changelog and we can compact the changelog with DBCompactTask.

progier389:
Looks like I should have read again the PR description! :wink:

progier389 (approved):
LGTM

PR #6263, opened 2024-07-16
Issue 6256 - nsslapd-numlisteners limit is not enforced

ssidhaye:
Description: Add a test to check if nsslapd-numlisteners value
can be set higher than 4

Relates: https://github.com/389ds/389-ds-base/issues/6256

Reviewed by: droideck

### DIFF
--- a/dirsrvtests/tests/suites/config/config_test.py
+++ b/dirsrvtests/tests/suites/config/config_test.py
@@ -11,6 +11,7 @@ import logging
 import pytest
 import os
 from lib389 import DirSrv, pid_from_file
+from lib389.dseldif import DSEldif
 from lib389.tasks import *
 from lib389.topologies import topology_m2, topology_st as topo
 from lib389.utils import *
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -741,6 +742,57 @@ def test_lmdb_config(create_lmdb_instance):
...
+    topo.standalone.config.set('nsslapd-numlisteners', '4')
+
+    # Check nsslapd-numlisteners value is set to 4
+    assert topo.standalone.config.get_attr_val_utf8('nsslapd-numlisteners') == '4'
+
+    # Check instance dse.ldif value is set to 5
+    inst = topo.standalone
+    dse_ldif = DSEldif(inst)
+    numlisteners = dse_ldif.get("cn='Directory Manager'", 'nsslapd-numlisteners')
+    assert numlisteners[0] == '5'
```
droideck: When I run the test, I get this failure:

```
        # Check instance dse.ldif value is set to 5
        inst = topo.standalone
        dse_ldif = DSEldif(inst)
        numlisteners = dse_ldif.get("cn='Directory Manager'", 'nsslapd-numlisteners')
>       assert numlisteners[0] == '5'
E       TypeError: 'NoneType' object is not subscriptable
```

droideck: Also, I wonder why `nsslapd-numlisteners` should be `5` when in `dse.ldif` but `4` via search?

droideck (changes requested):

dirsrvtests/tests/suites/config/config_test.py
jchapma: Hey @droideck 
There is a "feature" in the numlistener config code that allows an invalid ```nsslapd-numlisteners``` attribute value be written to dse.ldif before the config code gets a chance to sanitise it.

dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -741,6 +742,57 @@ def test_lmdb_config(create_lmdb_instance):
...
+    # Set nsslapd-numlisteners value to 4
+    topo.standalone.config.set('nsslapd-numlisteners', '4')
+
+    # Check nsslapd-numlisteners value is set to 4
+    assert topo.standalone.config.get_attr_val_utf8('nsslapd-numlisteners') == '4'
+
+    # Check instance dse.ldif value is set to 5
+    inst = topo.standalone
+    dse_ldif = DSEldif(inst)
+    numlisteners = dse_ldif.get("cn='Directory Manager'", 'nsslapd-numlisteners')
```
jchapma: I think the reason we get None here is down to the entrydn value in the call to dse_ldif.get(). The ```nsslapd-numlisteners``` attribute and value can be found in "cn=config".

dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -741,6 +742,57 @@ def test_lmdb_config(create_lmdb_instance):
...
+        6. nsslapd-numlisteners value should be successfully set to 5
+        7. nsslapd-numlisteners value is indeed set to 5
+    """
+    # Check default value for nsslapd-numlisteners is 1
+    assert topo.standalone.config.get_attr_val_utf8('nsslapd-numlisteners') == '1'
+
+    # Set nsslapd-numlisteners value to 4
+    topo.standalone.config.set('nsslapd-numlisteners', '4')
+
+    # Check nsslapd-numlisteners value is set to 4
```
jchapma, line 774: When we "set" the value for ````nsslapd-numlisteners``` we need to restart for the change to take effect, but maybe in this case where its just the config we are dealing with, we dont ?

jchapma (commented):
Hey @ssidhaye 

There is a bit of overlap between this PR and [6257](https://github.com/389ds/389-ds-base/pull/6257). 6257 contains a test similar to yours and a fix for this issue. Now that I think about it, is it better to push test cases for a particular fix in a separate PR. If that is the case then I can push the fix after you push this test case. Whatever you think.

dirsrvtests/tests/suites/config/config_test.py
jchapma: But I do prefer the use of DSEldif() over the way I handle dse.ldif in 6257

dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -741,6 +742,57 @@ def test_lmdb_config(create_lmdb_instance):
...
+    inst = topo.standalone
+    dse_ldif = DSEldif(inst)
+    numlisteners = dse_ldif.get("cn='Directory Manager'", 'nsslapd-numlisteners')
+    assert numlisteners[0] == '5'
+
+    # Restart instance
+    topo.standalone.restart()
+
+    # Check nsslapd-numlisteners value is set to 5
+    assert topo.standalone.config.get_attr_val_utf8('nsslapd-numlisteners') == '5'
```
jchapma: The max value for the ```nsslapd-numlisteners``` attr is 4, so if we get 5 we are in trouble

dirsrvtests/tests/suites/config/config_test.py
ssidhaye: I forgot to replace  `Directory Manager` with `DN_CONFIG`.
Sorry about the check for 5. I was trying out a few things and did not change it back to 4.
Fixed it

dirsrvtests/tests/suites/config/config_test.py
ssidhaye: Yes I changed it to use `DN_CONFIG`

dirsrvtests/tests/suites/config/config_test.py
ssidhaye: Yes, fixed it. I misread it earlier.

droideck (approved):
LGTM

dirsrvtests/tests/suites/config/config_test.py
ssidhaye: Yes, we don't need to restart

PR #6266, opened 2024-07-16
Issue 6265 - lmdb - missing entries in range searches

### LINKED ISSUE #6265: with lmdb a range search may fail to return some entries leading to truncated bulk import
**Issue Description**
Several issue seen after generating ldif with 2000 users and importing it in a replica:

1. The entryid attribute in missing in the suffix entry.
2. Access log shows that the internal search looking for "(parentid>=1)" is not returning all entries but one.
3. When initializing a replica through a replication agreement some entries are missing (because of 2)
4. Once 2. get fixed, the bulk import still fails because the default values for nsds5ReplicaFlowControlWindow and nsds5ReplicaFlowControlPause are not adapted to lmdb (supplier sent the entry faster than bdb and the target replica import them slower

 
 - Platform: Fedora
 - Package and version: 3.0.1
 - Browser: N/A

**Steps to Reproduce**
1. Configure two suppliers replication
2. Generate an ldif with 2000 users
3. Import the ldif
4.  search for entryid attribute in the suffix entry
5. search for entries with objectclass=*
6. search for entries with parentid>=1
7. Initialize second replica through first replica agreement.
8.search for entries with objectclass=* on second replica

**Expected results**

- [4] entryid should be 1
- [5] Should have one more entry than [6]
- [7] Initialization should be successful
- [8]  Should return sam,e number of entries than [5]

progier389:
Several issue seen after generating ldif with 2000 users and importing it in a replica:

1. The entryid attribute in missing in the suffix entry.
2. Access log shows that the internal search looking for "(parentid>=1)" is not returning all entries but one.
3. When initializing a replica through a replication agreement some entries are missing (because of 2)
4. Once 2. get fixed, the bulk import still fails because the default values for nsds5ReplicaFlowControlWindow and nsds5ReplicaFlowControlPause are not adapted to lmdb (supplier sent the entry faster than bdb and the target replica import them slower.

The fix is about 
1. Ensuring that the operational attribute are properly set when importing the suffix entry.
2. Avoid using database bulk operation when computing range unless we are sure that bdb is used. 
3. Fixed by 2
4. Change the default values for nsds5ReplicaFlowControlWindow and nsds5ReplicaFlowControlPause if agreement is on a lmdb backend.

Issue: #6265 

Reviewed by: @vashirov, @droideck  (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -276,10 +276,9 @@ def preserve_topo_m2(topo_m2, request):
     if not DEBUGGING:
         request.addfinalizer(fin)
 
-    bindcn = "replication managerg"
+    bindcn = "replication manager"
     binddn = f"cn={bindcn},cn=config"
     bindpw = ''.join(random.choices(string.ascii_letters + string.digits, k=15))
-    bindpw = "secret12"
     for inst in topo_m2:
         backup_dir = f'{inst.ds_paths.backup_dir}/topo_bak'
         try:
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -15,6 +15,10 @@ import ldap
 import pytest
 import subprocess
 import time
+import random
+import string
+from shutil import rmtree
+from lib389.dbgen import dbgen_users
 from lib389.idm.user import TEST_USER_PROPERTIES, UserAccount, UserAccounts
 from lib389.pwpolicy import PwPolicyManager
 from lib389.utils import *
@@ -25,7 +29,7 @@ from lib389.idm.group import Groups, Group
 from lib389.idm.domain import Domain
 from lib389.idm.directorymanager import DirectoryManager
 from lib389.idm.services import ServiceAccounts, ServiceAccount
-from lib389.replica import Replicas, ReplicationManager, ReplicaRole
+from lib389.replica import Replicas, ReplicationManager, ReplicaRole, BootstrapReplicationManager
 from lib389.agreement import Agreements
 from lib389 import pid_from_file
 from lib389.dseldif import *
@@ -256,6 +260,47 @@ def topo_with_sigkill(request):
     request.addfinalizer(fin)
 
     return topology
+
+
+@pytest.fixture(scope="function")
+def preserve_topo_m2(topo_m2, request):
+    """Backup the topology and restore it at the end."""
+
+    saves = []
+
+    def fin():
+        for inst,backup_dir in saves:
+            inst.tasks.bak2db(backup_dir=backup_dir, args={TASK_WAIT: True})
+            rmtree(backup_dir)
+
+    if not DEBUGGING:
+        request.addfinalizer(fin)
+
+    bindcn = "replication managerg"
+    binddn = f"cn={bindcn},cn=config"
+    bindpw = ''.join(random.choices(string.ascii_letters + string.digits, k=15))
+    bindpw = "secret12"
+    for inst in topo_m2:
+        backup_dir = f'{inst.ds_paths.backup_dir}/topo_bak'
+        try:
+            rmtree(backup_dir)
+        except FileNotFoundError:
+            pass
+        inst.tasks.db2bak(backup_dir=backup_dir, args={TASK_WAIT: True})
+        # Ensure that we are not using group bind dn
+        # because the test may delete credentials
+        replmgr = BootstrapReplicationManager(inst, dn=binddn)
+        if replmgr.exists():
+            replmgr.replace('userPassword', bindpw)
+        else:
+            replmgr.create(properties={'cn':bindcn, 'userPassword':bindpw})
+        replica = Replicas(inst).get(DEFAULT_SUFFIX)
+        replica.replace(REPL_BINDDN, binddn);
+        replica.remove_all(REPL_BIND_GROUP);
+        for agmt in replica.get_agreements().list():
+            agmt.replace_many((AGMT_CRED, bindpw), (REPL_BINDDN, binddn))
+        saves.append((inst, backup_dir))
+    return topo_m2
 
 
 @pytest.fixture()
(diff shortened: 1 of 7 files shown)


dirsrvtests/tests/suites/replication/regression_m2_test.py
```diff
@@ -258,6 +262,47 @@ def fin():
...
+
+    def fin():
+        for inst,backup_dir in saves:
+            inst.tasks.bak2db(backup_dir=backup_dir, args={TASK_WAIT: True})
+            rmtree(backup_dir)
+
+    if not DEBUGGING:
+        request.addfinalizer(fin)
+
+    bindcn = "replication managerg"
```
vashirov: Typo
```suggestion
    bindcn = "replication manager"
```

dirsrvtests/tests/suites/replication/regression_m2_test.py
```diff
@@ -258,6 +262,47 @@ def fin():
...
+    binddn = f"cn={bindcn},cn=config"
+    bindpw = ''.join(random.choices(string.ascii_letters + string.digits, k=15))
+    bindpw = "secret12"
```
vashirov: bindpwd is replaced here, was it used for debugging?

vashirov (commented):
I reviewed only the tests, just 2 minor comments.

progier389:
Fixed the two issues that @vashirov has found.

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -670,9 +703,294 @@ idl_new_range_fetch(
...
+                      "Database %s index is corrupt; key %s has a data item with the wrong size (%ld)\n",
+                      rctx->index_id, (char *)key->data, data->size);
+        rctx->flag_err = LDAP_UNWILLING_TO_PERFORM;
+        return DBI_RC_NOTFOUND;
+    }
+    memcpy(&id, data->data, sizeof(ID));
+    if (id == rctx->lastid) {
+        slapi_log_err(SLAPI_LOG_TRACE, "idl_lmdb_range_fetch",
+                      "Detected duplicate id %d due to DB_MULTIPLE error - skipping\n", id);
+        return 0;
```
droideck: Shouldn't it be set to DBI_RC_SUCCESS?

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -670,9 +703,294 @@ idl_new_range_fetch(
...
+    if ((NEW_IDL_NO_ALLID != rctx->flag_err) && rctx->ai && (rctx->idl != NULL) &&
+        idl_new_exceeds_allidslimit(rctx->count, rctx->ai, rctx->allidslimit)) {
+        rctx->idl->b_nids = 1;
+        rctx->idl->b_ids[0] = ALLID;
+        return DBI_RC_NOTFOUND; /* fool the code below into thinking that we finished the dups */
+    }
+#endif
+
+    rctx->count++;
+    return 0;
```
droideck: Same here with DBI_RC_SUCCESS

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -670,9 +703,294 @@ idl_new_range_fetch(
...
+        dblayer_read_txn_commit(be, &s_txn);
+    }
+    if (0 == idl_range_ctx.flag_err) {
+        idl_range_ctx.flag_err = ret;
+slapi_log_err(SLAPI_LOG_INFO, "idl_lmdb_range_fetch", "flag_err=%d\n", idl_range_ctx.flag_err);
+    }
+
+    /* sort idl */
+    if (!ALLIDS(idl_range_ctx.idl) && !(operator&SLAPI_OP_RANGE_NO_IDL_SORT)) {
+        qsort((void *)&idl_range_ctx.idl->b_ids[0], idl_range_ctx.idl->b_nids, (size_t)sizeof(ID), idl_sort_cmp);
```
droideck: Is there any reason to typecast (size_t) on sizeof(ID)? I think it returns size_t as is.

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -358,8 +364,33 @@ agmt_new_from_entry(Slapi_Entry *e)
...
+        }
+        be = slapi_be_select(ra->replarea);
+    }
+    if (!be) {
+        slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name,
+                      "agmt_new_from_entry - Failed to get backend for agreement %s on replicated suffix %s).\n",
+                      slapi_entry_get_dn(e), tmpstr ? tmpstr : "<NULL>");
+        goto loser;
+    }
+    use_lmdb = dblayer_is_lmdb(be);
```
droideck: Minor nitpick. I think it'll be a bit more readable if `use_lmdb = dblayer_is_lmdb(be);` line will be right before `ra->flowControlWindow = use_lmdb ? LMDB_DEFAULT_FLOWCONTROL_WINDOW : DEFAULT_FLOWCONTROL_WINDOW;` line

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -670,9 +703,294 @@ idl_new_range_fetch(
...
+        slapi_log_err(SLAPI_LOG_TRACE, "idl_range_add_id", "lookthrough_limit exceeded\n");
+        rctx->flag_err = LDAP_ADMINLIMIT_EXCEEDED;
+        return DBI_RC_NOTFOUND;
+    }
+    if ((rctx->sizelimit > 0) && (rctx->idl->b_nids > (ID)rctx->sizelimit)) {
+        slapi_log_err(SLAPI_LOG_TRACE, "idl_range_add_id", "sizelimit exceeded\n");
+        rctx->flag_err = LDAP_SIZELIMIT_EXCEEDED;
+        return DBI_RC_NOTFOUND;
+    }
+    if ((rctx->idl->b_nids & 0xff) == 0 &&
```
droideck: Do you mean SLAPI_OP_RANGE here?

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -670,9 +703,294 @@ idl_new_range_fetch(
...
+    if ((rctx->upperkey != NULL) && (rctx->upperkey->data != NULL)) {
+        if ((rctx->operator & SLAPI_OP_RANGE) == SLAPI_OP_LESS) {
+            if (keycmp(key, rctx->upperkey, rctx->ai->ai_key_cmp_fn) >= 0) {
+                return DBI_RC_NOTFOUND;
+            }
+        } else {
+            if (keycmp(key, rctx->upperkey, rctx->ai->ai_key_cmp_fn) > 0) {
+                return DBI_RC_NOTFOUND;
+            }
+        }
```
droideck: This `if` seems a bit cryptic (at least to me). Could you please add a small comment explaining the difference between `>=` and `>` and how it's related to if rctx->operator is SLAPI_OP_LESS?

droideck (changes requested):

ldap/servers/slapd/back-ldbm/idl_new.c
progier389: Nothing really complicated here:
 In a range, the subfilter operator is either SLAPI_OP_LESS or SLAPI_OP_LESS_OR_EQUAL,
   the code just check for the last key accordingly
but I can add a comment telling that the else case is (rctx->operator & SLAPI_OP_RANGE) == SLAPI_OP_LESS_OR_EQUAL)

ldap/servers/slapd/back-ldbm/idl_new.c
progier389: Only because this line (and most of this function logic) is copied from idl_new_range_fetch

ldap/servers/slapd/back-ldbm/idl_new.c
progier389: But anyway the logic is the same as idl_new_range_fetch

ldap/servers/slapd/back-ldbm/idl_new.c
progier389: No, only testing the time limit every 256 candidates to speed up things

progier389:
Fixed @droideck review comments

droideck (approved):
LGTM

Issue #6245, opened 2024-07-01
Fix some coverity scan issues


progier389:
**Issue Description**
Fix the issue rated as "important"  reported by coverity scan static analyzer

**Package Version and Platform:**
 - Platform:  Fedora
 - Package and version: main branch 

**Additional context**

Using some internal tools to run coverity scan 
`osh-cli mock-build --config=fedora-rawhide-x86_64 $WS/dist/srpms/389-ds-base-*.src.rpm
`
But I guess this should also be visible  in the  github coverity scan test results

Triaged: https://issues.redhat.com/browse/IDMDS-4302

PR #6273, opened 2024-07-24
Issue 6245 - Fix some other coverity scan regressions

progier389:
Fixing some of the new detected issues in
    https://scan6.scan.coverity.com/#/project-view/59218/11545?selectedIssue=1558490

Issue: #6245 

Reviewed by:  @droideck


lib/libadmin/util.c
```diff
@@ -142,7 +150,15 @@ cookieValue(char *var, char *val)
             cookie = STRDUP(cookie);
             numVars = 0;
             vars = (char **)MALLOC(sizeof(char *));
+            if (vars == NULL) {
+                return NULL;
+            }
             vals = (char **)MALLOC(sizeof(char *));
+            if (vals == NULL) {
+                FREE(vars);
+                return NULL;
```
droideck: Should we also free `cookie` before the returns?

droideck (approved):
The rest looks good!

lib/libadmin/util.c
progier389: Good catch I missed the STRDUP.
We should also check if STRDUP fails ....

Issue #6280, opened 2024-07-31
Changelog trims updates from a given RID even if a consumer has not received any of them


tbordaz:
**Issue Description**
If a consumer has not received any of the updates from a given RID, the RUV (in the replication agreement) contains no csn. In such case, the changelog trimming thread compute the starting purge csn using the supplier RUV instead of  the RA RUV. Later the supplier will miss this csn to update the consumer and replication is broken

**Package Version and Platform:**
 - all versions
**Steps to Reproduce**

1. script shell allows to reproduce it systematically. Need to translate it into CI

**Expected results**
supplier should keep the CSN that have not yet been seen by a consumer

**Additional context**

tbordaz:
e70c22fa0..0cfe9f268  main
d2b736d25..a9cbacdae  389-ds-base-3.0
308942491..bd7ca1e9b  389-ds-base-2.5
4a8ee574a..8edf050a9  389-ds-base-2.4
8f42d3ce8..4a1c4ba49  389-ds-base-2.3
6ddec9446..1bd36669c  389-ds-base-2.2
2cec5cea7..387c9b8a3  389-ds-base-2.1
692babe2d..945838780  389-ds-base-2.0
85cb1f0f5..e7ac6d875  389-ds-base-1.4.4
7f0b5363e..976acfd18  389-ds-base-1.4.3

PR #6281, opened 2024-07-31
Issue 6280 - Changelog trims updates from a given RID even if a consu…

tbordaz:
…mer has not received any of them

Bug description:
        If a consumer has not received any of the updates from a given RID, the RUV
        (in the replication agreement) contains no csn.
        In such case, the changelog trimming thread compute the starting purge csn
        using the supplier RUV instead of the RA RUV.
        Later the supplier will miss this csn to update the consumer and
        replication is broken

Fix description:
	The function that iterates the elements of the replication agreement RUV
        should call the callback (_cl5EnumConsumerRUV) with all elements even those
        containing empty CSN.
	The callback (_cl5EnumConsumerRUV) should remove the RID
	from the replica RUV in case the RA.RUV is empty.

relates: #6280

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/replication/changelog_trimming_test.py
+++ b/dirsrvtests/tests/suites/replication/changelog_trimming_test.py
@@ -14,9 +14,14 @@ import time
 from lib389._constants import *
 from lib389.properties import *
 from lib389.topologies import topology_m1 as topo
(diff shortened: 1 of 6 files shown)


progier389 (approved):
LGTM

droideck (approved):
Looks good!

PR #6285, opened 2024-08-02
Issue 6284 - BUG - freelist ordering causes high wtime

### LINKED ISSUE #6284: high wtimes on high connection load.
**Issue Description**
Following #4812 a lot of improvements were made to scalability. Currently a major custom is testing 389-ds and is continuing to hit limitations. This mostly is around high wtimes under high connection load. The user is also seeing low CPU consumption and low etimes, indicating that issues still persist around the connection/worker handling code.

I plan to either backport missing performance fixes to 2.2, or to improve and add more improvements to 2.2 (and main). 

 - Platform: SUSE Linux Enterprise 15 SP5
 - Package and version: 389-ds-base-2.2

Firstyear:
Bug Description: Multithread listeners were added to support having multiple threads able to perform IO for connections to reduce wtime. However, when the server is first started if the number of clients is less than the size of a single listeners conntable this causes extremely high wtimes.

This is because the initial connection freelist
construction adds every connection in order of *listener* so new connections always fill the first listener thread, then only once that listener is full it spills over to the next.

If the conntable size is small, this isn't noticeable, but an example is if your contable is 16384 elements with 4 listeners, then the first 4096 connections will always go to the first listener.

Fix Description: When the freelist is constructed rather than iterate over each listener *then* each slot, we iterate over each slot then each listenr. This causes the initial freelist to be interleaved between all listeners so that even with a small number of connections the work is spread fairly without lasering a single listener.

A supplied test from a user shows a significant drop in wtime and an increase in throughput of operations with this change.

fixes: https://github.com/389ds/389-ds-base/issues/6284

Review by: ???

### DIFF
--- a/ldap/servers/slapd/conntable.c
+++ b/ldap/servers/slapd/conntable.c
@@ -115,20 +115,26 @@ connection_table_new(int table_size)
 {
     Connection_Table *ct;
     size_t i = 0;
-    int ct_list = 0;
-    int free_idx = 0;
+    size_t ct_list = 0;
+    size_t free_idx = 0;
     ber_len_t maxbersize = config_get_maxbersize();
     ct = (Connection_Table *)slapi_ch_calloc(1, sizeof(Connection_Table));
     ct->list_num = config_get_num_listeners();
     ct->num_active = (int *)slapi_ch_calloc(1, ct->list_num * sizeof(int));
+    /* Connection table size must be divisible by the number of listeners */
     ct->size = table_size - (table_size % ct->list_num);
     /* Account for first slot of each list being used as head (c_next).  */
     ct->list_size = (ct->size/ct->list_num)+1;
+    /*
+     * Since there are extra list heads, in the case ct->size was a multiple of list_num, we
+     * need to expand ct->size to reflect the true allocation amount.
+     */
+    ct->size = ct->list_size * ct->list_num;
     ct->c = (Connection **)slapi_ch_calloc(1, ct->size * sizeof(Connection *));
     ct->fd = (struct POLL_STRUCT **)slapi_ch_calloc(1, ct->list_num * sizeof(struct POLL_STRUCT*));
     ct->table_mutex = PR_NewLock();
     /* Allocate the freelist (a slot for each connection plus another slot for the final NULL pointer) */
-    ct->c_freelist = (Connection **)slapi_ch_calloc(1, (ct->size+1) * sizeof(Connection *));
+    ct->c_freelist = (Connection **)slapi_ch_calloc(1, (ct->size + 1) * sizeof(Connection *));
     ct->conn_next_offset = 0;
 
     slapi_log_err(SLAPI_LOG_INFO, "connection_table_new", "Number of connection sub-tables %d, each containing %d slots.\n",
@@ -184,11 +190,26 @@ connection_table_new(int table_size)
 
             /* Ready to rock, mark as such. */
             ct->c[ct_list][i].c_state = CONN_STATE_INIT;
+        }
+    }
 
+    /*
+     * The freelist is how new connections are inserted into our ct to be used. We need to ensure
+     * these are *spread* initially over the set of listeners so that we distributed between handlers
+     * from the start. Over time as things are re-added to the freelist it will "shuffle" and we
+     * will effectively have randomised listener assignment.
+     *
+     * Without this it means that a single listener will be "focused on" until it's conntable
+     * is full at which point we spill out into the next listener.
+     *
+     * This is why in the subsequent loops we INVERT the loop to iterate over list_size
+     * first, and then the ct_list as the inner component so that the freelist initially
+     * alternates between the listeners first to help distribute requests.
+     */
+    for (i = 1; i < ct->list_size; i++) {
+        for (ct_list = 0; ct_list < ct->list_num; ct_list++) {
             /* Map multiple ct lists to a single freelist, but skip slot 0 of each list. */
-            if (i != 0) {
-                ct->c_freelist[free_idx++] = &(ct->c[ct_list][i]);
-            }
+            ct->c_freelist[free_idx++] = &(ct->c[ct_list][i]);
         }
     }
 


tbordaz (approved):
Nice finding !! that explains why we were not able to measure improvement on throughput. The initial clients were all hitting the same listener.

The patch looks good

A question though, my understanding is that wtime starts when the listener creates a pblock and adds it to the work queue. The problem being that the first listener is overloaded and slow, it should only add the pblock slower. I understand it impacts throughput but I miss how it impact the time spent in the workqueue by a specific pblock ?

Firstyear:
> Nice finding !! that explains why we were not able to measure improvement on throughput. The initial clients were all hitting the same listener.
> 
> The patch looks good
> 
> A question though, my understanding is that wtime starts when the listener creates a pblock and adds it to the work queue. The problem being that the first listener is overloaded and slow, it should only add the pblock slower. I understand it impacts throughput but I miss how it impact the time spent in the workqueue by a specific pblock ?



handle_pr_read_ready() is what we poll with the multiple listeners. This iterates over it's list of connections which calls connection_activity(). If activity is found, the connection is put on the work_q.

Then connection_threadmain picks up the operation and if it's the first iteration of the loop, initialises the operation. This is where wtime *starts*.

connection_threadmain then attempts connection_read_operation. If the PDU is *incomplete* then the connection needs more data, so it's unlocked. It is removed from the work_q and returned to be polled.

Then handle_pr_read_ready() runs again, when more data is available. connection_threadmain picks up the connection again and since the connection is already initialised the pblock remains. Now connection_read_operation completes the PDU and can proceed. wtime now *ends*.

This is how wtime is affected because if there is a single listener we have to wait longer for the second handle_pr_read_ready() as we have a longer head of connections in front that need to be handled first. 

By having multiple listeners parallel polling then in the case of incomplete PDU's, we return to poll the connection sooner reducing the wtime.

jchapma:
Why is it when someone explains something it seems so obvious. 

The intention was that when a connection becomes active, connection_table_get_list() will find the least used list to assign the connection to, balancing the connections across multi lists. 

I am very interested in doing a throughput comparison to measure the benefit of your fix.

jchapma (approved):

Firstyear:
No problem, happy to have found it. Depending on our customers testing and feedback there may be more coming to improve the problem :) 

I'm planning to backport this to 2.5 at least if that's no issue. Our customer is on 2.2 so we make need to backport to that too incase we aren't able to update to 2.5.

jchapma:
@Firstyear 
I cant see any issue with back porting to all relevant branches, which if I remember correctly is 2.2 upwards.

tbordaz:
> > Nice finding !! that explains why we were not able to measure improvement on throughput. The initial clients were all hitting the same listener.
> > The patch looks good
> > A question though, my understanding is that wtime starts when the listener creates a pblock and adds it to the work queue. The problem being that the first listener is overloaded and slow, it should only add the pblock slower. I understand it impacts throughput but I miss how it impact the time spent in the workqueue by a specific pblock ?
> 
> handle_pr_read_ready() is what we poll with the multiple listeners. This iterates over it's list of connections which calls connection_activity(). If activity is found, the connection is put on the work_q.
> 
> Then connection_threadmain picks up the operation and if it's the first iteration of the loop, initialises the operation. This is where wtime _starts_.
> 
> connection_threadmain then attempts connection_read_operation. If the PDU is _incomplete_ then the connection needs more data, so it's unlocked. It is removed from the work_q and returned to be polled.
> 
> Then handle_pr_read_ready() runs again, when more data is available. connection_threadmain picks up the connection again and since the connection is already initialised the pblock remains. Now connection_read_operation completes the PDU and can proceed. wtime now _ends_.
> 
> This is how wtime is affected because if there is a single listener we have to wait longer for the second handle_pr_read_ready() as we have a longer head of connections in front that need to be handled first.
> 
> By having multiple listeners parallel polling then in the case of incomplete PDU's, we return to poll the connection sooner reducing the wtime.

@Firstyear , thank you sooo much for this detailed explanations. So my understandings are

- In server startup this bug overloaded the primary listener by putting the free connection in its bucket
- initial clients are all hitting the same listener, preventing the benefit of the multi-listener and keeping the initial throughput as if there was a single listener
- If requests are large or just incomplete, then the listener has to do several poll/read on the same connection, the overloaded/slow listener is responsible of high wtime.

So high wtime are only seen when clients send large/incomple req. Am I correct ?

I am a bit surprise about the last bullet. I thought that it was the worker thread that poll/read the full request. This may have changed recently (@jchapma do you know ?).

Firstyear:
> I am a bit surprise about the last bullet. I thought that it was the worker thread that poll/read the full request. This may have changed recently (@jchapma do you know ?).

Only in turbo-mode, and only a subset of threads are elligible for this.

Firstyear:
```
   4d7218b75..146c6fde6  389-ds-base-2.2 -> 389-ds-base-2.2
   41add0d65..085b48957  389-ds-base-2.3 -> 389-ds-base-2.3
   84d6ae0f4..3427e2a14  389-ds-base-2.4 -> 389-ds-base-2.4
   161888969..9c5a0ce32  389-ds-base-2.5 -> 389-ds-base-2.5
   f029897e5..3ccf62745  389-ds-base-3.0 -> 389-ds-base-3.0
```

tbordaz:
> > I am a bit surprise about the last bullet. I thought that it was the worker thread that poll/read the full request. This may have changed recently (@jchapma do you know ?).
> 
> Only in turbo-mode, and only a subset of threads are elligible for this.

Okay. That make sense.
I think it may lead to is a significant support issue. So far we have been using wtime as a sign of worker starvation (i.e. time spent in the workq before getting a free worker) that we can address by increasing the number of workers. But because worker can put back an incomplete req in the workq, a ldap client can create fake high wtime. And so high wtime may also measure lazy client or large client req. In that last case an admin should not increase the number of worker.

IMHO wtime should stick to its initial purpose: detect worker starvation

Firstyear:
@jchapma and @tbordaz - our customer has now confirmed with these changes (and them reading their test numbers more closely) that 389-ds on like-for-like hardware etc is twice as fast at connection handling and remains stable to higher numbers than openldap. :)

Firstyear:
@tbordaz Related to your wtime comment, it was invaluable here to find the IO stall time, so we should have a new counter for IOwait as well.

tbordaz:
> @tbordaz Related to your wtime comment, it was invaluable here to find the IO stall time, so we should have a new counter for IOwait as well.

Although I see the benefit for wtime (it will focus wtime on the time to get a free worker), I am unsure on adding a new 'iotime' counter in the access log. If admin/support sees high 'iotime' (related to big req or lazy client), can he tune the server to protect it ? should we log a warning about those large iotime ?

progier389:
Although we cannot do anything to improve iotime on the server, IMHO it is still interesting to compute it: 
  - it is simple/cheap to compute
  - it drastically decrease the troubleshooting time in case of client problem
  - probably also helpful for some replication timing issues

Firstyear:
iotime on a single connection yields very little, but across a wide number of connections it can reveal a trend such as ioworker starvation, slow network, or something else. It gives us a metric we can use to measure improvements to connection handling. And it means wtime is the *true* wait time in queue for a worker, rather than the curernt io + wait that it is today.

As a contrast, to help explain why this is super useful, consider this log from a separate project.

```
Aug 20 10:34:37  INFO     ldap-request [ 1.00ms | 2.66% / 100.00% ]
Aug 20 10:34:37  INFO     ┕━ handle_ldaprequest [ 977µs | 1.86% / 97.34% ]
Aug 20 10:34:37  DEBUG       ┕━ do_search [ 958µs | 37.59% / 95.48% ]
Aug 20 10:34:37  DEBUG          ┝━ proxy_read [ 54.6µs | 4.78% / 5.44% ]
Aug 20 10:34:37  DEBUG          │  ┕━ read_acquire_ticket [ 6.64µs | 0.66% ]
Aug 20 10:34:37  DEBUG          ┝━ validate_ldap_session [ 138µs | 1.93% / 13.80% ]
Aug 20 10:34:37  DEBUG          │  ┕━ internal_search_uuid [ 119µs | 2.71% / 11.87% ]
Aug 20 10:34:37  DEBUG          │     ┕━ search [ 91.8µs | 2.68% / 9.15% ]
Aug 20 10:34:37  DEBUG          │        ┝━ be::search [ 35.1µs | 3.50% ]
Aug 20 10:34:37  DEBUG          │        ┕━ access::filter_entries [ 29.8µs | 2.97% ]
Aug 20 10:34:37  DEBUG          ┝━ search_ext [ 388µs | 1.35% / 38.64% ]
Aug 20 10:34:37  DEBUG          │  ┝━ search [ 253µs | 7.00% / 25.23% ]
Aug 20 10:34:37  DEBUG          │  │  ┝━ be::search [ 40.1µs | 4.00% ]
Aug 20 10:34:37  DEBUG          │  │  ┕━ access::filter_entries [ 143µs | 14.23% ]
Aug 20 10:34:37  DEBUG          │  ┕━ access::search_filter_entry_attributes [ 121µs | 12.06% ]
Aug 20 10:34:37  INFO           ┕━ ｉ [info]: LDAP Search Success -> number of entries | event_tag_id: 3 | nentries: 2
```

Having really good, segmented timings of different server segments like this is invaluable. You can see precisely where time was consumed in the operation and where. The request as a whole took 1.00ms, we can see it look 5% of that time to acquire a ticket (database lock), the search itself took 91.8us, and access controls dominated with 2.97% + 14.23% of our time consumed. 

So we can really easily see here that if we were to improve things we should look at session validation or access control execution.

This is why having timers across things like the iowait time, queue wait time, optime etc is so valuable.

tbordaz:
My question regarding  'iotime' was confusing. I agree we need precise counter and stick 'wtime' to time waiting for a worker. I agree that 'iotime' is valuable (to identify connection handling issue for example) but my question was should we change the access log format, adding iotime, or just log a warning (error log) when it goes over a limit.

Regarding the metrics they are very valuable, thanks sharing them. Not sure what is the validate_ldap_session, is it search/access to base entry ?

Firstyear:
> Regarding the metrics they are very valuable, thanks sharing them. Not sure what is the validate_ldap_session, is it search/access to base entry ?

It's an example, don't stress to much about it. It's from another project, not 389-ds :) 

The point was that having good timings can help identify issues much more accurately.

Issue #6288, opened 2024-08-06
dsidm and cockpit ui crashes when account policy is enabled (float() argument must be a string or a number, not 'NoneType')


navidys:
**Issue Description**
Followed instruction on RHDS-11 documentation in order to setup time-based account policy using CoS.
after setup cockpit UI crashes and also following dsidm command showing error:

```
dsidm  -b dc=MTTS mttsldp01li account entry-status cn=user01,ou=people,dc=MTTS
Error: float() argument must be a string or a number, not 'NoneType'
```

**Package Version and Platform:**
 - Platform: EL8
 - Package and version: 389-ds-base-1.4.3
 - Browser: chrome

**Steps to Reproduce**
1- Enable account policy plugin and do configuration: 
```
# dsconf -D "cn=Directory Manager" <instance> plugin account-policy enable

# dsconf -D "cn=Directory Manager" <instance>  plugin account-policy set --config-entry="cn=config,cn=Account Policy Plugin,cn=plugins,cn=config"

# dsconf -D "cn=Directory Manager" <instance> plugin account-policy config-entry set "cn=config,cn=Account Policy Plugin,cn=plugins,cn=config"  --alt-state-attr createTimestamp --spec-attr acctPolicySubentry --limit-attr accountInactivityLimit --always-record-login yes --state-attr lastLoginTime

# dsctl <instance> restart

```

2- create account inactivity policy

```
ldapadd -D "cn=Directory Manager" -W -H <ldaps instance URI> -x	  
dn: cn=Account Inactivation Policy,dc=MTTS
objectClass: top
objectClass: ldapsubentry
objectClass: extensibleObject
objectClass: accountpolicy
accountInactivityLimit: 20
cn: Account Inactivation Policy

dn: cn=TemplateCoS,dc=MTTS
objectClass: top
objectClass: ldapsubentry
objectClass: extensibleObject
objectClass: cosTemplate
acctPolicySubentry: cn=Account Inactivation Policy,dc=MTTS

dn: cn=DefinitionCoS,ou=People,dc=MTTS
objectClass: top
objectClass: ldapsubentry
objectclass: cosSuperDefinition
objectclass: cosPointerDefinition
cosTemplateDn: cn=TemplateCoS,dc=MTTS
cosAttribute: acctPolicySubentry default operational-default
```

3- create new user account

```
# dsidm -b dc=MTTS <instance> user create
....
```

[...]

PR #6292, opened 2024-08-09
Issue 6288 - dsidm crash with account policy plugin when alt-state-attr is disabled

navidys:
Issue 6288 - dsidm crash with account policy plugin when alt-state-attr is disabled

Bug Description:
If disable alt-state-attr for account policy plugin by using value 1.1 then dsidm command will crash if state-attr doesn't exit at that time.

Fix Description:
Check if alt-state-attri key exist under account data dict before getting its value.

relates: https://github.com/389ds/389-ds-base/issues/6288

Author: Navid Yaghoobi

Reviewed by: @progier389


progier389 (approved):
LGTM

Issue #6065, opened 2024-02-01
WebUI supports only instances with BDB


vashirov:
**Issue Description**
It's not possible to choose DB backend when a new instance is created.
And for existing instances WebUI expects `dsconf` to return BDB specific parameters. For example, a page throws an Oops... when I navigate to `Database` tab on an instance with LMDB backend:

https://github.com/389ds/389-ds-base/blob/1f95b57fb8b03f72c7d2a2ce5b14eed5ba320407/src/cockpit/389-console/src/database.jsx#L251

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at Function.<anonymous> (database.jsx:251:55)
    at process_queue (cockpit.js:1311:37)
    at cockpit.js:1324:35
    at later_drain (cockpit.js:1205:13)
```

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main branch
 - Browser: Firefox

**Steps to Reproduce**
1. Create a new instance with MDB backend
2. Go to Database tab
3. Observe an Oops...

**Expected results**
WebUI should support instances with both BDB and LMDB backends and not crash if some attributes are not available.

progier389:
Created IDMDS-4185 ticket

PR #6299, opened 2024-08-14
Issue 6056 - WebUI supports only instances with BDB

jchapma:
Description:
BDB specific attributes are not supported by the web console, resulting in an "Ooops" when the Database tab is selected.

Fix Description:
Add a new global Database configuration element for MDB layout. Add support to detect what DB engine is configured, and display the correct layout on rendering.

Relates: https://github.com/389ds/389-ds-base/issues/6065

Reviewed by:

### DIFF
--- a/src/cockpit/389-console/src/database.jsx
+++ b/src/cockpit/389-console/src/database.jsx
@@ -85,12 +85,12 @@ export class Database extends React.Component {
             globalDBConfig: {
                 activeTab: 0,
             },
+            backendImplement: BE_IMPL_MDB,
             configUpdated: 0,
             // Chaining Config
             chainingConfig: {},
             chainingUpdated: 0,
             chainingActiveKey: 0,
-            backendImplement: "mdb",
             // Chaining Link
             chainingLoading: false,
             // Suffix
@@ -250,7 +250,6 @@ export class Database extends React.Component {
                         dbimplement = attrs['nsslapd-backend-implement'][0];
                         this.setState({ backendImplement: dbimplement });
                     }
-
                     if (dbimplement === BE_IMPL_BDB) {
                         let db_cache_auto = false;
                         let import_cache_auto = false;
@@ -319,7 +318,6 @@ export class Database extends React.Component {
                                         mdbmaxreaders: attrs['nsslapd-mdb-max-readers'][0],
                                         mdbmaxdbs: attrs['nsslapd-mdb-max-dbs'][0],
                                         dbhomedir: dbhome,
-                                        importcachesize: attrs['nsslapd-import-cachesize'][0],
                                     },
                                 configUpdated: 1
                             }), () => { this.loadNDN() });
--- a/src/cockpit/389-console/src/database.jsx
+++ b/src/cockpit/389-console/src/database.jsx
@@ -5,7 +5,7 @@ import {
     ChainingConfig,
     ChainingDatabaseConfig
 } from "./lib/database/chaining.jsx";
-import { GlobalDatabaseConfig } from "./lib/database/databaseConfig.jsx";
+import { GlobalDatabaseConfig, GlobalDatabaseConfigMDB } from "./lib/database/databaseConfig.jsx";
 import { Suffix } from "./lib/database/suffix.jsx";
 import { Backups } from "./lib/database/backups.jsx";
 import { GlobalPwPolicy } from "./lib/database/globalPwp.jsx";
@@ -49,6 +49,8 @@ const CHAINING_CONFIG = "chaining-config";
 const BACKUP_CONFIG = "backups";
 const PWP_CONFIG = "pwpolicy";
 const LOCAL_PWP_CONFIG = "localpwpolicy";
+const BE_IMPL_BDB = "bdb";
+const BE_IMPL_MDB = "mdb";
 
 const _ = cockpit.gettext;
 
@@ -88,6 +90,7 @@ export class Database extends React.Component {
             chainingConfig: {},
             chainingUpdated: 0,
             chainingActiveKey: 0,
+            backendImplement: "mdb",
             // Chaining Link
             chainingLoading: false,
             // Suffix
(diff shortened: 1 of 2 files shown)


src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+                        </Text>
+                    </TextContent>
+
+                    <div className="ds-margin-top-lg">
+                        <Tabs isFilled activeKey={this.state.activeTabKey} onSelect={this.handleNavSelect}>
+                            <Tab eventKey={0} title={<TabTitleText>{_("Database Size")}</TabTitleText>}>
+                                <div className="ds-left-indent-md">
+                                    <Grid
+                                        title={_("Database maximum size in bytes. The practical maximum size of an LMDB database is limited by the system’s addressable memory (nsslapd-mdb-max-size).")}
+                                        className="ds-margin-top"
```
mreynolds389: I noticed several fields are very close to bottom of the tabs, I think it would look a little nicer if you used this class instead:  ds-margin-top-xlg

Do this for all the grids below so they are consistent, thanks!

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+                                                widthChars={10}
+                                            />
+                                        </GridItem>
+                                    </Grid>
+                                    <Grid
+                                        title={_("The maximum number of named database instances that can be included within the memory mapped database file (nsslapd-mdb-max-dbs).")}
+                                        className="ds-margin-top"
+                                    >
+                                        <GridItem className="ds-label" span={4}>
+                                            {_("Database Max DBS")}
```
vashirov:
```suggestion
                                            {_("Database Max DBs")}
```

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+                            {_("Global Database Configuration")}
+                        </Text>
+                    </TextContent>
+
+                    <div className="ds-margin-top-lg">
+                        <Tabs isFilled activeKey={this.state.activeTabKey} onSelect={this.handleNavSelect}>
+                            <Tab eventKey={0} title={<TabTitleText>{_("Database Size")}</TabTitleText>}>
+                                <div className="ds-left-indent-md">
+                                    <Grid
+                                        title={_("Database maximum size in bytes. The practical maximum size of an LMDB database is limited by the system’s addressable memory (nsslapd-mdb-max-size).")}
```
vashirov: Maximum size should be a multiple of the page size, which can vary (4K, 16K, 64K on different architectures). And as a user, I'd prefer to enter the size in megabytes, than calculate the right size.

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+                                            />
+                                        </GridItem>
+                                    </Grid>
+                                </div>
+                            </Tab>
+
+                            <Tab eventKey={5} title={<TabTitleText>{_("Advanced Settings")}</TabTitleText>}>
+                                <div className="ds-left-indent-md">
+                                    <Grid
+                                        title={_("Location for database memory mapped files.  You must specify a subdirectory of a tempfs type filesystem (nsslapd-db-home-directory).")}
```
vashirov:
```suggestion
                                        title={_("Location for database memory mapped files. You must specify a subdirectory of a tempfs type filesystem (nsslapd-db-home-directory).")}
```

vashirov (commented):
Few small issues. For the rest I need to try the PR locally.

src/cockpit/389-console/src/database.jsx
```diff
@@ -88,6 +90,7 @@ export class Database extends React.Component {
             chainingConfig: {},
             chainingUpdated: 0,
             chainingActiveKey: 0,
+            backendImplement: "mdb",
```
droideck: Why not to use `BE_IMPL_MDB` const here?

progier389:
Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation 
  I think it is dead code, neither the entries nor the index are cached in lmdb import )
  so IMHO it will be better to simply remove that tab

mreynolds389:
> Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation I think it is dead code, neither the entries nor the index are cached in lmdb import ) so IMHO it will be better to simply remove that tab

I think the import cache is used for other things that are not DB related (regardless of BDB/MDB) - so it still might be needed.  Just proceed with caution :-)

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+                                    >
+                                        <GridItem className="ds-label" span={4}>
+                                            {_("Database Home Directory")}
+                                        </GridItem>
+                                        <GridItem span={8}>
+                                            <TextInput
+                                                value={this.state.dbhomedir}
+                                                type="text"
+                                                id="dbhomedir"
+                                                aria-describedby="dbhomedir"
```
droideck, line 1537: For some reason, when I change dbhome directory, the button still stays grey. And if I change something else and then save, dbhome dir still resets to the default - `/var/lib/dirsrv/slapd-inst/db`

And it happens only for LMDB. BDB is fine

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -990,3 +990,580 @@ GlobalDatabaseConfig.defaultProps = {
...
+        this.setState({
+            [attr]: value,
+        }, () => { this.validateSaveBtn() });
+    }
+
+    handleTimeChange(value) {
+        this.setState({
+            compacttime: value,
+        }, () => { this.validateSaveBtn() });
+    }
```
droideck: I think we don't need this in LMDB part

droideck (changes requested):

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
jchapma: I agree, it looks way better

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
jchapma: Ok, the button change is a copy and paste error. 

Regarding the dbhome setting, when I try changing the db home directory with bdb, an unwilling to perform message is returned, in the error logs it mentions this config cannot be changed when the server is running. IIUC then do we need a config option for changing the db home in the UI at all? 

WDYT

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
jchapma: You are correct, removing this

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
jchapma: I like that idea, let me look at how to determine the system page size and make the max size a multiple of it

jchapma:
> > Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation I think it is dead code, neither the entries nor the index are cached in lmdb import ) so IMHO it will be better to simply remove that tab
> 
> I think the import cache is used for other things that are not DB related (regardless of BDB/MDB) - so it still might be needed. Just proceed with caution :-)

Let me look into this, cautiously...

src/cockpit/389-console/src/lib/database/databaseConfig.jsx
vashirov: 1MB contains 256 4K pages or 64 16K pages or 16 64K pages. So if we change the maximum size to use megabytes, it'd cover all cases.

progier389:
> > Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation I think it is dead code, neither the entries nor the index are cached in lmdb import ) so IMHO it will be better to simply remove that tab
> 
> I think the import cache is used for other things that are not DB related (regardless of BDB/MDB) - so it still might be needed. Just proceed with caution :-)

Apparently no, according to 'grep -ri import.cache'  it is only used by bdb:
  Once the config handling removed, you found reference in db-bdb code (and only one in db-mdb but only to set job_index_buffer_size which is itself no more used)
  
 And I do not think that removing a now obscure parameter from cockpit is really dangerous (people can still view/change it through dsconf anyway), I am afraid that if we left it, some people may ask questions about it and we will be in trouble to answer ...

jchapma:
> > > Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation I think it is dead code, neither the entries nor the index are cached in lmdb import ) so IMHO it will be better to simply remove that tab
> > 
> > 
> > I think the import cache is used for other things that are not DB related (regardless of BDB/MDB) - so it still might be needed. Just proceed with caution :-)
> 
> Apparently no, according to 'grep -ri import.cache' it is only used by bdb: Once the config handling removed, you found reference in db-bdb code (and only one in db-mdb but only to set job_index_buffer_size which is itself no more used)
> 
> And I do not think that removing a now obscure parameter from cockpit is really dangerous (people can still view/change it through dsconf anyway), I am afraid that if we left it, some people may ask questions about it and we will be in trouble to answer ...

Thanks for the clarification

mreynolds389:
> > > Not sure we want to configure nsslapd-import-cachesize in lmdb case (Although it is still used in some computation I think it is dead code, neither the entries nor the index are cached in lmdb import ) so IMHO it will be better to simply remove that tab
> > 
> > 
> > I think the import cache is used for other things that are not DB related (regardless of BDB/MDB) - so it still might be needed. Just proceed with caution :-)
> 
> Apparently no, according to 'grep -ri import.cache' it is only used by bdb: Once the config handling removed, you found reference in db-bdb code (and only one in db-mdb but only to set job_index_buffer_size which is itself no more used)

Right, the indexing buffer.  I knew it wasn't an internal DB cache, but forgot exactly how it was used.  Ok if it's not used by MBM then it's safe to drop.  Thanks!!

droideck (approved):
Looks good to me!
I think, @vashirov's ack is also needed, though. :)

vashirov (commented):
Database Maximum Size hint says the size is in bytes, but the form accepts megabytes now. 

If Database Max Readers set to one of the numbers within 1-25 range, it gets set to 26. Not sure if it's possible to exclude this range from setting in the first place, @droideck do you know?

Database Max DBs has a minimum value of 36 that can be set, but if set it to anything within 36-68 range, it gets reset to 69. Should 69 be the minimum then? I don't know where it comes from though.

jchapma:
> Database Maximum Size hint says the size is in bytes, but the form accepts megabytes now.
> 
> If Database Max Readers set to one of the numbers within 1-25 range, it gets set to 26. Not sure if it's possible to exclude this range from setting in the first place, @droideck do you know?
> 
> Database Max DBs has a minimum value of 36 that can be set, but if set it to anything within 36-68 range, it gets reset to 69. Should 69 be the minimum then? I don't know where it comes from though.

I will correct the first issue.

 I see the same behaviour when I run the associated  commands at the CLI level:
```
sudo dsconf ldapi://%2fvar%2frun%2fslapd-inst01.socket backend config set --mdb-max-readers=25
sudo dsconf ldapi://%2fvar%2frun%2fslapd-inst01.socket backend config get | grep readers
nsslapd-mdb-max-readers: 26
```

And the same for max dbs:
```
sudo dsconf ldapi://%2fvar%2frun%2fslapd-inst01.socket backend config get | grep "max-dbs"
nsslapd-mdb-max-dbs: 131
```

I suspect the issue is not at the UI level as it gets/sets values via dsconf over ldapi. Let me have a look into it.

droideck:
> Database Maximum Size hint says the size is in bytes, but the form accepts megabytes now.
> 
> If Database Max Readers set to one of the numbers within 1-25 range, it gets set to 26. Not sure if it's possible to exclude this range from setting in the first place, @droideck do you know?
> 
> Database Max DBs has a minimum value of 36 that can be set, but if set it to anything within 36-68 range, it gets reset to 69. Should 69 be the minimum then? I don't know where it comes from though.

Technically, yes, `<NumberInput>` allows to set `min` and `max` values. So, we can probably align them with the CLI output.

vashirov:
@droideck, does it allow to have a sparse range? I.e. allow to set 0 and then 26-512? If not, I'm fine with the current behaviour.

droideck:
> @droideck, does it allow to have a sparse range? I.e. allow to set 0 and then 26-512? If not, I'm fine with the current behaviour.

For that, I see two options:

1. We can just add a checkbox on the side that will force 0 (and it will disable the field when checked) which means this value is computed by the server.
2. We can modify the `<NumberInput>` callbacks to handle the sparse range. Currently, we call it like this:                                                 `onChange={(e) => { this.onConfigChange(e, "mdbmaxreaders", 0, 200) }}` And we can change its code any way we like.
    Same with `onMinus={() => { this.onMinusConfig("mdbmaxreaders") }}` and `onPlus={() => { this.onPlusConfig("mdbmaxreaders") }}`

jchapma:
I didn't realise there was a sparse range of values allowed for these mdb configs, I guess I could have fuzzed the CLI instead of pointing the finger at the layer underneath the UI, sorry @progier389 

I would prefer @droideck option 2, where we handle the logic in changeConfig(), it feels cleaner to me.
Thanks

vashirov (approved):
LGTM, thank you!

Issue #6296, opened 2024-08-13
Test `dirsrvtests/tests/suites/basic/basic_test.py::test_conn_limits` fails in main branch


vashirov:
**Issue Description**
Test `dirsrvtests/tests/suites/basic/basic_test.py::test_conn_limits` fails in main branch:

```
____________________ ERROR at setup of test_conn_limits[4] _____________________

request = <SubRequest 'dscreate_with_numlistener' for <Function test_conn_limits[4]>>
dscreate_custom_instance = <dirsrvtests.tests.suites.basic.basic_test.CustomSetup object at 0x7f98b8411eb0>

    @pytest.fixture(scope="module", params=set(range(1,5)))
    def dscreate_with_numlistener(request, dscreate_custom_instance):
        numlisteners = request.param
        dscreate_custom_instance.create_wrapper(maxfds=MAX_FDS)
        inst = dscreate_custom_instance.inst
        inst.stop()
        dse_ldif = DSEldif(inst)
        dse_ldif.replace('cn=config', 'nsslapd-numlisteners', str(numlisteners))
>       inst.start()

dirsrvtests/tests/suites/basic/basic_test.py:2395: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.12/site-packages/lib389/__init__.py:177: in inner
    return f(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <dirsrvtests.tests.suites.basic.basic_test.CustomSetup.CustomDirSrv object at 0x7f98b83c11c0>
timeout = 120, args = ()
tmp_env = environ({'PATH': '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'HOSTNAME': 'server.example.com', 'DI.../root', 'LC_CTYPE': 'C.UTF-8', 'PYTEST_CURRENT_TEST': 'tests/suites/basic/basic_test.py::test_conn_limits[4] (setup)'})
count = 119

    def start(self, timeout=120, *args):
        if self.status():
            return
        tmp_env = os.environ
        # Unset PYTHONPATH to avoid mixing old CLI tools and new lib389
[...]

PR #6300, opened 2024-08-15
Issue 6296 - basic_test.py::test_conn_limits fails in main branch

jchapma:
Description:
A CI test to validate connection management functionality by configuring an instance with a very small connection table fails, with the instance failing to start.

Fix description:
In a multi list connection table configuration, the connection table code has been updated to expand the connection table size to include a head for each list. This expanded size needs to be accounted for when we determine if we can accept new connections and on final check at server starup time.

Relates: https://github.com/389ds/389-ds-base/issues/6296

Reviewed by:


droideck (approved):
LGTM

tbordaz:
Hi @jchapma , does it apply to others branches than main ? I am see similar errors in 2.5 branch

PR #6308, opened 2024-08-22
Issue 6307 - Wrong set of entries returned for some search filters

tbordaz:
Bug description:
	When the server returns an entry to a search it
	checks both access and matching of the filter.
	When evaluating a '!' (NOT) logical expression the server,
	in a first phase evaluates ONLY the right to access the
	related component (and its subcomponents).
	Then in a second phase verifies the matching.
	If the related component is a OR, in the first phase it
	evaluates access AND matching, this even if the call was
        to evaluate only access.
	This result in incoherent results.

Fix description:
	Make sure that when the function vattr_test_filter_list_or
	is called to only check access, it does not evaluate the matching.

relates: # 6307

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/filter/filter_test.py
+++ b/dirsrvtests/tests/suites/filter/filter_test.py
@@ -10,12 +10,14 @@ import logging
 
 import pytest
 import time
+from ldap import SCOPE_SUBTREE
 from lib389.dirsrv_log import DirsrvAccessLog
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM

tbordaz:
Thanks @progier389  for the review, I will write a CI test before pushing it.

Issue #6312, opened 2024-08-28
In branch 2.5, healthcheck report an invalid warning regarding BDB deprecation


tbordaz:
**Issue Description**
During backport of LMDB fixes in branch 2.5, a test was erroneously backported.
BDB is still supported in 2.5

The commit are this [one](https://github.com/389ds/389-ds-base/commit/f48026f555628fafa538e29975b7cfdc8d1e26ce#diff-0b5030c27013744647254b2b3e798b5cac7135438425a4e39509e85e02109a44) and this [one](https://github.com/389ds/389-ds-base/commit/a7287a3daf8c97f3588ab97ec378f4aab3c06b13#diff-668fbad8fda81a5ffd009366e23f5faf373d6b7de853ce77b2eabb25389da047R80) 

**Package Version and Platform:**
Only 2.5

**Steps to Reproduce**
1. Install 2.5 instance
2. run 'dsctl instance healthcheck

**Expected results**
Should not contain the message "BDB is deprecated and should not be used as a backend."
DSBLE0006

PR #6313, opened 2024-08-28
Issue 6312 - In branch 2.5, healthcheck report an invalid warning reg…

tbordaz:
…arding BDB deprecation

Bug description:
	during healthcheck, _lint_backend_implementation checks that
	the instance is not running a BDB backend.
	This check only applies for instance after 3.0.0

Fix description:
	If the instance is newer than 3.0.0 the health check
	just returns

relates: #6312

Reviewed by:


droideck (approved):
Looks good!

Issue #6090, opened 2024-02-12
Fix dbscan man page


progier389:
**Issue Description**
dbscan man page must be re synchronized with dbscan binary options  

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version:  389-ds-base-3.0.0

**Steps to Reproduce**
1. Run: man dbscan
2. Run: dbscan -h
3. Check that all option described im the help is also in the man page

progier389:
dc952e22a..25e1d1688 main -> main     
a2e964925..f51cd0157  389-ds-base-3.0 -> 389-ds-base-3.0

progier389:
dc952e22a..25e1d1688  main -> main
   ab4a08f52..4790bfeb9  389-ds-base-3.0 -> 389-ds-base-3.0
   ce97827b5..2ce6458ed  389-ds-base-2.6 -> 389-ds-base-2.6
   662fb4dd1..f162b4660  389-ds-base-2.5 -> 389-ds-base-2.5

Issue #5609, opened 2023-01-17
dbscan crashes and deletes database file


mreynolds389:
**Issue Description**

dbscan is apprently broken and when it runs it deletes the index file it was looking at:

```
# dscreate from-file /data/dssetup.inf 
Starting installation ...
Validate installation settings ...
Create file system structures ...
Create self-signed certificate database ...
Perform SELinux labeling ...
Create database backend: dc=example,dc=com ...
Perform post-installation tasks ...
Completed installation for instance: slapd-localhost

[root@fedora var]# cd /var/lib/dirsrv/slapd-localhost/db/userroot/
[root@fedora userroot]# ls
aci.db         cn.db      entryrdn.db   id2entry.db    numsubordinates.db  parentid.db  uidnumber.db
ancestorid.db  DBVERSION  gidnumber.db  nsuniqueid.db  objectclass.db      uid.db

[root@fedora userroot]# dbscan -d /var/lib/dirsrv/slapd-localhost/db/userroot/  -f cn.db 
Segmentation fault (core dumped)

[root@fedora userroot]# ls
aci.db         __db.001  __db.003   entryrdn.db   id2entry.db    numsubordinates.db  parentid.db  uidnumber.db
ancestorid.db  __db.002  DBVERSION  gidnumber.db  nsuniqueid.db  objectclass.db      uid.db

[root@fedora userroot]# dbscan -d /var/lib/dirsrv/slapd-localhost/db/userroot/  -f id2entry.db 
Segmentation fault (core dumped)

[root@fedora userroot]# ls
aci.db         __db.001  __db.003   entryrdn.db   nsuniqueid.db       objectclass.db  uid.db
ancestorid.db  __db.002  DBVERSION  gidnumber.db  numsubordinates.db  parentid.db     uidnumber.db
```

Also noticed this on 380-ds-base-2.1

progier389:
Hum ! dbscan should not crash and should probably improve the database write and add an explicit option (maybe --do-it) for the options that modify the db.
but -d is the delete option (added because of lmdb support), IMHO you wanted to run:
 dbscan -L /var/lib/dirsrv/slapd-localhost/db/userroot/  -f cn.db

progier389:
The change of option for -d and adding --do-it for import and delete is handled by #6090 issue - closing this issue

PR #6315, opened 2024-09-02
Issue 6090 - Fix dbscan options and man pages

progier389:
dbscan -d option is dangerously confusing as it removes a database instance while in db_stat it identify the database 
(cf issue #5609 ).
This fix implements long options in dbscan, rename -d in --remove, and requires a new --do-it option for action that change the database content.
The fix should also align both the usage and the dbscan man page with the new set of options

Issue:  #6090 

### DIFF
--- a/dirsrvtests/tests/suites/clu/dbscan_test.py
+++ b/dirsrvtests/tests/suites/clu/dbscan_test.py
@@ -1,0 +1,253 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import os
+import pytest
+import re
+import subprocess
+import sys
+
+from lib389 import DirSrv
+from lib389._constants import PW_DM, DBSCAN
+from lib389.topologies import topology_m2 as topo_m2
+from difflib import context_diff
+
+pytestmark = pytest.mark.tier0
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+
+
+class CalledProcessUnexpectedReturnCode(subprocess.CalledProcessError):
+    def __init__(self, result, expected_rc):
+        super().__init__(cmd=result.args, returncode=result.returncode, output=result.stdout, stderr=result.stderr)
+        self.expected_rc = expected_rc
+        self.result = result
+
+    def __str__(self):
+        return f'Command {self.result.args} returned {self.result.returncode} instead of {self.expected_rc}'
+
+
+class DbscanPaths:
+    @staticmethod
+    def list_instances(inst, dblib, dbhome):
+        # compute db instance pathnames
+        instances = dbscan(['-D', dblib, '-L', dbhome], inst=inst).stdout
+        dbis = []
+        if dblib == 'bdb':
+            pattern = r'^ (.*) $'
+            prefix = f'{dbhome}/'
+        else:
+            pattern = r'^ (.*) flags:'
+            prefix = f''
+        for match in re.finditer(pattern, instances, flags=re.MULTILINE):
+            dbis.append(prefix+match.group(1))
+        return dbis
+
+    @staticmethod
+    def list_options(inst):
+        # compute supported options
+        options = []
+        usage = dbscan(['-h'], inst=inst, expected_rc=None).stdout
+        pattern = r'^\s+(?:(-[^-,]+), +)?(--[^ ]+).*$'
+        for match in re.finditer(pattern, usage, flags=re.MULTILINE):
+            for idx in range(1,3):
+                if not match.group(idx) is None:
+                    options.append(match.group(idx))
+        return options
+
+    def __init__(self, inst):
+        dblib = inst.get_db_lib()
+        dbhome = inst.ds_paths.db_home_dir
+        self.inst = inst
+        self.dblib = dblib
+        self.dbhome = dbhome
+        self.options = DbscanPaths.list_options(inst)
+        self.dbis = DbscanPaths.list_instances(inst, dblib, dbhome)
+        self.ldif_dir = inst.ds_paths.ldif_dir
+
+    def get_dbi(self, attr, backend='userroot'):
+        for dbi in self.dbis:
+            if f'{backend}/{attr}.'.lower() in dbi.lower():
+                return dbi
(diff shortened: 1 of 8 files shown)


progier389:
Fixed several issues because the testcase failed on bdb:
   - compute the right li_directory when opening the db
   - prevent race condition and double free with bdb (this cause a leak but it is in a dblayer function only called by dbscan ...
   - change the testcase to export/removeimport replication changelog instead of cn/ Add a limitation that only changelog and id2entry may be imported within dbscan (dbscan cannot open properly the db to support duplicate keys)

ldap/servers/slapd/back-ldbm/dbimpl.c
```diff
@@ -397,7 +397,44 @@ const char *dblayer_op2str(dbi_op_t op)
...
+    for (int i=0; i<3; i++) {
+        if (stat(lid, &sbuf) == 0) {
+            if (S_ISDIR(sbuf.st_mode)) {
+                return lid;
+            }
+            /* Non directory existing file could be regular
+             * at the first iteration otherwise it is an error.
+             */
+            if (i>0 || !S_ISREG(sbuf.st_mode)) {
+                break;
```
tbordaz, line 423: I am missing how "break;" goes to error handling ?

tbordaz (approved):
A minor question else LGTM

ldap/servers/slapd/back-ldbm/dbimpl.c
progier389: Originally there was a comment explaining that in error case the function returns a copy of the original string, since it is not a directory, there is a failure later on when trying to open the database
bit the comment get removed when I reorganized the code! :wink:
I will add it again. Anyway there iwill be a new commit because there s a regression about is_dbi_supported

dirsrvtests/tests/suites/clu/dbscan_test.py
```diff
@@ -0,0 +1,253 @@
...
+#
+import logging
+import os
+import pytest
+import re
+import subprocess
+import sys
+
+from lib389 import DirSrv
+from lib389._constants import PW_DM, DBSCAN
```
droideck: PW_DM is unused, I think?

dirsrvtests/tests/suites/clu/dbscan_test.py
```diff
@@ -0,0 +1,253 @@
...
+def dbscan(args, inst=None, expected_rc=0):
+    if inst is None:
+        prefix = os.environ.get('PREFIX', "")
+        prog = f'{prefix}/bin/dbscan'
+    else:
+        prog = os.path.join(inst.ds_paths.bin_dir, DBSCAN)
+    args.insert(0, prog)
+    output = subprocess.run(args, encoding='utf-8', stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    log.debug(f'{args} result is {output.returncode} output is {output.stdout}')
+    if expected_rc != output.returncode and not expected_rc is None:
```
droideck:
```suggestion
    if expected_rc != output.returncode and expected_rc is not None:
```

droideck: This is a minor nitpick, but I think `is not None` is better readable. `not expected_rc` is also viable, so `not expected_rc is None:` can be misread accidentally by humans. In this case, `is not None` is straightforward.

Feel free to leave it as it is if it's your style. It compiles to the same code anyway.

dirsrvtests/tests/suites/clu/dbscan_test.py
```diff
@@ -0,0 +1,253 @@
...
+    curdbis = DbscanPaths.list_instances(paths.inst, paths.dblib, paths.dbhome)
+    assert not cndbi in curdbis
+
+    # Run dbscan -I import_file ... --do-it
+    result = dbscan(['-D', paths.dblib, '-f', cndbi,
+                     '-I', export_cn, '--do-it'],
+                    inst=paths.inst, expected_rc=0)
+
+    # Check the error message about missing --do-it
+    assert not expected_msg in result.stdout
```
droideck: Same here
```suggestion
    assert expected_msg not in result.stdout
```

dirsrvtests/tests/suites/clu/dbscan_test.py
```diff
@@ -0,0 +1,253 @@
...
+    # Check that cn instance is still present
+    curdbis = DbscanPaths.list_instances(paths.inst, paths.dblib, paths.dbhome)
+    assert cndbi in curdbis
+
+    # Run dbscan --remove ... --doit
+    result = dbscan(['-D', paths.dblib, '--remove', '-f', cndbi, '--do-it'],
+                    inst=paths.inst, expected_rc=0)
+
+    # Check the error message about missing --do-it
+    assert not expected_msg in result.stdout
```
droideck: And here
```suggestion
    assert expected_msg not in result.stdout
```

ldap/servers/slapd/back-ldbm/dbimpl.c
```diff
@@ -397,7 +397,48 @@ const char *dblayer_op2str(dbi_op_t op)
...
+{
+    /*
+     * li_directory is an existing directory.
+     * it can be fname or its parent or its greatparent
+     * in case of problem returns the provided name
+     */
+    char *lid = slapi_ch_strdup(fname);
+    struct stat sbuf = {0};
+    char *pt = NULL;
+    for (int i=0; i<3; i++) {
```
droideck: Minor `size_t` for array indexing issue

ldap/servers/slapd/tools/dbscan.c
```diff
@@ -130,9 +133,43 @@ long ind_cnt = 0;
...
 static Slapi_Backend *be = NULL; /* Pseudo backend used to interact with db */
 
+/* For Long options without shortcuts */
+enum {
+    OPT_FIRST = 0x1000,
+    OPT_DO_IT,
+    OPT_REMOVE,
+};
+
+static struct option options[] = {
```
droideck: Another minor nitpick, probably we can make it `static const struct`

dirsrvtests/tests/suites/clu/dbscan_test.py
droideck: Also, probably, `expected_rc is not None` should be checked first
```suggestion
    if expected_rc is not None and expected_rc != output.returncode:
```

droideck (commented):
All really minor issues

ldap/servers/slapd/back-ldbm/dbimpl.c
progier389: But in this case i is not an array index ..., 
It is a count to know how many time we try to look for parent directories 
==> will keep int but rename i as count for better clarity

PR #6317, opened 2024-09-04
Issue 6188 automation - Check if nsslapd-haproxy-trusted-ip attribute is returned in default schema

ssidhaye:
Issue 6188 : Check if nsslapd-haproxy-trusted-ip
    attribute is returned in default schema
    
    Relates : https://github.com/389ds/389-ds-base/issues/6188
    
    Reviewed by: vashirov (Thanks)

### DIFF
--- a/dirsrvtests/tests/suites/basic/haproxy_test.py
+++ b/dirsrvtests/tests/suites/basic/haproxy_test.py
@@ -60,6 +60,8 @@ def test_haproxy_trust_ip_attribute(topo, setup_test):
 
     log.info("Check that nsslapd-haproxy-trusted-ip attribute is present")
     assert topo.standalone.config.present('nsslapd-haproxy-trusted-ip', '192.168.0.1')
+    # Check nsslapd-haproxy-trusted-ip attribute is present in schema
+    assert topo.standalone.schema.query_attributetype('nsslapd-haproxy-trusted-ip') is not None
 
     log.info("Delete nsslapd-haproxy-trusted-ip attribute")
     topo.standalone.config.remove_all('nsslapd-haproxy-trusted-ip')


dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -244,3 +245,28 @@ def test_check_replica_id_accepted_hub_consumer(topology_st, instance_role, repl
     log.info(f'output message : {msg}')
     assert "Replication successfully enabled for" in msg
     log.info(f"Test PASSED: --replica-id option is accepted for {instance_role}.")
+
+
+def test_def_rootdse_attr(topology_st):
+    """Test to check if nsslapd-haproxy-trusted-ip attribute is present in schema by default
+
+    :id: e8aca6d7-3c6f-47e6-ad6c-86fa223ddbac
+    :parameterized: no
```
bsimonova: It should be 'parametrized'. 
Or you can leave out that value, as it is not required when parametrize is not used in the test.

dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -244,3 +245,28 @@ def test_check_replica_id_accepted_hub_consumer(topology_st, instance_role, repl
...
+    :expectedresluts:
+        1. ldapsearch should return the attribute
+        2. The attribute should be present in default schema
+    """
+    topology_st.standalone.start()
+
+    log.info("Assert default schema has nsslapd-haproxy-trusted-ip attribute")
+    try:
+        haproxy = topology_st.standalone.schema.query_attributetype('nsslapd-haproxy-trusted-ip')
+        assert haproxy[0].names[0] == 'nsslapd-haproxy-trusted-ip'
```
bsimonova: We could check for syntax 1.3.6.1.4.1.1466.115.121.1.15 as well.

dirsrvtests/tests/suites/clu/dsconf_test.py
```diff
@@ -244,3 +245,28 @@ def test_check_replica_id_accepted_hub_consumer(topology_st, instance_role, repl
...
+    :expectedresults:
+        1. ldapsearch should return the attribute
+        2. The attribute should be present in default schema
+    """
+    topology_st.standalone.start()
+
+    log.info("Assert default schema has nsslapd-haproxy-trusted-ip attribute")
+    try:
+        haproxy = topology_st.standalone.schema.query_attributetype('nsslapd-haproxy-trusted-ip')
+        assert haproxy[0].names[0] == 'nsslapd-haproxy-trusted-ip'
```
vashirov: I think this check should be added to the existing [test_haproxy_trust_ip_attribute](https://github.com/389ds/389-ds-base/blob/a9c7ff9419a15508071a9ad43e4c7f0469938506/dirsrvtests/tests/suites/basic/haproxy_test.py#L46) test.

Is there a reason to use `assert` in `try` block and then raise an Exception in `except` block? `assert` raises `AssertionError`, I believe it should be enough to fail the test.

`schema.query_attributetype()` returns `None` if the attribute is not found, so we can assert that that it's not `None` instead.

ssidhaye: Good point. I'll move this check to the above test. Ah yes, AssertionError is enough here. Fixed it

bsimonova (dismissed):

vashirov (approved):

PR #6318, opened 2024-09-05
Issue 6316 - lmdb reindex is broken if index type is specified

### LINKED ISSUE #6316: lmdb reindex is broken if index type is specified
**Issue Description**
Some freeipa tests are now failing because a reindex is not properly done although a reindex task is performed.
The issue is related to the fact that the task contains subtype - 
Looks like it was a regression in some of the vlv fixes
 Anyway the index type should always be ignored in tasks because reindexing only some specific types has no sense (because reindex recreate the db instance from scratch anyway ...)  

 - Platform: . Fedora]
 - Package:  2.5 and 3.x  branches
 - Browser: N/A

**Steps to Reproduce**
1. Setup an instance using lmdb with default sample entries
2. Add description attribute in uid=demo_user, entry (or any other entry)
3. Add description index for equality and presence WITHOUT reindexing
4. Restart the server
5. Add description attribute to the suffix entry
6. Run reindex task with nsIndexAttribute: description:eq,pres
7. Run dbscan -f  /var/lib/dirsrv/slapd-standalone1/db/userroot/description.db -A 
8. See that there is only 1 key starting with = (the one for the suffix entry)

**Expected results**
Should have equality keys =value for both the suffix and the demo_user entry

**Additional context**
Solution is to ignore the type when handling reindex task

Triaged: https://issues.redhat.com/browse/IDMDS-4641

progier389:
While reindexing using task or offline reindex, if the attribute name contains the index type (for example :eq,pres)
Then the attribute is not reindexed. Problem occurs when lmdb is used, things are working fine with bdb. 
Solution: strip the index type in reindex as it is done in bdb case.
Anyway the reindex design requires that for a given attribute all the configured index types must be rebuild.

Issue: #6316 

Reviewed by: @tbordaz, @droideck   (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/indexes/regression_test.py
+++ b/dirsrvtests/tests/suites/indexes/regression_test.py
@@ -10,6 +10,9 @@ import time
 import os
 import pytest
 import ldap
+import logging
+import glob
+import re
 from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
 from lib389.backend import Backend, Backends, DatabaseConfig
 from lib389.cos import  CosClassicDefinition, CosClassicDefinitions, CosTemplate
@@ -31,6 +34,8 @@ SUFFIX2 = 'dc=example2,dc=com'
 BENAME2 = 'be2'
 
 DEBUGGING = os.getenv("DEBUGGING", default=False)
+logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
 
 
 @pytest.fixture(scope="function")
@@ -83,6 +88,7 @@ def add_a_group_with_users(request, topo):
             'cn': USER_NAME,
             'uidNumber': f'{num}',
             'gidNumber': f'{num}',
+            'description': f'Description for {USER_NAME}',
             'homeDirectory': f'/home/{USER_NAME}'
         })
         users_list.append(user)
@@ -95,9 +101,10 @@ def add_a_group_with_users(request, topo):
         # If the server crashed, start it again to do the cleanup
         if not topo.standalone.status():
             topo.standalone.start()
-        for user in users_list:
-            user.delete()
-        group.delete()
+        if not DEBUGGING:
+            for user in users_list:
+                user.delete()
+            group.delete()
 
     request.addfinalizer(fin)
 
@@ -123,6 +130,38 @@ def set_small_idlistscanlimit(request, topo):
         topo.standalone.restart()
 
     request.addfinalizer(fin)
+
+
+@pytest.fixture(scope="function")
+def set_description_index(request, topo, add_a_group_with_users):
+    """
+    Set some description values and description index without reindexing.
+    """
+    inst = topo.standalone
+    backends = Backends(inst)
+    backend = backends.get(DEFAULT_BENAME)
+    indexes = backend.get_indexes()
+    attr = 'description'
+
+    def fin(always=False):
+        if always or not DEBUGGING:
+            try:
+                idx = indexes.get(attr)
+                idx.delete()
+            except ldap.NO_SUCH_OBJECT:
+                pass
+
+    request.addfinalizer(fin)
+    fin(always=True)
+    index = indexes.create(properties={
+        'cn': attr,
+        'nsSystemIndex': 'false',
+        'nsIndexType': ['eq', 'pres', 'sub']
+        })
+    # Restart needed with lmdb (to open the dbi handle)
+    inst.restart()
+    return (indexes, attr)
+
 
 #unstable or unstatus tests, skipped for now
 @pytest.mark.flaky(max_runs=2, min_passes=1)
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -346,6 +385,102 @@ def test_task_status(topo):
...
+    log.info(f"count_keys found {count} keys starting with '{prefix}' in {indexfile}")
+    return count
+
+
+def test_reindex_task_with_type(topo, set_description_index):
+    """Check that reindex task works as expected when index type is specified.
+
+    :id: 0c7f2fda-69f6-11ef-9eb8-083a88554478
+    :setup: Standalone instance
+             - with 100 users havibg description attribute
```
tbordaz: typo: havibg

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -346,6 +385,102 @@ def test_task_status(topo):
...
+        suffix=DEFAULT_SUFFIX,
+        attrname=f'{attr}:eq,pres',
+        args={TASK_WAIT: True}
+    )
+    reindex_task = Task(topo.standalone, tasks.dn)
+    assert reindex_task.status()
+    assert reindex_task.get_exit_code() == 0
+
+    keys2e = count_keys(inst, DEFAULT_BENAME, attr, prefix='=')
+    keys2p = count_keys(inst, DEFAULT_BENAME, attr, prefix='+') > 0
```
tbordaz: Sorry for the dummy question do you expect a count or a boolean ?

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -346,6 +385,102 @@ def test_task_status(topo):
...
+    reindex_task = Task(topo.standalone, tasks.dn)
+    assert reindex_task.status()
+    assert reindex_task.get_exit_code() == 0
+
+    keys3e = count_keys(inst, DEFAULT_BENAME, attr, prefix='=')
+    keys3p = count_keys(inst, DEFAULT_BENAME, attr, prefix='+') > 0
+    keys3s = count_keys(inst, DEFAULT_BENAME, attr, prefix='*') > 0
+    assert keys3e == keys2e
+    assert keys3p == 1
+    assert keys3s == keys2s
```
tbordaz, line 481: Weird. In keys2s I would expect no key as it was not indexed in substring. In keys3s I would expect some keys. In short how keys3s == keys2s ?

tbordaz (approved):
a typo and some questions.
Also is it related to the failure https://github.com/dogtagpki/pki/pull/4839 ?

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: a count

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: There are two lists of types:
  1 - The configured one in the index entry (i.e:  eq,pres,sub (cf set_description_index fixture))
  2 - The ones specified when re indexing.   (i.e: eq,pres in this case)
The test verify that the CONFIGURED list is systematically used when reindexing.

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: So the > 0 should not be there ...

progier389:
@tbordaz  It is the issue with discussed in https://redhat-internal.slack.com/archives/C04NTP68ZQB/p1724853098624349
Since you mentionned the same search request in both case, I think 
that https://github.com/dogtagpki/pki/pull/4839 is the same problem.

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -31,6 +34,8 @@
 BENAME2 = 'be2'
 
 DEBUGGING = os.getenv("DEBUGGING", default=False)
+logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
```
droideck, line 38: It's added but not used. It might be worth rewriting some code comments to be log.info messages. :)

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -124,6 +131,38 @@ def fin():
...
+        if always or not DEBUGGING:
+            try:
+                idx = indexes.get(attr)
+                idx.delete()
+            except ldap.NO_SUCH_OBJECT:
+                pass
+
+    request.addfinalizer(fin)
+    fin(always=True)
+    index = indexes.create(properties={
```
droideck, line 156: It might be worth using `ensure_state` instead of `create` here.
https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/_mapped_object.py#L1065

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -346,6 +385,102 @@ def test_task_status(topo):
...
+    """
+    (indexes, attr) = set_description_index
+    inst = topo.standalone
+    if not inst.is_dbi_supported():
+        pytest.skip('This test requires that dbscan supports -A option')
+    # modify indexed value
+    Domain(inst, DEFAULT_SUFFIX).replace(attr, f'test_before_reindex')
+
+    keys1 = count_keys(inst, DEFAULT_BENAME, attr, prefix='=')
+    assert keys1 <= 1
```
droideck, line 442: > Should be either no key (bdb) or a single one (lmdb)

Should we check exactly the value here for BDB and LMDB cases? (instead of just asserting `<= 1`)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -1150,14 +1150,22 @@ process_db2index_attrs(Slapi_PBlock *pb, ImportCtx_t *ctx)
...
+    char *pt = NULL;
     int i;
 
     slapi_pblock_get(pb, SLAPI_DB2INDEX_ATTRS, &attrs);
     /* Should perhaps add some tests to check that indexes are valid */
     for (i = 0; attrs && attrs[i]; i++) {
         switch (attrs[i][0]) {
         case 't': /* attribute type to index */
-            slapi_ch_array_add(&ctx->indexAttrs, slapi_ch_strdup(attrs[i] + 1));
+            attrname = slapi_ch_strdup(attrs[i] + 1);
```
droideck, line 1162: I wonder if static analysers will scream here for the case where `attrname` might be NULL after `slapi_ch_strdup`. We have the null check in `slapi_ch_strdup`, so probably not... (if they are smart enough)

droideck (approved):
A couple of minor polishing moments; otherwise - Looks Good!

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
progier389: slapi-plugin.h:char *slapi_ch_strdup(const char *s) __ATTRIBUTE__((returns_nonnull));
So it is explicitly said in the header.

Most static analyser tools handles the attributes and anyway those that does not spawn so much false positive that they are mostly useless
Coverity scan is smart enough, by I noticed that sometime adding useless test for NULL returns tends to increase the number of report of possible NULL dereference because it also considers that if return is tested for NULL, that means that the function may returns NULL. (in short you should either test everywhere or nowhere ...)

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: Maybe you have looked at an old commit:
log.info is used in count_keys and test_reindex_task_with_type

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: Not in this specific case:
    We have to explicitly delete the index if it exists (and is indexed) or step 5 will assert
     (because the index is already indexed while the test expect it is not the case)

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: We do not really care about the exact number. 
  The idea is to check that the index is not fully indexed (i.e that the 100 users description keys are not present )
 Furthermore we may decide later to change mdb behavior to also have 0 keys ( I do not like the fact that partially indexed database get usable after a restart) and we should probably change it when we find the time  )

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -346,6 +385,102 @@ def test_task_status(topo):
...
+    assert reindex_task.get_exit_code() == 0
+
+    keys2e = count_keys(inst, DEFAULT_BENAME, attr, prefix='=')
+    keys2p = count_keys(inst, DEFAULT_BENAME, attr, prefix='+')
+    keys2s = count_keys(inst, DEFAULT_BENAME, attr, prefix='*')
+    assert keys2e > keys1
+    assert keys2p > 0
+    assert keys2s > 0
+
+    # Reindex 'cn' without types
```
tbordaz: minor: this is "attr" not " 'cn'"

dirsrvtests/tests/suites/indexes/regression_test.py
droideck: Okay, I just wanted a bit more coverage when possible. But your point is valid. Sounds good to me!

PR #6320, opened 2024-09-05
Issue 6319 - bdb subpackage has `%description` in the wrong place

### LINKED ISSUE #6319: bdb subpackage has `%description` in the wrong place
**Issue Description**
rpminspect reports issues with 389-ds-base spec file:

> Subpackage 389-ds-base-bdb on aarch64 carries 'Requires: libback-ldbm.so()(64bit)' which comes from subpackage 389-ds-base but does not carry an explicit package version requirement. Please add 'Requires: 389-ds-base = %{version}-%{release}' to the spec file to avoid the need to test interoperability between various combinations of old and new subpackages.

Currently, `%description` section comes before `Requires` and `Provides`. This is a problem for rpmbuild because it is still reading lines for the description when it encounters the `Requires` and `Provides` lines. So those never end up in the subpackage RPM header which causes the rpminspect error. 

The `%description` section needs to come after the `Requires` and `Provides` lines.

Related: https://github.com/rpminspect/rpminspect/issues/1427

vashirov:
Bug Description:
Currently, `%description` section comes before `Requires` and `Provides`. This is a problem for rpmbuild because it is still reading lines for the description when it encounters the `Requires` and `Provides` lines. So those never end up in the subpackage RPM header which causes the rpminspect error.

Fix Description:
The %description section needs to come after the Requires and Provides lines.

Fixes: https://github.com/389ds/389-ds-base/issues/6319
Relates: https://github.com/rpminspect/rpminspect/issues/1427

Reviewed by:

### DIFF
--- a/rpm/389-ds-base.spec.in
+++ b/rpm/389-ds-base.spec.in
@@ -284,19 +284,20 @@ SNMP Agent for the 389 Directory Server base package.
 %if %{with bundle_libdb}
 %package          bdb
 Summary:          Berkeley Database backend for 389 Directory Server
-%description      bdb
-Berkeley Database backend for 389 Directory Server
-Warning! This backend is deprecated in favor of lmdb and its support
-may be removed in future versions.
 
 Requires:         %{name} = %{version}-%{release}
+Requires:         %{name}-libs = %{version}-%{release}
 # Berkeley DB database libdb was marked as deprecated since F40:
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

droideck (approved):
Looks good!
probably, also makes sense to mention in the commit message that `Requires:         %{name}-libs = %{version}-%{release}` was added as it was missing in the specfile.

PR #6322, opened 2024-09-06
Issue 6321 - lib389 get_db_lib function may returns the wrong db type

### LINKED ISSUE #6321: lib389 get_db_lib function may returns the wrong db type
**Issue Description**
lib389 get_db_lib function may returns the default database type instead of the configured one.

 - Platform: CentOS 9
 - Package and version: 389-ds-base-2.5.2-1.el9
 - Browser: N/A

**Steps to Reproduce**
1. Setup an lmdb instance
2. Run dsctl instance healthcheck 
3. See DSBLE0006 and DSBLE0005 errors

**Expected results**
Healthcheck should not display any errors.

**Additional context**
Problem is that  `from backend import DatabaseConfig` generate an ImportError so the call to DatabaseConfig(self).get_db_lib() is bypassed.
It should rather be:`from lib389.backend import DatabaseConfig`

progier389:
get_db_lib returns the default db type instead of the configured one because of  caught exceptions
Fix the import issue if instance is online
Fix the DSEldif parameter for the offline case

Issue: #6321 

### DIFF
--- a/dirsrvtests/tests/suites/lib389/config_compare_test.py
+++ b/dirsrvtests/tests/suites/lib389/config_compare_test.py
@@ -16,6 +16,10 @@ from lib389.dseldif import DSEldif
 
 pytestmark = pytest.mark.tier1
 
+# set_db_type_and_state fixture parameters
+db_types_and_states = [('bdb,stopped'), ('mdb,stopped'), ('bdb,started'), ('mdb,started')]
+
+
 def test_config_compare(topology_i2):
     """
     Compare test between cn=config of two different Directory Server intance.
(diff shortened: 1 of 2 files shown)


progier389:
Added a test case and realized that there was also a problem in the off line case.

progier389:
Hum, 
At firts, have I chosen to havoid parametrization to avoid too many restart by I realized that I have to test both offline and online results so the server is restarted at every steps anyway.  I think I should better use a parmetrized fixture to set the db type in dse.ldif and the server state.

tbordaz (approved):
LGTM

progier389:
Commited new version with parametrized test (which is indeed simpler)

PR #6325, opened 2024-09-12
Issue 6324 - Provide more information in the error message during setup_ol_tls_conn()

### LINKED ISSUE #6324: Provide more information in the error message during setup_ol_tls_conn()
**Issue Description**
When there's a problem with creating a new TLS context, we just fail with -1 error code.
We can improve it by providing more information that can be extracted from LDAP_OPT_DIAGNOSTIC_MESSAGE.

droideck:
Description: When there's a problem with creating a new TLS context, we just fail with -1 error code. We can improve it by providing more information that can be extracted from LDAP_OPT_DIAGNOSTIC_MESSAGE or LDAP_OPT_ERROR_STRING.

Use slapi_ldap_get_lderrno to get more information from the ld structure. Add a test case.

Fixes: https://github.com/389ds/389-ds-base/issues/6324

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/tls/tls_repl_clientauth_test.py
+++ b/dirsrvtests/tests/suites/tls/tls_repl_clientauth_test.py
@@ -107,12 +107,15 @@ def test_new_tls_context_error(topo_tls_ldapi):
     if not ca_files:
         pytest.skip("No Self-Signed-CA.pem files found. Skipping test.")
 
+    log.info(f'Found {len(ca_files)} Self-Signed-CA.pem files')
+    log.info(f'CA files are: {", ".join(ca_files)}')
+
     for ca_file in ca_files:
         try:
             os.remove(ca_file)
             log.info(f"Removed file: {ca_file}")
         except OSError as e:
-            log.warning(f"Error removing file {ca_file}: {e}")
+            log.info(f"Error removing file {ca_file}: {e}")
 
     log.info('Reinit agreement')
     replica_m1 = Replicas(m1).get(DEFAULT_SUFFIX)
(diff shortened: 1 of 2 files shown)


droideck:
I'm not sure why but the test passes on my machine (and on a virtual machine). But it can't find the error message in GitHub CI.
Probably, it's because we have Docker here, and the issue is triggered when certificate is removed from systems tmp.

progier389:
At least the C code looks good. 
About the failing test test I find a bit surprising that we are not seeing any log about the deleted cert files:
cacert list should not be empty of the test would be skipped. And neither successful deletion or failed one are logged.
So I wonder if the delete operation has not failed and our logger does ignore the warning level 
Maybe you should rather use log.info (as we know that info are logged!) 
And probably worth to log,info(f'ca_files={ca_files}') to help debugging the testcase

droideck:
Okay, the issue was that I tried to grab the wrong error.

The error I need to check for is: `error:80000002:system library::No such file or directory`. It is the error that happens during the TLS context set when the cert is missing.

But there's another error that happens earlier in the function. And they are different between GitHub CI and my virtual machine mock run.

GitHub CI - `error:0200008A:rsa routines::invalid padding`
My env - `error:0A000086:SSL routines::certificate verify failed`

Why it's different, it's another story I am not sure still.

vashirov:
What version of openssl do you have? In the CI image I see openssl-3.2.2-3.fc40

droideck:
> What version of openssl do you have? In the CI image I see openssl-3.2.2-3.fc40

It's the same on my machine - openssl-3.2.2-3.fc40.x86_64

progier389 (approved):
LGTM

PR #6332, opened 2024-09-20
Issue 6331 - UI - Instance fails to load when DB backup directory doesn't exist

### LINKED ISSUE #6331: UI - Instance fails to load when DB backup directory doesn't exist
**Issue Description**
RHDS Console will fail to load an instance if the DB backup directory ( nsslapd-bakdir ) is set to a non-existing directory.
The Console will show a spinner and won't allow to select any other RHDS instances.

**Steps to Reproduce**
1. Change back up directory to non-existing one.
```
$ dsconf -D "cn=Directory Manager" ldap://localhost:38901 config replace nsslapd-bakdir=/DOES_NOT_EXIST
Enter password for cn=Directory Manager on ldap://localhost:38901:
Successfully replaced "nsslapd-bakdir"
```
2. Try to load RHDS instance.
3. Progress bar goes to 100% but instance won't load.

**Expected results**
Instance is loaded.

droideck:
Description: RHDS Console will fail to load an instance if the DB backup directory ( nsslapd-bakdir ) is set to a non-existing directory. The Console will show a spinner and won't allow to select any other RHDS instances.

Correctly process 'failed' case, update the progress bar and send a notification. Also, validate all settings tabs after the load, so the error is visible right away.

Fixes: https://github.com/389ds/389-ds-base/issues/6331

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/ds.jsx
+++ b/src/cockpit/389-console/src/ds.jsx
@@ -445,30 +445,42 @@ export class DSInstance extends React.Component {
     loadBackups() {
         let cmd = ["dsctl", "-j", this.state.serverId, "backups"];
         log_cmd("loadBackups", "Load Backups", cmd);
-        cockpit.spawn(cmd, { superuser: true, err: "message" }).done(content => {
-            this.updateProgress(25);
-            const config = JSON.parse(content);
-            const rows = [];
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM

PR #6333, opened 2024-09-22
Fix duplicate detection logic in slapi_str2charray_ext by ensuring exact match on string length

### LINKED ISSUE #6338: Fix multi valued parameters configuration code
**Issue Description**
In slapi_str2charray_ext function the duplicate value detection is flawed so some parameters values may be wrongly ignored
(This impacts the configuration parameters that accepts several values separated by commas )
 See PR #6333 

 - Platform: Any
 - Package and version: All

PavlNekrasov:
Updated the condition for checking duplicates to ensure that the lengths of the candidate and original strings are equal. This prevents false positives by confirming that the strings are identical at the start and match in length, enforcing an exact duplicate check.

Found by Linux Verification Center (linuxtesting.org) with SVACE.
Reporter: Pavel Nekrasov ([p.nekrasov@fobos-nt.ru](mailto:p.nekrasov@fobos-nt.ru)).

issue: #6338 

Reviewed by: @progier

### DIFF
--- a/ldap/servers/slapd/charray.c
+++ b/ldap/servers/slapd/charray.c
@@ -354,8 +354,9 @@ slapi_str2charray_ext(char *str, char *brkstr, int allow_dups)
         /* Always copy the first value into the array */
         if ((!allow_dups) && (i != 0)) {
             /* Check for duplicates */
+            size_t len_s = strlen(s);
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

PR #6335, opened 2024-09-23
fix buffer owerflow in the function `parseVariant`

DmitriyFedin:
`variant[0]` can have a maximum value of 72. `field->var = variant0 - VAR_MIN;`, so  a maximum value `field->var = 7`

This is a buffer overflow in the line
 `obj->var[field->var] = (char *)malloc(MAX_FILTER)`

if it accesses the array at index 7, an overflow will occur.

Found by Linux Verification Center (linuxtesting.org) with SVACE.
Reporter: Dmitriy Fedin ([d.fedin@fobos-nt.ru](mailto:d.fedin@fobos-nt.ru)).

### DIFF
--- a/ldap/servers/slapd/tools/ldclt/parser.c
+++ b/ldap/servers/slapd/tools/ldclt/parser.c
@@ -125,7 +125,7 @@ parseVariant(
    * Maybe a variable ?
    */
     if (variant[1] == '\0') {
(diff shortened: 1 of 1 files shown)


progier389 (changes requested):
Sorry but I do not think that this fix is correct:
As the error message said, variable are expected to be within between VAR_MIN and VAR_MAX. 
If there is a buffer overflow that is the array size that must be changed rather than limiting the variable range ...

Issue #6064, opened 2024-02-01
bdb2mdb shows errors


vashirov:
**Issue Description**
On an instance without replication set up, bdb2mdb shows errors about missing `replication_changelog.db`:

```
[root@fedora ~]# dsctl localhost dblib bdb2mdb 
cleanup dbmapdir=/var/lib/dirsrv/slapd-localhost/db dbhome=/dev/shm/slapd-localhost dblib=bdb
Required space for LDIF files is about 24.0 KB
Required space for DBMAP files is about 240.0 KB
Required number of dbi is 128
Backends exportation 0.000000% (userroot)
ldiffile: /var/lib/dirsrv/slapd-localhost/ldif/__dblib-userroot.ldif
run_dbscan failed: cmd is ['/bin/dbscan', '-D', 'bdb', '-f', '/var/lib/dirsrv/slapd-localhost/db/userroot/replication_changelog.db', '-X', '/var/lib/dirsrv/slapd-localhost/ldif/__dblib-userroot.cl5.dbtxt'] return code is 1 output is Can't initialize db plugin: bdb

Command '['/bin/dbscan', '-D', 'bdb', '-f', '/var/lib/dirsrv/slapd-localhost/db/userroot/replication_changelog.db', '-X', '/var/lib/dirsrv/slapd-localhost/ldif/__dblib-userroot.cl5.dbtxt']' returned non-zero exit status 1.
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/lib389/cli_ctl/dblib.py", line 152, in run_dbscan
    output = subprocess.check_output(args, encoding='utf-8', stderr=subprocess.STDOUT)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
[...]

PR #6341, opened 2024-09-24
Issue 6064 - bdb2mdb shows errors

progier389:
Fix several issues with dsctl dblib migration tools:
Fix some problem related to dbscan -D bdb --import:
 - random crash when closing  the database - Fixed by fully cleaning the bdb framework if open in read-write mode
 - random records were missing - Fixed by fully starting the bdb framework when it is open in read-write mode
   and by using txn
 Fixed error messages displayed when trying to reopen ldap connection on restarted instance by not reopening the connection that are not needed.
 Fixed error messages displayed when there is no replication changelog by detecting the presence of replication changelog before trying to export it.

Issue: #6064 

Reviewed by: @tbordaz (Thans!)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
@@ -6905,39 +6905,51 @@ bdb_public_private_open(backend *be, const char *db_filename, int rw, dbi_env_t 
     int flags;
     int rc;
 
-    /* Either filename is an existing regular file
-     *  or the "home" directory where txn logs are
-     */
+    slapi_ch_free_string(&conf->bdb_dbhome_directory);
+    if (li->li_directory == NULL) {
+        /* Either filename is an existing regular file
+         *  or the "home" directory where txn logs are
+         */
 
-    PL_strncpyz(dbhome, db_filename, MAXPATHLEN);
-    if (stat(dbhome, &st) == 0) {
-        if (S_ISDIR(st.st_mode)) {
-            li->li_directory = slapi_ch_strdup(dbhome);
-        } else if (S_ISREG(st.st_mode)) {
-            getdir(dbhome, NULL);
-            li->li_directory = slapi_ch_strdup(db_filename);
-            getdir(dbhome, NULL);
+        PL_strncpyz(dbhome, db_filename, MAXPATHLEN);
+        if (stat(dbhome, &st) == 0) {
+            if (S_ISDIR(st.st_mode)) {
+                li->li_directory = slapi_ch_strdup(dbhome);
+            } else if (S_ISREG(st.st_mode)) {
+                getdir(dbhome, NULL);
+                li->li_directory = slapi_ch_strdup(db_filename);
+                getdir(dbhome, NULL);
+            } else {
+                fprintf(stderr, "bdb_public_private_open: Unable to determine dbhome from %s\n", db_filename);
+                return EINVAL;
+            }
         } else {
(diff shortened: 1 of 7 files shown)


ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -6905,39 +6905,51 @@ bdb_public_private_open(backend *be, const char *db_filename, int rw, dbi_env_t
...
-        if (S_ISDIR(st.st_mode)) {
-            li->li_directory = slapi_ch_strdup(dbhome);
-        } else if (S_ISREG(st.st_mode)) {
+        PL_strncpyz(dbhome, db_filename, MAXPATHLEN);
+        if (stat(dbhome, &st) == 0) {
+            if (S_ISDIR(st.st_mode)) {
+                li->li_directory = slapi_ch_strdup(dbhome);
+            } else if (S_ISREG(st.st_mode)) {
+                getdir(dbhome, NULL);
+                li->li_directory = slapi_ch_strdup(db_filename);
```
tbordaz, line 6920: The patch does not change that code but I do no understand this line: li_directory refers to a file not a directory ?
Does it work ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -6905,39 +6905,51 @@ bdb_public_private_open(backend *be, const char *db_filename, int rw, dbi_env_t
...
-    conf->bdb_dbhome_directory = slapi_ch_strdup(dbhome);
     if (rw) {
         /* Setup a fully transacted environment */
         priv->dblayer_env = NULL;
-        conf->bdb_enable_transactions = 0;
+        conf->bdb_enable_transactions = 1;
         conf->bdb_tx_max = 50;
         rc = bdb_start(li, DBLAYER_NORMAL_MODE);
+        if (rc == 0) {
+            bdb_env = ((struct bdb_db_env*)(priv->dblayer_env))->bdb_DB_ENV;
```
tbordaz, line 6951: May be misleading, I think you set 'priv->dblayer_env' to NULL just above !

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
progier389: li->li_directory is supposed to be the instance directory (i.e .../db/userroot )
bdb_public_private_open is only called by dbscan to open the database.
And the db_filename is the instance path (i.e: .../db/userroot/cn.db) or it could be the global database (i.e .../db when trying to list all the existing instances)

 The puzzling line is  handling an error case (the provided path is wrong) as we cannot determine the li_directory, we provides directly the path (that will  generates an errors later on with a reasonable error message )
 Does it works: ?  No but that is expected. and it provides a reasonable error ( Not a directory) :
 
dbscan -D bdb -f $DB/userRoot/cn.db/foo.db 
bdb_public_private_open failed with db error 20 : Not a directory
Can't initialize db plugin: bdb

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
progier389: IMHO it is a pretty standard pattern: 
The value is first initialized to NULL to ensure it is not a random value (and make sure we do not try to free random address later on) 
Then it is set when the value once it is known

tbordaz (approved):
LGTM

PR #6349, opened 2024-09-30
Issue 6347 - VLV search sometime fails with operation error

### LINKED ISSUE #6347: VLV search sometime fails with operation error
**Issue Description**
https://issues.redhat.com/browse/RHCS-5299
VLV search sometime fails with LDAP_OPERATION error 
And error log contains 
`ERR - vlv_build_idl - Can't follow db cursor (err -12797)
`

 - Platform: [e.g. Fedora] Fedora 40 
 - Package and version: [e.g. 389-ds-base-3.0.1.2

**Steps to Reproduce**
1.  Prepare a F40 test machine
`1minutetip 1MT-Fedora-40`
2. . Install and launch the following script
```
#!/bin/bash
git clone https://github.com/freeipa/freeipa.git
cd freeipa/
dnf builddep -D "with_wheels 1" -D "with_lint 1" -D "with_doc 1" --spec freeipa.spec.in --best --allowerasing --setopt=install_weak_deps=False -y
bash makerpms.sh
dnf localinstall dist/rpms/* -y
ipa-server-install -a password -p dmpassword -r EXAMPLE.TEST -U --setup-dns --allow-zone-overlap --no-forwarders -N --auto-reverse --random-serial-numbers
mkdir ~/.ipa
cp /etc/ipa/default.conf  ~/.ipa
cp /etc/ipa/ca.crt  ~/.ipa
echo password | kinit admin
PYTHONPATH=. ipatests/ipa-run-tests -sv test_xmlrpc/test_cert_plugin.py
grep vlv_build_idl /var/log/dirsrv/slapd-EXAMPLE-TEST/errors
```
3. See the errors:
   - Some tests are FAIL
   - Error log show message like:
`
 [17/Sep/2024:18:54:05.250571209 +0000] - ERR - vlv_build_idl - Can't follow db cursor (err -12797)
`
   - Access log shows that a VLV operation fails with err=1 
```
   [26/Sep/2024:10:32:12.175793074 -0400] conn=17 op=315 SRCH base="ou=certificateRepository,ou=ca,o=ipaca" scope=1 filter="(certStatus=*)" attrs=ALL
[26/Sep/2024:10:32:12.179149635 -0400] conn=17 op=315 VLV 0:100:A None
[26/Sep/2024:10:32:12.179157931 -0400] conn=17 op=315 SORT serialno
[26/Sep/2024:10:32:12.179836172 -0400] conn=17 op=315 RESULT err=1 tag=101 nentries=0 wtime=0.000129631 optime=0.004046465 etime=0.004173324
```
Note: the VLV and SORT control may be missing becaise of https://github.com/389ds/389-ds-base/issues/6328
    Error lo

**Expected result**
All tests are PASS
Search should success
And no error should be logged. 

***See Also ***
Triaged: https://issues.redhat.com/browse/IDMDS-4698

progier389:
if  enabling certificate pruning freeipa tests sometime fails because VLV search sometime fails with err=1 and
error logs show - ERR - vlv_build_idl - Can't follow db cursor (err -12797)
Cause: The end of records is reached when trying to get a record in vlv index using its position
Solution: avoid to return MDB_NOT_FOUND in such case but return the last record instead

Issue: #6347 

Reviewed by: tbordaz (Thanks!)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
@@ -2387,6 +2387,10 @@ int dbmdb_cursor_set_recno(dbi_cursor_t *cursor, MDB_val *dbmdb_key, MDB_val *db
         rce->recno++;
         rc = MDB_CURSOR_GET(cursor->cur, &rce->key, &rce->data, MDB_NEXT);
     }
+    if (rc == MDB_NOTFOUND) {
+        /* Stay on last record if there are no more records */
+        rc = 0;
+    }
     if (rc == 0 && dbmdb_data->mv_size == rce->data.mv_size) {
         /* Should always be the case */
         memcpy(dbmdb_data->mv_data , rce->data.mv_data, dbmdb_data->mv_size);


ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
```diff
@@ -2387,6 +2387,10 @@ int dbmdb_cursor_set_recno(dbi_cursor_t *cursor, MDB_val *dbmdb_key, MDB_val *db
         rce->recno++;
         rc = MDB_CURSOR_GET(cursor->cur, &rce->key, &rce->data, MDB_NEXT);
     }
+    if (rc == MDB_NOTFOUND) {
```
tbordaz, line 2390: That means that the cursor failed to find rce->recno-- (its value was increased before cursor_get).
If (recno >= rce->recno--), that mean we were looking for a requested recno but did not find it. Should not it return NOTFOUND ?

Why not 
```
if ((rc == MDB_NOTFOUND) && (recno < (rce->recno - 1)) {
  rc = 0;
}
```

progier389: Though logically we should sometime returns MDB_NOTFOUND, I am a bit afraid about it:
 [1] this function is only used for VLV and I do not think that VLV search should fail because the index is out of the range. (I do not think it fails on bdb)
 [2] I am not 100% sure that we cannot have a bigger shift even if so far we have always seen rce->recno - 1) in that test
 So IMHO ignoring the MDB_NOTFOUND seems the safest solution ...

tbordaz: I agree. Sorry for the delay

tbordaz (approved):
LGTM

PR #6350, opened 2024-09-30
Issue 6343 - Improve onlne import robustness when the server is under load

### LINKED ISSUE #6343: Improve onlne import robustness when the server is under load
**Issue Description**

We have a check in the import code that will abort the import if there is a pending DB operation.  It does this check at the very start of the import process before the backend is disabled, etc.  The online import code should be more robust, and perform this check _after_ the backend/mapping tree is disabled.  It should also loop until all operations are complete (backend ref count is zero).

mreynolds389:
Description:

When the server is under load is can sometimes cause an online import to fail with an error message about "pending operations".

Move this ref count check (pending op check) after the backend is disabled, and retry for a few seconds.  If there are still pending operations then abort the import, otherwise continue.

Fixes: https://github.com/389ds/389-ds-base/issues/6343


progier389 (approved):
LGTM

Issue #6243, opened 2024-06-28
dsctl bdb2mdb should cleanly fail if bundled libdb is not avaiable


progier389:
**Issue Description**
When 389-ds-base-bdb is not installed (i.e if libdb is not installed) the database format conversion tools(dsctl bdb2mdb and dsctl mdb2bdb  should be cleanly disabled 

**Package Version and Platform:**
 - Platform: CentOS/Redhat 10
 - Package and version: 3.1

**Additional context**

Be careful thar the code should work for 2.5, 3.0 on Fedora (that requires the libdb rpm and ELN 10 that relie on the optionnal 389-ds-base-bdb rpm that contains the bundled library

progier389:
Triaged: https://issues.redhat.com/browse/IDMDS-4494

progier389:
5fa26eaf6..92eadd1ec  main -> main
   3be01de5a..2b58908a8  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6351, opened 2024-10-01
Issue 6243 - dsctl bdb2mdb should cleanly fail if bundled libdb is not available

progier389:
Determine what kind of Berkeley Database library is available
 and display a clear error message if it is not available or not usable 

Issue: #6243 

Reviewed by: @tbordaz , @vashirov , @droideck (Thanks!)

### DIFF
--- a/src/lib389/lib389/_constants.py
+++ b/src/lib389/lib389/_constants.py
@@ -379,7 +379,7 @@ CONTAINER_TLS_PWDFILE = '/data/config/pwdfile.txt'
 # Describe what kind of Berkeley Database library is available
 BDB_IMPL_STATUS = Enum('BDB_IMPL_STATUS', [
                        'UNKNOWN',   # Unable to discover
-                       'RPM',       # os libdb rpm is installed
+                       'STANDARD',  # os libdb rpm is installed
                        'BUNDLED',   # lib389 bundled rpm is installed
                        'READ_ONLY', # Read-only version is available
                        'NONE' ])    # bdb is not usasable
--- a/src/lib389/lib389/_constants.py
+++ b/src/lib389/lib389/_constants.py
@@ -375,3 +375,11 @@ CONTAINER_TLS_SERVER_KEY = '/data/tls/server.key'
 CONTAINER_TLS_SERVER_CERT = '/data/tls/server.crt'
 CONTAINER_TLS_SERVER_CADIR = '/data/tls/ca'
 CONTAINER_TLS_PWDFILE = '/data/config/pwdfile.txt'
+
+# Describe what kind of Berkeley Database library is available
+BDB_IMPL_STATUS = Enum('BDB_IMPL_STATUS', [
+                       'UNKNOWN',   # Unable to discover
+                       'RPM',       # os libdb rpm is installed
+                       'BUNDLED',   # lib389 bundled rpm is installed
+                       'READ_ONLY', # Read-only version is available
+                       'NONE' ])    # bdb is not usasable
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -51,7 +51,7 @@ class FakeArgs(dict):
 
 def get_bdb_impl_status():
     backldbm = 'libback-ldbm'
-    bundledbdb_plugin = 'libback-bdb'
+    bundledbdb_plugin = 'libback-ldbm'
     robdb_symbol = 'bdbro_getcb_vector'
     libdb = 'libdb-'
     plgstrs = check_plugin_strings(backldbm, [bundledbdb_plugin, robdb_symbol, libdb])
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -65,7 +65,7 @@ def get_bdb_impl_status():
         return BDB_IMPL_STATUS.READ_ONLY
     if plgstrs[libdb] is True:
         # standard bdb package build
-        return BDB_IMPL_STATUS.STABDARD
+        return BDB_IMPL_STATUS.STANDARD
     # Unable to find libback-ldbm plugin
     return BDB_IMPL_STATUS.UNKNOWN
 
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -22,7 +22,7 @@ from errno import ENOSPC
 from lib389.cli_base import CustomHelpFormatter
 from lib389._constants import DEFAULT_LMDB_SIZE, BDB_IMPL_STATUS, DN_CONFIG
 from lib389.dseldif import DSEldif
-from lib389.utils import parse_size, format_size
+from lib389.utils import parse_size, format_size, check_plugin_strings, find_plugin_path
 from pathlib import Path
 
 
@@ -35,14 +35,6 @@ MDB_MAP = "data.mdb"
 MDB_LOCK = "lock.mdb"
 
 LDBM_DN = "cn=config,cn=ldbm database,cn=plugins,cn=config"
-BDB_BUNDLED_RPM = '389-ds-base-bdb'
-BDB_RPM = 'libdb'
-TESTED_PACKAGES = ( BDB_BUNDLED_RPM, BDB_RPM )
-BDBRO_SYMBOL = 'bdbro_getcb_vector'
-BDB_SYMBOL = 'bdb_start'
-TESTED_SYMBOLS = ( BDB_SYMBOL, BDBRO_SYMBOL )
-
-
 CL5DB='replication_changelog.db'
 
 _log = None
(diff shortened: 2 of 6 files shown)


src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -234,6 +289,25 @@ def dblib_bdb2mdb(inst, log, args):
...
+
+    status = get_bdb_impl_status()
+    if status is BDB_IMPL_STATUS.UNKNOWN:
+        log.warning('Unable to determine if Berkeley Database library is available. If it is not the case, the migration will fail.')
+    elif status is BDB_IMPL_STATUS.NONE:
+        log.error('Berkeley Database library is not available. Maybe 389-ds-base-bdb rpm should be installed.')
+        raise RuntimeError('Berkeley Database library is not available')
+
+if pkgs[BDB_BUNDLED_RPM] is True:
+        return BDB_IMPL_STATUS.BUNDLED
```
droideck: Looks like the indentation is broken here.

droideck (changes requested):

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -48,6 +62,47 @@ def __init__(self, *args, **kwargs):
...
+    plugin = None
+    BDB_IMPL_STATUS
+    for pdir in PLUGIN_PATHS:
+        for ppath in glob.glob(f'{pdir}/libback-ldbm.*'):
+            plugin = ppath
+    if plugin and os.path.isfile(OBJDUMP_TOOL):
+        rc = subprocess.run(['objdump', '-R', plugin], text=True, capture_output=True)
+        symbs = { symb: symb in rc.stdout for symb in TESTED_SYMBOLS }
+    else:
+        symbs = { symb: None for symb in TESTED_SYMBOLS }
```
droideck: I think it'll be nice (but not required, - it's up to you) to make these two code blocks a separate helper function for readability and better maintainability.

LIke:
```
    pkgs = check_installed_packages()
    symbs = check_plugin_symbols()
```

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -48,6 +62,47 @@ def __init__(self, *args, **kwargs):
...
+
+    if pkgs[BDB_BUNDLED_RPM] is True:
+        return BDB_IMPL_STATUS.BUNDLED
+    if symbs[BDBRO_SYMBOL] is True:
+        return BDB_IMPL_STATUS.READ_ONLY
+    if pkgs[BDB_RPM] is not False and symbs[BDB_SYMBOL] is True:
+        return BDB_IMPL_STATUS.RPM
+    if symbs[BDB_SYMBOL] is False and pkgs[BDB_BUNDLED_RPM] is False:
```
tbordaz: I think at this step pkgs[BDB_BUNDLED_RPM] is false so we may not test it.

Why if we have not the bdb symbols should we fallback to NONE. Having BDB_RPM is not enough to decide it is RPM ?

src/lib389/lib389/cli_ctl/dblib.py
progier389: Good catch!  in fact this part of code copied from  get_bdb_impl_status() should be removed ....

src/lib389/lib389/cli_ctl/dblib.py
progier389: No. FYI: the values have a 3 way logic:   True,False or None  (if we are not able to determine if 
 the rpm is installed or not  ==> not True does not mean False 
 
 No if ds389 is build to use Bundle RPM or Read-Obly bdb, even if someone somehow install libdb it will not be used because the backend is not linked with libdb  ...

src/lib389/lib389/cli_ctl/dblib.py
progier389: I will do that but probably better to use:
```
    pkgs = check_installed_packages(TESTED_PACKAGES)
    symbs = check_plugin_symbols('libback-ldbm', TESTED_SYMBOLS)
```
To have reusable code (and move the helper function in utils.py

src/lib389/lib389/_constants.py
```diff
@@ -375,3 +375,11 @@ class AccessLog(IntEnum):
 CONTAINER_TLS_SERVER_CERT = '/data/tls/server.crt'
 CONTAINER_TLS_SERVER_CADIR = '/data/tls/ca'
 CONTAINER_TLS_PWDFILE = '/data/config/pwdfile.txt'
+
+# Describe what kind of Berkeley Database library is available
+BDB_IMPL_STATUS = Enum('BDB_IMPL_STATUS', [
+                       'UNKNOWN',   # Unable to discover
+                       'RPM',       # os libdb rpm is installed
```
vashirov: Not all distros use RPM. I think we should make this distro-agnostic.

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -33,6 +36,17 @@
...
+BDB_BUNDLED_RPM = '389-ds-base-bdb'
+BDB_RPM = 'libdb'
+TESTED_PACKAGES = ( BDB_BUNDLED_RPM, BDB_RPM )
+RPM_TOOL = '/usr/bin/rpm'
+PLUGIN_PATHS = ( '/usr/lib64/dirsrv/plugins/', '/usr/lib/dirsrv/plugins/' )
+BDBRO_SYMBOL = 'bdbro_getcb_vector'
+BDB_SYMBOL = 'bdb_start'
+TESTED_SYMBOLS = ( BDB_SYMBOL, BDBRO_SYMBOL )
+
+OBJDUMP_TOOL = '/usr/bin/objdump'
```
vashirov: objdump is part of binutils. In a proper prod environment it should not be present on the system.

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -48,6 +62,47 @@ def __init__(self, *args, **kwargs):
...
+
+    # Check libback-ldbm plugin content
+    p = Paths()
+    plugin = None
+    BDB_IMPL_STATUS
+    for pdir in PLUGIN_PATHS:
+        for ppath in glob.glob(f'{pdir}/libback-ldbm.*'):
+            plugin = ppath
+    if plugin and os.path.isfile(OBJDUMP_TOOL):
+        rc = subprocess.run(['objdump', '-R', plugin], text=True, capture_output=True)
```
vashirov: Instead of testing for symbols, can we make a decision about type of libdb based on `/proc/$(pidof ns-slapd)/maps` contents?

For example:
* bundled libdb
```
grep libdb /proc/$(pidof ns-slapd)/maps
7fdd4d9eb000-7fdd4da0f000 r--p 00000000 fd:02 4623526                    /usr/lib64/dirsrv/libdb-5.3-389ds.so
7fdd4da0f000-7fdd4db61000 r-xp 00024000 fd:02 4623526                    /usr/lib64/dirsrv/libdb-5.3-389ds.so
7fdd4db61000-7fdd4dba7000 r--p 00176000 fd:02 4623526                    /usr/lib64/dirsrv/libdb-5.3-389ds.so
7fdd4dba7000-7fdd4dbb1000 r--p 001bb000 fd:02 4623526                    /usr/lib64/dirsrv/libdb-5.3-389ds.so
7fdd4dbb1000-7fdd4dbb2000 rw-p 001c5000 fd:02 4623526                    /usr/lib64/dirsrv/libdb-5.3-389ds.so
```

* system libdb
```
grep libdb /proc/$(pidof ns-slapd)/maps
ffffa5470000-ffffa549b000 r--p 00000000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa549b000-ffffa54a0000 ---p 0002b000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa54a0000-ffffa55d4000 r-xp 00030000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa55d4000-ffffa55e0000 ---p 00164000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa55e0000-ffffa562a000 r--p 00170000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa562a000-ffffa5636000 ---p 001ba000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa5636000-ffffa5640000 r--p 001c6000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
ffffa5640000-ffffa5641000 rw-p 001d0000 fd:00 1181522                    /usr/lib64/libdb-5.3.so
```

And with no libdb available (or if it's already an instance with mdb), there wouldn't be any output.
The only caveat is that the instance needs to be running...

progier389: Interesting idea,. IMHO we can even keep the benefit of  both solutions:
  Keep current code and if status is UNKNOWN then try to  get the process maps 
  
  Another solution would be to use ldd on libback-ldbm.so since ldd is provided by glibc-common that should always be installed

progier389:
Here is a version  that directly for strings in libback-ldbm.so binary (without needing any tools)

progier389:
Fixed two issues:
   wrong bundle bdb plugin name (No idea why I named the plugin containing bundle-bdb glue: libback-lmdb :wink: )
   using glob.glob anywhere because mixing some modules using `import glob` and  other using `from glob import glob` cause trouble. Since there was far less file using the second form, I moved them to the first form.

src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -48,6 +49,27 @@ def __init__(self, *args, **kwargs):
...
+        # bundled bdb build
+        if find_plugin_path(bundledbdb_plugin):
+            return BDB_IMPL_STATUS.BUNDLED
+        return BDB_IMPL_STATUS.NONE
+    if plgstrs[robdb_symbol] is True:
+        # read-only bdb build
+        return BDB_IMPL_STATUS.READ_ONLY
+    if plgstrs[libdb] is True:
+        # standard bdb package build
+        return BDB_IMPL_STATUS.STABDARD
```
droideck: Probably a type (`STANDARD`?).
Also, it's not in `src/lib389/lib389/_constants.py`, I think.

droideck (approved):
Besides the minor issue, looks good!

src/lib389/lib389/cli_ctl/dblib.py
progier389: :+1: Good catch!  i fixed the typo and renamed RPM into STANDARD in the Enum definition.

Issue #6328, opened 2024-09-19
vlv control may not be logged


progier389:
**Issue Description**
vlv_print_access_log that logs the VLV and SORT lines in access log are not called in all failure cases
Typically we see:
```
[19/Sep/2024:09:01:09.958889789 -0400] conn=174 op=622 SRCH base="ou=certificateRepository,ou=ca,o=ipaca" scope=1 filter="(certStatus=*)" attrs=ALL
[19/Sep/2024:09:01:09.959904830 -0400] conn=174 op=622 RESULT err=1 tag=101 nentries=0 wtime=0.000116293 optime=0.001016967 etime=0.001131581 
```
and:
`[19/Sep/2024:09:01:09.959091978 -0400] - ERR - vlv_build_idl - Can't follow db cursor (err -12797)
`
So it is clearly a VLV search that is failing (because vlv_build_idl is used) but it is not reported in the access log. 

The source code shows that vlv_print_access_log is only called once the response control has been created (and  only in some branches of code) 
**Package Version and Platform:**
 - Platform: Fedora
 - Package and version:  main branch (and all older versions)

**Steps to Reproduce**
Need to be  investigated

**Expected results**
To help troubleshooting easier, vlv_print_access_log should be systematically called before
 returning from the search
   (if response control is not yet known (i.e: NULL) we should log None in the access VLV lines instead of the vlv control result data

progier389:
d862f3a81..b808a2de5  main -> main
  aaf38ee7e..e4f6ad35e  389-ds-base-3.0 -> 389-ds-base-3.0
  8aa9e3423..9572c251d  389-ds-base-2.5 -> 389-ds-base-2.5

PR #6354, opened 2024-10-03
Issue 6328 - vlv control may not be logged

progier389:
VLV control is logged only after succesfully sending the VLV response control 
which means that VLV is not visible if the search fail which make troubleshooting harder.
This fix, logs the VLV (with None as vlv result) everytime a VLV is used

This fix also improve lmdb debug log about VLVs

Issue: #6328 

Reviewed by: @tbordaz, @droideck  (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/vlv/regression_test.py
+++ b/dirsrvtests/tests/suites/vlv/regression_test.py
@@ -11,7 +11,8 @@ import os
 import glob
 import ldap
 from shutil import copyfile, rmtree
-from contextlib import contextmanager
+from datetime import datetime
+from contextlib import contextmanager, suppress
 from lib389.tasks import *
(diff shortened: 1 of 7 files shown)


droideck (approved):
The code looks good to me!

However, I'm slightly concerned about potential performance impact...
Maybe we could add an on/off switch for detailed VLV logging?

But if you've already checked and it's not really slowing things down, then nevermind!

tbordaz (approved):
LGTM

progier389:
About the performance concern:
This fix only impacts the failing VLV searches which means that in practice in a working production server it seldom triggers. Getting performance issue would mean an heavy load of failing VLV searches and probably that something is seriously broken. And IMHO in this case it is better to be aware that the failing  search are VLV one ...

PR #6357, opened 2024-10-08
Issue 6356 - On LMDB, after an update the impact VLV index, the vlv r…

### LINKED ISSUE #6356: On LMDB, after an update the impact VLV index, the vlv recno cache is not systematically cleared
**Issue Description**
The VLV manages/uses an index database. In LMDB in addition to the index there is one 'recno cache' database per VLV index. This recno cache, related to a given VLV index, is cleared when an update impacts the VLV index.
The recno cache is not systematically cleared upon update of the VLV index

 - since 2.5

**Steps to Reproduce**
install ipa-server and run test_xmlrpc/test_cert_plugin.py in a loop (see IDMDS-4688)
symptoms are 

- search failure 'RESULT err=1 tag=101 nentries=0'
- 'ERR - vlv_build_idl - Can't follow db cursor (err -12797)'

**Expected results**
VLV search should not fail

tbordaz:
…ecno cache is not systematically cleared

Bug description:
	The VLV manages/uses an index database.
	In LMDB in addition to the index there is one 'recno cache' database per VLV index.
	This recno cache, related to a given VLV index, is cleared when an update impacts the VLV index.
	The recno cache is not systematically cleared upon update of the VLV index.

	The way to clear the cache is to delete the record with key "OK".
	When the 'recno cache' does not contain this key, the next
	lookup to the cache, clears all entries and rebuild the cache
	from the vlv index.
	The deletion is done with mdb_del with both KEY and DATA
	refering to the same mdb_val. This means it deletes the records
	matching if both KEY/DATA.
	It should not match the DATA as the "OK" record is just
	a flag and all entries with that key should be removed.

Fix description:
	The fix consist to call mdb_del only with key

fixes: #6356

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
@@ -2882,7 +2882,7 @@ dbmdb_public_clear_vlv_cache(Slapi_Backend *be, dbi_txn_t *txn, dbi_db_t *db)
     ok.mv_size = 2;
     rc = dbmdb_open_dbi_from_filename(&rcdbi, be, rcdbname, NULL, 0);
     if (rc == 0) {
-        rc = MDB_DEL(TXN(txn), rcdbi->dbi, &ok, &ok);
+        rc = MDB_DEL(TXN(txn), rcdbi->dbi, &ok, NULL);
     }
     slapi_ch_free_string(&rcdbname);
     return rc;


progier389:
Have you tried  the fix ? 
I am a bit surprised because as far as I know the data associated with OK key is always OK so
IMHO this change does nothing ...
```
dbscan -f ./~recno-cache/vlv#allcertspkitomcatindex.db -A
...
KEY: OK	DATA: OK
...
```

progier389:
Maybe you get tricked by https://github.com/389ds/389-ds-base/pull/6349 that work around the issue ..

progier389 (approved):
Apparently, for some unknown reason, the change modify the behavior.
And since there is nothing wrong with deleting any record with "OK" key, lets accept this change

PR #6362, opened 2024-10-16
Issue 6067 - Update dsidm to prioritize basedn from .dsrc over interactive input

droideck:
Description: Modify dsidm CLI tool to check for the basedn in the .dsrc configuration file
when the -b option is not provided.
Previously, users were required to always specify the basedn interactively if -b was omitted,
even if it was available in .dsrc.
Now, the basedn is determined by first checking the -b option, then the .dsrc file, and finally
prompting the user if neither is set.

Related: https://github.com/389ds/389-ds-base/issues/6067

Reviewed by: ?

### DIFF
--- a/src/lib389/cli/dsidm
+++ b/src/lib389/cli/dsidm
@@ -128,7 +128,7 @@ if __name__ == '__main__':
     result = False
     try:
         inst = connect_instance(dsrc_inst=dsrc_inst, verbose=args.verbose, args=args)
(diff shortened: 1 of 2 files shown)


Firstyear (approved):
Thank you! I think this is a really good way to handle this :)

Firstyear:
If it's not too much trouble, can we back port this to 2.2?

PR #6363, opened 2024-10-16
Issue 6349 - RFE - Use previously extracted key path

### LINKED ISSUE #6340: excessive log warnings during certificate extraction
**Issue Description**

- SUSE 15 SP5

Firstyear:
Bug Description: slapd_SSL_client_auth uses the values of KeyExtractFile and CertExtractFile from the configuration entry, and each time it's called attempts to determine if this is an absolute or relative path. If the path is relative, it prepends a /tmp location based on if the running system /tmp is a private namespace or not. This causes a replication client that uses TLS certificate authentication to repeatedly emit warnings that the key extraction occurred to non-private tmp in a
container.

Fix Description: During key extraction, we extract keys and files using the "last token" from nss as the item that we extract. Because of this, we know that there can only be a single extracted key and cert, allowing us to extract these and store the full abs path of the extracted files rather than deriving them during each client iteration.

Review by: ???

### DIFF
--- a/ldap/servers/slapd/ssl.c
+++ b/ldap/servers/slapd/ssl.c
@@ -75,6 +75,10 @@ static int stimeout;
 static char *ciphers = NULL;
 static char *configDN = "cn=encryption,cn=config";
 
+/* The paths of extracted key and certs if any. */
+static char *key_extract_file = NULL;
+static char *cert_extract_file = NULL;
+
 
 /* Copied from libadmin/libadmin.h public/nsapi.h */
 #define SERVER_KEY_NAME "Server-Key"
@@ -1341,6 +1345,13 @@ slapd_ssl_init()
     return 0;
 }
 
+void
+slapd_ssl_destroy(void)
+{
+    slapi_ch_free_string(&key_extract_file);
+    slapi_ch_free_string(&cert_extract_file);
+}
+
 /*
  * val:   sslVersionMin/Max value set in cn=encryption,cn=config (INPUT)
  * rval:  Corresponding value to set SSLVersionRange (OUTPUT)
@@ -2063,8 +2074,6 @@ slapd_SSL_client_auth(LDAP *ld)
     SVRCOREStdPinObj *StdPinObj;
     SVRCOREError err = SVRCORE_Success;
     char *finalpersonality = NULL;
-    char *CertExtractFile = NULL;
-    char *KeyExtractFile = NULL;
 
     if ((family_list = getChildren(configDN))) {
         char **family;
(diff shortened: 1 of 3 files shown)


Firstyear:
Tested with suites/replication/tls_client_auth_repl_test.py and suites/basic

progier389:
Looks good except a minor point:
 I wonder if we should not have some cleanup code to free key_extract_file and cert_extract_file  called during shutdown (to avoid polluting libasan reports :wink:)

ldap/servers/slapd/ssl.c
```diff
@@ -2659,6 +2636,9 @@ slapd_extract_cert(Slapi_Entry *entry, int isCA)
         }
     }
     rv = SECSuccess;
+
+    cert_extract_file = slapi_ch_strdup(certfile);
```
tbordaz, line 2647: slapd_extract_cert is called several times. For example for 'cn=encryption,cn=config' and for each of its children. Should not  you call slapi_ch_free(&cert_extract_file) before setting it.

ldap/servers/slapd/ssl.c
```diff
@@ -3012,6 +2992,9 @@ slapd_extract_key(Slapi_Entry *entry, char *token __attribute__((unused)), PK11S
     PR_fprintf(outFile, "\n%s\n", KEY_TRAILER);
 #endif
     rv = SECSuccess;
+
+    key_extract_file = slapi_ch_strdup(keyfile);
```
tbordaz, line 3004: Same as above for cert_extract_file.

tbordaz (commented):
The change looks good except a doubt of a small memory leak.
If the concern was about the verbose warning in check_private_certdir you can also log it once, making the change smaller. 
So your fix is to avoid those annoying  logs and cleanup the code. Correct ?

Firstyear:
Good catch @tbordaz I have fixed both.

Firstyear:
> Looks good except a minor point: I wonder if we should not have some cleanup code to free key_extract_file and cert_extract_file called during shutdown (to avoid polluting libasan reports 😉)

We don't seem to have a shutdown call that I can use (unless I'm silly and couldn't see it).

droideck (approved):
LGTM!

progier389 (approved):
LGTM

tbordaz (approved):
LGTM as well

Firstyear:
```
   6fa06a719..aaf38ee7e  389-ds-base-3.0 -> 389-ds-base-3.0
   74f2f25a4..8aa9e3423  389-ds-base-2.5 -> 389-ds-base-2.5
   7715ff141..8d82ebf06  389-ds-base-2.4 -> 389-ds-base-2.4
   545947493..b6b0db91d  389-ds-base-2.3 -> 389-ds-base-2.3
   62c77454f..20ce92890  389-ds-base-2.2 -> 389-ds-base-2.2
   f6563c36b..b8a017bf3  389-ds-base-2.1 -> 389-ds-base-2.1
   9f45be601..49d3c04b9  389-ds-base-2.0 -> 389-ds-base-2.0
```

PR #6382, opened 2024-10-24
Issue 6381 - CleanAllRUV - move changelog purging to the very end of the task

### LINKED ISSUE #6381: CleanAllRUV - move changelog purging to the very end of the task
**Issue Description**

There are deadlock situations that can occur when cleanAllRUV is removing the clean task attribute (nsds5ReplicaCleanRUV) from the  replica config, while the change log purging is occurring.  Instead do the the changelog purge after everything else is done and have the changelog purging code remove the rid from the cleaned list once it finishes.

mreynolds389:
Description:

There are deadlock situations that can occur when cleanAllRUV is removing the clean task attribute (nsds5ReplicaCleanRUV) from the replica config, while the change log purging is occurring. Instead do the the changelog purge after everything else is done and have the changelog purging code remove the rid from the cleaned list once it finishes.

Also improved the task logging.

Fixes: https://github.com/389ds/389-ds-base/issues/6381


progier389 (approved):
LGTM

PR #6394, opened 2024-11-06
Issue 6340 - RFE - extract keys once (#6394)

### LINKED ISSUE #6340: excessive log warnings during certificate extraction

Firstyear:
Bug Description: Keys/Certs are extracted to PEM
repeatedly causing many warnings during outbound TLS authenticated replication

Fix Description: slapd_ssl_init() is called every time an outbound TLS connection is made, and will trigger a key extraction. If key extraction is already complete then the extraction is skipped.

Review by: ???

### DIFF
--- a/ldap/servers/slapd/ssl.c
+++ b/ldap/servers/slapd/ssl.c
@@ -1325,7 +1325,8 @@ slapd_ssl_init()
                 slapi_ch_free((void **)&token);
                 return -1;
             }
-            if (config_get_extract_pem()) {
+            /* Only extract the keys/cert *once* */
+            if (config_get_extract_pem() && _security_library_initialized == 0) {
                 /* Get Server{Key,Cert}ExtractFile from cn=Cipher,cn=encryption entry if any. */
                 slapd_extract_cert(entry, PR_FALSE);
                 slapd_extract_key(entry, isinternal ? internalTokenName : token, slot);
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

tbordaz (commented):
The function slapd_extract_cert is also called twice slapd_ssl_init. The first one to extract the CA cert and the second for the others certs. The fix prevents (re)extraction for the other certs. It could also prevent the re-extraction of the CA.

The test of _security_library_initialized is valid, you may also use slapd_security_library_is_initialized()

Firstyear:
slapd_security_library_is_initialized() is just "return _security_library_initialized" and given this is within the function that also sets _security_library_initialized, I think just checking that value is sufficient :)

Firstyear:
```
   fba88f397..97e910632  389-ds-base-3.0 -> 389-ds-base-3.0
   d1ea962ef..8d2300a43  389-ds-base-2.5 -> 389-ds-base-2.5
   0614f646f..24e598098  389-ds-base-2.4 -> 389-ds-base-2.4
   e6e2fad60..dd9d04a80  389-ds-base-2.3 -> 389-ds-base-2.3
   706fcca08..05707019d  389-ds-base-2.2 -> 389-ds-base-2.2
```

PR #6399, opened 2024-11-11
Issue 5842 - Add log buffering for error log

mreynolds389:
Description:

Adding log buffering for error log.  Also refactored and cleaned up some of the code.

relates: https://github.com/389ds/389-ds-base/issues/5842

### DIFF
--- a/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
+++ b/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
@@ -1369,8 +1369,46 @@ def test_missing_backend_suffix(topology_st, request):
         log.info('Restore dse.ldif')
         topology_st.standalone.stop()
         shutil.copy(dse_ldif + '.correct', dse_ldif)
+        topology_st.standalone.start()
 
     request.addfinalizer(fin)
+
+
+def test_errorlog_buffering(topology_st, request):
+    """Test log buffering works as expected when on or off
+
+    :id: 324ec5ed-c8ec-49fe-ab20-8c8cbfedca41
+    :setup: Standalone Instance
+    :steps:
+        1. Set buffering on
+        2. Reset logs and restart the server
+        3. Check for logging that should be buffered (not found)
+        4. Disable buffering
+        5. Reset logs and restart the server
+        6. Check for logging that should be found
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Success
+    """
+
+    # Configure instance
+    inst = topology_st.standalone
+    inst.config.replace('nsslapd-errorlog-logbuffering', 'on')
+    inst.deleteErrorLogs(restart=True)
+
+    time.sleep(1)
+    assert not inst.ds_error_log.match(".*slapd_daemon - slapd started.*")
+
+    inst.config.replace('nsslapd-errorlog-logbuffering', 'off')
+    inst.deleteErrorLogs(restart=True)
+
+    time.sleep(1)
+    assert inst.ds_error_log.match(".*slapd_daemon - slapd started.*")
+
 
 if __name__ == '__main__':
     # Run isolated
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -48,6 +48,8 @@ struct logbufinfo *logbuf_accum;
 static struct logging_opts loginfo;
 static int detached = 0;
 
+typedef int open_log(int32_t state, int32_t flags);
+
 //extern int slapd_ldap_debug;
 
 /*
@@ -121,11 +123,12 @@ static int vslapd_log_security(const char *log_data);
 static void log_convert_time(time_t ctime, char *tbuf, int type);
 static time_t log_reverse_convert_time(char *tbuf);
 static LogBufferInfo *log_create_buffer(size_t sz);
-static void log_append_buffer2(time_t tnl, LogBufferInfo *lbi, char *msg1, size_t size1, char *msg2, size_t size2);
+static void log_append_access_buffer(time_t tnl, LogBufferInfo *lbi, char *msg1, size_t size1, char *msg2, size_t size2);
 static void log_append_security_buffer(time_t tnl, LogBufferInfo *lbi, char *msg, size_t size);
 static void log_append_audit_buffer(time_t tnl, LogBufferInfo *lbi, char *msg, size_t size);
 static void log_append_auditfail_buffer(time_t tnl, LogBufferInfo *lbi, char *msg, size_t size);
-static void log_flush_buffer(LogBufferInfo *lbi, int type, int sync_now);
+static void log_append_error_buffer(time_t tnl, LogBufferInfo *lbi, char *msg, size_t size, int locked);
+static void log_flush_buffer(LogBufferInfo *lbi, int type, int sync_now, int locked);
 static void log_write_title(LOGFD fp);
 static void log__error_emergency(const char *errstr, int reopen, int locked);
 static void vslapd_log_emergency_error(LOGFD fp, const char *msg, int locked);
@@ -235,6 +238,8 @@ slapd_log_error_proc_internal(
  * LOG_WRITE_NOW(fd, buffer, size, headersize, err) writes into a LOGFD and
  *  flushes the buffer if necessary
  * LOG_CLOSE(fd) closes the logfile
+ *
+ * Note - the LOG_WRITE macros set "rc"
  */
 #define LOG_OPEN_APPEND(fd, filename, mode)                              \
     (((fd) = PR_Open((filename), PR_WRONLY | PR_APPEND | PR_CREATE_FILE, \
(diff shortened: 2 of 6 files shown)


dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
```diff
@@ -1369,9 +1369,47 @@ def fin():
...
+        6. Success
+    """
+
+    # Configure instance
+    inst = topology_st.standalone
+    inst.config.replace('nsslapd-errorlog-logbuffering', 'on')
+    inst.deleteErrorLogs(restart=True)
+
+    time.sleep(1)
+    assert not inst.ds_error_log.match(".*connection_table_new.*")
```
tbordaz: In case the log message change (no more connection_table_new) we may use what is in log_write_title ("-Directory/")

ldap/servers/slapd/log.h
```diff
@@ -183,7 +183,8 @@ struct logging_opts
     time_t log_error_ctime;             /* log creation time */
     LogFileInfo *log_error_logchain;    /* all the logs info */
     char *log_errorinfo_file;           /* error log rotation info file */
-    Slapi_RWLock *log_error_rwlock;     /* lock on error*/
+    Slapi_RWLock *log_error_rwlock;     /* lock on error */
+    LogBufferInfo *log_error_buffer;    /* buffer for access log */
```
tbordaz: The comment -> error log

ldap/servers/slapd/log.c
```diff
@@ -418,6 +429,10 @@ g_log_init()
     if ((loginfo.log_error_rwlock = slapi_new_rwlock()) == NULL) {
         exit(-1);
     }
+    loginfo.log_error_buffer = log_create_buffer(LOG_BUFFER_MAXSIZE);
+    if ((loginfo.log_error_buffer->lock = PR_NewLock()) == NULL) {
```
tbordaz, line 433: Should also test loginfo.log_error_buffer != NULL

ldap/servers/slapd/log.c
```diff
@@ -2245,12 +2260,16 @@ log_write_title(LOGFD fp)
...
     char buff[512];
     int bufflen = sizeof(buff);
+    int rc = 0;
 
     PR_snprintf(buff, bufflen, "\t%s B%s\n",
                 fe_cfg->versionstring ? fe_cfg->versionstring : CAPBRAND "-Directory/" DS_PACKAGE_VERSION,
                 buildnum ? buildnum : "");
     LOG_WRITE_NOW_NO_ERR(fp, buff, strlen(buff), 0);
-
+    if (rc != 0) {
```
tbordaz, line 2269: 'rc' is not set

ldap/servers/slapd/log.c
```diff
@@ -2803,6 +2819,44 @@ get_log_sev_name(int loglevel, char *sev_name)
...
+static void
+log_append_error_buffer(time_t tnl, LogBufferInfo *lbi, char *msg, size_t size, int locked)
+{
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+    char *insert_point = NULL;
+
+    /* While holding the lock, we determine if there is space in the buffer for our payload,
+       and if we need to flush.
+     */
+    PR_Lock(lbi->lock);
```
tbordaz, line 2831: is the flag 'locked' related to that lock ? shouldn't we test the flag

ldap/servers/slapd/log.c
```diff
@@ -3326,8 +3363,12 @@ log__open_accesslogfile(int logfile_state, int locked)
         PR_snprintf(buffer, sizeof(buffer), "LOGINFO:%s%s.%s (%lu) (%" PRId64 ")\n", PREVLOGFILE, loginfo.log_access_file, tbuf,
                     logp->l_ctime, logp->l_size);
         LOG_WRITE(fpinfo, buffer, strlen(buffer), 0);
+        if (rc != 0) {
```
tbordaz, line 3366: 'rc' is not set.

ldap/servers/slapd/log.c
```diff
@@ -3490,8 +3531,12 @@ log__open_securitylogfile(int logfile_state, int locked)
                     PREVLOGFILE, loginfo.log_security_file, tbuf,
                     logp->l_ctime, logp->l_size);
         LOG_WRITE(fpinfo, buffer, strlen(buffer), 0);
+        if (rc != 0) {
```
tbordaz, line 3534: I am missing where 'rc' is set

ldap/servers/slapd/log.c
```diff
@@ -6380,8 +6429,12 @@ log__open_auditlogfile(int logfile_state, int locked)
         PR_snprintf(buffer, sizeof(buffer), "LOGINFO:%s%s.%s (%lu) (%" PRId64 "d)\n", PREVLOGFILE, loginfo.log_audit_file, tbuf,
                     logp->l_ctime, logp->l_size);
         LOG_WRITE(fpinfo, buffer, strlen(buffer), 0);
+        if (rc != 0) {
```
tbordaz, line 6432: the same

ldap/servers/slapd/log.c
```diff
@@ -243,30 +248,36 @@ slapd_log_error_proc_internal(
...
+#define LOG_WRITE(fd, buffer, size, headersize)                                                   \
+    if (slapi_write_buffer((fd), (buffer), (PRInt32)(size)) != (PRInt32)(size)) {                 \
+        PRErrorCode prerr = PR_GetError();                                                        \
+        syslog(LOG_ERR, "Failed to write log, " SLAPI_COMPONENT_NAME_NSPR " error %d (%s): %s\n", \
+               prerr, slapd_pr_strerror(prerr), (buffer) + (headersize));                         \
+        rc = -1;                                                                                  \
+    }
+#define LOG_WRITE_NOW(fd, buffer, size, headersize, err)                                              \
+    do {                                                                                              \
+        (err) = 0;                                                                                    \
```
tbordaz, line 260: Instead of having LOG_WRITE_NOW_NO_ERR, we could test (err !=0) in LOG_WRITE_NOW and only use LOG_WRITE_NOW.

tbordaz (commented):
Fix looks good. minor remarks and a question about locking

ldap/servers/slapd/log.c
mreynolds389: > 'rc' is not set

The LOG_WRITE macros now set "rc"

ldap/servers/slapd/log.c
mreynolds389: > Instead of having LOG_WRITE_NOW_NO_ERR, we could test (err !=0) in LOG_WRITE_NOW and only use LOG_WRITE_NOW.

I'm a little weary of changing the behavior too much as I found that if we check all errors then things break.  Basically all the changes are just copy and paste of how it worked before.  I just organized it better.  Perhaps in a  new ticket we can look at some of the changes you are proposing?

ldap/servers/slapd/log.c
mreynolds389: > is the flag 'locked' related to that lock ? shouldn't we test the flag

Actually in my attempt to reduce all the duplicate code, and there's a lot more, we need to pass that lock flag but it's really only used for the error log when it comes down through slapd_log_error_proc_internal().

ldap/servers/slapd/log.c
mreynolds389: > Should also test loginfo.log_error_buffer != NULL

Perhaps, again this was copy and paste code.  So it's been working for a long time, but I'll take a look at it today.

dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
mreynolds389: > In case the log message change (no more connection_table_new) we may use what is in log_write_title ("-Directory/")

It needs to be logging text that happens further down after startup, so I can't use the header in this case.  Yeah it's weird how the error log works - some initial text always gets flushed right away.  So we need to check for text that gets logged "later" which makes the buffering more noticeable and testable.  

Anyway I will see what other "text" I can check for that is is more stable.

ldap/servers/slapd/log.c
mreynolds389: Hmm, if log_error_buffer would be NULL we would already crash in log_create_buffer().  Since this happens at server startup I don't think this is a necessary check as the server would be having all kinds of problems prior to hitting this code :-)

ldap/servers/slapd/log.c
mreynolds389: > > 'rc' is not set
> 
> The LOG_WRITE macros now set "rc"

Maybe we drop the macros and just write new functions to make this cleaner and easier to read?

mreynolds389: > > > 'rc' is not set
> > 
> > 
> > The LOG_WRITE macros now set "rc"
> 
> Maybe we drop the macros and just write new functions to make this cleaner and easier to read?

Actually reading the macro comments I now strongly feel we should clean it up.  It's talking about how NSPR performs on Windows NT :-)  I don't think these are things we need to be concerned about anymore :-)  But it should be cleaned up in a different ticket

progier389 (approved):
LGTM

ldap/servers/slapd/log.c
progier389: Cannot be done anyway because err must be an lvalue: (err) = .. will fail to compile if err is NULL

Issue #6374, opened 2024-10-21
nsslapd-mdb-max-dbs autotuning doesn't work properly


progier389:
**Issue Description**

[1] After restarting the server nsslapd-mdb-max-dbs may not be high enough to add a new backend
     because the value computation is wrong.

[2] dbscan fails to open the database if nsslapd-mdb-max-dbs has been increased.

[3] dbscan crash when closing the database (typically when using -S)

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main

**Steps to Reproduce**
1. See testcase dirsrvtests/tests/suites/config/config_test.py::test_lmdb_autotuned_maxdbs in the PR

**Expected results**
testcase should PASS
libasan should only report leaks 

**Additional context**

A run of the testcase with libasan howed several heap related issues:

 ```
   #0 0x7f5abe4c7d57 in import_abort_all /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/back-ldbm/import.c:162
    #1 0x7f5abe4c7e1a in factory_destructor /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/back-ldbm/import.c:195
    #2 0x7f5ac569ac4c in factory_destroy_extension /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/factory.c:366
    #3 0x42982a in connection_cleanup /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/connection.c:179
    #4 0x43565b in connection_table_move_connection_out_of_active_list /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/conntable.c:470
    #5 0x437fc7 in setup_pr_read_pds /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/daemon.c:1566
    #6 0x437fc7 in ct_list_thread /home/progier/sb/i6361/tst/source/389-ds-base/ldap/servers/slapd/daemon.c:1352
    #7 0x7f5ac4e193b6 in _pt_root (/lib64/libnspr4.so+0x2b3b6) (BuildId: e6dfe24882ade9ec65f7f1591a3e9e249e5a5780)
    #8 0x7f5ac5e5df95 in asan_thread_start(void*) (/lib64/libasan.so.8+0x5df95) (BuildId: a4ad7eb954b390cf00f07fa10952988a41d9fc7a)
[...]

PR #6400, opened 2024-11-12
Issue 6374 - nsslapd-mdb-max-dbs autotuning doesn't work properly

progier389:
Several issues:
- After restarting the server nsslapd-mdb-max-dbs may not be high enough to add a new backend
because the value computation is wrong.
 - dbscan fails to open the database if nsslapd-mdb-max-dbs has been increased.
- dbscan crashes when closing the database (typically when using -S)
When starting the instance the nsslapd-mdb-max-dbs  parameter is increased to ensure that a new backend may be added.
When dse.ldif path is not specified, the db environment is now open using the INFO.mdb data instead of using the default values.
synchronization between thread closure and database context destruction is hardened

Issue: #6374 

Reviewed by:  @tbordaz , @vashirov (Thanks!)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
@@ -99,7 +99,7 @@ dbmdb_compute_limits(struct ldbminfo *li)
      *  But some tunable may be autotuned.
      */
     if (dbmdb_count_config_entries("(objectClass=nsMappingTree)", &nbsuffixes) ||
-        dbmdb_count_config_entries("(objectClass=nsIndex)", &nbsuffixes) ||
+        dbmdb_count_config_entries("(objectClass=nsIndex)", &nbindexes) ||
         dbmdb_count_config_entries("(objectClass=vlvIndex)", &nbvlvs) ||
         dbmdb_count_config_entries("(objectClass=nsds5replicationagreement)", &nbagmt)) {
         /* error message is already logged */
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
@@ -83,7 +83,7 @@ dbmdb_compute_limits(struct ldbminfo *li)
     uint64_t total_space = 0;
     uint64_t avail_space = 0;
     uint64_t cur_dbsize = 0;
-    int nbchangelogs = 0;
+    int nbvlvs = 0;
     int nbsuffixes = 0;
     int nbindexes = 0;
     int nbagmt = 0;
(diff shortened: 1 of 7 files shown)


ldap/servers/slapd/back-ldbm/import.c
```diff
@@ -27,6 +27,8 @@
 #define NEED_DN_NORM_SP -25
 #define NEED_DN_NORM_BT -26
 
+pthread_mutex_t import_ctx_mutex = PTHREAD_MUTEX_INITIALIZER; /* Protect agains import context destruction */
```
tbordaz: static ?

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
```diff
@@ -100,7 +100,7 @@ dbmdb_compute_limits(struct ldbminfo *li)
      */
     if (dbmdb_count_config_entries("(objectClass=nsMappingTree)", &nbsuffixes) ||
         dbmdb_count_config_entries("(objectClass=nsIndex)", &nbsuffixes) ||
```
vashirov:
```suggestion
        dbmdb_count_config_entries("(objectClass=nsIndex)", &nbindexes) ||
```

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -4280,9 +4280,12 @@ dbmdb_import_init_writer(ImportJob *job, ImportRole_t role)
...
 dbmdb_free_import_ctx(ImportJob *job)
 {
-    if (job->writer_ctx) {
-        ImportCtx_t *ctx = job->writer_ctx;
-        job->writer_ctx = NULL;
+    ImportCtx_t *ctx = NULL;
+    pthread_mutex_lock(get_import_ctx_mutex());
+    ctx = job->writer_ctx;
+    job->writer_ctx = NULL;
+    pthread_mutex_unlock(get_import_ctx_mutex());
```
tbordaz, line 4287: Why not moving the unlock up to the end of the function ?

tbordaz (approved):
LGTM

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: The writer_ctx is reset in the lock, so we are sure that the other threads cannot use it and 
 there is no reason to block these threads longer than necessary ...
Furthermore keeping the critical section trivial avoid to have to check if there is a risk of deadlock which is not so obvious if you unlock at the end (because dbmdb_import_q_destroy performs some locking ... )

Issue #6401, opened 2024-11-12
Remove logging macros


mreynolds389:
**Issue Description**

We have these LOG_WRITE macros that are no longer needed.  There were used when we still supported Windows NT, and older UNIX platforms.

The code will easier to read and maintain if we move these macros to standard functions.

mreynolds389:
6c2cef0d4..e99668a61  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6402, opened 2024-11-12
Issue 6401 - Remove logging macros

mreynolds389:
Description:

Remove the log macros as they no longer serve their original purpose of dealing with issue on Windows NT.  LOG_WRITE_NOW was not even used.

Also did basic code cleanup.

relates: https://github.com/389ds/389-ds-base/issues/6401

### DIFF
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -91,6 +91,8 @@ static int slapi_log_map[] = {
 #define SLAPI_LOG_MIN SLAPI_LOG_FATAL /* from slapi-plugin.h */
 #define SLAPI_LOG_MAX SLAPI_LOG_DEBUG /* from slapi-plugin.h */
 #define LOG_CHUNK 16384 /* zlib compression */
+#define FLUSH PR_TRUE
+#define NO_FLUSH PR_FALSE
 
 /**************************************************************************
  * PROTOTYPES
@@ -209,13 +211,6 @@ compress_log_file(char *log_name)
     return 0;
 }
 
-int
-loglevel_is_set(int level)
-{
-    return (0 != (slapd_ldap_debug & level));
-}
-
-
 static int
 slapd_log_error_proc_internal(
     int loglevel,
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/log.c
```diff
@@ -224,64 +217,41 @@ slapd_log_error_proc_internal(
...
-    PR_Close((fd))
 
+static int32_t
+log_write(LOGFD fd, char *buffer, int32_t size, int32_t headersize)
+{
+    if (slapi_write_buffer((fd), (buffer), (PRInt32)(size)) != (PRInt32)(size)) {
+        PRErrorCode prerr = PR_GetError();
+        syslog(LOG_ERR, "Failed to write log, " SLAPI_COMPONENT_NAME_NSPR " error %d (%s): %s\n",
+               prerr, slapd_pr_strerror(prerr), (buffer) + (headersize));
+        return -1;
```
droideck: Minor nitpick: For consistency, we can use the `rc` here, too (the same way it's used in `log_write_now`).

ldap/servers/slapd/log.c
```diff
@@ -224,64 +217,41 @@ slapd_log_error_proc_internal(
...
- * unix platforms if there are more than 128 files open.
- *
- * LOG_OPEN_APPEND(fd, filename, mode) returns true if successful.  'fd' should
- *    be of type LOGFD (check log.h).  the file is open for appending to.
- * LOG_OPEN_WRITE(fd, filename, mode) is the same but truncates the file and
- *    starts writing at the beginning of the file.
- * LOG_WRITE(fd, buffer, size, headersize) writes into a LOGFD
- * LOG_WRITE_NOW(fd, buffer, size, headersize, err) writes into a LOGFD and
- *  flushes the buffer if necessary
- * LOG_CLOSE(fd) closes the logfile
```
droideck, line 240: I think it's worth leaving at least some of the comments and putting them around the new functions.

droideck (approved):
The rest LGTM!

ldap/servers/slapd/log.c
mreynolds389: > Minor nitpick: For consistency, we can use the `rc` here, too (the same way it's used in `log_write_now`).

Fixed.

ldap/servers/slapd/log.c
mreynolds389: > I think it's worth leaving at least some of the comments and putting them around the new functions.

There were not many comments about LOG_WRITE* but I added them back

ldap/servers/slapd/log.c
```diff
@@ -225,63 +218,47 @@ slapd_log_error_proc_internal(
...
+    PR_Sync(fd);
+
+    return rc;
+}
+
+int
+loglevel_is_set(int level)
+{
+    return (0 != (slapd_ldap_debug & level));
+}
```
tbordaz, line 249: It should be 'slapd_ldap_debug & slapi_log_map[level]' or better use slapi_is_loglevel_set

ldap/servers/slapd/log.c
```diff
@@ -3337,12 +3338,11 @@ log__open_accesslogfile(int logfile_state, int locked)
         return LOG_UNABLE_TO_OPENFILE;
     }
 
-
     /* write the header in the log */
     now = slapi_current_utc_time();
     log_convert_time(now, tbuf, 2 /* long */);
     PR_snprintf(buffer, sizeof(buffer), "LOGINFO:Log file created at: %s (%lu)\n", tbuf, now);
-    LOG_WRITE(fpinfo, buffer, strlen(buffer), 0);
+    log_write(fpinfo, buffer, strlen(buffer), 0);
```
tbordaz: check the return code ?

ldap/servers/slapd/log.c
```diff
@@ -3504,12 +3509,11 @@ log__open_securitylogfile(int logfile_state, int locked)
...
```
tbordaz: checking the return code ?

ldap/servers/slapd/log.c
```diff
@@ -6249,7 +6255,7 @@ log__open_errorlogfile(int logfile_state, int locked)
     now = slapi_current_utc_time();
     log_convert_time(now, tbuf, 2 /*long */);
     PR_snprintf(buffer, sizeof(buffer), "LOGINFO:Log file created at: %s (%lu)\n", tbuf, now);
-    LOG_WRITE(fpinfo, buffer, strlen(buffer), 0);
+    log_write(fpinfo, buffer, strlen(buffer), 0);
```
tbordaz: return code

ldap/servers/slapd/log.c
```diff
@@ -6408,7 +6420,7 @@ log__open_auditlogfile(int logfile_state, int locked)
...
```
tbordaz: return code

ldap/servers/slapd/log.c
```diff
@@ -6567,7 +6585,7 @@ log__open_auditfaillogfile(int logfile_state, int locked)
...
```
tbordaz: return code

ldap/servers/slapd/log.c
```diff
@@ -225,63 +218,47 @@ slapd_log_error_proc_internal(
...
+    }
+
+    return rc;
+}
+
+/*
+ * Writes into a LOGFD and flushes the buffer if necessary
+ */
+static int32_t
+log_write_now(LOGFD fd, char *buffer, int32_t size, int32_t headersize)
```
tbordaz: Instead of two functions, we could only use log_write and add a flag 'flush' to Pr_sync.

ldap/servers/slapd/log.c
mreynolds389: > It should be 'slapd_ldap_debug & slapi_log_map[level]' or better use slapi_is_loglevel_set

This is the original code, I don't think I should change it.   I just moved it to a new location

ldap/servers/slapd/log.c
mreynolds389: @tbordaz No, errors here are acceptable.  Aborting on an error breaks things - trust me I tried it before and the CI test fail.  We really only need the error checking for the error log where we need to shutdown if we can't log to it.  In my effort to reduce duplicate code we have to return an "rc" but it doesn't really apply to most logs.

 In the cases you mentioned it is the rotationinfo file that it is working on and errors are expected under certain conditions.  Again, I'm trying not to change the existing behavior because it break things.

ldap/servers/slapd/log.c
mreynolds389: > Instead of two functions, we could only use log_write and add a flag 'flush' to Pr_sync.

Sure I can do that

mreynolds389: > Instead of two functions, we could only use log_write and add a flag 'flush' to Pr_sync.

Done!

tbordaz (approved):
LGTM

PR #6403, opened 2024-11-13
Issue 6367 - RFE support of Session Tracking Control internet draft

### LINKED ISSUE #6367: RFE support of Session Tracking Control internet draft
It is helpful for ldapclient to get a tracking identifier logged, in the LDAP server logs, along with operations. The [internet draft](https://datatracker.ietf.org/doc/html/draft-wahl-ldap-session-03) propose an implementation

Implement a limited version of that draft focus on logging the session tracking identifier

Logging all information proposed in the draft is overkill and would impact the readingness of the logs

**Additional context**
replication will take benefit of this RFE to help diagnostic
see https://www.port389.org/docs/389ds/design/session-identifier-in-logs.html

tbordaz:
Bug description:
	This RFE is to support https://datatracker.ietf.org/doc/html/draft-wahl-ldap-session-03
	In short, it allows a client to send strings in a control.
	Those strings are added to the operation result logged in the
	access logs.
	Those strings are meaningful for the client (debug,
	kmonitoring,...).

Fix description:
	The design is https://www.port389.org/docs/389ds/design/session-identifier-in-logs.html

fixes: #6367

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/session_tracking/session_test.py
+++ b/dirsrvtests/tests/suites/session_tracking/session_test.py
@@ -1,0 +1,1576 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 RED Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ----
+
+import pytest, os, re, time
+from lib389.tasks import *
+from lib389.utils import *
+from lib389 import Entry
+from ldap import SCOPE_SUBTREE, ALREADY_EXISTS
+from ldap.controls import SimplePagedResultsControl
+from ldap.controls.sessiontrack import SessionTrackingControl, SESSION_TRACKING_CONTROL_OID
+from  ldap.extop import ExtendedRequest
+
+from lib389._constants import DEFAULT_SUFFIX, PW_DM, PLUGIN_MEMBER_OF
+from lib389.topologies import topology_st
+from lib389.plugins import MemberOfPlugin
+
+from lib389.schema import Schema
+from lib389.idm.user import UserAccount, UserAccounts
+from lib389.idm.account import Accounts
+from lib389.idm.account import Anonymous
+
+SESSION_SOURCE_IP = '10.0.0.10'
+SESSION_SOURCE_NAME = 'host.example.com'
+SESSION_TRACKING_FORMAT_OID = SESSION_TRACKING_CONTROL_OID + ".1234"
+
+pytestmark = pytest.mark.tier0
+
+def test_short_session_tracking_srch(topology_st, request):
+    """Verify that a short session_tracking string
+    is added (not truncate) during a search
+
+    :id: c9efc1cc-03c7-42b7-801c-440f7a11ee13
+    :customerscenario: False
+    :setup: Standalone instance, default backend
+    :steps:
+        1. Do a search with a short session tracking string
+        2. Restart the instance to flush the log
+        3. Check the exact same string is present in the access log
+    :expectedresults:
+        1. Search should succeed
+        2. success
+        3. Log should contain one log with that session
+    """
+
+
+    SESSION_TRACKING_IDENTIFIER = "SRCH short"
+    st_ctrl = SessionTrackingControl(
+      SESSION_SOURCE_IP,
+      SESSION_SOURCE_NAME,
+      SESSION_TRACKING_FORMAT_OID,
+      SESSION_TRACKING_IDENTIFIER
+    )
+
+    topology_st.standalone.search_ext_s(DEFAULT_SUFFIX,
+                                        ldap.SCOPE_SUBTREE,
+                                        '(uid=*)',
+                                        serverctrls=[st_ctrl])
+    topology_st.standalone.restart(timeout=10)
+    access_log_lines = topology_st.standalone.ds_access_log.match('.*tag=101.* sid="%s".*' % SESSION_TRACKING_IDENTIFIER)
+    assert len(access_log_lines) == 1
+
+    def fin():
+        pass
+
+    request.addfinalizer(fin)
+
+def test_short_session_tracking_add(topology_st, request):
+    """Verify that a short session_tracking string
+    is added (not truncate) during a add
+
+    :id: 04afd3de-365e-485f-9e00-913d913af931
+    :customerscenario: False
+    :setup: Standalone instance, default backend
+    :steps:
+        1. Add a test entry with a short session tracking
+        2. Restart the instance to flush the log
+        3. Check the exact same string is present in the access log
+    :expectedresults:
+        1. Add should succeed
+        2. success
+        3. Log should contain one log with that session
+    """
+
+
+    SESSION_TRACKING_IDENTIFIER = "ADD short"
+    st_ctrl = SessionTrackingControl(
+      SESSION_SOURCE_IP,
+      SESSION_SOURCE_NAME,
+      SESSION_TRACKING_FORMAT_OID,
+      SESSION_TRACKING_IDENTIFIER
+    )
+    TEST_DN = "cn=test_add," + DEFAULT_SUFFIX
+    try:
+        ent = topology_st.standalone.add_ext_s(TEST_DN,
+                                               [
+                                                    ('objectClass', b'person'),
+                                                    ('sn', b'test_add'),
+                                                    ('cn', b'test_add'),
+                                                    ('userPassword', b'test_add'),
+                                               ],
+                                               serverctrls=[st_ctrl])
+    except ldap.LDAPError as e:
+        log.fatal('Failed to add user1 entry, error: ' + e.message['desc'])
+        assert False
+
+    topology_st.standalone.restart(timeout=10)
+    access_log_lines = topology_st.standalone.ds_access_log.match('.*tag=105.* sid="%s".*' % SESSION_TRACKING_IDENTIFIER)
+    assert len(access_log_lines) == 1
+
+    def fin():
+        try:
+            topology_st.standalone.delete_s(TEST_DN)
+        except:
+            pass
+
+    request.addfinalizer(fin)
+
+def test_short_session_tracking_del(topology_st, request):
+    """Verify that a short session_tracking string
+    is added (not truncate) during a del
+
+    :id: a1391fbc-2107-4474-aaaf-088c393767a6
+    :customerscenario: False
+    :setup: Standalone instance, default backend
+    :steps:
+        1. Add a test entry
+        2. Delete the test entry with a short session tracking
+        3. Restart the instance to flush the log
+        4. Check the exact same string is not present in the access log for the ADD
+        5. Check the exact same string is present in the access log for the DEL
+    :expectedresults:
+        1. Add should succeed
+        2. DEL should succeed
+        3. success
+        4. Log should not contain a log with that session for the ADD
+        5. Log should contain one log with that session for the DEL
+    """
+
+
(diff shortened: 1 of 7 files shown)


Firstyear:
This is potentially an RCE due to ascii control characters when viewed by an admin.

tbordaz:
Thank @Firstyear for the rapid review. I assume RCE means remote code execution. I agree that taking client data requires careful checking. This data is ascii printable, less than 64K long. It is truncated to 15 first chars that are print in the log. Those chars (15bytes) are stored in the pblock. How can it be a RCE ?

progier389:
Not sure if it is really a RCE but anyway I think that @Firstyear concern is valid:
Logging directly the tag could be tricky and generates problems.
We should better escape the tag when logging it in access log to avoid bad surprise.

ldap/servers/slapd/result.c
```diff
@@ -2153,10 +2153,22 @@ log_result(Slapi_PBlock *pb, Operation *op, int err, ber_tag_t tag, int nentries
...
 
     get_internal_conn_op(&connid, &op_id, &op_internal_id, &op_nested_count);
     slapi_pblock_get(pb, SLAPI_PAGED_RESULTS_INDEX, &pr_idx);
     slapi_pblock_get(pb, SLAPI_PAGED_RESULTS_COOKIE, &pr_cookie);
+    slapi_pblock_get(pb, SLAPI_SESSION_TRACKING, &sessionTrackingId);
+    if (sessionTrackingId) {
+        int sz;
+        sz = strlen(sessionTrackingId) + 8;
+        session_str = slapi_ch_malloc(sz);
+        snprintf(session_str, sz, " sid=\"%s\"", sessionTrackingId);
```
progier389: should better avoid needless malloc/free and use directly a buffer of [6+15+1] char
i.e: char session_str[22] = "";

ldap/servers/slapd/control.c
```diff
@@ -162,6 +166,105 @@ slapi_get_supported_controls_copy(char ***ctrloidsp, unsigned long **ctrlopsp)
...
+    }
+    slapi_ch_free_string(&formatOID.bv_val); /* discard the value anyway */
+
+    /* Get sessionTrackingIdentifier and truncate it if needed */
+    ber_rc = ber_scanf(ber, "o}", &sessionTrackingIdentifier);
+    if ((ber_rc == LBER_ERROR) || (sessionTrackingIdentifier.bv_len > 65536)) {
+        rc = LDAP_PROTOCOL_ERROR;
+        goto free_and_return;
+    }
+    if (sessionTrackingIdentifier.bv_len > SESSION_ID_STR_SZ) {
```
progier389, line 223: should escape the session identifier with escape_string()  before checking if you need to truncate it and storing it in pblock.

tbordaz: I was thinking to only accept the control if the session id, provided by the client,  contains chars [32..126]. So printable chars only. Its purpose being to be logged in access log, I believe it is fair to restrict SID to this range. If we accept non printable chars then a byte can result with 4 chars to \\xx, so the SID final string would be very long.

ldap/servers/slapd/control.c
```diff
@@ -358,6 +467,25 @@ get_ldapmessage_controls_ext(
...
         slapi_pblock_set(pb, SLAPI_MANAGEDSAIT, &managedsait);
+        if (slapi_control_present(ctrls,
+                                  LDAP_CONTROL_X_SESSION_TRACKING, &session_tracking_spec, NULL)) {
+
+            parse_rc = parse_sessiontracking_ctrl(session_tracking_spec, &session_tracking_id);
+            if (parse_rc != LDAP_SUCCESS) {
+                Operation *pb_op = NULL;
+                slapi_pblock_get(pb, SLAPI_OPERATION, &pb_op);
+
+                slapi_log_err(SLAPI_LOG_WARNING, "get_ldapmessage_controls_ext", "Warning: conn=%" PRIu64 " op=%d failed to parse SessionTracking control (%d)\n",
```
progier389, line 477: IMHO the message should be a warning if criticallity flag is false and an error otherwise (in this case the operation should return LDAP_UNAVAILABLE_CRITICAL_EXTENSION

ldap/servers/slapd/control.c
progier389: That was why I proposed to escape the string before truncating it !-;)
I do not like the idea of rejecting non printable string:
   - double quote will still be a problem
   - application may be in trouble especially if the tag is in non ascii characters (better log diificult to read escaped tag than nothing !)
   - the rfc explicitly says that tag may be multibyte utf-8 characters

Firstyear:
@tbordaz ascii control characters - there is a method of adding them into a log such than an admin who views the log will trigger code to exec in their shell. 

You need to filter our ascii control chars is generally the easiest fix.

ldap/servers/slapd/control.c
tbordaz: Yes it makes sense. For adoption, application should not be limited on tags that they may not control.

ldap/servers/slapd/result.c
tbordaz: done

dirsrvtests/tests/suites/session_tracking/session_test.py
```diff
@@ -0,0 +1,1164 @@
...
+      SESSION_TRACKING_FORMAT_OID,
+      SESSION_TRACKING_IDENTIFIER
+    )
+
+    topology_st.standalone.search_ext_s(DEFAULT_SUFFIX,
+                                        ldap.SCOPE_SUBTREE,
+                                        '(uid=*)',
+                                        serverctrls=[st_ctrl])
+    topology_st.standalone.restart(timeout=10)
+    access_log_lines = topology_st.standalone.ds_access_log.match('.*tag=101.* sid="%s".*' % SESSION_TRACKING_IDENTIFIER)
```
progier389, line 64: IMHO the test may hit an older operation  result,
You should better generate a timestamp before doing the operation like:
```
from datetime import datetime
def log_timestamp(inst):
    current_datetime = datetime.now()
    timestamp = f'tsattr-{current_datetime.timestamp()}'.replace(".", "-")
    inst.search_s(DEFAULT_SUFFIX, SCOPE_BASE, f"(cn={timestamp})", escapehatch='i am sure')
    return timestamp

timestamp = log_timestamp(inst)
perform_operation_inst(inst)
...
access_log_lines = inst.ds_access_log.match('.*tag=101.* sid="%s".*' % SESSION_TRACKING_IDENTIFIER,
                                            after_pattern=timestamp)
```

progier389: Another way would be to use different sid for each operations

progier389: And that is what you do. Forget these 3 comment ! :wink:

ldap/servers/slapd/control.c
```diff
@@ -162,6 +166,124 @@ slapi_get_supported_controls_copy(char ***ctrloidsp, unsigned long **ctrlopsp)
...
+    sid = (char *) slapi_ch_calloc(1, sid_escaped_sz + 1);
+
+    /* Lets copy the escaped SID into the buffer */
+    if (sid_escaped_sz > SESSION_ID_STR_SZ) {
+        memcpy(sid, buf_sid_escaped, SESSION_ID_STR_SZ);
+        memset(sid + SESSION_ID_STR_SZ, '.', NB_DOTS); /* ending the string with "..." */
+    } else {
+        memcpy(sid, buf_sid_escaped, sid_escaped_sz);
+    }
+
```
progier389, line 250: Missing \0 ?
I know that calloc initialize to 0 but some code verification tools looking for vulnerability will generate a warning if the value is not explicitly set. so sid[sid_escaped_sz] = '\0'; is a good idea ...

tbordaz: Indeed this is a good idea. Thanks

ldap/servers/slapd/control.c
```diff
@@ -358,6 +488,31 @@ get_ldapmessage_controls_ext(
...
                                             LDAP_CONTROL_MANAGEDSAIT, NULL, NULL);
         slapi_pblock_set(pb, SLAPI_MANAGEDSAIT, &managedsait);
+        if (slapi_control_present(ctrls,
+                                  LDAP_CONTROL_X_SESSION_TRACKING, &session_tracking_spec, &iscritical)) {
+            Operation *pb_op = NULL;
+            slapi_pblock_get(pb, SLAPI_OPERATION, &pb_op);
+
+            if (iscritical) {
+                /* It must not be critical */
+                slapi_log_err(SLAPI_LOG_ERR, "get_ldapmessage_controls_ext", "conn=%" PRIu64 " op=%d Does not support critical SessionTracking control\n",
```
progier389: I feel that the error message is ambiguous.
To avoid misinterpretation, I would explicitly say that the application is buggy::
  "Does not support critical SessionTracking control (The application is not conform to https://datatracker.ietf.org/doc/html/draft-wahl-ldap-session-03 )."

ldap/servers/slapd/control.c
```diff
@@ -162,6 +166,125 @@ slapi_get_supported_controls_copy(char ***ctrloidsp, unsigned long **ctrlopsp)
...
+    }
+    ber = ber_init(session_tracking_spec);
+    if ((ber == NULL) || (session_tracking_id == NULL)) {
+        return LDAP_OPERATIONS_ERROR;
+    }
+
+    *session_tracking_id = NULL;
+
+    /* Get the sessionSourceIp but discard it as we are not using it */
+    ber_rc = ber_scanf(ber, "{o", &sessionSourceIp);
```
progier389: Should probaly better use directly:
 /* ignore sessionSourceIp, sessionSourceName, formatOID and get sessionTrackingIdentifier */
 ber_scanf(ber, "{xxxo}", &sessionTrackingIdentifier);

ldap/servers/slapd/abandon.c
```diff
@@ -87,6 +94,25 @@ do_abandon(Slapi_PBlock *pb)
 
     slapi_log_err(SLAPI_LOG_ARGS, "do_abandon", "id %d\n", id);
 
+    slapi_pblock_get(pb, SLAPI_SESSION_TRACKING, &sessionTrackingId);
+
+    /* prepare session_str to be logged */
+    if (sessionTrackingId) {
+        if (sizeof(session_str) < (strlen(sessionTrackingId) + 10 + 1)) {
```
progier389, line 101: Is it needed to check again the sessionTrackingId size ? Unless mistaken it is already truncated when stored to the pblock ...

ldap/servers/slapd/control.c
progier389: or maybe: `SessionTracking critical flag must be unset` 
FYI:  my first feeling when seeing this message is that there was a problem within ds389 ...

progier389:
Looks good except a few minor points:
  code simplification (ber parsing and redundant  size testing) 
 and more importantly an error message that IMHO should be changed

ldap/servers/slapd/control.c
tbordaz: Yes. This message is more explicit about what is failing

ldap/servers/slapd/control.c
tbordaz: indeed, a comment is better than useless code ;)

ldap/servers/slapd/abandon.c
tbordaz: As we have a static array. It is to prevent the risk that we increase the size of the sessionID (e.g. 15->20) without updating the array in abandon.c

progier389: ok.

progier389 (approved):
LGTM

PR #6261, opened 2024-07-12
Issue 6258 - Add sleep for ramdom failure in two tests

mmatsuya:
Bug Description: The following test fails frequently. In the first one, the access log is checked before it's written. In the second one, multiple threads with the same file descriptor for access log file writes the logs with out of order.

dirsrvtests/tests/suites/healthcheck/health_config_test.py::test_healthcheck_notes_unindexed_search dirsrvtests/tests/suites/paged_results/paged_results_test.py::test_multi_suffix_search

Fix Description: Adding sleep mitigates the failure.

Fixes: https://github.com/389ds/389-ds-base/issues/6258

(cherry-picked from the commit 1bd2e6e2a95e3f373fbdb7f5f79764687cae46e9)

PR #6409, opened 2024-11-19
Issue 6258 - Resolve race condition in paged_results_test.py

abeisemb:
The regression test dirsrvtests/tests/suites/paged_results/paged_results_test.py::test_multi_suffix_search has a race condition causing it to fail due to multiple queries potentially writing their logs out of chronological order.

This failure is mitigated by adding a boolean parameter to the paged_search() function so that, when the parameter is true, the function will sleep for a second after each query to avoid this race condition. The parameter is necessary because paged_search() is also used in a stress test, and sleeping after each query would result in that test being slowed down quite a bit. This new parameter defaults to False, so by default paged_search() will not sleep.

Helps fix: https://github.com/389ds/389-ds-base/pull/6261

### DIFF
--- a/dirsrvtests/tests/suites/paged_results/paged_results_test.py
+++ b/dirsrvtests/tests/suites/paged_results/paged_results_test.py
@@ -26,6 +26,8 @@ from lib389.idm.organizationalunit import OrganizationalUnit
 from lib389.backend import Backends
 from lib389._mapped_object import DSLdapObject
 
+from time import sleep
+
 pytestmark = pytest.mark.tier1
 
 DEBUGGING = os.getenv('DEBUGGING', False)
@@ -215,7 +217,7 @@ def change_conf_attr(topology_st, suffix, attr_name, attr_value):
     return attr_value_bck
 
 
-def paged_search(conn, suffix, controls, search_flt, searchreq_attrlist, abandon_rate=0):
+def paged_search(conn, suffix, controls, search_flt, searchreq_attrlist, abandon_rate=0, avoid_race_con=False):
     """Search at the DEFAULT_SUFFIX with ldap.SCOPE_SUBTREE
     using Simple Paged Control(should the first item in the
     list controls.
@@ -237,6 +239,10 @@ def paged_search(conn, suffix, controls, search_flt, searchreq_attrlist, abandon
     msgid = conn.search_ext(suffix, ldap.SCOPE_SUBTREE, search_flt, searchreq_attrlist, serverctrls=controls)
     log.info('Getting page %d' % (pages,))
     while True:
+        if avoid_race_con:
+            # This should only be true when calling from test_multi_suffix_search() to avoid a race condition
+            # with multiple queries.
+            sleep(1)
         try:
             rtype, rdata, rmsgid, rctrls = conn.result3(msgid, timeout=0.001)
         except ldap.TIMEOUT:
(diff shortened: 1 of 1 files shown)


abeisemb:
The full run for paged_results_test.py comes in at 6 minutes, 51 seconds with these changes.

dirsrvtests/tests/suites/paged_results/paged_results_test.py
```diff
@@ -1117,7 +1123,7 @@ def test_multi_suffix_search(topology_st, create_user, new_suffixes):
     try:
         req_ctrl = SimplePagedResultsControl(True, size=page_size, cookie='')
 
-        all_results = paged_search(topology_st.standalone, NEW_SUFFIX_1, [req_ctrl], search_flt, searchreq_attrlist)
+        all_results = paged_search(topology_st.standalone, NEW_SUFFIX_1, [req_ctrl], search_flt, searchreq_attrlist, 0, True)
```
progier389: Should better name explicitly the optional attribute instead on relying on its position:
`all_results = paged_search(topology_st.standalone, NEW_SUFFIX_1, [req_ctrl], search_flt, searchreq_attrlist, avoid_race_con=True)`

abeisemb: Thanks, I've updated the call in a commit just now. Full test runtime remains in the range of 6-7 minutes.

progier389 (changes requested):
But anyway I have a global bad feeling for this change.
 - Adding sleep seldom fix a race condition, it only decrease the probability of hitting hit.
 - Adding sleep increase the test duration 
 - The sleep does not seems to be at the right place, (should probably be before the while True:
There should be a better way to avoid such test failure:
   (since it is about log order, maybe by sorting access_log_lines according the opid ?)

abeisemb:
Understood. I'll look further into this, see if there's a better way to avoid this race condition.

dirsrvtests/tests/suites/paged_results/paged_results_test.py
```diff
@@ -237,6 +239,10 @@ def paged_search(conn, suffix, controls, search_flt, searchreq_attrlist, abandon
     msgid = conn.search_ext(suffix, ldap.SCOPE_SUBTREE, search_flt, searchreq_attrlist, serverctrls=controls)
     log.info('Getting page %d' % (pages,))
     while True:
+        if avoid_race_con:
+            # This should only be true when calling from test_multi_suffix_search() to avoid a race condition
+            # with multiple queries.
+            sleep(1)
         try:
             rtype, rdata, rmsgid, rctrls = conn.result3(msgid, timeout=0.001)
```
mreynolds389, line 247: What about increasing the result3() timeout?  Would that help?

abeisemb: Raising the timeout doesn't seem to make much of a difference, unfortunately. I previously raised it to 1, and then 10, and the test still failed due to log order. I'm thinking progier389's idea of reordering logs by opid might be the best solution here.

mreynolds389: Gotcha, was just throwing out ideas :-)

abeisemb:
Upon review, the additional commit breaks standard practice in this repository. Will close this MR, make a new branch+MR with proper commit practice, and submit the fix once I finish it.

PR #6410, opened 2024-11-19
Issue 6302 - Allow to run replication status without a prompt

### LINKED ISSUE #6302: web console is hanging.
**Issue Description**

web  console menu Monitoring --> Repliocation 

click  Repliocation --> webconsole hanging.

web console is hanging. because "dsconf replication monitor" use ~/.dsrc
but "dsconf replication status --suffix=xxxxxx" not use
so request binddn, bindpw

 - Platform: RHEL 8.9
 - Package and version: python3-lib389-1.4.3.39-7.module+el8.10.0+21985+3665ccdb.noarch.rpm
 - Browser chrome

**Screenshots**
image

droideck:
Description: We should allow running replication status and other similar commands without requesting a password and bind DN.

This way, the current instance's root DN and root PW will be used on other instances when requesting CSN info. If they are incorrect, then the info won't be printed, but otherwise, the agreement status will be displayed correctly.

Fixes: https://github.com/389ds/389-ds-base/issues/6302

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -317,12 +317,9 @@ def list_suffixes(inst, basedn, log, args):
 def get_repl_status(inst, basedn, log, args):
     replicas = Replicas(inst)
     replica = replicas.get(args.suffix)
-    pw_and_dn_prompt = False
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

Issue #6368, opened 2024-10-18
Add option to write access log in JSON


mreynolds389:
Add a configuration option to write the access log in JSON.  Just like what was done with the audit log recently

mreynolds389:
8bcd1b97b..ad75529c7  389-ds-base-3.0 -> 389-ds-base-3.0

mreynolds389:
b2a3b7078..3edcb75a3  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6412, opened 2024-11-20
Issue  6268 - write access log format infrastructure (part 1)

mreynolds389:
Description:

Create the config and layout for implementing a new logging format.

Also fixed some compiler warnings and minor UI bugs.

Relates: https://github.com/389ds/389-ds-base/issues/6368

### DIFF
--- a/ldap/servers/slapd/accesslog.c
+++ b/ldap/servers/slapd/accesslog.c
@@ -1,0 +1,517 @@
+/** BEGIN COPYRIGHT BLOCK
+ * Copyright (C) 2024 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * License: GPL (version 3 or any later version).
+ * See LICENSE for details.
+ * END COPYRIGHT BLOCK **/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "slap.h"
+
+#define JBUFSIZE 75
+#define KEY_SIZE 50
+
+/*
+ * ABANDON
+ *
+ * int32_t log_format
+ * time_t conn_time
+ * uint64_t conn_id
+ * int32_t op_id
+ * char* target_op
+ * char *sid
+ * int32_t msgid
+ * int32_t nentries
+ * int64_t tv_sec  - for etime
+ * int64_t tv_nsec - for etime
+ */
+
+int32_t
+slapd_log_access_abandon(slapd_log_pblock *logpb)
+{
+    int32_t rc = 0;
+    char etime[BUFSIZ] = {0};
+    char conn_key[KEY_SIZE] = {0};
+
+    /* Build the connection key */
+    PR_snprintf(conn_key, sizeof(conn_key), "%d-%" PRIu64,
+                (int32_t)logpb->conn_time, logpb->conn_id);
+
+    /* Construct the abandon etime */
+    if (logpb->tv_sec != -1) {
+        PR_snprintf(etime, sizeof(etime), "%" PRId64 ".%010" PRId64,
+                    logpb->tv_sec, logpb->tv_nsec);
+    }
+    if (logpb->nentries != -1) {
+
+    }
+
+    return rc;
+}
+
+/*
+ * ADD
+ *
+ * int32_t log_format
+ * time_t conn_time
+ * uint64_t conn_id
+ * int32_t op_id
+ * int32_t op_internal_id
+ * int32_t op_nested_count
+ * const char *dn
+ * char *authzid
+ * char *msg
+ * LDAPControl **controls
+ */
+int32_t
+slapd_log_access_add(slapd_log_pblock *logpb)
+{
+    int32_t rc = 0;
+
+    return rc;
+}
+
+/*
+ * AUTOBIND
+ *
+ * int32_t log_format
+ * time_t conn_time
+ * uint64_t conn_id
+ * char *bind_dn
+ */
+int32_t
+slapd_log_access_autobind(slapd_log_pblock *logpb)
+{
+    int32_t rc = 0;
+
+    return rc;
+}
+
+/*
+ * BIND
+ *
+ * int32_t log_format
+ * time_t conn_time
+ * uint64_t conn_id
+ * int32_t op_id
+ * const char *dn
+ * int32_t version
+ * const char *method
+ * const char *mech
+ * const char *msg
+ * LDAPControl **controls
+ */
+int32_t
+slapd_log_access_bind(slapd_log_pblock *logpb)
+{
+    int32_t rc = 0;
+
+    return rc;
+}
+
+/*
+ * UNBIND
+ *
+ * int32_t log_format
+ * time_t conn_time
+ * uint64_t conn_id
+ * int32_t op_id
+ * int32_t err
+ * char *close_err
+ * char *msg
+ */
+int32_t
+slapd_log_access_unbind(slapd_log_pblock *logpb)
+{
+    int32_t rc = 0;
+
+    /*
+     "conn=%" PRIu64 " op=%d UNBIND, error processing controls - error %d (%s)\n",
+     "conn=%" PRIu64 " op=%d UNBIND, decoding error: UnBindRequest not null\n"
+    */
+
+    return rc;
+}
+
+/*
+ * Disconnect
+ *
+ * int32_t log_format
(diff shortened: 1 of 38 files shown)


mreynolds389:
No regressions found in CI or in ASAN (although ASAN is finding valid leaks and Rust aborts - will open in a new ticket)

tbordaz (commented):
The change is large and the overall looks good. I think extended operation are missing (slapd_log_pblock/extendop.c)

mreynolds389:
> The change is large and the overall looks good. I think extended operation are missing (slapd_log_pblock/extendop.c)

Wow, yeah not sure how I missed extended ops.  I'll work on that...

ldap/servers/slapd/log.c
```diff
@@ -6647,7 +6647,7 @@ log_create_buffer(size_t sz)
  * logs_flush(). Tests show this speeds up searches by 10% on 4-way
  * systems.
  */
-static void
```
droideck: I might just be missing something in the code logic, but why did you change the `log_append_access_buffer` visibility? I don't see it being used anywhere in your patch...

ldap/servers/slapd/libglobs.c
```diff
@@ -2340,6 +2349,53 @@ config_get_auditlog_display_attrs()
...
+        slapi_ch_free_string(&slapdFrontendConfig->accesslog_log_format);
+        slapdFrontendConfig->accesslog_log_format = slapi_ch_strdup(value);
+        CFG_UNLOCK_WRITE(slapdFrontendConfig);
+    }
+
+    return LDAP_SUCCESS;
+}
+
+int32_t
+config_set_accesslog_time_format(const char *attrname, char *value, char *errorbuf, int apply)
```
droideck, line 2397: I think it'll be nice to add some kind of format validation here

ldap/servers/slapd/accesslog.c
```diff
@@ -0,0 +1,495 @@
...
+ * See LICENSE for details.
+ * END COPYRIGHT BLOCK **/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "slap.h"
+
+#define JBUFSIZE 75
```
droideck, line 15: That probably will be changed in the future, but just a note, 75 seems pretty small.

ldap/servers/slapd/result.c
```diff
@@ -2388,23 +2531,48 @@ static void
...
-                         op->o_connid, op->o_opid);
+        if (log_format != LOG_FORMAT_DEFAULT) {
+            slapd_log_pblock logpb = {0};
+
+            logpb.conn_time = op->o_conn_starttime;
+            logpb.conn_id = op->o_connid;
+            logpb.op_id = op->o_opid;
+            logpb.op_internal_id = -1;
+            logpb.op_nested_count = -1;
+            slapd_log_access_referral(&logpb);
```
droideck, line 2565: Init like this repeated a lot. Probably, it'll be better as a helper init function.

droideck (commented):
A few minor comments:)

ldap/servers/slapd/accesslog.c
mreynolds389: This is what we use for the audit log JSON format.  It's not used yet, but it will be and I think the size is ok for what it's used for.  See auditlog.c

ldap/servers/slapd/libglobs.c
mreynolds389: > I think it'll be nice to add some kind of format validation here

Hmmm, ok let me look into that...

ldap/servers/slapd/log.c
mreynolds389: > I might just be missing something in the code logic, but why did you change the `log_append_access_buffer` visibility? I don't see it being used anywhere in your patch...

Oh I was originally thinking I needed to call this function from outside log.c.  So I changed it from "static void" to "void", but I don't think I need to(not yet at least).  I'm surprised it didn't trigger a compiler warning.  Anyway nice catch I'll fix that!

ldap/servers/slapd/result.c
mreynolds389: > Init like this repeated a lot. Probably, it'll be better as a helper init function.

Hmmm, well every log event requires initializing the log pblock, but each event is a little different.  I kind of like having it all in one spot so you know exactly what it is doing, but I see what you are saying.  I can do that, have an internal op & external op init functions, but there will still be other commonly used/initialized keys, and I don't want the init functions to start to take a lot of parameters - defeats the point of using a pblock.

droideck: Sure! No strong opinion on this one! The `logpb.something = some_other_thing` pattern isn't too bad, yeah.
Feel free to handle it however you think best fits your coding style :)

mreynolds389: You have a valid suggestion, and I'm going to try it out since those first 5 parameters are very common (but not universal).  I will try it.  Originally I just had the logging functions taking 7-10 parameters but then it became a nightmare if you need to tweak the design or extend it.  So a "pblock" style seemed to be the best approach for maintainability.

mreynolds389: Sorry I'm going to skip this.  It doesn't really help much as not every logging call can use this consistently.  It looks kind of funny, and can be confusing.  I feel the cons outweigh the pros in this case.

mreynolds389:
@tbordaz @droideck - Ok rebased for session id tracking, added extended op logging, and added config validation in libglobs.c.  Please review...

ldap/servers/slapd/add.c
```diff
@@ -523,30 +543,57 @@ op_shared_add(Slapi_PBlock *pb)
...
+
         if (!internal_op) {
-            slapi_log_access(LDAP_DEBUG_STATS, "conn=%" PRIu64 " op=%d ADD dn=\"%s\"%s\n",
-                             pb_conn ? pb_conn->c_connid : -1,
-                             operation->o_opid,
-                             slapi_entry_get_dn_const(e),
-                             proxystr ? proxystr : "");
+            if (log_format != LOG_FORMAT_DEFAULT) {
+                /* JSON logging */
+                logpb.conn_time = pb_conn ? pb_conn->c_starttime : operation->o_conn_starttime;
```
progier389: I think that we should have a function to initialize the logpb that  that set all fields needed for most of the operations 
void init_logpb(slapd_log_pblock *logpb, slapd_pblock *pb, int is_internal_op)
That will ensure that we are not forgetting initializing some fields like the sid and that will make the code shorter

mreynolds389: But we are already NULLifying all values:  "slapd_log_pblock logpb = {0}"

mreynolds389: Also the pblock is not always available every time we make a log call.  Sometimes the conn and op structs are passed as arguments to the function that is making the logging call.  I'm just not sure we can consistently use a meaningful/useful init function as we still need to set unique struct elements for each logging call.  Sure, I can do:

    logpb_init(&logpb, start_time, conn_id, op_id, internal_op, inernal_nested_count)

But we will still have a bunch of other initializations to do for each logging call.  Thought it would be simpler and easier to read if it's all just done in the one spot.

progier389: Really ? Since the pblock is used for most of the  operations / internal operations I suspect that in most case, we should have it (In fact, I do not know when we do not have it  but  for the few cases were it is not there we can always initializing all fields specifically as of today).
And the init fonctions should only initialize the parameter that are used in all operations
 isome connection/operation data,
  including incomming controls,
   sid,
   staring time ...
The operation specific one should be initialized on per case basic 
i.e: 
    logpb_init(&logpb, pb);
    logpb.bind_dn = dn;
    logpb.method = "sasl";

"simpler and easier to read" is arguable,  but probably not so easy to maintain ...
if we have to change something (like adding a parameter)  it will again impact a lot of code in different  files
having a generic initialization will ensure that the change occurs in a few places

mreynolds389: Ok I'll work on this and see how it unfolds :-)

mreynolds389: @progier389 ok changes made, and the CI failures are not reproducible on my side.

progier389 (approved):
LGTM

PR #6413, opened 2024-11-21
Issue 6349 - RFE - extract keys once (#6363) (#6394)

### LINKED ISSUE #6340: excessive log warnings during certificate extraction

Firstyear:
Bug Description: Keys/Certs are extracted to PEM
repeatedly causing many warnings during outbound TLS authenticated replication

Fix Description: After more testing, if the connection is dropped and restarted, the certpath is retrieved but re-extraction does not occur. This still triggers the warning however. To resolve this, we only warn about the tpm namespace during library initialisation.

I really hope I got it right this time :(

Review by: ???

### DIFF
--- a/ldap/servers/slapd/ssl.c
+++ b/ldap/servers/slapd/ssl.c
@@ -966,9 +966,12 @@ check_private_certdir()
 
     if (!tmp_private) {
         /* tmp is not a private name space */
-        slapi_log_err(SLAPI_LOG_WARNING, "Security Initialization",
+        if (_security_library_initialized == 0) {
+            /* only alert about this the first time around */
+            slapi_log_err(SLAPI_LOG_WARNING, "Security Initialization",
                 "%s is not a private namespace. pem files not exported there\n",
                 private_mountpoint);
+        }
         return NULL;
     }
 


vashirov:
Minor nitpick: your commit and PR description point to a different issue. Judging by the title it should be #6394 instead of #6349.

progier389:
> Minor nitpick: your commit and PR description point to a different issue. Judging by the title it should be https://github.com/389ds/389-ds-base/pull/6394 instead of https://github.com/389ds/389-ds-base/pull/6349.

Should not it point to an issue rather than a PR ?

ldap/servers/slapd/ssl.c
```diff
@@ -966,9 +966,12 @@ check_private_certdir()
 
     if (!tmp_private) {
         /* tmp is not a private name space */
-        slapi_log_err(SLAPI_LOG_WARNING, "Security Initialization",
+        if (_security_library_initialized != 0) {
```
progier389: Is not the test inverted ?  
 The comment says the first time but (_security_library_initialized != 0) means every time except the first! ...

Firstyear: Indeed, and in my second test I misread the invalid output as the correct and vice versa. I feel quite silly now :(

Firstyear:
Fixed up both issue. So sorry to have missed this, it was a very long day :(

progier389 (approved):
LGTM.

vashirov:
Hey @Firstyear, I'm preparing new upstream releases and would like to include this fix as well.
I noticed that 7478498e26f5bf138c25225c3ad1c3e703be7e26 is not present in 2.0 and 2.1, while the previous fix https://github.com/389ds/389-ds-base/commit/b1b4356c4d92835ba9b03b90824f69334342d4a8 is present. Could you please also backport it?

Firstyear:
Yep, can do.

Firstyear:
```
   8330cbc6f..e56ce1187  389-ds-base-3.0 -> 389-ds-base-3.0
      979fb5db7..ff9b90991  389-ds-base-2.5 -> 389-ds-base-2.5
   1643e9868..2b040a9c1  389-ds-base-2.4 -> 389-ds-base-2.4
   bb9ca40d1..0fe972fe2  389-ds-base-2.3 -> 389-ds-base-2.3
   8afb46f71..9b5efd9fa  389-ds-base-2.2 -> 389-ds-base-2.2
   2b541c64b..b03d88e6e  389-ds-base-2.1 -> 389-ds-base-2.1
   4b34443ce..a3a738781  389-ds-base-2.0 -> 389-ds-base-2.0
```

PR #6418, opened 2024-11-22
Issue 6417 - If an entry RDN is identical to the suffix, then Entryrd…

### LINKED ISSUE #6417: If an entry RDN is identical to the suffix, then Entryrdn gets broken during a reindex
**Issue Description**

 - all versions (IIUC)

**Steps to Reproduce**
will provide a CI testcase

**Expected results**

tbordaz:
…n gets broken during a reindex

Bug description:
	During a reindex, the entryrdn index is built at the end from
	each entry in the suffix.
	If one entry has a RDN that is identical to the suffix DN,
	then entryrdn_lookup_dn may erroneously return the suffix DN
	as the DN of the entry.

Fix description:
	When the lookup entry has no parent (because index is under
	work) the loop lookup the entry using the RDN.
	If this RDN matches the suffix DN, then it exits from the loop
	with the suffix DN.
	Before exiting it checks that the original lookup entryID
	is equal to suffix entryID. If it does not match
	the function fails and then the DN from the entry will be
	built from id2enty

fixes: #6417

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/indexes/entryrdn_test.py
+++ b/dirsrvtests/tests/suites/indexes/entryrdn_test.py
@@ -11,11 +11,16 @@ import os
 import pytest
 import ldap
 import logging
+from lib389 import Entry
 from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
-from lib389.backend import Backends
+from lib389.backend import Backends, Backend
+from lib389.mappingTree import MappingTrees
+from lib389.configurations.sample import create_base_domain
+from lib389.idm.domain import Domain
 from lib389.idm.user import UserAccounts, UserAccount
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.topologies import topology_m2 as topo_m2
+from lib389.topologies import topology_st
 from lib389.agreement import Agreements
 from lib389.utils import ds_is_older, ensure_bytes
 from lib389.tasks import Tasks,ExportTask, ImportTask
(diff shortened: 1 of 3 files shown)


progier389 (changes requested):
Nice finding and nice test case but the test case fails on lmdb :wink:
IMHO entryrdn_lookup_dn is not even called in lmdb case ...

I wonder if the bdb fix could not be more simple. Something like:
```
         if (maybesuffix) {
+            if (id  >= 1) { 
+                rc = -1;
+                goto bail;
+            }
```

dirsrvtests/tests/suites/indexes/entryrdn_test.py
```diff
@@ -283,6 +288,105 @@ def test_long_rdn(topo_m2):
...
+
+    # Create the domain entry 'dc=dup_rdn,dc=dup_rdn'
+    trigger_entry = suffix + "," + suffix
+    topology_st.standalone.add_s(Entry((
+            trigger_entry, {
+            "objectClass": "top",
+            "objectClass": "domain",
+            "dc": dc_value,
+        }
+    )))
```
droideck: You can create it using `Domain` class - something like this:

```
# Create the domain entry 'dc=dup_rdn,dc=dup_rdn'
trigger_entry = suffix + "," + suffix
domain = Domain(inst, dn=trigger_entry)
domain.create(properties={
    'dc': dc_value,
    'description': 'Entry with RDN identical to suffix'
})
```

droideck (changes requested):
And yep, looks like LMDB wasn't able to reindex... The ds2index output in the LMDB test:

```
2
  ID: 2; RDN: "ou=my_org"; NRDN: "ou=my_org"
C1
  ID: 2; RDN: "ou=my_org"; NRDN: "ou=my_org"
P2
  ID: 1; RDN: "dc=dup_rdn"; NRDN: "dc=dup_rdn"
dc=dup_rdn
  ID: 1; RDN: "dc=dup_rdn"; NRDN: "dc=dup_rdn"
  ```

dirsrvtests/tests/suites/indexes/entryrdn_test.py
tbordaz: Oppss. I forgot to unset NSSLAPD_DB_LIB and run mdb. Indeed it fails, still something broken. I continue investigations

tbordaz:
Absolutely... it fixes my original testcase but broke everything else ;)
I am investigating a better fix

progier389 (approved):
LGTM, This time all the tests are PASS except the backup one which is fixed by PR #6423

dirsrvtests/tests/suites/indexes/entryrdn_test.py
droideck: My only concern is the comment above as I think it is still valid (using `Domain()` instead of `add_s(Entry)`)

tbordaz: Oppss good catch !
Actually I already did your suggested change and also updated the test for mdb specific message. But a wrong manipulation restored the original test file.
I restored the changes

droideck (approved):
Looks good! Ack

Issue #6420, opened 2024-11-25
Update dsidm user get_dn test after BZ 1893667 fix


mirielka:
Update dsidm user get_dn test that was marked as expected fail due to BZ 1893667, which is already fixed.

Issue #5373, opened 2022-07-20
dsidm user get_dn fails with search_ext() argument 1 must be str, not function


mreynolds389:
**Issue Description**

    # dsidm -v localhost -b dc=example,dc=com  user get_dn

...
DEBUG: _gen_dn filter = (&(objectclass=nsPerson)(objectclass=nsAccount)(objectclass=nsOrgPerson)(objectclass=posixAccount))
DEBUG: _gen_dn dn = <function get_dn.<locals>.<lambda> at 0x7f91dd05cdc0>
DEBUG: search_ext() argument 1 must be str, not function
Traceback (most recent call last):
  File "/usr/sbin/dsidm", line 141, in <module>
    result = args.func(inst, dsrc_inst['basedn'], log, args)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/cli_idm/user.py", line 38, in get_dn
    _generic_get_dn(inst, basedn, log.getChild('_generic_get_dn'), MANY, dn, args)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/cli_idm/__init__.py", line 101, in _generic_get_dn
    o = mc.get(dn=dn)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/_mapped_object.py", line 1185, in get
    results = self._get_dn(dn)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/_mapped_object.py", line 1203, in _get_dn
    return _search_ext_s(self._instance,
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/_mapped_object.py", line 92, in _search_ext_s
    return _ldap_op_s(inst, inst.search_ext_s, 'search_ext_s', *args, **kwargs)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/_mapped_object.py", line 74, in _ldap_op_s
    return f(*args, **kwargs)
  File "/home/mareynol/source/ds389/389-ds-base/src/lib389/lib389/__init__.py", line 176, in inner
    return f(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/ldap/ldapobject.py", line 863, in search_ext_s
    msgid = self.search_ext(base,scope,filterstr,attrlist,attrsonly,serverctrls,clientctrls,timeout,sizelimit)
[...]

PR #6421, opened 2024-11-25
Issue 6420 - Update dsidm user get_dn test after BZ 1893667 fix

mirielka:
Description: user get_dn test was marked as expected fail due to BZ issue 1893667. Since the issue is fixed, the test is being updated to working status.

Relates: https://github.com/389ds/389-ds-base/issues/6420

Author: Lenka Doudova

Reviewed by: ???


droideck (approved):
I don't think we should link BZ here.
And it'll be better to link the upstream issue where the fix is - https://github.com/389ds/389-ds-base/issues/5373

And the BZ can be linked there (I'll do it).

So please, fix the commit message (remove BZ number and add the above upstream issue link)
The rest, LGTM!

mirielka:
Thanks for review, commit message changed as requested.

Issue #6422, opened 2024-11-25
test_db_home_dir_online_backup CI test break other tests


progier389:
**Issue Description**
test_db_home_dir_online_backup CI test change the dbhome to a directory that 
 is removed at the test completion but the dbhome is not restored.
 So further testcases using the same topology fails
 
 Instead of using tempfile.TemporaryDirectory as context manager we should use a more complex one
that also set/reset  nsslapd-db-home-directory

**Package Version and Platform:**
 - Platform: [e.g. Fedora] Fedora
 - Package and version: main branch

**Steps to Reproduce**
1. Run backup CI tests on bdb
2. See error (test_after_db_log_rotation is failing)

**Expected results**
All tests should PASS

PR #6423, opened 2024-11-25
Issue 6422 - test_db_home_dir_online_backup CI test break other tests

progier389:
test_db_home_dir_online_backup CI test change the dbhome to a directory that
is removed at the test completion but the dbhome is not restored.
The solution is to use a python context manager to change and restore the dse.ldif

Issue: #6422 

Reviewed by: @tbordaz , @droideck (Thanks!)


droideck (approved):
Looks good!

tbordaz (approved):
LGTM as well

Issue #6424, opened 2024-11-27
Add basic dsidm group tests


mirielka:
Create basic dsidm group tests.

PR #6425, opened 2024-11-27
Issue 6424 - Add basic dsidm group tests

mirielka:
New basic dsidm group tests - create, delete, list, get, get_dn, modify, rename, rename with keep old rdn option.

Relates: https://github.com/389ds/389-ds-base/issues/6424

Author: Lenka Doudova

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsidm_group_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_group_test.py
@@ -1,0 +1,421 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+
+import pytest
+import logging
+import os
+
+from lib389 import DEFAULT_SUFFIX
+from lib389.cli_idm.group import list, get, get_dn, create, delete, modify, rename
+from lib389.topologies import topology_st
+from lib389.cli_base import FakeArgs
+from lib389.utils import ds_is_older, ensure_str
+from lib389.idm.group import Groups
+from . import check_value_in_log_and_reset
+
+pytestmark = pytest.mark.tier0
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+group_name = "test_group_2000"
+
+@pytest.fixture(scope="function")
+def create_test_group(topology_st, request):
+    groups = Groups(topology_st.standalone, DEFAULT_SUFFIX)
+
+    log.info('Create test group')
+    if groups.exists(group_name):
+        test_group = groups.get(group_name)
+        test_group.delete()
+
+    properties = FakeArgs()
+    properties.cn = group_name
+    create(topology_st.standalone, DEFAULT_SUFFIX, topology_st.logcap.log, properties)
+    test_group = groups.get(group_name)
+
+    def fin():
+        log.info('Delete test group')
+        if test_group.exists():
+            test_group.delete()
+
+    request.addfinalizer(fin)
+
+
+@pytest.fixture(scope="function")
+def add_description(topology_st, request):
+    args = FakeArgs()
+    args.selector = group_name
+    args.changes = ['add:description:{}'.format(group_name)]
+    modify(topology_st.standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args, warn=False)
+
+
+@pytest.mark.skipif(ds_is_older("1.4.2"), reason="Not implemented")
+def test_dsidm_group_create(topology_st):
+    """ Test dsidm group create option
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/idm/group.py
```diff
@@ -112,6 +112,23 @@ def __init__(self, instance, basedn, rdn=DEFAULT_BASEDN_RDN):
         else:
             self._basedn = '{},{}'.format(ensure_str(rdn), ensure_str(basedn))
 
+    def create_test_group(self, gid=2000):
```
progier389: I do not think this function is used. BTW as it is only used for test it should not be in lib389 but rather in the test

dirsrvtests/tests/suites/clu/dsidm_group_test.py
```diff
@@ -0,0 +1,416 @@
...
+@pytest.fixture(scope="function")
+def create_test_group(topology_st, request):
+    group_name = "test_group_2000"
+    groups = Groups(topology_st.standalone, DEFAULT_SUFFIX)
+
+    log.info('Create test group')
+    if groups.exists(group_name):
+        test_group = groups.get(group_name)
+        test_group.delete()
+    else:
```
progier389: deleting the group if is exists seems OK
but you should the create the group unconditionally (rather than in the "else" part)

dirsrvtests/tests/suites/clu/dsidm_group_test.py
```diff
@@ -0,0 +1,416 @@
...
+    def fin():
+        log.info('Delete test group')
+        if test_group.exists():
+            test_group.delete()
+
+    request.addfinalizer(fin)
+
+
+@pytest.mark.skipif(ds_is_older("1.4.2"), reason="Not implemented")
+def test_dsidm_group_create(topology_st):
```
progier389, line 61: I wonder if you should not use the create_test_group fixture then assert that the group exists

dirsrvtests/tests/suites/clu/dsidm_group_test.py
```diff
@@ -0,0 +1,416 @@
...
+
+pytestmark = pytest.mark.tier0
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="function")
+def create_test_group(topology_st, request):
+    group_name = "test_group_2000"
```
tbordaz: The group name (i.e 'test_group_2000') is used at several places in the file. You may use a global variable

dirsrvtests/tests/suites/clu/dsidm_group_test.py
```diff
@@ -0,0 +1,416 @@
...
+def test_dsidm_group_delete(topology_st, create_test_group):
+    """ Test dsidm group delete option
+
+    :id: d285db5d-8efd-4775-b5f4-47bf27106da6
+    :setup: Standalone instance
+    :steps:
+        1. Run dsidm group delete on a created group
+        2. Check that a message is provided on deletion
+        3. Check that group does not exist
+    :expected results:
```
bsimonova: The docstring needs to be :expectedresults: without space.

dirsrvtests/tests/suites/clu/dsidm_group_test.py
```diff
@@ -0,0 +1,416 @@
...
+@pytest.mark.skipif(ds_is_older("1.4.2"), reason="Not implemented")
+def test_dsidm_group_modify(topology_st, create_test_group):
+    """ Test dsidm group modify option
+
+    :id: 43611907-b477-4793-9ed6-8cf2d43ddd6b
+    :setup: Standalone instance
+    :steps:
+        1. Run dsidm group modify add description value
+        2. Run dsidm group modify replace description value
+        3, Run dsidm group modify delete description value
```
bsimonova: There needs to be "3. Run dsidm" with dot.

bsimonova (changes requested):

dirsrvtests/tests/suites/clu/dsidm_group_test.py
mirielka: Thank you for review. I left the creation test as it was, since it also checks that the correct messages are generated, which is not covered by the fixture. It seems redundant to include it to the fixture, as it would perform the check every time it's used, therefore in every single test.

dirsrvtests/tests/suites/clu/dsidm_group_test.py
mirielka: Fixed, thank you

dirsrvtests/tests/suites/clu/dsidm_group_test.py
mirielka: Fixed, thank you

progier389 (approved):
LGTM

bsimonova (approved):

bsimonova:
LGTM, just edit the commit message with reviewers.

PR #6426, opened 2024-11-29
Issue 5798 - Fix dsconf config multi-valued attr operations

### LINKED ISSUE #5798: CLI - dsconf inst config should support multi-valued operations

droideck:
Description: Fix add operation to properly handle multi-valued attributes so they persist after server restart.
Add support for multiple values at once via dsconf config. Handle delete operation in a more flexible way.
Refactor attribute handling code for better error handling and consistency.

Fixes: https://github.com/389ds/389-ds-base/issues/5798

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsconf_config_test.py
+++ b/dirsrvtests/tests/suites/clu/dsconf_config_test.py
@@ -1,0 +1,347 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2024 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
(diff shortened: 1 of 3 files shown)


progier389:
Looks good, but we should perhaps add some ci test ...

progier389 (approved):
LGTM

PR #6433, opened 2024-12-04
Issue 6258 - Mitigate race condition in paged_results_test.py

### LINKED ISSUE #6258: Random failure in test_healthcheck_notes_unindexed_search and test_multi_suffix_search
**Issue Description**
The following tests fails frequently. In the first one, the access log was checked before it's written and it failed to find an expected word in it. In the second one, some multiple threads have the same file descriptor to write access log, and those ones write it with their own timing. There is no guarantee that the log for the earlier queries are written into the log files earlier. Adding sleep mitigates this failure.

- dirsrvtests/tests/suites/healthcheck/health_config_test.py::test_healthcheck_notes_unindexed_search
- dirsrvtests/tests/suites/paged_results/paged_results_test.py::test_multi_suffix_search

 - Package and version: 389-ds-base-2.2.4-8.el9_2.x86_64 and 389-ds-base-2.5.1-1.el9.x86_64

**Steps to Reproduce**

1. py.test -v dirsrvtests/tests/suites/healthcheck/health_config_test.py::test_healthcheck_notes_unindexed_search
2. py.test -v dirsrvtests/tests/suites/paged_results/paged_results_test.py::test_multi_suffix_search

**Expected results**
Tests passed successfully with higher probability

**Additional context**
Access Log from the second test (test_multi_suffix_search):

```
# cat /var/log/dirsrv/slapd-standalone1/access |grep cookie
[11/Jul/2024:10:18:46.798464473 -0400] conn=1 op=63 RESULT err=0 tag=101 nentries=4 wtime=0.000050861 optime=0.001933251 etime=0.001982161 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=0
[11/Jul/2024:10:18:46.806845399 -0400] conn=1 op=66 RESULT err=0 tag=101 nentries=4 wtime=0.000041498 optime=0.002442879 etime=0.002482475 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=0
[11/Jul/2024:10:18:46.813039784 -0400] conn=1 op=67 RESULT err=0 tag=101 nentries=2 wtime=0.000034341 optime=0.004714516 etime=0.004746569 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=-1
[11/Jul/2024:10:18:46.800891409 -0400] conn=1 op=64 RESULT err=0 tag=101 nentries=2 wtime=0.000044929 optime=0.002472561 etime=0.002516322 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=0
[11/Jul/2024:10:18:46.803033964 -0400] conn=1 op=65 RESULT err=0 tag=101 nentries=4 wtime=0.000044943 optime=0.002182204 etime=0.002225282 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=0
[11/Jul/2024:10:18:46.796576171 -0400] conn=1 op=62 RESULT err=0 tag=101 nentries=4 wtime=0.000058250 optime=0.002272217 etime=0.002328974 notes=U,P details="Partially Unindexed Filter,Paged Search" pr_idx=0 pr_cookie=0
```
The op= numbers in this log are out of order.

abeisemb:
The regression test dirsrvtests/tests/suites/paged_results/paged_results_test.py::test_multi_suffix_search has a race condition causing it to fail due to multiple queries potentially writing their logs out of chronological order.

This failure is mitigated by sorting the retrieved access_log_lines by their "op" value. This ensures the log lines are in chronological order, as expected by the assertions at the end of test_multi_suffix_search().

Helps fix: https://github.com/389ds/389-ds-base/issues/6258

Reviewed by: @droideck , @progier389  (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/paged_results/paged_results_test.py
+++ b/dirsrvtests/tests/suites/paged_results/paged_results_test.py
@@ -7,6 +7,7 @@
 # --- END COPYRIGHT BLOCK ---
 #
 import socket
+import re
 from random import sample, randrange
 
 import pytest
@@ -1126,6 +1127,8 @@ def test_multi_suffix_search(topology_st, create_user, new_suffixes):
         topology_st.standalone.restart(timeout=10)
 
         access_log_lines = topology_st.standalone.ds_access_log.match('.*pr_cookie=.*')
+        # Sort access_log_lines by op number to mitigate race condition effects. 
+        access_log_lines.sort(key=lambda x: int(re.search(r"op=(\d+) RESULT", x).group(1)))
         pr_cookie_list = ([line.rsplit('=', 1)[-1] for line in access_log_lines])
         pr_cookie_list = [int(pr_cookie) for pr_cookie in pr_cookie_list]
         log.info('Assert that last pr_cookie == -1 and others pr_cookie == 0')


abeisemb:
Amended commit to move a helper function inside the test function for readability.

progier389 (approved):
LGTM
FYI:  rather than using indexes to extract the string you can also use directly the [regular expression module](https://docs.python.org/3/library/re.html). Here is an use example:

```
import re

str="xxxx op=123 conn=567 RESULT xxx"
result = re.search(r'(op=\d+ conn=\d+)', str)
print(result.group(1))
```

vashirov:
I agree with @progier389, this should be a regex.
Minor nit: in the commit message instead of "Helps fix" please use "Fixes:", see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

progier389:
Looks good.

abeisemb:
Added missing re import.

dirsrvtests/tests/suites/paged_results/paged_results_test.py
```diff
@@ -1126,6 +1127,8 @@ def test_multi_suffix_search(topology_st, create_user, new_suffixes):
         topology_st.standalone.restart(timeout=10)
 
         access_log_lines = topology_st.standalone.ds_access_log.match('.*pr_cookie=.*')
+        # Sort access_log_lines by op number to mitigate race condition effects. 
+        access_log_lines.sort(key=lambda x: int(re.search(r"op=(.*?) RESULT", x).group(1)))
```
droideck: Minor nitpick, but I think we should be as specific as possible to avoid any weird issues:

```suggestion
        access_log_lines.sort(key=lambda x: int(re.search(r"op=(\d+) RESULT", x).group(1)))
```

droideck (approved):
Besides that, LGTM!

dirsrvtests/tests/suites/paged_results/paged_results_test.py
abeisemb: Good catch! The commit is amended now with that change.

abeisemb: Full test runtime comes in at 7 minutes 2 seconds.

Issue #6432, opened 2024-12-04
Crash during bind when acct policy plugin does not have "alwaysrecordlogin" set


mreynolds389:
**Issue Description**

A crash occurs when account policy is configured, but "alwaysrecordlogin" is not set.  This leads to a NULL dereference:

```
#0  strcmpi_fast (dst=0x7fabb3c1e0e1 "bjectClass", src=0x1 <error: Cannot access memory at address 0x1>)
    at ../389-ds-base/ldap/servers/slapd/intrinsics.h:32
#1  0x00007fac57210fb5 in attrlist_find (a=0x7fabb3c24000, type=0x0) at ../389-ds-base/ldap/servers/slapd/attrlist.c:112
#2  0x00007fac57233fe0 in slapi_entry_attr_find (e=0x7fabb3c12000, type=0x0, a=0x7fabb53fc580) at ../389-ds-base/ldap/servers/slapd/entry.c:2237
#3  0x00007fac57235685 in slapi_entry_attr_has_syntax_value (e=0x7fabb3c12000, type=0x0, value=0x7fac55992640)
    at ../389-ds-base/ldap/servers/slapd/entry.c:3061
#4  0x00007fac55566a2b in acct_update_login_history (dn=0x7fabb3c62a80 "cn=mark,dc=example,dc=com", timestr=0x7fabb3c62ac0 "20241204152117Z")
    at ../389-ds-base/ldap/servers/plugins/acctpolicy/acct_plugin.c:233
#5  0x00007fac55567029 in acct_record_login (dn=0x7fabb3c62a80 "cn=mark,dc=example,dc=com")
    at ../389-ds-base/ldap/servers/plugins/acctpolicy/acct_plugin.c:375
#6  0x00007fac55567444 in acct_bind_postop (pb=0x7fabb3c00000) at ../389-ds-base/ldap/servers/plugins/acctpolicy/acct_plugin.c:527
#7  0x00007fac5729084b in plugin_call_func (list=0x7fac5599c900, operation=501, pb=0x7fabb3c00000, call_one=0)
    at ../389-ds-base/ldap/servers/slapd/plugin.c:2001
#8  0x00007fac572906ae in plugin_call_list (list=0x7fac5599c900, operation=501, pb=0x7fabb3c00000) at ../389-ds-base/ldap/servers/slapd/plugin.c:1944
[...]

PR #6434, opened 2024-12-04
Issue 6432 - Crash during bind when acct policy plugin does not have "alwaysrecordlogin" set

mreynolds389:
Description:

If alwaysrecordlogin is off then we dereference NULL ptr cfg->login_history_attr when trying to write the history/time value. Instead we should skip over this code if it is not set.

Relates: https://github.com/389ds/389-ds-base/issues/6432

### DIFF
--- a/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
+++ b/dirsrvtests/tests/suites/plugins/accpol_check_all_state_attrs_test.py
@@ -21,10 +21,11 @@ from lib389._constants import (
     PASSWORD,
     PLUGIN_ACCT_POLICY,
 )
-from lib389.idm.user import (UserAccount, UserAccounts)
+from lib389.idm.user import (UserAccount)
 from lib389.plugins import (AccountPolicyPlugin, AccountPolicyConfig)
+from lib389.cos import (CosTemplate, CosPointerDefinition)
 from lib389.idm.domain import Domain
-from datetime import datetime, timedelta
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tbordaz (approved):
LGTM

progier389:
Note: I just noticed that basic CI test are failing since: Issue  6268 - write access log format infrastructure (part 1) .
```
 -    def _search_be(belist, beinfo):
+    def _search_be(self, belist, beinfo):
```
change seems suspicious (because no calls to that function have not been changed ...)
 probably should have been:
```
 +    @classmethod
    def _search_be(belist, beinfo):
```
I am also a bit surprised with the:
```
     # Wait a bit till the task is created and available for searching
-    time.sleep(0.5)
+    time.sleep(1)
```
Seems weird that the task entry is not already created when the add operation that should add it returns.
is there a bug on ns-slapd side or should the sleep be removed ?

mreynolds389:
> Note: I just noticed that basic CI test are failing since: Issue 6268 - write access log format infrastructure (part 1) .
> 
> ```
>  -    def _search_be(belist, beinfo):
> +    def _search_be(self, belist, beinfo):
> ```
> 
> change seems suspicious (because no calls to that function have not been changed ...) probably should have been:
> 
> ```
>  +    @classmethod
>     def _search_be(belist, beinfo):
> ```
> 
> I am also a bit surprised with the:
> 
> ```
>      # Wait a bit till the task is created and available for searching
> -    time.sleep(0.5)
> +    time.sleep(1)
> ```
> 
> Seems weird that the task entry is not already created when the add operation that should add it returns. is there a bug on ns-slapd side or should the sleep be removed ?

I'll open a new issue/PR for this...

PR #6437, opened 2024-12-04
Issue 6436 - MOD on a large group slow if substring index is present

### LINKED ISSUE #6436: MOD on a large static group takes too long if substring index is present.
**Issue Description**
If the substring index is configured for the group membership attribute ( member or uniqueMember ),
the removal of a member from a large static group is pretty slow.

 - Platform: Red Hat Enterprise Linux Server release 7.8 Beta (Maipo)
 - Package and version: 389-ds-base-1.3.10.1-5.el7.tkt51166.x86_64

**Steps to Reproduce**
1. Create the substring index for the membership attribute.
2. Add a large static group.
3. Delete a member from the group.
4. The MOD operation to remove the value is too slow.

**Expected results**
Fast MOD operation.

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=1868348

jchapma:
Bug Description: If the substring index is configured for the group membership attribute ( member or uniqueMember ), the removal of a member from a large static group is pretty slow.

Fix Description: A solution to this issue would be to introduce a new index to track a membership atttribute index. In the interm, we add a check to healthcheck to inform the user of the implications of this configuration.

Fixes: https://github.com/389ds/389-ds-base/issues/6436

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_config_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_config_test.py
@@ -188,7 +188,7 @@ def test_healthcheck_RI_plugin_missing_indexes(topology_st):
 
 
 def test_healthcheck_MO_plugin_missing_indexes(topology_st):
-    """Check if HealthCheck returns DSMOLE0002 code
+    """Check if HealthCheck returns DSMOLE0001 code
 
     :id: 236b0ec2-13da-48fb-b65a-db7406d56d5d
     :setup: Standalone instance
@@ -203,8 +203,8 @@ def test_healthcheck_MO_plugin_missing_indexes(topology_st):
     :expectedresults:
         1. Success
         2. Success
-        3. Healthcheck reports DSMOLE0002 code and related details
-        4. Healthcheck reports DSMOLE0002 code and related details
+        3. Healthcheck reports DSMOLE0001 code and related details
+        4. Healthcheck reports DSMOLE0001 code and related details
         5. Success
         6. Healthcheck reports no issue found
         7. Healthcheck reports no issue found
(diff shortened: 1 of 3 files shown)


Firstyear (commented):
Looks good :)

dirsrvtests/tests/suites/healthcheck/health_config_test.py
```diff
@@ -236,6 +236,86 @@ def test_healthcheck_MO_plugin_missing_indexes(topology_st):
...
+    """
+
+    RET_CODE = 'DSMOLE0002'
+    MEMBER_DN = 'cn=member,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config'
+    UNIQUE_MEMBER_DN = 'cn=uniquemember,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config'
+
+    standalone = topology_st.standalone
+
+    log.info('Enable MO plugin')
+    plugin = MemberOfPlugin(standalone)
```
tbordaz, line 288: Is it required to enable memberof ?
IMHO healthcheck should only warn when a membership attribute is indexed in substring, independently of memberof being enabled/disabled

src/lib389/lib389/plugins.py
```diff
@@ -827,6 +827,56 @@ def _lint_member_attr_indexes(self):
...
+
+                        if not valid:
+                            report['detail'] = report['detail'].replace('ATTR', attr)
+                            report['detail'] = report['detail'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('ATTR', attr)
+                            report['fix'] = report['fix'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
+                            report['items'].append(suffix)
+                            report['items'].append(attr)
+                            report['check'] = f'memberof:substring_index'
```
tbordaz: Why 'memberof' ? I would expect the admin to check the need of 'attr' substring_index

src/lib389/lib389/plugins.py
```diff
@@ -827,6 +827,56 @@ def _lint_member_attr_indexes(self):
...
+                    except:
+                        # No index at all, bad
+                        report['detail'] = report['detail'].replace('ATTR', attr)
+                        report['detail'] = report['detail'].replace('BACKEND', suffix)
+                        report['fix'] = report['fix'].replace('ATTR', attr)
+                        report['fix'] = report['fix'].replace('BACKEND', suffix)
+                        report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
+                        report['items'].append(suffix)
+                        report['items'].append(attr)
+                        report['check'] = f'memberof:substring_index'
```
tbordaz: the same

dirsrvtests/tests/suites/healthcheck/health_config_test.py
jchapma: We discussed this on Tuesday and decided that is was easier to check for this configuration from the memberof plugin, similar to the check for the equality index type.

src/lib389/lib389/plugins.py
jchapma: Good point, I will change this

src/lib389/lib389/plugins.py
```diff
@@ -827,6 +827,56 @@ def _lint_member_attr_indexes(self):
                         report['check'] = f'memberof:attr_indexes'
                         yield report
 
+    def _lint_member_substring_index(self):
+        if self.status():
+            from lib389.backend import Backends
+            backends = Backends(self._instance).list()
+            attrs = self.get_attr_vals_utf8_l("memberofgroupattr")
+            container = self.get_attr_val_utf8_l("nsslapd-plugincontainerscope")
+            #import pdb; pdb.set_trace()
```
droideck: An unneeded leftover

src/lib389/lib389/plugins.py
```diff
@@ -827,6 +827,56 @@ def _lint_member_attr_indexes(self):
...
+                            report['detail'] = report['detail'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('ATTR', attr)
+                            report['fix'] = report['fix'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
+                            report['items'].append(suffix)
+                            report['items'].append(attr)
+                            report['check'] = f'attr:substring_index'
+                            yield report
+                    except:
+                        # No index at all, bad
```
droideck: If it's a 'No index at all' error, I think, we should be more specific with the `except:` and specify the exact exception expected.

droideck (changes requested):

src/lib389/lib389/plugins.py
jchapma: It is, my bad

src/lib389/lib389/plugins.py
jchapma: I  copied this from other memberof attr checks, but you're right. This test is to check if a membership attribute contains a substring index, if the index doesn't exist then we have nothing to report.

src/lib389/lib389/plugins.py
```diff
@@ -827,6 +827,46 @@ def _lint_member_attr_indexes(self):
...
+                            report['detail'] = report['detail'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('ATTR', attr)
+                            report['fix'] = report['fix'].replace('BACKEND', suffix)
+                            report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
+                            report['items'].append(suffix)
+                            report['items'].append(attr)
+                            report['check'] = f'attr:substring_index'
+                            yield report
+                    except:
+                        pass
```
droideck: No, what I mean is that we should not handle bare exceptions like that (and in Python, in general).

I think it should follow this logic (it's just an example, the actual exceptions should be something else):

```python
try:
    result = check()
except ValueError:
    handle_value_error()
except TypeError:
    handle_type_error()
```

tbordaz (approved):
LGTM

droideck (approved):
Looks good! Thank you!

PR #6445, opened 2024-12-10
Issue 6442 - Fix latest covscan memory leaks

### LINKED ISSUE #6442: Fix covscan memory leaks on main branch
**Issue Description**

Covscan reports 26 leaks on main branch, and other issues of course.   

```
ARRAY_VS_SINGLETON        9
BAD_CHECK_OF_WAIT_COND    11
COMPILER_WARNING          1
IDENTICAL_BRANCHES        1
INTEGER_OVERFLOW          5
LOCK_EVASION              9
OVERRUN                   35
RESOURCE_LEAK             26
USE_AFTER_FREE            25
Y2K38_SAFETY              2
```

mreynolds389:
Description:

Fixed a majority of the memory leaks (except ACL leaks), and some Y2K38 issues

Fixes: https://github.com/389ds/389-ds-base/issues/6442

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -540,8 +540,7 @@ def test_basic_import_export(topology_st, import_example_ldif):
     time.sleep(1)
 
     # Good as place as any to quick test the task has some expected attributes
-    if ds_is_newer('1.4.1.2'):
-        assert import_task.present('nstaskcreated')
+    assert import_task.present('nstaskcreated')
     assert import_task.present('nstasklog')
     assert import_task.present('nstaskcurrentitem')
     assert import_task.present('nstasktotalitems')
--- a/ldap/servers/plugins/acctpolicy/acct_plugin.c
+++ b/ldap/servers/plugins/acctpolicy/acct_plugin.c
@@ -272,7 +272,6 @@ acct_update_login_history(const char *dn, char *timestr)
         }
         /* first time round */
     } else if (cfg->login_history_size > 0) {
-        login_hist = (char **)slapi_ch_calloc(2, sizeof(char *));
         /* alloc new array and append latest value */
         login_hist = (char **)slapi_ch_realloc((char *)login_hist, sizeof(char *) * (num_entries + 2));
         login_hist[num_entries] = slapi_ch_smprintf("%s", timestr);
--- a/ldap/servers/plugins/acl/acllas.c
+++ b/ldap/servers/plugins/acl/acllas.c
@@ -4308,6 +4308,8 @@ acllas_replace_attr_macro(char *rule, lasInfo *lasinfo)
              * list which replaces the old one.
              */
             l = acl_strstr(&str[0], ")");
+            /* coverity false positive: l + 2 will always be positive */
+            /* coverity[integer_overflow] */
             macro_str = slapi_ch_malloc(l + 2);
             strncpy(macro_str, &str[0], l + 1);
             macro_str[l + 1] = '\0';
@@ -4323,17 +4325,18 @@ acllas_replace_attr_macro(char *rule, lasInfo *lasinfo)
 
             str++; /* skip the . */
             l = acl_strstr(&str[0], ")");
-            if (l == -1){
+            if (l == -1) {
                 slapi_log_err(SLAPI_LOG_ERR, plugin_name,
                               "acllas_replace_attr_macro - Invalid macro str \"%s\".", str);
                 slapi_ch_free_string(&macro_str);
                 charray_free(working_list);
                 return NULL;
             }
+            /* coverity false positive: l + 1 will always be positive */
+            /* coverity[integer_overflow] */
             macro_attr_name = slapi_ch_malloc(l + 1);
             strncpy(macro_attr_name, &str[0], l);
             macro_attr_name[l] = '\0';
-
 
             slapi_entry_attr_find(e, macro_attr_name, &attr);
             if (NULL == attr) {
--- a/ldap/servers/plugins/chainingdb/cb_conn_stateless.c
+++ b/ldap/servers/plugins/chainingdb/cb_conn_stateless.c
@@ -943,10 +943,10 @@ cb_update_failed_conn_cpt(cb_backend_instance *cb)
 {
     /* if the chaining BE is already unavailable, we do nothing*/
     time_t now;
+
+    slapi_lock_mutex(cb->monitor_availability.cpt_lock);
     if (cb->monitor_availability.farmserver_state == FARMSERVER_AVAILABLE) {
-        slapi_lock_mutex(cb->monitor_availability.cpt_lock);
         cb->monitor_availability.cpt++;
-        slapi_unlock_mutex(cb->monitor_availability.cpt_lock);
         if (cb->monitor_availability.cpt >= CB_NUM_CONN_BEFORE_UNAVAILABILITY) {
             /* we reach the limit of authorized failed connections => we setup the chaining BE state to unavailable */
             now = slapi_current_rel_time_t();
@@ -958,21 +958,22 @@ cb_update_failed_conn_cpt(cb_backend_instance *cb)
                           "cb_update_failed_conn_cpt - Farm server unavailable");
         }
     }
+    slapi_unlock_mutex(cb->monitor_availability.cpt_lock);
 }
 
 void
 cb_reset_conn_cpt(cb_backend_instance *cb)
 {
+    slapi_lock_mutex(cb->monitor_availability.cpt_lock);
     if (cb->monitor_availability.cpt > 0) {
-        slapi_lock_mutex(cb->monitor_availability.cpt_lock);
         cb->monitor_availability.cpt = 0;
         if (cb->monitor_availability.farmserver_state == FARMSERVER_UNAVAILABLE) {
             cb->monitor_availability.farmserver_state = FARMSERVER_AVAILABLE;
             slapi_log_err(SLAPI_LOG_PLUGIN, CB_PLUGIN_SUBSYSTEM,
                           "cb_reset_conn_cpt - Farm server is back");
         }
-        slapi_unlock_mutex(cb->monitor_availability.cpt_lock);
     }
+    slapi_unlock_mutex(cb->monitor_availability.cpt_lock);
 }
 
 int
--- a/ldap/servers/plugins/replication/repl5_replica.c
+++ b/ldap/servers/plugins/replication/repl5_replica.c
@@ -35,7 +35,7 @@ struct replica
     ReplicaUpdateDNList updatedn_list; /* list of dns with which a supplier should bind to update this replica */
     Slapi_ValueSet *updatedn_groups;   /* set of groups whose memebers are allowed to update replica */
     ReplicaUpdateDNList groupdn_list;  /* exploded listof dns from update group */
-    uint32_t updatedn_group_last_check;    /* the time of the last group check */
+    time_t updatedn_group_last_check;  /* the time of the last group check */
     int64_t updatedn_group_check_interval; /* the group check interval */
     ReplicaType repl_type;             /* is this replica read-only ? */
     ReplicaId repl_rid;                /* replicaID */
@@ -219,7 +219,7 @@ replica_new_from_entry(Slapi_Entry *e, char *errortext, PRBool is_add_operation,
          * during replica initialization
          */
         rc = _replica_update_entry(r, e, errortext);
-        /* add changelog config entry to config 
+        /* add changelog config entry to config
          * this is only needed for replicas logging changes,
          * but for now let it exist for all replicas. Makes handling
          * of changing replica flags easier
@@ -2544,7 +2544,7 @@ _replica_get_config_dn(const Slapi_DN *root)
     return dn;
 }
 /* when a replica is added the changelog config entry is created
- * it will only the container entry, specifications for trimming 
+ * it will only the container entry, specifications for trimming
  * or encyrption need to be added separately
  */
 static int
@@ -2876,6 +2876,7 @@ replica_update_state(time_t when __attribute__((unused)), void *arg)
                       "replica_update_state - Failed to get the config dn for %s\n",
                       slapi_sdn_get_dn(r->repl_root));
         replica_unlock(r->repl_lock);
+        slapi_mod_done(&smod);
         return;
     }
     pb = slapi_pblock_new();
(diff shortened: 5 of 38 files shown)


ldap/servers/plugins/acctpolicy/acct_plugin.c
```diff
@@ -272,9 +272,11 @@ acct_update_login_history(const char *dn, char *timestr)
         }
         /* first time round */
     } else if (cfg->login_history_size > 0) {
-        login_hist = (char **)slapi_ch_calloc(2, sizeof(char *));
-        /* alloc new array and append latest value */
-        login_hist = (char **)slapi_ch_realloc((char *)login_hist, sizeof(char *) * (num_entries + 2));
+        if (login_hist == NULL) {
```
progier389: FYI:  According to realloc man page, If ptr is NULL, then realloc(ptr, size) is equivalent to malloc(size)
so IMHO the proper code is:
```
/* alloc new array and append latest value */
login_hist = (char **)slapi_ch_realloc((char *)login_hist, sizeof(char *) * (num_entries + 2));
```
(i.e only removing useless line 275 ...)

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -537,11 +537,10 @@ def test_basic_import_export(topology_st, import_example_ldif):
     import_task.import_suffix_from_ldif(ldiffile=import_ldif, suffix=DEFAULT_SUFFIX)
 
     # Wait a bit till the task is created and available for searching
-    time.sleep(1)
+    time.sleep(2)
```
progier389: I do not think the changes in this file belong to this PR as it has nothing to do with covscan fixes ...
IMHO covscan fix is tricky enough to avoid the complexity of adding other changes in middle of them.
Better push them in another PR

ldap/servers/plugins/acl/acllas.c
```diff
@@ -4308,6 +4308,10 @@ acllas_replace_attr_macro(char *rule, lasInfo *lasinfo)
              * list which replaces the old one.
              */
             l = acl_strstr(&str[0], ")");
+            if (l + 2 < 0) {
```
progier389: test is a bit stange as l  +2 < 0 because acl_stestr either returns -1 or the string lenght which is >= 0  so l+2 >= 1
IMHO if (l<0) does makes more sense

ldap/servers/plugins/acl/acllas.c
```diff
@@ -4323,7 +4327,7 @@ acllas_replace_attr_macro(char *rule, lasInfo *lasinfo)
 
             str++; /* skip the . */
             l = acl_strstr(&str[0], ")");
-            if (l == -1){
+            if (l < 0) {
```
progier389: why changing that one ?

ldap/servers/plugins/chainingdb/cb_conn_stateless.c
```diff
@@ -958,21 +959,22 @@ cb_update_failed_conn_cpt(cb_backend_instance *cb)
                           "cb_update_failed_conn_cpt - Farm server unavailable");
         }
     }
+    slapi_unlock_mutex(cb->monitor_availability.cpt_lock);
```
progier389, line 961: lock unlocked twice: lines 950 and 962.
Should rather be in else par

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -35,7 +35,7 @@ struct replica
     ReplicaUpdateDNList updatedn_list; /* list of dns with which a supplier should bind to update this replica */
     Slapi_ValueSet *updatedn_groups;   /* set of groups whose memebers are allowed to update replica */
     ReplicaUpdateDNList groupdn_list;  /* exploded listof dns from update group */
-    uint32_t updatedn_group_last_check;    /* the time of the last group check */
+    uint64_t updatedn_group_last_check;    /* the time of the last group check */
```
progier389: Should be a time_t I think

ldap/servers/plugins/retrocl/retrocl_trim.c
```diff
@@ -239,7 +239,7 @@ trim_changelog(void)
     time_t now_maxage; /* used for checking if the changelog entry can be trimmed */
     changeNumber first_in_log = 0, last_in_log = 0;
     int num_deleted = 0;
-    int max_age, last_trim, trim_interval;
+    uint64_t max_age, last_trim, trim_interval;
```
progier389: time_t

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
```diff
@@ -1749,7 +1749,7 @@ bdb_import_push_progress_history(ImportJob *job, ID current_id, time_t current_t
 }
 
 static void
-bdb_import_calc_rate(ImportWorkerInfo *info, int time_interval)
+bdb_import_calc_rate(ImportWorkerInfo *info, uint64_t time_interval)
```
progier389: time_t ?

ldap/servers/slapd/plugin_internal_op.c
```diff
@@ -282,12 +276,6 @@ slapi_search_internal_set_pb_ext(Slapi_PBlock *pb, Slapi_DN *sdn, int scope, con
         return;
     }
 
-    /* Free op/attrs just in case this pb is being reused */
```
progier389, line 285: Not sure we should not keep these an rather add
 slapi_pblock_set(pb, SLAPI_OPERATION, NULL);
 slapi_pblock_get(pb, SLAPI_SEARCH_ATTRS, NULL);
 to avoid double free
 
 Same issue in two next changes

ldap/servers/slapd/util.c
```diff
@@ -1067,6 +1067,16 @@ slapi_rand_r(unsigned int *randx)
     return (int)(*randx & 0x7FFFFFFF);
 }
 
+int
+slapi_rand_r_64(uint64_t *randx)
```
progier389: Does this change belong to that PR ?

progier389 (changes requested):
one serious issues  (double unlock)
and lots of minor ones.

ldap/servers/plugins/acl/acllas.c
mreynolds389: Again this is just fixing a warning.  And I found it here:

```
Error: INTEGER_OVERFLOW ([CWE-190](https://cwe.mitre.org/data/definitions/190.html)): [[#def2]](https://cov01.lab.eng.brq2.redhat.com/osh/task/766355/log/389-ds-base-3.1.1.202412091343git4c51c6ea1-1.fc41/scan-results.html#def2) [important]
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4310:13: tainted_data_return: Called function "acl_strstr(&str[0], ")")", and a possible return value may be less than zero.
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4310:13: assign: Assigning: "l" = "acl_strstr(&str[0], ")")".
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4315:13: overflow: The expression "l + 2" is considered to have possibly overflowed.
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4315:13: overflow_sink: "l + 2", which might be negative, is passed to "slapi_ch_malloc(l + 2)".
# 4313|                   return NULL;
# 4314|               }
# 4315|->             macro_str = slapi_ch_malloc(l + 2);
# 4316|               strncpy(macro_str, &str[0], l + 1);
# 4317|               macro_str[l + 1] = '\0';

Error: INTEGER_OVERFLOW ([CWE-190](https://cwe.mitre.org/data/definitions/190.html)): [[#def3]](https://cov01.lab.eng.brq2.redhat.com/osh/task/766355/log/389-ds-base-3.1.1.202412091343git4c51c6ea1-1.fc41/scan-results.html#def3) [important]
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4329:13: tainted_data_return: Called function "acl_strstr(&str[0], ")")", and a possible return value may be less than zero.
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4329:13: assign: Assigning: "l" = "acl_strstr(&str[0], ")")".
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4337:13: overflow: The expression "l + 1" is considered to have possibly overflowed.
389-ds-base-3.1.1.202412091343git4c51c6ea1-build/389-ds-base-3.1.1.202412091343git4c51c6ea1/ldap/servers/plugins/acl/acllas.c:4337:13: overflow_sink: "l + 1", which might have overflowed, is passed to "slapi_ch_malloc(l + 1)".
# 4335|                   return NULL;
# 4336|               }
# 4337|->             macro_attr_name = slapi_ch_malloc(l + 1);
# 4338|               strncpy(macro_attr_name, &str[0], l);
# 4339|               macro_attr_name[l] = '\0';
```

dirsrvtests/tests/suites/basic/basic_test.py
mreynolds389: sounds good

ldap/servers/slapd/util.c
mreynolds389: Ok, so for Y2K38 time_t is going to become 64 bit (versus 32 bit).  So a lot of these changes are to address the upcoming Y2K38 issues.  So we need a 64 bit version of slapi_rand_r() for passwd_extop.c

ldap/servers/plugins/chainingdb/cb_conn_stateless.c
mreynolds389: > lock unlocked twice: lines 950 and 962. Should rather be in else par

Yeah I forgot to remove the unlock on line 950 when I moved it to after the "if" block.  Good catch.

ldap/servers/slapd/util.c
progier389: I do not think we really need to change that code as the time is only used to get a random seed value.
  That is the rare case were we could safely truncated the 64 bit time_t to a 32 bit unsigned integer   (upper bits are not really significant as they rarely change anyway) 
Using a 32 bit seed seems good enough especially as we want to generate a  small  random value ...

mreynolds389:
@progier389 Changes made, please review...

ldap/servers/slapd/passwd_extop.c
```diff
@@ -288,8 +288,8 @@ passwd_modify_generate_policy_passwd(passwdPolicy *pwpolicy,
     /* if only minalphas is set, divide it into minuppers and minlowers. */
     if (pwpolicy->pw_minalphas > 0 &&
         (my_policy[idx_minuppers] == 0 && my_policy[idx_minlowers] == 0)) {
-        unsigned int x = (unsigned int)time(NULL);
-        my_policy[idx_minuppers] = slapi_rand_r(&x) % pwpolicy->pw_minalphas;
+        uint64_t x = time(NULL);
```
progier389: should rather be uint32_t x = (uint32_t)time(NULL)
... slapi_rand_r(&x) ...

ldap/servers/slapd/passwd_extop.c
```diff
@@ -346,10 +346,10 @@ passwd_modify_generate_policy_passwd(passwdPolicy *pwpolicy,
     /* if password length is longer the sum of my_policy's,
        let them share the burden */
     if (passlen > tmplen) {
-        unsigned int x = (unsigned int)time(NULL);
+        uint64_t x = time(NULL);
```
progier389: same as above: uint32_t x = ...

progier389 (commented):
Still two changes needed passwd_extop.c around the slapi_rand_r
Casting int64 pointer to int32 pointer is a bad idea (depending on the endianess, we may ends up with a constand seed)  that is the value that should be cast.
 I wrote `uint32_t x = (uint32_t)time(NULL) `but reading the change again I have the feeling that the original code `unsigned int x = (unsigned int)time(NULL);` is correct.
(Sorry I should have been more precise when speaking about the cast in my previous week comments)
Everything else looks good ...

ldap/servers/slapd/passwd_extop.c
mreynolds389: I think coverity is still going to complain about this change as it wants a uint64_t for the time value.  I will try it, but... we do have some time (13 years) until a change is actually needed :-)  So we could skip this for now.

progier389: Maybe add a comment saying that we should ignore coverity scan warning (because we do not care about the 64-32 bits values when initializing the random seed)

mreynolds389: Reverted change and added comments.  Please review...

ldap/servers/slapd/passwd_extop.c
```diff
@@ -287,7 +287,9 @@ passwd_modify_generate_policy_passwd(passwdPolicy *pwpolicy,
 
     /* if only minalphas is set, divide it into minuppers and minlowers. */
     if (pwpolicy->pw_minalphas > 0 &&
-        (my_policy[idx_minuppers] == 0 && my_policy[idx_minlowers] == 0)) {
+        (my_policy[idx_minuppers] == 0 && my_policy[idx_minlowers] == 0))
+    {
+        /* For Y2K38 we will need to handle time() returning a uint64_t value */
```
progier389: I have the feeling that you missed my point: 
 For Y2K38 we do NOT have to do anything in this file:  time() will then return a uint64_t value that will then be truncated to a uint32_t but that is ok as we only want to generate 32 random  (more or less) bytes to initialize the pseudo random generator seed. (i.e: we do not care about the real date)
 The comment we should add is that we should ignore covscan Y2K38 warning. in theses two lines) !-;)
 
 Anyway I will approve the changes as it is as nothing is broken ...

progier389 (approved):
LGTM

ldap/servers/slapd/passwd_extop.c
mreynolds389: I added this above the two random function calls:

      /* coverity[store_truncates_time_t] */

PR #6448, opened 2024-12-12
Issue 6446 - on replica consumer, account policy plugin fails to mana…

### LINKED ISSUE #6446: on replica consumer, account policy plugin fails to manage the last login history
**Issue Description**
The RFE #5807 allows to manage the last login history. On a consumer in replicated topology, the update of the history fails.
The symptom are, lastloginTime is not updated and an error is logged like
`ERR - acct_update_login_history - Modify error 10 on entry
`

since 2.4.2

**Steps to Reproduce**
A reproducible testcase is provided with the PR

**Expected results**
lastloginTime should be updated and no error logged

tbordaz:
…ge the last login history

Bug description:
	An internal update needs the flag SLAPI_OP_FLAG_BYPASS_REFERRALS
	on a RO consumer. Else the server does not select the backend
	and returns LDAP_REFERRAL to the internal operation.
	In acct_update_login_history the flag is missing.

Fix description:
	Use the same flags in acct_update_login_history as it is
	used in acct_record_login

fixes: #6446

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/plugins/accpol_test.py
+++ b/dirsrvtests/tests/suites/plugins/accpol_test.py
@@ -10,7 +10,7 @@ import pytest
 import subprocess
 from lib389.tasks import *
 from lib389.utils import *
-from lib389.topologies import topology_st
+from lib389.topologies import topology_st, topology_m1c1
 from lib389.idm.user import (UserAccount, UserAccounts)
 from lib389.plugins import (AccountPolicyPlugin, AccountPolicyConfig, AccountPolicyConfigs)
 from lib389.cos import (CosTemplate, CosPointerDefinition)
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/plugins/accpol_test.py
```diff
@@ -1298,6 +1298,135 @@ def test_locact_modrdn(topology_st, accpol_local):
...
+        plugin.enable()
+        ACCPOL_DN = "cn={},{}".format(PLUGIN_ACCT_POLICY, DN_PLUGIN)
+        ACCP_CONF = "{},{}".format(DN_CONFIG, ACCPOL_DN)
+        plugin.set('nsslapd-pluginarg0', ACCP_CONF)
+        accp = AccountPolicyConfig(supplier, dn=ACCP_CONF)
+        accp.set('alwaysrecordlogin', 'yes')
+        accp.set('stateattrname', 'lastLoginTime')
+    except ldap.LDAPError as e:
+        log.error('Failed to enable Global Account Policy Plugin and Password policy attributes')
+        raise e
```
mreynolds389: Very minor, but I don't think we need these try/excepts - just let it raise the error when it happens.  Same for other code below.

mreynolds389 (approved):

Issue #6453, opened 2024-12-18
Fix memory leaks in ldbm_entryrdn.c


mreynolds389:
**Issue Description**

There are several memory leaks in ldbm_entryrdn.c with mdb and long DN's

To reproduce run:  dirsrvtests/tests/suites/indexes/entryrdn_test.py

mreynolds389:
cea43e0b0..be58fbec5  389-ds-base-3.0 -> 389-ds-base-3.0
39282eb5c..ffec7565c  389-ds-base-2.6 -> 389-ds-base-2.6
a303da80a..6d3ba79f2  389-ds-base-2.5 -> 389-ds-base-2.5

tbordaz:
7adcd8af0..11f473daa  main
9c2b6290d..b03738dff  389-ds-base-3.0
7f26842c5..5451a4abf  389-ds-base-2.6
856f2bd50..60b6a5837  389-ds-base-2.5

PR #6455, opened 2024-12-19
Issue 6453 - Fix memory leaks in entryrdn

mreynolds389:
Description:

We leak memory when processing long DN's (mdb).  We have to loop over long DN's and on each pass we leak the previous rdn element. In the tombstone case we just need to free the current childelem

Relates: https://github.com/389ds/389-ds-base/issues/6453

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
@@ -1534,6 +1534,7 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
     dbi_entryrdn_records_t rec = {0};
     int rc = 0;
     int dbop = 0;
+    void *prev_elem_data = NULL;
 
     if (NULL == ctx || NULL == key || NULL == data || NULL == elem ||
         NULL == comp_key) {
@@ -1548,7 +1549,7 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
         slapi_log_err(ENTRYRDN_LOGLEVEL(rc), "_entryrdn_get_elem",
                       "Backend %s suffix %s is too long.\n",
                       ctx->be->be_name, (char*)(key->data));
-        rc = DBI_RC_OTHER;;
+        rc = DBI_RC_OTHER;
         _ENTRYRDN_DEBUG_GOTO_BAIL();
         goto bail;
     }
@@ -1561,8 +1562,14 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
     /* Position cursor at the matching key */
     *elem = NULL;
     dbop = DBI_OP_MOVE_NEAR_DATA;
+
 retry_get:
+    if (*elem) {
+        /* This is a retry, so free the previous elem's data */
+        slapi_ch_free((void**)&prev_elem_data);
+    }
     rc = dblayer_cursor_op(&ctx->cursor, dbop, key, data);
+    prev_elem_data = data->data;  /* save pointer to data so we can free it on a retry */
     *elem = (rdn_elem *)data->data;
     dblayer_value_init(ctx->be, data);
 
@@ -1591,6 +1598,7 @@ retry_get:
     }
     if (*elem && RDN_IS_REDIRECT(*elem)) {
         rc = _entryrdn_resolve_redirect(ctx, elem, 1);
+        prev_elem_data = *elem;
         if (rc) {
             _ENTRYRDN_DEBUG_GOTO_BAIL();
             goto bail;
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -1570,12 +1576,15 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
                                          key->data, data->size, data->ulen, rc);
         }
         _ENTRYRDN_DEBUG_GOTO_BAIL();
+        slapi_ch_free((void**)elem);
```
progier389: This will prevent the log of the rdn with the error 
   slapi_log_err(SLAPI_LOG_TRACE, "_entryrdn_get_elem", "<-- _entryrdn_get_elem (*elem rdn=%s) rc=%d\n",
                      RDN_ADDR(*elem), rc); 
 IMHO we should rather call if (rc) { slapi_ch_free((void**)elem); } after login the error in the bail section.

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -1695,12 +1711,18 @@ _entryrdn_get_tombstone_elem(entryrdn_db_ctx_t *ctx,
             if (0 == strcmp(comma + 1, slapi_rdn_get_nrdn(srdn))) {
                 /* found and done */
                 _entryrdn_dup_rdn_elem((const void *)dataret.data, elem);
+                if (RDN_IS_REDIRECT(childelem)) {
+                    slapi_ch_free((void **)&childelem);
```
progier389: this sounds suspicious because  childelem is an array elems ...

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -1544,8 +1545,13 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
     /* Position cursor at the matching key */
     *elem = NULL;
     dbop = DBI_OP_MOVE_NEAR_DATA;
+
 retry_get:
+    if (*elem) {
+        slapi_ch_free((void**)&elem_data);
+    }
     rc = dblayer_cursor_op(&ctx->cursor, dbop, key, data);
+    elem_data = data->data;  /* save pointer to data so we can free it on a retry */
```
progier389: would not be slapi_ch_free((void**)elem);  sufficient ?

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
mreynolds389: > this sounds suspicious because childelem is an array elems ...

It fixes the leak.  Note childelem and dataret.data point to the same address.  I just chose to use childelem.

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
mreynolds389: No because elem ptr is changed after we grab the original ptr address (when a retry occurs for a long rdn).  So we need to free the previous value (elem_data).  I'm reworking the variable name to be "prev_elem_data" so it's more clear.

mreynolds389: TBH this was a really tricky set of leaks to fix.  Took me a long time to finalize the fix.

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -1544,8 +1545,14 @@ _entryrdn_get_elem(entryrdn_db_ctx_t *ctx,
     /* Position cursor at the matching key */
     *elem = NULL;
     dbop = DBI_OP_MOVE_NEAR_DATA;
+
 retry_get:
+    if (*elem) {
+        /* free the previous elem data */
+        slapi_ch_free((void**)&prev_elem_data);
+    }
```
droideck: I wonder if we also need to check and free it at the `bail:` section

mreynolds389: > I wonder if we also need to check and free it at the `bail:` section

Maybe :-)  Like I said this code is a bit tricky when it comes to handling long DN's vs not long DN's.  I'll try and force some failures and see what happens...

progier389: > Maybe :-) Like I said this code is a bit tricky when it comes to handling long DN's vs not long DN's. I'll try and force some failures and see what happens...
That is a bit worrisome: I have the feeling that neither of us really understand the full impact of this change. 
 (I certainly don't ! :frowning:  )

My gut feeling is that things should not be so tricky.
Either I missed something when designing that code (quite possible as long dn handling was done as an afterthought)  or we are not freeing Redirect rdn at the right place. (I wonder if in _entryrdn_resolve_redirect
```
         if (canfree) {
            slapi_ch_free((void**)elem);
        }
```
should not be moved around the end of the function (so that the redirect key get also freed in error case)
Or maybe the canfree flag is not set properly when calling _entryrdn_resolve_redirect ...

mreynolds389:
@progier389 @droideck  ok I have revised the fix after forcing failures at all the failure points.  This is did expose a flaw in the last patch (minor), but it allowed for a slightly more concise fix.  This is ready for review now.

droideck (approved):
Looks good to me!
I'll leave the final check to @progier389 though, as he's more proficient with the area.

progier389 (approved):
LGTM

Issue #6224, opened 2024-06-12
d2entry - Could not open id2entry err 0 - at startup when having sub-suffixes


progier389:
**Issue Description**
Description of problem:
Sub suffix causes the following error when the Directory Server starts
~~~
```
ERR - id2entry - Could not open id2entry err 0
ERR - dn2entry_ext - The dn "ou=test,dc=example,dc=com" was in the entryrdn index, but it did not exist in id2entry of instance test.
```
~~~

**Package Version and Platform:**
 - Platform: RHEL
 - Package and version: 389-ds-base-2.2.7-3.module+el9dsrv+18864+4949f8c5.x86_64

**Steps to Reproduce**
1. Install 389-ds-base-2.2.7-3.module+el9dsrv+18864+4949f8c5.x86_64
2. Set up as follows

```
DHOST=$(hostname -s)
cat << EOF > /root/$DHOST.inf
[general]
[slapd]
instance_name = $DHOST
root_password = password1
[backend-userroot]
create_suffix_entry = True
suffix = dc=example,dc=com
EOF
dscreate from-file /root/$DHOST.inf
dsconf $DHOST backend create --suffix="ou=test,dc=example,dc=com" --be-name="test" --parent-suffix="dc=example,dc=com" --create-suffix

```
3. Restart the Directory Server and find the ERRs.
```
dsctl $DHOST restart
grep -A1 "Could not open id2entry" /var/log/dirsrv/slapd-$DHOST/errors

```

**Expected results**
No error.

**Additional context**
RHEL Bug:   https://issues.redhat.com/browse/DIRSRV-53

Triaged: https://issues.redhat.com/browse/IDMDS-4306

progier389:
A minor issue related to the internal search that search for referral. 
I think that we should limit the search to a single backend

progier389:
0f46d4339..4e3dc9e88  main -> main
   8e6ce71d3..ae8ab2141  389-ds-base-3.0 -> 389-ds-base-3.0
   dc0d81844..0f35be137  389-ds-base-2.5 -> 389-ds-base-2.5
   9501c34df..43057ebc3  389-ds-base-2.4 -> 389-ds-base-2.4
   99c01b8b0..47672fb37  389-ds-base-2.3 -> 389-ds-base-2.3
   f9c420d11..560b13273  389-ds-base-2.2 -> 389-ds-base-2.2

progier389:
[...]

PR #6456, opened 2024-12-19
Issue 6224 - Remove test_referral_subsuffix from ds_logs_test.py

mmatsuya:
Bug Description:

test_referral_subsuffix test was removed from main branch and some other ones for higher versions. But, it was not removed from 389-ds-base-1.4.3 and 389-ds-base-2.1. The test doesn't work anymore with the fix for Issue 6224, because the added new control limited one backend for internal search. The test should be removed.

Fix Description:

remove the test from ds_logs_test.py

relates: https://github.com/389ds/389-ds-base/issues/6224


tbordaz (approved):
LGTM

PR #6460, opened 2024-12-20
Issue 6417 - (2nd) If an entry RDN is identical to the suffix, then E…

### LINKED ISSUE #6417: If an entry RDN is identical to the suffix, then Entryrdn gets broken during a reindex

tbordaz:
…ntryrdn gets broken during a reindex

Bug description:
	The primary fix has a flaw as it assumes that the
	suffix ID is '1'.
	If the RUV entry is the first entry of the database
	the server loops indefinitely

Fix description:
	Read the suffix ID from the entryrdn index

fixes: #6417

Reviewed by: Pierre Rogier (reviewed the first fix)

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
@@ -992,6 +992,7 @@ entryrdn_lookup_dn(backend *be,
     ID workid = id; /* starting from the given id */
     rdn_elem *elem = NULL;
     int maybesuffix = 0;
+    ID suffix_id = 1;
 
     slapi_log_err(SLAPI_LOG_TRACE, "entryrdn_lookup_dn",
                   "--> entryrdn_lookup_dn\n");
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -992,6 +992,7 @@ entryrdn_lookup_dn(backend *be,
     ID workid = id; /* starting from the given id */
     rdn_elem *elem = NULL;
     int maybesuffix = 0;
+    ID suffix_id;
```
progier389: Style: We usually initialize the variable at declaration time. 
 ID suffix_id = 1;

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -1032,6 +1033,20 @@ entryrdn_lookup_dn(backend *be,
...
+    if (rc) {
+        slapi_log_err(SLAPI_LOG_WARNING, "entryrdn_lookup_dn",
+                  "Fails to retrieve the ID of suffix %s - Assume it is '1'\n", slapi_sdn_get_ndn(be->be_suffix));
+        suffix_id = 1;
+    } else {
+        elem = (rdn_elem *)data.data;
+        suffix_id = id_stored_to_internal(elem->rdn_elem_id);
+        dblayer_value_free(be, &data);
+    }
+
```
progier389, line 1051: Memory leak: I think that both
```
dblayer_value_free(be, &data);
dblayer_value_free(be, &key);
```
are needed (whether rc==0 or not)

progier389 (changes requested):

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
tbordaz: Nice catch. Thanks.
I updated the fix

progier389 (approved):
LGTM

PR #6467, opened 2025-01-01
Issue 6368 - Fix crash when testing access json logging

### LINKED ISSUE #6368: Add option to write access log in JSON

mreynolds389:
Description:

An uninitialized pointer led to a crash when getting the bind dn from the pblock and then freeing it.

Fixes: https://github.com/389ds/389-ds-base/issues/6368


progier389 (approved):
LGTM

Issue #6468, opened 2025-01-02
RFE - CLI - Add logging features to dsconf


mreynolds389:
**Issue Description**

It would be nice to have a more friendly CLI interface for managing the server logging.  Specifically the log levels.

mreynolds389:
0e0140b5d..9c5d69c56  389-ds-base-3.0 -> 389-ds-base-3.0
c8eda6f92..e3fe9949a  389-ds-base-2.6 -> 389-ds-base-2.6
3477ba258..7a6130b4d  389-ds-base-2.5 -> 389-ds-base-2.5
ea04b0033..7a3de5ee4  389-ds-base-2.4 -> 389-ds-base-2.4
0cfe37159..384f5afe8  389-ds-base-2.3 -> 389-ds-base-2.3
ab9fe5160..260cfe7ae  389-ds-base-2.2 -> 389-ds-base-2.2

Just realized that this commit includes the access json logging settings, which do not exist in 2.x, so I will clean that up next

mreynolds389:
Remove access json settings as they don't apply

e3fe9949a..11670a36f  389-ds-base-2.6 -> 389-ds-base-2.6
7a6130b4d..1cf55fb6d  389-ds-base-2.5 -> 389-ds-base-2.5

Remove access and audit json settings as they don't apply

7a3de5ee4..f44d9997f  389-ds-base-2.4 -> 389-ds-base-2.4
384f5afe8..9c4ef7e95  389-ds-base-2.3 -> 389-ds-base-2.3
260cfe7ae..5333a3d56  389-ds-base-2.2 -> 389-ds-base-2.2

Remove wrong security log settings

a7f7cd944..6aac203ac  main
9c5d69c56..0de8c2144  389-ds-base-3.0 -> 389-ds-base-3.0
11670a36f..444354b1f  389-ds-base-2.6 -> 389-ds-base-2.6
1cf55fb6d..2ecd3921e  389-ds-base-2.5 -> 389-ds-base-2.5

mreynolds389:
Some platforms do not support python 3.10 so the use "match" will not work.  So some tweaks might be needed, reopening ticket for awareness...

vashirov:
> Some platforms do not support python 3.10 so the use "match" will not work. So some tweaks might be needed, reopening ticket for awareness...

Yeah, it fails to build in RHEL 9, I opened https://github.com/389ds/389-ds-base/pull/6556.

vashirov:
[...]

PR #6469, opened 2025-01-02
Issue 6468 - RFE - CLI - Add logging settings to dsconf

mreynolds389:
Description:

It would be nice to have a more friendly CLI interface for managing the server logging. Specifically the log levels. This PR adds a new subcommand "logging" that can handle each log type and all the settings for it.

relates: https://github.com/389ds/389-ds-base/issues/6468

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsconf_logging.py
+++ b/dirsrvtests/tests/suites/clu/dsconf_logging.py
@@ -1,0 +1,168 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import json
+import subprocess
+import logging
+import pytest
+from lib389._constants import DN_DM
+from lib389.topologies import topology_st as topo
+
+pytestmark = pytest.mark.tier1
+
+log = logging.getLogger(__name__)
+
+SETTINGS = [
+    ('logging-enabled', None),
+    ('logging-disabled', None),
+    ('mode', '700'),
+    ('compress-enabled', None),
+    ('compress-disabled', None),
+    ('buffering-enabled', None),
+    ('buffering-disabled', None),
+    ('max-logs', '4'),
+    ('max-logsize', '7'),
+    ('rotation-interval', '2'),
+    ('rotation-interval-unit', 'week'),
+    ('rotation-tod-enabled', None),
+    ('rotation-tod-disabled', None),
+    ('rotation-tod-hour', '12'),
+    ('rotation-tod-minute', '20'),
+    ('deletion-interval', '3'),
+    ('deletion-interval-unit', 'day'),
+    ('max-disk-space', '20'),
+    ('free-disk-space', '2'),
+]
+
+DEFAULT_TIME_FORMAT = "%FT%TZ"
+
+
+def execute_dsconf_command(dsconf_cmd, subcommands):
+    """Execute dsconf command and return output and return code"""
+
+    cmdline = dsconf_cmd + subcommands
+    proc = subprocess.Popen(cmdline, stdout=subprocess.PIPE)
+    out, _ = proc.communicate()
+    return out.decode('utf-8'), proc.returncode
+
+
+def get_dsconf_base_cmd(topo):
+    """Return base dsconf command list"""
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/clu/dsconf_logging.py
```diff
@@ -0,0 +1,164 @@
...
+def get_dsconf_base_cmd(topo):
+    """Return base dsconf command list"""
+    return ['/usr/sbin/dsconf', topo.standalone.serverid,
+            '-j', '-D', DN_DM, '-w', 'password', 'logging']
+
+
+def test_log_settings(topo):
+    """Test each log settings can be set successfully
+
+    :id: ffc912a6-c188-413d-9c35-7f4b3774d946
```
droideck: duplicate id

droideck (changes requested):
I think it should be possible to create helper functions to generate common parsers, reducing duplication. Most of the attributes repeat across different log types.

Something like?..
For all of them in a `for loop` (this part will really reduce the redundancy):
```
    _add_enable_disable_parsers(set_parsers, log_type)
    _add_location_parser(set_parsers, log_type)
    _add_compression_parsers(set_parsers, log_type)
    etc..
```

And the rest for special cases where we can't use `for loop` for all of them.

Another thing:
I think, we need to add a bit more validation where it's possible (time format and sizes?)

mreynolds389:
> Another thing: I think, we need to add a bit more validation where it's possible (time format and sizes?)

We already do the validation in the core server (libglobs.c).   I don't see the need to duplicate it in the CLI.  The CLI will correctly report the error message from the server.

dirsrvtests/tests/suites/clu/dsconf_logging.py
mreynolds389: Yeah meant to fix that but forgot.  This was a side project of mine that I've _slowly_ been working on for the last few months.

droideck:
> > Another thing: I think, we need to add a bit more validation where it's possible (time format and sizes?)
> 
> We already do the validation in the core server (libglobs.c). I don't see the need to duplicate it in the CLI. The CLI will correctly report the error message from the server.

Yeah, I agree. I just worry that UI depends on `desc` field. And when we get `Operations error`, it is the only thing that is displayed in UI. (i.e. `Error saving Audit Log settings - Operations error`)

Also, CLI output looks a bit not clean:

```
# dsconf inst  logging audit set max-logsize 2147483647

Error: 103 - 3 - 1 - Operations error - [] - nsslapd-auditlog-logmaxdiskspace: maxdiskspace "100 (MB)" is less than max log size "2147483647 (MB)" - modify_ext_s(('cn=config', [(2, 'nsslapd-auditlog-maxlogsize', [b'2147483647'])]),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance inst
```

IMHO, it'll be nice to have a clean `Error: nsslapd-auditlog-logmaxdiskspace: maxdiskspace "100 (MB)" is less than max log size "2147483647 (MB)"` formatted error in CLI and UI.

But we can get the validation error from the exception and parse it correctly in UI and CLI. And we can deal with it in a different ticket, probablly.

So, yeah, discard this concern. :)

mreynolds389:
> Also, CLI output looks a bit not clean:
> 
> ```
> # dsconf inst  logging audit set max-logsize 2147483647
> 
> Error: 103 - 3 - 1 - Operations error - [] - nsslapd-auditlog-logmaxdiskspace: maxdiskspace "100 (MB)" is less than max log size "2147483647 (MB)" - modify_ext_s(('cn=config', [(2, 'nsslapd-auditlog-maxlogsize', [b'2147483647'])]),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance inst
> ```
> 

Ugh that is ugly

> IMHO, it'll be nice to have a clean `Error: nsslapd-auditlog-logmaxdiskspace: maxdiskspace "100 (MB)" is less than max log size "2147483647 (MB)"` formatted error in CLI and UI.
> 
> But we can get the validation error from the exception and parse it correctly in UI and CLI. And we can deal with it in a different ticket, probablly.

Hmm, yeah we use generic error handling in the CLI.  It was never clean.  I think it would be nice to improve it - perhaps that will be my next side project :-)

mreynolds389:
@droideck Requested changes made.  I just lumped all the settings into a single loop instead of having separate functions for common parsers

droideck (approved):
Looks great! Ack

PR #6471, opened 2025-01-03
Issue 6470 - Some replication status data are reset upon a restart

### LINKED ISSUE #6470: Some replication status data are reset upon a restart
**Issue Description**
The replication agreement contains operational attributes related to the total init
- nsds5replicaLastInitStart
- nsds5replicaLastInitEnd
- nsds5replicaLastInitStatus

Those attributes are reset after restarting DS. So the consumer may erroneously appear not yet initialized and
we are loosing the time when it was initialized.

 - all versions

**Steps to Reproduce**
testcase provided

**Expected results**
The total init date (start/end) and the total init status should be preserved during restart

tbordaz:
Bug description:
	The replication agreement contains operational attributes
	related to the total init: nsds5replicaLastInitStart,
	nsds5replicaLastInitEnd, nsds5replicaLastInitStatus.
	Those attributes are reset at restart

Fix description:
	When reading the replication agreement from config
	(agmt_new_from_entry) restore the attributes into
	the in-memory RA.
	Updates the RA config entry from the in-memory RA
	during shutdown/cleanallruv/enable_ra

fixes: #6470

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/replication/single_master_test.py
+++ b/dirsrvtests/tests/suites/replication/single_master_test.py
@@ -13,6 +13,7 @@ from lib389.utils import *
 from lib389.idm.user import UserAccounts, TEST_USER_PROPERTIES
 
 from lib389.replica import ReplicationManager, Replicas
+from lib389.agreement import Agreements
 from lib389.backend import Backends
 
 from lib389.topologies import topology_m1c1 as topo_r # Replication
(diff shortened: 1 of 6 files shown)


dirsrvtests/tests/suites/replication/single_master_test.py
```diff
@@ -156,6 +157,136 @@ def test_lastupdate_attr_before_init(topo_nr):
...
+    assert post_restart_initEnd_2 == initEnd_2
+
+    # Check that values are updated by total init
+    log.info("Check values from first/second total init are different")
+    assert status_2 == status_1
+    assert initStart_2 != initStart_1
+    assert initEnd_2 != initEnd_1
+
+
+
```
droideck: Another new line we can live without:)

dirsrvtests/tests/suites/replication/single_master_test.py
```diff
@@ -156,6 +157,136 @@ def test_lastupdate_attr_before_init(topo_nr):
...
+    agmts_supplier = Agreements(supplier, replica_supplier.dn)
+    supplier_consumer = None
+    for agmt in agmts_supplier.list():
+        if (agmt.get_attr_val_utf8('nsDS5ReplicaPort') == str(consumer.port) and
+           agmt.get_attr_val_utf8('nsDS5ReplicaHost') == consumer.host):
+            supplier_consumer = agmt
+            break
+    assert supplier_consumer
+
+
```
droideck: Rogue new line:)

droideck (approved):
Couple of minor points, the rest looks good!

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -2527,6 +2551,108 @@ agmt_set_consumer_ruv(Repl_Agmt *ra, RUV *ruv)
...
+
+    if (ra->last_init_start_time) {
+        nb_mods++;
+    }
+    if (ra->last_init_end_time) {
+        nb_mods++;
+    }
+    if (ra->last_init_status[0] != '\0') {
+        nb_mods++;
+    }
```
droideck, line 2577:
```suggestion
    }
    if (nb_mods == 0) {
        PR_Unlock(ra->lock);
        return;
    }
```


I think it's better to `return` here if there are no mods. 
It feels a bit strange to check it at the end of the function and then call `slapi_mod_done(*);` even though no mods were done...

tbordaz: Good point. Thanks

Issue #6372, opened 2024-10-21
dsconf backup create hangs on a container created with 389ds/dirsrv:3.1 docker image


mayurcrunchr:
**Issue Description**

We have a cronjob that runs the following command to backup the ldap database online every hour at `55 minutes( 55 * * * *) `
`"docker exec -i ldap dsconf -D "cn=Directory Manager" localhost backup create /tmp/ansible.1bu66l05/db"`

The job takes backup successfuly and finishes successfully for few hours(23 hours in this case) and then hangs suddenly while copying backup files (after 24 iterations or 24 hours in this case)

` "[20/Oct/2024:22:55:33.029998779 +0000] - INFO - archive_copyfile - Copying /etc/dirsrv/slapd-localhost/slapd-collations.conf to /tmp/ansible.1bu66l05/db/config_files/slapd-collations.conf"`

We are using the docker image 389ds/dirsrv:3.1 on a Ubuntu 20.04.6 LTS host

**Package Version and Platform:**
 - Platform: openSUSE Tumbleweed , release: 20240726
 - Package and version: [389-ds  3.1.1~git0.aef1668-180.1 x86_64]

**Steps to Reproduce**
1. Create a ldap docker container using docker image 389ds/dirsrv:3.1 on a ubuntu 20.04.6 LTS host
2. Add a backend and set the nsslapd-backend-implement to bdb(we have also observed the same problem on fresh mdb database)
3. Create a test group 
4. Create a test user and add the test user to the test group created in step 3
5. Enable the cronjob  to backup the ldap db every hour at 55 minutes
`55 * * * * docker exec -i ldap dsconf -D "cn=Directory Manager" localhost backup create /tmp/ansible.1bu66l05/db`
6. After few hours(24 hours in our case), the dsconf backup hangs

**Screenshots**
![image](https://github.com/user-attachments/assets/274bce93-8eb0-4ac3-8fef-f543d7cede0f)
![image](https://github.com/user-attachments/assets/14979091-a08c-47fd-b6ab-a2e9b22823d5)
![image](https://github.com/user-attachments/assets/b6c0084c-7a64-4f86-ba5e-89f5e98476d2)

**Additional context**
Please find attached the backend config, nsslapd config params, access logs, error logs, db_stat output, gdb backtrace for ns-slapd, dsconf and dscontainer processes.

[...]

PR #6475, opened 2025-01-06
Issue 6372 - Deadlock while doing online backup

progier389:
Sometime server hangs during online backup because of a deadlock due to lock order inversion between the dse backup_lock mutex and the dse rwlock.

Solution:
  Add functions to manage the lock/unlock to ensure consistency.
  Ensure that threads always tries to lock dse_backup_lock mutex before the dse write lock
Code cleanup:
  - Move the backup_lock into the dse struct
  - Avoid the obsolete warning during tests (I think that we will have to do a second cleanup phase later to see if we could not replace self.conn.add_s by self.create .
  
  Issue: #6372 
  
  Closing: #6372 
  
  Reviewed by: @mreynolds389 (Thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/backups/backup_test.py
+++ b/dirsrvtests/tests/suites/backups/backup_test.py
@@ -17,15 +17,19 @@ import ldap
 from datetime import datetime
 from lib389._constants import DN_DM, PASSWORD, DEFAULT_SUFFIX, INSTALL_LATEST_CONFIG
 from lib389.properties import BACKEND_SAMPLE_ENTRIES, TASK_WAIT
-from lib389.topologies import topology_st as topo, topology_m2 as topo_m2
+from lib389.topologies import topology_st as topo, topology_m2 as topo_m2, set_timeout
 from lib389.backend import Backends, Backend
 from lib389.dbgen import dbgen_users
 from lib389.tasks import BackupTask, RestoreTask
 from lib389.config import BDB_LDBMConfig
+from lib389.idm.nscontainer import nsContainers
 from lib389 import DSEldif
 from lib389.utils import ds_is_older, get_default_db_lib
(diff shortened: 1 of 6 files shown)


mreynolds389 (approved):
Very nice, ack.  I know that historically in DS we use "int" for boolean purposes, but using a PRBool for "dse_backup_in_progress" would be nice.  I'm not asking for it, but it's something to consider.

progier389:
>  I know that historically in DS we use "int" for boolean purposes, but using a PRBool for "dse_backup_in_progress" would be nice. 
I have no strong opinion about that one. In fact I hesitated before choosing 0/1 !
I wonder if we should not rather use the C11 bool type and true/false (I think that these these days most compiler support it  and this looks more logic for the part of code that is independent of nspr ...) but once again I have no strong opinion ...

mreynolds389:
> > I know that historically in DS we use "int" for boolean purposes, but using a PRBool for "dse_backup_in_progress" would be nice.
> > I have no strong opinion about that one. In fact I hesitated before choosing 0/1 !
> > I wonder if we should not rather use the C11 bool type and true/false (I think that these these days most compiler support it  and this looks more logic for the part of code that is independent of nspr ...) but once again I have no strong opinion ...

Yeah I feel nspr is becoming less and less "needed" - if not already obsolete.  It would actually be nice to completely remove it from DS as some point :-)  So lets do this, lets use the C11 bool type here, so it's in the code for all to see, and then we can all start using this moving forward.

progier389:
Push the change with C11 bool

vashirov:
@progier389 after this commit backup tests started to fail in 3.0:
https://github.com/389ds/389-ds-base/actions/runs/12650325366/job/35250352696

I will run more tests with the sanitized build.

PR #6480, opened 2025-01-08
Issue 6417 - (3rd) If an entry RDN is identical to the suffix, then E…

### LINKED ISSUE #6417: If an entry RDN is identical to the suffix, then Entryrdn gets broken during a reindex

tbordaz:
…ntryrdn gets broken during a reindex

Bug description:
	The previous fix had a flaw.
	In case entryrdn_lookup_dn is called with an undefined suffix
	the lookup of the suffix trigger a crash.
	For example it can occur during internal search of an
	unexisting map (view plugin).
	The issue exists in all releases but is hidden since 2.3.

Fix description:
	testing the suffix is defined

fixes: #6417

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
@@ -1033,21 +1033,25 @@ entryrdn_lookup_dn(backend *be,
     dblayer_value_free(be, &data);
     dblayer_value_init(be, &data);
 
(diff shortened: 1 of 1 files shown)


progier389 (commented):
LGTM

progier389 (approved):

PR #6482, opened 2025-01-08
Issue 6481 - UI - When ports that are in use are used to update a DS instance the error message is not helpful

### LINKED ISSUE #6481: UI - When ports that are in use are used to update a DS instance the error message is not helpful
**Issue Description**
When ports that are in use are used to update a DS instance the error message is not helpful

**Steps to Reproduce**
1. Create 2 DS Instances on the Webconsole
2. Update the second instance to use the ports that are currently in use by the
1st instance
3. Follow the prompt to restart the instance for the changes to take effect.
4. Error displayed:
Screenshot from 2025-01-07 13-26-34

**Expected results**
The user should not be allowed to use a port number that is already in use

jchapma:
Bug description:
When updating port values on a DS instance, if the port value is already in use the error message displayed by the UI is not helpful.

Fix description:
Add a UI method that checks if the updated port value is already in use. If it is, disable the save button.

Fixes: https://github.com/389ds/389-ds-base/issues/6481

Reviewed by:

### DIFF
--- a/src/cockpit/389-console/src/lib/server/settings.jsx
+++ b/src/cockpit/389-console/src/lib/server/settings.jsx
@@ -1,6 +1,6 @@
 import cockpit from "cockpit";
 import React from "react";
-import { log_cmd, valid_dn, isValidIpAddress } from "../tools.jsx";
+import { log_cmd, valid_dn, isValidIpAddress, is_port_in_use } from "../tools.jsx";
 import {
 	Button,
 	Checkbox,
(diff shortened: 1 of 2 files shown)


mreynolds389 (changes requested):
You need to move the validation code from the state change function to validateSaveBtn() function, otherwise the NumberInput will not have the validated prop set correctly.

jchapma:
> You need to move the validation code from the state change function to validateSaveBtn() function, otherwise the NumberInput will not have the validated prop set correctly.

Initially I had the validation code in the validateSaveBtn function, then if the user was say decrementing to a port value and came across a port that was in use the -- button was disabled and the user couldn't go any lower. So I thought a better way was to do the validation in minus/plus/change functions, set the errObj for ('nsslapd-port') as true if the port was in use, then check errObj in validateSaveBtn. Or am i missing something ?

droideck:
> Initially I had the validation code in the validateSaveBtn function, then if the user was say decrementing to a port value and came across a port that was in use the -- button was disabled and the user couldn't go any lower. So I thought a better way was to do the validation in minus/plus/change functions, set the errObj for ('nsslapd-port') as true if the port was in use, then check errObj in validateSaveBtn. Or am i missing something ?

I think `validateSaveBtn` should disable `Save Settings` button, and it should not affect `plus` or `minus` buttons...

jchapma:
> > Initially I had the validation code in the validateSaveBtn function, then if the user was say decrementing to a port value and came across a port that was in use the -- button was disabled and the user couldn't go any lower. So I thought a better way was to do the validation in minus/plus/change functions, set the errObj for ('nsslapd-port') as true if the port was in use, then check errObj in validateSaveBtn. Or am i missing something ?
> 
> I think `validateSaveBtn` should disable `Save Settings` button, and it should not affect `plus` or `minus` buttons...

Ok, let me move the validation back to validateSaveBtn

mreynolds389:
Almost there :-)

This diff gets the NumberInput working as expected

```
diff --git a/src/cockpit/389-console/src/lib/server/settings.jsx b/src/cockpit/389-console/src/lib/server/settings.jsx
index 5de56dac6..568fd1a53 100644
--- a/src/cockpit/389-console/src/lib/server/settings.jsx
+++ b/src/cockpit/389-console/src/lib/server/settings.jsx
@@ -311,6 +311,10 @@ async validateSaveBtn(nav_tab, attr, value) {
                         const portInUse = await is_port_in_use(portValue);
                         if (portInUse) {
                             disableSaveBtn = true;
+                            if (value !== this.state['_' + attr]) {
+                                // New value, but it's already being used
+                                valueErr = true;
+                            }
                         }
                     } catch (error) {
                         console.error("Error checking port:", error);
```

mreynolds389:
Sorry another issue...  I also noticed that the "minus" buttons are disabled for port and secure port.  That should be fixed as well.  Thanks!

jchapma:
> Sorry another issue... I also noticed that the "minus" buttons are disabled for port and secure port. That should be fixed as well. Thanks!

The issue of the minus button being disabled existed before I started this work, I had a look at it yesterday and it seems the values of nsslapd-port and nsslapd-secureport are undefined at render time and only defined after loadConfig() is called in componentDidMount(). Let me have another look at it.

mreynolds389:
> > Sorry another issue... I also noticed that the "minus" buttons are disabled for port and secure port. That should be fixed as well. Thanks!
> 
> The issue of the minus button being disabled existed before I started this work, I had a look at it yesterday and it seems the values of nsslapd-port and nsslapd-secureport are undefined at render time and only defined after loadConfig() is called in componentDidMount(). Let me have another look at it.

Sorry I was not saying it was your fault, but might as well fix it while you are working on that code.

jchapma:
> > > Sorry another issue... I also noticed that the "minus" buttons are disabled for port and secure port. That should be fixed as well. Thanks!
> > 
> > 
> > The issue of the minus button being disabled existed before I started this work, I had a look at it yesterday and it seems the values of nsslapd-port and nsslapd-secureport are undefined at render time and only defined after loadConfig() is called in componentDidMount(). Let me have another look at it.
> 
> Sorry I was not saying it was your fault, but might as well fix it while you are working on that code.

Ye I knew you meant fix it while I was working on that area of code, thanks anyway though. 

I had a look around and found a few more disabled "minus" button issues, how about I create an issue for these and fix them all together ?

Server
Server Settings -> General Settings -> LDAP Port, LDAPS Port, 
Server Settings -> Disk Monitoring -> Disk Monitoring Threshold, Disk Monitoring Grace Period
Tuning and Limits -> Idle Connection Timeout, I/O Block Timeout
Logging -> Access Log Settings -> Maximum Number Of Logs
Logging -> Audit Log Settings -> Maximum Number Of Logs
Logging -> Audit Fail Log Settings  -> Maximum Number Of Logs
Logging -> Error Log Settings  -> Maximum Number Of Logs
Logging -> Security Log Settings  -> Maximum Number Of Logs

Database
Global Database Configuration -> Limits -> ID List Scan Limit
Global Database Configuration -> NDN Cache -> Normalized DN Cache Max Size
Global Database Configuration -> Advanced Settings -> Database Max DBs

mreynolds389 (approved):
Yeah, since there are other places where the minus button is the same just open a new ticket then.  Thanks, and ack!

droideck (approved):
Looks good!

Issue #6483, opened 2025-01-08
CI - Fix filter test - test_match_large_valueset


mreynolds389:
**Issue Description**

In test_match_large_valueset() the call to dbgen_users was using the wrong parameters.  The suffix and parent parameters were switched

mreynolds389:
f58023bd7..790d43177  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6484, opened 2025-01-08
Issue 6483 - CI - fix filter_test - test_match_large_valueset

mreynolds389:
Description:

Fix filter_test::test_match_large_valueset

The suffix parameter to dbgen_users was using the wrong DN (user ou versus DEFAULT suffix)

Relates: https://github.com/389ds/389-ds-base/issues/6483

### DIFF
--- a/dirsrvtests/tests/suites/filter/filter_test.py
+++ b/dirsrvtests/tests/suites/filter/filter_test.py
@@ -1,13 +1,13 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2020 Red Hat, Inc.
+# Copyright (C) 2020-2025 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 #
+import ldap
 import logging
-
 import pytest
 import time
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/filter/filter_test.py
```diff
@@ -373,15 +373,16 @@ def test_match_large_valueset(topology_st):
             config_ldbm = DSLdapObject(inst, DN_CONFIG_LDBM)
             config_ldbm.set('nsslapd-cache-autosize', '0')
         be_groups.replace('nsslapd-cachememsize', groups_entrycache)
-    except:
-        raise
+    except Exception as e:
+        raise e
```
droideck: I think this part is completely redundant...
It basically just catches and immediately re-raises the same exception without performing any additional actions.

Even though it's better than bare `except: raise` - still, I'd just remove it:)

droideck (approved):
Besides the minor concern, - Looks good!

dirsrvtests/tests/suites/filter/filter_test.py
mreynolds389: I was just fixing lint errors, I can take it out.

tbordaz (approved):
LGTM

Issue #6485, opened 2025-01-08
Fix double free in USN cleanup task


mreynolds389:
**Issue Description**

ASAN report shows double free of bind dn in the USN cleanup task data.  The bind dn was passed as a reference so it should never have to be freed by the cleanup task.

mreynolds389:
790d43177..90163fb54  389-ds-base-3.0 -> 389-ds-base-3.0
b49474d41..41d5e9a41  389-ds-base-2.6 -> 389-ds-base-2.6
456a07da4..afce4f176  389-ds-base-2.5 -> 389-ds-base-2.5
6a586990d..2ddd04e6b  389-ds-base-2.4 -> 389-ds-base-2.4
453ca5233..20d41f800  389-ds-base-2.3 -> 389-ds-base-2.3
caf5316c9..82339aa3d  389-ds-base-2.2 -> 389-ds-base-2.2
5ccb60646..0a0f1888d  389-ds-base-2.1 -> 389-ds-base-2.1
34f0f2326..f5880862e  389-ds-base-2.0 -> 389-ds-base-2.0
9846b17ba..04df2c8df  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6486, opened 2025-01-08
Issue 6485 - Fix double free in USN cleanup task

mreynolds389:
Description:

ASAN report shows double free of bind dn in the USN cleanup task data. The bind dn was passed as a reference so it should never have to be freed by the cleanup task.

Relates: https://github.com/389ds/389-ds-base/issues/6485

### DIFF
--- a/ldap/servers/plugins/usn/usn_cleanup.c
+++ b/ldap/servers/plugins/usn/usn_cleanup.c
@@ -240,7 +240,7 @@ usn_cleanup_add(Slapi_PBlock *pb,
     char *suffix = NULL;
     char *backend_str = NULL;
     char *maxusn = NULL;
-    char *bind_dn;
+    char *bind_dn = NULL;
     struct usn_cleanup_data *cleanup_data = NULL;
     int rv = SLAPI_DSE_CALLBACK_OK;
     Slapi_Task *task = NULL;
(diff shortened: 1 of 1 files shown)


tbordaz (commented):
Currently the bind_dn refers to the SLAPI_REQUESTOR_DN that was in the pblock when the task was spawn. Indeed it is freed when pblock_done and there is a double free in task_destructor. IMHO there is a risk of 'user after free' if  bind_dn after it has been free by pblock_done.

If I am not missing something, I would prefer that  the bind_dn is duplicated when creating the task data (usn_cleanup_add) and free during task_desctructor.

mreynolds389:
> Currently the bind_dn refers to the SLAPI_REQUESTOR_DN that was in the pblock when the task was spawn. Indeed it is freed when pblock_done and there is a double free in task_destructor. IMHO there is a risk of 'user after free' if bind_dn after it has been free by pblock_done.
> 
> If I am not missing something, I would prefer that the bind_dn is duplicated when creating the task data (usn_cleanup_add) and free during task_desctructor.

That crossed my mind, but I'm just always trying to avoid malloc/freeing when possible.  Anyway you are right.  There is a risk, although the window is small, it's still there.  Nice catch!  Fixed

tbordaz (approved):
LGTM

Issue #6509, opened 2025-01-16
Race condition with Paged Result searches causes a connection error T3


mreynolds389:
**Issue Description**

There is a race condition with Paged Result searches when a new operation comes in while a paged search is finishing.  This triggers an invalid time out error and closes the connection.

The problem is that we do not use the "PagedResult lock" when checking the connection's paged result data for a timeout event.  This causes the paged result timeout value to change unexpectedly and trigger a false timeout when a new operation arrives.

There are also memory leaks when an error occurs on a paged result search.

mreynolds389:
9d00b6d02..8773d9da1  389-ds-base-3.0 -> 389-ds-base-3.0
d6b11656a..541cd3836  389-ds-base-2.6 -> 389-ds-base-2.6
417efec1a..86dd51fd1  389-ds-base-2.5 -> 389-ds-base-2.5
e6eaa5c7a..59ffc64e3  389-ds-base-2.4 -> 389-ds-base-2.4
2767cfecd..fb1954f6d  389-ds-base-2.3 -> 389-ds-base-2.3
b0a3b38a0..ddd0ffab9  389-ds-base-2.2 -> 389-ds-base-2.2
318ed4170..4acbdc7c6  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6510, opened 2025-01-17
Issue 6509 - Race condition with Paged Result searches

mreynolds389:
Description:

There is a race condition with Paged Result searches when a new operation comes in while a paged search is finishing. This triggers an invalid time out error and closes the connection with a T3 code.

The problem is that we do not use the "PagedResult lock" when checking the connection's paged result data for a timeout event. This causes the paged result timeout value to change unexpectedly and trigger a false timeout when a new operation arrives.

This also fixes some memory leaks that occur when an error happens.

Relates: https://github.com/389ds/389-ds-base/issues/6509

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -1554,7 +1554,29 @@ setup_pr_read_pds(Connection_Table *ct, int listnum)
         if (c->c_state == CONN_STATE_FREE) {
             connection_table_move_connection_out_of_active_list(ct, c);
         } else {
-            /* we try to acquire the connection mutex, if it is already
+            /* Check for a timeout for PAGED RESULTS */
+            if (pagedresults_is_timedout_nolock(c)) {
+                /*
+                 * There could be a race condition so lets try again with the
+                 * right lock
+                 */
+                pthread_mutex_t *pr_mutex = pageresult_lock_get_addr(c);
+                if (pthread_mutex_trylock(pr_mutex) == EBUSY) {
+                    c = next;
+                    continue;
+                }
+                if (pagedresults_is_timedout_nolock(c)) {
+                    pthread_mutex_unlock(pr_mutex);
+                    disconnect_server(c, c->c_connid, -1,
+                                      SLAPD_DISCONNECT_PAGED_SEARCH_LIMIT,
+                                      0);
+                } else {
+                    pthread_mutex_unlock(pr_mutex);
+                }
+            }
+
+            /*
+             * we try to acquire the connection mutex, if it is already
              * acquired by another thread, don't wait
              */
             if (pthread_mutex_trylock(&(c->c_mutex)) == EBUSY) {
@@ -1562,35 +1584,24 @@ setup_pr_read_pds(Connection_Table *ct, int listnum)
                 continue;
             }
             if (c->c_flags & CONN_FLAG_CLOSING) {
-                /* A worker thread has marked that this connection
-                 * should be closed by calling disconnect_server.
-                 * move this connection out of the active list
-                 * the last thread to use the connection will close it
+                /*
+                 * A worker thread, or paged result timeout, has marked that
+                 * this connection should be closed by calling
+                 * disconnect_server(). Move this connection out of the active
+                 * list then the last thread to use the connection will close
+                 * it.
                  */
                 connection_table_move_connection_out_of_active_list(ct, c);
             } else if (c->c_sd == SLAPD_INVALID_SOCKET) {
                 connection_table_move_connection_out_of_active_list(ct, c);
             } else if (c->c_prfd != NULL) {
                 if ((!c->c_gettingber) && (c->c_threadnumber < c->c_max_threads_per_conn)) {
-                    int add_fd = 1;
-                    /* check timeout for PAGED RESULTS */
-                    if (pagedresults_is_timedout_nolock(c)) {
-                        /* Exceeded the paged search timelimit; disconnect the client */
-                        disconnect_server_nomutex(c, c->c_connid, -1,
-                                                  SLAPD_DISCONNECT_PAGED_SEARCH_LIMIT,
-                                                  0);
-                        connection_table_move_connection_out_of_active_list(ct,
-                                                                            c);
-                        add_fd = 0; /* do not poll on this fd */
-                    }
-                    if (add_fd) {
-                        ct->fd[listnum][count].fd = c->c_prfd;
-                        ct->fd[listnum][count].in_flags = SLAPD_POLL_FLAGS;
-                        /* slot i of the connection table is mapped to slot
-                         * count of the fds array */
-                        c->c_fdi = count;
-                        count++;
-                    }
+                    ct->fd[listnum][count].fd = c->c_prfd;
+                    ct->fd[listnum][count].in_flags = SLAPD_POLL_FLAGS;
+                    /* slot i of the connection table is mapped to slot
+                        * count of the fds array */
+                    c->c_fdi = count;
+                    count++;
                 } else {
                     if (c->c_threadnumber >= c->c_max_threads_per_conn) {
                         c->c_maxthreadsblocked++;
--- a/ldap/servers/slapd/opshared.c
+++ b/ldap/servers/slapd/opshared.c
@@ -251,7 +251,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
     char *errtext = NULL;
     int nentries, pnentries;
     int flag_search_base_found = 0;
-    int flag_no_such_object = 0;
+    bool flag_no_such_object = false;
     int flag_referral = 0;
     int flag_psearch = 0;
     int err_code = LDAP_SUCCESS;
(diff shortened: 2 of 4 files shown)


progier389 (approved):
LGTM

ldap/servers/slapd/opshared.c
```diff
@@ -914,12 +914,20 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                      * the error has already been sent
                      * stop the search here
                      */
+                    if (op_is_pagedresults(operation)) {
```
tbordaz: Is it needed ? We are in the switch/case handling an error of the backend search. Few lines above the search result was already 'pagedresults_set_search_result' to NULL . why not calling 'be->be_search_results_release(&sr)' few lines above ?

ldap/servers/slapd/pagedresults.c
```diff
@@ -924,19 +931,25 @@ pagedresults_is_timedout_nolock(Connection *conn)
 
     slapi_log_err(SLAPI_LOG_TRACE, "pagedresults_is_timedout", "=>\n");
 
+    pthread_mutex_lock(pageresult_lock_get_addr(conn));
```
tbordaz: I have concern with a comment in pageresult_lock_get_addr. it is said to not lock pageresult_lock while holding c_mutex. IIUC here we are holding c_mutex.

mreynolds389: This is required to fix the problem :-)  Other places in the code (pagedresult.c) do the exact same thing

ldap/servers/slapd/opshared.c
mreynolds389: It sets the backend to NULL (pagedresults_set_current_be) not the search result (pagedresults_set_search_result).  But that might work moving it there, but I was worried about the "flag_no_such_object" case were we want to try a different backend before quitting.  Might be safer to leave it as is?  WDYT?

ldap/servers/slapd/pagedresults.c
mreynolds389: Actually I'm not sure if other places do that, but this is the fix that worked for the customer.  I'll double check that...

mreynolds389: Well we also have a comment saying we can not call the the timedout function unless that lock is taken :-)

```
 *
 * must be called within pageresult_lock_get_addr(conn)
 */
int
pagedresults_is_timedout_nolock(Connection *conn)
```

mreynolds389:
Also, this is still a WIP :-)  Note the label.  The customer is still testing the fix.  I just wanted to get a PR filed in mean time

ldap/servers/slapd/pagedresults.c
progier389: I am afraid that @tbordaz is right: this change could lead to a deadlock:
       setup_pr_read_pds held the c_mutex when calling the pagedresults_is_timedout_nolock
IMHO that is not the case for the other 'exact same thing'

mreynolds389: But setup_pr_read_pds() **always** takes the conn lock before calling pagedresults_is_timedout_nolock, so why aren't we seeing a deadlock then?

progier389: IMHO both comment are correct and that is daemon.c code that is wrong:
  probably the page search timeout is not done at the right place |
  should be done before getting the c_mutex and it should try to get the   pageresult_lock_get_addr lock
   before calling pagedresults_is_timedout_nolock.
   
   For example:
```
        if (pagedresults_is_timedout_nolock(c))
        {
            /* There could be a race condition so lets try again with the right lock */
            pthread_mutex_t *mutex = pageresult_lock_get_addr(c);
            if (pthread_mutex_trylock(mutex)) == EBUSY) {
                c = next;
                continue;
            }
            if (pagedresults_is_timedout_nolock(c)) {
                pthread_mutex_free(mutex);
                disconnect_server(c, c->c_connid, -1,
                                  SLAPD_DISCONNECT_PAGED_SEARCH_LIMIT,
                                  0);
            } else {
                pthread_mutex_free(mutex);
            }
        }
        if (pthread_mutex_trylock(&(c->c_mutex)) == EBUSY) {
            c = next;
            continue;
        }

```

progier389: But setup_pr_read_pds() always takes the conn lock before calling pagedresults_is_timedout_nolock, so why aren't we seeing a deadlock then?

Because
 1 - before your change pagedresults_is_timedout_nolock was not taking the pageresult_lock_get_addr so no race condition
  2 - with your code there is a race condition but there must have other thread that hold both lock at the right time and the critical section is very small so the probability is very low

ldap/servers/slapd/daemon.c
```diff
@@ -1554,43 +1554,54 @@ setup_pr_read_pds(Connection_Table *ct, int listnum)
...
             connection_table_move_connection_out_of_active_list(ct, c);
         } else {
-            /* we try to acquire the connection mutex, if it is already
+            /* Check for a timeout for PAGED RESULTS */
+            if (pagedresults_is_timedout_nolock(c)) {
+                /*
+                 * There could be a race condition so lets try again with the
+                 * right lock
+                 */
+                pthread_mutex_t *mutex = pageresult_lock_get_addr(c);
```
tbordaz: Instead of 'mutex' you may name it something like 'pr_mutex', to make it clear it is protecting page result stuff.

ldap/servers/slapd/daemon.c
```diff
@@ -1554,43 +1554,54 @@ setup_pr_read_pds(Connection_Table *ct, int listnum)
         if (c->c_state == CONN_STATE_FREE) {
             connection_table_move_connection_out_of_active_list(ct, c);
         } else {
-            /* we try to acquire the connection mutex, if it is already
+            /* Check for a timeout for PAGED RESULTS */
+            if (pagedresults_is_timedout_nolock(c)) {
```
tbordaz, line 1558: I have a doubt here. Without holding the lock can we really trust the value.
I would rather systematically acquire the lock before calling this function.

tbordaz (commented):
The patch looks good to me. Just a question regarding the benefit of a test without holding the lock.

ldap/servers/slapd/daemon.c
mreynolds389: I had a similar concern, but this was Pierre's suggestion so I just went with it :-)  I "think" it's sort of a shortcut to avoid taking a lock at first.  The timer will only go from a value > 0 to 0.  If it's zero then we know its good because it never goes from 0 to > 0.  If it is > 0, then double check with the lock that it hasn't changed to zero.  Is this what you were thinking @progier389 ?

progier389: FYI: here is the rational for the double check first without lock then with lock:
 1) Timing is critical in this function: as previous issues showed a delay of a few ms leads to a drastic performance drop in term of number of operations p[er seconds
 2) Indeed the check without lock is unreliable (i.e: there are possible race condition) but such inconsistency is transient and:
        if we get True instead of False then the second test under lock will catch it. 
        if we get False instead of True then you will try again at next loop (i.e after 250ms in the worst case) as we are simply computing a timeout, it does not really matter if it is slightly delayed.
 3) Most operations are not page search so it would be a pity to impact them by always taking the lock

ldap/servers/slapd/opshared.c
tbordaz: I am fine to test it also at this location. Possibly add a comment why it is done only after checking 'flag_no_such_object'.

ldap/servers/slapd/daemon.c
tbordaz: Thanks for the explanation. I agree that it is a perf sensitive loop and avoiding a lock in the general case is good justification. Also the timeout will be eventually detected.

tbordaz (approved):
LGTM

ldap/servers/slapd/daemon.c
tbordaz: Sorry, I have another doubt.. (do you know columbo ;) )
pagedresults_mutex (lock of the paged result part in the connection) is held in op_shared_search to send back results. So there is a possibility that a worker acquires that lock and get blocked while sending back entries (until ioblocktimeout). It can hang the polling thread that needs to acquire the pr lock on that connection.

This scenario is possibly not created to that fix.

ldap/servers/slapd/opshared.c
mreynolds389: > I am fine to test it also at this location. Possibly add a comment why it is done only after checking 'flag_no_such_object'.

@tbordaz I'm sorry, I thought you were saying there were two calls to (be_search_results_release).  I read over your comment too quickly.  Yes, this was a duplicate check.  I have reworked the PR based on all your comments.  Please review.

ldap/servers/slapd/daemon.c
mreynolds389: > Sorry, I have another doubt.. (do you know columbo ;)

_"Oh just one more thing..."_  :laughing:

progier389 (approved):
LGTM

ldap/servers/slapd/daemon.c
mreynolds389: Talked with Thierry offline.  We will look into this issue in a new ticket.

Issue #6476, opened 2025-01-06
FTBFS with GCC 15


vashirov:
**Issue Description**
389-ds-base fails to build using GCC 15 or GCC 14 with `-std=gnu23` (which GCC 15 defaults to).

**Package Version and Platform:**
 - Platform: Fedora Rawhide
 - Package and version: `main` branch

**Steps to Reproduce**
1. Build using `CFLAGS` with `-std=gnu23`

Errors (with `-std=gnu23 -Wno-incompatible-pointer-types`: [stderr.txt](https://github.com/user-attachments/files/18320832/stderr.txt)

`-Wincompatible-pointer-types` is on by default in `gnu23` and also produces errors.

Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=2333039

vashirov:
GCC 15 is now in Fedora Rawhide.

https://gcc.gnu.org/gcc-15/porting_to.html

mreynolds389:
Working on this now...

mreynolds389:
d3c618713..502f3f6cd  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6527, opened 2025-01-21
Issue 6476 - Fix build failure with GCC 15

mreynolds389:
Description:

Most of the failures are our use of function pointer with a generic typedef with unknown parameters (e.g. IFP)

There are other IFP we use, but they are not triggering build failures yet.

Relates: https://github.com/389ds/389-ds-base/issues/6476


mreynolds389:
In the Issue there was a typo in the flag to use to reproduce this - so there are a lot more errors to fix.  Closing PR for now.

PR #6530, opened 2025-01-22
Issue 6446 - Fix test_acct_policy_consumer test to wait enough for la…

mmatsuya:
…stLoginHistory update

Adding a small sleep to wait for lastLoginHistory update after the bind operation.

Relates: #6446

Author: Masahiro Matsuya

Reviewed by: ???


tbordaz (approved):
LGTM
indeed lastLoginHistory is updated in a postop (bind) in acct_bind_postop so we need to wait a bit or do an checking loop.

PR #6532, opened 2025-01-22
Issue 6531 - Fix CI Tests - Decrease mdb map size for large topologies

### LINKED ISSUE #6531: Fix CI Tests - Decrease mdb map size for large topologies
**Issue Description**
github test shows random failure that does not reproduce on our test vm.
The test fails because some instance get unexpectedly stopped. (some 1 instance over the 4, sometime 2 instances)
It happens when the instance are idle after a test  finished successfully and before another test starts.
At this point I suspect that the system kill -9 the process because of OOM (The alternative hypotheses would be a crash after deleting a backend but ?I doubt it)
I propose to decrease the mdb map size to 10g when using topology with more than 3 instances

**Package Version and Platform:**s
 - Platform:  Fedora
 - Package and version: main

**Steps to Reproduce**
1. See CI test results and looks for error in replication/acceptance test_new_suffix

**Expected results**
The test should be stable

progier389:
Decrease mdb map size to 10 gb when using topologies whose number of instances is > 3

Issue: #6531

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -537,6 +537,8 @@ class DirSrv(SimpleLDAPObject, object):
                    SER_DEPLOYED_DIR: directory where 389-ds is deployed
                    SER_BACKUP_INST_DIR: directory where instances will be
                                         backed up
+                   SER_DB_LIB: database implementation: bdb/mdb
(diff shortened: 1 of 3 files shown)


tbordaz (approved):
LGTM

Issue #6546, opened 2025-01-27
Tests in clu/dsconf_config_test.py uses the lib389 library in /usr/local with pip


mmatsuya:
**Issue Description**

Tests in dirsrvtests/tests/suites/clu/dsconf_config_test.py added by https://github.com/389ds/389-ds-base/issues/5798
was not fail with the old package without the fix. Because the lib389 installed by pip was used for the test.

The clu test should use the system lib389 which was installed on the system. 
Because the command binary installed on the system is used for the test. 

At least in RHEL9, the default sys.path has /usr/local path, and it has higher priority than
/usr/lib{,64}/python3.x/site-packages which contains the system lib389. And, it can be 
changed with PYTHONPATH env variable. By setting PYTHONPATH without /usr/local,
the system lib389 is forcibly used for the test.

**Package Version and Platform:**
 - Platform: RHEL9
 - Package and version: 389-ds-base-2.4.5-12.el9_4.x86_64

**Steps to Reproduce**
1. Use the old version without the fix for Issue 5798
2. Run dirsrvtests/tests/suites/clu/dsconf_config_test.py test

**Expected results**
dirsrvtests/tests/suites/clu/dsconf_config_test.py fails with the old version.
dirsrvtests/tests/suites/clu/dsconf_config_test.py didn't fail with the new version with the fix for Issue 5798

vashirov:
PYTHONPATH workaround was only needed for RHEL7 where we didn't ship lib389. In other cases we don't recommend to run tests with lib389 installed in /usr/local. Closing as won't fix.

PR #6547, opened 2025-01-27
Remove /usr/local path from PYTHONPATH to use system lib389 for clu test

mmatsuya:
The clu test should use the lib389 python module from the installed one on the system, since the command also comes from the system. If PYTHONPATH env variable exists, remove /usr/local path from that. As a result, the path without /usr/local takes a priority in sys.path. If PYTHONPATH doesn't exist, use sys.path and set PYTHONPATH without the /usr/local path.

Relates to: #6546

Author: Masahiro Matsuya

Reviewed by: ???

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsconf_config_test.py
+++ b/dirsrvtests/tests/suites/clu/dsconf_config_test.py
@@ -10,6 +10,8 @@
 import subprocess
 import logging
 import pytest
+import os
+import sys
 from lib389._constants import DN_DM
 from lib389.topologies import topology_st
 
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/clu/dsconf_config_test.py
```diff
@@ -53,6 +55,29 @@ def get_dsconf_base_cmd(topology):
...
+        srcpath = sys.path
+    newpath = []
+    for path in srcpath:
+        if not path.startswith("/usr/local"):
+            newpath.append(path)
+    os.environ["PYTHONPATH"] = ":".join(newpath)
+    return savepath
+
+
+def use_systemlib389_end(savepath):
```
progier389, line 74: A bit suspicious: 
   This function will not be called if a test fails.

IMHO you should rather define use_systemlib389 as a function (or maybe module) fixture
 with a request parameter and add a finalizer top restore the environment
 
 BTW I am a bit surprised: savepath is undefined in use_systemlib389_end function ...

PR #6556, opened 2025-01-28
Issue 6468 - Fix building for older versions of Python

vashirov:
Bug Description:
Structural Pattern Matching has been added in Python 3.10, older version do not support it.

Fix Description:
Replace `match` and `case` statements with `if-elif`.

Relates: https://github.com/389ds/389-ds-base/issues/6468

Reviewed by:

### DIFF
--- a/src/lib389/lib389/cli_conf/logging.py
+++ b/src/lib389/lib389/cli_conf/logging.py
@@ -236,19 +236,20 @@ def get_log_config(inst, basedn, log, args):
     attr_map = {}
     levels = {}
 
-    match args.logtype:
-        case "access":
-            attr_map = ACCESS_ATTR_MAP
-            levels = ACCESS_LEVELS
-        case "error":
-            attr_map = ERROR_ATTR_MAP
-            levels = ERROR_LEVELS
-        case "security":
-            attr_map = SECURITY_ATTR_MAP
-        case "audit":
-            attr_map = AUDIT_ATTR_MAP
-        case "auditfail":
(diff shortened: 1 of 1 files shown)


droideck (approved):
LGTM!

src/lib389/lib389/cli_conf/logging.py
```diff
@@ -236,19 +236,20 @@ def get_log_config(inst, basedn, log, args):
...
-        case "error":
-            attr_map = ERROR_ATTR_MAP
-            levels = ERROR_LEVELS
-        case "security":
-            attr_map = SECURITY_ATTR_MAP
-        case "audit":
-            attr_map = AUDIT_ATTR_MAP
-        case "auditfail":
-            attr_map = AUDITFAIL_ATTR_MAP
+    if args.logtype == "access":
```
progier389, line 239: I wonder if we should not rather use a dict here ...

progier389:
```
LOGTYPE_DICT = {
    "access": (ACCESS_ATTR_MAP, ACCESS_LEVELS),
    "error": (ERROR_ATTR_MAP, ERROR_LEVELS),
    "security": (SECURITY_ATTR_MAP, None),
    "audit": (AUDIT_ATTR_MAP, None),
    "auditfail": (AUDITFAIL_ATTR_MAP, None),
}

try:
    attr_map,levels = LOGTYPE_DICT[args.logtype]
except KeyError:
    raise ValueError(f"Unknown logtype: {args.logtype}")
```

Issue #6555, opened 2025-01-28
Potential crash when deleting a replicated backend


progier389:
**Issue Description**
When removing a replicated backend (and probably also when disabling replication) a crash may occurs because the connection cleanup code attempt to use a free replica:"

```
#0  ___pthread_mutex_lock (mutex=mutex@entry=0x7373616c4374636d) at pthread_mutex_lock.c:80
#1  0x00007fba78007796 in PR_EnterMonitor (mon=0x7373616c43746365) at ../../../../nspr/pr/src/pthreads/ptsynch.c:563
#2  0x00007fba73c305c7 in replica_lock (lock=<optimized out>) at ldap/servers/plugins/replication/repl5_replica.c:109
#3  replica_relinquish_exclusive_access (r=0x7fb707089140, connid=0, opid=-1) at ldap/servers/plugins/replication/repl5_replica.c:676
#4  0x00007fba73c16475 in consumer_connection_extension_destructor (ext=<optimized out>, object=<optimized out>, parent=<optimized out>) at ldap/servers/plugins/replication/repl_connext.c:91
#5  0x00007fba789e0ad3 in factory_destroy_extension (type=<optimized out>, object=0x7fb710e01a90, parent=0x0, extension=0x7fb710e01bc8) at ldap/servers/slapd/factory.c:366
#6  factory_destroy_extension (type=<optimized out>, object=0x7fb710e01a90, parent=0x0, extension=0x7fb710e01bc8) at ldap/servers/slapd/factory.c:348
#7  0x0000555bc877d76c in connection_cleanup (conn=0x7fb710e01a90) at ldap/servers/slapd/connection.c:181
#8  0x0000555bc879117c in connection_table_move_connection_out_of_active_list.isra.0 (ct=ct@entry=0x7fba743977c0, c=c@entry=0x7fb710e01a90) at ldap/servers/slapd/conntable.c:470
#9  0x0000555bc87827db in setup_pr_read_pds (ct=0x7fba743977c0, listnum=<optimized out>) at ldap/servers/slapd/daemon.c:1596
#10 ct_list_thread (threadnum=<optimized out>) at ldap/servers/slapd/daemon.c:1356
#11 0x00007fba7800e3b7 in _pt_root (arg=0x7fba7722bbc0) at ../../../../nspr/pr/src/pthreads/ptthread.c:191
#12 0x00007fba78797057 in start_thread (arg=<optimized out>) at pthread_create.c:448
[...]

PR #6559, opened 2025-01-29
Issue 6555 - Potential crash when deleting a replicated backend

progier389:
Problem: connection cleanup may try to access a freed replica
Solution: Ensure that the replica is still existing before using its pointer

Issue: #6555 

Reviewed by: @tbordaz, @droideck  (Thanks!)

### DIFF
--- a/ldap/servers/plugins/replication/repl5.h
+++ b/ldap/servers/plugins/replication/repl5.h
@@ -29,6 +29,7 @@
 #include "repl5_ruv.h"
 #include "plstr.h"
 #include <pthread.h>
+#include <stdbool.h>
 
 #define START_UPDATE_DELAY                   2 /* 2 second */
 #define REPLICA_TYPE_WINDOWS                 1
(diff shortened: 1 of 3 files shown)


tbordaz (approved):
LGTM

ldap/servers/plugins/replication/repl5.h
```diff
@@ -710,6 +711,8 @@ PRBool replica_is_updatedn(Replica *r, const Slapi_DN *sdn);
 void replica_set_updatedn(Replica *r, const Slapi_ValueSet *vs, int mod_op);
 void replica_set_groupdn(Replica *r, const Slapi_ValueSet *vs, int mod_op);
 char *replica_get_generation(const Replica *r);
+bool replica_check_validity(Replica *replica);
```
droideck, line 714: It feels a bit inconsistent to use `bool` while every other line in the file uses `PRBool`... But I understand we might get rid of `PRBool` at some point.

I don't have a hard opinion here, but generally, I'd prefer consistency in this case (my 5 cents). :)

droideck (approved):
The change looks good!

ldap/servers/plugins/replication/repl5.h
progier389: FYI: I discussed that one with Mark in another review, and finally we decided to use stdbool in new code.
I agree it is  inconsistent, until we do a global cleanup phase (but frankly I am afraid that there are dependencies toward the plugin API and we will never be able to complete that move )

PR #6560, opened 2025-01-30
Issue 6465 - Create optional meta-package for replication reporting dependencies

### LINKED ISSUE #6465: Add a replication log analysis tool for performance monitoring
**Description**
Add a new ReplicationLogAnalyzer class and supporting infrastructure to analyze replication performance across multiple servers. The tool parses replication logs, tracks CSNs, calculates lag times, and generates comprehensive reports in multiple formats including interactive HTML visualizations.
Additionally, add CLI and UI implementations.

**Implementation References**
Base implementation can be adapted from: https://github.com/droideck/ansible-ds389-repl-monitoring/tree/main/roles/ds389_repl_monitoring/library

droideck:
Description: Introduce a new `python3-lib389-repl-reports` meta-package that bundles optional dependencies like Plotly and Matplotlib. This enables advanced replication reporting (HTML/PNG) without forcing these libraries on all installations. Update replication tests.
Remove direct references to Plotly, Matplotlib, and Numpy from the base package.

Fixes: https://github.com/389ds/389-ds-base/issues/6465

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/replication/repl_log_monitoring_test.py
+++ b/dirsrvtests/tests/suites/replication/repl_log_monitoring_test.py
@@ -24,6 +24,13 @@ from lib389.replica import ReplicationManager
 from lib389.repltools import ReplicationLogAnalyzer
 from lib389._constants import *
 
+try:
+    import plotly
+    import matplotlib
+    HTML_PNG_REPORTS_AVAILABLE = True
+except ImportError:
+    HTML_PNG_REPORTS_AVAILABLE = False
+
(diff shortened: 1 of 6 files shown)


vashirov (approved):
This made a difference for me on Fedora, thanks!
I tried to build it on CentOS Stream 10 and I realized that unfortunately we don't have plotly and matplotlib there... So I'm approving this to reduce the dependencies on Fedora in the short term, but we need to figure out how to make this work in EL10.

Issue #6562, opened 2025-01-31
Prevent potential undefined behaviour in ldap/servers/slapd/util.c in filter_stuff_func


Rx1513:
**Issue Description**
There's probability to get memcpy(ctx->buf + ctx->buf_size, NULL, 0) which is caused by escaping null character.
In [cppreference](https://en.cppreference.com/w/cpp/string/byte/memcpy) it is an undefined behaviour:

> If either dest or src is an [invalid or null pointer](https://en.cppreference.com/w/cpp/language/pointer#Pointers), the behavior is undefined, even if count is zero.

Although there're no mention in creference about having src and count both zero, I assume it's also an undefined behaviour.

**Steps to Reproduce**
Pass to filter_stuff_func f (see gdb log) and any of EQ and APPROX filters.

```
Breakpoint 4, filter_escape_filter_value (f=0x55fdadcd99a0, fmt=0x7f38fd37e367 "(%s=%s%s)", len=4) at ldap/servers/slapd/filter.c:132
132         filter_compute_hash(f);
(gdb) print *f
$9 = {f_flags = 0, f_choice = 163, f_hash = 0, assigned_decoder = 0x0, f_un = {f_un_type = 0x55fdadcd9a70 "", f_un_ava = {ava_type = 0x55fdadcd9a70 "", 
      ava_value = {bv_len = 0, bv_val = 0x55fdadcd9a50 ""}, ava_private = 0x0}, f_un_complex = 0x55fdadcd9a70, f_un_sub = {sf_type = 0x55fdadcd9a70 "", 
      sf_initial = 0x0, sf_any = 0x55fdadcd9a50, sf_final = 0x0, sf_private = 0x0}, f_un_extended = {mrf_type = 0x55fdadcd9a70 "", mrf_value = {bv_len = 0, 
        bv_val = 0x55fdadcd9a50 ""}, mrf_oid = 0x0, mrf_dnAttrs = 0 '\000', mrf_match = 0x0, mrf_index = 0x0, mrf_reusable = 0, mrf_reset = 0x0, 
      mrf_object = 0x0, mrf_destroy = 0x0}}, f_next = 0x0}
```

**Additional context**
Originally it was found during fuzzing of get_filter.

vashirov:
Unfortunately, this change introduced a regression, filter tests fail.
In particular, dirsrvtests/tests/suites/filter/bitw_filter_test.py::test_all_together causes heap-use-after-free:

```
[...]

PR #6563, opened 2025-01-31
Issue #6562 - Prevent undefined behaviour in in filter_stuff_func

Rx1513:
Bug Description:

Undefined behaviour caused by escaping null character.

Fix Description:

Add simple check when ldap_bv2escaped_filter_value return null terminator which return error.

relates: https://github.com/389ds/389-ds-base/issues/6562

Reviewed by: @progier389 (Thanks!)


Rx1513:
I guess there's much better solution than this. Like adding check right after function signature or making wrapper for ldap_bv2escaped_filter_value

progier389 (approved):
LGTM

Issue #6489, opened 2025-01-09
crash on logging 3.1.1


stanislavlevin:
**Issue Description**
`ns-slapd` sporadically crashes.

**Package Version and Platform:**
 - ALTLinux
 - 389-ds-base-3.1.1 (synced to f5732e1d59da0edada10d697420a5f6b34723a7a).

**Steps to Reproduce**
Can't provide the exact steps.

**Expected results**
No crash.

**Additional context**

[logging_crash.stacktrace.tar.gz](https://github.com/user-attachments/files/18363021/logging_crash.stacktrace.tar.gz)

```
#4  0x00007f5d15a3fb6b in PR_Assert (s=s@entry=0x7f5d15a6114e "I/O method is invalid", file=file@entry=0x7f5d15a61143 "priometh.c", ln=ln@entry=54) at io/prlog.c:571
#5  0x00007f5d15a4028c in _PR_InvalidInt () at io/priometh.c:54
#6  0x00007f5d162b8a5e in log_write (headersize=0, fd=0x7f5d1450ad60, buffer=0x7f59691f8b80 "\t389-Directory/3.1.1 B2024.359.1534\n", size=36, flush=1) at ldap/servers/slapd/log.c:226
        rc = 0
        rc = <optimized out>
        prerr = <optimized out>
#7  log_write_title (fp=0x7f5d1450ad60) at ldap/servers/slapd/log.c:2252
        fe_cfg = 0x7f5d164f73c0 <global_slapdFrontendConfig>
        buildnum = 0x7f597b0524d0 "2024.359.1534"
        buff = "\t389-Directory/3.1.1 B2024.359.1534\n\000\177\000\000\311\307W\026]\177\000\000\260\214\037iY\177\000\000-\215\037iY\177\000\000\260\224\037iY\177\000\000\000\000\000\000\000\000\000\000\002\000\000\000\000\000\000\000\360\213\037iY\177\000\000\200\213\037iY\177\000\000\320W\177g", '\000' <repeats 12 times>, "\005\000\000\000\t\000\000\000\000\000\000\000}\000\000\000\004\000\000\000\b", '\000' <repeats 19 times>, "\300\245\242\024]\177\000\000\300\306\037iY\177\000\000\000\000\000\000\000\000\000\000\240\341\240\024]\177\000\000\260\214\037iY\177\000\000"...
        rc = 0
[...]

PR #6565, opened 2025-01-31
Issue 6489 - After log rotation refresh the FD pointer

mreynolds389:
Description:

When flushing a log buffer we get a FD for log prior to checking if the log should be rotated.  If the log is rotated that FD reference is now invalid, and it needs to be refreshed before proceeding

Relates: https://github.com/389ds/389-ds-base/issues/6489

### DIFF
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -6750,6 +6750,23 @@ log_refresh_state(int32_t log_type)
         return 0;
     }
 }
+static LOGFD
+log_refresh_fd(int32_t log_type)
+{
+    switch (log_type) {
+    case SLAPD_ACCESS_LOG:
+        return loginfo.log_access_fdes;
+    case SLAPD_SECURITY_LOG:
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/log.c
```diff
@@ -6861,6 +6861,25 @@ log_flush_buffer(LogBufferInfo *lbi, int log_type, int sync_now, int locked)
                                                         rotationtime_secs);
         }
         log_state = log_refresh_state(log_type);
+
```
tbordaz: it looks good but similarly to the log_state you may move that code into a function and call
`fd = log_refresh_filedescriptor(log_type);`

tbordaz (approved):
LGTM

ldap/servers/slapd/log.c
mreynolds389: > it looks good but similarly to the log_state you may move that code into a function and call `fd = log_refresh_filedescriptor(log_type);`

Implemented

PR #6567, opened 2025-02-03
Issue 6566 - RI plugin failure to handle a modrdn for rename of member of multiple groups

### LINKED ISSUE #6566: RI plugin failure to handle a modrdn when user is in multiple groups
**Issue Description**
WIth AM and RI plugins enabled, the rename of a user that is part of multiple groups fails with a "value exists" error.
```
ipa user-mod --rename=dweathe dustin.weatherholtz
ipa: ERROR: Type or value exists
```
 - Package and version: 389-ds-base-1.4.3

**Steps to Reproduce**
1. Enable AM and RI plugins
2. Add AM rule
```
dsconf -D "cn=dm" ldap://localhost:389  -w password plugin automember  definition userswitha add   --scope "ou=People,dc=example,dc=com"  --filter "objectclass=posixAccount" --grouping-attr "member:dn"   
dsconf -D "cn=dm" ldap://localhost:389  -w password plugin automember definition userswitha regex userswitha add --target-group cn=userswitha,ou=groups,dc=example,dc=com --inclusive cn=.*a.*

dsconf -D "cn=dm" ldap://localhost:389  -w password plugin automember  definition userswithz add   --scope "ou=People,dc=example,dc=com"  --filter "objectclass=posixAccount" --grouping-attr "member:dn"   
dsconf -D "cn=dm" ldap://localhost:389  -w password plugin automember definition userswithz  regex userswithz add --target-group cn=userswithz,ou=groups,dc=example,dc=com --inclusive cn=.*z.*
```
4. Add three users with specific names
```
sudo dsidm -b 'dc=example,dc=com' inst01 user create --uid userwitha  --cn userwitha --displayName "userwitha" --uidNumber 1002 --gidNumber 1002 --homeDirectory /home/userwitha
sudo dsidm -b 'dc=example,dc=com' inst01 user create --uid userwithz  --cn userwithz --displayName "userwithz" --uidNumber 1003 --gidNumber 1003 --homeDirectory /home/userwithz
sudo dsidm -b 'dc=example,dc=com' inst01 user create --uid userwithaz  --cn userwithaz --displayName "userwithaz" --uidNumber 1004 --gidNumber 1004 --homeDirectory /home/userwithaz
```
5. Rename userwithaz
6. See error
```
[03/Feb/2025:00:28:17.090767901 +0000] - ERR - referint-plugin - _update_all_per_mod - Entry cn=userswithz,ou=groups,dc=example,dc=com failed (20)
```
7. **Expected results**
User name should be updated without issue

**Additional context**
Testcase copied from https://issues.redhat.com/browse/RHEL-18333

jchapma:
Bug description:
With AM and RI plugins enabled, the rename of a user that is part of multiple groups fails with a "value exists" error.

Fix description:
For a modrdn the RI plugin creates a new DN, before a modify is attempted check if the new DN already exists in the attr being updated.

Fixes: https://github.com/389ds/389-ds-base/issues/6566

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/referint/referint.c
+++ b/ldap/servers/plugins/referint/referint.c
@@ -924,6 +924,7 @@ _update_all_per_mod(Slapi_DN *entrySDN, /* DN of the searched entry */
 {
     Slapi_Mods *smods = NULL;
     char *newDN = NULL;
+    struct berval bv = {0};
     char **dnParts = NULL;
     char *sval = NULL;
     char *newvalue = NULL;
@@ -984,6 +985,16 @@ _update_all_per_mod(Slapi_DN *entrySDN, /* DN of the searched entry */
         /* newRDN and superior are already normalized. */
         newDN = slapi_ch_smprintf("%s,%s", newRDN, superior);
         slapi_dn_ignore_case(newDN);
+
+        /* Check if DN has been updated already (automember) */
+        bv.bv_val = newDN;
+        bv.bv_len = strlen(newDN);
+        if (VALUE_PRESENT == attr_value_find_wsi(attr, &bv, &v)) {
+            slapi_log_err(SLAPI_LOG_PLUGIN, REFERINT_PLUGIN_SUBSYSTEM,
+                          "_update_all_per_mod - Duplicate DN update detected, ignoring %s\n", newDN);
+            bv = (struct berval){0};
+            goto bail;
+        }
         /*
          * Compare the modified dn with the value of
          * the target attribute of referint to find out
--- a/ldap/servers/plugins/referint/referint.c
+++ b/ldap/servers/plugins/referint/referint.c
@@ -985,16 +985,6 @@ _update_all_per_mod(Slapi_DN *entrySDN, /* DN of the searched entry */
         /* newRDN and superior are already normalized. */
         newDN = slapi_ch_smprintf("%s,%s", newRDN, superior);
         slapi_dn_ignore_case(newDN);
-
-        /* Check if DN has been updated already (automember) */
-        bv.bv_val = newDN;
-        bv.bv_len = strlen(newDN);
-        if (VALUE_PRESENT == attr_value_find_wsi(attr, &bv, &v)) {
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

ldap/servers/plugins/referint/referint.c
```diff
@@ -984,6 +985,16 @@ _update_all_per_mod(Slapi_DN *entrySDN, /* DN of the searched entry */
...
         slapi_dn_ignore_case(newDN);
+
+        /* Check if DN has been updated already (automember) */
+        bv.bv_val = newDN;
+        bv.bv_len = strlen(newDN);
+        if (VALUE_PRESENT == attr_value_find_wsi(attr, &bv, &v)) {
+            slapi_log_err(SLAPI_LOG_PLUGIN, REFERINT_PLUGIN_SUBSYSTEM,
+                          "_update_all_per_mod - Duplicate DN update detected, ignoring %s\n", newDN);
+            bv = (struct berval){0};
+            goto bail;
```
tbordaz: you need to call slapi_ch_free_string(&newDN) before going to 'bail'

tbordaz (commented):
Although the fix looks good to me I have a doubt.
The block deletes the value of the oldDN and add the value of the newDN.
In short the fix skips the block if the newDN value exists. I have a doubt about not removing the oldDN, I wonder if in some cases the oldDN may remain in the entry.
Would it be possible to catch LDAP_TYPE_OR_VALUE_EXISTS and XXX in _do_modify (where XXX is the RC when the modify does not find the value to delete)

jchapma:
> Although the fix looks good to me I have a doubt. The block deletes the value of the oldDN and add the value of the newDN. In short the fix skips the block if the newDN value exists. I have a doubt about not removing the oldDN, I wonder if in some cases the oldDN may remain in the entry. Would it be possible to catch LDAP_TYPE_OR_VALUE_EXISTS and XXX in _do_modify (where XXX is the RC when the modify does not find the value to delete)

Yes, you are correct. With this change we are exiting before the delete of the oldDN is done. Let me change this.

tbordaz (approved):
LGTM

PR #6569, opened 2025-02-03
Issue 6004 - (2nd) idletimeout may be ignored

tbordaz:
Problem:
    multiple listener threads was implemented in 2.x and after
    This is missing in 1.4.3 so the cherry pick should be adapted
Fix:
   skip the loop with listeners

Issue #6004

Reviewed by:


jchapma (approved):
LGTM

Issue #6574, opened 2025-02-03
Move pblock access from switch to func table


mreynolds389:
**Issue Description**

Currently there are over 300 switch cases for getting/setting pblock parameters.

Instead we should have separate functions for each parameter, and a callback function pointer table.

mreynolds389:
b30473ccd..7f3760520  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6575, opened 2025-02-03
Issue 6574 - Move pblock access from switch to func table

mreynolds389:
Description

Currently there are over 300 switch cases for getting/setting pblock parameters.

Instead we should have separate functions for each parameter, and use a callback function pointer table.

This also combines all the parameter definitions to a single ordered list. This will make future changes much easier to implement.

Relates: https://github.com/389ds/389-ds-base/issues/6574

### DIFF
--- a/ldap/servers/slapd/pblock.c
+++ b/ldap/servers/slapd/pblock.c
@@ -30,9 +30,6 @@
 #define SLAPI_PW_ENTRY 2005
 #define SLAPI_TASK_WARNING 2006
 
-/* Used for checking assertions about pblocks in some cases. */
-#define SLAPI_HINT 9999
-
 static PRLock *pblock_analytics_lock = NULL;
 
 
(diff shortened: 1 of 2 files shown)


mreynolds389:
As for performance, with ldclt doing exact searches I do see a 5% - 8% performance improvement when the number of threads is increased to 16 (matching the hardware).  Under less load I don't really see an improvement.  So there is a benefit when the server is being hit hard.

Firstyear (approved):
Not every day you get a 5 to 8% improvement! That's amazing to see :)

mreynolds389:
> Not every day you get a 5 to 8% improvement! That's amazing to see :)

Well its only under certain load types - it's not across the board :-)  But yes I see an improvement when all the workers are working

ldap/servers/slapd/pblock.c
```diff
@@ -368,3865 +365,5944 @@ slapi_pblock_clone(Slapi_PBlock *pb)
...
-int
-slapi_pblock_get(Slapi_PBlock *pblock, int arg, void *value)
+static int32_t
+slapi_pblock_get_backend(Slapi_PBlock *pblock, void *value)
 {
+    Slapi_Backend *be = pblock->pb_backend;
 
-#ifdef PBLOCK_ANALYTICS
-    pblock_analytics_record(pblock, arg);
-#endif
```
droideck, line 377: This analytics code (and the one in `_set`) were removed. Should we still keep it? Or it's not used?

ldap/servers/slapd/slapi-plugin-compat4.h
```diff
@@ -31,16 +31,6 @@ extern "C" {
...
-/******* Deprecated (obsolete) pblock arg identifiers ************************/
-#define SLAPI_CONFIG_FILENAME 40 /* use config. entries instead */
-#define SLAPI_CONFIG_LINENO 41   /* use config. entries instead */
-#define SLAPI_CONFIG_ARGC 42     /* use config. entries instead */
-#define SLAPI_CONFIG_ARGV 43     /* use config. entries instead */
-#define SLAPI_CONN_AUTHTYPE 144  /* use SLAPI_CONN_AUTHTYPE */
-#define SLAPI_REQUESTOR_ISUPDATEDN SLAPI_IS_REPLICATED_OPERATION
-#define SLAPI_CONN_CLIENTIP 145 /* use SLAPI_CONN_CLIENTNETADDR */
-#define SLAPI_CONN_SERVERIP 146 /* use SLAPI_CONN_SERVERNETADDR */
-
```
droideck: I think we should mention the removal in the commit message.

But also, we have some leftovers in `test/libslapd/pblock/pblock_accessors*` and a bit in `_slapi_pblock_deprecated` in `ldap/servers/slapd/pblock_v3.h`.

IMO, it'll be cleaner to remove the code in a separate PR.

droideck (approved):
Besides a couple of minor issues, it looks like a simple but powerful fix! Nice find!

tbordaz (commented):
The change looks good and  perf improvement very nice. IIUC a former customer plugin will require to be rebuilt, correct ?

ldap/servers/slapd/slapi-plugin.h
```diff
@@ -6977,50 +6974,6 @@ slapi_timer_result slapi_timespec_expire_check(struct timespec *expire);
                                          * into tombstone
                                          */
 
-/*
```
progier389: I do not like the massive change in slapi-plugin.h 
  (looks like a risk of breaking existing custom plugin
    And after a closer look as you changed the #define values, it will definitively break the binary compatibility
     with existing custom plugin.
IMHO we should not do this. we should be able to split the switches without changing slapi-plugin.h

tbordaz: an option would be to have a conversion table that convert a former define value into an index into the callback table

progier389: FYI, or you can do a simple table:
 here is a python script that generate a skeleton of what it would looks like:
[split_pblock_switches.py.txt](https://github.com/user-attachments/files/18657166/split_pblock_switches.py.txt)

mreynolds389:
> The change looks good and perf improvement very nice. IIUC a former customer plugin will require to be rebuilt, correct ?

@progier389 @tbordaz Yeah you guys are right, but the problem is that the callback table will have thousands of empty slots to fill in the gaps between the #define numbers.  Maybe we can use a hash table instead?

progier389:
Having a single table with hole will be faster than an hash table (there will be a single indirection versus hash computation)
In fact  if we ignore SLAPI_HINT (which is purely internal and can be changed)  the maximum value is slightly below 2000
Having hole is not really an issue: holes are just a pointeur to a function that logs error

mreynolds389:
> Having a single table with hole will be faster than an hash table (there will be a single indirection versus hash computation) In fact if we ignore SLAPI_HINT (which is purely internal and can be changed) the maximum value is slightly below 2000 Having hole is not really an issue: holes are just a pointeur to a function that logs error

I guess I got caught up in maintainability and and trying to fix how sloppy our current code is (in regards to defines).  Anyway I'll undo the slapi-plugin change and adjust the tables in pblock.c

ldap/servers/slapd/pblock.c
mreynolds389: > This analytics code (and the one in `_set`) were removed. Should we still keep it? Or it's not used?

I'm not 100% what these are for.  I assumed it was for measuring how we accessed the pblock, but I have a feeling this is no longer needed.  Perhaps we can remove it in a separate PR?

ldap/servers/slapd/slapi-plugin-compat4.h
mreynolds389: I left this as is with the new commit

ldap/servers/slapd/pblock.c
```diff
@@ -4571,3 +6729,3949 @@ bind_credentials_set_nolock(Connection *conn, char *authtype, char *normdn, char
...
+    slapi_pblock_set_dbverify_dbdir,
+    slapi_pblock_set_plugin_ext_op_backend_fn,
+    slapi_pblock_set_paged_results_cookie,
+    slapi_pblock_set_usn_increment_for_tombstone,
+    slapi_pblock_set_memberof_deferred_task,
+    NULL, /* slot 1952 available */
+};
+
+int
+slapi_pblock_set(Slapi_PBlock *pblock, int arg, void *value)
```
droideck, line 10668: I think we still need `PR_ASSERT(NULL != pblock);` here.

mreynolds389: > I think we still need `PR_ASSERT(NULL != pblock);` here.

Agreed, done!

tbordaz (approved):
The (huge) change looks good to me and nice.
Minor question, any reason why keeping slapi_pblock_get/set as 'int' rather than int32_t ?

ldap/servers/slapd/pblock.c
```diff
@@ -4571,3 +6729,3952 @@ bind_credentials_set_nolock(Connection *conn, char *authtype, char *normdn, char
...
+{
+    PR_ASSERT(NULL != pblock);
+
+    if (arg > 0 && arg <= PR_ARRAY_SIZE(set_cbtable) && set_cbtable[arg] != NULL) {
```
progier389: Should be arg >= 0 && arg<PR_ARRAY_SIZE(set_cbtable) && ... to avoid array overflow ...
(maybe same issue on slapi_pblock_get ?)

mreynolds389: > Should be arg >= 0 && arg<PR_ARRAY_SIZE(set_cbtable) && ... to avoid array overflow ... (maybe same issue on slapi_pblock_get ?)

My bad, I used the code example _you_ sent me and I didn't fully look it over :-)  I'll get it fixed

mreynolds389:
@progier389 @tbordaz changes made, please review one last time :-)

Issue #6583, opened 2025-02-05
Fix CI on older branches


vashirov:
**Issue Description**
Test execution on older branches (1.4.x, some 2.x) fails. We use the same Fedora base image across all branches, but it contains newer toolchain and doesn't have some build dependencies.
Instead we should use appropriate images: EL8 for 1.4.x, EL9 for 2.x, EL10/Fedora for 3.x.

WebUI tests on EL8 can't be executed because playwright doesn't support Python 3.6 that is shipped there by default. And running tests with newer Python versions requires lib389 for the same version, which requires additional workarounds to install. WebUI tests will be addressed in another ticket.

PR #6584, opened 2025-02-05
Issue 6583 - Fix CI on older branches - 1.4.3

vashirov:
Bug Description:
Test execution on older branches (1.4.x, some 2.x) fails. We use the same Fedora base image across all branches, but it contains newer toolchain and doesn't have some build dependencies.

Fix Description:
Use appropriate images: EL8 for 1.4.x, EL9 for 2.x, EL10/Fedora for 3.x.

Fixes: https://github.com/389ds/389-ds-base/issues/6583


progier389 (approved):
LGTM

progier389:
Maybe we should  also disable C Lang builds that fails  (I do not think that they are really useful for rhel 8 tests)

vashirov:
Right, I think the whole Compile action can be disabled for older branches.

progier389:
FYI: I think that with 
https://github.com/389ds/389-ds-base/pull/6670 we may avoid some of this PR changes. 
 (compile action should still be removed )
  but I have nearly the same test result with the branch current image

PR #6585, opened 2025-02-05
Issue 6555 - V2 - Potential crash when deleting a replicated backend

progier389:
Problem: connection cleanup may try to access a freed replica
Solution: Ensure that the replica is still existing before using its pointer
Note: Second version because of a potential hang at shutdown 

Issue: https://github.com/389ds/389-ds-base/issues/6555


tbordaz (approved):
The problem occurred quit frequently in my previous tests. I would say 1 out 5 tries.
The patch works after more than 130 runs.

PR #6586, opened 2025-02-05
Issue 6583 - Fix CI on older branches - 1.4.4

vashirov:
Bug Description:
Test execution on older branches (1.4.x, some 2.x) fails. We use the same Fedora base image across all branches, but it contains newer toolchain and doesn't have some build dependencies.

Fix Description:
Use appropriate images: EL8 for 1.4.x, EL9 for 2.x, EL10/Fedora for 3.x.

Fixes: https://github.com/389ds/389-ds-base/issues/6583


progier389 (approved):
LGTM

Issue #6599, opened 2025-02-10
Implement option to write the access log in JSON


mreynolds389:
**Description**

Implement option to write the access log in JSON.

mreynolds389:
Design doc

https://www.port389.org/docs/389ds/design/accesslog-json-design.html

mreynolds389:
fec8bfb39..e7c6211af  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6600, opened 2025-02-10
Issue 6599 - Implement option to write access log in JSON

mreynolds389:
Description:

Add option to write access log in JSON

Relates: https://github.com/389ds/389-ds-base/issues/6599


mreynolds389:
Design doc:

https://www.port389.org/docs/389ds/design/accesslog-json-design.html

progier389 (approved):
LGTM, 
The only doubt I have is about the copyrights, 
I was thinking that now we tend to keep a single year for RH copyright , but I may be wrong ...

mreynolds389:
> LGTM, The only doubt I have is about the copyrights, I was thinking that now we tend to keep a single year for RH copyright , but I may be wrong ...

On IPA they use this range approach. I recall it following some "standard", but I'm not really sure.  Let me look into that...

mreynolds389:
> > LGTM, The only doubt I have is about the copyrights, I was thinking that now we tend to keep a single year for RH copyright , but I may be wrong ...
> 
> On IPA they use this range approach. I recall it following some "standard", but I'm not really sure. Let me look into that...

Not really seeing anything concrete on the matter.  Seems ranges, or lists are common, but not required.  Ultimately the result is the same as long as it has the current year.  So I'll change it back to a single year as I don't have a strong option on it.

jchapma:
I had a good play around with this, works great for me. It is going to make logconv's job a lot easier.

Issue #5841, opened 2023-07-14
dsconf incorrectly setting up Pass-Through Authentication


mreynolds389:
## Description of problem

dsconf incorrectly setting up Pass-Through Authentication

Version-Release number of selected component (if applicable):
389-ds-base.x86_64 (1.4.3.34-1.module+el8.7.0+18367+58a49cb0

## Steps to Reproduce:

    # dsconf corpldap plugin pam-pass-through-auth config 'IPA PAM PTA Config' add --exclude-suffix='cn=config' --id_map_method='ENTRY' --id-attr='uid' --filter="(&(!(uid=newrelicmon))(!(uid=bugzilla))(!(uid=applmgr)))" --fallback='TRUE' --secure='TRUE' --service='ldapserver' --missing-suffix='ALLOW' --include-suffix='ou=Users,dc=redhat,dc=com'
Successfully created the cn=IPA PAM PTA Config,cn=PAM Pass Through Auth,cn=plugins,cn=config

## Actual results:
fallback, id_map_method and id_attr being incorrectly set

## Additional info:

    # dsconf corpldap plugin pam-pass-through-auth config 'IPA PAM PTA Config' add --exclude-suffix='cn=config' --id_map_method='ENTRY' --id-attr='uid' --filter="(&(!(uid=newrelicmon))(!(uid=bugzilla))(!(uid=applmgr)))" --fallback='TRUE' --secure='TRUE' --service='ldapserver' --missing-suffix='ALLOW' --include-suffix='ou=Users,dc=redhat,dc=com'
    Successfully created the cn=IPA PAM PTA Config,cn=PAM Pass Through Auth,cn=plugins,cn=config

It created successfully, but showing the output results in fallback, id_map_method and id_attr being incorrectly set:

```
# dsconf corpldap plugin pam-pass-through-auth show
dn: cn=PAM Pass Through Auth,cn=plugins,cn=config
cn: PAM Pass Through Auth
nsslapd-plugin-depends-on-type: database
nsslapd-pluginDescription: PAM pass through authentication plugin
nsslapd-pluginEnabled: on
nsslapd-pluginId: pam_passthruauth
nsslapd-pluginInitfunc: pam_passthruauth_init
nsslapd-pluginPath: libpam-passthru-plugin
nsslapd-pluginType: betxnpreoperation
nsslapd-pluginVendor: 389 Project
nsslapd-pluginVersion: 1.4.3.34
nsslapd-pluginloadglobal: true
[...]

PR #6601, opened 2025-02-11
Issue 5841 - dsconf incorrectly setting up Pass-Through Authentication

jchapma:
Bug description:
During init, PAMPassThroughAuthConfigs defines an "objectclass=nsslapdplugin" plugin object. During filter creation, dsconf fails as objectclass=nsslapdplugin is not present in the PAM PT config entry. This objectclass has been removed in all other branches, branch 1.4.3 was skipped as there are cherry pick conflicts.

Fix description:
Remove nsslapdplugin from the plugin object, objectclass list.

Fixes: https://github.com/389ds/389-ds-base/issues/5841

Reviewed by:


progier389 (approved):
LGTM

PR #6605, opened 2025-02-12
Issue 6604 - Fix coverity scan issues CID 1591261 - CID 1591269

### LINKED ISSUE #6604: Fix coverity scan issues CID 1591261 - CID 1591269
**Issue Description**
Fix latest coverity scan new regression reported this week (i.e: CID 1591259 - CID 1591269 in https://scan.coverity.com/projects/389ds-389-ds-base?tab=overview)
 Note: CID 1591259 and CID 1591260 are not fixed because they are false positive

 - Platform: Fedora

**Steps to Reproduce**
1. See coverity scan result in https://scan.coverity.com/projects/389ds-389-ds-base?tab=overview

**Additional context**

```
*** CID 1591260:  Concurrent data access violations  (MISSING_LOCK)
/ldap/servers/plugins/statechange/statechange.c: 273 in statechange_post_op()
267     
268                 if (execute && notify->filter) {
269                     /* next the filter */
270                     int filter_test = 0;
271     
272                     /* need to test entry both before and after op */
>>>     CID 1591260:  Concurrent data access violations  (MISSING_LOCK)
>>>     Accessing "notify->realfilter" without holding lock "buffer_lock". Elsewhere, "_statechange_notify.realfilter" is written to with "buffer_lock" held 2 out of 2 times.
273                     if (e_before && !slapi_filter_test_simple(e_before, notify->realfilter))
274                         filter_test = 1;
275     
276                     if (!filter_test && e_after && !slapi_filter_test_simple(e_after, notify->realfilter))
277                         filter_test = 1;
278     
```

But line 2744 buffer_lock is held in read mode: `slapi_rwlock_rdlock(buffer_lock);`
** CID 1591259:  Memory - illegal accesses  (OVERRUN)
/ldap/servers/plugins/memberof/memberof.c: 3780 in memberof_replace_list()

________________________________________________________________________________________________________
```
*** CID 1591259:  Memory - illegal accesses  (OVERRUN)
/ldap/servers/plugins/memberof/memberof.c: 3780 in memberof_replace_list()
3774     #pragma GCC diagnostic pop
3775                         rc = memberof_del_one(pb, config, group_sdn, sdn);
3776     
3777                         pre_index++;
3778                     } else {
3779                         /* decide what to do */
>>>     CID 1591259:  Memory - illegal accesses  (OVERRUN)
>>>     Overrunning dynamic array "pre_array" at offset corresponding to index variable "pre_index".
3780                         int cmp = memberof_compare(
3781                             config,
3782                             &(pre_array[pre_index]),
3783                             &(post_array[post_index]));
3784     
3785                         if (cmp < 0) {
```
coverity scan suspect that pre_index may be > post_total ( which is true if we considers only the if/else
 but the algorithm prevents it (pre_index is no more increased once post_total is reached.

progier389:
Fix trivial coverity scan issues

Issue:   #6604 

Reviewed by: @droideck, @tbordaz  (Thanks!)

### DIFF
--- a/ldap/servers/plugins/retrocl/retrocl_po.c
+++ b/ldap/servers/plugins/retrocl/retrocl_po.c
@@ -691,6 +691,12 @@ retrocl_postob(Slapi_PBlock *pb, int optype)
         (void)slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &post_entry);
         break;
     }
+    if (entry == NULL) {
+        /* To avoid coverity scan warning, but in practice, it should never happen. */
+        slapi_log_err(SLAPI_LOG_ERR, RETROCL_PLUGIN_NAME,
+                      "retrocl_postob - Operation on NULL entry\n", rc);
+        return 0;
+    }
 
     /* check if we should log change to retro changelog, and
      * if so, do it here */
--- a/ldap/servers/plugins/retrocl/retrocl_po.c
+++ b/ldap/servers/plugins/retrocl/retrocl_po.c
@@ -694,7 +694,7 @@ retrocl_postob(Slapi_PBlock *pb, int optype)
     if (entry == NULL) {
         /* To avoid coverity scan warning, but in practice, it should never happen. */
         slapi_log_err(SLAPI_LOG_ERR, RETROCL_PLUGIN_NAME,
-                      "retrocl_postob - Operation on NULL entry\n", rc);
+                      "retrocl_postob - Operation on NULL entry\n");
         return 0;
     }
 
--- a/ldap/servers/slapd/back-ldbm/ldbm_add.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_add.c
@@ -77,7 +77,7 @@ ldbm_back_add(Slapi_PBlock *pb)
     int rc = 0;
     int addingentry_id_assigned = 0;
     Slapi_DN *sdn = NULL;
-    Slapi_DN parentsdn;
+    Slapi_DN parentsdn = {0};
     Slapi_Operation *operation;
     int is_replicated_operation = 0;
     int is_resurect_operation = 0;
--- a/ldap/servers/slapd/back-ldbm/ldbm_delete.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_delete.c
@@ -57,7 +57,7 @@ ldbm_back_delete(Slapi_PBlock *pb)
     int is_ruv = 0; /* True if the current entry is RUV */
     int is_replicated_operation = 0;
     int is_tombstone_entry = 0;     /* True if the current entry is alreday a tombstone        */
-    int is_internal;
+    int is_internal = 0;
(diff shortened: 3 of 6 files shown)


ldap/servers/plugins/retrocl/retrocl_po.c
```diff
@@ -691,6 +691,12 @@ retrocl_postob(Slapi_PBlock *pb, int optype)
         (void)slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &post_entry);
         break;
     }
+    if (entry == NULL) {
+        /* To avoid coverity scan warning, but in practice, it should never happen. */
+        slapi_log_err(SLAPI_LOG_ERR, RETROCL_PLUGIN_NAME,
+                      "retrocl_postob - Operation on NULL entry\n", rc);
+        return 0;
```
tbordaz, line 698: So the operation is not logged in retroCL. Shouldn't it return an error (LDAP_OPERATIONS_ERROR ?)

tbordaz (approved):
LGTM. Minor comment

ldap/servers/plugins/retrocl/retrocl_po.c
progier389: I do not think it really matters, 
  Only abandon and extended operations (that are not logged in the retrocl anyway) does have a target entry)
  So the test is always false unless there is an heap/memory corruption.
    ( and frankly, at this point what we do does not matter much as the instance will probably crash anyway)
BTW if you really want to retuirn an error, LDAP_UNWILLING_TO_PERFORM will be better. Returning LDAP_OPERATIONS_ERROR in such case is an abuse. (LDAP_OPERATIONS_ERROR (i.e ldap operationms sequence erro) should only be returned by bind operation)

ldap/servers/plugins/retrocl/retrocl_po.c
```diff
@@ -691,6 +691,12 @@ retrocl_postob(Slapi_PBlock *pb, int optype)
         (void)slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &post_entry);
         break;
     }
+    if (entry == NULL) {
+        /* To avoid coverity scan warning, but in practice, it should never happen. */
+        slapi_log_err(SLAPI_LOG_ERR, RETROCL_PLUGIN_NAME,
+                      "retrocl_postob - Operation on NULL entry\n", rc);
```
droideck: I think it needs `%d` for the `rc`

droideck (approved):
Minor issue, the rest looks good to me!

ldap/servers/plugins/retrocl/retrocl_po.c
progier389: You are right: there is a problem, IMHO rc should better be removed  (no special rc in this test)

progier389: fixed in https://github.com/389ds/389-ds-base/pull/6605/commits/ee5d4d0fcd27eaa1c3f272124e3501bbd1bda3db commit

PR #6607, opened 2025-02-13
Issue 6553 - Update concread to 0.5.4 and refactor statistics tracking

### LINKED ISSUE #6553: Increased memory consumption caused by NDN cache
**Issue Description**

During online export, an increased memory consumption is observed. This is especially noticable with larger databases (1+ millions of entries), where database doesn't fully fit in memory, and during the export OOM-killer is invoked, killing the server.

I bisected the change that introduced the issue to https://github.com/389ds/389-ds-base/issues/4410.
If NDN cache is turned off during the export (`dsconf localhost config replace nsslapd-ndn-cache-enabled=off`), there is no increased memory consumption. Memory stays under 1.4GB, while with NDN cache it grows continiously. 

ASAN build doesn't show any leaks, but crashes with panic somewhere in concread code, when the export is done twice:
```
Jan 25 10:56:39 ns-slapd[444223]: [25/Jan/2025:10:56:39.595584783 -0500] - INFO - export_one_entry - export secauthority: Processed 42000 entries (3%).
Jan 25 10:56:39 ns-slapd[444223]: thread '<unnamed>' panicked at /root/ds/rpmbuild/BUILD/389-ds-base-1.4.3.39.20250125gitd0d980098/vendor/concread/src/arcache/mod.rs:1353:17:
Jan 25 10:56:39 ns-slapd[444223]: assertion failed: inner.ghost_rec.len() >= freq_to_len
Jan 25 10:56:39 ns-slapd[444223]: stack backtrace:
Jan 25 10:56:40 ns-slapd[444223]:    0: rust_begin_unwind
Jan 25 10:56:40 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/std/src/panicking.rs:652:5
Jan 25 10:56:40 ns-slapd[444223]:    1: core::panicking::panic_fmt
Jan 25 10:56:40 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/core/src/panicking.rs:72:14
Jan 25 10:56:40 ns-slapd[444223]:    2: core::panicking::panic
Jan 25 10:56:40 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/core/src/panicking.rs:146:5
Jan 25 10:56:41 ns-slapd[444223]:    3: concread::arcache::ARCache<K,V>::evict
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/vendor/concread/src/arcache/mod.rs:474:9
Jan 25 10:56:41 ns-slapd[444223]:    4: concread::arcache::ARCache<K,V>::commit
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/vendor/concread/src/arcache/mod.rs:1457:9
Jan 25 10:56:41 ns-slapd[444223]:    5: concread::arcache::ARCacheWriteTxn<K,V>::commit
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/vendor/concread/src/arcache/mod.rs:1506:9
Jan 25 10:56:41 ns-slapd[444223]:    6: concread::arcache::ARCache<K,V>::try_quiesce
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/vendor/concread/src/arcache/mod.rs:796:13
Jan 25 10:56:41 ns-slapd[444223]:    7: <concread::arcache::ARCacheReadTxn<K,V> as core::ops::drop::Drop>::drop
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/vendor/concread/src/arcache/mod.rs:1929:9
Jan 25 10:56:41 ns-slapd[444223]:    8: core::ptr::drop_in_place<concread::arcache::ARCacheReadTxn<alloc::ffi::c_str::CString,alloc::ffi::c_str::CString>>
Jan 25 10:56:41 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/core/src/ptr/mod.rs:514:1
Jan 25 10:56:41 ns-slapd[444223]:    9: core::ptr::drop_in_place<rslapd::cache::ARCacheCharRead>
Jan 25 10:56:41 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/core/src/ptr/mod.rs:514:1
Jan 25 10:56:41 ns-slapd[444223]:   10: core::ptr::drop_in_place<alloc::boxed::Box<rslapd::cache::ARCacheCharRead>>
Jan 25 10:56:41 ns-slapd[444223]:              at /builddir/build/BUILD/rustc-1.79.0-src/library/core/src/ptr/mod.rs:514:1
Jan 25 10:56:41 ns-slapd[444223]:   11: cache_char_read_complete
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/src/librslapd/src/cache.rs:92:5
Jan 25 10:56:41 ns-slapd[444223]:   12: ndn_cache_add
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/dn.c:3192:5
Jan 25 10:56:41 ns-slapd[444223]:   13: slapi_dn_normalize_ext
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/dn.c:1186:13
Jan 25 10:56:41 ns-slapd[444223]:   14: slapi_sdn_get_dn
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/dn.c:2374:18
Jan 25 10:56:41 ns-slapd[444223]:   15: slapi_sdn_get_ndn_len
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/dn.c:2590:11
Jan 25 10:56:41 ns-slapd[444223]:   16: slapi_sdn_get_size
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/dn.c:2751:11
Jan 25 10:56:41 ns-slapd[444223]:   17: backdn_init
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/back-ldbm/backentry.c:127:20
Jan 25 10:56:41 ns-slapd[444223]:   18: bdb_db2ldif
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c:1222:27
Jan 25 10:56:41 ns-slapd[444223]:   19: ldbm_back_ldbm2ldif
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/back-ldbm/ldif2ldbm.c:198:12
Jan 25 10:56:41 ns-slapd[444223]:   20: task_export_thread
Jan 25 10:56:41 ns-slapd[444223]:              at /usr/src/debug/389-ds-base-1.4.3.39-20250125gitd0d980098.el8.x86_64/ldap/servers/slapd/task.c:1189:32
Jan 25 10:56:41 ns-slapd[444223]:   21: <unknown>
Jan 25 10:56:41 ns-slapd[444223]:   22: start_thread
Jan 25 10:56:42 ns-slapd[444223]:   23: __clone
Jan 25 10:56:42 ns-slapd[444223]: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Jan 25 10:56:42 ns-slapd[444223]: fatal runtime error: failed to initiate panic, error 5
```

This applies to all versions after 1.4.3.32+.

droideck:
Description: Implement new cache statistics tracking with atomic counters
and dedicated stats structs.
Update concread dependency to 0.5.4 for improved cache performance.
Add tests for cache statistics functionality.

Fixes: https://github.com/389ds/389-ds-base/issues/6553

Reviewed by: ?

### DIFF
--- a/src/librslapd/src/cache.rs
+++ b/src/librslapd/src/cache.rs
@@ -1,38 +1,160 @@
 // This exposes C-FFI capable bindings for the concread concurrently readable cache.
+use concread::arcache::stats::{ARCacheWriteStat, ReadCountStat};
 use concread::arcache::{ARCache, ARCacheBuilder, ARCacheReadTxn, ARCacheWriteTxn};
-use std::convert::TryInto;
+use concread::cowcell::CowCell;
 use std::ffi::{CStr, CString};
 use std::os::raw::c_char;
 
+#[derive(Clone, Debug, Default)]
+struct CacheStats {
+    reader_hits: u64,      // Hits from read transactions (main + local)
+    reader_includes: u64,  // Number of includes from read transactions
+    write_hits: u64,       // Hits from write transactions
+    write_inc_or_mod: u64, // Number of includes/modifications from write transactions
+    freq_evicts: u64,      // Number of evictions from frequent set
+    recent_evicts: u64,    // Number of evictions from recent set
+}
+
+impl CacheStats {
+    fn new() -> Self {
+        CacheStats::default()
+    }
+
+    fn update_from_read_stat(&mut self, stat: ReadCountStat) {
+        self.reader_hits += stat.main_hit + stat.local_hit;
+        self.reader_includes += stat.include + stat.local_include;
+    }
+
+    fn update_from_write_stat(&mut self, stat: &FFIWriteStat) {
+        self.write_hits += stat.read_hits;
+        self.write_inc_or_mod += stat.includes + stat.modifications;
+        self.freq_evicts += stat.freq_evictions;
+        self.recent_evicts += stat.recent_evictions;
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct FFIWriteStat {
+    pub read_ops: u64,
+    pub read_hits: u64,
+    pub p_weight: u64,
+    pub shared_max: u64,
+    pub freq: u64,
+    pub recent: u64,
+    pub all_seen_keys: u64,
+    pub includes: u64,
+    pub modifications: u64,
+    pub freq_evictions: u64,
+    pub recent_evictions: u64,
+    pub ghost_freq_revives: u64,
+    pub ghost_rec_revives: u64,
+    pub haunted_includes: u64,
+}
+
+impl<K> ARCacheWriteStat<K> for FFIWriteStat {
+    fn cache_clear(&mut self) {
+        self.read_ops = 0;
+        self.read_hits = 0;
+    }
+
+    fn cache_read(&mut self) {
+        self.read_ops += 1;
+    }
+
+    fn cache_hit(&mut self) {
+        self.read_hits += 1;
+    }
+
+    fn p_weight(&mut self, p: u64) {
+        self.p_weight = p;
+    }
+
+    fn shared_max(&mut self, i: u64) {
+        self.shared_max = i;
+    }
+
+    fn freq(&mut self, i: u64) {
+        self.freq = i;
+    }
+
+    fn recent(&mut self, i: u64) {
+        self.recent = i;
+    }
+
+    fn all_seen_keys(&mut self, i: u64) {
+        self.all_seen_keys = i;
+    }
+
+    fn include(&mut self, _k: &K) {
+        self.includes += 1;
+    }
+
+    fn include_haunted(&mut self, _k: &K) {
+        self.haunted_includes += 1;
+    }
+
+    fn modify(&mut self, _k: &K) {
+        self.modifications += 1;
+    }
+
+    fn ghost_frequent_revive(&mut self, _k: &K) {
+        self.ghost_freq_revives += 1;
+    }
+
+    fn ghost_recent_revive(&mut self, _k: &K) {
+        self.ghost_rec_revives += 1;
+    }
+
+    fn evict_from_recent(&mut self, _k: &K) {
+        self.recent_evictions += 1;
+    }
+
+    fn evict_from_frequent(&mut self, _k: &K) {
+        self.freq_evictions += 1;
+    }
+}
+
 pub struct ARCacheChar {
     inner: ARCache<CString, CString>,
+    stats: CowCell<CacheStats>,
 }
 
 pub struct ARCacheCharRead<'a> {
-    inner: ARCacheReadTxn<'a, CString, CString>,
+    inner: ARCacheReadTxn<'a, CString, CString, ReadCountStat>,
+    cache: &'a ARCacheChar,
 }
 
 pub struct ARCacheCharWrite<'a> {
-    inner: ARCacheWriteTxn<'a, CString, CString>,
+    inner: ARCacheWriteTxn<'a, CString, CString, FFIWriteStat>,
+}
+
+impl ARCacheChar {
+    fn new(max: usize, read_max: usize) -> Option<Self> {
+        ARCacheBuilder::new()
+            .set_size(max, read_max)
+            .set_reader_quiesce(false)
+            .build()
+            .map(|inner| Self {
+                inner,
+                stats: CowCell::new(CacheStats::new()),
+            })
+    }
 }
 
 #[no_mangle]
 pub extern "C" fn cache_char_create(max: usize, read_max: usize) -> *mut ARCacheChar {
-    let inner = if let Some(cache) = ARCacheBuilder::new().set_size(max, read_max).build() {
-        cache
+    if let Some(cache) = ARCacheChar::new(max, read_max) {
+        Box::into_raw(Box::new(cache))
     } else {
-        return std::ptr::null_mut();
-    };
-    let cache: Box<ARCacheChar> = Box::new(ARCacheChar { inner });
-    Box::into_raw(cache)
+        std::ptr::null_mut()
+    }
 }
 
 #[no_mangle]
 pub extern "C" fn cache_char_free(cache: *mut ARCacheChar) {
-    // Should we be responsible to drain and free everything?
     debug_assert!(!cache.is_null());
     unsafe {
-        let _drop = Box::from_raw(cache);
+        drop(Box::from_raw(cache));
     }
 }
 
@@ -53,22 +175,26 @@ pub extern "C" fn cache_char_stats(
 ) {
     let cache_ref = unsafe {
         debug_assert!(!cache.is_null());
-        &(*cache) as &ARCacheChar
+        &(*cache)
     };
-    let stats = cache_ref.inner.view_stats();
-    *reader_hits = stats.reader_hits.try_into().unwrap();
-    *reader_includes = stats.reader_includes.try_into().unwrap();
-    *write_hits = stats.write_hits.try_into().unwrap();
-    *write_inc_or_mod = (stats.write_includes + stats.write_modifies)
-        .try_into()
-        .unwrap();
-    *shared_max = stats.shared_max.try_into().unwrap();
-    *freq = stats.freq.try_into().unwrap();
-    *recent = stats.recent.try_into().unwrap();
-    *freq_evicts = stats.freq_evicts.try_into().unwrap();
-    *recent_evicts = stats.recent_evicts.try_into().unwrap();
-    *p_weight = stats.p_weight.try_into().unwrap();
-    *all_seen_keys = stats.all_seen_keys.try_into().unwrap();
+
+    // Get accumulated stats
+    let stats_read = cache_ref.stats.read();
+    *reader_hits = stats_read.reader_hits;
+    *reader_includes = stats_read.reader_includes;
+    *freq_evicts = stats_read.freq_evicts;
+    *recent_evicts = stats_read.recent_evicts;
+    *write_inc_or_mod = stats_read.write_inc_or_mod;
+    *write_hits = stats_read.write_hits;
+
+    // Get the cache stats that are valid through the whole cache's lifetime
+    let write_txn = cache_ref.inner.write_stats(FFIWriteStat::default());
+    let write_stats = write_txn.commit();
+    *shared_max = write_stats.shared_max;
+    *freq = write_stats.freq;
+    *recent = write_stats.recent;
+    *p_weight = write_stats.p_weight;
+    *all_seen_keys = write_stats.all_seen_keys;
 }
 
 // start read
@@ -79,7 +205,8 @@ pub extern "C" fn cache_char_read_begin(cache: *mut ARCacheChar) -> *mut ARCache
         &(*cache) as &ARCacheChar
     };
     let read_txn = Box::new(ARCacheCharRead {
-        inner: cache_ref.inner.read(),
+        inner: cache_ref.inner.read_stats(ReadCountStat::default()),
+        cache: cache_ref,
     });
     Box::into_raw(read_txn)
 }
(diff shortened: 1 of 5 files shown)


droideck:
@Firstyear hi! Could you please take a look at the PR, as I'm not sure how to get:
`include, modify, evict_from_recent, and evict_from_frequent` working...

I've tried to implement the stats for them here, but I think I missed something, as I don't see them in the result stats.

Also, note that I created `CacheStats` for read_stats tracking because they are not collected from cache when I call `read_stats -> finish` in `cache_char_stats`. But I get a singular increase from the actual read transaction, so there is a way.:)

src/librslapd/src/cache.rs
```diff
@@ -78,17 +238,29 @@ pub extern "C" fn cache_char_read_begin(cache: *mut ARCacheChar) -> *mut ARCache
...
 }
 
 #[no_mangle]
 pub extern "C" fn cache_char_read_complete(read_txn: *mut ARCacheCharRead) {
     debug_assert!(!read_txn.is_null());
+
     unsafe {
-        let _drop = Box::from_raw(read_txn);
+        let read_txn_box = Box::from_raw(read_txn);
+        read_txn_box.cache.inner.try_quiesce();
```
Firstyear: you need to do try_quiesce_stats(); and you need to do it *after* you call .finish.

Firstyear:
> @Firstyear hi! Could you please take a look at the PR, as I'm not sure how to get: `include, modify, evict_from_recent, and evict_from_frequent` working...
> 
> I've tried to `impl<K> ARCacheWriteStat<K> for` them here, but I think I missed something, as I don't see them in the result stats.
> 
> Also, note that I created `CacheStats` for read_stats tracking because they are not collected from cache when I call `read_stats -> finish` in `cache_char_stats`. But I get a singular increase from the actual read transaction, so there is a way.:)

I think you may be misunderstanding how the stats api is working.

The stats struct you pass in is just triggered on those *events*. The cache itself *tracks nothing*. You need to track all the state and values. 

So this:  https://github.com/389ds/389-ds-base/pull/6607/files#diff-3292c7c1da4d212ee8ca277908c4b63bdd14216f92e4ea623782cd3f65fadad4R224 does *nothing* because between the start of the txn to the commit you never read or write from the txn, so the stats shows zero - because nothing was done during the transaction!!!!

You actually need to be putting in a stats tracker https://github.com/389ds/389-ds-base/pull/6607/files#diff-3292c7c1da4d212ee8ca277908c4b63bdd14216f92e4ea623782cd3f65fadad4R257 and then using that to adjust your hit/load amounts as they change. Because the quiesce step is where the read transaction inclusions/evictions are actually triggered. 

PS: Your cache stats don't actually need to be a struct of Atomics, you can make it a Struct inside a CowCell and then when someone wants to view the stats you read it, when a thread wants to update you take a write on the cow cell. That way it never blocks anything.

Firstyear:
> > So this: https://github.com/389ds/389-ds-base/pull/6607/files#diff-3292c7c1da4d212ee8ca277908c4b63bdd14216f92e4ea623782cd3f65fadad4R224 does _nothing_ because between the start of the txn to the commit you never read or write from the txn, so the stats shows zero - because nothing was done during the transaction!!!!
> 
> Sure, I totally understand the last part. I see, though, that during w_txn.,commit, we can get some stats that are valid through the whole cache's lifetime:
> 
> ```rust
>         stats.p_weight(inner.p as u64);
>         stats.shared_max(shared.max as u64);
>         stats.freq(inner.freq.len() as u64);
>         stats.recent(inner.rec.len() as u64);
>         stats.all_seen_keys(cache.len() as u64);
> ```

Still not correct - those are the current value of those stats at *this exact point in time*. They are not lifetime stats. You need to stop thinking that these values exist and are stored and tracked but the arcache, they aren't, you need to store them. The arcache doesn't track any statistics at all, the whole point of the stats api is to let you build something that tracks the values you are interested in. 

For example, between the first write txn and the second, the p_weight can move up, down, or stay the same. We don't tell you which way it went. All we do is tell you "this is the p_weight at this point in time". The same with freq/rec lengths, these values move both up and down. and "all seen keys" will either go up or stay the same, but we don't tell you the difference, just the value as it exists right now. You need to calculate the difference between the previous write txn and this write txn. 

For example if the p_weight moves down, that indicates pressure on the recent side of the cache, which causes the frequent side to be evicted. So you can use this api to work out "oh P weight has dropped by 10% of shared max, there must be pressure". But that requires you to check the p value between the first txn and the second, then to calculate that difference and ratio to shared max between both values, and make that assessment. We don't do that for you, we just show you the values at those points in time. 

Does that make more sense? 

> 
> IIUC, it's safe to get them only on the `cache_char_stats` call, as getting them each time we do a write txn's commit will be overkill. Right?

No, you have to do it each write txn commit. As mentioned the values are *not* lifetime stats, they are the exact values at *this* point in time, and you need to track those events yourself, including calculating the differences as those values at those points in time change. 

Cache_char_stats should do nothing except return the current value of stats *you* have been tracking.

Each time you call .commit or .finish, you need to update those stats, else your values will not be correct. 


> 
> Please, review! And thank you for your help with the developing concread understanding. I really appreciate that!

concread is a "fun one" because you have to think concurrently, about everything happening all at once. It can be tricky, but it's worth it. 

You can tihnk of a cowcell as a rwlock btw, where read/writes never block (unlike a rwlock).

tbordaz:
@Firstyear , reading your explanations my understanding is that Cache_char_stats will get sample of values. The values are valid but there may be small/significant variations between the two samples. Am I correct ?
If we want to compute a hit ratio (or eviction rate) during a specific testcase should we increase the sampling frequency ? Can intensive sampling  impact the cache efficiency itself ?

droideck:
@Firstyear Okay... Following your guidance, I've updated the code to collect stats from both `try_quiesce_stats` and read transaction `.finish()` results. And I've added a call for `cache_char_write_commit` even though we don't use it.

Another thing.
Interestingly, I've tried to remove stats collection and set `set_reader_quiesce(true)`, -> I saw a 10% performance improvement (I tested on a large export task).
So what I think, we could either:
1. Follow @tbordaz's suggestion to limit quiesce frequency (I'm afraid, though, it can affect the cache validity)
2. Make NDN cache stats collection optional via a config setting (could be a separate ticket)

BTW, I'm also seeing concerning memory behavior that needs investigation. With NDN cache enabled, memory usage grows significantly during exports and isn't freed afterward (unless we restart, of course). With NDN cache **disabled**, memory usage stays normal. ASAN shows no issues. Also, I've checked if all write and read transactions are freed on the 389 DS side, and all is good (I counted new and closed txns and the numbers where equal after the massive online export).

This suggests either a concread internal issue or misuse on our part. My current thoughts:
- Could unlimited ghost set growth be the culprit? Is there an automatic ghost-cleaning mechanism, or should we call something with a `clean` flag somewhere?

Would appreciate any insights on the memory growth issue, but I'll keep investigating regardless...

Firstyear:
> @Firstyear Okay... Following your guidance, I've updated the code to collect stats from both `try_quiesce_stats` and read transaction `.finish()` results. And I've added a call for `cache_char_write_commit` even though we don't use it.
> 
> Another thing. Interestingly, I've tried to remove stats collection and set `set_reader_quiesce(true)`, -> I saw a 10% performance improvement (I tested on a large export task). So what I think, we could either:
> 
>     1. Follow @tbordaz's suggestion to limit quiesce frequency (I'm afraid, though, it can affect the cache validity)

Collect, if you don't quiesce then new cache items will never actually be included or accessible, you need to try_quiesce after any read op. 

Consider reading https://github.com/kanidm/concread/blob/master/CACHE.md to understand why. The inclusion and tracking queues are bounded by the way, so they can only "buffer" so much. 

> 
>     2. Make NDN cache stats collection optional via a config setting (could be a separate ticket)

Yeah, that sounds optional and better done later. 

> 
> 
> BTW, I'm also seeing concerning memory behavior that needs investigation. With NDN cache enabled, memory usage grows significantly during exports and isn't freed afterward (unless we restart, of course). With NDN cache **disabled**, memory usage stays normal. ASAN shows no issues. Also, I've checked if all write and read transactions are freed on the 389 DS side, and all is good (I counted new and closed txns and the numbers where equal after the massive online export).
> 
> This suggests either a concread internal issue or misuse on our part. My current thoughts:
> 
>     * Could unlimited ghost set growth be the culprit? Is there an automatic ghost-cleaning mechanism, or should we call something with a `clean` flag somewhere?
> 
> Would appreciate any insights on the memory growth issue, but I'll keep investigating regardless...

https://github.com/kanidm/concread/?tab=readme-ov-file#what-is-concurrently-readable

"This is a space-time trade off, using more memory to achieve better parallel behaviour."

Which is what you are seeing here. We have to keep the ghost lists to track *time* as well, because we need to be able to order cache events based on time too. 

We *could* investigate ways to trim the ghost lists more effectively based on epochs or similar. But yes, this is the trade of concread, you gain by having non-blocking parallel operations, but you are paying by using more ram. 

Realistically, the cache will grow to some X point, and then plateau because the ghost lists then settle. The amount in them shouldn't be so huge that it's a problem, and if it is then perhaps expanding the cache is worth more. This is why you could consider looking at the ghost lists as a stat to collect too. 

Anyway, I've opened https://github.com/kanidm/concread/issues/128 which I'll work on today for you.

Firstyear (approved):
Outside of the changes I need to make in concread, looks good.

Firstyear:
@droideck https://github.com/kanidm/concread/pull/129 is done, will merge and release today. Should just be a version bump to 5.4 over 5.3 for you.

Firstyear:
@droideck 5.4 is out, give it a try :)

droideck:
> @droideck 5.4 is out, give it a try :)

Much better! The instance used only 5% or my machine's memory during the huge export. (it was nearly 19% before the update)
And I see 5% performance improvement for the task.
Thanks!

Firstyear (approved):
Then lets merge it :)

Issue #6611, opened 2025-02-14
db_home_test.py CI tests should be skipped on lmdb


progier389:
**Issue Description**
dirsrvtests/tests/suites/setup_ds/db_home_test.py are failing on 389-ds-base-3.0.6
Since dbhome is meaningless on lmdb (because there is no database cache nor txn log files), these tests should be skipped.
And these tests should also be fixed for non standard cases 
  (non root install, prefix install)

**Package Version and Platform:**
 - Platform: RHEL 10 image mode
 - Package and version: 389-ds-base-3.0.6

**Steps to Reproduce**
1. Install 389-ds
2. Run setup_ds CI tests

**Expected results**
All tests should pass

progier389:
7018e91a7..aabfec1d6  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6612, opened 2025-02-14
Issue 6611 - db_home_test.py CI tests should be skipped on lmdb

progier389:
dbhome setup tests should be skipped on lmdb (dbhome is meaningless)

Also fix the tests for non standard deployment (typically when using PREFIX / non root install)

Issue: #6611 

Reviewed by:  @droideck, @tbordaz (Thanks!)


droideck (approved):
Works on my machine! (LMDB is not ran, BDB is green)
Ack.

tbordaz (approved):
LGTM

PR #6620, opened 2025-02-17
Issue 6619 - test_dblib_migration fails on RHEL10

### LINKED ISSUE #6619: test_dblib_migration fails on RHEL10
**Issue Description**
CI tests/suites/clu/dsctl_dblib_test.py::test_dblib_migration test fails on RHEL 10 image mode
because 389ds instances cannot start in bdb mode (which is not supported by default on RHEL10)

==> The test should be skipped  if bundled bdb is not installed.

 - Platform: Redhat 10
 - Package and version: 389-ds-base-3.0.6

**Steps to Reproduce**
1. Run CI tests on RHEL 10 

**Expected results**
Test should be skipped

**Additional context**
Note: IMHO the fix impacts both CI test and lib389

progier389:
Test test_dblib_migration fails on RHEL10 because bdb is not supported. 
Test should be skipped in that case and libdb should provide a function to check if server can starts on bdb

Issue: #6619 

Reviewed by: @jchapma (Thanks!)

### DIFF
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -51,7 +51,7 @@ class FakeArgs(dict):
 
 def get_bdb_impl_status():
     backldbm = 'libback-ldbm'
-    bundledbdb_plugin = 'libback-ldbm'
+    bundledbdb_plugin = 'libback-bdb'
     robdb_symbol = 'bdbro_getcb_vector'
     libdb = 'libdb-'
     plgstrs = check_plugin_strings(backldbm, [bundledbdb_plugin, robdb_symbol, libdb])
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -70,6 +70,14 @@ def get_bdb_impl_status():
     return BDB_IMPL_STATUS.UNKNOWN
 
 
+def is_bdb_supported(read_write=True):
+    bdbok = [BDB_IMPL_STATUS.BUNDLED, BDB_IMPL_STATUS.STANDARD]
+    if noread_write:
```
jchapma: typo

progier389: Now fixed.

jchapma (approved):
LGTM

Issue #6571, opened 2025-02-03
Nested group does not receive memberOf attribute.


egerber86:
Problem repeats constantly in the next case (group names are important):

There are three groups: group1, group2, group21. There is one user: user1.

group21 is a member of group2.
user1 is a member of group1 and group21.

When I add group2 into group1, group21 does not receive memberof group1.

The problem can be solved by commenting the next block of code: https://github.com/389ds/389-ds-base/blob/main/ldap/servers/plugins/memberof/memberof.c#L3401C1-L3414C6

Does this check is correct and necessary ?

dorokhinra:
@progier389 In other words, the error should be looked for in why dn got into Folder_seen_ndn_vals ? And the check itself https://github.com/389ds/389-ds-base/blob/main/ldap/servers/plugins/memberof/memberof.c#L3401C1-L3414C6 is it correct?

egerber86:
Hello, 

I found that group21 get into already_seen_ndn_vals during processing of user1. 

That behaivior looks suspicious because users are not groups and they are always leafs in a nested groups tree.

It is possible to forbid filling already_seen_ndn_vals during processing of "leafs" objects. 

As an option, we can add additional attribute to the memberof config with objectClass that described which objects are "leafs" or with a path where "leaf"-objects are stored.

tbordaz:
@egerber86 , thanks for investigations and perfect testcase. I agree there is a bug and I can reproduce it easily.
I submitted another PR and will update your PR.

tbordaz:
d3f58b483..a9509045f  main
0299ce5f3..7a720f4d7  389-ds-base-3.0
772e29a9f..a87bf2eb1  389-ds-base-2.6
e18ca4809..96d0c77bc  389-ds-base-2.5
aeb632807..fd08e52f8  389-ds-base-2.4
e5e43ef97..952112fdd  389-ds-base-2.3
288ad36ed..993424afe  389-ds-base-2.2
bc2e91ef4..7cc3ee8bd  389-ds-base-2.1
e47219610..32a773538  389-ds-base-2.0
c687eafd2..f94c1d3ad  389-ds-base-1.4.4
f760548c0..dd337449e  389-ds-base-1.4.3

tbordaz:
It fixes the regression [6692](https://github.com/389ds/389-ds-base/issues/6692) 
5078fd321..2e108e625  main
ded674a5c..13cb8bedd  389-ds-base-3.0
[...]

PR #6628, opened 2025-02-19
Issue 6571. Do not add groups to already_seen for users

egerber86:
Variable already_seen_ndn_val does not filling during the users processing.

Issue: https://github.com/389ds/389-ds-base/issues/6571

Reviewed by: @progier389

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -2288,6 +2288,44 @@ memberof_get_groups_r(MemberOfConfig *config, Slapi_DN *member_sdn, memberof_get
     return rc;
 }
 
+int 
+check_if_member_is_person(Slapi_Value* member) {
+    int found = 0;
+    if (member == NULL) {
+        return found;
+    }
+    
+    const char * member_str = slapi_value_get_string(member);
+    Slapi_Entry *member_entry = NULL;
+    Slapi_PBlock *member_pb = NULL;
+    Slapi_DN *member_sdn = NULL;
+    member_sdn = slapi_sdn_new_normdn_byref(member_str);
+
+    slapi_search_get_entry(&member_pb, member_sdn, NULL, &member_entry, memberof_get_plugin_id());
+    if (member_entry) {
+        Slapi_Attr *attr = NULL;
+        slapi_entry_attr_find(member_entry, "objectclass", &attr);
+        if (attr) {
+            Slapi_Value *sval = NULL;
+            int i;
+            const struct berval *attrVal;
+            i = slapi_attr_first_value(attr, &sval);
+            while (i != -1) {
+                attrVal = slapi_value_get_berval(sval);
+                if (strcasecmp(attrVal->bv_val, "person") == 0) {
+                    found = 1;
+                    break;
+                }
+                i = slapi_attr_next_value(attr, i, &sval);
+            }
+        }
+    }
+    slapi_sdn_free(&member_sdn);
+    slapi_search_get_entry_done(&member_pb);
+    
+    return found;
+}
+
 /* memberof_get_groups_callback()
  *
  * Callback to perform work of memberof_get_groups()
@@ -2367,8 +2405,10 @@ memberof_get_groups_callback(Slapi_Entry *e, void *callback_data)
         slapi_valueset_add_value_ext(groupvals, group_dn_val, SLAPI_VALUE_FLAG_PASSIN);
 
         /* push this ndn to detect group recursion */
-        already_seen_ndn_val = slapi_value_new_string(group_ndn);
-        slapi_valueset_add_value_ext(already_seen_ndn_vals, already_seen_ndn_val, SLAPI_VALUE_FLAG_PASSIN);
+        if (check_if_member_is_person(((memberof_get_groups_data *)callback_data)->memberdn_val) == 0) {
+            already_seen_ndn_val = slapi_value_new_string(group_ndn);
+            slapi_valueset_add_value_ext(already_seen_ndn_vals, already_seen_ndn_val, SLAPI_VALUE_FLAG_PASSIN);
+        }
     }
     if (!config->skip_nested || config->fixup_task) {
         /* now recurse to find ancestors groups of e */


egerber86:
Hello @progier389, 

Do you have any comments about PR ?

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -2288,6 +2288,43 @@ memberof_get_groups_r(MemberOfConfig *config, Slapi_DN *member_sdn, memberof_get
...
+                attrVal = slapi_value_get_berval(sval);
+                if (strcasecmp(attrVal->bv_val, "person") == 0) {
+                    found = 1;
+                    break;
+                }
+                i = slapi_attr_next_value(attr, i, &sval);
+            }
+        }
+    }
+    slapi_sdn_free(&member_sdn);
```
droideck, line 2323: I think we also need to call `slapi_search_get_entry_done(&member_pb);` here

droideck (changes requested):
I don't have deep memberof knowledge though, so I'll leave the `if (check_if_member_is_person(((memberof_get_groups_data *)callback_data)->memberdn_val) == 0) {` part to @progier389 (I'll dive deeper into the issue, if time allows later this sprint).

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -2288,6 +2288,43 @@ memberof_get_groups_r(MemberOfConfig *config, Slapi_DN *member_sdn, memberof_get
...
 }
 
+int 
+check_if_member_is_person(Slapi_Value* member) {
+    int found = 0;
+    if (member == NULL) {
+        return found;
+    }
+    
+    char * member_str = slapi_value_get_string(member);
```
dorokhinra: I think const char* member_str : slapi_value_get_string return const char*,  also slapi_sdn_new_normdn_byref(const char *)

egerber86: Done

ldap/servers/plugins/memberof/memberof.c
egerber86: Done

egerber86 (commented):
I have fixed all comments

progier389:
I have some bad feelings about check_if_member_is_person:
  Group members are not always user accounts (even if it is the common case) so it seems wrong to user "person" objectclass as discriminant.
  A minor point: should better use slapi_attr_value_find instead of walking the values and comparing them 

But I wonder if there are not some accelerator using wrongly already_seen_ndn_val that cause the problem.,
IMHO the problem is that the group is added in already_seen_ndn_val before getting fully processed
(but on the other way it is understandable because doing this avoid loops if there are circular membership definition

We should probably spend the time to try to understand why its happen before being able to fix it. 
(and maybe lets wait for @tbordaz feedback because he has better knowledge of this plugin)

dorokhinra:
@progier389 The memberof_get_groups_callback function is focused on working with group objects and is not aimed at processing the person object. Since the person object itself eliminates the occurrence of recursion. You cannot "lock" an object with a reference to yourself. Maybe it's better to exclude the person object from entering this function?

tbordaz:
already_seen_ndn_val was introduced in https://fedorahosted.org/389/ticket/49031 in order to [improve performance](https://www.port389.org/docs/389ds/design/memberof-scalability.html#fixup-get_groups-structure) of memberof plugin.

One of the main contributor of the perf hit was that a same entry was fixup (evaluating/updating the groups it was member of) several times.
In fact there are possibly multiple paths from a top group to a leaf. For example, adding 'team_T' to the 'CI' then in the following membships, the entry contributor_A, team_T, CI and company were fixup several times.
<pre>
company->engineering->CI->team_T->contributor_A
company->sales->projects_B->CI->team_T->contributor_A
</pre>
 To prevent to do and redo the fixup of a same entry the already_seen_ndn_val keeps, along the recursion, the already fixup (seen) entries. 

I agree that vast the majority of the cost was due to several fixup of the same groups (intermediate nodes). But already_seen_ndn_val was also intended to contain leaf entries (e.g. 'person' entry).

So IMHO the suggested code cleanup will not give much perf improvement and also allows to fixup several times the same 'person'

progier389:
@tbordaz 
Thanks for the explanation (so if I understanding rightly this PR will probably have a noticeable performance impact ...)
but FYI, it is not about code cleanup but about fixing a real bug (as explained in the issue, the membership is not rightly evaluated)
So there is still some more work to do on that one (to find a better fix).
IMHO the first step will be to understand why already_seen_ndn_val prevent the member to be added as it should in the test case scenario ...

egerber86:
Hello, I read the code and found that memberof_call_foreach_dn function either calls memberof_get_groups_callback or take groupvals from the cache. 

I found that user1 is processed first just because it is the first member in group1. During that processing creates cache for group2 and group21. Cache for group21 is not fully filled probably because group2 is a member of group1, so we can see a message "Possible group recursion cn=group1...". 

I attach my log for the problematic case.

```
Start processing sdn uid=user1,cn=users,cn=accounts,dc=example,dc=com 
group_ndn_val 1 
group_ndn_val cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
groupvals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_norm_vals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_ndn_val 2 
group_ndn_val cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
groupvals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
groupvals cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
group_norm_vals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_norm_vals cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
group_ndn_val 1 
group_ndn_val cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
groupvals cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
group_norm_vals cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
group_ndn_val 1 
group_ndn_val cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
Possible group recursion cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
groups====== cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
groups====== cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
groups====== cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
loop
Start processing sdn cn=group21,cn=groups,cn=accounts,dc=example,dc=com 
cached===== cn=group21,cn=groups,cn=accounts,dc=example,dc=com
groups====== cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
loop
Start processing sdn cn=group2,cn=groups,cn=accounts,dc=example,dc=com 
group_ndn_val 1 
group_ndn_val cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_sdn cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
groupvals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
group_norm_vals cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
already_seen cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
cached===== cn=group1,cn=groups,cn=accounts,dc=example,dc=com
groups====== cn=group1,cn=groups,cn=accounts,dc=example,dc=com 
loop
```


I also would like to notice that, If I create a group with name "user1" everything will be OK.

So, I still suppose that excluding "person" objects from filling already_seen_ndn_val is a good idea.

tbordaz:
Hi @egerber86 , thank you for all your analyze/data. I understand that your fix is working but it remains unclear to me why it is working. This code is quite old in my mind I am likely missing some points. I plan to work on the ticket this week.

droideck (commented):
The change I requested is done.
But yeah, we still need to investigate the issue deeper. Hence, not setting `Approve`.

tbordaz:
This PR was a proposal for fixing the bug #6571 in 2.3. The PR fixed the testcase but was not fully valid as it did not add 'person' in already_seen table where it should be added since we already processed it.
Later another fix was pushed it is [here](https://github.com/389ds/389-ds-base/commit/952112fdd673ce4775403013898d010d423e1985#diff-88d1c22c85907bacbe38a1d5d9c4f3935f44f721784a2235fd1d48a049c75638). This current PR is no longer applicable and I propose to close it.

Issue #6327, opened 2024-09-18
Using a pointer size instead of buffer size at function get_ip_str


PavlNekrasov:
**Issue Description**

https://github.com/389ds/389-ds-base/blob/82dcdab11a6b9b6baeaa7af8738443cf94a469bb/ldap/servers/slapd/connection.c#L239-L260
 In this case, sizeof(str) returns the size of the pointer, but this does not check for the actual buffer length. 

Also, the if statement `(sizeof(str) < INET_ADDRSTRLEN)` and `(sizeof(str) < INET6_ADDRSTRLEN)`  will always evaluate to true and the get_ip_str function will return the original str.

**Package Version and Platform:**
 - Platform: ALT Linux 10
 - Package and version: main

**Additional context**
Found by Linux Verification Center (linuxtesting.org) with SVACE.
Reporter: Pavel Nekrasov ([p.nekrasov@fobos-nt.ru](mailto:p.nekrasov@fobos-nt.ru)).
Organization: Fobos-NT ([info@fobos-nt.ru](mailto:info@fobos-nt.ru)).

PR #6629, opened 2025-02-19
Fix incorrect `sizeof()` usage for pointer in `get_ip_str()` function

PavlNekrasov:
Fix for issue: https://github.com/389ds/389-DS-Base/issues/6327


progier389 (approved):
LGTM

Issue #6626, opened 2025-02-18
dsctl healthcheck should not report busy replica as ERROR


progier389:
**Issue Description**
If there is more than two suppliers, a replica agreement status can be in replica busy 
(meaning  that on last attempt the target replica was busy (receiving updates from another node))
In this case `dsctl instance healthcheck` is reporting an error:
 ```
 {
    "source": "ipahealthcheck.ds.replication",
    "check": "ReplicationCheck",
    "result": "ERROR",        <------------------------
    "uuid": "blah-blah-blah-blah",
    "when": "20250205035523Z",
    "duration": "0.199898",
    "kw": {
      "key": "EXAMPLE007",
      "items": [
        "Replication",
        "Agreement"
      ],
      "msg": "The replication agreement (bond.spy.example.com) under \"dc=ipa,dc=spy,dc=example,dc=com\" is not in synchronization.\nStatus message: error (1) can't acquire busy replica (unable to acquire replica: the replica is currently being updated by another supplier.)"
    }
```
Although it is a normal situation

**Package Version and Platform:**
 - Platform:  Fedora
 - Package and version:   main branch

**Steps to Reproduce**
1. Setup a three supplier with fully meshed agreement 
2. Generate a constant modify load on suppliers 1 and 2
3. Perform a change on supplier 3
4. Run dsctl supplier3 healthcheck
5. See that healthcheck report an error aboput busy replica

**Expected results**
No errors should be reported

**Additional context**
Also reported on RHEL 9: https://issues.redhat.com/browse/RHEL-79673

Triaged: https://issues.redhat.com/browse/IDMDS-5275

progier389:
f8c9704c7..bc22cfa61  main -> main
   f16e1870e..0ab605e44  389-ds-base-3.0 -> 389-ds-base-3.0
   feb341c08..2763de6cf  389-ds-base-2.6 -> 389-ds-base-2.6

tbordaz:
404a71564..a415141f8  389-ds-base-2.5
75b4e5014..9c9bcfb89  389-ds-base-2.4
776bd7165..7cbca1471  389-ds-base-2.3
050c93c3d..aa5d0ecbf  389-ds-base-2.2

PR #6630, opened 2025-02-20
Ignore replica busy condition in healthcheck

progier389:
Replica Busy condition is expected when there is more than 2 suppliers so healthcheck should not report any error for such condition.
Fixed issue in CI tests:
  - test_healthcheck_replication_out_of_sync_not_broken was unstable and redundant with test_healthcheck_replica_busy so I moved it back in health_repl_test.py and rewrite it to test working replication whose replica are not in sync (healthcheck should not report anything) 
  - some tests (not always the same) were randomly failing with an empty log (which is unexpected because healthcheck output should never be empty.
   I suspect the log capture mechanism so health_repl_test.py now run dsctl instance healthcheck using subprocess module and capture the output. So far I have not changed the other files because I have not noticed any failure.

Issue: #6626 


tbordaz (approved):
LGTM
I think you also need to update existing calls to run_healthcheck_and_flush_log to set isnot=True|False.

progier389:
test_healthcheck_replication_out_of_sync_broken is failing:
`ERROR:dirsrvtests.tests.suites.healthcheck.health_repl_test:DSREPLLE0001 is not in healthcheck output:  `
Both in bdb and lmdb
Maybe there is a side effect ?

progier389:
Apparently that is a test case issue: test_healthcheck_replication_out_of_sync_broken has a race condition allowing to run healthcheck before replication is out of sync (because agmt fro M1 to M2,M3 are not broken.
Changing the test to do updates on all instances, and restore broken replica once the test is finished.

dirsrvtests/tests/suites/healthcheck/health_repl_test.py
```diff
@@ -265,25 +371,52 @@ def test_healthcheck_replication_out_of_sync_broken(topology_m3):
...
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Healthcheck reports no issue found
+        4. Healthcheck reports no issue found
+    """
+
+    RET_CODE = CMD_OUTPUT
+    # Is DSREPLLE0003 ignored if replica is busy ?
+    ignored = not ds_is_older("2.7")
```
progier389: needless leftover ...

progier389:
Test are now pass except an unrelated issue. ( https://github.com/389ds/389-ds-base/issues/6671 )
As there was quite some change in the CI test since thierry reviewed it, I would that someone has a look at it before doing the commit ...

PR #6631, opened 2025-02-23
Issue 6625 - UI - fix next round of bugs

### LINKED ISSUE #6625: UI - fix various issues
**Issue Description**

There area bunch of minor issues to fix around ldap editor editors and alert messages

mreynolds389:
Description:

This fixes the following issues:

- Subtrees field must not be mandatory If you configure the Attribute Uniqueness plug-in over object classes
- MemberOf configuration window you can select empty values for Subtree Scope and Exclude Subtree fields
- Database suffix field accepts dash symbol which ends in error
- Enable logs checkboxes do not display properly after changes
- Instance fails to load when DB backup directory doesn't exist
- Improvement for Create New Instance form
- Database Name is not validated
- Database Suffix is not validated properly
- Failing to update NDN cache size
- Import cache size is not easy to set
- Import cache autosize off/on fails
- Statistic about worker threads is incorrect
- Database index table searching does not work
- Schema attribute table searching does not work
- TimePicker component needs updating for PF5
- Fix Treeview appearance for browser Dark Mode
- Add attribute syntax oid name to table

Fixes: https://github.com/389ds/389-ds-base/issues/6625

### DIFF
--- a/src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
+++ b/src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
@@ -1,16 +1,21 @@
 import cockpit from "cockpit";
 import React from "react";
 import {
+    Bullseye,
 	Button,
 	Checkbox,
 	Form,
+    FormHelperText,
 	FormSelect,
 	FormSelectOption,
 	Grid,
 	GridItem,
+    HelperText,
+    HelperTextItem,
 	Modal,
 	ModalVariant,
 	TextInput,
+    Spinner,
 	Switch,
 	ValidatedOptions
 } from '@patternfly/react-core';
@@ -40,6 +45,8 @@ class AttributeUniqueness extends React.Component {
         super(props);
         this.state = {
             firstLoad: true,
+            saving: false,
+            loading: true,
             configRows: [],
             modalChecked: false,
             modalSpinning: false,
@@ -52,6 +59,8 @@ class AttributeUniqueness extends React.Component {
             attrNames: [],
             subtrees: [],
             subtreesOptions: [],
+            excludeSubtrees: [],
+            excludeSubtreesOptions: [],
             acrossAllSubtrees: false,
             topEntryOc: "",
             subtreeEnriesOc: "",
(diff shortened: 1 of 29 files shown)


src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
```diff
@@ -121,18 +133,23 @@ class AttributeUniqueness extends React.Component {
 
         // Subtrees
         this.handleSubtreesSelect = (event, selection) => {
+            console.error
```
droideck: Looks like a stray console.error typo

src/cockpit/389-console/src/lib/tools.jsx
```diff
@@ -243,9 +243,48 @@ export function valid_dn(dn) {
...
+    return true;
+}
+
+export function valid_db_name(name) {
+    // Validate name is a valid backend name
+    if (name === "") {
+        return false;
+    }
+
+    const name_regex = /^[A-Za-z_\-]+$/;
```
droideck: IIRC, we accept numbers in the name too but it should start with a letter.

src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
```diff
@@ -439,49 +527,75 @@ class AttributeUniqueness extends React.Component {
...
+                    cmd = [...cmd, "delete"];
+                } else {
+                    for (const value of subtrees) {
+                        cmd = [...cmd, value];
+                    }
+                }
+            }
+            if (!listsEqual(this.state['excludeSubtrees'], this.state['_excludeSubtrees'])) {
+                cmd = [...cmd, "--exclude-subtree"];
+                if (exclude_subtrees.length === 0) {
```
droideck: Should it be `excludeSubtrees` instead of `exclude_subtrees`?

droideck (approved):
A couple of minor issues.
The rest looks great! Ack

src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
mreynolds389: > Should it be `excludeSubtrees` instead of `exclude_subtrees`?

Nice catch!

mreynolds389:
> A couple of minor issues. The rest looks great! Ack

Thanks for the review!  Got one more PR coming after this one and then we should be good!

PR #6633, opened 2025-02-25
Issue 6632 - Replication init fails with ASAN build

### LINKED ISSUE #6632: Replication init fails with ASAN build
**Issue Description**
After the increase of DEFAULT_PBKDF2_ROUNDS to 100_000, bind operation on ASAN build takes more time. On some systems it's more time than the default timeouts for replication agreements, causing replication init to fail.

vashirov:
Bug Description:
After the increase of DEFAULT_PBKDF2_ROUNDS to 100_000, bind operation on ASAN build takes more time. On some systems it's more time than the default timeouts for replication agreements, causing replication init to fail.

Fix Description:
Introduce a new helper function `get_timeout_scale()` and update hardcoded timeouts to use it for scaling. For ASAN builds it returns 2.0 instead of default 1.0.
Factor returned by `get_timeout_scale()` can be adjusted using environment variable `DS_TIMEOUT_SCALE`.

Fixes: https://github.com/389ds/389-ds-base/issues/6632

### DIFF
--- a/src/lib389/lib389/replica.py
+++ b/src/lib389/lib389/replica.py
@@ -30,7 +30,7 @@ from lib389.properties import REPLICA_OBJECTCLASS_VALUE, REPLICA_OBJECTCLASS_VAL
 
 from lib389.utils import (normalizeDN, escapeDNValue, ensure_bytes, ensure_str,
                           ensure_list_str, ds_is_older, copy_with_permissions,
-                          ds_supports_new_changelog)
+                          ds_supports_new_changelog, get_timeout_scale)
 from lib389 import DirSrv, Entry, NoSuchEntryError, InvalidArgumentError
 from lib389._mapped_object import DSLdapObjects, DSLdapObject
 from lib389.passwd import password_generate
@@ -2074,6 +2074,13 @@ class ReplicationManager(object):
 
         agmt_name = self._inst_to_agreement_name(to_instance)
 
+        # Default timeout
+        timeout = 5
+        # Scaled timeout
+        timeout = round(timeout * get_timeout_scale())
+        # Ensure the timeout is not too small
+        timeout = max(timeout, 1)
+
         # add a temp agreement from A -> B
         from_agreements = from_replica.get_agreements()
         temp_agmt = from_agreements.create(properties={
@@ -2082,7 +2089,7 @@ class ReplicationManager(object):
             'nsDS5ReplicaBindDN': brm.dn,
             'nsDS5ReplicaBindMethod': 'simple' ,
             'nsDS5ReplicaTransportInfo': 'LDAP',
-            'nsds5replicaTimeout': '5',
+            'nsds5replicaTimeout': str(timeout),
             'description': "temp_%s" % agmt_name,
             'nsDS5ReplicaHost': to_instance.host,
             'nsDS5ReplicaPort': str(to_instance.port),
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/replica.py
```diff
@@ -2362,13 +2365,16 @@ def ensure_agreement(self, from_instance, to_instance, init=False):
         assert dn is not None
         assert creds is not None
 
+        timeout = 5
+        timeout = int(timeout * get_timeout_scale())
```
droideck: I think `round()` will give more correct results.

Like, if DS_TIMEOUT_SCALE=1.6 and `timeout` is `6`, then:
- float is 9.6
- `int(timeout * get_timeout_scale()) == 9`
- `round(timeout * get_timeout_scale()) == 10`

src/lib389/lib389/utils.py
```diff
@@ -2013,3 +2013,14 @@ def check_plugin_strings(plugin_name, tested_strings):
...
+
+def get_timeout_scale():
+    """
+    Get the timeout scale factor
+    :return float:
+    """
+    scale_factor = 1.0
+    if Paths().asan_enabled:
+        scale_factor = 2.0
+    return float(os.getenv('DS_TIMEOUT_SCALE', default=scale_factor))
```
droideck: Might worth to validate the env var here. It'll raise a bit cryptic (for user) message otherwise (for example, if `DS_TIMEOUT_SCALE=testvar`):

```
ValueError: could not convert string to float: 'testvar'
```

droideck (changes requested):
Couple of small issues, - otherwise, looks good!

src/lib389/lib389/replica.py
vashirov: It's not meant to be precise, since it applies to many different timeouts. And I used float to also be able to reduce the default timeouts, for example 0.5 to cut the default timeouts in half. For ASAN builds on slower VMs you want to use 2 or 3 just to make the tests pass.

src/lib389/lib389/utils.py
vashirov: What behaviour do you want in case the validation fails? Set the timeout scale to 1.0 or log some error? This env var is not meant to be used by the end user, it's for us to deal with occasional slowness of the test infra.

src/lib389/lib389/replica.py
droideck: Sure, I'm okay with both options. I just thought that the intention was to round the number. If we are okay with just discarding any remainder, I see no issues with using `int()`.

src/lib389/lib389/utils.py
droideck: I think an error that mentions the env var will work the best. It will help us to see exactly where is issue (and it'll be future-proof, in case anyone else will use the internal feature for some reason)

progier389: IMHO we should check that it is a float and its value is >= 1/timeout (and just tell that a float >= 0.2 is expected 
Even if it is only for tester, a typo is so quick to do and a clear message will avoid us spending too much time to understand whats is happening!

vashirov: I've added an error with a fallback to the default value in case it's an invalid value:
```
ERROR:lib389.utils:DS_TIMEOUT_SCALE should be a valid float. Using default value: 2.0
```

src/lib389/lib389/replica.py
vashirov: Ok, I replaced it with `round()` everywhere.

vashirov:
Requested changes implemented, please review.

droideck (approved):
LGTM!

PR #6634, opened 2025-02-25
Revert "Issue #6562 - Prevent undefined behaviour in in filter_stuff_func (#6563)"

progier389:
This reverts commit c27fe457431b78c4a7d9e8681a9462def67a76ba.

Reverting c27fe457431b78c4a7d9e8681a9462def67a76ba that is causing
 an heap corruption: buffer overflow

Issue: #6562

Reviewed by: @tbordaz , @droideck (Thanks!)

### DIFF
--- a/ldap/servers/slapd/util.c
+++ b/ldap/servers/slapd/util.c
@@ -338,10 +338,6 @@ filter_stuff_func(void *arg, const char *val, PRUint32 slen)
             } else {
                 filter_len = escaped_filter.bv_len;
                 buf = escaped_filter.bv_val;
-                if (buf == NULL) {
-                    slapi_log_err(SLAPI_LOG_TRACE, "filter_stuff_func", "Attempt to copy from NULL pointer\n");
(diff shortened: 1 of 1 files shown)


progier389:
Reverting the commit, but this is strange as the change seems trivial.
I wonder if this change is not exposing another bug since the function returns -1 in some other cases ...

progier389:
Indeed the way get_filter_internal and get_filter_list handles fstr/ftmp seems a bit fishy 
I think that original change is fine but we are mishandling things in error case.

progier389:
I do not understand why filter_escape_filter_value_extended is trying to use a ftmp value (it should only access to the filter data but not to the filter string ....)

progier389:
I still propose to revert the original fix (to avoid having failing tests)
then we will need some more work to understand the reason of the heap corruption

droideck (approved):
LGTM

tbordaz (approved):
LGTM

PR #6638, opened 2025-02-27
Issue 6625 - UI - various fixes part 3

mreynolds389:
Description:

This PR addresses the following issues:

- Add new user type "traditional" for legacy objectclasses
- Display entry types in wizard header when adding new users, groups, and roles
- Update lib389 and monitor charts to use psutil instead of the UI directly calling "ps", "netstat", and "top"
- Improved format of numbers displayed in monitor: adding commas and converting bytes to more readable values
- Improved replication manager table to allow password edits

Relates: https://github.com/389ds/389-ds-base/issues/6625

### DIFF
--- a/src/cockpit/389-console/src/lib/monitor/monitorTables.jsx
+++ b/src/cockpit/389-console/src/lib/monitor/monitorTables.jsx
@@ -5,29 +5,25 @@ import {
     Grid,
     GridItem,
     Pagination,
-    PaginationVariant,
     SearchInput,
     Text,
     TextContent,
     TextVariants,
 } from '@patternfly/react-core';
 import {
-	expandable,
-	TableVariant,
-	sortable,
-	SortByDirection,
-	Table,
-	Thead,
-	Tr,
-	Th,
-	Tbody,
-	Td,
-	ExpandableRowContent,
+    SortByDirection,
+    Table,
+    Thead,
+    Tr,
+    Th,
+    Tbody,
+    Td,
+    ExpandableRowContent,
     ActionsColumn
 } from '@patternfly/react-table';
 import { ExclamationTriangleIcon } from '@patternfly/react-icons/dist/js/icons/exclamation-triangle-icon';
 import PropTypes from "prop-types";
-import { get_date_string } from "../tools.jsx";
+import { get_date_string, numToCommas } from "../tools.jsx";
 
 const _ = cockpit.gettext;
 
@@ -72,9 +68,9 @@ class AbortCleanALLRUVTable extends React.Component {
     getLog(log) {
         return (
             <TextContent>
-                <Text 
+                <Text
                     component={TextVariants.pre}
-                    style={{ 
+                    style={{
                         whiteSpace: 'pre-wrap',
                         wordBreak: 'break-word'
                     }}
@@ -88,7 +84,7 @@ class AbortCleanALLRUVTable extends React.Component {
     createRows(tasks) {
         let rows = [];
         let columns = [...this.state.columns];
-        
+
         for (const task of tasks) {
             rows.push({
                 isOpen: false,
@@ -101,12 +97,12 @@ class AbortCleanALLRUVTable extends React.Component {
                 originalData: task
             });
         }
-        
+
         if (rows.length === 0) {
             rows = [{ cells: [_("No Tasks")] }];
             columns = [{ title: _("Abort CleanAllRUV Tasks") }];
         }
-        
+
         return { rows, columns };
     }
 
@@ -189,7 +185,7 @@ class AbortCleanALLRUVTable extends React.Component {
                         <Tr>
                             {!hasNoTasks && <Th screenReaderText="Row expansion" />}
                             {columns.map((column, columnIndex) => (
-                                <Th 
+                                <Th
                                     key={columnIndex}
                                     sort={column.sortable ? {
                                         sortBy,
@@ -207,7 +203,7 @@ class AbortCleanALLRUVTable extends React.Component {
                             <React.Fragment key={rowIndex}>
                                 <Tr>
                                     {!hasNoTasks && (
-                                        <Td 
+                                        <Td
                                             expand={{
                                                 rowIndex,
                                                 isExpanded: row.isOpen,
@@ -222,7 +218,7 @@ class AbortCleanALLRUVTable extends React.Component {
                                 {row.isOpen && row.originalData && (
                                     <Tr isExpanded={true}>
                                         <Td />
-                                        <Td 
+                                        <Td
                                             colSpan={columns.length + 1}
                                             noPadding
                                         >
@@ -291,9 +287,9 @@ class CleanALLRUVTable extends React.Component {
     getLog(log) {
         return (
             <TextContent>
-                <Text 
+                <Text
                     component={TextVariants.pre}
-                    style={{ 
+                    style={{
                         whiteSpace: 'pre-wrap',
                         wordBreak: 'break-word'
                     }}
@@ -307,7 +303,7 @@ class CleanALLRUVTable extends React.Component {
     createRows(tasks) {
         let rows = [];
         let columns = [...this.state.columns];
-        
+
         for (const task of tasks) {
             rows.push({
                 isOpen: false,
@@ -320,12 +316,12 @@ class CleanALLRUVTable extends React.Component {
                 originalData: task
             });
         }
-        
+
         if (rows.length === 0) {
             rows = [{ cells: [_("No Tasks")] }];
             columns = [{ title: _("CleanAllRUV Tasks") }];
         }
-        
+
         return { rows, columns };
     }
 
@@ -408,7 +404,7 @@ class CleanALLRUVTable extends React.Component {
                         <Tr>
                             {!hasNoTasks && <Th screenReaderText="Row expansion" />}
                             {columns.map((column, columnIndex) => (
-                                <Th 
+                                <Th
                                     key={columnIndex}
                                     sort={column.sortable ? {
                                         sortBy,
@@ -426,7 +422,7 @@ class CleanALLRUVTable extends React.Component {
                             <React.Fragment key={rowIndex}>
                                 <Tr>
                                     {!hasNoTasks && (
-                                        <Td 
+                                        <Td
                                             expand={{
                                                 rowIndex,
                                                 isExpanded: row.isOpen,
@@ -441,7 +437,7 @@ class CleanALLRUVTable extends React.Component {
                                 {row.isOpen && row.originalData && (
                                     <Tr isExpanded={true}>
                                         <Td />
-                                        <Td 
+                                        <Td
                                             colSpan={columns.length + 1}
                                             noPadding
                                         >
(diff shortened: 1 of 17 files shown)


src/cockpit/389-console/src/lib/monitor/monitorTables.jsx
```diff
@@ -983,30 +979,29 @@ class ConnectionTable extends React.Component {
...
-                <GridItem span={6}><b>{parts[6] === "1" ? _("Yes") : _("No")}</b></GridItem>
-                <GridItem span={6}>{_("Number Of Times Connection Hit Max Threads:")}</GridItem>
-                <GridItem span={6}><b>{parts[7]}</b></GridItem>
-                <GridItem span={6}>{_("Number Of Operations Blocked By Max Threads:")}</GridItem>
-                <GridItem span={6}><b>{parts[8]}</b></GridItem>
+                <GridItem span={2}><b>{numToCommas(parts[4])}</b></GridItem>
+                <GridItem className="ds-margin-top-lg" span={5}>{_("Connection Currently At Max Threads:")}</GridItem>
+                <GridItem className="ds-margin-top-lg" span={7}><b>{parts[6] === "1" ? _("Yes") : _("No")}</b></GridItem>
+                <GridItem span={5}>{_("Number Of Times Connection Hit Max Threads:")}</GridItem>
+                <GridItem span={7}><b>numToCommas({parts[7]})</b></GridItem>
```
droideck: Missed curly braces around `numToCommas`

src/lib389/lib389/monitor.py
```diff
@@ -1,12 +1,13 @@
...
-# Copyright (C) 2021 Red Hat, Inc.
+# Copyright (C) 2025 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 
 import copy
+import psutil
```
droideck, line 10: `mock` build fails because of the missing dependency. So we might need to introduce it...
Probably both, - `setup.py` addition and `pythoc3-psutil` package.

src/lib389/lib389/monitor.py
```diff
@@ -86,8 +87,52 @@ def get_statistics(self):
...
+            sslport = str(self._instance.sslport)
+
+        conn_count = 0
+        conns = psutil.net_connections()
+        for conn in conns:
+            if len(conn[4]) > 0:
+                conn_port = str(conn[4][1])
+                if conn_port in (port, sslport):
+                    conn_count += 1
+                    break
```
droideck: Because of the `break`, it will always return `connection_count=1`, or do I miss something?

droideck (changes requested):

mreynolds389:
@droideck ok all those issues should be fixed now

src/lib389/lib389/monitor.py
```diff
@@ -86,8 +87,52 @@ def get_statistics(self):
...
+        p = psutil.Process(pid)
+        memory_stats = p.memory_info()
+
+        # Get memory & CPU stats
+        stats['total_mem'] = [str(total_mem)]
+        stats['rss'] = [str(memory_stats[0])]
+        stats['vms'] = [str(memory_stats[1])]
+        stats['mem_percent'] = [str(int(p.memory_percent("vms")))]  # Round up
+        stats['total_threads'] = [str(p.num_threads())]
+        stats['cpu_usage'] = [str(int(p.cpu_percent()))]  # Round up
```
droideck: `round()` would provide more accurate percentages that `int()` (as `int()` will just discard a remainder)

src/cockpit/389-console/src/lib/replication/replModals.jsx
```diff
@@ -1593,18 +1593,19 @@ export class AddManagerModal extends React.Component {
...
-            error
+            error,
+            edit,
         } = this.props;
-        let saveBtnName = _("Add Replication Manager");
+        let saveBtnName = this.props.edit ? "Save Replication Manager" : _("Add Replication Manager");
         const extraPrimaryProps = {};
         if (spinning) {
-            saveBtnName = _("Adding Replication Manager ...");
+            saveBtnName = this.props.edit ? "Saving replication Manager ..." : _("Adding Replication Manager ...");
```
droideck: `replication`, probably, should be capitalized too:)

src/cockpit/389-console/src/lib/replication/replConfig.jsx
```diff
@@ -243,18 +260,22 @@ export class ReplConfig extends React.Component {
             passwd: this.state.manager_passwd,
             addNotification: this.props.addNotification,
             msg: _("Replication Manager"),
-            success_msg: _("Successfully added Replication Manager"),
-            error_msg: _("Failure adding Replication Manager"),
+            success_msg: this.state.editManager ?
+                "Successfully edited Replication Manager" :
+                _("Successfully added Replication Manager"),
+            error_msg:this.state.editManager ?
+                "Failure editing Replication Manager" :
```
droideck, line 267: Missing internationalization tags - `_("")`

droideck (approved):
Besides a couple of minor issues, LGTM!

I'd leave the PR for @vashirov to double-check the part of the additional `psutil` dependency, though... I think, all is good, but we might need to update some other places.

src/cockpit/389-console/src/lib/replication/replModals.jsx
mreynolds389: > `replication`, probably, should be capitalized too:)

Yup...

src/cockpit/389-console/src/lib/replication/replConfig.jsx
mreynolds389: > Missing internationalization tags - `_("")`

We don't have the corresponding PO mapping, so it wouldn't do anything

mreynolds389: Leaving it out is one way to tell that it would need to be done, but we need someone who knows Japanese to do it

src/lib389/lib389/monitor.py
```diff
@@ -86,8 +87,52 @@ def get_statistics(self):
...
+        pid = self._instance.get_pid()
+        total_mem = psutil.virtual_memory()[0]
+        p = psutil.Process(pid)
+        memory_stats = p.memory_info()
+
+        # Get memory & CPU stats
+        stats['total_mem'] = [str(total_mem)]
+        stats['rss'] = [str(memory_stats[0])]
+        stats['vms'] = [str(memory_stats[1])]
+        stats['mem_percent'] = [str(round(p.memory_percent("vms")))]
```
vashirov:
```suggestion
        stats['mem_percent'] = [str(round(p.memory_percent("rss")))]
```

I have doubts about this metric. I think VSZ is a not a good indicator. I suggest to use RSS instead.
Previously we used BDB, and VSZ was about 1GB for a fresh instance. With LMDB it's slightly more than the max map size:
```
# grep MAXSIZE /var/lib/dirsrv/slapd-localhost/db/INFO.mdb
MAXSIZE=15569256448
# cat /proc/$(pidof ns-slapd)/stat | cut -d" " -f23
15950184448
```

So now it looks a bit scary:
image

And with AddressSanitizer, which allocates 20TB of virtual memory, it's even more scarier:
image

With the above change it's a bit better:

image

But the virtual memory doesn't change much over time, unlike resident, but it's hard to see this on the graph because of the difference in scale.

vashirov (commented):
I'm really happy with the change :) Especially the integration of psutil in dsconf monitor is a nice quality of life improvement.

src/lib389/lib389/monitor.py
progier389: I do not know:
 IMHO both VSS and RSS are interesting statistics:
The problem with RSS is that it may hide a slow memory leak (because unused blocks may get paged out) 
 IMHO the 707 % is a hint that lmdb map size is probably too big (and there will be some heavy swapping if the db grows around its maximum)
 
BTW libasan is really a special case and we should not worry too much about it ...

mreynolds389: Ok I've rewored the monitro page as below.  I don't have a "global" memory usage stat anymore, and I added some other satts that i can get from psutil.  See screenshot (FYI I added fake numbers to show all the stats)

monitor

mreynolds389: image

mreynolds389: Dark mode example

vashirov: I tried on this on a system with 8GB of RAM, database with 1M entries, primed:
image
Few things:
1. Virtual memory is still dominating the graph. But I think it's fine, since this is not a full blown monitoring tool, but a quick overview. Thanks for adding additional stats!
2. `dsconf localhost monitor server` shows CPU usage as 0, also visible on this graph. `cpu_percent` requires an interval: https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent I think 0.1 should be fine.
4. Time scale shows only last 10 seconds, and wraps at 99. Perhaps we should use a similar approach as Networking tab? I.e. show last 5 minutes and actual time stamps:
image

mreynolds389: @vashirov hmmm the memory graph was not supposed to grow and distort above 100% - it was supposed to be capped.  I must have broken it when working on something else.  I should be able to add the time interval string on the axis.  I'll work on this and let you guys know when its ready.

mreynolds389: @vashirov OK all changes have been made, please test and review one more time

vashirov: Now I see Virtual Memory is capped at 100%:
image

But it shows 715% on the left, while Resident and Swap sizes are the same as on the graph, which is a bit confusing. I propose to drop percentages in the left column, as with the Virtual memory it can be misleading. 
Not sure what to do about the graph though. For me it would be useful to see how RSS and swap usage changes over time compared to the total available memory, not to the VSZ of the process. @progier389, WDYT?

Minor issue: time scale has numbers first, then they get replaced with timestamps.

mreynolds389: > Minor issue: time scale has numbers first, then they get replaced with timestamps.

You can't have the same x value, so each interval must be different (can't use "0:00:00".  Can't use a time, because it simply is not recorded yet.  I thought the countdown was cool, but I will remove it.  I can just let it build up gradually (the intervals are cumulative), but there is not going to be a perfect solution here.

As for capping memory inside the chart at 100% I could expand the chart to allow more than 100% (like I do with CPU), but I feel it's still going to look odd in your test case, but lets try it. :-)

vashirov: @mreynolds389 I apologize for steering this review into bikeshedding territory :) I think we can merge it as it is, and then address the graph separately.

vashirov (approved):

src/lib389/lib389/monitor.py
mreynolds389: @vashirov ok just pushed another change, thoughts?

vashirov: Looks great on both small RAM (2GB) and large RAM (32GB) VMs, thanks! :)

Issue #6639, opened 2025-02-27
remove code related to entryrdn_get_switch()


mreynolds389:
**Issue Description**

There is no real need for switching between entryrdn and entrydn.  We can clean up a lot of code by removing all the code related to  entryrdn_get_switch().

mreynolds389:
Hmm this was definitely tested, but I'll check it out

mreynolds389:
be639119e..778361f67  389-ds-base-3.0 -> 389-ds-base-3.0

mreynolds389:
> Commit [ed386e1](https://github.com/389ds/389-ds-base/commit/ed386e16d283a73588aa7fdb1988ca61b6d9ef62) cause a regression cf [#6671](https://github.com/389ds/389-ds-base/issues/6671)

Investigating...

PR #6640, opened 2025-02-27
Issue 6639 - remove all the code related to entryrdn_get_switch

mreynolds389:
Description:

There is no real benefit to switching between entryrdn and entrydn formats so remove all the code related to this.

Relates: https://github.com/389ds/389-ds-base/issues/6639

### DIFF
--- a/ldap/ldif/template-dse-minimal.ldif.in
+++ b/ldap/ldif/template-dse-minimal.ldif.in
@@ -626,7 +626,6 @@ objectclass: extensibleObject
 cn: config
 nsslapd-mode: 600
 nsslapd-directory: %db_dir%
-nsslapd-subtree-rename-switch: on
 
 dn: cn=monitor, cn=ldbm database, cn=plugins, cn=config
 objectclass: top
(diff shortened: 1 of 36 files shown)


droideck (approved):
Wow, that's quite a PR. Everything looks good, though! Ack

jchapma:
This PR builds and tests fine for me. 

To help my understanding in this area, when we get the entry id for a particular DN we only use the entryrdn index and not the entrydn index. Why is this, is the entrydn index a legacy thing that we dont need anymore because we can easily construct the full DN from the entryrdn index using parent/child keys.

mreynolds389:
One possible addition we could do here is in the upgrade code we could remove the old config attribute if present.  Right now it is silently ignored.

jchapma:
> One possible addition we could do here is in the upgrade code we could remove the old config attribute if present. Right now it is silently ignored.

Makes sense to me

mreynolds389:
> > One possible addition we could do here is in the upgrade code we could remove the old config attribute if present. Right now it is silently ignored.
> 
> Makes sense to me

Added

PR #6643, opened 2025-02-28
Issue 6641 - modrdn fails when a user is member of multiple groups

### LINKED ISSUE #6641: modrdn fails when a user is member of multiple groups
**Issue Description**
Rename of a user that is member of multiple AM groups fail when MO and RI plugins are enabled.

**Steps to Reproduce**
1. Enable AM plugin
2. Create definition and regex rule for users with A in the name
3. Create definition and regex rule for users with Z in the name
4. Enable MO plugin
5. Enable RI plugin
6. Create groups and users (userwitha, userwithz, userwithaz)
7. Rename userwithaz

**Expected results**
Group member attributes should be updated with new rdn

**Actual results**
Type or value exists error generated, group member rename fails

jchapma:
Bug description:
Rename of a user that is member of multiple AM groups fail when MO and RI plugins are enabled.

Fix description:
MO plugin - After updating the entry member attribute, check the return value. Retry the delete if the attr value exists and retry the add if the attr value is missing.

RI plugin - A previous commit checked if the attr value was not present before adding a mod. This commit was reverted in favour of overriding the internal op return value, consistent with other plugins.

CI test from Viktor Ashirov <vashirov@redhat.com>

Fixes: https://github.com/389ds/389-ds-base/issues/6641
Relates: https://github.com/389ds/389-ds-base/issues/6566

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/plugins/modrdn_test.py
+++ b/dirsrvtests/tests/suites/plugins/modrdn_test.py
@@ -123,8 +123,8 @@ def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
 
     # Create groups
     groups = Groups(inst, DEFAULT_SUFFIX)
-    groups.create(properties={"cn": "userswithA"})
-    groups.create(properties={"cn": "userswithZ"})
+    groupA = groups.create(properties={"cn": "userswithA"})
+    groupZ = groups.create(properties={"cn": "userswithZ"})
 
     # Create users
     users = nsUserAccounts(inst, DEFAULT_SUFFIX)
@@ -156,7 +156,18 @@ def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
         }
     )
     user = users.create(properties=user_props)
+    user_orig_dn = user.dn
 
     # Rename userwithaz
     user.rename(new_rdn="uid=userwith")
+    user_new_dn = user.dn
+
     assert user.get_attr_val_utf8("uid") != "userwithaz"
+
+    # Check groups contain renamed username
+    assert groupA.is_member(user_new_dn)
+    assert groupZ.is_member(user_new_dn)
+
+    # Check groups dont contain original username
+    assert not groupA.is_member(user_orig_dn)
+    assert not groupZ.is_member(user_orig_dn)
--- a/dirsrvtests/tests/suites/plugins/modrdn_test.py
+++ b/dirsrvtests/tests/suites/plugins/modrdn_test.py
@@ -1,0 +1,162 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import pytest
+from lib389.topologies import topology_st
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.idm.group import Groups
+from lib389.idm.user import nsUserAccounts
+from lib389.plugins import (
+    AutoMembershipDefinitions,
+    AutoMembershipPlugin,
+    AutoMembershipRegexRules,
+    MemberOfPlugin,
+    ReferentialIntegrityPlugin,
+)
+
+pytestmark = pytest.mark.tier1
+
+USER_PROPERTIES = {
+    "uid": "userwith",
+    "cn": "userwith",
+    "uidNumber": "1000",
+    "gidNumber": "2000",
+    "homeDirectory": "/home/testuser",
+    "displayName": "test user",
+}
+
+
+def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
+    """Test that a member of 2 automember groups can be renamed
+    :id: 0e40bdc4-a2d2-4bb8-8368-e02c8920bad2
+
+    :setup: Standalone instance
+
+    :steps:
+        1. Enable automember plugin
+        2. Create definiton for users with A in the name
+        3. Create regex rule for users with A in the name
+        4. Create definiton for users with Z in the name
+        5. Create regex rule for users with Z in the name
+        6. Enable memberof plugin
+        7. Enable referential integrity plugin
+        8. Restart the instance
+        9. Create groups
+        10. Create users userwitha, userwithz, userwithaz
+        11. Rename userwithaz
+
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Success
+        7. Success
+        8. Success
+        9. Success
+        10. Success
+        11. Success
+    """
+    inst = topology_st.standalone
+
+    # Enable automember plugin
+    automember_plugin = AutoMembershipPlugin(inst)
+    automember_plugin.enable()
+
+    # Create definiton for users with A in the name
+    automembers = AutoMembershipDefinitions(inst)
+    automember = automembers.create(
+        properties={
+            "cn": "userswithA",
+            "autoMemberScope": DEFAULT_SUFFIX,
+            "autoMemberFilter": "objectclass=posixAccount",
+            "autoMemberGroupingAttr": "member:dn",
+        }
+    )
+
+    # Create regex rule for users with A in the name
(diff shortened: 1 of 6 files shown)


ldap/servers/plugins/referint/referint.c
```diff
@@ -726,6 +726,22 @@ _do_modify(Slapi_PBlock *mod_pb, Slapi_DN *entrySDN, LDAPMod **mods)
     slapi_modify_internal_pb(mod_pb);
     slapi_pblock_get(mod_pb, SLAPI_PLUGIN_INTOP_RESULT, &rc);
 
+    /* Do we need to override the return value */
+    if (rc) {
+        for (size_t i = 0; mods && mods[i] != NULL; i++) {
+            int mod_type = mods[i]->mod_op & LDAP_MOD_OP;
+            if (mod_type == LDAP_MOD_DELETE) {
+                if (rc == LDAP_NO_SUCH_ATTRIBUTE) {
+                    rc = LDAP_SUCCESS;
```
tbordaz: This is changing the 'rc' value, so why not breaking from the loop ?
I have a mix feeling about the loop. If there is
MOD_DEL -> LDAP_SUCCESS
MOD_DEL -> NO_SUCH_ATTR
MOD_ADD-> OPERATION_ERROR
It will hide the OPERATION_ERROR.

dirsrvtests/tests/suites/plugins/modrdn_test.py
```diff
@@ -0,0 +1,162 @@
...
+        {
+            "uid": USER_PROPERTIES["uid"] + "az",
+            "cn": USER_PROPERTIES["cn"] + "az",
+        }
+    )
+    user = users.create(properties=user_props)
+
+    # Rename userwithaz
+    user.rename(new_rdn="uid=userwith")
+    assert user.get_attr_val_utf8("uid") != "userwithaz"
```
tbordaz, line 166: May be good to also verify that the renamed entry is in the groups userswithA and userswithB. Also checking the for former dn (before rename) is no longer in the groups.

tbordaz: In addition, I suggest to run IDM CI tests. freeipa is relying over MO and automember.

ldap/servers/plugins/referint/referint.c
jchapma: This function can be invoked for either "delete" mode where there is a single MOD_DEL, or "modrdn" mode with MOD_DEL and MOD_ADD operations. So IIUC, i don't think the scenario you described can happen. 

Regarding why I didn't break out of the loop:
In "delete" mode a single MOD_DEL is used and the loop will exit anyway. 
In "modrdn" mode where there are 2 mods, MOD_DEL and MOD_ADD, in a worse case scenario where both mods fail the pblock intop result will be set to the MOD_ADD error code. When this is detected, the loop will be on its final iteration. 

But I can add breaks if you prefer.

progier389: Anyway the loop logic is broken:
   if the internal modify fails no change is done and you will return success without adding/removing the other values that must be added/deleted.
   
When we look at _do_modify I see it ts called in two ways:
   [1] with a single modifier (in that case, we should use a _do_single_modify function that calls _do_modify and returns LDAP_SUCCES if return code is LDAP_NO_SUCH_ATTRIBUTE or  LDAP_TYPE_OR_VALUE_EXISTS
   [2] with a list of modifier. The fact that the value is already present/deleted is already tested when building the modifier list means that there is nothing else to do. just go on to call do_modifier in that case.

jchapma: > Anyway the loop logic is broken:
> if the internal modify fails no change is done and you will return success without adding/removing the other values that must be added/deleted.

In the context of updating group member attribute values that can also be updated by AM and MO plugins, it is my understanding that our goal here is to gracefully handle a failed modify with these mod op -> error code characteristics, any other mod op error should be returned to the caller:
MOD_DEL -> LDAP_NO_SUCH_ATTRIBUTE
MOD_ADD -> LDAP_TYPE_OR_VALUE_EXISTS

progier389: I think that the term "gracefully handle" is a bit unclear:
  - it means the the plugin should not fail and abort the operation in that case
    and I guess we agree on that point because your commit achieve that goal.
  - but it should not either left thing in inconsistent state and aborts all the plugin changes because one of them is failing.

progier389: BTW: I think we have the same problem with the memberof_add_memberof_attr change.

I wonder if the easiest solution is not to perform the internal modify as it is today 
 and if it fails with LDAP_NO_SUCH_ATTRIBUTE/LDAP_TYPE_OR_VALUE_EXISTS:
   if there is only 1 modifier:  do return success
   otherwise performs one internal modify operation  per modifier ignoring   LDAP_NO_SUCH_ATTRIBUTE/LDAP_TYPE_OR_VALUE_EXISTS error. 
   
   Anyway the error case is quite rare so the perf impact should stay reasonable,
     and the change in both plugins could be simple if a new slapi_special_internal_modify operation is added in libslapd

progier389: Another way it to add an new operation flags and test it in modify operation to ignores the errors ...
(That will probably to more efficient)

tbordaz: I initially had a performance concern with the idea of one internal_modify by  MOD (e.g. MOD_DEL, MOD_ADD). At a second though I agree with @progier389 that the performance hit is rare and limited. MODRDN is not that frequent and limited because the number of groups that contain a given entry is usually limited. So having two MOD rather than one looks okay.
If MOD_DEL fails because of LDAP_NO_SUCH_ATTRIBUTE, just ignore the error, if MOD_ADD fails because of LDAP_TYPE_OR_VALUE_EXISTS just ignore the error. This is more expensive but makes the code easier to maintain.

dirsrvtests/tests/suites/plugins/modrdn_test.py
```diff
@@ -0,0 +1,173 @@
...
+    "uidNumber": "1000",
+    "gidNumber": "2000",
+    "homeDirectory": "/home/testuser",
+    "displayName": "test user",
+}
+
+
+def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
+    """Test that a member of 2 automember groups can be renamed
+    :id: 0e40bdc4-a2d2-4bb8-8368-e02c8920bad2
```
vashirov:
```suggestion
    """Test that a member of 2 automember groups can be renamed
    
    :id: 0e40bdc4-a2d2-4bb8-8368-e02c8920bad2
```
Small change to make test metadata validator happy.

jchapma:
I made a mistake with the last commit, let me fix it and re push.

jchapma:
rebase

tbordaz (commented):
Overall patch looks good to me. Did you get a chance to run IDM CI ?

jchapma:
> Overall patch looks good to me. Did you get a chance to run IDM CI ?

Hi @tbordaz I did not, let me give it a go. Thanks

jchapma:
The last commit fixes the issue causing the CI test to fail in deferred memberof

progier389 (approved):
LGTM but lets wait for IPA tests result before merging this one

jchapma:
> LGTM but lets wait for IPA tests result before merging this one

I have run this PR on the IPA CI framework, all tests pass: https://github.com/freeipa-pr-ci2/freeipa/pull/4315

But, when I created a new PR with more relevant tests there are timeout issues: https://github.com/freeipa-pr-ci2/freeipa/pull/4316

I wonder is the performance impact of breaking a modrdn operation into 2 separate operations, a good idea. WDYT ?

jchapma:
rebase

tbordaz:
If performance is an issue, an option would be (in memberof) to read the attribute values from the target entry and check if it requires a MOD_DEL/MOD_ADD to do a single modify.

tbordaz (approved):
Indeed sidgen plugin brings a sidgen task that generate sid for accounts not having a 'ipantsecurityidentifier' value. As the test is runnning fine on your VM and fails because of timeout on IDM CI it looks fine to me to push your fix.
Thanks for all your investigations

PR #6654, opened 2025-03-05
Issue 6090 - dbscan: use bdb by default

vashirov:
Bug Description:
dbscan started to use mdb by default on versions where it's not the default.

Fix Description:
Use bdb by default on 2.x versions.

Relates: https://github.com/389ds/389-ds-base/issues/6090


vashirov:
This also should go in 2.5 branch.

mreynolds389 (approved):

Issue #6655, opened 2025-03-06
Replication release replica decoding error - Unable to parse the response


mreynolds389:
**Issue Description**

#### Supplier 1

```
[05/Mar/2025:21:26:45.697952963 -0500] - DEBUG - NSMMReplicationPlugin - conn_read_result_ex - agmt="cn=002" (fedora:39002): ldap_parse_extended_result did not return data - errmsg (unsupported extended operation) error 2
[05/Mar/2025:21:26:45.706707732 -0500] - ERR - decode_repl_ext_response - decoding failed: response_code (Ok) ruv_bervals (Ok) data_guid (Ok) data (Ok) bvdata (No data)
[05/Mar/2025:21:26:45.709707733 -0500] - ERR - NSMMReplicationPlugin - release_replica - agmt="cn=002" (localhost:389): Unable to parse the response  to the endReplication extended operation.
```

#### Supplier 2

```
[05/Mar/2025:21:26:45.651516142 -0500] conn=7 op=19 EXT oid="2.16.840.1.113730.3.5.12" name="replication-multisupplier-extop" --> START SESSION
[05/Mar/2025:21:26:45.666796859 -0500] conn=7 op=20 EXT oid="2.16.840.1.113730.3.5.5" name="replication-multisupplier-extop" --> END SESSION
[05/Mar/2025:21:26:45.694423788 -0500] conn=7 op=20 RESULT err=2 tag=120 nentries=0 wtime=0.000123842 optime=0.027627701 etime=0.027750453 --> END RESULT
[05/Mar/2025:21:26:45.664625578 -0500] conn=7 op=19 RESULT err=0 tag=120 nentries=0 wtime=0.000175835 optime=0.013111958 etime=0.013287409 --> START END
```

When trying to end a session, if that replica is already acquired (another replica already has exclusive access) we return error 2 (protocol error).  This leads to "no data" in the response and we get the decoding error.

So there is some race condition where we are not waiting for the response of the START SESSION before trying to END SESSION.

mreynolds389:
More info, so when a replica receives an "end" request we check for exclusive access in repl_extop.c:1294 in multisupplier_extop_EndNSDS50ReplicationRequest()

      connext = consumer_connection_extension_acquire_exclusive_access(conn, connid, opid);

This returns NULL, and then we return err=2 (or SLAPI_PLUGIN_EXTENDED_NOT_HANDLED) which is seen as "unsupported extended operation" and leads to the decoding error.

[...]

PR #6658, opened 2025-03-06
Issue 6655 - fix replication release replica decoding error

mreynolds389:
Description:

When a start replication session extended op is received acquire and release exclusive access before returning the result to the client. Otherwise there is a race condition where a "end" replication extended op can arrive before the replica is released and that leads to a decoding error on the other replica.

Relates: https://github.com/389ds/389-ds-base/issues/6655

### DIFF
--- a/dirsrvtests/tests/suites/replication/acceptance_test.py
+++ b/dirsrvtests/tests/suites/replication/acceptance_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2017 Red Hat, Inc.
+# Copyright (C) 2025 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
@@ -452,6 +452,13 @@ def test_multi_subsuffix_replication(topo_m4):
                             f"User {user_dn} on supplier {user_obj._instance.serverid} "
                             f"still has 'Description {j}'"
                         )
+
+        # Check there are no decoding errors
+        assert not topo_m4.ms["supplier1"].ds_error_log.match('.*decoding failed.*')
+        assert not topo_m4.ms["supplier2"].ds_error_log.match('.*decoding failed.*')
+        assert not topo_m4.ms["supplier3"].ds_error_log.match('.*decoding failed.*')
+        assert not topo_m4.ms["supplier4"].ds_error_log.match('.*decoding failed.*')
+
     finally:
         for suffix, test_users in test_users_by_suffix.items():
             for user in test_users:
@@ -505,6 +512,7 @@ def test_new_suffix(topo_m4, new_suffix):
 
     repl.remove_supplier(m1)
     repl.remove_supplier(m2)
+
 
 def test_many_attrs(topo_m4, create_entry):
     """Check a replication with many attributes (add and delete)
--- a/ldap/servers/plugins/replication/repl_extop.c
+++ b/ldap/servers/plugins/replication/repl_extop.c
@@ -1124,6 +1124,12 @@ send_response:
         slapi_pblock_set(pb, SLAPI_EXT_OP_RET_OID, REPL_NSDS50_REPLICATION_RESPONSE_OID);
     }
 
+    /* connext (release our hold on it at least) */
+    if (NULL != connext) {
+        /* don't free it, just let go of it */
+        consumer_connection_extension_relinquish_exclusive_access(conn, connid, opid, PR_FALSE);
+    }
+
     slapi_pblock_set(pb, SLAPI_EXT_OP_RET_VALUE, resp_bval);
     slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name,
                   "multisupplier_extop_StartNSDS50ReplicationRequest - "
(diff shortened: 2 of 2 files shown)


dirsrvtests/tests/suites/replication/acceptance_test.py
```diff
@@ -704,6 +704,25 @@ def test_warning_for_invalid_replica(topo_m4):
     assert topo_m4.ms["supplier1"].ds_error_log.match('.*nsds5ReplicaBackoffMax.*10.*invalid.*')
 
 
+def test_decoding_errors(topo_m4):
+    """Check that there are no descoding errors in the logs
```
droideck:
```suggestion
    """Check that there are no decoding errors in the logs
```

dirsrvtests/tests/suites/replication/acceptance_test.py
```diff
@@ -704,6 +704,25 @@ def test_warning_for_invalid_replica(topo_m4):
...
+
+    This test should not be run by itself, but with all the previous tests
```
droideck: It sounds a bit troubling as test ordering dependencies can be a bit brittle. 

I never used it myself but maybe we can use something like this?
https://pytest-dependency.readthedocs.io/en/stable/usage.html
or
https://pytest-order.readthedocs.io/en/stable/usage.html

@vashirov or @bsimonova might have better idea...

droideck (approved):
C part looks good to me!

dirsrvtests/tests/suites/replication/acceptance_test.py
mreynolds389: It's not really the ordering that matters, but the issue would only show itself after running the entire test suite.  It's also not always 100% reproducible so that makes this tricky.  I almost didn't even do a test case, but I figured you'd want one :-)

vashirov: @droideck I didn't know about these options, nice find! 
@mreynolds389 I can reliably reproduce the issue by running `dirsrvtests/tests/suites/replication/acceptance_test.py::test_multi_subsuffix_replication` alone, on both small and large VM instances. Perhaps we can move the check there?

ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1124,6 +1124,12 @@ multisupplier_extop_StartNSDS50ReplicationRequest(Slapi_PBlock *pb)
         slapi_pblock_set(pb, SLAPI_EXT_OP_RET_OID, REPL_NSDS50_REPLICATION_RESPONSE_OID);
     }
 
+    /* connext (release our hold on it at least) */
```
tbordaz, line 1127: Sorry I do not understand why doing this test/relinquish few lines before could prevent to send the message on a closed session. Does it prevent the message on the acceptance_test ?

do you think that  reread connext, (with repl_con_get_ext(REPL_CON_EXT_CONN, conn)) before sending the relinquish, could help ?.

tbordaz (commented):
The fix is in multisupplier_extop_StartNSDS50ReplicationRequest, should not it also be in multisupplier_extop_EndNSDS50ReplicationRequest ?

ldap/servers/plugins/replication/repl_extop.c
mreynolds389: I just moved the "release exclusive access" to above the send_ldap_result() which prevents the race condition.  I'm not sure what you mean by re-read the connext, or how that would fix this issue.

I'll check the "end" replication code to see if it's still applicable there.

mreynolds389: Ok made some changes, but I'll try to explain how the fix works...

Previously we did not release exclusive access until after we sent the result.  This allowed a replica to try and end a session, but when it tries to end the session it sees that exclusive access is already taken.  This results in an extended op result that does not contain any data/payload, which in turn causes a decoding error on the other replica as it expected a "successful" extended op to have a payload.  The real issue here is that we should have released exclusive access before sending the result otherwise there is this tiny window where this problem can occur.  I hope that helps explain things better.

tbordaz: My first comment was wrong as I believed the error can from invalid value of connext. Just ignore it.

Following your explanations, I understand that the problem is that the consumer sends an empty extop result (no payload). Am I correct ? Is the payload 'resp_bval' ?

mreynolds389: When a replica is already "exclusively acquired" we do not return a payload in multisupplier_extop_EndNSDS50ReplicationRequest() when consumer_connection_extension_acquire_exclusive_access() returns NULL (when it's already acquired).

This is fine, this is not the problem but the symptom.  The real problem is that it is already acquired but it is in the process of being released.  It should have been released in multisupplier_extop_StartNSDS50ReplicationRequest() _before_ telling the client it was done.  This is the issue, and the fix is simple, just release it before telling the client all is ok to proceed.  Otherwise the client/replica can send the next extended op before the "exclusive release" in the original "start" finishes.  It's a small window where this can happen, but it seems to happen a lot.

tbordaz: ack thanks for the explanation

tbordaz (approved):

ldap/servers/plugins/replication/repl_extop.c
mreynolds389: Now I also wonder why replication is sending start/end ops back to back like that, but I think that is a different issue.

Issue #6653, opened 2025-03-05
Cleanup error messages


mreynolds389:
**Issue Description**

There are some error messages that are unnecessarily alarming.  We should adjust severity level, or move them to a different log level

```
[05/Mar/2025:15:27:16.883240267 -0500] - ERR - attrcrypt_cipher_init - No symmetric key found for cipher AES in backend userroot, attempting to create one...
[05/Mar/2025:15:27:16.905173224 -0500] - ERR - attrcrypt_cipher_init - No symmetric key found for cipher 3DES in backend userroot, attempting to create one...

[05/Mar/2025:15:27:16.905173224 -0500] - ERR - bdb_version_write - Could not open file "/dev/shm/slapd-NWRA-COM/DBVERSION" for writing Netscape Portable Runtime -5950 (File not found.)
```

 Originally reported in: https://github.com/389ds/389-ds-base/issues/6637

mreynolds389:
e1f23ff33..c3255ef28  389-ds-base-3.0 -> 389-ds-base-3.0
20ee732e6..653cef687  389-ds-base-2.6 -> 389-ds-base-2.6
8af3fd328..48983fe08  389-ds-base-2.5 -> 389-ds-base-2.5
19a275763..b66b2c9df  389-ds-base-2.4 -> 389-ds-base-2.4

PR #6659, opened 2025-03-06
Issue  6653 - Cleanup error messages

mreynolds389:
Description:

Change the attr encryption symmetric key error message to use a notice severity level instead of a full blown error.

Do not log by default an error when the DBVERSION file is missing as this is expected after a system reboot.

And fix compiler warning in ssl.c

Relates: https://github.com/389ds/389-ds-base/issues/6653

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_version.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_version.c
@@ -53,7 +53,12 @@ bdb_version_write(struct ldbminfo *li, const char *directory, const char *datave
     /* Open the file */
     if ((prfd = PR_Open(filename, PR_RDWR | PR_CREATE_FILE | PR_TRUNCATE,
                         SLAPD_DEFAULT_FILE_MODE)) == NULL) {
-        slapi_log_err(SLAPI_LOG_ERR, "bdb_version_write",
+        int sev_level = SLAPI_LOG_ERR;
+        if (PR_GetError() == -5950 /* not found */ && strstr("/dev/shm/", filename)) {
+            /* After a server reboot it is expected for this file to not be present */
+            sev_level = SLAPI_LOG_DEBUG;
+        }
+        slapi_log_err(sev_level, "bdb_version_write",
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/back-ldbm/db-bdb/bdb_version.c
```diff
@@ -53,7 +53,12 @@ bdb_version_write(struct ldbminfo *li, const char *directory, const char *datave
     /* Open the file */
     if ((prfd = PR_Open(filename, PR_RDWR | PR_CREATE_FILE | PR_TRUNCATE,
                         SLAPD_DEFAULT_FILE_MODE)) == NULL) {
-        slapi_log_err(SLAPI_LOG_ERR, "bdb_version_write",
+        int sev_level = SLAPI_LOG_ERR;
+        if (PR_GetError() == -5950 /* not found */ && strstr("/dev/shm/", filename)) {
+            /* After a server reboot it is expected for this file to not be present */
+            sev_level = SLAPI_LOG_DEBUG;
```
tbordaz, line 59: As it is a normal event, we can also log it as SLAPI_LOG_INFO and with a less alarming message like "/dev/shm/slapd-foo not present"

tbordaz (approved):
LGTM

ldap/servers/slapd/back-ldbm/db-bdb/bdb_version.c
mreynolds389: > As it is a normal event, we can also log it as SLAPI_LOG_INFO and with a less alarming message like "/dev/shm/slapd-foo not present"

So don't log it as a DEBUG message (not logged by default) to logging it every time?  My idea was to not log it by default as it's just more noise that people complain about.

tbordaz: Oopsss my bad. I did not realize that. I agree it is better to not log it (unless debug) as it is expected behavior on reboot. Sorry for the noise.

mreynolds389: No worries, thanks for the review!

PR #6661, opened 2025-03-07
Issue 6660 - UI - Replication Monitoring Lag Report Feature

### LINKED ISSUE #6660: UI - Replication Monitoring Lag Report Feature
**Description**
Create a Web UI tab in the Monitoring - Replication tab to generate and visualize replication lag reports. It should include interactive charts, metrics, and downloadable reports.

It also should include all existing features from https://github.com/389ds/389-ds-base/issues/6465 

Base implementation was adapted from: https://github.com/droideck/ansible-ds389-repl-monitoring/tree/main/roles/ds389_repl_monitoring/library

droideck:
Description: Add a new "Log Analysis" tab in the Replication Monitoring section of the 389 DS Cockpit interface to visualize and analyze replication performance.

Implement a UI for configuring replication log analysis parameters. Add file browser for selecting log directories and output locations. Create interactive PatternFly charts to visualize replication lag metrics. Support multiple output formats (JSON, HTML, PNG, CSV). Include filtering options for troubleshooting (time range, lag thresholds, etc.) Add ability to download and view generated reports.

Fixes: https://github.com/389ds/389-ds-base/issues/6660

Reviewed by: ?

### DIFF
--- a/src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
+++ b/src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
@@ -79,7 +79,7 @@ export class ReplLogAnalysis extends React.Component {
             // Form inputs
             logDirs: "",
             logDirsList: [],
-            suffixesList: props.suffixDN ? [props.suffixDN] : [],
+            suffixesList: [],
             currentSuffixInput: "",
             anonymizeOption: false,
             replicationFilter: "all", // 'all', 'only-fully', 'only-not'
--- a/src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
+++ b/src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
@@ -8,13 +8,8 @@ import {
     Card,
     Checkbox,
     DatePicker,
-    Flex,
-    FlexItem,
     Form,
     FormGroup,
-    FormHelperText,
-    FormSelect,
-    FormSelectOption,
     Grid,
     GridItem,
     HelperText,
@@ -26,8 +21,6 @@ import {
     Radio,
     Split,
     SplitItem,
-    Stack,
-    StackItem,
     Text,
     TextContent,
     TextInput,
@@ -57,7 +50,6 @@ import {
     PlusCircleIcon,
     TrashIcon,
     FolderCloseIcon,
-    HistoryIcon
 } from "@patternfly/react-icons";
 import { ReportModal } from "./monitorModals.jsx";
 
@@ -121,6 +113,7 @@ export class ReplLogAnalysis extends React.Component {
             reportError: null,
             startDate: null,
             endDate: null,
+            isReplReportsPackageInstalled: false
         };
 
         // Bind methods
@@ -172,6 +165,8 @@ export class ReplLogAnalysis extends React.Component {
 
     componentDidMount() {
         this.props.enableTree();
+        this.validateDirectoryExists("/var/log/dirsrv");
+        this.checkReplReportsPackage();
     }
 
     validateDirectoryExists(path) {
@@ -186,6 +181,27 @@ export class ReplLogAnalysis extends React.Component {
                 .then(() => resolve(true))
                 .catch(() => resolve(false));
         });
+    }
+
+    checkReplReportsPackage() {
+        // Check if python3-lib389-repl-reports package is installed
+        cockpit.spawn(["rpm", "-q", "python3-lib389-repl-reports"], { err: "ignore" })
+            .then(() => {
+                // Package is installed
+                this.setState({ isReplReportsPackageInstalled: true });
+            })
+            .catch(() => {
+                // Package is not installed
+                this.setState(prevState => ({
+                    isReplReportsPackageInstalled: false,
+                    // Disable HTML and PNG format options if package is not installed
+                    formatOptions: {
+                        ...prevState.formatOptions,
+                        html: false,
+                        png: false
+                    }
+                }));
+            });
     }
 
     handleInputChange(event, value) {
@@ -230,7 +246,12 @@ export class ReplLogAnalysis extends React.Component {
     }
 
     handleFormatOptionChange(format) {
-        // Allow all format options to be toggled
+        // For HTML and PNG formats, check if the package is installed
+        if ((format === 'html' || format === 'png') && !this.state.isReplReportsPackageInstalled) {
+            // Don't allow toggling these formats if the package is not installed
+            return;
+        }
+        
         this.setState(prevState => ({
             formatOptions: {
                 ...prevState.formatOptions,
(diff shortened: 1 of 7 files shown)


src/cockpit/389-console/src/monitor.jsx
```diff
@@ -11,6 +11,7 @@ import AuditLogMonitor from "./lib/monitor/auditlog.jsx";
 import AuditFailLogMonitor from "./lib/monitor/auditfaillog.jsx";
 import ErrorLogMonitor from "./lib/monitor/errorlog.jsx";
 import SecurityLogMonitor from "./lib/monitor/securitylog.jsx";
+import ReplLogAnalysis from "./lib/monitor/replLogAnalysis.jsx";
```
vashirov, line 14: Is it possible to import this only when repl-tools are available on the system? For example, on RHEL we don't have the required dependencies for repl-tools, so this tab would be not fully functional and misleading to the users.

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
```diff
@@ -0,0 +1,1912 @@
...
+                                        <GridItem span={8}>
+                                            <Split hasGutter>
+                                                <SplitItem isFilled>
+                                                    <TextInput
+                                                        id="currentSuffixInput"
+                                                        value={currentSuffixInput}
+                                                        type="text"
+                                                        onChange={this.handleSuffixInputChange}
+                                                        onKeyDown={this.handleSuffixKeyPress}
+                                                        placeholder={_("Enter a suffix (e.g., dc=example,dc=com)")}
```
vashirov: Would it be possible to prepopulate from the existing suffixes on the server? Or it can be used with logs from any other servers?
Also, it allows me to enter any text as the suffix.

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
```diff
@@ -0,0 +1,1912 @@
...
+                html: true,       // HTML format - default to true
+                png: true,        // PNG format - default to true
+                csv: true         // CSV format - default to true
+            },
+            useCustomOutputDir: false,
+            customOutputDir: "",
+
+            // File browser state
+            showFileBrowser: false,
+            currentPath: "/var/log/dirsrv",
```
vashirov, line 102: When I go to some other directory, but then press Cancel or close the dialog, next time I open the dialog, it still shows the last directory. I think it should reset to the default location.

src/cockpit/389-console/src/monitor.jsx
droideck: Good point! I'll disable HTML and PNG options in case the package is missing. Patternfly Charts and CSV should work without it.

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
droideck: It's actually by design. I imagine it can be convenient to have a "working directory" while the report is being configured. For example, someone might choose a directory but then misclicked or decided to include another directory and it's convenient if they continue from the same place.

It is reset to the default location if the tab was changed (and then returned back to Log Analysis), though.

droideck:
Fixed.
I've also started to add an Existing Report Table/Modal so it's possible to load older reports.

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
vashirov: I agree that the directory should stay, when there was an affirmative action like Add Current Directory or Add Selected. But if I press Cancel or close the dialog, I expect the form to reset to the previous known good state. Maybe this recording will be more helpful:
[dialog.webm](https://github.com/user-attachments/assets/b83d559d-88ac-45d4-b402-19196c62afaf)

droideck: Actually, I've played around more with the feature, and I fully agree with your original statement.
I think we should always reset to the default location.

Even if you've already chosen some directories, we might want to add more directories from another place. And if we need to navigate back to the already chosen location, we can just copy-paste the path in the text field and hit enter.

So many cases can be, and as of now, I think the most natural solution is to reset to the default. I'm open to change, though.

droideck:
The functionality for Table/Modal for reopening old reports is complete.
And I fixed other reported issues.

Please, review.

droideck:
> @droideck - shouldn't "Logs directory" default to the servers' log directory? It's not clear what the directory should be and what should be in it.

"Log Directories" are server directories. So when we choose the set of directories to process, I default to "/var/log/dirsrv". But the report can be run on any server so we can, say, be on `standalone1` instance but run the report on `supplier1`, `supplier2` and `consumer1` log directories.

I added a short note about the report logic (and more tooltips throughout the whole UI report, so it should be easier to use now). 

> "Server Anonymization" & "Replication Status Filter" sections should have more vertical separation - its very crowded and runs together

It's done. However, I also reworked the UI and made it into a single 'grid-column.'
 
> The Report format area has a redundant "Report Format ... Report Format [check boxes]" I don't see a need for the second Report Format on the same line. It should be the label and then just the checkboxes

I renamed them. So it should be better now (I still want to keep the section that unifies all Output options)

> Lastly I think we need some tiptools or helper documentation to describe how to use the page. For example, It's confusing what the UTC offset is for and how it works. What should be in the "Logs directory". Etc, etc. Yes this can be documented in the admin guide, but I think the UI should be more intuitive/helpful to an Admin. To be honest most people don't read the docs anyway, so the UI should be as easy to use/understand as possible.

Done.

droideck:
I did a few more minor fixes.
Please review.

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
```diff
@@ -0,0 +1,2140 @@
...
+                                                            label={_("Show actual server names")}
+                                                            isChecked={!anonymizeOption}
+                                                            onChange={() => this.handleAnonymizeChange(false)}
+                                                        />
+                                                    </FormGroup>
+
+                                                    <FormGroup
+                                                        label={_("Replication Status Filter")}
+                                                        fieldId="replication-filter"
+                                                        className="ds-margin-top"
```
mreynolds389: Would you mind making this "ds-margin-top-lg"?  I think a little more separation would be nice

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
```diff
@@ -0,0 +1,2140 @@
...
+                                                                    id="currentSuffixInput"
+                                                                    value={currentSuffixInput}
+                                                                    type="text"
+                                                                    onChange={this.handleSuffixInputChange}
+                                                                    onKeyDown={this.handleSuffixKeyPress}
+                                                                    placeholder={_("Enter a suffix (e.g., dc=example,dc=com)")}
+                                                                    validated={errors.suffixes ? "error" : "default"}
+                                                                    aria-label="Suffix input"
+                                                                />
+                                                            </SplitItem>
```
mreynolds389, line 1377: Perhaps this could be in a followup PR (as this PR is already very large), but I think you should have a multi-select list populated with existing suffixes.   You shouldn't require someone to type out a suffix if all available suffixes are available to the UI.

droideck: Sure, I'll create a new PR later next week (so, the next sprint)!

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
droideck: Agree, it looks better! Done

src/cockpit/389-console/src/lib/monitor/replLogAnalysis.jsx
droideck: Actually, sorry for the fuss; it was just a bug in the already existing feature.
P.S. note, it populates with existing replicated suffixes (not all suffixes)

mreynolds389 (approved):

Issue #6665, opened 2025-03-11
UI - Need to refresh log settings after saving


mreynolds389:
**Issue Description**

While do we reload the config from the parent component it is not resetting the child component as expected

mreynolds389:
778361f67..3e7710114  389-ds-base-3.0 -> 389-ds-base-3.0
79b9ef901..67daf0d45  389-ds-base-2.6 -> 389-ds-base-2.6
2e02ff34e..8cc2568a1  389-ds-base-2.5 -> 389-ds-base-2.5
e7d847aec..22682d75c  389-ds-base-2.4 -> 389-ds-base-2.4

PR #6666, opened 2025-03-11
Issue 6665 - UI - Need to refresh log settings after saving

mreynolds389:
Description:

While do we reload the config from the parent component it is not resetting the child component as expected.

Relates: https://github.com/389ds/389-ds-base/issues/6665


droideck (approved):
LGTM!

PR #6667, opened 2025-03-11
Issue 6639 - Fix crash in upgrade when removing subtree name attribute

mreynolds389:
Description:

Needed to set the deleted values to NULL before issuing the delete operation

Relates: https://github.com/389ds/389-ds-base/issues/6639


progier389 (approved):
LGTM

PR #6670, opened 2025-03-12
Issue 6668 - Fix build break on 1.4.3 branch

### LINKED ISSUE #6668: Build break on 1.4.3 branch
**Issue Description**
Build break on 1.4.3 branch.
While cherry picking issue #6155 to 1.4.3 I noticed a compiulation error in a file that is not impacted by that cherry-pick 

```
/home/progier/sb/t1.4.3/tst/source/389-ds-base/ldap/servers/slapd/entry.c: In function 'slapi_entry_attr_get_valuearray':
/home/progier/sb/t1.4.3/tst/source/389-ds-base/ldap/servers/slapd/entry.c:2894:34: error: returning 'struct slapi_value **' from a function with incompatible return type 'const struct slapi_value **' [-Wincompatible-pointer-types]
 2894 |     return attr->a_present_values.va;

```

 - Platform: Fedora
 - Package and version:  remotes/origin/389-ds-base-1.4.3

**Steps to Reproduce**
1. clone 389-ds-base
2. switch to 389-ds-base-1.4.3 branch
3. Try to build the rpms

**Expected results**
No errors

progier389:
Fixing four issues preventing to build 1.4.3 branch.
  - A missing cast
  - A nasty conflict between statvfs struct and slapd.h f_type macro
  - A missing external declaration in pw_verify.c preventing make -f rpm.mk rpms to complete
  - PREFIX environment variable not taken in account by setup.py 
Now both local build and rpm.mk builds on Fedora should work.

Issue: #6668 

Reviewed by: @droideck

### DIFF
--- a/src/lib389/setup.py
+++ b/src/lib389/setup.py
@@ -13,7 +13,8 @@
 #
 
 from setuptools import setup, find_packages
-from os import path
+from os import path, environ
+from sys import argv
 import build_manpages as bm
 if bm.__version__ < '2.1':
     from build_manpages import build_manpages as bm
@@ -29,6 +30,18 @@ version = "1.4.0.1"
 
 with open(path.join(here, 'README.md'), 'r') as f:
     long_description = f.read()
+
+#
+# For some historical reason when using prefix install
+#  files that should be in /usr/sbin/, /usr/share/ are
+#  in $PREFIX/sbin, $PREFIX/share
+# So lets mimick this behavior
+
+prefix=environ.get('INSTALL_PREFIX', '/usr')
+print(f"argv={argv}")
+if prefix != "/usr" and 'install' in argv and '--prefix' not in argv:
+    argv.append('--prefix')
+    argv.append(prefix)
 
 setup(
     name='lib389',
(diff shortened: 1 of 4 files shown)


progier389:
Apparently there is another problem - maybe about github action as compile fails with
```
checking for libevent... no
configure: error: Package requirements (libevent) were not met:
```

That said, the test action is now able to properly build the rpms
 so I think that we may want to remove the compile action in 1.4.3 as @vashirov  planned to do in  
 https://github.com/389ds/389-ds-base/pull/6584

src/lib389/setup.py
```diff
@@ -30,6 +31,18 @@
...
     long_description = f.read()
 
+#
+# For some historical reason when using prefix install
+#  files that should be in /usr/sbin/, /usr/share/ are
+#  in $PREFIX/sbin, $PREFIX/share
+# So lets mimick this behavior
+
+prefix=environ.get('INSTALL_PREFIX', '/usr')
+print(f"argv={argv}")
```
droideck, line 41: Is it a debug statement?

droideck (approved):
The rest of the changes look good!

src/lib389/setup.py
```diff
@@ -59,19 +72,19 @@
 
     # find lib389/clitools -name ds\* -exec echo \''{}'\', \;
     data_files=[
-        ('/usr/sbin/', [
+        (prefix+'/sbin/', [
```
droideck, line 75: I think `os.path.join()` should be used for paths here

src/lib389/setup.py
progier389: Indeed it is displayed in the output of `make lib389-install` that said, it is also present in the main branch

src/lib389/setup.py
progier389: Technically right, but there is not much interest as we do not support (nor plan to support) Windows anyway. Furthermore prefix+'/sbin/' is used in the main branch ...

progier389:
FYI: About setup.py , I just merged the changes from main branch. The files are now pretty similar:

```
[389-ds-base]$diff  /home/progier/sb/i6668/389-ds-base/src/lib389/setup.py /home/progier/sb/389/389-ds-base/src/lib389/setup.py.in 
26,27d25
< # with open(path.join(here, 'VERSION'), 'r') as version_file:
< # version = version_file.read().strip()
29c27
< version = "1.4.0.1"
---
> version = "@RPM_VERSION@"
79a78
>             'cli/openldap_to_ds',
85a85
>             'man/openldap_to_ds.8',
101c101,102
<         'cryptography'
---
>         'cryptography',
>         'psutil'
```

PR #6673, opened 2025-03-13
Issue 6669 - logconv.py updates

### LINKED ISSUE #6669: logconv.py updates
**Issue Description**
logconv.py fails when specific timestamps are provided

**Steps to Reproduce**
1. Setup ds instance interactively ( with dc=example,dc=com) 
2. Use ldifgen to generate user data
```
dsctl dirsrv ldifgen groups test --number 1 --num-members 50000 --create-members --suffix dc=example,dc=com --ldif-file /var/lib/dirsrv/slapd-dirsrv/ldif/ldifgen.ldif 
```
3. Run logconv.py with start and end time ( preferably use the timestamps mentioned in the access logs)
```
logconv.py -S "[12/Mar/2025:14:06:29.636514463 +0000]" --endTime "[12/Mar/2025:14:21:02.974469376 +0000]" /var/log/dirsrv/slapd-logconv-test/access
```
5. See timestamp error
```
logAnalyser - ERROR - Timestamp 2025-03-12 14:21:03.084300+00:00 outside of range (2025-03-12 14:06:29.636514+00:00 - 2025-03-12 14:21:02.974469+00:00)
```

**Expected results**
Timestamps outside of the specified range should be ignored.

**Additional context**
Reported by https://issues.redhat.com/browse/RHEL-81277

jchapma:
Bug Description:
logconv.py fails when specific timestamps are provided

Fix Description:
Rework time range behaviour, filemagic updates, minor cosmetic updates

Fixes: https://github.com/389ds/389-ds-base/issues/6669

Reviewed by:

### DIFF
--- a/ldap/admin/src/logconv.py
+++ b/ldap/admin/src/logconv.py
@@ -204,7 +204,7 @@ class logAnalyser:
         # Init regex patterns and corresponding actions
         self.regexes = self._init_regexes()
         # Init logger
-        self.logger = self._setup_logger()
+        self.logger = self._setup_logger(logging.ERROR)
 
     def _get_stats_info(self,
                         report_stats_sec: str,
@@ -247,7 +247,7 @@ class logAnalyser:
         except IOError as io_err:
             raise IOError(f"Failed to open file '{stats_file}' for writing: {io_err}")
 
-    def _setup_logger(self):
+    def _setup_logger(self, log_level):
         """
         Setup logging
         """
@@ -257,7 +257,7 @@ class logAnalyser:
         handler = logging.StreamHandler()
         handler.setFormatter(formatter)
 
-        logger.setLevel(logging.WARNING)
+        logger.setLevel(log_level)
         logger.addHandler(handler)
 
         return logger
@@ -585,7 +585,8 @@ class logAnalyser:
 
     def _match_line(self, line: str, bytes_read: int):
         """
-        Do some general maintance on the match and pass it to its match handler function.
+        Process a single line from an access log, match it against predefined patterns,
+        and handle the match by calling its corresponding handler function.
 
         Args:
             line (str): A single line from the access log.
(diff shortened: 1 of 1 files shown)


ldap/admin/src/logconv.py
```diff
@@ -619,13 +619,13 @@ def _match_line(self, line: str, bytes_read: int):
                 if parse_start and parse_stop:
                     if parse_start.microsecond or parse_stop.microsecond:
                         if not parse_start <= norm_timestamp <= parse_stop:
-                            self.logger.error(f"Timestamp {norm_timestamp} outside of range ({parse_start} - {parse_stop})")
-                            return False
+                            self.logger.debug(f"Timestamp {norm_timestamp} outside of range ({parse_start} - {parse_stop})")
+                            return
```
progier389: _return_ seems a bit suspicious as __match_line_ function is supposed to return a boolean 
 as the time does not change in the _pattern,action_ loop maybe return False was ok.
 BTW __match_line_ function should also return False if the _pattern,action_ loops ends without any matches
 (i.e a final return False is missing in the __match_line_ function)

ldap/admin/src/logconv.py
```diff
@@ -1905,13 +1901,15 @@ def convert_timestamp_to_string(self, timestamp: str):
...
+            return ""
 
-        return formatted_timestamp
+        try:
+            timestamp = timestamp[:26] + timestamp[29:]  # Adjust timestamp format
+            dt = datetime.strptime(timestamp, "%d/%b/%Y:%H:%M:%S.%f %z")
+            return dt.strftime("%d/%b/%Y:%H:%M:%S")
+        except ValueError as e:
+            print(f"Timestamp format error: {e}")
+            return ""
```
droideck: It looks very unusual. We don't raise the error here but we just return an empty string (and print the error). But then, we have a `try-except ValueError` wrapper when `log_start_time` and `log_end_time` are calculated. But they will never have the ValueError because it's silenced.

ldap/admin/src/logconv.py
```diff
@@ -2239,14 +2235,12 @@ def main():
...
     try:
-        log_start_time = db.convert_timestamp_to_string(db.server['first_time'])
-        log_end_time = db.convert_timestamp_to_string(db.server['last_time'])
+        log_start_time = db.convert_timestamp_to_string(db.server.get('first_time', 0))
+        log_end_time = db.convert_timestamp_to_string(db.server.get('last_time', 0))
     except ValueError as e:
-        db.logger.error(f"Converting timestamp to datetime object failed")
-        log_start_time = 'Error'
-        log_end_time = 'Error'
+        db.logger.error(f"Converting timestamp to datetime object failed - {e}" )
```
droideck: As of now, I think, this line will never happen.

ldap/admin/src/logconv.py
```diff
@@ -2202,7 +2199,6 @@ def main():
             if os.path.isfile(accesslog):
                 db.process_file(num, accesslog)
             else:
-                # print(f"Invalid file: {accesslog}")
                 db.logger.error(f"Invalid file:{accesslog}")
```
droideck: Missing space after `:`

ldap/admin/src/logconv.py
droideck: Also, if you keep `print(f"Timestamp format error: {e}")` - it probably should go to error level.

droideck (changes requested):

ldap/admin/src/logconv.py
progier389: :+1: I agree, BTW catching the ValueError is a be a good idea but in this case:
[1]   you should print a clear error message (f'Cannot convert timestamp {timesamp} to a date. {e})
[2]   you should reraise the exception 
BTW may be use logger error rather than print for error messages ...

ldap/admin/src/logconv.py
```diff
@@ -247,7 +247,7 @@ def _init_csv_writer(self, stats_file: str):
         except IOError as io_err:
             raise IOError(f"Failed to open file '{stats_file}' for writing: {io_err}")
 
-    def _setup_logger(self):
+    def _setup_logger(self, log_level):
```
droideck: Type hint needed

droideck (approved):
Ack.

P.S. I think we really need test for logconv.py. The code a bit hard to review as some methods are pretty complex and big.

jchapma:
> Ack.
> 
> P.S. I think we really need test for logconv.py. The code a bit hard to review as some methods are pretty complex and big.

I completely agree with you, we do. I'm not quiet sure how to do it though, any ideas ?

droideck:
> > Ack.
> > P.S. I think we really need test for logconv.py. The code a bit hard to review as some methods are pretty complex and big.
> 
> I completely agree with you, we do. I'm not quiet sure how to do it though, any ideas ?

We have some old tests in `dirsrvtests/tests/tickets/ticket47910_test.py` we can try to port.
But also, I think integration tests shouldn't be too hard to develop: generate predictable log data, run the tool, and validate the output. Or do I miss something?

jchapma:
> > > Ack.
> > > P.S. I think we really need test for logconv.py. The code a bit hard to review as some methods are pretty complex and big.
> > 
> > 
> > I completely agree with you, we do. I'm not quiet sure how to do it though, any ideas ?
> 
> We have some old tests in `dirsrvtests/tests/tickets/ticket47910_test.py` we can try to port. But also, I think integration tests shouldn't be too hard to develop: generate predictable log data, run the tool, and validate the output. Or do I miss something?

That sounds good, let me have a look at it. Thanks

Issue #6613, opened 2025-02-14
test_reindex_task_creates_abandoned_index_file CI test is failing


progier389:
**Issue Description**
test_reindex_task_creates_abandoned_index_file CI test is failing
because this test is not adapted to lmdb
`       assert os.path.exists(f"{inst.ds_paths.db_dir}/{DEFAULT_BENAME}/{attr_name.lower()}.db")
`
So the test check database instance existence by checking its file.
but that is not supported when using lmdb (because there is only one unique db file)
should use a dblib agnostic method to test the filename
like checking if one of the backend db instance in `get_mdb_dbis(inst.dbdir)['userrroot'].keys()`  ends with:
`/{attr_name.lower()}` 

**Package Version and Platform:**
 - Platform: RHEL
 - Package and version: 389-ds-base-3.0.6

**Steps to Reproduce**
1. Run dirsrvtests/tests/suites/indexes tests

**Additional context**
Triaged: https://issues.redhat.com/browse/IDMDS-5156

Issue #6672, opened 2025-03-12
Fix test - clu/dbscan_test.py::test_dbscan_destructive_actions fails on BDB


bsimonova:
**Issue Description**
Test clu/dbscan_test.py::test_dbscan_destructive_actions fails on **BDB** with KeyError: 'Unknown dbi userroot/replication_changelog'
After investigation, the path for dbscan list on BDB is set to "/var/lib/dirsrv/slapd-supplier1/" instead of "/var/lib/dirsrv/slapd-supplier1/db" on BDB. 
This results in empty output. LMDB has correct path.

**Package Version and Platform:**
 - Platform: RHEL 9.6
 - Package and version: 389-ds-base-2.6.1-4.el9_6.x86_64

**Steps to Reproduce**
1. Install 389ds on RHEL 9.6
2. Run suites/clu/dbscan_test.py::test_dbscan_destructive_actions

tbordaz:
3b62b9949..d2900a3d3  389-ds-base-2.6
cf7ee2e73..f25a8b72a  389-ds-base-2.5

Issue #6671, opened 2025-03-12
tombstone_fixup_test sometime fails on bdb


progier389:
**Issue Description**
In https://github.com/389ds/389-ds-base/actions/runs/13794554391/job/38639849305#step:7:535 on a fix that does not impact any c files:

```
      # Verify that all tombstones have a nsTombstoneCSN
        for ts in tombstones.list():
            log.info(f'Checking nsTombstoneCSN on tombstone {ts}')
>           assert ts.present("nsTombstoneCSN")
E           assert False
E            +  where False = <bound method Tombstone.present of <class 'lib389.tombstone.Tombstone'>(instance_name="supplier1", dn="ea98e91d-ff5411ef-b0a8f52e-c4f77034")>('nsTombstoneCSN')
E            +    where <bound method Tombstone.present of <class 'lib389.tombstone.Tombstone'>(instance_name="supplier1", dn="ea98e91d-ff5411ef-b0a8f52e-c4f77034")> = <class 'lib389.tombstone.Tombstone'>(instance_name="supplier1", dn="ea98e91d-ff5411ef-b0a8f52e-c4f77034").present
...
FAILED dirsrvtests/tests/suites/replication/tombstone_fixup_test.py::test_precise_tombstone_purging
```

The error seems a bit worrisome as there should not exists any tombstone without nsTombstoneCSN

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main branch

**Steps to Reproduce**
1. Looks at bdb github actions results.

**Expected results**
Test should be reliable

progier389:
Regression is caused by: ed386e16d283a73588aa7fdb1988ca61b6d9ef62  Issue #6639 - remove all the code related to entryrdn_get_switch

mreynolds389:
e99668a61..1ef05b27e  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6674, opened 2025-03-13
Issue 6613 - test_reindex_task_creates_abandoned_index_file fails

progier389:
Test is failing because it was marked as flaky and was never tried on lmdb.
Problem was that the database instance for an index existence was still tested  by looking at the files.
Solution: use dbscan to determine the dbi existence as it was done in clu/dbscan_test.py 
Was done by moving (and adapting) the helper class from clu/dbscan_test.py  to lib389 and reusing it in the test_reindex_task_creates_abandoned_index_file test.

Issue:  #6613

Reviewed by: @droideck (Thank!)

### DIFF
--- a/dirsrvtests/tests/suites/indexes/regression_test.py
+++ b/dirsrvtests/tests/suites/indexes/regression_test.py
@@ -16,6 +16,7 @@ import re
 from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
 from lib389.backend import Backend, Backends, DatabaseConfig
 from lib389.cos import  CosClassicDefinition, CosClassicDefinitions, CosTemplate
+from lib389.cli_ctl.dblib import DbscanHelper
 from lib389.dbgen import dbgen_users
 from lib389.idm.domain import Domain
 from lib389.idm.group import Groups, Group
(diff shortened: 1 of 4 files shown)


src/lib389/lib389/cli_ctl/dblib.py
```diff
@@ -170,7 +258,7 @@ def get_backends(log, dse, tmpdir):
     # now that we finish reading the dse.ldif we may update it if needed.
     for dn, dir in update_dse:
         dse.replace(dn, 'nsslapd-directory', dir)
-    log.debug(f'lib389.cli_ctl.dblib.get_backends returns: {str(res)}')
+    self.log.debug(f'lib389.cli_ctl.dblib.get_backends returns: {str(res)}')
```
droideck: Probably, it should be just `log.debug(f'lib389...`, as there's no `self` in here.

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -163,8 +164,28 @@ def fin(always=False):
     return (indexes, attr)
 
 
-#unstable or unstatus tests, skipped for now
-@pytest.mark.flaky(max_runs=2, min_passes=1)
+def chech_dbi(dbsh, attr_name, expected = True, lowercase=False):
```
droideck: Probably a typo - `chech_dbi -> check_dbi`

And then in other places where it's used

droideck (commented):
The concerns I have are minor.
But the tests fail... I've tried to look a bit and, I think, dbhome path looks strange... But I'm not 100% sure.

dirsrvtests/tests/suites/indexes/regression_test.py
progier389: :+1: At least the typo was consistent! :wink:

progier389:
Fixed the minor issue and I will have a look about the failing tests. I suspect it may be the issue that @bsimonova is seeing in https://github.com/389ds/389-ds-base/issues/6672 (maybe db_home_dir versus db_dir ?)

progier389:
All test pass except one due to a known issue: https://github.com/389ds/389-ds-base/issues/6671

droideck (approved):
LGTM!

PR #6679, opened 2025-03-19
Issue 6571 - Nested group does not receive memberOf attribute

### LINKED ISSUE #6571: Nested group does not receive memberOf attribute.

tbordaz:
Bug description:
	There is a risk to create a loop in group membership.
	For example G2 is member of G1 and G1 is member of G2.
	Memberof plugins iterates from a node to its ancestors
	to update the 'memberof' values of the node.
	The plugin uses a valueset ('already_seen_ndn_vals')
	to keep the track of the nodes it already visited.
	It uses this valueset to detect a possible loop and
	in that case it does not add the ancestor as the
	memberof value of the node.
	This is an error in case there are multiples paths
	up to an ancestor.

Fix description:
	The ancestor should be added to the node systematically,
	just in case the ancestor is in 'already_seen_ndn_vals'
	it skips the final recursion

fixes: #6571

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -1601,7 +1601,7 @@ memberof_call_foreach_dn(Slapi_PBlock *pb __attribute__((unused)), Slapi_DN *sdn
         ht_grp = ancestors_cache_lookup(config, (const void *)ndn);
         if (ht_grp) {
 #if MEMBEROF_CACHE_DEBUG
-            slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM, "memberof_call_foreach_dn: Ancestors of %s already cached (%x)\n", ndn, ht_grp);
+            slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM, "memberof_call_foreach_dn: Ancestors of %s already cached (%lx)\n", ndn, (ulong) ht_grp);
 #endif
             add_ancestors_cbdata(ht_grp, callback_data);
             *cached = 1;
@@ -1609,7 +1609,7 @@ memberof_call_foreach_dn(Slapi_PBlock *pb __attribute__((unused)), Slapi_DN *sdn
         }
     }
 #if MEMBEROF_CACHE_DEBUG
-    slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM, "memberof_call_foreach_dn: Ancestors of %s not cached\n", ndn);
+    slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM, "memberof_call_foreach_dn: Ancestors of %s not cached\n", slapi_sdn_get_ndn(sdn));
 #endif
 
     /* Escape the dn, and build the search filter. */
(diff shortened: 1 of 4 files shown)


progier389 (approved):
LGTM

progier389 (changes requested):
Finally although the changes looks good, the CI tests shows regressions 
 memberof tests now fails on bdb ...

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -3422,9 +3404,21 @@ memberof_get_groups_callback(Slapi_Entry *e, void *callback_data)
...
-        already_seen_ndn_val = slapi_value_new_string(group_ndn);
-        slapi_valueset_add_value_ext(already_seen_ndn_vals, already_seen_ndn_val, SLAPI_VALUE_FLAG_PASSIN);
+        if (slapi_valueset_find(
+            ((memberof_get_groups_data *)callback_data)->config->group_slapiattrs[0], already_seen_ndn_vals, group_ndn_val)) {
+            /* The group group_ndn_val has already been processed
+             * skip the final recursion to prevent infinite loop
+             */
+            slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM,
+                          "memberof_get_groups_callback - detecting a loop in group %s (stop building memberof)\n",
+                           group_ndn);
```
mreynolds389, line 3414: Indentation :-)

mreynolds389 (approved):
LGTM

tbordaz:
Indeed there are many failures... Checking...

progier389:
The most worrisome failure is
dirsrvtests/tests/suites/slapi_memberof/basic_interface_test.py::test_slapi_memberof_reuse_only_2 on bdb

The tombstone test failure is a known issue: https://github.com/389ds/389-ds-base/issues/6671

The vlv reindex failure on lmdb is also a bit surprising (that is the first time I see it) but I do not understand how it could be related to this PR change that only impact a pluigin not involved in reindex task ...
(Maybe the group member have changed and it trigger a bug in reindex code ?) 

I think there is a bug in test_healthcheck_replication_out_of_sync_broken 
 IMHO healthcheck should be run on the same supplier than the one on which the user was added. 
 will create an issue for that one ...
 
 memberof test on lmdb is known to sometime fail but this look the same as the bdb error ...

tbordaz:
> FYI: I reproduce it systematically
> 
> git pull origin pull/6679/head # build ds export NSSLAPD_DB_LIB=bdb cd dirsrvtests/tests/suites/slapi_memberof pytest -x $PWD -k test_slapi_memberof_reuse_only_2
> 
> (After fixing the test case to use the test file path instead of looking in argv) : [patch.txt](https://github.com/user-attachments/files/19374248/patch.txt)
> 
> Not sure why you do not reproduce it. Maybe timing is involved (I am using ASAN build) although logically it should not ...

My bad, I was able to reproduce it and investigate the failure. Actually the test was written on an invalid assumption. The test verifies that slapi_memberof can reuse the results computed by memberof.
The problem was that the test was using 'manager' as membership attribute and it does not work with the way memberof is designed. The link (unique value) 'manager', processed by memberof plugin, add into a given entry all persons having this entry as direct/indirect manager. While the test was designed the other way around to retrieve in a given entry all managers it refers to.
I changed the test, using the regular 'member' membership value.

progier389 (approved):
Tests are now OK. 1 failure that is a known issue.

PR #6681, opened 2025-03-19
Issue 6680 - instance read-only mode is broken

### LINKED ISSUE #6680: instance read-only mode is broken.
**Issue Description**
Server fails to start after turning on nsslapd-readonly=on in cn=config.

 - Platform: Fedora

**Steps to Reproduce**
1. Configure a 389-ds-base instance
2. dsconf _instance_ config replace nsslapd-readonly=on
3. dsctl _instance_ restart
4. See error: The server fails to starts and errors logs shows errors

**Expected results**
Server should be able to start without errors

**Additional context**
Triaged: https://issues.redhat.com/browse/IDMDS-4841

progier389:
Read only mode is broken because some plugins fails to starts as they are not able to create/updates some entries in the dse backend.
Solution is to allow interrnal operations to write in dse.backend but not modify the dse.ldif (except for the special case when trying to modify nsslapd-readonly flags (to be allowed to set/unset the readonly mode)

Issue: #6680 

Reviewed by: @droideck, @tbordaz  (thanks!)

### DIFF
--- a/dirsrvtests/tests/suites/config/regression_test.py
+++ b/dirsrvtests/tests/suites/config/regression_test.py
@@ -28,6 +28,8 @@ CUSTOM_MEM = '9100100100'
 IDLETIMEOUT = 5
 DN_TEST_USER = f'uid={TEST_USER_PROPERTIES["uid"]},ou=People,{DEFAULT_SUFFIX}'
 
+RO_ATTR = 'nsslapd-readonly'
+
 
 @pytest.fixture(scope="module")
 def idletimeout_topo(topo, request):
@@ -188,3 +190,41 @@ def test_idletimeout(idletimeout_topo, dn, expected_result):
     except ldap.SERVER_DOWN:
         result = True
     assert expected_result == result
+
+
+def test_instance_readonly_mode(topo):
+    """Check that readonly mode is supported
+
+    :id: 34d2e28e-04d7-11f0-b0cf-482ae39447e5
+    :setup: Standalone Instance
+    :steps:
+        1. Set readonly mode
+        2. Restart the instance
+        3. Check that readonly mode is set
+        4. Try to modify another config attribute
+        5. Unset readonly mode
+        6. Restart the instance
+        7. Check that readonly mode is unset
+        8. Try to modify another config attribute
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Should get ldap.UNWILLING_TO_PERFORM exception
+        5. Success
+        6. Success
+        7. Success
+        8. Success
+    """
+
+    inst = topo.standalone
+    inst.config.replace(RO_ATTR, 'on')
+    inst.restart
+    assert inst.config.get_attr_val_utf8(RO_ATTR) == "on"
+    attr = 'nsslapd-errorlog-maxlogsize'
+    val = inst.config.get_attr_val_utf8(attr)
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        inst.config.replace(attr, val)
+    inst.config.replace(RO_ATTR, 'off')
+    inst.restart
+    assert inst.config.get_attr_val_utf8(RO_ATTR) == "off"
--- a/dirsrvtests/tests/suites/config/regression_test.py
+++ b/dirsrvtests/tests/suites/config/regression_test.py
@@ -199,32 +199,52 @@ def test_instance_readonly_mode(topo):
     :setup: Standalone Instance
     :steps:
         1. Set readonly mode
-        2. Restart the instance
-        3. Check that readonly mode is set
-        4. Try to modify another config attribute
-        5. Unset readonly mode
-        6. Restart the instance
-        7. Check that readonly mode is unset
+        2. Stop the instance
+        3. Get dse.ldif modification time
+        4. Start the instance
+        5. Get dse.ldif modification time
+        6. Check that modification time has not changed
+        7. Check that readonly mode is set
         8. Try to modify another config attribute
+        9. Unset readonly mode
+        10. Restart the instance
+        11. Check that modification time has not changed
+        12. Check that modification time has changed
+        13. Check that readonly mode is unset
+        14. Try to modify another config attribute
     :expectedresults:
         1. Success
         2. Success
         3. Success
-        4. Should get ldap.UNWILLING_TO_PERFORM exception
+        4. Success
         5. Success
         6. Success
         7. Success
-        8. Success
+        8. Should get ldap.UNWILLING_TO_PERFORM exception
+        9. Success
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/dse.c
```diff
@@ -1031,6 +1031,114 @@ dse_check_for_readonly_error(Slapi_PBlock *pb, struct dse *pdse)
...
+    Slapi_Regex *re_isro = recomp("^" CONFIG_READONLY_ATTRIBUTE ":\\s+on\\s*$");
+    Slapi_Regex *re_eoe = recomp("^$");
+    bool isconfigentry = false;
+    bool isro = false;
+    FILE *fdse = NULL;
+    char line[128];
+    char *error = NULL;
+    const char *regexp = "";
+
+    if (dsename) {
```
droideck: Should this be `if (!dsename) {` - or do I miss something?

ldap/servers/slapd/dse.c
```diff
@@ -1031,6 +1031,114 @@ dse_check_for_readonly_error(Slapi_PBlock *pb, struct dse *pdse)
...
+        slapi_log_err(SLAPI_LOG_ERR, "is_readonly_set_in_dse",
+                      "Failed to compile '%s' regular expression. Error is %s\n",
+                      regexp, error);
+    }
+    slapi_ch_free_string(&error);
+    return re;
+}
+
+/*
+ * Check if "nsslapd-readlony: on" is in cn-config in dse.ldif file
```
droideck: Typo - `readlony -> readonly` in a few places

droideck (approved):
Besides the minor issues, LGTM!

ldap/servers/slapd/dse.c
progier389: Good catch but now I wonder why the testcase is working ?

ldap/servers/slapd/dse.c
```diff
@@ -1031,6 +1031,114 @@ dse_check_for_readonly_error(Slapi_PBlock *pb, struct dse *pdse)
...
+        return false;
+    }
+    if (!slapi_config_get_readonly()) {
+        ro = false;
+        return ro;
+    }
+    if (ro) {
+        /* read-only mode and dse is up to date ==> Do not modify it. */
+        return ro;
+    }
```
droideck, line 1130: I think the tests worked because we return here.
I can't recall exactly the order config init operates on the start-up, but I think it has the config read into memory at this point.

And the core part of the fix is `is_rejected_op` func.

dirsrvtests/tests/suites/config/regression_test.py
```diff
@@ -188,3 +190,41 @@ def test_idletimeout(idletimeout_topo, dn, expected_result):
...
+    inst.config.replace(RO_ATTR, 'on')
+    inst.restart
+    assert inst.config.get_attr_val_utf8(RO_ATTR) == "on"
+    attr = 'nsslapd-errorlog-maxlogsize'
+    val = inst.config.get_attr_val_utf8(attr)
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        inst.config.replace(attr, val)
+    inst.config.replace(RO_ATTR, 'off')
+    inst.restart
+    assert inst.config.get_attr_val_utf8(RO_ATTR) == "off"
```
tbordaz, line 249: Is it missing the step 7 - update a config attribute when RO is disabled

ldap/servers/slapd/dse.c
```diff
@@ -1031,6 +1031,114 @@ dse_check_for_readonly_error(Slapi_PBlock *pb, struct dse *pdse)
...
+
+/*
+ * Check if "nsslapd-readonly: on" is in cn-config in dse.ldif file
+ * ( If the flag is set in memory but on in the file, the file should
+ *   be written (to let dsconf able to modify the nsslapd-readonly flag)
+ */
+static bool
+is_readonly_set_in_dse(const char *dsename)
+{
+    Slapi_Regex *re_config = recomp("^dn:\\s+cn=config\\s*$");
```
tbordaz, line 1057: Shouldn't it match only for a line ending with 'cn=config' ?

ldap/servers/slapd/mapping_tree.c
```diff
@@ -2058,6 +2058,82 @@ slapi_dn_write_needs_referral(Slapi_DN *target_sdn, Slapi_Entry **referral)
...
+    /* Read operations are not rejected */
+    if ((be_op_type == SLAPI_OPERATION_SEARCH) ||
+        (be_op_type == SLAPI_OPERATION_COMPARE) ||
+        (be_op_type == SLAPI_OPERATION_BIND) ||
+        (be_op_type == SLAPI_OPERATION_UNBIND)) {
+        return false;
+    }
+
+    /* Tombstone fixup are not rejected. */
+    if (operation_is_flag_set(op, OP_FLAG_TOMBSTONE_FIXUP)) {
```
tbordaz, line 2093: Should we also accept RUV update (OP_FLAG_REPL_RUV) ? As there are no update, RUV should not change so it is probably fine to ignore RUV.
Or it may be handle by later test that accept all internal operation.

tbordaz (approved):
LGTM

ldap/servers/slapd/mapping_tree.c
progier389: Internal op are accepted only on  dse backendthat is not replicated and has no tombstone ...
Since old code was already only checking OP_FLAG_TOMBSTONE_FIXUP, I think that your first assumption is right: no  write operation means no RUV nor tombstone change

ldap/servers/slapd/dse.c
progier389: The impact of the dsename bug is that _is_readonly_set_in_dse_ always returns _false_, so ro variable stays _false_ and internal operation get written in the dse.file. 
I think that the reason the testcase pass is because it does not check that dse file is not modified after the restart.
I will improve that.

dirsrvtests/tests/suites/config/regression_test.py
progier389: Sorry, I do not understand what do you mean. Could you precise ?
|Oh, I guess you meant Step 8: Try to modify another config attribute
  (That is now step 14. I will add it)

ldap/servers/slapd/dse.c
progier389: Matching a line ending with 'cn=config' would also work because cn=config is always the first entry that match the _ending with 'cn=config'_ condition
But since the function is searching for nsslapd-readonly in cn=config entry, doing an exact match is more logical. 
FYI: nsslapd-readonly is also an attribute of the backend config entry and we should ignore it.

Issue #6683, opened 2025-03-20
test_healthcheck_replication_out_of_sync_broken sometimes fails


progier389:
**Issue Description**
dirsrvtests/tests/suites/healthcheck/health_repl_test.py::test_healthcheck_replication_out_of_sync_broken sometime fails as seen in https://github.com/389ds/389-ds-base/actions/runs/13970232962/job/39110682229.

I believe that the issue is that healthcheck is run on the wrong replica 
 (it should be run on the replica on which the user was added ...)

**Package Version and Platform:**
 - Platform: [e.g. Fedora]
 - Package and version: main branch

**Steps to Reproduce**
1. See CI tests results in github

PR #6684, opened 2025-03-20
Issue 6683 - test_healthcheck_replication_out_of_sync_broken mail fail

progier389:
Testcase sometime fail because of a race condition.
The fix is to make sure that replications sessions are restarted after breaking the replication.

Issue: #6683 


droideck (approved):
LGTM!

Issue #6494, opened 2025-01-10
Various errors when using extended matching rule on vlv sort filter


progier389:
**Issue Description**
libasan reports several issues when trying to configre a vlv whose sort filter contains an extended matching rule.
While fixing the test I saw crashes SIGSEGV, use-after-free, double-free, read overflow and memory leaks.

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: At least main branch.

**Steps to Reproduce**
1. Install ds version with libasan
2. Run following script:
```
find $PREFIX | grep asan | xargs /bin/rm -f
INSTANCE=i
SUFFIX='dc=com,dc=example'
S1=$PREFIX/var/run/slapd-$INSTANCE.socket
Socket=$(echo $S1 | sed 's,/,%2f,g')

echo "
[general]
[slapd]
instance_name=i

[backend-userroot]
create_suffix_entry=yes
sample_entries=yes
suffix= $SUFFIX
" >/tmp/temp.ldap

set -x
dsctl $INSTANCE remove --do-it
dscreate from-file /tmp/temp.ldap

echo "TEST 1 ==> ds should crash"
dsconf $INSTANCE backend vlv-index add-search --name "VLV People" --search-base "ou=People,$SUFFIX" --search-filter "cn=*" --search-scope 2 userRoot
dsconf $INSTANCE backend vlv-index add-index --index-name "SORT_UID_emr" --sort "uid:2.5.13.2" --parent-name "VLV People" --index-it "$SUFFIX"

echo "TEST 2 ==> ds should hang + merror after restart"
dsconf $INSTANCE backend vlv-index add-index --index-name "SORT_UID" --sort "uid" --parent-name "VLV People" --index-it "dc=$SUFFIX"
echo quit | ldapsearch -Q -LLL -Y EXTERNAL -H ldapi://$Socket -b "ou=People,$SUFFIX" "cn=*" "distinguishedName" -E 'sss=uid:2.5.13.2' -E 'vlv=0/999999/0/0'
dsconf $INSTANCE backend vlv-index add-index --index-name "SORT_UID" --sort "uid" --parent-name "VLV People" --index-it "$SUFFIX"
echo quit | ldapsearch -Q -LLL -Y EXTERNAL -H ldapi://$Socket -b "ou=People,$SUFFIX" "cn=*" "distinguishedName" -E 'sss=uid:2.5.13.2' -E 'vlv=0/999999/0/0'
[...]

PR #6685, opened 2025-03-21
Issue 6494 - (3rd) Various errors when using extended matching rule on vlv sort filter

mmatsuya:
Bug description:

dirsrvtests/tests/suites/vlv/regression_test.py::test_vlv_with_mr failed on 389-ds-base-2.2

This 3rd commit (in Issue 6494) is required for 389-ds-base-2.2, since DEMO_PW definition is missing.
Also, topology_st is used without import. It's fixed in Issue 1925.

Fixes: https://github.com/389ds/389-ds-base/issues/6494
Related: https://github.com/389ds/389-ds-base/issues/1925

### DIFF
--- a/dirsrvtests/tests/suites/vlv/regression_test.py
+++ b/dirsrvtests/tests/suites/vlv/regression_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2018 Red Hat, Inc.
+# Copyright (C) 2023 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

mmatsuya:
To fix the vlv test failure in pipeline, I changed the two following points.

- To add the suffix argument into add_users. With this fix, test_vlv_with_mr test doesn't use the default suffix used by other tests. 
As a result, the failure doesn't happen during test_vlv_with_mr test. This follows the current implementation in newer branch (including main).
But, it may happen in other tests or a test added newly in the future. So, the next one is required.
- To suppress ldap.ALREADY_EXISTS during user creation. In case that the user entry exists, just it's ignored.

mmatsuya:
waiting for the vlv result.

mmatsuya:
vlv test passed.

progier389:
New changes still looks good !

PR #6690, opened 2025-03-24
Issue 6671 - tombstone_fixup_test sometime fails on bdb

mreynolds389:
Description:

The diff is confusing but it basically just removed an "if" evaluation that was supposed to be always true:

    -        if (entryrdn_get_switch() ||
    -            !slapi_entry_flag_is_set(fi->entry->ep_entry,
    +        if (!slapi_entry_flag_is_set(fi->entry->ep_entry,
                                          SLAPI_ENTRY_FLAG_TOMBSTONE)) {

This was incorrect, this code should always be run regardless of the entry flags

relates: https://github.com/389ds/389-ds-base/issues/6639
relates: https://github.com/389ds/389-ds-base/issues/6671

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_import_threads.c
@@ -2409,117 +2409,114 @@ bdb_import_foreman(void *param)
             goto error;
         }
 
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM, test is now passing. There are still some failing tests but they are known issues

Issue #6695, opened 2025-03-26
UI - fix more minor issues


mreynolds389:
**Issue Description**

Need to fix:

- MemberOf page - subtree scopes - allows you to create blank values, and does not validate suffixes
- Local password policy - copy/paste error added a duplicate label for a checkbox (send expiring warning)
- Create instance modal - did not validate database name
- Update connection monitor chart to include detailed information for connection status (established, close wait, and time wait)

PR #6696, opened 2025-03-26
Issue 6695 - UI - fix more minor issues

mreynolds389:
Description:

fix the following items:

- MemberOf page - subtree scopes - allows you to create blank values, and does not validate suffixes 
- Local password policy - copy/paste error added a duplicate label for a checkbox (send expiring warning) 
- Create instance modal - did not validate database name 
- Update connection monitor chart to include detailed information for connection status (established, close wait, and time wait)

relates: https://github.com/389ds/389-ds-base/issues/6695

### DIFF
--- a/src/cockpit/389-console/src/dsModals.jsx
+++ b/src/cockpit/389-console/src/dsModals.jsx
@@ -4,7 +4,13 @@ import PropTypes from "prop-types";
 import { DoubleConfirmModal } from "./lib/notifications.jsx";
 import { BackupTable } from "./lib/database/databaseTables.jsx";
 import { BackupModal } from "./lib/database/backups.jsx";
-import { log_cmd, bad_file_name, valid_dn, callCmdStreamPassword } from "./lib/tools.jsx";
+import {
+    log_cmd,
+    bad_file_name,
+    valid_dn,
+    valid_db_name,
+    callCmdStreamPassword
+} from "./lib/tools.jsx";
 import {
     Button,
     Checkbox,
@@ -103,10 +109,6 @@ export class CreateInstanceModal extends React.Component {
             'createDM'
         ];
 
-        const optionalAttrs = [
-            'createDBName'
-        ];
-
         // Handle server ID
         if (this.state.createServerId !== "") {
             if (this.state.createServerId.length > 80) {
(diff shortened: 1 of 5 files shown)


src/cockpit/389-console/src/dsModals.jsx
```diff
@@ -142,11 +144,9 @@ export class CreateInstanceModal extends React.Component {
         }
 
         if (this.state.createDBCheckbox) {
-            for (const attr of optionalAttrs) {
-                if (this.state[attr] === "") {
-                    all_good = false;
-                    errObj[attr] = true;
-                }
+            if (!valid_db_name(this.state.createDBName) || this.state.createDBName === "") {
```
droideck: Probably, `this.state.createDBName === ""` can be removed as valid_db_name checks for an empty string.

droideck (approved):
I see a strange duplication in UI when I point to 00:00:00 at the corner of the chart (when it changes to the current time at the corner - the issue disappears):

image

Besides these couple of minor issues - looks good!

mreynolds389:
> I see a strange duplication in UI when I point to 00:00:00 at the corner of the chart (when it changes to the current time at the corner - the issue disappears):
> 

This is unavoidable, it's just how it works in PF.  The charts always do this and I spent alot of time trying to fix it before, but it's just how it works.

mreynolds389:
> > I see a strange duplication in UI when I point to 00:00:00 at the corner of the chart (when it changes to the current time at the corner - the issue disappears):
> 
> This is unavoidable, it's just how it works in PF. The charts always do this and I spent alot of time trying to fix it before, but it's just how it works.

Ok I sort of improved it.  @vashirov and I were debating this before.  Basically to get rid of it you can't have intervals with the same value - that's why they are currently stacked.  You also need to preset the number of intervals you want in the chart otherwise it gets all messed up.  It's annoying. So I added 1 second intervals until the real data comes in, but it also looks odd.  

Screenshot From 2025-03-27 08-29-24

mreynolds389:
> > > I see a strange duplication in UI when I point to 00:00:00 at the corner of the chart (when it changes to the current time at the corner - the issue disappears):
> > 
> > 
> > This is unavoidable, it's just how it works in PF. The charts always do this and I spent alot of time trying to fix it before, but it's just how it works.
> 
> Ok I sort of improved it. @vashirov and I were debating this before. Basically to get rid of it you can't have intervals with the same value - that's why they are currently stacked. You also need to preset the number of intervals you want in the chart otherwise it gets all messed up. It's annoying. So I added 1 second intervals until the real data comes in, but it also looks odd.

Changed the interval to match he refresh rate (currently 10 seconds)

droideck (approved):
Looks much better! Ack

PR #6697, opened 2025-03-27
Issue 6571 - (2nd) Nested group does not receive memberOf attribute

### LINKED ISSUE #6571: Nested group does not receive memberOf attribute.

tbordaz:
Bug description:
        erroneous debug change made in previous fix
	where cache_ancestors is called with the wrong parameter

Fix description:
        Restore the orginal param 'member_data'
	Increase the set of tests around multipaths

fixes: #6571

### DIFF
--- a/dirsrvtests/tests/suites/memberof_plugin/regression_test.py
+++ b/dirsrvtests/tests/suites/memberof_plugin/regression_test.py
@@ -595,6 +595,8 @@ def test_multipaths(topology_st, request):
                              'homeDirectory': '/home/user1'
                              })
     group = Groups(inst, SUFFIX, rdn=None)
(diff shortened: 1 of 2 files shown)


droideck (approved):
LGTM

Issue #6698, opened 2025-03-27
NPE after configuring invalid filtered role


progier389:
**Issue Description**
Null pointer exception may occur after creating a filtered role with an invalid filter.
More detail in https://issues.redhat.com/browse/RHEL-67019

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version:  main branch but all existing version are impacted.

**Steps to Reproduce**
1. See RHEL-67019

**Expected results**

**Screenshots**

**Additional context**
Triaged: https://issues.redhat.com/browse/IDMDS-4865

progier389:
ed8987efe..31e120d23  main -> main
   a4b2def77..0b3342a6c  389-ds-base-3.0 -> 389-ds-base-3.0
   c237c8d74..3e7ba4d84  389-ds-base-2.6 -> 389-ds-base-2.6
   790655f9d..30776f908  389-ds-base-2.5 -> 389-ds-base-2.5
   9bf5707fb..48d6ed56e  389-ds-base-2.4 -> 389-ds-base-2.4
   dd337449e..c6b0eb629  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6699, opened 2025-03-27
Issue 6698 - NPE after configuring invalid filtered role

progier389:
Server crash when doing search after configuring filtered role with invalid filter.
Reason: The part of the filter that should be overwritten are freed before knowing that the filter is invalid.
Solution: Check first that the filter is valid before freeing the filtere bits

Issue: #6698 

Reviewed by: @tbordaz , @mreynolds389  (Thanks!)

### DIFF
--- a/ldap/servers/slapd/filter.c
+++ b/ldap/servers/slapd/filter.c
@@ -1033,9 +1033,11 @@ slapi_filter_get_subfilt(
 }
 
 /*
- * Before calling this function, you must free all the parts
+ * The function does not know how to free all the parts
  * which will be overwritten (i.e. slapi_free_the_filter_bits),
- * this function dosn't know how to do that
+ * so the caller must take care of that.
+ * But it must do so AFTER calling slapi_filter_replace_ex to
+ * avoid getting invalid filter if slapi_filter_replace_ex fails.
  */
 int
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):

ldap/servers/slapd/filter.c
```diff
@@ -1094,8 +1096,15 @@ slapi_filter_free_bits(Slapi_Filter *f)
 int
 slapi_filter_replace_strfilter(Slapi_Filter *f, char *strfilter)
 {
-    slapi_filter_free_bits(f);
-    return (slapi_filter_replace_ex(f, strfilter));
+    /* slapi_filter_replace_ex may fail and we cannot
```
tbordaz, line 1099: I agree with the fact to check the filter before freeing the bits. Shouldn't we also fix slapi_filter_replace_strfilter ?

tbordaz (approved):

ldap/servers/slapd/filter.c
progier389: I do not understand your comment:
The fix is already inside slapi_filter_replace_strfilter and apparently that is the only place where slapi_filter_free_bits is called...

tbordaz: Oooppps.... sorry I misread the function name. forget my comment :)

PR #6709, opened 2025-03-28
Issue 6595 - Regression test in betxn_test.py failing due to busy LDAP server. 

### LINKED ISSUE #6595: Regression test in betxn_test.py failing due to busy LDAP server
**Issue Description**
The regression test dirsrvtests/tests/suites/betxns/betxn_test.py::test_betxn_modrdn_memberof_cache_corruption is failing with the following error:
```
E         ldap.BUSY: {'msgtype': 109, 'msgid': 12, 'result': 51, 'desc': 'Server is busy', 'ctrls': [], 'info': 'target entry busy because of a canceled operation'}

/usr/local/lib64/python3.9/site-packages/ldap/ldapobject.py:128: BUSY
```
This looks to be due to the server still processing the first group.rename() call, which is meant to fail with an ldap.OBJECT_CLASS_VIOLATION exception, as the second group.rename() call is sent in. 
I mitigated this on the test environment with a time.sleep(2), but there should be a better way to fix this. Will keep looking. 

 - Platform: RHEL 9
 - Package and version: Multiple builds, but I found it on 389-ds-base-2.4.5-12.el9_4

**Steps to Reproduce**
1. Run the regression test on a RHEL9 test VM. The test is failing on the main branch currently. 
2. The test will fail. 

**Expected results**
The test passes, assuming the build used doesn't cause the test to fail in other ways. 

**Additional context**
I haven't seen this issue documented elsewhere, so I'm opening it here for reference while I investigate further.

abeisemb:
The regression test dirsrvtests/tests/suites/betxns/betxn_test.py::test_betxn_modrdn_memberof_cache_corruption throws an ldap.BUSY error due to two intentionally failing group.rename() calls in quick succession. The topology is still processing the first failed call when the second one hits. This fix catches the ldap.BUSY and sleeps for a short time before trying again to give the topology time to finish.

Resolves: https://github.com/389ds/389-ds-base/issues/6595

### DIFF
--- a/dirsrvtests/tests/suites/betxns/betxn_test.py
+++ b/dirsrvtests/tests/suites/betxns/betxn_test.py
@@ -29,6 +29,16 @@ logging.getLogger(__name__).setLevel(logging.DEBUG)
 log = logging.getLogger(__name__)
 USER_PASSWORD = 'password'
 
+def retry_if_busy(f):
+    for x in range(1000):
+        try:
+            return f()
+        except ldap.BUSY:
+            time.sleep(0.02)
+
+    log.error("Too many retries!")
+    assert False
+
 
 def test_betxt_7bit(topology_st):
     """Test that the 7-bit plugin correctly rejects an invalid update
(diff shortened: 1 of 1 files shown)


progier389:
Change looks good. Waiting for the test result before approving the PR.

dirsrvtests/tests/suites/betxns/betxn_test.py
```diff
@@ -29,6 +29,13 @@
 log = logging.getLogger(__name__)
 USER_PASSWORD = 'password'
 
+def retry_if_busy(f):
+    while True:
```
tbordaz: Do you think it worth adding a max retry ? Just to prevent an infinite loop of the test if similar problem popup again ?

progier389: Probably a good idea (I needed to add such limit in my testcase because first patch attempt was always returning LDAP_BUSY).
Note: the change is pretty simple:
 I replaced the `while True:`  by: `for x in range(1000):` and 
```
log.error('Too many retries')
assert False
```
 after the loop

abeisemb: Thanks for the feedback. I've amended the commit with the recommended changes.

progier389 (approved):
LGTM

Issue #6710, opened 2025-03-28
test_vlv_scope_one_on_two_backends fails in 2.6


tbordaz:
**Issue Description**
test_vlv_scope_one_on_two_backends remove the database files during the test. In 2.6, database is bdb and in addition to the .db files it requires to remove the env files under '/dev/shm/slapd-instance'. If they are not removed, at startup bdb tries to retrieve .db files that have been remove and fails to start 

**Package Version and Platform:**
 related to test_vlv_scope_one_on_two_backends

**Steps to Reproduce**
Run the CI test_vlv_scope_one_on_two_backends tests in 2.x

**Expected results**
Test should succeeds

tbordaz:
e10553ba7..e6827695e  389-ds-base-2.6
3396f8683..b8ca6fba0  389-ds-base-2.5

PR #6711, opened 2025-03-28
Issue 6710 - test_vlv_scope_one_on_two_backends fails in 2.6

tbordaz:
Bug description:
	remove_database removes database files but with bdb
	it does not remove the DB_ENV (/dev/shm/slapd-inst).
	At startup the DB detect incoherency between ENV
	and (not) existing database files then fail

Fix description:
	remove_database also remove env files with bdb

fixes: #6710

Reviewed by:


progier389 (approved):
LGTM

Issue #6700, opened 2025-03-27
CLI - schema - objectclass info should also include superior objectclass's allowed and requires attrs


mreynolds389:
**Issue Description**

When you get/list an objectclass it only lists its level of allowed and required objectclasses, but it should also include all its superior objectclasses' allowed and required attributes.  

This will also fix some issues in the Web UI when adding objectclasses to entries, where there are attributes that are required by the superior objectclasses.

A good example is "inetOrgPerson" which has superior objectclasses: organizationalPerson and Person.  "Person" requires "sn" and "cn" and this trips up the UI when adding this OC as it does not know "sn" and "cn" are required.

mreynolds389:
13cb8bedd..a6a631042  389-ds-base-3.0 -> 389-ds-base-3.0
e6827695e..3b62b9949  389-ds-base-2.6 -> 389-ds-base-2.6
b8ca6fba0..cf7ee2e73  389-ds-base-2.5 -> 389-ds-base-2.5
963ff7127..94acbe701  389-ds-base-2.4 -> 389-ds-base-2.4

PR #6712, opened 2025-03-30
Issue 6700 - CLI/UI - include superior objectclasses' allowed and requires attrs

mreynolds389:
Description:

When you get/list an objectclass it only lists its level of allowed and required objectclasses, but it should also include all its superior objectclasses' allowed and required attributes.

Added an option to the CLI to also include all the parent/superior required and allowed attributes

Relates: https://github.com/389ds/389-ds-base/issues/6700

### DIFF
--- a/src/lib389/lib389/schema.py
+++ b/src/lib389/lib389/schema.py
@@ -116,15 +116,66 @@ class Schema(DSLdapObject):
             result = ATTR_SYNTAXES
         return result
 
-    def _get_schema_objects(self, object_model, json=False):
-        """Get all the schema objects for a specific model: Attribute, Objectclass,
-        or Matchingreule.
+    def gather_oc_sup_attrs(self, oc, sup_oc, ocs, processed_ocs=None):
+        """
+        Recursively build up all the objectclass superiors' may/must
+        attributes
+
+        @param oc - original objectclass we are building up
+        @param sup_oc - superior objectclass that we are gathering must/may
+                        attributes from, and for following its superior
+                        objectclass
+        @param ocs - all objectclasses
+        @param processed_ocs - list of all the superior objectclasees we have
+                               already processed. Used for checking if we
+                               somehow get into an infinite loop
+        """
+        if processed_ocs is None:
+            # First pass, init our values
+            sup_oc = oc
+            processed_ocs = [sup_oc['names'][0]]
+        elif sup_oc['names'][0] in processed_ocs:
+            # We're looping, need to abort. This should never happen because
+            # of how the schema is structured, but perhaps a bug was
+            # introduced in the server schema handling?
+            return
+
+        # update processed list to prevent loops
+        processed_ocs.append(sup_oc['names'][0])
+
+        for soc in sup_oc['sup']:
+            if soc.lower() == "top":
+                continue
+            # Get sup_oc
+            for obj in ocs:
+                oc_dict = vars(ObjectClass(obj))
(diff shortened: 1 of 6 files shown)


mreynolds389:
Still need to write a test...

src/lib389/lib389/schema.py
```diff
@@ -116,15 +116,51 @@ def get_attr_syntaxes(json=False):
...
+
+                    # Sort the tuples
+                    may = list(oc['may'])
+                    may.sort()
+                    oc['may'] = tuple(may)
+                    must = list(oc['must'])
+                    must.sort()
+                    oc['must'] = tuple(must)
+
+                    # Now recurse and check this objectclass
```
tbordaz, line 166: I had a doubt about a possible infinite loop if for example OC1 -- sup --> OC2 -- sup --> OC3 -- sup --> OC1. But I was not able to find a testcase to produce this state (need to define superior before child and DEL/ADD superior if it has a child).

mreynolds389: Yeah this is possible with custom schema - can't trust an admin to be an LDAP expert.  I can keep track of OC's looked at, and break out if we see the same OC again

mreynolds389: Seems like the server will not allow me to create schema in this nested/looping fashion.

droideck: I agree it'll be nice to add some `visited = set()` - just in case, as a safe precaution for a clean code.
Even though we see that the server guards it now.

src/lib389/lib389/schema.py
```diff
@@ -116,15 +116,51 @@ def get_attr_syntaxes(json=False):
...
+                        if attr not in oc['may']:
+                            oc['may'] = oc['may'] + (attr,)
+                    for attr in oc_dict['must']:
+                        if attr not in oc['must']:
+                            oc['must'] = oc['must'] + (attr,)
+
+                    # Sort the tuples
+                    may = list(oc['may'])
+                    may.sort()
+                    oc['may'] = tuple(may)
```
droideck: Minor neat pick, but I think it'll be a bit cleaner to use `sorted()` directly instead of `list() -> .sort()` conversion.
And it can be done with one line - `oc['may'] = tuple(sorted(oc['may']))`

droideck (approved):
The rest looks good to me!

src/lib389/lib389/schema.py
mreynolds389: I have code to check it, but I seriously can not get the schema into a state where it can happen.  So I don't want to add that code as it's misleading since it's not possible.

droideck: Yep, I also wasn't able to reproduce the loop. But adding the visited set is just good defensive programming - not misleading code at all, IMO.

If schema validation changes in the future or a bug slips in, we'd already be protected against the recursive loops. It's minimal code overhead for solid protection, and a pretty standard recursion pattern anyway.

mreynolds389: Fair enough, I'll add the code - it's not a big deal.  I guess I'm scarred by the fact that this product is so old that when you see the code doing some kind of check, you assume that it was there for a valid reason - that it was preventing a real problem.  That's why I used the word "misleading".  So I'll add some kind of comment clarifying that it's just being cautious.  I appreciate your persistence/intentions, and it is good practice as long as it's clear to the future reader.

PR #6714, opened 2025-03-31
Issue 6713 - ns-slapd crash during mdb offline import

### LINKED ISSUE #6713: ns-slapd crash during mdb offline import
**Issue Description**
A segmentation fault is triggered in dbmdb_import_prepare_worker_entry() during an  mdb offline import

**Steps to Reproduce**
1. Create a fresh instance
2. Generate the ldif file with 2 million users
```
dsctl inst01 ldifgen users --number 2_000_000 --suffix dc=example,dc=com --ldif-file /var/lib/dirsrv/slapd-inst01/ldif/ldifgen.ldif
``` 
3. Stop the instance
4. Import the generated ldif file 
```
dsctl inst01 ldif2db userRoot /var/lib/dirsrv/slapd-inst01/ldif/ldifgen.ldif
```
5. Notice SEGV
```
Thread 4 "ns-slapd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffb640006c0 (LWP 122046)]
__strncmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283

Thread 4 (Thread 0x7ffb640006c0 (LWP 122046) "ns-slapd"):
#0  __strncmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283
#1  0x00007ffff3d70dce in dbmdb_import_prepare_worker_entry (wqelmnt=0x897cd0) at ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c:1398
#2  0x00007ffff3d73531 in dbmdb_import_worker (param=<optimized out>) at ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c:3299
#3  0x00007ffff7454749 in _pt_root (arg=0x86b4c0) at ../../../../nspr/pr/src/pthreads/ptthread.c:201
#4  0x00007ffff7903947 in start_thread (arg=<optimized out>) at pthread_create.c:444
#5  0x00007ffff7989970 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

jchapma:
Bug description: A segmentation fault is triggered in dbmdb_import_prepare_worker_entry() during an mdb offline import.

The import producer thread parses, validates and writes ldif entries to the worker queue, while the import worker threads simultaneously read, format and index entries before adding them to the DB. A race condition occurs when a worker thread reads an entry before the producer has fully written it, leading to a segmentation fault.

Fix description: Ensure thread safe access by locking the worker queue before writing entries.

Fixes: https://github.com/389ds/389-ds-base/issues/6713

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
@@ -372,7 +372,9 @@ void dbmdb_dup_worker_slot(struct importqueue *q, void *from_slot, void *to_slot
     char *from = from_slot;
     char *to = to_slot;
     int offset = offsetof(WorkerQueueData_t, wait_id);
+    pthread_mutex_lock(&q->mutex);
     memcpy(to+offset, from+offset, (sizeof (WorkerQueueData_t))-offset);
+    pthread_mutex_unlock(&q->mutex);
 }
 
 /* Find a free slot once used_slots == max_slots */
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
@@ -372,9 +372,7 @@ void dbmdb_dup_worker_slot(struct importqueue *q, void *from_slot, void *to_slot
     char *from = from_slot;
     char *to = to_slot;
     int offset = offsetof(WorkerQueueData_t, wait_id);
-    pthread_mutex_lock(&q->mutex);
     memcpy(to+offset, from+offset, (sizeof (WorkerQueueData_t))-offset);
-    pthread_mutex_unlock(&q->mutex);
 }
 
 /* Find a free slot once used_slots == max_slots */
@@ -419,13 +417,14 @@ dbmdb_import_workerq_push(ImportQueue_t *q, WorkerQueueData_t *data)
             safe_cond_wait(&q->cv, &q->mutex);
         }
     }
-    pthread_mutex_unlock(&q->mutex);
     if (q->job->flags & FLAG_ABORT) {
         /* in this case worker thread does not free the data so we should do it */
         dbmdb_import_workerq_free_data(data);
+        pthread_mutex_unlock(&q->mutex);
         return -1;
     }
     dbmdb_dup_worker_slot(q, data, slot);
+    pthread_mutex_unlock(&q->mutex);
     return 0;
 }
 


jchapma:
Something we still need to understand is why an MDB offline import runs faster when we use CLOCK_MONOTONIC instead of CLOCK_REALTIME as the clock id. @vashirov discovered this a while ago.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -372,7 +372,9 @@ void dbmdb_dup_worker_slot(struct importqueue *q, void *from_slot, void *to_slot
     char *from = from_slot;
     char *to = to_slot;
     int offset = offsetof(WorkerQueueData_t, wait_id);
+    pthread_mutex_lock(&q->mutex);
```
progier389: Two remarks:
1.  protecting only the memcpy seems a bit weird. I have the feeling that there could still be race condition.
2. Why add a new lock/unlock (performance impact!) ?
IMHO , we should rather modify the caller function (i.e: dbmdb_import_workerq_push) and extends its critical to include the call to  dbmdb_dup_worker_slot

progier389 (changes requested):

jchapma:
> > Something we still need to understand is why an MDB offline import runs faster when we use CLOCK_MONOTONIC instead of CLOCK_REALTIME as the clock id. @vashirov discovered this a while ago.
> 
> Probably worth to open a Jira ticket to investigate the import performance issue in general and note that CLOCK_MONOTONIC is more efficient than CLOCK_REALTIME (IMHO the reason could simply be due the cost of accessing these different clocks) What puzzles me is that the import statistics are quite different from what I got when I implemented the import code.

I agree, let me open an investigate ticket for this.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
jchapma: > Two remarks:
> 
> 1. protecting only the memcpy seems a bit weird. I have the feeling that there could still be race condition.
> 2. Why add a new lock/unlock (performance impact!) ?
>    IMHO , we should rather modify the caller function (i.e: dbmdb_import_workerq_push) and extends its critical to include the call to  dbmdb_dup_worker_slot

1. The reason I added the lock around the memcpy() is that this is where we write to the worker queue. I ran tsan before and after the change, the after tsan report didnt highlight this area as a potential race condition.
2. I think I tried extending the critical section to include dbmdb_dup_worker_slot() but there was a perf degradation. But I need to confirm this.

tbordaz: @progier389 , something not clear to me. a 'slot' is available when wait_id is 0.
Before selecting this slot [here](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c#L414), shouldn't we check that that wait_id is 0 ?

progier389: There are 2 cases: either some slots are never been used  (i.e: q->used_slots < q->max_slots) and wait_id is 0 because it was never changed (FYI that is the case in the line 414 that you are pointing) or all slots have been used and  dbmdb_get_free_worker_slot is called to look for a slot with wait_is == 0

tbordaz: Okay. Thanks for the explanation.
As wait_id is the trigger of processing by the worker. Would it be an option to memcpy the struct except wait_id. Then only after copying the wait_id ?

progier389: Hum, I do not think so:  
If memcpy and wait_id assignment are in the same critical section, then order does not matter but if they are not:
- if memcpy is done after wait_id assignment, there is a risk that the slot reader read incomplete data. (That is the cause of the crash)
- but if it is done before there is a risk that another thread will also pick the same slot and one of the data will be overwritten ...

jchapma: > * protecting only the memcpy seems a bit weird. I have the feeling that there could still be race condition.
> * Why add a new lock/unlock (performance impact!) ?
>   IMHO , we should rather modify the caller function (i.e: dbmdb_import_workerq_push) and extends its critical to include the call to  dbmdb_dup_worker_slot

Performance comparison when importing 2M user ldif.

With locking around the memcpy to the worker queue:
ThreadSanitizer: reported 35 warnings
average rate 1106.2/sec
Import writer thread usage: run: 15.44% read: 4.86% write: 76.64% pause: 0.02% txnbegin: 0.00% txncommit: 3.04%

With dbmdb_dup_worker_slot() moved into the critical section of dbmdb_import_workerq_push():
ThreadSanitizer: reported 61 warnings
average rate 1128.0/sec
Import writer thread usage: run: 15.09% read: 4.31% write: 77.57% pause: 0.02% txnbegin: 0.00% txncommit: 3.01%

progier389: That the rate is slightly faster is expected as less lock/unlock are done but I am surprised with the increased ThreadSanitizer reported warnings. (maybe something is wrong with your change ? (in error handling ?)

jchapma:
Moved the lock into dbmdb_import_workerq_push(), we can look into the tsan report as part of the later investigation.

progier389 (approved):
LGTM

Issue #6715, opened 2025-03-31
dsconf backend replication monitor fails if replica id starts with 0


progier389:
**Issue Description**
dsconf monitor reports Unavailable consumer if replica id starts with 0 
 (like '001' )

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: latest and previous versions

**Steps to Reproduce**
1. Setup a MMR between two nodes
2. Set the replica id to 010 and 020
3. Initialize the replication
4. Run dsconf supplier1 replication monitor -c "_host_:_PortSupplier2_:cn=directory manager:password"
5. Check the replication status

**Expected results**
Should not see 'Unavailable' in the monitoring output

progier389:
RHEL Bug: https://issues.redhat.com/browse/RHEL-67003

Triaged: https://issues.redhat.com/browse/IDMDS-4861

progier389:
815a0f9e0..d19a4f995  main -> main
   961b9e2e3..bcc855236  389-ds-base-3.0 -> 389-ds-base-3.0
   3e4a95a0a..9489cf310  389-ds-base-2.6 -> 389-ds-base-2.6
   a8df8dcca..da48b44a2  389-ds-base-2.5 -> 389-ds-base-2.5
   57e30687b..f986a5bb6  389-ds-base-2.4 -> 389-ds-base-2.4

progier389:
The regression in 2.4 is due to the fact that issue #6736 must be cherry-picked too

PR #6716, opened 2025-04-01
Issue 6715 - dsconf backend replication monitor fails if replica id starts with 0

progier389:
lib389 fails to retrioeve the csn if the replicaid is not exactly the normalized form of a number
Typically because 010 does not match with the RUV value.
Solution: Ensure that rid get normalized before comparing them or using them in a dict

Issue: #6715 

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -1258,14 +1258,14 @@ def test_normalized_rid_dict():
         nrd[key] = val
 
     # Check that normalization do something
-    assert nkeys != [ sd.keys() ]
+    assert nkeys != list(sd.keys())
 
     # Check that key stored in NormalizedRidDict are normalized
     for key in nrd.keys():
         assert key in nkeys
 
     # Check that normalized and non normalized keys have the same value
-    for keyl in sd.items():
+    for key,val in sd.items():
         nkey = NormalizedRidDict.normalize_rid(key)
         assert nrd[key] == val
         assert nrd[nkey] == val
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -30,7 +30,10 @@ from lib389.idm.group import Groups, Group
 from lib389.idm.domain import Domain
 from lib389.idm.directorymanager import DirectoryManager
 from lib389.idm.services import ServiceAccounts, ServiceAccount
-from lib389.replica import Replicas, ReplicationManager, ReplicationMonitor, ReplicaRole, BootstrapReplicationManager
+from lib389.replica import (
+    Replicas, ReplicationManager, ReplicationMonitor, ReplicaRole,
+    BootstrapReplicationManager, NormalizedRidDict
+)
 from lib389.agreement import Agreements
 from lib389 import pid_from_file
 from lib389.dseldif import *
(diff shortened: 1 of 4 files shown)


tbordaz (approved):
LGTM

src/lib389/lib389/replica.py
```diff
@@ -822,6 +822,25 @@ def demote(self, suffix, newrole):
...
+
+    def __getitem__(self, key):
+        nkey = int(key)
+        self._log.debug(f'MYDBG: __getitem__ {key} {nkey} {self.__getitem__(nkey)}')
+        return self.__getitem__(nkey)
+
+    def __setitem__(self, key, value):
+        nkey = int(key)
+        self._log.debug(f'MYDBG: __setitem__ {key} {nkey} {value}')
+        self.d.__setitem__(nkey, value)
```
droideck: I am a bit concerned with the dict implementation.

At least, I think, it needs to use `super().` in the returns and in the init - `super().__init__()`

Also, the debug messages with `MYDBG` look strange. If we want to keep the messages, I'd at least change the `MYDBG` part:)

droideck (changes requested):

src/lib389/lib389/replica.py
droideck: Also, maybe rename it to something like `NormalizedRidDict`?

progier389: Ok with all your remarks. 
In fact the way I wrote _rdict, it is not really a dict but an object containing a dict. It works because we only get/set elements in this dict but it is clearly inconsistent ...

I will change that and also add a testcase to check the  NormalizedRidDict behavior

dirsrvtests/tests/suites/replication/regression_m2_test.py
```diff
@@ -1144,6 +1148,129 @@ def test_bulk_import(preserve_topo_m2):
...
+
+    # Check that normalization do something
+    assert nkeys != [ sd.keys() ]
+
+    # Check that key stored in NormalizedRidDict are normalized
+    for key in nrd.keys():
+        assert key in nkeys
+
+    # Check that normalized and non normalized keys have the same value
+    for keyl in sd.items():
```
droideck: probably, it should be `for key, val in sd.items():`

dirsrvtests/tests/suites/replication/regression_m2_test.py
```diff
@@ -1144,6 +1148,129 @@ def test_bulk_import(preserve_topo_m2):
...
+    nsd = { NormalizedRidDict.normalize_rid(key): val for key,val in sd.items() }
+    nkeys = list(nsd.keys())
+
+    # Initialize a NormalizedRidDict
+    nrd = NormalizedRidDict()
+    for key,val in sd.items():
+        nrd[key] = val
+
+    # Check that normalization do something
+    assert nkeys != [ sd.keys() ]
```
droideck: I think it should be `assert nkeys != list(sd.keys())`

Otherwise, it looks like this:

```
>>> test1 = {"key1": 23, "key2": 333}
>>> test2 = test1.keys()
>>> test2
dict_keys(['key1', 'key2'])
>>> test3 = [test1.keys()]
>>> test3
[dict_keys(['key1', 'key2'])]
```

And that's more like intended:

```
>>> test4 = list(test1.keys())
>>> test4
['key1', 'key2']
```

droideck (approved):
Besides the minor issues - LGTM!

P.S. CI is very slow today...

dirsrvtests/tests/suites/replication/regression_m2_test.py
progier389: Ok. 
I  do not understand how the test could PASS with such an error ...
dirsrvtests/tests/suites/replication/regression_m2_test.py::test_normalized_rid_dict PASSED [ 93%]

Finally I understand: there was two error, keyl instead of key and the missing val 
so the test was using the same values for key and val in each loop ...

PR #6723, opened 2025-04-02
Issue 6481 - When ports that are in use are used to update a DS instance the error message is not helpful

### LINKED ISSUE #6481: UI - When ports that are in use are used to update a DS instance the error message is not helpful

jchapma:
Bug description: The initial fix for this issue used sudo lsof to determineif a server port was in use. lsof is not installed by default and using sudo can ask for a password.

Fix description: Use ss -ntplu (which is installed by default) to detect if a port is taken.

Fixes: https://github.com/389ds/389-ds-base/issues/6481

Fix Author:  Viktor Ashirov <vashirov@redhat.com>

Reviewed by:

### DIFF
--- a/src/cockpit/389-console/src/lib/tools.jsx
+++ b/src/cockpit/389-console/src/lib/tools.jsx
@@ -221,7 +221,7 @@ export function is_port_in_use(port) {
             return;
         }
 
-        let cmd = ['bash', '-c', `sudo lsof -i :${port} || echo "free"`];
+        let cmd = ['bash', '-c', `ss -ntplu | grep -w :${port} || echo "free"`];
         log_cmd("is_port_in_use", cmd);
 
         cockpit


progier389 (commented):
Several remarks:

- As I recently commented in RHEL bug, maybe we should considers if we should not try first to use lsof (because it provides the name of the command that use the port unlike ss)  (and using sudo -n ... to avoid trying to ask for pasword) then if the result is "free" (i.e port not found or lsof not installed or sudo failed) then double check with ss

vashirov:
`ss` provides process name if it's executed with elevated privileges. But why do we need to know the process name in the first place? If the port is taken by any process, we shouldn't attempt to bind to it.

vashirov:
The way it currently works, if the new port is taken, Save button is disabled. We don't convey to the user that it's taken by some other process. So I don't think we need to get this information.

progier389:
Ok, I was tricked because the command output get in cmd but as it is is only used to check if it is "free" or not we do not care. 
So current change looks good.

progier389 (approved):
LGTM

PR #6726, opened 2025-04-04
Issue 5356 - Set DEFAULT_PASSWORD_STORAGE_SCHEME to PBKDF2-SHA512 in tests

ssidhaye:
Now the default password storage scheme is PBKDF2-SHA512
Hence updating it in the test to reflect current state.
DEFAULT_PASSWORD_STORAGE_SCHEME is now set to PBKDF2-SHA512

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_security_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_security_test.py
@@ -99,14 +99,10 @@ def test_healthcheck_insecure_pwd_hash_configured(topology_st):
     run_healthcheck_and_flush_log(topology_st, standalone, json=False, searched_code=RET_CODE)
     run_healthcheck_and_flush_log(topology_st, standalone, json=True, searched_code=RET_CODE)
 
-    if is_fips():
-        log.info('Set passwordStorageScheme and nsslapd-rootpwstoragescheme to SSHA512 in FIPS mode')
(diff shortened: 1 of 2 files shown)


jchapma (approved):
LGTM

droideck (changes requested):
Could you please also `grep is_fips` through the code base and check other places?
Like https://github.com/389ds/389-ds-base/blob/a42f464c3faf6d4c4cb52fada8be7ccf1169374a/dirsrvtests/tests/suites/healthcheck/health_security_test.py#L102-L105

Firstyear:
> Now the default password storage scheme is PBKDF2-SHA512

If we update default password storage it should be to argon2id, not pbkdf2, since argon2id gives greater cryptographic resistance to attacks on GPUs.

vashirov:
> > Now the default password storage scheme is PBKDF2-SHA512
> 
> If we update default password storage it should be to argon2id, not pbkdf2, since argon2id gives greater cryptographic resistance to attacks on GPUs.

This PR is to update the tests to match default storage scheme that was updated in 2022: https://github.com/389ds/389-ds-base/commit/99a74d7b6c3c003c23205c5d2db246ef61862ffb 
And we don't support argon2id as of now.

vashirov (approved):

droideck (approved):
LGTM!

PR #6729, opened 2025-04-04
Issue 6728 - CLI - Issue with user rename operation

### LINKED ISSUE #6728: CLI - Issue with user rename operation
**Issue Description**
The rename functionality in dsidm CLI incorrectly handles RDN values when rename operations are performed, especially with the keep-old-rdn option. This causes issues with user identification after renaming and produces inconsistent results when listing or searching for renamed users.

droideck:
Description: Fix the RDN extraction method in _generic_list to correctly display renamed user entries regardless of keep-old-rdn status.

Improve the _generic_rename function to properly handle nonexistent users and detect when no actual changes are made.

Add test cases for various rename scenarios including nonexistent users, duplicate names, keep-old-rdn behavior, and listing after rename. Enhance error handling for edge cases like renaming to an existing username.

Fixes: https://github.com/389ds/389-ds-base/issues/6728

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_idm/__init__.py
+++ b/src/lib389/lib389/cli_idm/__init__.py
@@ -111,7 +111,10 @@ def _generic_list(inst, basedn, log, manager_class, args=None):
         if args and args.json:
             json_result = {"type": "list", "items": []}
         for o in ol:
-            o_str = o.__unicode__()
+            # Get the RDN value directly from the DN using ldap.dn.str2dn
+            # This ensures we get the actual RDN value even if keep-old-rdn was used
+            rdn_components = ldap.dn.str2dn(o.dn)[0]
+            o_str = rdn_components[0][1]  # Get the value part of the RDN
             if args and args.json:
                 json_result['items'].append(o_str)
             else:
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/cli_idm/__init__.py
```diff
@@ -111,7 +111,10 @@ def _generic_list(inst, basedn, log, manager_class, args=None):
         if args and args.json:
             json_result = {"type": "list", "items": []}
         for o in ol:
-            o_str = o.__unicode__()
+            # Get the RDN value directly from the DN using ldap.dn.str2dn
+            # This ensures we get the actual RDN value even if keep-old-rdn was used
+            rdn_components = ldap.dn.str2dn(o.dn)[0]
+            o_str = rdn_components[0][1]  # Get the value part of the RDN
```
progier389: How are rdn with multiple AVAs handled ? (i.e: `cn=John Doe+mail=john.doe@example.com,dc=example,dc=com` ) ?

I wonder if there is not a flaw in the was we handle rdn in __mapped_object ...
And seeing the comment in DSLdapObject.__unicode__is suspect that the answer is ARGHHH ! :wink: 

We could fix (or not) that in another issue but in that case, I think it is better to 
create a get_rdn_from_dn function in DSLdapObject class and use it.
 (will be helpful if we decide to look for the value associated with the primary attribute in the rdn AVA list)

droideck: For multiple AVAs handling, we have this ticket opened: https://github.com/389ds/389-ds-base/issues/3329

So yeah, I think it's best to be addressed separately:) I've put the issue number in 'TODO' comment in the code, though.

mreynolds389 (approved):
Since some of the other issues will be addressed in other tickets I'll ack this

progier389 (approved):
LGTM

Issue #6515, opened 2025-01-20
Json option does not work with "dsidm account get-by-dn" command


mirielka:
**Issue Description**
Running "dsidm account get-by-dn" with json option (-j) does not provide output in the form of JSON object

**Package Version and Platform:**
389-ds-base-3.0.5-2.el10.x86_64

**Steps to Reproduce**
1. Run "dsidm -j account get-by-dn" on an existing account

**Expected results**
Output in the form of JSON object

**Actual results**
```
# dsidm -j localhost -b dc=example,dc=com account get-by-dn uid=demo_user,ou=people,dc=example,dc=com

dn: uid=demo_user,ou=people,dc=example,dc=com
cn: Demo User
displayName: Demo User
gidNumber: 99998
homeDirectory: /var/empty
legalName: Demo User Name
loginShell: /bin/false
objectClass: top
objectClass: nsPerson
objectClass: nsAccount
objectClass: nsOrgPerson
objectClass: posixAccount
uid: demo_user
uidNumber: 99998
```

mreynolds389:
40434bd19..0a066b552  389-ds-base-3.0 -> 389-ds-base-3.0
0bbe5f799..ec44ed81b  389-ds-base-2.6 -> 389-ds-base-2.6
12f5b005b..a2b055784  389-ds-base-2.5 -> 389-ds-base-2.5
13a9e5554..5994557ac  389-ds-base-2.4 -> 389-ds-base-2.4
d7629c6d4..6fed77659  389-ds-base-2.3 -> 389-ds-base-2.3
ac0a09047..d65f8cc9b  389-ds-base-2.2 -> 389-ds-base-2.2
e097f395f..91aca7968  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6731, opened 2025-04-08
Issue 6515 - CLI - dsidm get_dn does not return JSON format

mreynolds389:
Description:

The get_dn() functions do not check for the JSON flag.  Also added a LogCapture function to return the the raw results so we can do json validation.  Also did some other minor python lint cleanup.

Relates: https://github.com/389ds/389-ds-base/issues/6515

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsidm_account_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_account_test.py
@@ -1,21 +1,24 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2023 Red Hat, Inc.
+# Copyright (C) 2025 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 #
-import ldap
-import time
-import subprocess
-import pytest
 import logging
 import os
-
+import json
+import pytest
+import ldap
 from lib389 import DEFAULT_SUFFIX
-from lib389.cli_idm.account import list, get_dn, lock, unlock, delete, modify, rename, entry_status, \
-    subtree_status, reset_password, change_password
+from lib389.cli_idm.account import (
+    get_dn,
+    lock,
+    unlock,
+    entry_status,
+    subtree_status,
+)
 from lib389.topologies import topology_st
 from lib389.cli_base import FakeArgs
 from lib389.utils import ds_is_older
@@ -33,6 +36,7 @@ def create_test_user(topology_st, request):
     log.info('Create test user')
     users = nsUserAccounts(topology_st.standalone, DEFAULT_SUFFIX)
     test_user = users.create_test_user()
+    log.info('Created test user: %s', test_user.dn)
 
     def fin():
         log.info('Delete test user')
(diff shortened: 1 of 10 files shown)


dirsrvtests/tests/suites/clu/dsidm_account_test.py
```diff
@@ -1,21 +1,30 @@
...
+    get_dn,
+    lock,
+    unlock,
+    # delete,
+    # modify,
+    # rename,
+    entry_status,
+    subtree_status,
+    # reset_password,
+    # change_password
```
droideck: If we don't need imports, I think it's better just remove them

src/lib389/lib389/cli_base/__init__.py
```diff
@@ -34,7 +33,7 @@ def _get_arg(args, msg=None, hidden=False, confirm=False):
                     x = getpass("%s : " % msg)
                     y = getpass("CONFIRM - %s : " % msg)
                     if x != y:
-                        log.info("Passwords do not match, try again.")
+                        print("Passwords do not match, try again.")
```
droideck, line 36: Is it a typo? Other places (for both JSON and non-JSON) were changed to `log.info`.

dirsrvtests/tests/suites/clu/dsidm_role_test.py
```diff
@@ -771,9 +771,18 @@ def test_dsidm_role_subtree_status(topology_st, create_test_managed_role):
...
     subtree_status(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+    result = topology_st.logcap.get_raw_outputs()
     check_value_in_log_and_reset(topology_st, check_value=output)
 
+    log.info('Test dsidm role subtree-status - json')
+    args.json = True
+    subtree_status(standalone, DEFAULT_SUFFIX, topology_st.logcap.log, args)
+    result = topology_st.logcap.get_raw_outputs()
+    json_result = json.loads(result[0])
+    assert json_result['entries'][0]['dn'] == test_role.dn
```
droideck: I see the test fails here:

```
>       assert json_result['entries'][0]['dn'] == test_role.dn
E       AssertionError: assert 'cn=nsManaged...xample,dc=com' == 'cn=test_mana...xample,dc=com'
E         - cn=test_managed_role,dc=example,dc=com
E         ?    ^^ ^^^      ^^
E         + cn=nsManagedDisabledRole,dc=example,dc=com
E         ?    ^ ^      ^^^^^^^^^
```

droideck (approved):
Besides the minor issues and test, looks good to me!

dirsrvtests/tests/suites/clu/dsidm_account_test.py
mreynolds389: > If we don't need imports, I think it's better just remove them

I guess I saw this as a todo list :-)

src/lib389/lib389/cli_base/__init__.py
mreynolds389: > Is it a typo? Other places (for both JSON and non-JSON) were changed to `log.info`.

We don't have "log" available.  Not sure why it previously worked, but it caused a lint error so I just cleaned it up

dirsrvtests/tests/suites/clu/dsidm_role_test.py
mreynolds389: Odd, I'll look into this...  But I think I was just testing the function and not the entire file - so other roles got added from previous tests.  Anyway I'll fix it.

src/lib389/lib389/cli_base/__init__.py
droideck: Oh, ok, got it!

Issue #6735, opened 2025-04-08
CLI - dsidm provide option to set decription when creating an entry


mreynolds389:
**Issue Description**

In the Web UI when you create a group, and other entries, you can set the description during the "add", but dsidm only allows you to set it after creating the entry.

mreynolds389:
ffa7a57a9..117b63abf  389-ds-base-3.0 -> 389-ds-base-3.0
cecb72cf7..2bd87b603  389-ds-base-2.6 -> 389-ds-base-2.6
e513efc21..02d3ddd6a  389-ds-base-2.5 -> 389-ds-base-2.5
627468eb8..5ed1c90ea  389-ds-base-2.4 -> 389-ds-base-2.4
c1ecfaf73..d06915e8b  389-ds-base-2.3 -> 389-ds-base-2.3
b4d683917..494ea090c  389-ds-base-2.2 -> 389-ds-base-2.2
903e0398e..ececf7020  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6737, opened 2025-04-09
Issue 6735 - CLI - dsidm provide option to set decription when creating an entry

mreynolds389:
Description:

In the Web UI when you create a group, and other entries, you can set the description during the "add", but dsidm only allows you to set it after creating the entry. To be consistent UI and CLI should work the same way

Relates: https://github.com/389ds/389-ds-base/issues/6735

### DIFF
--- a/src/lib389/lib389/cli_idm/group.py
+++ b/src/lib389/lib389/cli_idm/group.py
@@ -8,7 +8,7 @@
 # --- END COPYRIGHT BLOCK ---
 
 import json
-from lib389.idm.group import Group, Groups, MUST_ATTRIBUTES
+from lib389.idm.group import Group, Groups, MAY_ATTRIBUTES, MUST_ATTRIBUTES
 from lib389.cli_base import populate_attr_arguments, _generic_modify, CustomHelpFormatter
 from lib389.cli_idm import (
     _generic_list,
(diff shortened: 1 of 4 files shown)


src/lib389/lib389/cli_idm/group.py
```diff
@@ -44,7 +44,7 @@ def get_dn(inst, basedn, log, args):
 
 
 def create(inst, basedn, log, args):
-    kwargs = _get_attributes(args, MUST_ATTRIBUTES)
+    kwargs = _get_attributes(args, MUST_ATTRIBUTES, MAY_ATTRIBUTES)
```
droideck, line 47: I'm not sure why but when I run `dsidm inst group create` without any optional arguments, there's no interactive `Enter value for description:` prompt

mreynolds389: > I'm not sure why but when I run `dsidm inst group create` without any optional arguments, there's no interactive `Enter value for description:` prompt

Right, that's because it's not required.  Otherwise it breaks CI tests when description is not set

mreynolds389: So you can also do this and it will still interactively ask for requires attributes : 

    # dsidm inst group create --description="test"
    Enter value for cn:

droideck: Okay, thinking more about that, that's good too.
I have no hard preference here:)

mreynolds389: > Okay, thinking more about that, that's good too. I have no hard preference here:)

Yeah it's an odd issue.  Would be nice to prompt for it, but it's not required, and everywhere else we only prompt for required attributes.

droideck (approved):
LGTM

Issue #6734, opened 2025-04-08
CLI - dsidm should have option to list DN's


mreynolds389:
**Issue Description**

Implement "list dn" option in dsidm.  In some cases you can have objects with the same RDN, and in such cases you need to know the full DN to use the "get-by-dn" option

mreynolds389:
d6beb024e..09708ffec  389-ds-base-3.0 -> 389-ds-base-3.0
2bd87b603..92db91166  389-ds-base-2.6 -> 389-ds-base-2.6
02d3ddd6a..b73b97218  389-ds-base-2.5 -> 389-ds-base-2.5
5ed1c90ea..1dd7f130e  389-ds-base-2.4 -> 389-ds-base-2.4
d06915e8b..3ceac184b  389-ds-base-2.3 -> 389-ds-base-2.3
494ea090c..b0c7110a3  389-ds-base-2.2 -> 389-ds-base-2.2
ececf7020..fe3ef99f0  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

Firstyear:
This appears to have broken 2.2 and greater:

```
09708ffec9 (Mark Reynolds   2025-04-08 14:12:36 -0400 1304)             raise ldap.UNWILLING_TO_PERFORM(f"Too many objects matched selection criteria: {criteria} {search_filter}"
09708ffec9 (Mark Reynolds   2025-04-08 14:12:36 -0400 1305)                                             f" - Please use 'get-by-dn' to specify which entry to get:\n    {'\n    '.join(entry_dn)}")
```

```
[  206s]   File "/home/abuild/rpmbuild/BUILD/389-ds-base-2.2.10~git89.e25be4bc/src/lib389/lib389/__init__.py", line 72, in <module>
[  206s]     from lib389.tasks import BackupTask, RestoreTask
[  206s]   File "/home/abuild/rpmbuild/BUILD/389-ds-base-2.2.10~git89.e25be4bc/src/lib389/lib389/tasks.py", line 14, in <module>
[  206s]     from lib389._mapped_object import DSLdapObject
[  206s]   File "/home/abuild/rpmbuild/BUILD/389-ds-base-2.2.10~git89.e25be4bc/src/lib389/lib389/_mapped_object.py", line 1265
[  206s]     raise ldap.UNWILLING_TO_PERFORM(f"Too many objects matched selection criteria: {criteria} {search_filter}"
[  206s]                                    ^
[...]

PR #6738, opened 2025-04-09
Issue 6743 - CLI - dsidm add option to list DN's

mreynolds389:
Description: Add an option to list the full DN instead of the RDN

Relates: https://github.com/389ds/389-ds-base/issues/6734

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsidm_role_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_role_test.py
@@ -19,7 +19,7 @@ from lib389.cli_idm.role import (
     )
 from lib389.topologies import topology_st
 from lib389.cli_base import FakeArgs
-from lib389.utils import ds_is_older
+from lib389.utils import ds_is_older, is_a_dn
 from lib389.idm.role import Roles, ManagedRoles, FilteredRoles, NestedRoles
 from . import check_value_in_log_and_reset, check_value_in_log
 
@@ -266,7 +266,6 @@ def test_dsidm_role_delete(topology_st, create_test_managed_role):
     assert not test_managed_role.exists()
 
 
-@pytest.mark.xfail(reason="DS6488")
 @pytest.mark.skipif(ds_is_older("1.4.2"), reason="Not implemented")
 def test_dsidm_role_list(topology_st, create_test_managed_role):
     """ Test dsidm role list option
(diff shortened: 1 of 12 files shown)


src/lib389/lib389/cli_idm/account.py
```diff
@@ -235,6 +235,8 @@ def create_parser(subparsers):
 
     list_parser = subcommands.add_parser('list', help='list accounts that could login to the directory', formatter_class=CustomHelpFormatter)
     list_parser.set_defaults(func=list)
+    list_parser.add_argument('--full-dn', action='store_true',
```
droideck: I'm not sure we need it here as it returns DN by default (and there's no way to return a name).
And I think it's okay as it returns all accounts (not only users). So just DN makes sense.

dirsrvtests/tests/suites/clu/dsidm_role_test.py
```diff
@@ -289,11 +290,13 @@ def test_dsidm_role_list(topology_st, create_test_managed_role):
         5. Success
         6. Success
         7. Success
+        8. Success
     """
 
     standalone = topology_st.standalone
     args = FakeArgs()
     args.json = False
+    args.full_dn = False
```
droideck, line 299: I think we need to update other tests' `FakeArgs` too by adding the `args.full_dn = False`

droideck (changes requested):

src/lib389/lib389/cli_idm/account.py
mreynolds389: > I'm not sure we need it here as it returns DN by default (and there's no way to return a name). And I think it's okay as it returns all accounts (not only users). So just DN makes sense.

I'm confused by your comment, it returns the RDN "value" by default _not_ the DN.  That's the whole point of this feature/PR is to have an option to list the DN's.

dirsrvtests/tests/suites/clu/dsidm_role_test.py
mreynolds389: > I think we need to update other tests' `FakeArgs` too by adding the `args.full_dn = False`

Yeah sounds good, and same for the JSON option....

mreynolds389: > I think we need to update other tests' `FakeArgs` too by adding the `args.full_dn = False`

Right, there are other CI failures because of this.  I'll get that fixed...

src/lib389/lib389/cli_idm/account.py
mreynolds389: Whoops!  "account list" does show the full DN by default.  Nice catch!!  Fixed

droideck (approved):
Looks good!

Issue #6736, opened 2025-04-09
Exception thrown by dsconf instance repl get_ruv


progier389:
**Issue Description**
Exception thrown by dsconf instance repl get_ruv
test_ruv_url_not_added_if_different_uuid CI test failure 

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: main branch

**Steps to Reproduce**
1. install 389-ds-base
2. configure 2 suppliers and their agreements
3. Run dsconf supplier1 replication get-ruv --suffix dc=example,dc=com
4. See error: (no RUV but getting an error message)

Steps to reproduce the behavior with the CI test:
1. install 389-ds-base
2. git clone https://github.com/389ds/389-ds-base.git
3. cd 389-ds-base
4. pytest -x dirsrvtests/tests/suites/replication -k test_ruv_url_not_added_if_different_uuid
5. See error:
```
dirsrvtests/tests/suites/replication/regression_m2c2_test.py::test_ruv_url_not_added_if_different_uuid FAILED [100%]
...
            log.error(f"No URL for RID {replica_m2.get_rid()} in RUV")
>           assert False
E           assert False
```

**Expected results**
Test should PASS

**Additional context**
Behavior changed recently  (Probably a regression ...)

Apparently the RUV object has no URL

DEBUG    dirsrvtests.tests.suites.replication.regression_m2c2_test:regression_m2c2_test.py:136 c1 RUV: {'data_generation': {'name': '67f68823000000010000', 'value': None}, 'ruvs': [{'raw_ruv': None, 'rid': '1', 'url': None, 'csn': '1970-01-01 00:00:00', 'raw_csn': '00000000000000000000', 'maxcsn': '1970-01-01 00:00:00', 'raw_maxcsn': '00000000000000000000', 'modts': '1970-01-01 00:00:00'}]}

I suspect a change in lib389:
get-ruv seems broken too:

```
dsconf consumer2 replication get-ruv --suffix dc=example,dc=com
Error: can only concatenate str (not "NoneType") to str

```

progier389:
the consumer RUV looks OK: 
```
dn: cn=replica,cn=dc\3Dexample\2Cdc\3Dcom,cn=mapping tree,cn=config
nsds50ruv: {replicageneration} 67f6882a000000020000
nsds50ruv: {replica 2 ldap://linux.home:39002} 67f68838000000020000 67f68838000000020000
```
so the problem is likely in lib389.replica

progier389:
[...]

PR #6742, opened 2025-04-10
Issue 6736 - Exception thrown by dsconf instance repl get_ruv

progier389:
After issue #6715 commit `dsconf instance replication get_ruv` fails with error
because by default NormalizedRidDict does not calling __getitem__ as I expected.
The solution is to overwrite the get function to do things properly.

Issue: #6736
Relates: #6715 

Reviewed by: @droideck (Thanks!)

### DIFF
--- a/src/lib389/lib389/replica.py
+++ b/src/lib389/lib389/replica.py
@@ -841,13 +841,11 @@ class NormalizedRidDict(dict):
         nkey = NormalizedRidDict.normalize_rid(key)
         super().__setitem__(nkey, value)
 
-    def get(self, key, *args):
-        if len(args) > 0:
-            try:
-                return self[key]
-            except KeyError:
-                return args[0]
-        return self[key]
+    def get(self, key, vdef=None):
+        try:
+            return self[key]
+        except KeyError:
+            return vdef
 
 
 class RUV(object):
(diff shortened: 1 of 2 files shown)


progier389:
test_ruv_url_not_added_if_different_uuid is still failing! :disappointed:

progier389:
Ok. the problem is that None is expected when there is no default rather than KeyError ...

src/lib389/lib389/replica.py
```diff
@@ -841,6 +841,12 @@ def __setitem__(self, key, value):
         nkey = NormalizedRidDict.normalize_rid(key)
         super().__setitem__(nkey, value)
 
+    def get(self, key, vdef=None):
+        try:
+            return self[key]
+        except KeyError:
+            return vdef
+
```
droideck, line 849: I have a concern that now `.get()` and `.__getitem__()` have different behaviour. One raises a `KeyError`, and another one returns `None/vdef`. 

Personally, I'd change `get()` to (and then, adapt the usage everywhere it's needed):

```
    nkey = NormalizedRidDict.normalize_rid(key)
    return super().get(nkey, vdef)
```

But I don't have a hard opinion here. We can go either way for `NormalizedRidDict`.

progier389: This behavior difference also exists when using plain dict.
I agree that it is a bit confusing as I blindly fall into that pitfall ! :wink: 
But that is the way python implemented dict:

```
[src]$cat dict.py 
d = { 'a':'b' }
print( d.get('k') )
try:
    print( d['k'] )
except KeyError:
    print("d['k'] raised KeyError")
try:
    print( d.__getitem__('k') )
except KeyError:
    print("d.__getitem__('k') raised KeyError")
```
```
[src]$python3 dict.py 
None
d['k'] raised KeyError
d.__getitem__('k') raised KeyError
[src]$
```

droideck: Okay, never mind... 
Looks like it's by design! Oh, well

droideck (approved):
LGTM!

Issue #6501, opened 2025-01-13
dsidm role rename-by-dn does not rename a role


mirielka:
**Version-Release number of the selected component**
389-ds-base-3.0.5-2.el10.x86_64

**Steps to reproduce**
1. Create a role (e.g. managed)
2. Try to rename a role using "dsidm role rename-by-dn"
3. Check for original/renamed role

**Actual results**

Message about successful rename is produced but with the original dn, and the entry still exists only under the original dn.

```
# dsidm localhost -b dc=example,dc=com role create-managed
Enter value for cn : test_role
Successfully created test_role

# dsidm localhost -b dc=example,dc=com role list
test_role

# dsidm localhost -b dc=example,dc=com role get test_role
dn: cn=test_role,dc=example,dc=com
cn: test_role
description: test
objectClass: nsSimpleRoleDefinition
objectClass: nsManagedRoleDefinition
objectClass: nsRoleDefinition
objectClass: ldapSubEntry
objectClass: top

# dsidm localhost -b dc=example,dc=com role rename-by-dn cn=test_role,dc=example,dc=com cn=renamed,dc=example,dc=com
Successfully renamed to cn=test_role,dc=example,dc=com 

# dsidm localhost -b dc=example,dc=com role list
test_role

# dsidm localhost -b dc=example,dc=com role get test_role
dn: cn=test_role,dc=example,dc=com
cn: test_role
description: test
objectClass: nsSimpleRoleDefinition
objectClass: nsManagedRoleDefinition
objectClass: nsRoleDefinition
objectClass: ldapSubEntry
objectClass: top

# dsidm localhost -b dc=example,dc=com role get-by-dn cn=renamed,dc=example,dc=com
Error: No such object
```

**Expected results**
Role should have new cn and dn.

**Additional information**
Output is the same when using the keep-old-rdn option - no changes are made.

PR #6748, opened 2025-04-14
Issue 6501 - CLI - dsidm role rename was not working

mreynolds389:
Description:

dsidm role rename says it's successful but nothing is changed. This is because the wrong object is used when performing the rename

Relates: https://github.com/389ds/389-ds-base/issues/6501

### DIFF
--- a/src/lib389/lib389/cli_idm/role.py
+++ b/src/lib389/lib389/cli_idm/role.py
@@ -75,7 +75,7 @@ def delete(inst, basedn, log, args, warn=True):
     dn = _get_dn_arg(args.dn, msg="Enter dn to delete")
     if warn:
         _warn(dn, msg="Deleting %s %s" % (SINGULAR.__name__, dn))
-    _generic_delete(inst, basedn, log.getChild('_generic_delete'), Role, dn, args)
+    _generic_delete(inst, basedn, log.getChild('_generic_delete'), SINGULAR, dn, args)
 
 
 def modify(inst, basedn, log, args, warn=True):
--- a/dirsrvtests/tests/suites/clu/dsidm_role_test.py
+++ b/dirsrvtests/tests/suites/clu/dsidm_role_test.py
@@ -547,7 +547,6 @@ def test_dsidm_role_modify_by_dn(topology_st, create_test_managed_role):
     assert not test_role.present('description', 'modified managed role')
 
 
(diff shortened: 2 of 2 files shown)


src/lib389/lib389/cli_idm/role.py
```diff
@@ -85,7 +85,7 @@ def modify(inst, basedn, log, args, warn=True):
 
 def rename(inst, basedn, log, args, warn=True):
     dn = _get_dn_arg(args.dn, msg="Enter dn to modify")
-    _generic_rename_dn(inst, basedn, log.getChild('_generic_rename_dn'), MANY, dn, args)
+    _generic_rename_dn(inst, basedn, log.getChild('_generic_rename_dn'), Roles, dn, args)
```
droideck: I'm a bit confused. `MANY == Roles` - at least, it should be, as the approach is used everywhere in our CLI.

mreynolds389: > I'm a bit confused. `MANY == Roles` - at least, it should be, as the approach is used everywhere in our CLI.

Well this has been needed to fix many other bugs around this area.  It comes down to setting "self._protected = False" otherwise with "MANY" deletes are silently ignored as protected remains True for these objects

mreynolds389: Sorry, and "renames", not just "deletes", also check this "protected" variable and will silently be ignored when set to True.

droideck: What I mean is that the change in `src/lib389/lib389/cli_idm/role.py` doesn't change anything, unless I miss something...

In the beginning of the file:
```
MANY = Roles
SINGULAR = Role
```

And this PR doesn't change anything around `protected`.
But the `xfail` removal makes sense to me.

droideck (approved):
LGTM

src/lib389/lib389/cli_idm/role.py
mreynolds389: Ahhh ok let me double check it, I had written this particular fix before other PR's (around Roles) got merged.

mreynolds389: Yeah this issue got "accidentally" fixed in a previous PR, so the original change here was never needed.  Revised PR, and will merge.  Thanks!

PR #6749, opened 2025-04-15
Issue 6734 - BUG - format strings may not contain backslash

### LINKED ISSUE #6734: CLI - dsidm should have option to list DN's

Firstyear:
Bug Description: Format strings may not contain backslashes in older versions of python 3 (3.6/3.9 for example).

Fix Description: pre-format the strings.

fixes: https://github.com/389ds/389-ds-base/issues/6734

Review by: ???

### DIFF
--- a/src/lib389/lib389/_mapped_object.py
+++ b/src/lib389/lib389/_mapped_object.py
@@ -1301,8 +1301,9 @@ class DSLdapObjects(DSLogging, DSLints):
             raise ldap.NO_SUCH_OBJECT(f"No object exists given the filter criteria: {criteria} {search_filter}")
         if len(results) > 1:
             entry_dn = [e.dn for e in results]
+            entry_dns_pretty = '\n    '.join(entry_dn)
             raise ldap.UNWILLING_TO_PERFORM(f"Too many objects matched selection criteria: {criteria} {search_filter}"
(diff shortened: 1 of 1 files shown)


vashirov (approved):
Build no longer fails on EL9, thanks!

src/lib389/lib389/_mapped_object.py
```diff
@@ -1301,8 +1301,9 @@ def get(self, selector=[], dn=None, json=False):
             raise ldap.NO_SUCH_OBJECT(f"No object exists given the filter criteria: {criteria} {search_filter}")
         if len(results) > 1:
             entry_dn = [e.dn for e in results]
+            entry_dns_pretty = '\n    '.join(entry_dn);
```
droideck: Looks like there's an accidental semicolon (`;`) - probably just muscle memory from another language! ;)
It works fine but doesn't match Python's usual style. :)

droideck (approved):
Otherwise, the fix looks good!

src/lib389/lib389/_mapped_object.py
Firstyear: I wonder what other language that could be .... ;)

PR #6750, opened 2025-04-17
Issue 4989 - Confusing error message from dsconf plugin set --enabled

### LINKED ISSUE #4989: Confusing error message from dsconf plugin set --enabled
**Issue Description**
I was trying to enable/disable plugins that are not part of a list from `dsconf localhost plugin --help`. In particular, I needed to disable plugin 7-bit check:
```
# dsconf localhost plugin set "7-bit check" --enabled off
Successfully changed the cn=7-bit check,cn=plugins,cn=config
```
If I run it again, I get a confusing error message:
```
# dsconf localhost plugin set "7-bit check" --enabled off
Error: There is nothing to set in the cn=7-bit check,cn=plugins,cn=config plugin entry
```
I think we can do a better error message saying that the plugin is already enabled/disabled.

droideck:
Description: Enhance the generic_object_edit function to detect when a plugin is already in the requested enabled/disabled state and provide a more helpful error message.
Check specifically if the plugin state was changed and report the result.

Fixes: https://github.com/389ds/389-ds-base/issues/4989

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/cli_conf/__init__.py
+++ b/src/lib389/lib389/cli_conf/__init__.py
@@ -127,13 +127,8 @@ def generic_object_edit(dsldap_object, log, args, arg_to_attr):
         log.info("Successfully changed the %s", dsldap_object.dn)
     else:
         # Check if we're trying to enable/disable a plugin and it's already in that state
-        enabled_attr = None
-        for arg, attr in arg_to_attr.items():
-            if arg == 'enabled' and attr in unchanged_attrs:
-                enabled_attr = attr
-                break
-
-        if enabled_attr is not None:
+        enabled_attr = arg_to_attr.get('enabled', None)
+        if enabled_attr in unchanged_attrs:
             # Get the current state for a more informative message
             current_state = ensure_list_str(existing_attributes[enabled_attr])[0]
             if current_state.lower() == 'on':
--- a/src/lib389/lib389/cli_conf/__init__.py
+++ b/src/lib389/lib389/cli_conf/__init__.py
@@ -106,6 +106,8 @@ def generic_object_edit(dsldap_object, log, args, arg_to_attr):
     existing_attributes = dsldap_object.get_all_attrs()
 
     modlist = []
+    unchanged_attrs = {}
+
     for attr, value in attrs.items():
         # Delete the attribute only if the user set it to 'delete' value
         if value in ("delete", ["delete"]):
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/cli_conf/__init__.py
```diff
@@ -116,11 +118,30 @@ def generic_object_edit(dsldap_object, log, args, arg_to_attr):
...
+                # Track attributes that didn't change for better error messages
+                unchanged_attrs[attr] = value
+
     if len(modlist) > 0:
         dsldap_object.apply_mods(modlist)
         log.info("Successfully changed the %s", dsldap_object.dn)
     else:
-        raise ValueError("There is nothing to set in the %s plugin entry" % dsldap_object.dn)
+        # Check if we're trying to enable/disable a plugin and it's already in that state
+        enabled_attr = None
```
tbordaz: If this portion of code is to detect unchanged 'nsslapd-pluginEnabled' why not directly checking this name in unchanged_attrs ?
Else you may need to test 'arg in ['enabled','disabled'].

droideck: > Else you may need to test 'arg in ['enabled','disabled'].

The change is specifically for `generic_object_edit`, which has `--enabled` attr (`nsslapd-pluginEnabled`). I don't think we have `nsslapd-pluginDisabled` attr:)

> If this portion of code is to detect unchanged 'nsslapd-pluginEnabled' why not directly checking this name in unchanged_attrs ?

It's more of a style decision, I guess?
I'd like to avoid using attribute names in CLI directly, so the following loop logic handles args and the associated attributes in a "shielded" fashion. It just feels "cleaner" to me.
```
       for arg, attr in arg_to_attr.items():
            if arg == 'enabled' and attr in unchanged_attrs:
                enabled_attr = attr
                break
```

It can also be extended in the future for other attributes if needed (even though it's unlikely for this portion of code).

tbordaz: I agree. Thanks

tbordaz (approved):
LGTM

src/lib389/lib389/cli_conf/__init__.py
```diff
@@ -116,11 +118,30 @@ def generic_object_edit(dsldap_object, log, args, arg_to_attr):
...
+        for arg, attr in arg_to_attr.items():
+            if arg == 'enabled' and attr in unchanged_attrs:
```
progier389: Iterating on dict items then testing equality on the key looks pretty unusual.
Why not getting directly the value from the key: (using get() or []) as usual ?
  attr = arg_to_attr.get('enabled', None)
  if attr in unchanged_attrs:
                enabled_attr = attr
  or:
    with contextlib.suppress(KeyError):
        if arg_to_attr['enabled'] in unchanged_attrs:
            enabled_attr =  arg_to_attr['enabled']

droideck: Good point! Fixed

progier389 (approved):
LGTM

Issue #6464, opened 2024-12-25
Fix spelling in code and text in UI and CLI


droideck:
**Issue Description**

A number of spelling errors exist in the **cockpit 389-console** project, particularly in messages displayed to users. [[typos-cli](https://github.com/crate-ci/typos)](https://github.com/crate-ci/typos) can be utilized to locate and correct these spelling mistakes. 

Code typos:
- One notable typo, `"Interractive"`, appears throughout the code and messages;

- In the following files, the command-line switch `pwdmincatagories` should be changed to `pwdmincategories`:

   ```
   src/cockpit/389-console/src/lib/database/localPwp.jsx
   src/lib389/lib389/pwpolicy.py
   src/lib389/lib389/cli_conf/pwpolicy.py
   ```

vashirov:
a6a631042..0cee387e8  389-ds-base-3.0 -> 389-ds-base-3.0
d2900a3d3..ed0b749b2  389-ds-base-2.6 -> 389-ds-base-2.6
f25a8b72a..c87f50efa  389-ds-base-2.5 -> 389-ds-base-2.5
94acbe701..c9f11b0f7  389-ds-base-2.4 -> 389-ds-base-2.4
3c910b69a..a1dcd97fc  389-ds-base-2.3 -> 389-ds-base-2.3
9a2c85244..8669799de  389-ds-base-2.2 -> 389-ds-base-2.2

Issue #6276, opened 2024-07-26
Schema lib389 object is not keeping custom schema data upon editing


droideck:
Description of problem:
While editing a custom schema attribute, lib389 and Web UI is not keeping some defined data:

1) if the "Description" is updated, the "OID" is reverted to "<ATTRIBUTE_NAME>-oid"

2) if the "OID" is updated, then the "Description" field shows "No description".

How reproducible:
Always.

Steps to Reproduce:
1. Create a custom attribute ( Schema tab / Attributes tab / "Add Attribute" button )
2. Edit the newly created attribute and change either the description or the OID
3. Save
4. Only the updated field is kept. The non-updated field is reverted to its default value.

Actual results:
Custom datum is lost.

Expected results:
Keep all custom values.

Additional info:
1) A workaround is to update both fields at the same time:

Set the two fields ( Description and OID ) to dummy values and save.
Edit the attribute again
Set the two fields to the correct values and save
2) JS Console:

```
CMD: cmdOperationAttribute: Do the replace operation on Attribute ==> dsconf -j ldapi://%2fvar%2frun%2fslapd-localhost.socket schema attributetypes replace customAttr --syntax 1.3.6.1.4.1.1466.115.121.1.15 --single-value --user-mod --oid 1.2.3.4

CMD: cmdOperationAttribute: Do the replace operation on Attribute ==> dsconf -j ldapi://%2fvar%2frun%2fslapd-localhost.socket schema attributetypes replace customAttr --syntax 1.3.6.1.4.1.1466.115.121.1.15 --single-value --user-mod --desc customAttr_Desc
```

droideck:
bcc61cdda..c9b3141b9  389-ds-base-3.0 -> 389-ds-base-3.0
   e633e341b..10fb16c04  389-ds-base-2.5 -> 389-ds-base-2.5
   07686abcc..ac27dec0b  389-ds-base-2.4 -> 389-ds-base-2.4
   e3226057d..141fd5514  389-ds-base-2.3 -> 389-ds-base-2.3
   0a3eeac6e..6da56381f  389-ds-base-2.2 -> 389-ds-base-2.2
   b6523e37e..c63847924  389-ds-base-2.1 -> 389-ds-base-2.1
   4a13e4d7b..9bcf4ae9e  389-ds-base-2.0 -> 389-ds-base-2.0
   09fb5006c..9a4748b4c  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

[...]

PR #6751, opened 2025-04-17
Issue 6276 - UI - schema editing and memberof shared config not working correctly

mreynolds389:
Description:

While editing a custom schema attribute the UI is not keeping some defined data (issue 6276).

Fixed issue with memberOf shared config area not being properly handled when added and deleted.

Also fixed a type with "interractive" (issue 6464)

Relates: https://github.com/389ds/389-ds-base/issues/6276
Relates: https://github.com/389ds/389-ds-base/issues/6464

### DIFF
--- a/src/cockpit/389-console/src/lib/plugins/memberOf.jsx
+++ b/src/cockpit/389-console/src/lib/plugins/memberOf.jsx
@@ -822,34 +822,40 @@ class MemberOf extends React.Component {
                 configAllBackends ? "on" : "off",
                 "--skipnested",
                 configSkipNested ? "on" : "off",
-                "--autoaddoc",
-                configAutoAddOC || action === "add" ? configAutoAddOC : "delete",
             ];
 
             // Delete attributes if the user set an empty value to the field
-            cmd = [...cmd, "--scope"];
             if (configEntryScope.length !== 0) {
+                cmd = [...cmd, "--scope"];
                 for (const value of configEntryScope) {
                     cmd = [...cmd, value];
                 }
-            } else {
-                cmd = [...cmd, "delete"];
+            } else if (action !== "add") {
+                cmd = [...cmd, "--scope", "delete"];
             }
-            cmd = [...cmd, "--exclude"];
+
+            if (configAutoAddOC !== "") {
+                cmd = [...cmd, "--autoaddoc", configAutoAddOC];
+            } else if (action !== "add") {
+                cmd = [...cmd, "--autoaddoc", "delete"];
+            }
+
             if (configEntryScopeExcludeSubtree.length !== 0) {
+                cmd = [...cmd, "--exclude"];
                 for (const value of configEntryScopeExcludeSubtree) {
                     cmd = [...cmd, value];
                 }
-            } else {
-                cmd = [...cmd, "delete"];
+            } else if (action !== "add") {
+                cmd = [...cmd, "--exclude", "delete"];
             }
-            cmd = [...cmd, "--groupattr"];
+
             if (configGroupAttr.length !== 0) {
+                cmd = [...cmd, "--groupattr"];
                 for (const value of configGroupAttr) {
                     cmd = [...cmd, value];
                 }
-            } else {
-                cmd = [...cmd, "delete"];
+            } else if (action !== "add") {
+                cmd = [...cmd, "--groupattr", "delete"];
             }
 
             this.setState({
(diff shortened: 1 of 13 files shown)


src/lib389/lib389/cli_conf/plugins/memberof.py
```diff
@@ -160,11 +163,11 @@ def create_parser(subparsers):
     fixup.add_argument('--wait', action='store_true',
                        help="Wait for the task to finish, this could take a long time")
     fixup.add_argument('--timeout', type=int, default=0,
-                        help="Sets the task timeout. ,Default is 0 (no timeout)")
+                       help="Sets the task timeout. ,Default is 0 (no timeout)")
```
vashirov:
```suggestion
                       help="Sets the task timeout. Default is 0 (no timeout)")
```

vashirov (approved):
There is one more typo:
```
src/cockpit/389-console/src/lib/monitor/monitorTables.jsx:1780:                const pwInteractive = rowCopy.pwInputInterractive;
```
Besides these 2 small issues, LGTM!

src/cockpit/389-console/src/lib/plugins/memberOf.jsx
```diff
@@ -1638,10 +1644,9 @@ class MemberOf extends React.Component {
                             <GridItem className="ds-left-margin" span={3}>
                                 <Button
                                     variant="primary"
-                                    isDisabled={memberOfConfigEntry === "" || !valid_dn(memberOfConfigEntry)}
```
droideck, line 1641: I think we should still validate the DN here. It'll allow to avoid situation, when we already have a `Shared Config Entry` but then the user enters an invalid `Shared Config Entry` DN and then presses `Manage` button.

But the case probably should be like this (`!==` instead of `===`):

```js
isDisabled={(memberOfConfigEntry !== "" && !valid_dn(memberOfConfigEntry))}
```

droideck (approved):
The rest LGTM!

src/cockpit/389-console/src/lib/plugins/memberOf.jsx
mreynolds389: Heh, whoops :-)

mreynolds389: Actually I made the text field to always be disabled - forcing the user to use the modal to create, edit, and delete the shared config entry.  I think it's more intuitive and removes the chance to use it incorrectly.

droideck: What I mean, if you first Create the entry, and then back on the main screen will try to modify the `Shared Config Entry` it will allow to Manage Invalid DN

droideck: image
image

mreynolds389: I don't allow the user to enter/change a value on the main page anymore.

droideck: > I don't allow the user to enter/change a value on the main page anymore.

Sounds good.
I'm a bit concerned that it'll make the logic in other `Shared Config Entry` different which is a bit confusing. Should we then change all of the  `Shared Config Entry`  field logic? I think it's only `Account Policy` plugin, though.

Another thing, if we always disable the field, then it doesn't make sense to have it as an input field as it's less readable. Probably, it can be changed to `Text`? I don't have a hard opinion here, though.

mreynolds389: > > I don't allow the user to enter/change a value on the main page anymore.
> 
> Sounds good. I'm a bit concerned that it'll make the logic in other `Shared Config Entry` different which is a bit confusing. Should we then change all of the `Shared Config Entry` field logic? I think it's only `Account Policy` plugin, though.

Yes we should make then all the same, and it's an easy change, so I'll do that quick.

> 
> Another thing, if we always disable the field, then it doesn't make sense to have it as an input field as it's less readable. Probably, it can be changed to `Text`? I don't have a hard opinion here, though.

Yeah this crossed my mind, I thought the "disabled input" was done in other areas for read-only values but it was changed at point.  So I'll redo this as well.

mreynolds389: @droideck ok made all three plugins (account policy, MO, and RI) all behave and look thd same now.  Please review

mreynolds389:
@droideck I think I fixed everything.  I also had to do lib389 updates for account policy and RI plugins.  I removed the ability to change the config entry DN - so that should address the rename issue.  Note - I don't think cn=config supports modrdn.  So a rename would have to be a delete/add.  I also did not see any issues when adding a config entry fails.

vashirov:
Regarding this use case https://github.com/389ds/389-ds-base/issues/6276#issuecomment-2792415891
It's fixed in WebUI where the existing values are passed to the dsconf command. But not in CLI. Would it be possible to make it work there as well?

vashirov:
Ah, ok. In that case we should document how to edit attributes correctly. Currently we only cover creation of a new schema attribute. I'll open a doc bug.

mreynolds389:
> Ah, ok. In that case we should document how to edit attributes correctly. Currently we only cover creation of a new schema attribute. I'll open a doc bug.

Yeah while it would be nice to rework how we can edit schema, it would be a big change to how its been working for years.  Maybe if we got complaints about it we could look into revising it, but I think for now just updating the docs is best.

droideck:
LGTM!

PR #6754, opened 2025-04-23
Port ticket 47560 test

mirielka:
Port test related to ticket 47560 to use DSLdapObject.
Move the test from dirsrvtests/tests/tickets/ticket47560_test.py to dirsrvtests/tests/suites/memberof_plugin/fixup_test.py


dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+    assert not group.present('member', member.dn)
+
+    _enable_disable_mbo('on')
+
+    log.info('Verify the memberOf attribute is still present')
+    assert member.present('memberof', group.dn)
+    assert not group.present('member', member.dn)
+
+    log.info('Run the fixup task')
+    topology_st.standalone.tasks.fixupMemberOf(suffix=DEFAULT_SUFFIX, args={TASK_WAIT: True})
```
droideck: We shouldn't use the old API.
This is the way:
https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/plugins.py#L1050

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+    group = groups.get(group_name)
+
+    users = nsUserAccounts(topology_st.standalone, DEFAULT_SUFFIX)
+    member = users.get(member_name)
+
+    log.info('Add nsMemberOf objectclass to member entry')
+    member.set('objectclass', 'nsMemberOf', ldap.MOD_ADD)
+
+    log.info('Add memberOf attribute to member entry')
+    member.set('memberOf', group.dn, ldap.MOD_ADD)
```
droideck: I think `member.add` will be a shorter and a nicer way here:
https://github.com/389ds/389-ds-base/blob/main/src/lib389/lib389/_mapped_object.py#L338

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+import time
 from lib389.plugins import MemberOfPlugin
-from lib389.idm.user import UserAccounts
+from lib389.idm.user import UserAccounts, nsUserAccounts
 from lib389.idm.group import Groups
+from lib389 import DEFAULT_SUFFIX
+from lib389._constants import *
+from lib389.properties import *
+from lib389.topologies import topology_st
+from lib389.utils import *
```
droideck: If it can be avoided, please, don't import everything (`__all__`) with `*`.

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+            topology_st.standalone.plugins.enable(name=PLUGIN_MEMBER_OF)
+        else:
+            topology_st.standalone.plugins.disable(name=PLUGIN_MEMBER_OF)
+
+        log.info("MemberOf plugin: {}".format(value))
+
+        topology_st.standalone.stop(timeout=120)
+        time.sleep(1)
+        topology_st.standalone.start(timeout=120)
+        time.sleep(3)
```
droideck: I'm not sure I understand the purpose of this code snippet... Why not just a restart?

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+
+        # need to reopen a connection toward the instance
+        topology_st.standalone.open()
```
droideck: It is also part of restart (generally).

droideck (changes requested):
The PR also needs a proper commit message.

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
```diff
@@ -10,17 +10,146 @@
...
+            test_group.delete()
+        log.info('Delete member user')
+        if member.exists():
+            member.delete()
+
+    request.addfinalizer(fin)
+
+
+def test_memberof_fixup_task(topology_st, prepare_entries):
+    """ Test memberOf fixup task
```
tbordaz, line 71: I think the description of the test can be more specific. It checks that memberof fixup task can repair an entry (member) that is erroneously set as member of a group.

mirielka:
Thanks for reviews, most of these were copies from original test that I failed to identify as obsolete (old API) or no longer used (imports).
Commit message is present, but failed to copy into the issue description.

droideck (approved):
For the commit message, we also specify the subject like this:

```
Issue 6753 - Port ticket 47560 test
```

And generally, we have a word 'Description: ' in the commit body.
But yeah, it's minor:)

I think Thierry's comment is still not fixed (I unresolved it).
But besides that - LGTM!

dirsrvtests/tests/suites/memberof_plugin/fixup_test.py
mirielka: Renamed the test

Issue #6727, opened 2025-04-04
Compact interval timer should be persistent


mreynolds389:
** Description **

Currently the compaction interval is set at server startup and we just start a slapi_eq_once_rel().  The default interval is 30 days, so if you restart your server every 29 days the database/changelog will never compact.

We need to write the "start time" into the config so that we know where to pick up from after a restart

mreynolds389:
6069d2ce0..f18581e06  389-ds-base-3.0 -> 389-ds-base-3.0
0fbd56357..a6e4ae9b1  389-ds-base-2.6 -> 389-ds-base-2.6
a415141f8..9fcbe7d04  389-ds-base-2.5 -> 389-ds-base-2.5
9c9bcfb89..fcd60d63b  389-ds-base-2.4 -> 389-ds-base-2.4

PR #6760, opened 2025-04-29
Issue 6727 - RFE - database compaction interval should be persistent

mreynolds389:
Description:

Record when we start the compaction interval in the ldbm config entry so if we restart the server compaction will still take place at the intended interval

Relates: https://github.com/389ds/389-ds-base/issues/6727

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
@@ -805,6 +805,32 @@ done:
 }
 
 static void *
+bdb_config_db_compactdb_starttime_get(void *arg)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+
+    return (void *)((uintptr_t)(BDB_CONFIG(li)->bdb_compactdb_starttime));
+}
+
+static int32_t
+bdb_config_db_compactdb_starttime_set(void *arg,
+                                      void *value,
+                                      char *errorbuf __attribute__((unused)),
+                                      int phase __attribute__((unused)),
+                                      int apply)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+    int32_t retval = LDAP_SUCCESS;
+    uint64_t val = (uint64_t)((uintptr_t)value);
+
+    if (apply) {
+        BDB_CONFIG(li)->bdb_compactdb_starttime = val;
+    }
+
+    return retval;
+}
+
+static void *
 bdb_config_db_page_size_get(void *arg)
 {
     struct ldbminfo *li = (struct ldbminfo *)arg;
@@ -1600,6 +1626,7 @@ static config_info bdb_config_param[] = {
     {CONFIG_DB_CHECKPOINT_INTERVAL, CONFIG_TYPE_INT, "60", &bdb_config_db_checkpoint_interval_get, &bdb_config_db_checkpoint_interval_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_COMPACTDB_INTERVAL, CONFIG_TYPE_INT, "2592000" /*30days*/, &bdb_config_db_compactdb_interval_get, &bdb_config_db_compactdb_interval_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_COMPACTDB_TIME, CONFIG_TYPE_STRING, "23:59", &bdb_config_db_compactdb_time_get, &bdb_config_db_compactdb_time_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
+    {CONFIG_DB_COMPACTDB_STARTTIME, CONFIG_TYPE_UINT64, "0" , &bdb_config_db_compactdb_starttime_get, &bdb_config_db_compactdb_starttime_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_TRANSACTION_BATCH, CONFIG_TYPE_INT, "0", &bdb_get_batch_transactions, &bdb_set_batch_transactions, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_TRANSACTION_BATCH_MIN_SLEEP, CONFIG_TYPE_INT, "50", &bdb_get_batch_txn_min_sleep, &bdb_set_batch_txn_min_sleep, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_TRANSACTION_BATCH_MAX_SLEEP, CONFIG_TYPE_INT, "50", &bdb_get_batch_txn_max_sleep, &bdb_set_batch_txn_max_sleep, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
@@ -3801,27 +3801,10 @@ bdb_compact(time_t when, void *arg)
         slapi_log_err(SLAPI_LOG_NOTICE, "bdb_compact", "Compacting DB start: %s\n",
                       inst->inst_name);
 
-        rc = bdb_db_compact_one_db(db, inst);
-        if (rc) {
-            slapi_log_err(SLAPI_LOG_ERR, "bdb_compact",
-                          "Failed to compact id2entry for %s; db error - %d %s\n",
-                          inst->inst_name, rc, db_strerror(rc));
-            break;
-        }
-
         /* Time to compact the DB's */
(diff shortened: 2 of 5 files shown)


ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.h
```diff
@@ -91,8 +91,9 @@ typedef struct bdb_config
     int bdb_previous_lock_config;  /* Max lock count when we last shut down--
                                       * used to determine if we delete the mpool */
     u_int32_t bdb_deadlock_policy; /* i.e. the atype to DB_ENV->lock_detect in bdb_deadlock_threadmain */
-    int bdb_compactdb_interval;    /* interval to execute compact id2entry dbs */
-    char *bdb_compactdb_time;       /* time of day to execute compact id2entry dbs */
+    int bdb_compactdb_interval;       /* interval to execute compact id2entry dbs */
+    char *bdb_compactdb_time;         /* time of day to execute compact id2entry dbs */
+    int32_t bdb_compactdb_starttime; /* the time the interval was started */
```
droideck: Should we have it as int64_t right away to avoid 2038 issue? Or, probably, the correct option will be a `time_t`, unless I miss something.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -3847,6 +3847,41 @@ bdb_start_checkpoint_thread(struct ldbminfo *li)
...
+ * a delayed event because at server startup the checkpoint thread can run
+ * before all the plugins have been started which causes invalid memory reads.
+ */
+static void
+bdb_write_compact_start_time(time_t when, void *arg)
+{
+    struct ldbminfo *li = (struct ldbminfo *)arg;
+    Slapi_PBlock *mod_pb = slapi_pblock_new();
+    Slapi_Mods smods;
+    char start_time_str[12] = {0};
```
droideck: I don't think 12 is enough for `uint64_t`. I think it's 20.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -4010,9 +4074,13 @@ bdb_checkpoint_threadmain(void *param)
...
-            /* reset interval timer */
-            compactdb_interval = compactdb_interval_update;
-            slapi_timespec_expire_at(compactdb_interval, &compactdb_expire);
+            /* reset compact interval timer */
+            compactdb_interval_orig = compactdb_interval_update;
+            slapi_timespec_expire_at(compactdb_interval_update, &compactdb_expire);
+            /* lastly update the config */
+            compactdb_start_time = slapi_current_utc_time();
+            slapi_eq_once_rel(bdb_write_compact_start_time, (void *)li,
+                              slapi_current_rel_time_t() + 3);
```
droideck, line 4077: I'm not very proficient with BDB, but this 3-second delay seems a bit strange to me.
Why exactly 3 seconds? And if it's a precautioon, can we do it any other way? (like check when the BDB is ready or something with some 100ms retries?)

mreynolds389: > I'm not very proficient with BDB, but this 3-second delay seems a bit strange to me. Why exactly 3 seconds? And if it's a precautioon, can we do it any other way? (like check when the BDB is ready or something with some 100ms retries?)

I'm not sure if there is an easy way to tell if all the plugins have been started from inside the check pointing thread.  The race condition I am avoiding usually all happens within the same second.  But the mod update doesn't need to happen right away, so I figured waiting 3 seconds was a safe bet

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -3891,11 +3928,31 @@ bdb_checkpoint_threadmain(void *param)
...
+        time_t curr_time = slapi_current_utc_time();
+        if (compactdb_interval < (curr_time - compactdb_start_time)) {
+            /* the interval has now been passed, trigger compaction right away */
+            compactdb_interval = 1;
+        } else {
+            compactdb_interval = compactdb_interval - (curr_time - compactdb_start_time);
+        }
+    }
+
     /* assumes bdb_force_checkpoint worked */
```
tbordaz, line 3939: Something not clear to me. Now compactdb_interval is not read from config and only set if compact_db_start_time is set. So there is a possibility that compactdb_interval==0. So it will trigger a compaction immediately in the first checkpointing. Is that expected ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
```diff
@@ -1600,6 +1626,7 @@ static config_info bdb_config_param[] = {
     {CONFIG_DB_CHECKPOINT_INTERVAL, CONFIG_TYPE_INT, "60", &bdb_config_db_checkpoint_interval_get, &bdb_config_db_checkpoint_interval_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_COMPACTDB_INTERVAL, CONFIG_TYPE_INT, "2592000" /*30days*/, &bdb_config_db_compactdb_interval_get, &bdb_config_db_compactdb_interval_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
     {CONFIG_DB_COMPACTDB_TIME, CONFIG_TYPE_STRING, "23:59", &bdb_config_db_compactdb_time_get, &bdb_config_db_compactdb_time_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
+    {CONFIG_DB_COMPACTDB_STARTTIME, CONFIG_TYPE_UINT64, "0" , &bdb_config_db_compactdb_starttime_get, &bdb_config_db_compactdb_starttime_set, CONFIG_FLAG_ALWAYS_SHOW | CONFIG_FLAG_ALLOW_RUNNING_CHANGE},
```
tbordaz, line 1629: Is it required that 'starttime' can be updated (allow running change) ? Not a big concern but that means that admin can change this operational attribute and simulate any starting time (difficult to test)

tbordaz (commented):
Patch looks good to me. Just a minor question

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: compactdb_interval is still read from the config:

    3931:  compactdb_interval_orig = compactdb_interval = (time_t)BDB_CONFIG(li)->bdb_compactdb_interval;

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
mreynolds389: I guess I assumed we need that flag set so we can internally change it.  Sorry didn't look at this part of the change that closely.  Yeah if its not needed then we can drop that flag.

mreynolds389: Ok confirmed we do need this set to allow running change.  Otherwise this will block the internal modify from happening:

    [07/May/2025:10:45:47.637629125 -0400] - ERR - bdb_config_set - nsslapd-db-compactdb-starttime can't be modified while the server is running.
    [07/May/2025:10:45:47.646675737 -0400] - ERR - bdb_write_compact_start_time - failed to modify config_entry, err=53

mreynolds389:
@tbordaz @droideck while working on the CI test I found that it looks like we were calling bdb_db_compact_one_db() before and after we called bdb_db_compact().  However, in bdb_do_compact() we were already calling bdb_db_compact_one_db() for eqch backend.  So we were calling bdb_db_compact_one_db() a total of 3 times for the same exact backend.  Not sure why as it didn't compact any additional pages, so I removed those extra calls to bdb_db_compact_one_db().  Just wanted to mention this as I'd like that change reviewed.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_config.c
tbordaz: Thanks

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -3801,27 +3801,10 @@ bdb_compact(time_t when, void *arg)
         slapi_log_err(SLAPI_LOG_NOTICE, "bdb_compact", "Compacting DB start: %s\n",
                       inst->inst_name);
 
-        rc = bdb_db_compact_one_db(db, inst);
```
tbordaz, line 3804: This part of code was done when CL was integrated into the main DB ([here](https://github.com/389ds/389-ds-base/commit/0f1ab5f0531eca6af07aeda59bf4c8f97f59357f#diff-e8c324db8e63fb5573f5be42df0c6adfae82cffa6851bad15c3ad029fae233a6R3784) ). At that time it is said to first compact id2entry and the second one to compact the CL.
It would worth verifying that with that change, compaction id2entry/CL are still done.

mreynolds389: > This part of code was done when CL was integrated into the main DB ([here](https://github.com/389ds/389-ds-base/commit/0f1ab5f0531eca6af07aeda59bf4c8f97f59357f#diff-e8c324db8e63fb5573f5be42df0c6adfae82cffa6851bad15c3ad029fae233a6R3784) ). At that time it is said to first compact id2entry and the second one to compact the CL. It would worth verifying that with that change, compaction id2entry/CL are still done.

It does report that pages are freed from userroot and changelog

tbordaz (approved):
LGTM

Issue #6761, opened 2025-05-01
Password modify extended operation should skip password policy checks when executed by root DN


mreynolds389:
**Issue Description**

When the LDAP password policy extended operation is executed by root DN on a regular user under constraints of a password policy, eg password history, it should skip password policy checks because the root DN should be allowed to set a regular user password to any value however this is currently not the case.

The root DN user can still modify userPassword attribute directly via regular modify operation, which can be used as workaround of sorts however it does not play nice with various password policy flags/op attributes ie it can be done but requires more effort and essentially negates password policy extended operation functionality.

Steps to Reproduce:

1. Have a password policy in place with a specific constraint eg password history.
2. Use ldappasswd tool, bind as root DN, attempt to change user's password to one that is already in history (violate password policy constraints).

Actual results:

Result: Constraint violation (19)
Additional info: Failed to update password

Expected results:

The root DN should be able to violate password policy constraints and change user password regardless.

Additional info:

It should probably apply to a password administrator user as well if one is defined for password policy configuration.

PR #6762, opened 2025-05-01
Issue 6761 - Password modify extended operation should skip password policy checks when executed by root DN

mreynolds389:
Description:

When the LDAP password policy extended operation is executed by root DN on a regular user under constraints of a password policy, eg password history, it should skip password policy checks because the root DN should be allowed to set a regular user password to any value however this is currently not the case.

The fix is to check if the bind is a pwd admin and set SLAPI_REQUESTOR_ISROOT for the new internal operation that will update the password. This does not bypass aci evalaution.

relates: https://github.com/389ds/389-ds-base/issues/6761

### DIFF
--- a/ldap/servers/slapd/modify.c
+++ b/ldap/servers/slapd/modify.c
@@ -1374,6 +1374,15 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
             rc = -1;
             goto done;
         }
+    } else if (pw_is_pwp_admin(pb, pwpolicy, PWP_ADMIN_OR_ROOTDN)) {
+        /* This is an internal operation, but we still need to check if this
+         * is a password admin */
+        if (!SLAPI_IS_MOD_DELETE(mod->mod_op) && pwpolicy->pw_history) {
+            /* Updating pw history, get the old password */
+            get_old_pw(pb, &sdn, old_pw);
+        }
+        rc = 1;
+        goto done;
     }
 
     /* check if password is within password minimum age;
--- a/ldap/servers/slapd/passwd_extop.c
+++ b/ldap/servers/slapd/passwd_extop.c
@@ -120,6 +120,7 @@ passwd_apply_mods(Slapi_PBlock *pb_orig, const Slapi_DN *sdn, Slapi_Mods *mods, 
 {
     LDAPControl **req_controls_copy = NULL;
     LDAPControl **pb_resp_controls = NULL;
+    passwdPolicy *pwpolicy = NULL;
     int ret = 0;
 
     slapi_log_err(SLAPI_LOG_TRACE, "passwd_apply_mods", "=>\n");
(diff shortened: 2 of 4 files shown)


ldap/servers/slapd/passwd_extop.c
```diff
@@ -132,6 +133,14 @@ passwd_apply_mods(Slapi_PBlock *pb_orig, const Slapi_DN *sdn, Slapi_Mods *mods,
         }
 
         Slapi_PBlock *pb = slapi_pblock_new();
+        pwpolicy = new_passwdPolicy(pb_orig, slapi_sdn_get_ndn(sdn));
+        if (pw_is_pwp_admin(pb_orig, pwpolicy, PWP_ADMIN_OR_ROOTDN)) {
+            /* If this is root DN or password admin set is_requestor_root so
+             * set this in the new pblock so password updates are correctly
+             * applied */
+            int is_requestor_root = 1;
+            slapi_pblock_set(pb, SLAPI_REQUESTOR_ISROOT, &is_requestor_root);
```
tbordaz, line 142: Weird, IIUC pw_is_pwp_admin return True if pblock[SLAPI_REQUESTOR_ISROOT] is already 1. Is that setting required ?

ldap/servers/slapd/modify.c
```diff
@@ -1323,6 +1323,15 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
             rc = -1;
             goto done;
         }
+    } else if (pw_is_pwp_admin(pb, pwpolicy, PWP_ADMIN_OR_ROOTDN)) {
```
tbordaz, line 1377: This is called for direct operation only. Should not we call for both internal/direct ?
Possibly checking that this is
<pre>
slapi_pblock_get(pb, SLAPI_EXT_OP_REQ_OID, &oid);
strcasecmp(oid, EXTOP_PASSWD_OID);
</pre>

ldap/servers/slapd/passwd_extop.c
mreynolds389: > Weird, IIUC pw_is_pwp_admin return True if pblock[SLAPI_REQUESTOR_ISROOT] is already 1. Is that setting required ?

It is the easiest way to set the "new" pblock used for the mod that updates the password.  So we check the "current/old" pblock to see if it's a password admin, if it is then we set the SLAPI_REQUESTOR_ISROOT to 1 in the "new" pblock so the password is updated as expected

tbordaz: oppsss.. my bad, I read too fast and missed there were two pblocks. Ignore my comment

ldap/servers/slapd/modify.c
mreynolds389: @tbordaz Sorry I'm not sure what you are wanting to do here.  The code does check both direct/internal ops.  Do you want to make sure the internal op is definitely a "password modify extended op" before checking if it's a password admin?
 
    +    } else if (strcmp(oid, EXTOP_PASSWD_OID) == 0 && pw_is_pwp_admin(pb, pwpolicy, PWP_ADMIN_OR_ROOTDN)) {

droideck (approved):
Looks good to me!

ldap/servers/slapd/modify.c
tbordaz: Sorry my remark was really confusing.
What I meant is minor comment. In the else branch this is for internal operation. Should we bother testing ROOT_DN ? for example acl are ignored for internal op so we could just assume that an internal op is done by root.
If test is preferable then  could we also check the EXTOP oid ?

mreynolds389: @tbordaz Ok the OID is not passed back to the server in this case.  I did confirm we need to check if it is a password admin regardless if its an internal operation because a non-root DN user will still trigger the internal operation.  

Anyway I'm not sure I see a need to check the OID, and trying to set it is proving difficult - looks like it getting stripped along the way and I'm not sure it's worth pursuing as I don't see a need for it.

tbordaz: Okay. Thanks for your explanations. This is fine to me

tbordaz (approved):
LGTM

PR #6765, opened 2025-05-05
Issue 6764 - statistics about index lookup report a wrong duration

### LINKED ISSUE #6764: statistics about index lookup report a wrong duration
**Issue Description**
When nsslapd-statlog-level is set to '1' it reports statistics about indexes lookup. A statistic is the duration of the lookup. This duration is wrong as it only reports the duration of the last index lookup.
Indeed, a filter contains several components and the duration should be displayed for each of them.

 - since 2.4 ( https://github.com/389ds/389-ds-base/issues/3729)

**Steps to Reproduce**
set nsslapd-statlog-level=1 and do a search with multiple components. check that a single duration is reported

**Expected results**
We should report a duration per component

tbordaz:
Bug description:
	During a SRCH statistics about indexes lookup
	(when nsslapd-statlog-level=1) reports a duration.
	It is wrong because it should report a duration per filter
	component.

Fix description:
	Record a index lookup duration per key
	using key_lookup_start/key_lookup_end

fixes: #6764

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/back-ldbm/filterindex.c
+++ b/ldap/servers/slapd/back-ldbm/filterindex.c
@@ -1105,11 +1105,14 @@ keys2idl(
         struct component_keys_lookup *key_stat;
         int key_len;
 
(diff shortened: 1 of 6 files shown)


progier389 (approved):
LGTM

PR #6766, opened 2025-05-06
Issue 6755 - RFE use of Session Tracking Control in replication agree…

### LINKED ISSUE #6755: RFE use of Session Tracking Control in replication agreement
This RFE is the client side of RFE #6367 .
To help debugging of replication the replication agreements need to log a unique identifier both onreplication supplier side (in error logs) and on the consumer side (in access logs). 

Design to come

nonr

**Additional context**
To enable client side logging the server side must implement #6367

tbordaz:
…ment

Bug description:
	Replication agreement can use Session Tracking Control
	to correlated replication sessions between suppliers
	and consumers.
	Since https://github.com/389ds/389-ds-base/issues/6367 the
	server sides (consumer) logs session tracking control in
	access logs.
	This ticket is to implement the client side (supplier)

Fix description:
	This is described  https://www.port389.org/docs/389ds/design/session-identifier-clients.html
	Each replication agreement implement a session identifier
	string that is sent for each replicated operation
	including extended op.
	The session identifier is logged, on client side,
	if the replication debug log is turned on

fixes: #6755

Reviewed by:

### DIFF
--- a/ldap/servers/plugins/replication/repl5_agmt.c
+++ b/ldap/servers/plugins/replication/repl5_agmt.c
@@ -58,6 +58,9 @@
 #include "slapi-plugin.h"
 #include "slap.h"
 #include "../../slapd/back-ldbm/dbimpl.h"          /* for dblayer_is_lmdb */
+#include <pk11func.h>
+#include <pk11pqg.h>
+#include <plbase64.h>
 
 #define DEFAULT_TIMEOUT 120                  /* (seconds) default outbound LDAP connection */
 #define DEFAULT_FLOWCONTROL_WINDOW      1000 /* #entries sent without acknowledgment (bdb) */
@@ -95,6 +98,10 @@ typedef struct repl5agmt
     const Slapi_DN *dn;                    /* DN of replication agreement entry */
     const Slapi_RDN *rdn;                  /* RDN of replication agreement entry */
     char *long_name;                       /* Long name (rdn + host, port) of entry, for logging */
+    int session_id_cnt;                    /* use to differentiate sessions */
+    char *session_id_prefix;               /* use for debugging purpose on server/client sides */
+#define SESSION_ID_STR_SZ 15
+    char session_id[SESSION_ID_STR_SZ + 49];
     Repl_Protocol *protocol;               /* Protocol object - manages protocol */
     struct changecounter **changecounters; /* changes sent/skipped since server start up */
     int64_t num_changecounters;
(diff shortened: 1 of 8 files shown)


ldap/servers/plugins/replication/repl5_inc_protocol.c
```diff
@@ -325,8 +325,8 @@ repl5_inc_result_threadmain(void *param)
 
             conn_get_error_ex(conn, &operation_code, &connection_error, &ldap_error_string);
             slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name,
-                          "repl5_inc_result_threadmain - Result %d, %d, %d, %d, %s\n",
-                          operation_code, connection_error, conres, message_id, ldap_error_string);
+                          "repl5_inc_result_threadmain - sid \"%s\" - Result %d, %d, %d, %d, %s\n",
```
mreynolds389: Nit pick here, but I can you change the format to be:

    sid=\"%s\"

It's cleaner, and easier to parse if needed.  Do this for all the logging calls.

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -180,7 +191,105 @@ nsds5ReplicaBusyWaitTime - time to wait after getting a REPLICA BUSY from the co
...
+void
+agmt_set_session_id(Repl_Agmt *ra)
+{
+    if (ra->session_id_cnt == 999) {
+        ra->session_id_cnt = 1;
+    } else {
+        ra->session_id_cnt++;
+    }
+    strncpy(ra->session_id, ra->session_id_prefix, SESSION_ID_STR_SZ - 4);
+    sprintf(&ra->session_id[SESSION_ID_STR_SZ - 4], " %3d", ra->session_id_cnt);
```
mreynolds389: Why all the padding?

    sid "OBWBnz8jniH   4"

Can't we just use:

    sprintf(&ra->session_id[SESSION_ID_STR_SZ - 4], " %d", ra->session_id_cnt);
   
    sid "OBWBnz8jniH 4"

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -95,6 +98,14 @@ typedef struct repl5agmt
     const Slapi_DN *dn;                    /* DN of replication agreement entry */
     const Slapi_RDN *rdn;                  /* RDN of replication agreement entry */
     char *long_name;                       /* Long name (rdn + host, port) of entry, for logging */
+#define SID_SUPPORT_UNKNOWN 0
+#define SID_SUPPORT_NO      1
+#define SID_SUPPORT_YES     2
```
progier389: Just curious. Why bothering checking if control is supported ?
You could systematically send the control with CRITICAL flag unset and the consumer server will ignore it if it is not supported

tbordaz: Good point I will remove this from the patch

ldap/servers/plugins/replication/repl5_inc_protocol.c
tbordaz: That is right I changed it.

ldap/servers/plugins/replication/repl5_agmt.c
tbordaz: Indeed it was over complex. I fixed it.

ldap/servers/plugins/replication/repl5_agmt.c
```diff
@@ -180,7 +187,87 @@ nsds5ReplicaBusyWaitTime - time to wait after getting a REPLICA BUSY from the co
...
+
+void
+agmt_set_session_id(Repl_Agmt *ra)
+{
+    if (ra->session_id_cnt == 999) {
+        ra->session_id_cnt = 1;
+    } else {
+        ra->session_id_cnt++;
+    }
+    sprintf(ra->session_id, "%s %3d", ra->session_id_prefix, ra->session_id_cnt);
```
mreynolds389, line 263: Looks like you are still padding the ID count, why?

tbordaz: The id count ranges [1..999] and represent an unique outgoing replication connection (RA and cleanallruv). It rollback from 999 to 1 if there are more that 999 connections during an instance life cycle.
In my mind the padding would help debug using grep. The following 'grep' only matches 1rst session
`grep '<sid>   1' errors`
while the following pattern matches many more ('1', '10'..'19', '100'..'199')
`grep '<sid> 1' errors`

mreynolds389: Thanks for the explanation, makes sense now.

mreynolds389 (approved):

PR #6773, opened 2025-05-07
Issue 6772 - dsconf - Replicas with the "consumer" role allow for viewing and modification of their changelog.

### LINKED ISSUE #6772: dsconf - Replicas with the "consumer" role allow for viewing and modification of their changelog.
**Issue Description**
If I understand LDAP correctly, a replica with a consumer role should not have a changelog. Only suppliers and hubs should. However, dsconf get-changelog and set-changelog allow for operations on consumer replicas which can lead to confusing situations, such as a consumer having a changelog with an invalid max age; the changelog, which was not properly created, does not have the validation function callbacks defined in this case. 
It's not a huge problem, but it can lead to some confusion on the end user's side if they don't fully understand the distinction between consumers and suppliers/hubs. 

 - Platform: RHEL 10
 - Package and version: Latest build off main, at this time 3.1.2. 

**Steps to Reproduce**
1. Create an instance using dscreate with a suffix entry. 
2. Enable replication on the instance with the "consumer" role. 
3. Run dsconf replication set-changelog and/or get-changelog on this instance. You'll see that this consumer instance has a changelog. 
 
**Expected results**
I believe the consumer instance should show a message saying that the consumer replica does not support a changelog. 

**CLI Output**
Here is an example of commands and output showing the existence of a changelog on a consumer replica. 
```
# dscreate from-file tmp.inf
Starting installation ...
Validate installation settings ...
Create file system structures ...
Create self-signed certificate database ...
Perform SELinux labeling ...
Create database backend: dc=example,dc=com ...
Perform post-installation tasks ...
Completed installation for instance: slapd-localhost
# dsconf -D "cn=Directory Manager" localhost replication enable --suffix "dc=example,dc=com" --role "consumer" --bind-dn "cn=replication manager,cn=config" --bind-passwd "password"
Replication successfully enabled for "dc=example,dc=com"
# dsconf -D "cn=Directory Manager" localhost replication get-changelog --suffix "dc=example,dc=com"
dn: cn=changelog,cn=userroot,cn=ldbm database,cn=plugins,cn=config
cn: changelog
nsslapd-changelogmaxage: 7d
objectClass: top
objectClass: extensibleObject
```

**Additional context**
I first ran into this issue when trying to set up an instance with improperly configured single supplier replication. https://github.com/389ds/389-ds-base/issues/6752

abeisemb:
dsconf currently allows users to set and retrieve changelogs in consumer replicas, which do not have officially supported changelogs. This can lead to undefined behavior and confusion. This commit prints a warning message if the user tries to interact with a changelog on a consumer replica.

Resolves: https://github.com/389ds/389-ds-base/issues/6772

### DIFF
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -686,6 +686,9 @@ def set_per_backend_cl(inst, basedn, log, args):
     replace_list = []
     did_something = False
 
+    if (is_replica_role_consumer(inst, suffix)):
+        log.info("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
+
     if args.encrypt:
         cl.replace('nsslapd-encryptionalgorithm', 'AES')
         del args.encrypt
@@ -715,6 +718,10 @@ def set_per_backend_cl(inst, basedn, log, args):
 # that means there is a changelog config entry per backend (aka suffix)
 def get_per_backend_cl(inst, basedn, log, args):
     suffix = args.suffix
+
+    if (is_replica_role_consumer(inst, suffix)):
+        log.info("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
+
     cl = Changelog(inst, suffix)
     if args and args.json:
         log.info(cl.get_all_attrs_json())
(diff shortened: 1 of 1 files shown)


abeisemb:
Currently this is a conservative approach: instead of killing the process entirely, I print out a warning message to let the user know there may be unintended behavior such as a changelog with an invalid max-age. 
However, it's worth noting that a consumer replica likely doesn't actually have a changelog initialized during creation, as I discovered when trying to trace through the max-age validation process. With that in mind, a more aggressive approach (i.e. killing the process) may be warranted.

droideck (approved):
LGTM

abeisemb:
I was mistaken: checking out to the commit immediately before mine results in the ds_logs_test.py::test_internal_log_server_level_4 test still failing on a test VM. Additionally, building off of upstream main and running the ds_logs test results in the same assertion failure. 
Whatever's causing this error, it isn't this PR.

PR #6774, opened 2025-05-08
Issue 6768 - ns-slapd crashes when a referral is added

### LINKED ISSUE #6768: ns-slapd crashes when a referral is added
**Issue Description**
ns-slapd crashes when a paged result search is run on a referral suffix

**Steps to Reproduce**
1. Create instance1 with dc=example1,dc=com suffix
2. Create instance2 with dc=example2,dc=com suffix
3. Add a referral from instance1 to instance 2
4. Run a paged result search on instance1 with instance2 suffix as base
5. ns-slapd SEGV

**Expected results**
A referral search result (10) should be returned, or if chase referrals are enabled on the client, then the requested info from instance2 suffix

jchapma:
Bug description: When a paged result search is successfully run on a referred suffix, we retrieve the search result set from the pblock and try to release it. In this case the search result set is NULL, which triggers a SEGV during the release.

Fix description: If the search result code is LDAP_REFERRAL, skip deletion of the search result set. Added test case.

Fixes: https://github.com/389ds/389-ds-base/issues/6768

Reviewed by:

### DIFF
--- a/ldap/servers/slapd/opshared.c
+++ b/ldap/servers/slapd/opshared.c
@@ -875,7 +875,11 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                  * sent a result */
                 rc = SLAPI_FAIL_GENERAL;
                 slapi_pblock_get(pb, SLAPI_RESULT_CODE, &err);
-                if (err == LDAP_NO_SUCH_OBJECT) {
+                if (err == LDAP_REFERRAL) {
+                    /* manage a referral search result code. */
+                    flag_referral++;
+                }
+                else if (err == LDAP_NO_SUCH_OBJECT) {
                     /* may be the object exist somewhere else
                      * wait the end of the loop to send back this error
                      */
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/opshared.c
```diff
@@ -875,7 +875,11 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                  * sent a result */
                 rc = SLAPI_FAIL_GENERAL;
                 slapi_pblock_get(pb, SLAPI_RESULT_CODE, &err);
-                if (err == LDAP_NO_SUCH_OBJECT) {
+                if (err == LDAP_REFERRAL) {
+                    /* manage a referral search result code. */
+                    flag_referral++;
```
tbordaz, line 880: I am not really fan about reusing this variable. Are you sure it has the same purpose as the one you need ?
May be you can create your own variable.

ldap/servers/slapd/opshared.c
```diff
@@ -892,7 +896,7 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
                 slapi_pblock_get(pb, SLAPI_RESULT_CODE, &err);
                 /* PAGED RESULTS */
                 /* Explicit ctrlp test avoid gcc -fanalyzer warning */
-                if (op_is_pagedresults(operation)) {
+                if (op_is_pagedresults(operation) && !flag_referral) {
```
tbordaz, line 899: I have a doubt here. My understanding is that calling be_search_results_release/delete_search_result_set with a NULL result set lead to sigsev. Am I correct ?
Would it be possible to harden delete_search_result_set so that it does not crash ?

PR #6775, opened 2025-05-12
Issue 4253 - Container bind mount schema

### LINKED ISSUE #4253: Container bind mount schema
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/51200
- Created at 2020-07-02 01:04:18 by [firstyear](https://pagure.io/user/firstyear) (@Firstyear)
- Assigned to nobody

---

#### Issue Description
A early adopter of the 389 container noticed that if you bind mount in schema, this causes first run to fail as the "directory already exists". We should probably be more careful about this scenario in our dscontainer script and allow some locaitons to "already exist" if they are known config injection points.

vashirov:
Bug Description:
When you mount a volume with schema into container, instance creation fails with:
```
  File "/usr/libexec/dirsrv/dscontainer", line 489, in <module>
    begin_magic()
    ~~~~~~~~~~~^^
  File "/usr/libexec/dirsrv/dscontainer", line 279, in begin_magic
    if not sds.create_from_args(g2b.collect(), s2b.collect()):
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/lib389/instance/setup.py", line 759, in create_from_args
    self._install_ds(general, slapd, backends)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/lib389/instance/setup.py", line 903, in _install_ds
    _ds_shutil_copytree(os.path.join(slapd['sysconf_dir'], 'dirsrv/schema'), slapd['schema_dir'])
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/lib389/__init__.py", line 230, in _ds_shutil_copytree
    os.makedirs(dst)
    ~~~~~~~~~~~^^^^^
  File "<frozen os>", line 227, in makedirs
FileExistsError: [Errno 17] File exists: '/etc/dirsrv/slapd-localhost/schema'
```

`_ds_shutil_copytree()` is an outdated copy of the `shutil.copytree()` from Python 3.5. In the later versions `dirs_exist_ok` parameter was introduced, but it was never backported to our implementation.

Fix Description:
Instead of the custom implementation, use standard `shutil.copytree()` function, but ignore existing directories (`dirs_exist_ok=True`) and use `copy` instead of `copy2` to ignore file metadata.

Fixes: https://github.com/389ds/389-ds-base/issues/4253

### DIFF
--- a/src/lib389/lib389/instance/setup.py
+++ b/src/lib389/lib389/instance/setup.py
@@ -18,7 +18,7 @@ import socket
 import subprocess
 import getpass
 import configparser
-from lib389 import _ds_shutil_copytree, DirSrv
+from lib389 import DirSrv
 from lib389._constants import *
 from lib389.properties import *
 from lib389.passwd import password_hash, password_generate
@@ -900,7 +900,13 @@ class SetupDs(object):
         #  This is a little fragile, make it better.
         # It won't matter when we move schema to usr anyway ...
 
-        _ds_shutil_copytree(os.path.join(slapd['sysconf_dir'], 'dirsrv/schema'), slapd['schema_dir'])
+        shutil.copytree(
+            os.path.join(slapd["sysconf_dir"], "dirsrv/schema"),
+            slapd["schema_dir"],
+            copy_function=shutil.copy,
+            # Schema directory might be bind mounted, ingore it
+            dirs_exist_ok=True,
+        )
         os.chown(slapd['schema_dir'], slapd['user_uid'], slapd['group_gid'])
         os.chmod(slapd['schema_dir'], 0o770)
 
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -214,61 +214,6 @@ def pid_from_file(pidfile):
         pass
     return pid
 
-def _ds_shutil_copytree(src, dst, symlinks=False, ignore=None, copy_function=copy2,
-             ignore_dangling_symlinks=False):
-    """Recursively copy a directory tree.
-    This is taken from /usr/lib64/python3.5/shutil.py, but removes the
-    copystat function at the end. Why? Because in a container without
-    privileges, we don't have access to set xattr. But copystat attempts to
-    set the xattr when we are root, which causes the copy to fail. Remove it!
-    """
-    names = os.listdir(src)
-    if ignore is not None:
-        ignored_names = ignore(src, names)
-    else:
-        ignored_names = set()
-
-    os.makedirs(dst)
-    errors = []
-    for name in names:
-        if name in ignored_names:
-            continue
-        srcname = os.path.join(src, name)
-        dstname = os.path.join(dst, name)
-        try:
-            if os.path.islink(srcname):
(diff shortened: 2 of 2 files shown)


mreynolds389 (approved):

src/lib389/lib389/instance/setup.py
```diff
@@ -900,7 +900,12 @@ def _install_ds(self, general, slapd, backends):
         #  This is a little fragile, make it better.
         # It won't matter when we move schema to usr anyway ...
 
-        _ds_shutil_copytree(os.path.join(slapd['sysconf_dir'], 'dirsrv/schema'), slapd['schema_dir'])
+        shutil.copytree(
+            os.path.join(slapd["sysconf_dir"], "dirsrv/schema"),
+            slapd["schema_dir"],
+            copy_function=shutil.copy,
+            dirs_exist_ok=True,
```
droideck, line 908: I think it's worth to add a small comment here that `dirs_exist_ok` is for bind-mounted volumes handling.

droideck (approved):
Looks good!

But just to confirm, `dirs_exist_ok` was added in Python 3.8.
Should we state some where that we deprecate all older Python version starting 3.1/3.2? (it might be already true, though, for other reasons)

src/lib389/lib389/instance/setup.py
vashirov: Thanks, added.

vashirov:
> But just to confirm, `dirs_exist_ok` was added in Python 3.8.
> Should we state some where that we deprecate all older Python version starting 3.1/3.2? (it might be already true, though, for other reasons)

We support Python 3.6 on 1.4.3, Python 3.9 on 2.x+.
But you're making a good point. I scanned our lib389 code with [vermin](https://pypi.org/project/vermin/), it found that most of our code is compatible with 3.6, with the following exceptions:
3.7 - https://github.com/389ds/389-ds-base/blob/18b24dc3a52be5a7e17d9d8bde43c84e45a4ca0a/src/lib389/lib389/perftools.py#L401
3.7 - https://github.com/389ds/389-ds-base/blob/18b24dc3a52be5a7e17d9d8bde43c84e45a4ca0a/src/lib389/lib389/repltools.py#L410
3.8 - https://github.com/389ds/389-ds-base/blob/18b24dc3a52be5a7e17d9d8bde43c84e45a4ca0a/src/lib389/lib389/cli_ctl/instance.py#L221
3.11 - https://github.com/389ds/389-ds-base/blob/18b24dc3a52be5a7e17d9d8bde43c84e45a4ca0a/src/lib389/lib389/replica.py#L920

I think the last one needs to be addressed, because we need to backport it to older versions.

Issue #6753, opened 2025-04-23
Port ticket tests


mirielka:
Port ticket tests to use DSLdapObject and to be moved from dirsrvtests/tests/tickets into dirsrvtests/tests/suites/

PR #6777, opened 2025-05-13
Issue 6753 - Remove ticket test 47573

mirielka:
Description:
The tests in dirsrvtests/tests/tickets/ticket47573_test.py are duplicated by tests in dirsrvtests/tests/suites/schema/schema_replication_test.py, therefore the ticket test is being removed entirely. Also removing a few deprecation warnings caused directly by remaining test code.

Relates: #6753

Author: Lenka Doudova

Reviewer: ???


droideck (changes requested):
I don't think we should remove the test.
It checks for the case when an attribute is moved from MAY to MUST.

Unless I miss something, this scenario is not tested in the `schema_replication_test.py`.

mirielka:
> I don't think we should remove the test. It checks for the case when an attribute is moved from MAY to MUST.
> 
> Unless I miss something, this scenario is not tested in the `schema_replication_test.py`.

Unless I'm mistaken the ticket test code tests the other way around - it moved MUST attribute to MAY. Comments in the code describe it the wrong way though. The original ticket also puts it as an issue with moving MUST attribute to MAY.

droideck:
> Unless I'm mistaken the ticket test code tests the other way around - it moved MUST attribute to MAY. Comments in the code describe it the wrong way though. The original ticket also puts it as an issue with moving MUST attribute to MAY.

Right... I think it has some issues with what is written in the comments and the code (and the intention of the code).
Still, my original thought, I think, is still correct. We shouldn't remove the test as it has its unique value that is not represented in schema_replication_test.

mirielka:
> > Unless I'm mistaken the ticket test code tests the other way around - it moved MUST attribute to MAY. Comments in the code describe it the wrong way though. The original ticket also puts it as an issue with moving MUST attribute to MAY.
> 
> Right... I think it has some issues with what is written in the comments and the code (and the intention of the code). Still, my original thought, I think, is still correct. We shouldn't remove the test as it has its unique value that is not represented in schema_replication_test.

Ah yes, my bad, I mixed it up! Will rework the test.

tbordaz:
> > Unless I'm mistaken the ticket test code tests the other way around - it moved MUST attribute to MAY. Comments in the code describe it the wrong way though. The original ticket also puts it as an issue with moving MUST attribute to MAY.
> 
> Right... I think it has some issues with what is written in the comments and the code (and the intention of the code). Still, my original thought, I think, is still correct. We shouldn't remove the test as it has its unique value that is not represented in schema_replication_test.

@droideck , unless I'm mistaken 'test_ticket47573_one' is done in 'test_schema_replication_one' and 'test_ticket47573_two' is done in 'test_schema_replication_four'. So IIUC the two testcases can be removed.
What do you mean 'it has its unique value not represented' ?

dirsrvtests/tests/suites/schema/schema_replication_test.py
```diff
@@ -675,22 +692,117 @@ def test_schema_replication_nine(topology_m1c1, schema_replication_init):
...
     regex = re.compile(r"must not be overwritten \(set replication log for additional info\)")
     res = pattern_errorlog(topology_m1c1.ms["supplier1"].errorlog_file, regex)
-    if res is not None:
+    assert res is None
+
+    log.info('Testcase PASSED')
+
+
+def test_move_must_attribute_to_may(topology_m1c1, schema_replication_init):
+    """ Check that if a MUST attribute is moved to MAY on supplier, but remains enchanged on the consumer,
```
droideck:
```suggestion
    """ Check that if a MUST attribute is moved to MAY on supplier, but remains unchanged on the consumer,
```

droideck (approved):
Besides the minor-minor issue, looks good!

PR #6780, opened 2025-05-13
Issue 6768 - ns-slapd crashes when a referral is added

### LINKED ISSUE #6768: ns-slapd crashes when a referral is added

jchapma:
Bug description: When a paged result search is successfully run on a referred suffix, we retrieve the search result set from the pblock and try to release it. In this case the search result set is NULL, which triggers a SEGV during the release.

Fix description: If the search result code is LDAP_REFERRAL, skip deletion of the search result set. Added test case.

Fixes: https://github.com/389ds/389-ds-base/issues/6768

Reviewed by: @progier389, @tbordaz

### DIFF
--- a/dirsrvtests/tests/suites/paged_results/paged_results_test.py
+++ b/dirsrvtests/tests/suites/paged_results/paged_results_test.py
@@ -1271,6 +1271,52 @@ def test_search_stress_abandon(create_40k_users, create_user):
         paged_search(conn, create_40k_users.suffix, [req_ctrl], search_flt, searchreq_attrlist, abandon_rate=abandon_rate)
 
 
+def test_search_referral(topology_st):
+    """Test a paged search on a referred suffix doesnt crash the server.
+
+    :id: c788bdbf-965b-4f12-ac24-d4d695e2cce2
+
+    :setup: Standalone instance
+
+    :steps:
+         1. Configure a default referral.
+         2. Create a paged result search control.
+         3. Paged result search on referral suffix (doesnt exist on the instance, triggering a referral).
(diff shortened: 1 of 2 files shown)


jchapma:
I opened the initial PR on upstream instead of my fork, correcting this

jchapma:
@tbordaz, in a previous comment you said:
I have a doubt here. My understanding is that calling be_search_results_release/delete_search_result_set with a NULL result set lead to sigsev. Am I correct ?
Would it be possible to harden delete_search_result_set so that it does not crash ?

When I looked into modifying be_search_results_release() to handle a NULL search result set, I discovered that this function pointer was indeed NULL. I understand this is because we are using the default backend for a referral. So the fix is now just to check be_search_results_release != NULL.

jchapma:
Rebase.

Hi @tbordaz , @progier389 

You both reviewed the original PR  for this issue  (https://github.com/389ds/389-ds-base/pull/6774). But I incorrectly opened on upstream instead of my fork, so opened this PR.  This PR is hanging around for a while now, it would be great if you review. 

Thank you

progier389 (approved):
LGTM

Issue #6782, opened 2025-05-15
Improve paged result locking


mreynolds389:
**Issue Description**

pagedresults_lock() and pagedresults_unlock() are reading from a critical section (prp->pr_mutex) without holding necessary lock.  

There is also a deadlock possibility when checking for abandoned PR search in opshared.c

mreynolds389:
f4a089f6..681a16f7c  389-ds-base-3.0 -> 389-ds-base-3.0
90e16e9e9..9e3fc6409  389-ds-base-2.6 -> 389-ds-base-2.6
dea349a42..44be1ab4b  389-ds-base-2.5 -> 389-ds-base-2.5
1a194ef32..eb25463a3  389-ds-base-2.4 -> 389-ds-base-2.4
6b8a0adad..d82f03602  389-ds-base-2.3 -> 389-ds-base-2.3
185f0b49d..80e963660  389-ds-base-2.2 -> 389-ds-base-2.2
f5ccc9437..74262f2ce  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6783, opened 2025-05-15
Issue 6782 - Improve paged result locking

mreynolds389:
Description:

pagedresults_lock() and pagedresults_unlock() are reading from a critical section (prp->pr_mutex) without holding necessary lock.

There is also a deadlock possibility when checking for abandoned PR search in opshared.c

Relates: https://github.com/389ds/389-ds-base/issues/6782

### DIFF
--- a/ldap/servers/slapd/opshared.c
+++ b/ldap/servers/slapd/opshared.c
@@ -619,6 +619,14 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
             int32_t tlimit;
             slapi_pblock_get(pb, SLAPI_SEARCH_TIMELIMIT, &tlimit);
             pagedresults_set_timelimit(pb_conn, operation, (time_t)tlimit, pr_idx);
+            /* When using this mutex in conjunction with the main paged
+             * result lock, you must do so in this order:
+             *
+             * --> pagedresults_lock()
+             *    --> pagedresults_mutex
+             *    <-- pagedresults_mutex
+             * <-- pagedresults_unlock()
+             */
             pagedresults_mutex = pageresult_lock_get_addr(pb_conn);
         }
 
@@ -744,11 +752,11 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
             pr_search_result = pagedresults_get_search_result(pb_conn, operation, 1 /*locked*/, pr_idx);
             if (pr_search_result) {
                 if (pagedresults_is_abandoned_or_notavailable(pb_conn, 1 /*locked*/, pr_idx)) {
+                    pthread_mutex_unlock(pagedresults_mutex);
                     pagedresults_unlock(pb_conn, pr_idx);
                     /* Previous operation was abandoned and the simplepaged object is not in use. */
                     send_ldap_result(pb, 0, NULL, "Simple Paged Results Search abandoned", 0, NULL);
                     rc = LDAP_SUCCESS;
-                    pthread_mutex_unlock(pagedresults_mutex);
                     goto free_and_return;
                 } else {
                     slapi_pblock_set(pb, SLAPI_SEARCH_RESULT_SET, pr_search_result);
--- a/ldap/servers/slapd/pagedresults.c
+++ b/ldap/servers/slapd/pagedresults.c
@@ -48,7 +48,6 @@ pageresult_lock_get_addr(Connection *conn)
 static void
 _pr_cleanup_one_slot(PagedResults *prp)
 {
-    PRLock *prmutex = NULL;
     if (!prp) {
         return;
     }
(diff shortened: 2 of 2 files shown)


ldap/servers/slapd/pagedresults.c
```diff
@@ -1018,10 +1018,10 @@ pagedresults_lock(Connection *conn, int index)
     }
     pthread_mutex_lock(pageresult_lock_get_addr(conn));
     prp = conn->c_pagedresults.prl_list + index;
-    pthread_mutex_unlock(pageresult_lock_get_addr(conn));
     if (prp->pr_mutex) {
         PR_Lock(prp->pr_mutex);
     }
+    pthread_mutex_unlock(pageresult_lock_get_addr(conn));
```
progier389: A closer look confirms my unease about that one:
There is a deadlock risk:
  The page search per connection mutex is now locked then the prp mutex
  but op_shared_search locks them in the opposite order:
  	    pagedresults_lock(pb_conn, pr_idx);                   ===> The prp mutex 
            pthread_mutex_lock(pagedresults_mutex);        ===> The per connection mutex

ldap/servers/slapd/opshared.c
```diff
@@ -744,11 +744,11 @@ op_shared_search(Slapi_PBlock *pb, int send_result)
             pr_search_result = pagedresults_get_search_result(pb_conn, operation, 1 /*locked*/, pr_idx);
             if (pr_search_result) {
                 if (pagedresults_is_abandoned_or_notavailable(pb_conn, 1 /*locked*/, pr_idx)) {
+                    pthread_mutex_unlock(pagedresults_mutex);
```
progier389, line 755: Nice catch !

progier389 (commented):
_pr_cleanup_one_slot zeroes the prp->mutex then restore it.
I wonder if it is it possible that it somehow get called while the prp is still used.
 (and having a thread that has not fully finished its work and released the mutex)
 I wonder if we should not avoid overwriting the mutex ...

ldap/servers/slapd/pagedresults.c
mreynolds389: I'm not seeing this.  pagedresults_lock() releases the per_conn lock before it returns.  So you have to call pagedresults_lock() before taking the per_conn mutex.  Everywhere I checked this is the order that is used:

pagedresults_lock(conn, idx)
  --> pthread_mutex_lock(pagedresults_mutex);
  <-- pthread_mutex_unlock(pagedresults_mutex);
pagedresults_unlock(conn, idx)

Yes pagedresults_lock/pagedresults_unlock takes the per_conn mutex while it locks/unlock the prp_mutex, but it shouldn't deadlock.  I'm probably missing the scenario you are worried about, can you point out where this is happening?

mreynolds389:
> _pr_cleanup_one_slot zeroes the prp->mutex then restore it. I wonder if it is it possible that it somehow get called while the prp is still used. (and having a thread that has not fully finished its work and released the mutex) I wonder if we should not avoid overwriting the mutex ...

Interesting observation, and that would be an easy fix.  I'll look into this...

ldap/servers/slapd/pagedresults.c
mreynolds389: Ok I see how it can happen.  Thread 1 has prp lock, but doesn't have the per_conn lock yet, thread 2 takes per_conn lock (via pagedresults_lock()) as it tries to get the prp lock and blocks.  Then thread 1 tries to get the per_conn lock and we deadlock.  Ugh, too bad :-(

progier389 (approved):
Last commit looks good.
Just lets hope that it fix the customer issue and that we are not missing some other hole somewhere else 😉

droideck (approved):
Also looks good to me (AFAIU).
I wonder if it'll be beneficial to add a bit more comments around lock functions and explain lock order to prevent future dev accidentally creating deadlocks.

Do I understand correctly that the lock order in the `op_shared_search` section is this:
To get a lock:
1. First acquire pageresult_lock_get_addr(conn) via pagedresults_lock()
2. Then acquire pagedresults_mutex

And when releasing:
1. First release pagedresults_mutex
2. Then release the connection-specific lock via pagedresults_unlock()

progier389:
> Also looks good to me (AFAIU). I wonder if it'll be beneficial to add a bit more comments around lock functions and explain lock order to prevent future dev accidentally creating deadlocks.

I agree it would be a good idea to document the lock order when we know it.
it is difficult to fully order all the locks (there are too many of them and with the plugin callbacks things are not always obvious ...)
(I was sometime surprised to see deadlock between locks from different components that are apparently unrelated)

> 
> Do I understand correctly that the lock order in the `op_shared_search` section is this: To get a lock:
> 
> 1. First acquire pageresult_lock_get_addr(conn) via pagedresults_lock()
> 2. Then acquire pagedresults_mutex

No, it is the opposite: 
  When both locks are needed: 
    the prp mutex is acquired first 
    then the per connection mutex is acquired after
    (as it is done in opshared.c)
    
 When calling pagedresults_lock 
   the per connection mutex is acquired then released  and only after the prp mutex is lock
  (so only a single mutex is locked at any time by this function)   

> 
> And when releasing:
> 
> 1. First release pagedresults_mutex
> 2. Then release the connection-specific lock via pagedresults_unlock()

Note: The release order does not really matter (unlock is not blocking so both lock will be released whatever the unlock order is)
 but yes the locking order is important and should always be the same.
 And order also matters when using pthread_rwlock_rdlock locks because it may block if another thread is trying to acquire the lock in write mode)

progier389:
Thanks for adding the comments. LGTM

Issue #6784, opened 2025-05-16
Improve entry cache LRU to limit large group evictions


progier389:
**Is your feature request related to a problem? Please describe.**
Large group entries (having several thousands of members) often get quickly evicted from the entry cache when memberof plugin is enabled because all its members are visited and that fill the cache.
 Current LRU algorithm evict entries based on how long they were loaded in the cache. So the group is soon getting the oldest loaded entry.
This has several drawbacks:
   - time to load again the entry
   - reloading the entry also mean rebuilding its associate acl group cache 

**Describe the solution you'd like**
  - The idea is to keep a few entries in the cache as long as possible by associating a weight with each entry
     based on their loading time and their number of members
       (i.e: weight = load_time_in_micro_seconds * log²(nb_members) if there is more than 512 members)
       and keep the n first entries having the higher weight 
       (n is configurable with a default value of 10 )

**Describe alternatives you've considered**
     More details is discussed in the design document PR:  https://github.com/389ds/389ds.github.io/pull/20 
     and should ends up in https://www.port389.org/docs/389ds/design/entrycache-eviction.html once the PR get merged.

PR #6785, opened 2025-05-16
Issue 6784 - Support of Entry cache pinned entries

progier389:
Entry cache eviction improvement to limit large group evictions
Design is https://www.port389.org/docs/389ds/design/entrycache-pinned.html
Issue: #6784 

Reviewed by: @mreynolds389 , @tbordaz , @Firstyear  (Thanks!)

## Summary by Sourcery

Implement weighted, pinned-entry cache eviction and associated configuration, refactor cache statistics handling, enhance logging, and add integration tests for eviction behavior.

New Features:
- Add weighted, pinned-entry cache eviction supported by a new pinned-context list and entry weights computed from load time and group size.
- Introduce configuration options nsslapd-cache-pinned-entries and nsslapd-cache-debug-pattern (with regex) for controlling preserved entries and targeted cache logging.
- Expose average entry load time and pinned-entry metrics in cache monitoring output.

Enhancements:
- Replace individual counters and limits with a unified cache_stats structure and simplify cache_init to accept instance context.
- Refactor eviction logic to use cache_stats and pinned-context, with regex-filtered logging via LOGPATTERN and average weight details.
- Allow MDB backend size tuning in topology scripts via DS389_MDB_MAX_SIZE environment variable.

Tests:
- Add entrycache_eviction_test.py integration tests verifying pinned-entry caching and eviction thresholds.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -1,4 +1,4 @@
-/** BEGIN COPYRIGHT BLOCK
+/*
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
  * Copyright (C) 2025 Red Hat, Inc.
  * All rights reserved.
@@ -145,6 +145,7 @@ typedef unsigned short u_int16_t;
 #define DEFAULT_CACHE_SIZE       (uint64_t)0
 #define DEFAULT_CACHE_SIZE_STR   "0"
 #define DEFAULT_CACHE_ENTRIES    -1 /* no limit */
+#define DEFAULT_CACHE_PRESERVED_ENTRIES_STR "10"
 #define DEFAULT_DNCACHE_SIZE     (uint64_t)16777216
 #define DEFAULT_DNCACHE_SIZE_STR "16777216"
 #define DEFAULT_DNCACHE_MAXCOUNT -1 /* no limit */
@@ -341,6 +342,7 @@ struct backentry
     void *ep_id_link;               /*     tables used for */
     void *ep_uuid_link;             /*     looking up entries */
     PRMonitor *ep_mutexp;           /* protection for mods; make it reentrant */
+    uint64_t ep_weight;             /* for cache eviction */
 };
 
 /* From ep_type through ep_create_time MUST be identical to backcommon */
@@ -358,26 +360,39 @@ struct backdn
     void *dn_id_link;               /* for hash table */
 };
 
+/* Entry Cache statistics */
+struct cache_stats
+{
+    uint64_t hits;            /* for analysis of hits/misses */
+    uint64_t tries;
+    uint64_t nentries;        /* current # entries in cache */
+    int64_t  maxentries;      /* max entries allowed (-1: no limit) */
+    uint64_t size;            /* current size in bytes */
+    uint64_t maxsize;         /* max size in bytes */
+    uint64_t weight;          /* Total weight of all entries */
+    uint64_t nehw;            /* current # entries having weight in cache */
+                              /* weight/nehw is the average time in
+                               * microseconds needed to load an entry
+                               * in the cache
+                               */
+};
+
 /* for the in-core cache of entries */
 struct cache
 {
-    uint64_t c_maxsize;       /* max size in bytes */
-    uint64_t c_config_maxsize; /* manually configured value */
-    Slapi_Counter *c_cursize; /* size in bytes */
-    int64_t c_maxentries;     /* max entries allowed (-1: no limit) */
-    int64_t c_config_maxentries; /* manually configured value */
-    uint64_t c_curentries;    /* current # entries in cache */
     Hashtable *c_dntable;
     Hashtable *c_idtable;
 #ifdef UUIDCACHE_ON
     Hashtable *c_uuidtable;
 #endif
-    Slapi_Counter *c_hits; /* for analysis of hits/misses */
-    Slapi_Counter *c_tries;
-    struct backcommon *c_lruhead; /* add entries here */
-    struct backcommon *c_lrutail; /* remove entries here */
+    struct backcommon c_lrus[2];  /* Standard and preserved lru looped double linked lists */
     PRMonitor *c_mutex;           /* lock for cache operations */
+    uint64_t c_config_maxsize;    /* manually configured value */
+    int64_t c_config_maxentries;  /* manually configured value */
     PRLock *c_emutexalloc_mutex;
+    struct cache_stats c_stats;
+    struct ldbm_instance *c_inst;
+    size_t c_nbpreserved;
 };
 
 #define CACHE_ADD(cache, p, a) cache_add((cache), (void *)(p), (void **)(a))
@@ -385,6 +400,11 @@ struct cache
 #define CACHE_REMOVE(cache, p) cache_remove((cache), (void *)(p))
 #define CACHE_LOCK(cache)      cache_lock((cache))
 #define CACHE_UNLOCK(cache)    cache_unlock((cache))
+
+/* For backentry_compute_weight implementation */
+typedef struct timespec BackEntryWeightData;
+
+
 
 /* various modules keep private data inside the attrinfo structure */
 typedef struct dblayer_private     dblayer_private;
@@ -788,6 +808,11 @@ typedef struct ldbm_instance
     int require_internalop_index;    /* set to 1 to require an index be used in an internal search */
     struct cache inst_dncache;       /* The dn cache for this instance. */
     uint32_t inst_page_count;        /* page count used for cache autotuning */
+    int cache_preserved_entries;     /* Number of entries to preserve during cache eviction */
+    char *cache_debug_pattern;       /* Entries whose dn matche this pattern are logged as INFO
+                                      * when they get added/removed from entry cache
+                                      */
+    Slapi_Regex *cache_debug_re;     /* Compiled version of cache_debug_pattern */
 } ldbm_instance;
 
 /*
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -1,4 +1,4 @@
-/*
+/** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
  * Copyright (C) 2025 Red Hat, Inc.
  * All rights reserved.
@@ -145,7 +145,7 @@ typedef unsigned short u_int16_t;
 #define DEFAULT_CACHE_SIZE       (uint64_t)0
 #define DEFAULT_CACHE_SIZE_STR   "0"
 #define DEFAULT_CACHE_ENTRIES    -1 /* no limit */
-#define DEFAULT_CACHE_PRESERVED_ENTRIES_STR "0"
+#define DEFAULT_CACHE_PINNED_ENTRIES_STR "0"
 #define DEFAULT_DNCACHE_SIZE     (uint64_t)16777216
 #define DEFAULT_DNCACHE_SIZE_STR "16777216"
 #define DEFAULT_DNCACHE_MAXCOUNT -1 /* no limit */
@@ -320,6 +320,8 @@ struct backcommon
 #define ENTRY_STATE_CREATING   0x2  /* entry is being created; don't touch it */
 #define ENTRY_STATE_NOTINCACHE 0x4  /* cache_add failed; not in the cache */
 #define ENTRY_STATE_INVALID    0x8  /* cache entry is invalid and needs to be removed */
+#define ENTRY_STATE_UNAVAILABLE 0xf /* entry is not fully created or is deleted */
+#define ENTRY_STATE_PINNED     0x10 /* cache entry is pinned (never removed by the lru) */
     int32_t ep_refcnt;              /* entry reference cnt */
     size_t ep_size;                 /* for cache tracking */
     struct timespec ep_create_time; /* the time the entry was added to the cache */
@@ -369,7 +371,7 @@ struct cache_stats
     int64_t  maxentries;      /* max entries allowed (-1: no limit) */
     uint64_t size;            /* current size in bytes */
     uint64_t maxsize;         /* max size in bytes */
-    uint64_t weight;          /* Total weight of all entries */
+    uint64_t weight;          /* total weight of all entries */
     uint64_t nehw;            /* current # entries having weight in cache */
                               /* weight/nehw is the average time in
                                * microseconds needed to load an entry
@@ -385,14 +387,15 @@ struct cache
 #ifdef UUIDCACHE_ON
     Hashtable *c_uuidtable;
 #endif
-    struct backcommon c_lrus[2];  /* Standard and preserved lru looped double linked lists */
+    struct backcommon *c_lruhead; /* add entries here */
+    struct backcommon *c_lrutail; /* remove entries here */
     PRMonitor *c_mutex;           /* lock for cache operations */
     uint64_t c_config_maxsize;    /* manually configured value */
     int64_t c_config_maxentries;  /* manually configured value */
     PRLock *c_emutexalloc_mutex;
     struct cache_stats c_stats;
     struct ldbm_instance *c_inst;
-    size_t c_nbpreserved;
+    struct pinned_ctx  *c_pinned_ctx; /* Pinned entries handler context */
 };
 
 #define CACHE_ADD(cache, p, a) cache_add((cache), (void *)(p), (void **)(a))
@@ -403,8 +406,6 @@ struct cache
 
 /* For backentry_compute_weight implementation */
 typedef struct timespec BackEntryWeightData;
-
-
 
 /* various modules keep private data inside the attrinfo structure */
 typedef struct dblayer_private     dblayer_private;
@@ -808,7 +809,7 @@ typedef struct ldbm_instance
     int require_internalop_index;    /* set to 1 to require an index be used in an internal search */
     struct cache inst_dncache;       /* The dn cache for this instance. */
     uint32_t inst_page_count;        /* page count used for cache autotuning */
-    int cache_preserved_entries;     /* Number of entries to preserve during cache eviction */
+    int cache_pinned_entries;        /* Number of entries to preserve during cache eviction */
     char *cache_debug_pattern;       /* Entries whose dn matche this pattern are logged as INFO
                                       * when they get added/removed from entry cache
                                       */
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -145,7 +145,7 @@ typedef unsigned short u_int16_t;
 #define DEFAULT_CACHE_SIZE       (uint64_t)0
 #define DEFAULT_CACHE_SIZE_STR   "0"
 #define DEFAULT_CACHE_ENTRIES    -1 /* no limit */
-#define DEFAULT_CACHE_PRESERVED_ENTRIES_STR "10"
+#define DEFAULT_CACHE_PRESERVED_ENTRIES_STR "0"
 #define DEFAULT_DNCACHE_SIZE     (uint64_t)16777216
 #define DEFAULT_DNCACHE_SIZE_STR "16777216"
 #define DEFAULT_DNCACHE_MAXCOUNT -1 /* no limit */
--- a/ldap/servers/slapd/back-ldbm/backentry.c
+++ b/ldap/servers/slapd/back-ldbm/backentry.c
@@ -132,3 +132,47 @@ backdn_init(Slapi_DN *sdn, ID id, int to_remove_from_hash)
 
     return (bdn);
 }
+
+void
+backentry_init_weight(BackEntryWeightData *starttime)
+{
+    clock_gettime(CLOCK_MONOTONIC, starttime);
+}
+
+void
+backentry_compute_weight(struct backentry *e, const BackEntryWeightData *starttime)
+{
+    struct timespec now = {0};
+    struct timespec delta = {0};
+    unsigned int nbmembers = 0;
+    int n = 0;
+    Slapi_Attr *a = NULL;
+
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    slapi_timespec_diff(&now, (BackEntryWeightData*)starttime, &delta);
+    e->ep_weight = delta.tv_sec * 1000000UL + delta.tv_nsec / 1000UL;
+    if (e->ep_weight == 0) {
+        /* Ensure that entries with very small weight can be distinguished
+         * from those without any weight.
+         */
+        e->ep_weight = 1L;
+    }
+    /* Let count the number of members */
+    if (e->ep_entry) {
+        slapi_entry_attr_find(e->ep_entry, "uniquemember", &a);
+        slapi_attr_get_numvalues(a, &n);
+        nbmembers += (unsigned int)n;
+        slapi_entry_attr_find(e->ep_entry, "member", &a);
+        slapi_attr_get_numvalues(a, &n);
+        nbmembers += (unsigned int)n;
+    }
+    /* Compute log²(nbmembers) */
+    for (n=0; nbmembers>0; nbmembers>>=1) n++;
+    /* And increase significantly the loading time for large groups
+     * so that large groups weight get higer than standard entries one
+     * (even if they got a CPU context switch while being loaded)
+     */
+    if ( n > 8 ) {
+        e->ep_weight *= n;
+    }
+}
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -59,7 +59,6 @@
 #define LOG(...)
 #endif
 #define LOGPATTERN(cache, dn, msg, ...) { if (debug_pattern_matches(cache, dn)) { slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, "CACHE DEBUG: " msg, __VA_ARGS__); } }
-#define LOG2(...) slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, __VA_ARGS__)
 
 
 struct pinned_ctx {
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -19,8 +19,6 @@
 #define LDAP_CACHE_DEBUG
 /* #define LDAP_CACHE_DEBUG_LRU * causes slowdown */
 /* #define CACHE_DEBUG * causes slowdown */
-#define LDAP_CACHE_DEBUG_LRU  1
-#define CACHE_DEBUG  1
 #endif
 
 
@@ -633,10 +631,8 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                     entry->ep_refcnt++;
                     if (entry->ep_state & ENTRY_STATE_PINNED) {
                         pinned_remove(cache, laste);
-                        lru_delete(cache, laste);
-                    } else {
-                        lru_delete(cache, laste);
                     }
+                    lru_delete(cache, laste);
                     if (type == ENTRY_CACHE) {
                         entrycache_remove_int(cache, laste);
                         entrycache_return(cache, (struct backentry **)&laste, PR_TRUE);
@@ -681,10 +677,8 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                         entry->ep_refcnt++;
                         if (entry->ep_state & ENTRY_STATE_PINNED) {
                             pinned_remove(cache, laste);
-                            lru_delete(cache, laste);
-                        } else {
-                            lru_delete(cache, laste);
                         }
+                        lru_delete(cache, laste);
                         entrycache_remove_int(cache, laste);
                         entrycache_return(cache, (struct backentry **)&laste, PR_TRUE);
                     } else {
@@ -730,7 +724,7 @@ cache_init(struct cache *cache, struct ldbm_instance *inst, uint64_t maxsize, in
     cache->c_lruhead = cache->c_lrutail = NULL;
     cache_make_hashes(cache, type);
     cache->c_pinned_ctx = (struct pinned_ctx*)slapi_ch_calloc(1, sizeof (struct pinned_ctx));
-    
+
     if (((cache->c_mutex = PR_NewMonitor()) == NULL) ||
         ((cache->c_emutexalloc_mutex = PR_NewLock()) == NULL)) {
         slapi_log_err(SLAPI_LOG_ERR, "cache_init", "PR_NewMonitor failed\n");
@@ -1185,7 +1179,7 @@ pinned_remove(struct cache *cache, void *ptr)
 }
 
 /* Ensure pinned entries respects the cache memory and maxentrie limits
- * May put entries back in the lru so this function should be called 
+ * May put entries back in the lru so this function should be called
  * just before calling entrycache_flush
  */
 void
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -1081,7 +1081,7 @@ entrycache_remove_int(struct cache *cache, struct backentry *e)
 
     LOGPATTERN(cache, backentry_get_ndn(e),
                "Cache average weight is %lu . Removing entry from "
-               " cache with size: %lu, weight: %lu, dn:%s\n",
+               "cache with size: %lu, weight: %lu, dn:%s\n",
                AV_WEIGHT(cache), e->ep_size, e->ep_weight,
                backentry_get_ndn(e));
     LOG("=> entrycache_remove_int (%s) (%u) (%u)\n", backentry_get_ndn(e), e->ep_id, e->ep_refcnt);
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -17,9 +17,10 @@
 
 #ifdef DEBUG
 #define LDAP_CACHE_DEBUG
-/* #define LDAP_CACHE_DEBUG_LRU_EVICTION * causes slowdown */
 /* #define LDAP_CACHE_DEBUG_LRU * causes slowdown */
 /* #define CACHE_DEBUG * causes slowdown */
+#define LDAP_CACHE_DEBUG_LRU  1
+#define CACHE_DEBUG  1
 #endif
 
 
@@ -51,29 +52,32 @@
     } while (0)
 
 #define LOG(...) slapi_log_err(SLAPI_LOG_CACHE, (char *)__func__, __VA_ARGS__)
-// #define LOG2(msg, ...) { if (cache->c_inst->cache_debug_re) slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, "CACHE DEBUG: " msg, __VA_ARGS__); }
+//# define LOG(_a, _x1, _x2, _x3)  slapi_log_err(SLAPI_LOG_CACHE, _a, _x1, _x2, _x3)
 #else
 #define ASSERT(_x) ;
 //#define LOG(_a, _x1, _x2, _x3)  ;
 #define LOG(...)
 #endif
-#define LOGPATTERN(cache, dn, msg, ...) { if (debug_pattern_matches(cache, dn)) slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, "CACHE DEBUG: " msg, __VA_ARGS__); }
+#define LOGPATTERN(cache, dn, msg, ...) { if (debug_pattern_matches(cache, dn)) { slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, "CACHE DEBUG: " msg, __VA_ARGS__); } }
+#define LOG2(...) slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, __VA_ARGS__)
+
+
+struct pinned_ctx {
+    struct backentry *head;
+    struct backentry *tail;
+    uint64_t npinned;
+    uint64_t size;
+};
+
 
 typedef enum {
     ENTRY_CACHE,
     DN_CACHE,
 } CacheType;
 
-typedef enum {
-    LRU_REGULAR,
-    LRU_PRESERVED,
-} LRUQueueType;
-
 #define LRU_DETACH(cache, e) lru_detach((cache), (void *)(e))
-#define CACHE_LRU_Q(cache, lruid) (&(cache)->c_lrus[lruid])
-#define CACHE_LRU_EOQ(cache, lruid, e) (((struct backcommon *)(e)) == CACHE_LRU_Q(cache, lruid))
-#define CACHE_LRU_HEAD(cache, lruid, type) ((type)(CACHE_LRU_Q(cache, lruid)->ep_lrunext))
-#define CACHE_LRU_TAIL(cache, lruid, type) ((type)(CACHE_LRU_Q(cache, lruid)->ep_lruprev))
+#define CACHE_LRU_HEAD(cache, type) ((type)((cache)->c_lruhead))
+#define CACHE_LRU_TAIL(cache, type) ((type)((cache)->c_lrutail))
 #define BACK_LRU_NEXT(entry, type) ((type)((entry)->ep_lrunext))
 #define BACK_LRU_PREV(entry, type) ((type)((entry)->ep_lruprev))
 
@@ -87,10 +91,7 @@ static int entrycache_add_int(struct cache *cache, struct backentry *e, int stat
 static struct backentry *entrycache_flush(struct cache *cache);
 static bool debug_pattern_matches(struct cache *cache, const char *dn);
 #ifdef LDAP_CACHE_DEBUG_LRU
-static void entry_lru_verify(struct cache *cache, struct backentry *e, bool in, LRUQueueType lruid);
-#endif
-#ifdef LDAP_CACHE_DEBUG_LRU_EVICTION
-static void debug_verify_entrycache_lru(int level, struct cache *cache, LRUQueueType lruid, int lineno);
+static void entry_lru_verify(struct cache *cache, struct backentry *e, int in);
 #endif
 
 static int dn_same_id(const void *bdn, const void *k);
@@ -102,8 +103,10 @@ static int dncache_replace(struct cache *cache, struct backdn *olddn, struct bac
 static int dncache_add_int(struct cache *cache, struct backdn *bdn, int state, struct backdn **alt);
 static struct backdn *dncache_flush(struct cache *cache);
 static int cache_is_in_cache_nolock(void *ptr);
+void pinned_remove(struct cache *cache, void *ptr);
+void pinned_flush(struct cache *cache);
 #ifdef LDAP_CACHE_DEBUG_LRU
-static void dn_lru_verify(struct cache *cache, struct backdn *dn, bool in);
+static void dn_lru_verify(struct cache *cache, struct backdn *dn, int in);
 #endif
 
 /***** tiny hashtable implementation *****/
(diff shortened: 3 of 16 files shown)


Firstyear:
fyi the design doc link is dead mate :)

Firstyear (dismissed):
The risk here is that this is going to make it *extremely* difficult to evict groups from the cache due to the weighting. 

I think that perhaps the approach here may be wrong - LRU is known to have issues with situations like these, it's prone to churn. Rather than LRU, it's probably better to change algorithm to ARC which is resistant to behaviour like this.

progier389:
> The risk here is that this is going to make it _extremely_ difficult to evict groups from the cache due to the weighting.

That is the point:
Keeping the large groups (until they got modified) in the cache is the intended goal of this feature.
And the number of preserved entry is limited and very small (10 entries by default) and is configurable
(most groups will get in the the regular LRU and handled as it is today ...)

> 
> I think that perhaps the approach here may be wrong - LRU is known to have issues with situations like these, it's prone to churn. Rather than LRU, it's probably better to change algorithm to ARC which is resistant to behaviour like this.

I do not think that adaptative replacement cache really solves the problem:
The problem that we are trying to solve is not that some entries are more frequently used than other but that some entries are much more costly to load than others ...

But anyway I will update the design document to mention this alternative ...

Firstyear:
> > The risk here is that this is going to make it _extremely_ difficult to evict groups from the cache due to the weighting.
> 
> That is the point: Keeping the large groups (until they got modified) in the cache is the intended goal of this feature. And the number of preserved entry is limited and very small (10 entries by default) and is configurable (most groups will get in the the regular LRU and handled as it is today ...)

Yes I know it's the point, but that itself creates issues because *after* this operation is complete, you want your cache to be able to evict these entries, and use that memory for other active tasks. Second, some operations may require *more* than 10 entries for their operation to complete efficiently. This is why investing in a better algorithm in the cache will be more effective than trying to hack in a fix like this.

> 
> > I think that perhaps the approach here may be wrong - LRU is known to have issues with situations like these, it's prone to churn. Rather than LRU, it's probably better to change algorithm to ARC which is resistant to behaviour like this.
> 
> I do not think that adaptative replacement cache really solves the problem: The problem that we are trying to solve is not that some entries are more frequently used than other but that some entries are much more costly to load than others ...

But that is fundamentally the point of a cache is that things are expensive to load, so we want to avoid loading them. So ARC does solve this problem, because LRU is not resistant to the type of scanning and eviction patterns you describe here, where ARC *is* resistant and so would not evict the items. ARC also can handle when you would need more than 10 entries available for the operation, and then evicting them post operation as well. 

I think this approach has a risk of reducing the effectiveness of the cache by causing items to be pinned and unable to be evicted long term, which will negatively impact other operations, and the small size of the pinned set may mean that this does not solve the problem for memberof in other environments.

progier389:
I have the feeling that we are speaking about really different problems (Sorry, the problem description was probably not clear enough!)
The point we want to address is not so much that entries get evicted during memberof. 
    (anyway if the group is huge there is no way that we can load up all its member entries in the cache. So eviction and potential reload will happen)
We are trying to limit the fact that some entries are much more expensive to load in the cache than most of the other ones
  (because of the cost of normalizing then sorting thousands of DN to create the member sorted valueset)

| you want your cache to be able to evict these entries
No. That is exactly that we do not want! We want to dedicate part of memory to huge groups 
 and avoid having to reload these one.

| Second, some operations may require more than 10 entries for their operation to complete efficiently
So ? These entries will go in regular queue and handled by the LRU as it is currently done.
I hope that no operation requires to load 10 huge groups and anyway if it is the case that only means that the administrator should adjust the 10 limit to something larger. (remember that the value is configurable and could even be set to 0 to disable the feature if needed. 10 is really an arbitrary (and arguable) default value. I wonder if the default should not depends on the cache size. Not sure we want to preserve any entries on tiny containers for example)

| But that is fundamentally the point of a cache is that things are expensive to load, so we want to avoid loading them. 
True.
| So ARC does solve this problem
False.
 ARC is based of access frequency and has no knowledge of how expensive it is to load an entry in the cache.
 So an entry that is accessing 10 time less often than another one will be evicted even if its loading time is 100 time higher

I do not say that using ARC instead of LRU is a bad idea. (I agree with you that it could probably improve things)
 I only say it does not solve the point I was trying to solve (avoiding to have to reload very expensive entries)
 And BTW separating a huge groups entries from the others is not incompatible with using ARC
 (in other word, in the future, nothing prevent us to replace the LRU algorithm used for the regular queue by ARC)

Firstyear:
> I have the feeling that we are speaking about really different problems (Sorry, the problem description was probably not clear enough!) The point we want to address is not so much that entries get evicted during memberof. (anyway if the group is huge there is no way that we can load up all its member entries in the cache. So eviction and potential reload will happen) We are trying to limit the fact that some entries are much more expensive to load in the cache than most of the other ones (because of the cost of normalizing then sorting thousands of DN to create the member sorted valueset)

No I know exactly what the problem is. 

The problem is *two fold*. You have solved "how do we make sure that the entry we WANT in the cache is in the cache". I am telling you that your solution breaks another property - that is "Once we are *finished* with that operation, how can we use the cache for other entries that are in demand". 

Which means if you pin even a few entries, that causes the cache to be *unable* to bring in other entries. You have improved memberof, at the expense of other operations, because you are focused on the member of behaviour, not on the behaviour *after* that operation completes. 

Second, you also are only solving a subset of the cases - a small number of large groups. Not what happens if you have a large number of large groups. That also is a use case, and this change doesn't address it. 

> 
> | you want your cache to be able to evict these entries No. That is exactly that we do not want! We want to dedicate part of memory to huge groups and avoid having to reload these one.
> 
> | Second, some operations may require more than 10 entries for their operation to complete efficiently So ? These entries will go in regular queue and handled by the LRU as it is currently done. I hope that no operation requires to load 10 huge groups and anyway if it is the case that only means that the administrator should adjust the 10 limit to something larger. (remember that the value is configurable and could even be set to 0 to disable the feature if needed. 10 is really an arbitrary (and arguable) default value. I wonder if the default should not depends on the cache size. Not sure we want to preserve any entries on tiny containers for example)

But again, this doesn't fix the problem then. If you have memberof with 100 large groups, then you haven't fixed the problem for that user? I think you are so focused on this one small problem that you are potentially missing the bigger picture and negative consequences of this change. 

> 
> | But that is fundamentally the point of a cache is that things are expensive to load, so we want to avoid loading them. True. | So ARC does solve this problem False. ARC is based of access frequency and has no knowledge of how expensive it is to load an entry in the cache. So an entry that is accessing 10 time less often than another one will be evicted even if its loading time is 100 time higher

ARC does solve this - it's a combination of LRU and LFU (in a manner). So when you have a *high cost* operation like the memberof change, ARC *adapts* on the fly to ensure that your large groups will be retained in the frequent set (at the expense of the recent set). Once that operation completes, overtime the adaptation adjusts to be able to remove items from that frequent set so that other entries can be brought into the cache. 

ARC will ensure that your 100 groups for memberof are in the frequent set, and that once that is finished, new entries can come in and be cached instead. This is *exactly what you want* and you don't need any tuning or hardcoding of magic values like "10 large entries". ARC was literally designed for this exact use case in mind, and to defend from both cache scanning attacks, but also frequent usage item pinning in the cache preventing evictions. 

> 
> I do not say that using ARC instead of LRU is a bad idea. (I agree with you that it could probably improve things) I only say it does not solve the point I was trying to solve (avoiding to have to reload very expensive entries) And BTW separating a huge groups entries from the others is not incompatible with using ARC (in other word, in the future, nothing prevent us to replace the LRU algorithm used for the regular queue by ARC)

It does solve the problem. I have spent unreasonable amounts of time studying caching, and I can tell you, it does what you want here, in a more flexible manner, and a safer one too. 



The second problem here is member of being slow in general - I did propose a new algorithm to solve this a number of years ago and it was not adopted. We use it in kanidm and we are able to handle thousands of recursively nested large groups with MO and 100,000s of users and do the updates in a fraction of the time of 389-ds. So the other side of this is that memberof and it's processing algorithm needs to be updated too, because that is also a source of issues. That alone would potentially even remove the need to change the cache behaviour. So also consider that MO needs an update to how it processes entries.

tbordaz:
@Firstyear, let's assume we have 10 large (1M) groups. Each of them taking ~1min to upload in the entry cache. The workload hit those 10 entries but then stop accessing them for a long time (1h) while the entry cache is full. LRU will evict those 10 entries, what mechanism in ARC prevents to evict them as well, as then are not frequently accessed ?

If the memory footprint of the 10 "saved" highest expensive entries, is not accounted from the entry cache. That would preserve the current entry cache behavior. That would extend, behind the scene, the configured/tuned entry cache by the size of the 10 largest entries. Would it be an option ?

Firstyear:
> @Firstyear, let's assume we have 10 large (1M) groups. Each of them taking ~1min to upload in the entry cache. The workload hit those 10 entries but then stop accessing them for a long time (1h) while the entry cache is full. LRU will evict those 10 entries, what mechanism in ARC prevents to evict them as well, as then are not frequently accessed ?

ARC has "ghost lists" that track keys that were previously in the cache, so that if they are re-accessed they can be brought back immediately into the frequent/recent sets and adjust weighting of those sets accordingly. That's how ARC works, it's the best of both LFU and LRU, but without the penalties of either. It's cheap resource wise, and has extremely good cache accuracy generally, it works for a variety of workloads. Outside of clock-pro, ARC is probably the best cache algorithm you can choose. 

I also want to point out a different side of this which you may not be considering - the issue here isn't the need to pin 10 entries for memberof, it's the fact that with LRU and memberof, you are performing a cache scan/churn attack on yourself (which by the way, ARC resists and would not be affected by, again solving the problem). Because the current memberof algorithm is updating all leaf entries as it proceeds, you are churning the cache as you walk all the leaf entries, which is evicting all the groups and entries as you go. This forces you to need to evict and bring back groups multiple times to the entry cache. So you're blowing up the cache on yourself. I presented an improved memberof algorthim nearly 8 years ago which resolves all the groups memberOf first, then once that is stable, updates entries *once* with the correct memberOf data.

So instead of doing this, why not fix the problem which is MemberOf's algorithm? The algorithm I presented would avoid the need to pin any groups at all, because it doesn't blow the cache up every time it runs?

Similar, why not use ARC - you are performing a linear scan over the cache within memberOf, which causes all your entries to churn and be evicted. Why not use a cache algorithm that resists that scanning attack and will keep the groups in working memory as you perform memberOf, and then will adapt once that operation finishes to contain entries that are now used in searches? 

> 
> If the memory footprint of the 10 "saved" highest expensive entries, is not accounted from the entry cache. That would preserve the current entry cache behavior. That would extend, behind the scene, the configured/tuned entry cache by the size of the 10 largest entries. Would it be an option ?

But the 10 saved entries may not be accounted in the entry cache, but you have finite ram. You're saying " lets not track these in the entry cache" but in reality it's "lets sacrifice the size of the entry cache and avoid looking at the problem we are making". More concrete - lets say we have 2GB of ram. Entry cache uses what ... 1.5GB? Now you update with these "10 fixed entries". To fit them in ram, you make the entry cache 1.2GB or less. Sure, you may not account for these entries in the entry cache, but you have to *reduce* the size of the entry cache anyway to make sure that they can always fit. You're still sacrificing the entry cache size for all other operations. 

And what if there are more than 10 large groups. What if it's 20? 30? What now? You hardcoded 10 as a hack, to avoid one situational problem, do you add more hacks on top? Or do you fix the *real problems* - the real problems being the use of LRU not ARC, and the memberof algo not having been updated with the performance improvements I suggested?


Right now I think both you and pierre are so focused on this narrow section of the problem (make memberof fast when you have up to 10 large groups), that you are missing the real cause of the problems (memberof algorthim scans all entries evicting the whole cache repeatedly, that the current MO algorithm is poor in performance). But additionally, this narrow focus means you aren't seeing the consequences of this change (MO won't be faster if you have greater than 10 large groups, you are permanently sacrificing entry cache space for a fraction of operations, you're adding a hack to work around a narrow problem rather than fix the real causes). 

Something I don't get is why you are so laser focused on this as the solution. I'm telling you from experience in a parallel and similar project that we don't have these problems *because* we use ARC and we have the improved MO algorithm. Why do you not want to adopt these things when they work? Why are you so focused on "this must be correct"?

tbordaz:
Thanks @Firstyear for your detailled explanations and the time you spent describing them. There are 4 points that I want to make clear

- The concern is the free of an expensive entry. It is related to time/cpu spent creating a sorted valueset when uploading an entry in the entry cache and so when we have spent a significant amount of time it looks problematic to free the entry.
- This design/patch has nothing to do with memberof plugin. Upload of a large group is needed independently of member that is a post-op. If it is a search or an update that need to access a group it is done before (in backend code) and independently of memberof. The CU cases that raised this issue were solved with a early priming of the big groups (several millions of members) because the workload timeout before the entry was finally loaded.
- I agree that ARC looks more efficient that current LRU and we should investigate the change of the algorithm but (sorry) I am still missing how ARC will solve the current issue (see below)
- I agree that the internal algorithm of memberof is less efficient than the one you proposed years ago and there is no evidence it is more robust. However it is not related to the concern (free of an expensive entry)

Regarding ARC "ghost list" that keeps tracks of keys previously in the cache. What are the keys ? Are the hash keys of a DN/nsuniqueid, are they the backentry struct or something else. If it is the backentry then when are they free ?

Regarding the memory footprint of the "preserved" entries. A production server running millions of users is not 2Gb ram. It is more in a range of > 100Gb. The tuning of the entry cache is usually not expecting to fit all entries and there are doubts that a cache above 20Gb is really beneficial. So for those host, booking few (1G-5G) extra Gb for "preserved" entries will not limit the real entry cache.

Firstyear:
> Thanks @Firstyear for your detailled explanations and the time you spent describing them. There are 4 points that I want to make clear
> 
>     * The concern is the free of an expensive entry. It is related to time/cpu spent creating a sorted valueset when uploading an entry in the entry cache and so when we have spent a significant amount of time it looks problematic to free the entry.

Yes, and this is what caching solves. We agree there.

> 
>     * This design/patch has nothing to do with memberof plugin. Upload of a large group is needed independently of member that is a post-op. If it is a search or an update that need to access a group it is done before (in backend code) and independently of memberof. The CU cases that raised this issue were solved with a early priming of the big groups (several millions of members) because the workload timeout before the entry was finally loaded.

MemberOf is related - discussed below.

> 
>     * I agree that ARC looks more efficient that current LRU and we should investigate the change of the algorithm but (sorry) I am still missing how ARC will solve the current issue (see below)
> 
>     * I agree that the internal algorithm of memberof is less efficient than the one you proposed years ago and there is no evidence it is more robust. However it is not related to the concern (free of an expensive entry)

But it's very related Thierry - why are we freeing the expensive entry in the first place is *the real problem* here. Why is the cache not retaining it? 

That's why I'm trying to say. You're looking at "how can we keep the expensive entry in memory". I'm looking at "why is it being evicted in the first place". 

That difference in perspective is what is leading us to different answers. You are saying you want to keep expensive entries in the cache, I'm arguing that we should prevent them being evicted in the first place. And currently there are multiple reasons why it could be evicted.

* LRU is a bad algorithm and prone to early eviction of entries that you still need
* MemberOf blasts the entry cache and pretty much guarantees you evict everything in the cache.

Your solution will work for a small number of large groups, but it won't work for large numbers of medium groups, or large numbers of large groups. What I am suggesting resolves all these cases. Your solution has more *negative* consequences that are concerning. Yes it fixes your problem, but it introduces 3 new problems. 

But this problem has another dimension - make entry loading faster - surely there is a way to make valueset faster for loading so that it's not as expensive ... 

> 
> 
> Regarding ARC "ghost list" that keeps tracks of keys previously in the cache. What are the keys ? Are the hash keys of a DN/nsuniqueid, are they the backentry struct or something else. If it is the backentry then when are they free ?

The keys are the same as whatever the normal keys are. It could be a DN, an entry id, nsuniqueid. Whatever the current LRU uses for it's lookup, is the key that is in the ghost list. The ghost list size is still bounded, it's not unlimited. But it's large enough to know that something did exist in the cache before, and where. But it does NOT keep the whole entry in the ghost sets, the ghost sets just track something with that key did exist, but not the data associated to it. 

> 
> Regarding the memory footprint of the "preserved" entries. A production server running millions of users is not 2Gb ram. It is more in a range of > 100Gb. The tuning of the entry cache is usually not expecting to fit all entries and there are doubts that a cache above 20Gb is really beneficial. So for those host, booking few (1G-5G) extra Gb for "preserved" entries will not limit the real entry cache.

Then if you have > 100GB of ram, why is anything being evicted *at all*. Make the entry cache bigger and be done with it. And if you have 100GB of ram you absolutely do expect the entry cache to hold all the content of the database in memory. I don't know any admin that would see think "I have all this ram, lets not use it". 

And if you "doubt" that the entry cache benefits over 20Gb, then that is a sign that there is a problem with the entry cache. There should be no reason we can't use nearly all of that ram for the entry cache to fit our dataset in and prevent evictions if there is that much ram. 

If we have a smaller amount of ram, then it becomes more important that we are able to use all of it effectively - and not pin things needlessly. Again, better caching fixes all of the above. 

So what is the resistance here? I'm proposing a better option that fixes not just this problem, but a whole class of cache related issues.

Now for the REALLY long part. 

>     * I agree that ARC looks more efficient that current LRU and we should investigate the change of the algorithm but (sorry) I am still missing how ARC will solve the current issue (see below)

I assume you know how LRU works. If we have a cache of 4 elements, then when we need to evict, we evict the "oldest" used item. Here we evict element `1` and we add `5` to the end. 

```
┌───┐   ┌───┐   ┌───┐   ┌───┐        
│ 1 │──▶│ 2 │──▶│ 3 │──▶│ 4 │        
└───┘   └───┘   └───┘   └───┘        
                                     
        ┌───┐   ┌───┐   ┌───┐   ┌───┐
        │ 2 │──▶│ 3 │──▶│ 4 │──▶│ 5 │
        └───┘   └───┘   └───┘   └───┘
```

Now what LRU is vulnerable to a specific type of attack (if you want to call it that) when the cache is *smaller* than the possible set of elements that exist. For example our database has 20 entries, but the cache only fits 5 of them. 

So lets imagine here we have an ALLIDS search. We are going to load elements 1 through 20. Let's imagine our cache as before:

```
┌───┐   ┌───┐   ┌───┐   ┌───┐
│ 1 │──▶│ 2 │──▶│ 3 │──▶│ 4 │
└───┘   └───┘   └───┘   └───┘
```

For elements 1 through 4 we are fine - we scan go through the cache and everything we want to access is there and found.

Now what happens when we access element 5? Well, we evict 1 and add 5. 

But what about 6? Well, now we evict 2, and add 6. What about 7? we evict 3 and add 7. 

This repeats and goes on until the whole set of the cache has been churned - now the cache is elements 17 through 20. 

```
┌───┐   ┌───┐   ┌───┐   ┌───┐                
│ 1 │──▶│ 2 │──▶│ 3 │──▶│ 4 │                
└───┘   └───┘   └───┘   └───┘                
                                             
        ┌───┐   ┌───┐   ┌───┐   ┌───┐        
        │ 2 │──▶│ 3 │──▶│ 4 │──▶│ 5 │        
        └───┘   └───┘   └───┘   └───┘        
                                             
                ┌───┐   ┌───┐   ┌───┐   ┌───┐
                │ 3 │──▶│ 4 │──▶│ 5 │──▶│ 6 │
                └───┘   └───┘   └───┘   └───┘
                                             
                                             
                                             
                ┌───┐   ┌───┐   ┌───┐   ┌───┐
                │17 │──▶│18 │──▶│19 │──▶│20 │
                └───┘   └───┘   └───┘   └───┘
```


But before this ALLIDS scan, the cache had entries 1 to 4 - they were the items being "used the most" that's why they were there. So now we have to evict 17 through 20 and re-load 1 to 4. 

Sounds familiar to your problem doesn't it ;) you have some entries (1 to 4) that you want to prefer in the cache, but they are easily able to be removed because LRU is vulnerable to scanning. 

Another situation in DS where this occurs is MemberOf - because so many entries have to be updated multiple times, this effectively acts as a scanning attack on the cache. Everything is evicted and then has to be re-included. That's why I'm bringing it up, because MemberOf is a great way to blow the cache content away. 


So you asked, how does ARC protect from this? 

ARC works by having *4* entry lists at a time. The frequent set, the recent set, the ghost frequent set, and the ghost recent set. 

Initially these lists are all balanced. To make this shorter, I'll skip the nice graphics sorry.

```
      recent   [ X  X  X  X  ]
ghost recent   [ X  X  X  X  ]
      frequent [ X  X  X  X  ]
ghost frequent [ X  X  X  X  ]
```

First we add some elements, lets say 1 to 4 - they all go to the *recent* set first. 

```
      recent   [ 1  2  3  4  ]
ghost recent   [ X  X  X  X  ]
      frequent [ X  X  X  X  ]
ghost frequent [ X  X  X  X  ]
```

Now these items get accessed many times. The next time they are accessed, they are promoted to frequent.

```
      recent   [ X  X  X  X  ]
ghost recent   [ X  X  X  X  ]
      frequent [ 1  2  3  4  ]
ghost frequent [ X  X  X  X  ]
```

Now lets say that we have a new item, 5 that is added. It goes from recent to frequent the same way, evicting item `1` in the process. Because item 1 was in frequent when evicted, it's key is added to the ghost set. 

```
      recent   [ X  X  X  X  ]
ghost recent   [ X  X  X  X  ]
      frequent [ 5  2  3  4  ]
ghost frequent [ 1  X  X  X  ]
```

At this point, we go to request item 1 from the cache again. Uh oh! It was evicted. But we find it in the ghost set. This indicates to the cache that there is *demand* on frequent items, and that we evicted too early. To respond to this, the cache now adjusts it's list sizes to allow more in frequent, at the expense of recent items. 

```
      recent   [ X  X  X  ]
ghost recent   [ X  X  X  X  X  ]
      frequent [ 1  2  3  4  5  ]
ghost frequent [ X  X  X  ]
```

You'll notice some changes here - the recent set got smaller, but the ghost recent set got bigger, and the ghost frequest set got smaller. I won't go into all the details about why, but lets just focus that the recent set is smaller now. We have adapted to the work load which wants more frequent items to be in the cache. 

So now, lets consider that ALLIDS situation again. As we churn through ALLIDS, we get items 1 to 5 as before from the frequent set. But once we get to item 6, what happens? Before, this caused 1 through 5 to be evicted with LRU. 

But not in ARC.

Here, item 6 is added to the recent set. I added 6 through 8 here:

```
      recent   [ 6  7  8  ]
ghost recent   [ X  X  X  X  X  ]
      frequent [ 1  2  3  4  5  ]
ghost frequent [ X  X  X  ]

```

Now we get to `9`. What now? Now, we evict from the recent set and the keys move to the ghost set. 

```
      recent   [ 9  10 11 ]
ghost recent   [ 6  7  8  X  X  ]
      frequent [ 1  2  3  4  5  ]
ghost frequent [ X  X  X  ]
```

See how now, we aren't evicting the items in frequent? Even though we are "churning" the recent set, the frequent set is *protected*, and those large expensive entries stay in the frequent set. No longer are we vulnerable to the same scanning attack that LRU is. 

Anyway, after the ALLIDS scan, here is what the cache would look like

```
      recent   [ 18 19 20 ]
ghost recent   [ 13 14 15 16 17 ]
      frequent [ 1  2  3  4  5  ]
ghost frequent [ X  X  X  ]
```

Again, see that the frequent set is protected, we didn't evict those large groups that are really important to keep cached. That's what you wanted here is it not? You wanted those expensive groups protected and cached :)  

Now, your next question might be - okay, but what if something is no longer being access from the frequent set? Can it be evicted? That was your concern after all William, we don't want to pin entries forever. 

Well, now is when the ghost recent set comes in. Lets say that entry 17 is accessed again. Now we look it up and it's in the ghost recent set. This indicates a demand on the recent set, that an entry should have been there that isn't. At this point, we adjust the cache sizes again and include 17 again. 

```
      recent   [ 17 18 19 20 ]
ghost recent   [ 13 14 15 16 ]
      frequent [ 1  2  3  4  ]
ghost frequent [ 5  X  X  X  ]
```

This can only happy if entry 17 is access multiple times in quick succession, showing that the content of the frequent set may not be as important as including new items now. So the weights shift to the recent set. If more items in the ghost recent set were accessed, the weight can continue to shift:

```
      recent   [ 16 17 18 19 20 ]
ghost recent   [ 13 14 15 ]
      frequent [ 1  2  3  ]
ghost frequent [ 4  5  X  X  X  ]
```

This is why the cache is called adaptive - it adapts between recent (new item inclusions) being the priority, or frequent and heavily used items being the priority. It protects against scanning attacks (ALLIDS or large MemberOf updates) , and unlike LFU it's not vulnerable to pinning attacks (where you request an item repeatedly so that it can never be freed). 

This is why I'm suggesting that ARC may be a better option here - it will keep large items in the cache if they are heavily accessed (frequent), while also being able to release them if there are heavy demands on other entries, allowing new entries to be included into the cache (recent). 

Hope that helps explain why this is a better option than pinning entries.

Firstyear:
> The huge load time creates disrupt the application leading to spurious authentication failure.
> 
> Neither ARC nor LRU can handle that because they do not make differences between entries load time ...
> 
> Pinning such entries in the cache is helpful even if it may decrease the overall performance when acceding the other entries but if the cache is large enough the degradation is small enough to not impair the service. (users does not really care if they have to wait 1 or 2 ms more when they log in but they complain when waiting more than a few seconds or getting error messages!
> 
> Not sure what else can be done to solve this specific user case. Maybe trying to decrease the load time (by ensuring that large groups members are normalized and sorted when writing then on id2entry db (so that we do not have to normalize and sort them when loading them but such a change may be disruptive for some other customers having application that are not strictly following the ldap recommandations relying on exact string value or order)

Any valueset greater that 10 elements is sorted. So we can do this. See VALUESET_ARRAY_SORT_THRESHOLD in valueset.c

Additionally, this may also impact the load time if the build of the valueset is using exact allocations, or repeatedly sorting. So I think there are optimisations here ... 

> 
> At least I am sure that pinning a specified number of entries would not impact other customers (especially if we set the default value to 0)

If the default is 0 that's okay, but what's the heuristic that chooses if the entry stays pinned?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -361,119 +374,237 @@ lru_verify(struct cache *cache, void *ptr, int in)
...
+    struct backcommon *e = (struct backcommon *)ptr;
+
     if (NULL == ptr) {
         LOG("=> lru_delete\n<= lru_delete (null entry)\n");
         return;
     }
-    e = (struct backcommon *)ptr;
+    if (type == ENTRY_CACHE) {
+        struct backentry *pe0 = CACHE_LRU_HEAD(cache, LRU_PRESERVED, struct backentry *);
+        struct backentry *e2 = (struct backentry *)ptr;
```
tbordaz: Why not using 'e' rather than 'e2' ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -361,119 +374,237 @@ lru_verify(struct cache *cache, void *ptr, int in)
...
-    if (e->ep_lrunext)
-        e->ep_lrunext->ep_lruprev = e;
-    if (!cache->c_lrutail)
-        cache->c_lrutail = e;
-#ifdef LDAP_CACHE_DEBUG_LRU
-    lru_verify(cache, e, 1);
-#endif
+    if (type == ENTRY_CACHE) {
+        struct backentry  *e2 = (struct backentry *)e;
+        if (e2->ep_weight == 0 || cache->c_inst->cache_preserved_entries == 0) {
```
tbordaz: Shouldn't we also add those with e2->ep_weight==1L. There is no need to bother with them for reserved entries

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -361,119 +374,237 @@ lru_verify(struct cache *cache, void *ptr, int in)
...
+            lru_add_before(cache, e, CACHE_LRU_HEAD(cache, LRU_REGULAR, struct backcommon *), LRU_REGULAR, type);
+            return;
+        }
+
+        /* Let move first preserved entry in regular list until there is enough
+         * space to add a new entry
+         */
+        while (cache->c_nbpreserved >= cache->c_inst->cache_preserved_entries) {
+            pe = CACHE_LRU_HEAD(cache, LRU_PRESERVED, struct backentry *);
+            if (e2->ep_weight <= pe->ep_weight) {
```
tbordaz: Any reason to use 'e2' rather than directly 'e' ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1236,6 +1451,16 @@ entrycache_replace(struct cache *cache, struct backentry *olde, struct backentry
...
         }
     }
+    /* Lets propagate the weight */
+    if (newe->ep_weight == 0) {
+        newe->ep_weight = olde->ep_weight;
+    }
+    if (olde->ep_weight) {
+        cache->c_stats.nehw--;
+    }
+    if (newe->ep_weight) {
```
tbordaz: if ep_weight differs between 'olde' and 'newe', should not we update cache->c_stats.weight ?

ldap/servers/slapd/back-ldbm/ldbm_instance_config.c
```diff
@@ -325,6 +325,95 @@ ldbm_instance_config_require_internalop_index_set(void *arg,
...
+{
+    struct ldbm_instance *inst = (struct ldbm_instance *)arg;
+
+    int64_t val = (int64_t)((uintptr_t)value);
+    if (val < 0) {
+        slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE,
+                              "Error: Invalid value for %s (%ld). The value must not be negative\n",
+                              CONFIG_INSTANCE_CACHE_PRESERVED_ENTRIES, val);
+        slapi_log_err(SLAPI_LOG_ERR, "ldbm_config_cache_weight_threshold_set",
+                      "Invalid value for %s (%ld). The value must be greater than \"100\"\n",
```
tbordaz: The message is rather "not negative"

ldap/servers/slapd/back-ldbm/ldbm_instance_config.c
```diff
@@ -325,6 +325,95 @@ ldbm_instance_config_require_internalop_index_set(void *arg,
...
+ldbm_config_cache_preserved_entries_set(void *arg,
+                               void *value,
+                               char *errorbuf,
+                               int phase __attribute__((unused)),
+                               int apply)
+{
+    struct ldbm_instance *inst = (struct ldbm_instance *)arg;
+
+    int64_t val = (int64_t)((uintptr_t)value);
+    if (val < 0) {
```
tbordaz, line 348: Shouldn't limit the max number of entries ? For example less than 1000 for memory footprint and computation time.

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -145,6 +145,7 @@ typedef unsigned short u_int16_t;
 #define DEFAULT_CACHE_SIZE       (uint64_t)0
 #define DEFAULT_CACHE_SIZE_STR   "0"
 #define DEFAULT_CACHE_ENTRIES    -1 /* no limit */
+#define DEFAULT_CACHE_PRESERVED_ENTRIES_STR "10"
```
tbordaz: We agreed to '0' by default

ldap/servers/slapd/back-ldbm/ldbm_instance_config.c
progier389: computation time is not really an issue.
Memory footprint is, but it is always hard to set a limit. (You can be sure that someone will want a bit more) My feeling is that we should let the value up to the administrator

ldap/servers/slapd/back-ldbm/cache.c
progier389: to avoid having to cast e everywhere  (backcommon versus backentry types )

ldap/servers/slapd/back-ldbm/cache.c
progier389: e has backcommon type which does not has ep_weight member

ldap/servers/slapd/back-ldbm/cache.c
progier389: There is a difference of nature: 0 means that no weight was computed so we should ignore them
1 means that there is a very small weight - the entry will probably not be pinned (or stay pinned for long)  but it is handled as any other entry with a weight )

progier389:
I have rewritten the changes to:
   - explicitly said that the entry are pinned in the cache
   - Do not embed the change in the LRU code so that we can easily replace it
   - Set the default to 0 
   - Entries are now put in the pinned list (or not) just before trying to add it in the LRU 
   - The pinned entries have now a specific flag and are stored in a doubly linked list

progier389:
Most tests are passing but there is an issue with the new test when running the test on github
(Looks like a test case setup issue ...)

E         ldap.UNWILLING_TO_PERFORM: {'msgtype': 103, 'msgid': 13, 'result': 53, 'desc': 'Server is unwilling to perform', 'ctrls': [], 'info': 'Error: "nsslapd-cachememsize" can not be updated while "nsslapd-cache-autosize" is set in "cn=config,cn=ldbm database,cn=plugins,cn=config".', 'ldap_request': "modify_ext_s(('cn=entrycache_test,cn=ldbm database,cn=plugins,cn=config', [(2, 'nsslapd-cachememsize', [b'8000000'])]),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance standalone1"}

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -56,6 +58,17 @@
 //#define LOG(_a, _x1, _x2, _x3)  ;
 #define LOG(...)
 #endif
+#define LOGPATTERN(cache, dn, msg, ...) { if (debug_pattern_matches(cache, dn)) { slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, "CACHE DEBUG: " msg, __VA_ARGS__); } }
+#define LOG2(...) slapi_log_err(SLAPI_LOG_INFO, (char *)__func__, __VA_ARGS__)
```
tbordaz: is it a leftover, do you mean 'LOG' ?

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -319,6 +320,8 @@ struct backcommon
 #define ENTRY_STATE_CREATING   0x2  /* entry is being created; don't touch it */
 #define ENTRY_STATE_NOTINCACHE 0x4  /* cache_add failed; not in the cache */
 #define ENTRY_STATE_INVALID    0x8  /* cache entry is invalid and needs to be removed */
+#define ENTRY_STATE_UNAVAILABLE 0xf /* entry is not fully created or is deleted */
```
tbordaz, line 323: It looks to me this state is tested, reset but never set.

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -358,24 +362,38 @@ struct backdn
...
     Hashtable *c_uuidtable;
 #endif
-    Slapi_Counter *c_hits; /* for analysis of hits/misses */
-    Slapi_Counter *c_tries;
     struct backcommon *c_lruhead; /* add entries here */
     struct backcommon *c_lrutail; /* remove entries here */
     PRMonitor *c_mutex;           /* lock for cache operations */
     PRLock *c_emutexalloc_mutex;
+    struct cache_stats c_stats;
+    struct ldbm_instance *c_inst;
```
tbordaz, line 397: In erase_cache should not we set c_inst to NULL ?
No operation should access the cache when the backend is destroyed but just in case it looks better to sigsev on a NULL pointer than accessing a freed struct.

ldap/servers/slapd/back-ldbm/backentry.c
```diff
@@ -132,3 +132,47 @@ backdn_init(Slapi_DN *sdn, ID id, int to_remove_from_hash)
...
+    Slapi_Attr *a = NULL;
+
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    slapi_timespec_diff(&now, (BackEntryWeightData*)starttime, &delta);
+    e->ep_weight = delta.tv_sec * 1000000UL + delta.tv_nsec / 1000UL;
+    if (e->ep_weight == 0) {
+        /* Ensure that entries with very small weight can be distinguished
+         * from those without any weight.
+         */
+        e->ep_weight = 1L;
```
tbordaz, line 158: in cache.c we often test ep_weight==0 to ignore a (small) entry.
Here (adding an entry in the entry cache) it is never 0.
Should we test ep_weight==1 in cache.c to ignore small entries

tbordaz (dismissed):
LGTM. Few minor comments

ldap/servers/slapd/back-ldbm/backentry.c
progier389: IMHO no:
The weight is computed when the entry get loaded (i.e by find_xxx functions) so the entry that are directly added (through add operation or import) have no weight 
==> 0 does not mean that the entry is small but that its weight is not available

The small entries are usually ignored later when their weight is compared to the first entry in the pinned list
small entry may get transiently pinned but they will be quickly ejected by entries with higher weight

ldap/servers/slapd/back-ldbm/back-ldbm.h
progier389: The cache struct is a direct member of the ldbm_instance struct so if the instance
So cache is freed when the instance is freed

ldap/servers/slapd/back-ldbm/back-ldbm.h
progier389: Indeed.
 this value is not settable and is used to test a set of states:
      ENTRY_STATE_DELETED|ENTRY_STATE_CREATING|ENTRY_STATE_NOTINCACHE|ENTRY_STATE_INVALID
 (otherwise its value would have been 0x10 :wink: )

ldap/servers/slapd/back-ldbm/cache.c
progier389: It is a debugging leftover. will remove it.

progier389:
@sourcery-ai review

ldap/servers/slapd/back-ldbm/ldbm_instance_config.c
```diff
@@ -326,6 +326,95 @@ ldbm_instance_config_require_internalop_index_set(void *arg,
...
+                          error ? error: "None");
+            return LDAP_UNWILLING_TO_PERFORM;
+        }
+    }
+    if (apply) {
+        if (val) {
+            val = slapi_ch_strdup(val);
+        }
+        slapi_ch_free_string(&inst->cache_debug_pattern);
+        slapi_ch_free((void**)&inst->cache_debug_re);
```
tbordaz: should be slapi_re_free(inst->cache_debug_re); inst->cache_debug_re=NULL;

ldap/servers/slapd/back-ldbm/ldbm_instance_config.c
```diff
@@ -326,6 +326,95 @@ ldbm_instance_config_require_internalop_index_set(void *arg,
...
+                                  error ? error: "None");
+            slapi_log_err(SLAPI_LOG_ERR, "ldbm_config_cache_weight_threshold_set",
+                          "Invalid value for %s (%s). The value must be"
+                          " a valid regular expression. Error is: %s\n",
+                          CONFIG_INSTANCE_CACHE_DEBUG_PATTERN, val,
+                          error ? error: "None");
+            return LDAP_UNWILLING_TO_PERFORM;
+        }
+    }
+    if (apply) {
```
tbordaz, line 405: if apply is false, we need to free 're'

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -632,6 +678,9 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                     entry->ep_state |= ENTRY_STATE_INVALID;
                     if (entry->ep_refcnt == 0) {
                         entry->ep_refcnt++;
+                        if (entry->ep_state & ENTRY_STATE_PINNED) {
```
tbordaz, line 681: Something not clear to me, refcnt is increased before calling 'pinned_remove'. So the entry will not go on the lru although I believe it should.

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -589,6 +632,9 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                 entry->ep_state |= ENTRY_STATE_INVALID;
                 if (entry->ep_refcnt == 0) {
                     entry->ep_refcnt++;
+                    if (entry->ep_state & ENTRY_STATE_PINNED) {
```
tbordaz, line 635: Something not clear to me, refcnt is increased before calling 'pinned_remove'. So the entry will not go on the lru although I believe it should.

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -902,8 +938,10 @@ cache_set_max_entries(struct cache *cache, int64_t entries, bool autotuned)
     }
 
     /* check for full cache, and clear out if necessary */
-    if (CACHE_FULL(cache))
+    if (CACHE_FULL(cache)) {
+        pinned_flush(cache);
         eflush = entrycache_flush(cache);
```
tbordaz, line 941: pinned_flush should be called before calling entrycache_flush. Why not calling it at the beginning of entrycache_flush ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1365,13 +1566,21 @@ entrycache_return(struct cache *cache, struct backentry **bep, PRBool locked)
                 }
                 backentry_free(bep);
             } else {
-                lru_add(cache, e);
+                pinned_verify(cache, __LINE__);
+                if (!pinned_add(cache, e)) {
```
tbordaz, line 1568: Do you want to add the entry on lru only if the entry is not pinned ? Then why not testing 'e->ep_state & ENTRY_STATE_PINNED

tbordaz (commented):
Few more questions

ldap/servers/slapd/back-ldbm/cache.c
progier389: I do not understand your point:
We are flushing the entries out of the cache because of a rollback and found an entry ready to be released (i.e: recent enough (remove_it flag) and no more used (refcount == 0) Since this thread use it refcount++ (that is the old code) ,then if the entry is pinned it is removed (anyway it is invalid) from pinned queue and moved in lru then it is removed from lru then freed

ldap/servers/slapd/back-ldbm/cache.c
progier389: Same answer than previews comment

ldap/servers/slapd/back-ldbm/cache.c
progier389: That is pinned_add(() that determine if the entry should be pinned and set the ENTRY_STATE_PINNED flag ...

ldap/servers/slapd/back-ldbm/cache.c
tbordaz: At the end of pinned_remove, the entry is move to lru (lru_add), at the condition refcnt==0.
So if we increase refcnt before calling pinned_remove then pinned_remove will not move the entry to LRU. Is that expected ?

progier389: It looks normal: 
  - LRU only contains entries with refcnt == 0 (because entry that are being used should not be removed by the LRU)
  -  The entry will be added in the LRU later on (by entrycache_return when refcnt will be 0)

tbordaz: Thanks for checking and explanations

tbordaz (approved):
LGTM

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -340,6 +356,24 @@ hash_stats(Hashtable *ht, u_long *slots, int *total_entries, int *max_entries_pe
 /***** add/remove entries to/from the LRU list *****/
 
 #ifdef LDAP_CACHE_DEBUG_LRU
+static void
+pinned_verify(struct cache *cache, int lineno)
```
mreynolds389, line 360: line number is not used

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1045,6 +1079,11 @@ entrycache_remove_int(struct cache *cache, struct backentry *e)
     const char *uuid;
 #endif
 
+    LOGPATTERN(cache, backentry_get_ndn(e),
+               "Cache average weight is %lu . Removing entry from "
+               " cache with size: %lu, weight: %lu, dn:%s\n",
```
mreynolds389: extra leading space

src/lib389/lib389/cli_conf/backend.py
```diff
@@ -886,6 +888,7 @@ def create_parser(subparsers):
     set_backend_parser.add_argument('--disable', action='store_true', help='Disables the backend database')
     set_backend_parser.add_argument('--cache-size', help='Sets the maximum number of entries to keep in the entry cache')
     set_backend_parser.add_argument('--cache-memsize', help='Sets the maximum size in bytes that the entry cache can grow to')
+    set_backend_parser.add_argument('--cache-preserved-entries', help='Sets the maximum number of entries that are not evicted from the cache when trying to make space. This is typically used to keep very large groups in the cache')
```
mreynolds389, line 891: Do we want to expose this to the UI?  Or, is it best to keep it on the obscure side and just be available in CLI?

mreynolds389 (approved):
LGTM!

src/lib389/lib389/cli_conf/backend.py
progier389: No strong opinion about it. But I think that very few customer will need to change the default value so we may let it only CLI for now. It will always be time to improve the UI if we see that lot of users use that feature.

Issue #6787, opened 2025-05-19
Improve  error message when bulk import connection is closed


mreynolds389:
**Issue Description**

If an online replication initialization connection is closed a vague error message is reported when the init is aborted:  

    [02/Apr/2025:11:31:52.948428656 +0300] - ERR - factory_destructor - ERROR bulk import abandoned

It should be clear that the import is being abandoned because the connection was closed

mreynolds389:
f18581e06..51024f1b9  389-ds-base-3.0 -> 389-ds-base-3.0
a6e4ae9b1..d2128145d  389-ds-base-2.6 -> 389-ds-base-2.6
9fcbe7d04..072eddef5  389-ds-base-2.5 -> 389-ds-base-2.5
fcd60d63b..52adcc2ca  389-ds-base-2.4 -> 389-ds-base-2.4
7cbca1471..73ed7f6e6  389-ds-base-2.3 -> 389-ds-base-2.3
aa5d0ecbf..dd4f79dc4  389-ds-base-2.2 -> 389-ds-base-2.2
5d6b70423..d472dd83d  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6788, opened 2025-05-19
Issue 6787 - Improve error message when bulk import connection is closed

mreynolds389:
Description:

If an online replication initialization connection is closed a vague error message is reported when the init is aborted:

    factory_destructor - ERROR bulk import abandoned

It should be clear that the import is being abandoned because the connection was closed and identify the conn id.

relates: https://github.com/389ds/389-ds-base/issues/6787


progier389 (approved):
LGTM

PR #6790, opened 2025-05-20
Issue 6789 - Default PAM PTA config can't be accessed/modified

### LINKED ISSUE #6789: Default PAM PTA config can't be accessed/modified
**Issue Description**
Trying access default PAM PTA plugin configuration results in an error:
```
# dsconf localhost plugin pam-pass-through-auth config 'PAM Pass Through Auth' show
Error: No object exists given the filter criteria: PAM Pass Through Auth (&(&(objectclass=top)(objectclass=extensibleObject)(objectclass=pamConfig))(|(cn=PAM Pass Through Auth)))
```

While children entries can be accessed without any issues.

vashirov:
Bug Description:
Trying access default PAM PTA plugin configuration results in an error:
```
Error: No object exists given the filter criteria: PAM Pass Through Auth (&(&(objectclass=top)(objectclass=extensibleObject)(objectclass=pamConfig))(|(cn=PAM Pass Through Auth)))
```

While children entries can be accessed without any issues.

Fix Description:
Change scope for `PAMPassThroughAuthConfigs` to subtree.

Fixes: https://github.com/389ds/389-ds-base/issues/6789

### DIFF
--- a/dirsrvtests/tests/suites/plugins/pam_pta_test.py
+++ b/dirsrvtests/tests/suites/plugins/pam_pta_test.py
@@ -1,0 +1,45 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import pytest
+import logging
+from lib389.plugins import PAMPassThroughAuthPlugin, PAMPassThroughAuthConfigs
+from lib389.topologies import topology_st
+
+pytestmark = pytest.mark.tier1
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+def test_default_config_entry(topology_st):
+    """Verify that both default PAM PTA config entry and child entries can be searched.
+
+    :id: e1f3a04b-deed-42b0-a98d-768f5b7f3fdf
+    :setup: Standalone instance
+    :steps:
+        1. Create custom PAM PTA config entry
+        2. Read attributes of the custom config entry
+        3. Read attributes of the default config entry
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+    """
+    inst = topology_st.standalone
+    pam_pta = PAMPassThroughAuthPlugin(inst)
+    configs = PAMPassThroughAuthConfigs(inst)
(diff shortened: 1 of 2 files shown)


jchapma (approved):
This works for me, ack

jchapma:
Now the default config is displayed and editable in the UI.

vashirov:
@jachapma, PAM PTA configuration is a can of worms :upside_down_face: I'm not sure now about this fix.
I checked the docs all the way to RHDS 9 (RHEL 6):
> PAM pass-through authentication is configured in child entries beneath the PAM Pass-Through Authentication plugin container entry.

But child entries is "new" functionality that was [added](https://github.com/389ds/389-ds-base/commit/dbcd4481c8a2f38dc947dc728e30aac5e0a5761a) in 2012. Configuration via the parent entry was left for backwards compatibility.
So now we have a situation where default config entry exists with the following options:
```
pamMissingSuffix: ALLOW
pamExcludeSuffix: cn=config
pamIDMapMethod: RDN
pamIDAttr: notUsedWithRDNMethod
pamFallback: FALSE
pamSecure: TRUE
pamService: ldapserver
```
But it's hidden, can't be disabled, and can interfere with the additional config entries (as you saw in the customer case).

If it's exposed (with this PR), it can also be deleted through dsconf of WebUI, causing another way to break your instance.

Now I'm leaning towards disabling default entry config (no-op) and allow configuration only through child entries.

WDYT?

mreynolds389:
@vashirov Personally I like to disable it, and log a message or return an error when trying to modify the main entry.  I'm going to send an email to 389-users and see if anyone responds with how they use the plugin....

jchapma:
> Now I'm leaning towards disabling default entry config (no-op) and allow configuration only through child entries.
> 
> WDYT?

I agree with you both, does this mean we can remove these config values from template-dse.ldif or might that break something for users depending on the default config ?

mreynolds389:
> > Now I'm leaning towards disabling default entry config (no-op) and allow configuration only through child entries.
> > WDYT?
> 
> I agree with you both, does this mean we can remove these config values from template-dse.ldif: pamMissingSuffix: ALLOW pamExcludeSuffix: cn=config pamIDMapMethod: RDN pamIDAttr: notUsedWithRDNMethod pamFallback: FALSE pamSecure: TRUE pamService: ldapserver
> 
> or might that break something for users depending on the default config ?

Lets wait for feedback before proposing too many changes (but yes I would think we could get rid of it).  Ideally we can add an upgrade function to move the default config into a new config child entry (if none exist).

Issue #6791, opened 2025-05-21
crash in liblmdb during instance shutdown


progier389:
**Issue Description**

389ds crash seen after shutdown:

#0  0x00007ff2339ba239 in mdb_env_reader_dest () at /lib64/liblmdb.so.0.0.0
#1  0x00007ff2380b8ff1 in __nptl_deallocate_tsd () at /lib64/libc.so.6
#2  0x00007ff2380bbedb in start_thread () at /lib64/libc.so.6
#3  0x00007ff23812cc3c in clone3 () at /lib64/libc.so.6FYI:

**Package Version and Platform:**
 - Platform: RHEL 10
 - Package and version: 389-ds-base-3.0.6-3.el10_0.x86_64

**Steps to Reproduce**
1. Install IPA server on RHEL10.
2. Stop IPA services and start it again
3. Check dmesg

**Expected results**
Server should stop without crashing

**Additional context**
Original issue reported in https://issues.redhat.com/browse/RHEL-86878

Note: I failed to reproduce. Anyway it is a race condition and it is possible that 
 the reproducer state is incomplete 
(Maybe there should be a lingering ldaps connection as seen in the core file during the shutdown 
  or maybe it is a race condition that does not occurs on the machine that I am testing)
But the core is pretty clear:
 Most threads are already gone.
 The main thread is waiting to close the connection poll
The crashing is stopping (outside the thread callback function) and handling the private data cleanup
 and crashing in lmdb when trying to access the environ but the lmdb db is already closed

Traiged: https://issues.redhat.com/browse/IDMDS-5830

progier389:
9206ee101..d7ba84082  main -> main
   5fb95fb80..21ca962fe  389-ds-base-3.0 -> 389-ds-base-3.0

tbordaz:
@progier389 , do you know if this bug applies on 2.6 and 2.7 ?

progier389:
No idea. So far we have only report about it on RHEL 10 (i.e 3.x)
That said that part of the code is quite similar so it probably apply  to.
On the other side as the  crash occurs after the db is closed, the impact is minimal (core file generated and a record reporting the crash in the system journal) so IMHO we could afford not to backport it ... )(

droideck:
Should we back port it to the LMDB Tech Preview relevant `389-ds-base-2.x` branches?

PR #6793, opened 2025-05-22
Issue 6791 - crash in liblmdb during instance shutdown

progier389:
Sometime ns-slapd process crashes during the shutdown.
The core stacks shows that a thread attempts to use liblmdb after lmdb environment get closed 
in one of thread specific data destructor callback.
The fix ensure that lmdb is not called after its environment get closed

Issue: #6791 

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
@@ -97,6 +97,9 @@ typedef struct {
 static dbmdb_dbi_t *dbi_slots;    /* The alloced slots */
 static int dbi_nbslots;           /* Number of available slots in dbi_slots */
 
+static int32_t g_mdb_env_is_open = false;
+static void dbmdb_set_is_env_open(bool is_open);
+
 /*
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -97,6 +97,9 @@ typedef struct {
 static dbmdb_dbi_t *dbi_slots;    /* The alloced slots */
 static int dbi_nbslots;           /* Number of available slots in dbi_slots */
 
+static int32_t g_mdb_env_is_open = false;
```
droideck, line 100: Will it work correctly for multiple backends? (I might misunderstand how we handle multiple LMDB environments though. So, just making sure its thought through)

progier389: There is a single environment (except for the temporary db used during import to handle entryrdn but that is another part of the code).
txn may need to handle different backends to keep consistency (moddn operation or some plugins like referntial interghrity plugin or memberof) so we need a single environment

tbordaz (approved):
LGTM

PR #6804, opened 2025-06-02
Issue 6753 - Remove ticket 47808 test

mirielka:
Removing ticket 47808 test as duplicate of dirsrvtests/tests/suites/betxns/betxn_test.py::test_betxn_attr_uniqueness

Relates: #6753

Author: Lenka Doudova

Reviewer: ???


tbordaz:
I agree those two tests are close but they are not strictly identical.
The differences of the test ticket 47808 vs test betxn are

- the ticket uses for the plugin configuration  an old configuration method 'nsslapd-pluginargxx' while betxn uses 'add_unique_attribute' (i.e.  uniqueness-attribute-name) and 'add_unique_subtree'.
- the ticket checks 'sn' that is not RDN while betxn test use 'uid' that is RDN

For backward compatibility I have a doubt if it worth continue testing 'nsslapd-pluginargxx'. We may decide to port this test (old config, non RDN attribute) in betxn. Let's wait for others feedback

mirielka:
> I agree those two tests are close but they are not strictly identical. The differences of the test ticket 47808 vs test betxn are
> 
>     * the ticket uses for the plugin configuration  an old configuration method 'nsslapd-pluginargxx' while betxn uses 'add_unique_attribute' (i.e.  uniqueness-attribute-name) and 'add_unique_subtree'.
> 
>     * the ticket checks 'sn' that is not RDN while betxn test use 'uid' that is RDN
> 
> 
> For backward compatibility I have a doubt if it worth continue testing 'nsslapd-pluginargxx'. We may decide to port this test (old config, non RDN attribute) in betxn. Let's wait for others feedback

I amended the PR by parametrizing the betxn test to test using both 'uid' and 'sn' attribute. I consulted the testing using 'nsslapd-pluginargxx' attribute and it wasn't recommended as this way of changing configuration should be deprecated.

droideck (changes requested):
I think to fully cover it, we need to add a restart between user creates and verify with a search that duplicate entry doesn't exist. (one assert with `.exists()` and another search with `ldap.NO_SUCH_OBJECT` expectations.

droideck (approved):
Looks good! Ack

PR #6806, opened 2025-06-06
Issue 6753 - Port ticket 47640 test

mirielka:
Porting ticket 47640 test to dirsrvtests/tests/suites/betxns/betxn_test.py::test_linked_attributes_plugin.

Relates: #6753

Author: Lenka Doudova

Reviewer: ???


dirsrvtests/tests/suites/betxns/betxn_test.py
```diff
@@ -559,6 +560,51 @@ def fin():
...
+    # Add the linked attrs config entry
+    la_configs = LinkedAttributesConfigs(topology_st.standalone)
+    la_configs.create(properties={'cn': 'Manager Link',
+                                  'linkType': 'seeAlso',
+                                  'managedType': 'seeAlso'})
+
+    # Add an entry that has a link to an entry that does not exist
+    users = UserAccounts(topology_st.standalone, DEFAULT_SUFFIX)
+    manager = users.create_test_user(uid=999)
+    manager.add('objectclass', 'extensibleObject')
```
droideck: I think we shouldn't use `extensibleObject` when it's not really needed. If you add it to user, then `person` objectclass should work, IIRC.

droideck (approved):
Otherwise, looks good!

dirsrvtests/tests/suites/betxns/betxn_test.py
mirielka: Thank you, you're right, `extensibleObject` is not necessary in this case. Amended.

Issue #6805, opened 2025-06-05
Multiple backend entry cache tuning


mreynolds389:
**Issue Description**
 
Currently all backends are treated as equals when doing the entry/dn cache autotuning.  But in the case of multiple backends, some are larger than others, and the autotuning should take this into account and give the bigger backends larger cache sizes

PR #6808, opened 2025-06-06
Issue 6805 - RFE - Multiple backend entry cache tuning

mreynolds389:
Description:

Currently all backends are treated as equals when doing the entry/dn cache autotuning. But in the case of multiple backends, some are larger than others, and the autotuning should take this into account and give the bigger backends larger cache sizes.

We take the page count from each backend to detemine what percentage of the available memory each backend can use.

For MDB we don't have a cache autosize setting so that needed to be added.

Relates: https://github.com/389ds/389-ds-base/issues/6805


mreynolds389:
More work is needed in this PR, but pre-review can be done.  While the changes in mdb_instance.c are needed, so import_test.py::test_import_with_index passes, I'm not confident it's correct.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
```diff
@@ -2648,6 +2648,53 @@ dbmdb_get_entries_count(dbi_db_t *db, dbi_txn_t *txn, int *count)
...
+ */
+uint32_t
+dbmdb_get_inst_page_count(struct ldbminfo *li, ldbm_instance *inst)
+{
+    if (!inst->inst_id2entry) {
+        dbmdb_ctx_t *ctx = MDB_CONFIG(li);
+        int rc = 0;
+
+        if (!ctx->env) {
+            return 0;
```
progier389, line 2681: in bdb case, I think that you create the env, maybe we should do the same with lmdb if it is not already open ...

mreynolds389: > in bdb case, I think that you create the env, maybe we should do the same with lmdb if it is not already open ...

We do in dbmdb_start_autotune()  line 202

In bdb we have to do it for each backend (or so it appears - I could definitely be wrong).  So its in a different function and in a loop, but in mdb we can do it once for all the backends (before we loop over each backend).

ldap/servers/slapd/back-ldbm/db-mdb/mdb_config.c
```diff
@@ -544,7 +544,38 @@ dbmdb_ctx_t_serial_lock_set(void *arg,
     return LDAP_SUCCESS;
 }
 
+static void *
```
progier389, line 549: I wonder if  backend.py should not be modified too:  maybe nsslapd-cache-autosize should be moved from bdb only to global

mreynolds389: yeah definitely and the UI will probably need an update as well

mreynolds389: BTW I'm almost done with the PR, just have to handle the in-memory only behavior

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
```diff
@@ -2595,10 +2595,18 @@ bdb_public_bdb_import_main(void *arg)
             slapi_task_finish(job->task, ret);
         }
     } else {
+        dblayer_private *priv = NULL;
```
tbordaz, line 2571: Something not clear to me is that this part of code is in the import 'error' handling. Should not it be done only after the creation of the 'ancestorid' index ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -1778,6 +1778,112 @@ bdb_instance_start(backend *be, int mode)
     return return_value;
 }
 
+static int
+bdb_get_page_count(dbi_db_t *db, dbi_txn_t *txn, uint32_t *count)
```
tbordaz: Always called with NULL txn, can we get rid of it ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -1778,6 +1778,112 @@ bdb_instance_start(backend *be, int mode)
...
+}
+
+/*
+ * Get the page count for this backend instance
+ * If any error occurs just return 0
+ */
+uint32_t
+bdb_get_inst_page_count(struct ldbminfo *li, ldbm_instance *inst)
+{
+    if (!inst->inst_id2entry) {
```
tbordaz, line 1807: Sorry I missed where it is set in bdb code? It am under the impression it is always NULL in bdb.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
```diff
@@ -251,45 +258,12 @@ bdb_start_autotune(struct ldbminfo *li)
...
-            /* If we want to clamp down, remove the "+1". This would change the above from 510mb -> 448mb. */
-            clamp_div = (dn_size / (64 * MEGABYTE)) + 1;
-            dn_size = clamp_div * (64 * MEGABYTE);
-        }
-    }
-
-    slapi_log_err(SLAPI_LOG_NOTICE, "bdb_start_autotune", "found %" PRIu64 "k physical memory\n", mi->system_total_bytes / 1024);
-    slapi_log_err(SLAPI_LOG_NOTICE, "bdb_start_autotune", "found %" PRIu64 "k available\n", mi->system_available_bytes / 1024);
+    slapi_log_err(SLAPI_LOG_NOTICE, "bdb_start_autotune",
+                  "found %" PRIu64 "k physical memory\n",
```
tbordaz: Could be in Mega rather in Kilo

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
```diff
@@ -383,8 +422,7 @@ bdb_start_autotune(struct ldbminfo *li)
         slapi_log_err(SLAPI_LOG_WARNING, "bdb_start_autotune", "Total entry cache size: %" PRIu64 " B; dbcache size: %" PRIu64 " B; available memory size: %" PRIu64 " B; \n",
                       total_cache_size, (uint64_t)li->li_dbcachesize, mi->system_available_bytes);
         slapi_log_err(SLAPI_LOG_WARNING, "bdb_start_autotune", "%s\n", msg);
-        /* WB 2016 - This should be UNCOMMENTED in a future release */
-        /* return SLAPI_FAIL_GENERAL; */
+        return SLAPI_FAIL_GENERAL;
```
tbordaz: Not sure it should fail. It is just a warning but we may let the admin set this value.

dirsrvtests/tests/suites/roles/basic_test.py
```diff
@@ -748,6 +748,7 @@ def test_not_such_entry_role_rewrite(topo, request):
     log.info("Duration of the search was %f", duration)
     assert(len(entries) == 4)
     assert (duration < 1)
+    topo.standalone.config.loglevel(vals=(ErrorLog.DEFAULT,))
```
progier389: Just curious: why changing the loglevel ? Is it related to this feature ?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
progier389: @tbordaz , 
There is no need to autotune if the import fails so that is why the test is done here in the "else" part of the error handling code (in other words  when imports completes successfully.)
So its looks a right place to auto tune the cache

tbordaz: Yes that is correct. This part of code, although after 'error:' label, handles import failures and successes. Thanks

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
progier389: What are you speaking about ? inst ? 
 if that is the case, you are mistaken: probably doing a confusion between:
       ldbm  ( the back-ldbm code for backend backed on databases)
       lmdb : the lmdb database
 injst is the be->be_private data containing ldbm context and is never NULL

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
progier389: In any case the warning message is not in sync with the behavior:
You should not say: 'In a future release this WILL prevent server start up' then Fail to start ! :wink: 
==> Either we keep current behavior (no fail) or we should also change the message text and level

dirsrvtests/tests/suites/roles/basic_test.py
mreynolds389: plugin logging was previously enabled which breaks other tests, but yeah not sure if it's related to this PR or not ( did this change some time ago) :-)  I'll double check this...

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: It's not about inst, its about "inst_id2entry".  When a backend is not initialized this is NULL so we must check for this  to properly handle the count.  This is not a mistake.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
mreynolds389: > Could be in Mega rather in Kilo

Probably :-), yeah I'd like to start displaying these numbers in more modern/useful units.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: Also to Thierry we set this in:  dbbdb_create_db_for_open()

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
mreynolds389: > Always called with NULL txn, can we get rid of it ?

Yeah we can drop it.  In early development it was in place where it was being set, but not anymore.

dirsrvtests/tests/suites/roles/basic_test.py
progier389: OK. if it is not related, IMHO it is better to put in a different PR:
Q/E may have to cherry-pick it without wanting to merge the big RFE  :wink:

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
progier389: On second thought, my feeling is that we should keep the original behavior without returning.
If we decide to change the behavior we should better to do it in another PR
 (especially if we need to undo it afterwards ! :wink: )

mreynolds389: > On second thought, my feeling is that we should keep the original behavior without returning. If we decide to change the behavior we should better to do it in another PR (especially if we need to undo it afterwards ! 😉 )

Agreed

ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c
```diff
@@ -1778,6 +1778,113 @@ bdb_instance_start(backend *be, int mode)
...
+        } else {
+            return 0;
+        }
+
+        if ((mypEnv->bdb_DB_ENV->open)(mypEnv->bdb_DB_ENV,
+                                       inst_dirp,
+                                       mypEnv->bdb_openflags,
+                                       priv->dblayer_file_mode))
+        {
+            return 0;
```
droideck, line 1851: Do I understand correctly that if we exit here - we don't release mypEnv? (possible memory leak)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
```diff
@@ -2648,6 +2648,54 @@ dbmdb_get_entries_count(dbi_db_t *db, dbi_txn_t *txn, int *count)
...
+        }
+
+        rc = dbmdb_instance_start(inst->inst_be, DBLAYER_NORMAL_MODE);
+        if (rc == 0 && inst->inst_id2entry) {
+             dbmdb_get_page_count(inst->inst_id2entry, &inst->inst_page_count);
+             dblayer_instance_close(inst->inst_be);
+        } else {
+            slapi_log_err(SLAPI_LOG_ALERT, "dbmdb_get_inst_page_count",
+                          "error returning count: %d (%p)\n", rc, inst->inst_id2entry);
```
droideck, line 2689: In case of the error, we don't do `dblayer_instance_close(inst->inst_be);`. Why is it?
Maybe I miss something (probably, worth adding a comment here)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_misc.c
```diff
@@ -144,9 +144,225 @@ dbmdb_count_config_entries(char *filter, int *nbentries)
...
+            cache_set_max_entries(&(inst->inst_cache), -1);
+            cache_set_max_size(&(inst->inst_cache), ec_size + 1, CACHE_TYPE_ENTRY);
+        }
+        if (cache_size >= 0 && li->li_cache_autosize >= 0) {
+            slapi_log_err(SLAPI_LOG_NOTICE, "mdb_start_autotune",
+                          "cache autosizing: %s dn cache (%" PRIu64 " total): %s\n",
+                          inst->inst_name, backend_count,
+                          convert_bytes_to_str((double)dn_size, size_buffer, 0));
+            cache_set_max_entries(&(inst->inst_dncache), -1);
+            cache_set_max_size(&(inst->inst_dncache), dn_size + 1, CACHE_TYPE_DN);
```
droideck: Why in LMDB case we do '+1' but in BDB, we don't. It's not fully clear...

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -360,7 +360,7 @@ add_index_dbi(caddr_t attr, caddr_t otx)
     if (ai->ai_indexmask & INDEX_ANY) {
         octx->rc = add_dbi(octx, octx->be, ai->ai_type, flags);
         octx->ai = NULL;
-        return octx->rc ? STOP_AVL_APPLY : 0;
+        return octx->rc && octx->rc != MDB_NOTFOUND ? STOP_AVL_APPLY : 0;
```
droideck, line 363: Could you please add a comment here on why we continue in case of MDB_NOTFOUND only? Is it because index DB file is not present and that's okay?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
```diff
@@ -314,37 +291,101 @@ bdb_start_autotune(struct ldbminfo *li)
...
+                          "cache autosizing: %s entry cache (%" PRIu64 " total): %s\n",
+                          inst->inst_name, backend_count,
+                          convert_bytes_to_str((double)ec_size, size_buffer, 0));
             cache_set_max_entries(&(inst->inst_cache), -1);
-            cache_set_max_size(&(inst->inst_cache), li->li_cache_autosize_ec, CACHE_TYPE_ENTRY);
+            cache_set_max_size(&(inst->inst_cache), ec_size, CACHE_TYPE_ENTRY);
         }
-        if (dncache_size == 0 || dncache_size == MINCACHESIZE || li->li_cache_autosize > 0) {
-            slapi_log_err(SLAPI_LOG_NOTICE, "bdb_start_autotune", "cache autosizing: %s dn cache (%" PRIu64 " total): %" PRIu64 "k\n", inst->inst_name, backend_count, dn_size / 1024);
+        if (cache_size > 0 && li->li_cache_autosize >= 0) {
```
droideck: In both `ldbm_instance_config_cachesize_set()` and `ldbm_instance_config_cachememsize_set():`
we have `if (val > 0 && inst->inst_li->li_cache_autosize > 0)`

But here, we have `if (cache_size > 0 && li->li_cache_autosize >= 0) {` - it seems inconsistent, unless I miss something.

(`>= 0` vs `> 0`)

dirsrvtests/tests/suites/config/autotuning_test.py
```diff
@@ -359,6 +368,276 @@ def test_cache_autosize_invalid_values(topo, invalid_value):
...
+        1. Set in the cn=config,cn=ldbm database,cn=plugins,cn=config:
+        3. Set in the cn=UserRoot,cn=ldbm database,cn=plugins,cn=config:
+           nsslapd-cachememsize: 0 and some same value
+        4. Restart the instance
+        5. Check nsslapd-dbcachesize and nsslapd-cachememsize
+    :expectedresults:
+        1. nsslapd-cache-autosize is successfully set
+        3. nsslapd-cachememsize is successfully set
+        4. The instance should be successfully restarted
+        5. nsslapd-cachememsize should set to value greater than 512KB
```
droideck: Step ordering typo - and in `test_cache_autosize_multi_backends`

ldap/servers/slapd/back-ldbm/db-mdb/mdb_misc.c
mreynolds389: That's a debugging artifact.  I need to remove it.    Note this is a still a WIP :-)

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
mreynolds389: It's when the database is not initialized yet (at server install and new backend creation).  I will add a comment but I wanted Pierre to review this part  - it is needed, but it could ignore valid errors?

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
mreynolds389: sorry, again this was for debugging.  This PR is not done yet.  I still have to handle previous config settings, etc.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_misc.c
droideck: Oh! Sorry! I'll hold my horses a bit then!:D

mreynolds389: It's all good and I appreciate the comments, I'll get a bunch of this stuff cleaned up today as most of the debugging code is not needed now.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
progier389: Looks good: 
add_dbi returns the mdb_dbi_open return code: MDB_NOT_FOUND means that the database instance does not exists (and the open flag is not asking to create it). Other errors have a different code
http://www.lmdb.tech/doc/group__mdb.html#gac08cad5b096925642ca359a6d6f0562a

ldap/servers/slapd/back-ldbm/db-mdb/mdb_layer.c
progier389: An error is meaning that we fail to open the instance ==> there is no need to close it since it is not open ...

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -910,7 +918,12 @@ cache_get_max_size(struct cache *cache)
     uint64_t n = 0;
 
     cache_lock(cache);
-    n = cache->c_maxsize;
+    if (g_get_shutdown()) {
+        /* We are shutting down, return the manually set value */
```
tbordaz, line 922: If autotune==False, I think this value is unset. Shouldn't we test autotune as well

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
```diff
@@ -142,22 +142,20 @@ bdb_start_autotune(struct ldbminfo *li)
     ldbm_instance *inst = NULL;
```
tbordaz, line 142: bdb_start_autotune and dbmdb_start_autotune are looking very similar, except the way to collect inst page count. Is it better to have separated functions as the autotuning of the entrycache should be close given the size of available memory.

tbordaz (approved):
The overall C/python parts are looking good to me.

ldap/servers/slapd/back-ldbm/cache.c
mreynolds389: > If autotune==False, I think this value is unset. Shouldn't we test autotune as well

Not sure I understand the concern, and the code has changed since you left your comment, but I will try to answer it.  We do not need to adjust "c_config_maxentries" if we are autotuning because "c_config_maxentries" was already set at startup when the server does it autotuning.  So we only need to adjust "c_config_maxentries" if we are manually setting the value after startup.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
mreynolds389: > bdb_start_autotune and dbmdb_start_autotune are looking very similar, except the way to collect inst page count. Is it better to have separated functions as the autotuning of the entrycache should be close given the size of available memory.

They do have a lot of similarities of course, but there are some big differences (db cache, import cache, and the underlying db library).  I think trying to create a single function that handles both bdb/mdb will be much harder to read and complex.  The hope is that bdb will go away, and then we don't have to worry about it, but trying to handle both will add more work later when bdb is officially gone.  So that being said, I'd prefer to leave it as two separate functions.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
```diff
@@ -314,37 +291,100 @@ bdb_start_autotune(struct ldbminfo *li)
...
+    {
         inst = (ldbm_instance *)object_get_data(inst_obj);
         uint64_t cache_size = (uint64_t)cache_get_max_size(&(inst->inst_cache));
         uint64_t dncache_size = (uint64_t)cache_get_max_size(&(inst->inst_dncache));
+        uint64_t ec_size = MINCACHESIZE;
+        uint64_t dn_size = DEFAULT_DNCACHE_SIZE;
+        uint64_t clamp_div = 0; /* For clamping the autotune value to a 64Mb boundary */
+
+        /* Calculate this instance's slice of memory it can use */
+        if (total_pages) {
```
droideck, line 305: Do I understand correctly that `total_pages == 0` when DB doesn't have any entries?

So if we create a DB with no entries and then start to add them, we might end up in a situation when the cache is incorrectly auto tuned.
I wonder if should add some kind of warning when `total_pages == 0` (i.e. "No pages found in any backend, using minimum cache sizes"). Or something like that.

I understand it might be an edge case which will be resolved on restart, but still I wonder.

progier389: The drawback you mention is intrinsic to the notion of comparing the database entry sizes: if we add/delete a lot of entries  then the cache size will get wrong. So IMHO the empty database is not really an exception and does not deserve a special warning. (Furthermore in most case the backend is populated through import vor bulk import and the cache size is recomputed afterward.)

mreynolds389: If the DB is empty we basically default to 64 MB.  When you initialize a DB it re-autosizes all backends, and when you restart the server it also autosizes.  The error log will record everything autosizing does - so it is always logged - IMO I don''t see a need to say a backend is empty.

progier389 (approved):
LGTM

ldap/servers/slapd/back-ldbm/db-bdb/bdb_misc.c
droideck: Sounds good, and makes sense!

droideck (approved):
Ack from me too!

PR #6809, opened 2025-06-09
Issue 6753 - Port ticket test 47927

mirielka:
Description:
Porting ticket 47927 test to
dirsrvtests/tests/suites/plugins/attruniq_test.py

Relates: #6753

Author: Lenka Doudova

Reviewer: ???


dirsrvtests/tests/suites/plugins/attruniq_test.py
```diff
@@ -80,4 +95,161 @@ def test_modrdn_attr_uniqueness(topology_st):
...
+    UNIQUE_VALUE = '1234'
+
+    log.info('Create user with unique attribute')
+    user1.add('telephonenumber', UNIQUE_VALUE)
+    assert user1.present('telephonenumber', UNIQUE_VALUE)
+
+    log.info('Add exclude subtree')
+    topology_st.standalone.modify_s("cn=attruniq,cn=plugins,cn=config",
+                                    [(ldap.MOD_ADD, 'uniqueness-exclude-subtrees',
+                                      ensure_bytes(EXCLUDED_CONTAINER_DN))])
```
mirielka: I haven't found a method in attribute uniqueness plugin code for easy configuration of exclude subtrees similar to add_unique_subtree, would it be beneficial to create it in the scope of this PR? So far it would only be used in this test.

mreynolds389: I'm ok with you adding it in this PR, perhaps rework the title of the commit to focus on that change (so it's easier to track when looking at the git log)?

mirielka: Thank you, good idea! Amended.

src/lib389/lib389/plugins.py
```diff
@@ -175,6 +175,9 @@ def disable_all_subtrees(self):
 
         self.set('uniqueness-across-all-subtrees', 'off')
 
+    def add_exclude_subtree(self, basedn):
+        self.add('uniqueness-exclude-subtrees', basedn)
```
droideck: If we are adding a new method, I think we should also add `remove_exclude_subtree` for completeness

dirsrvtests/tests/suites/plugins/attruniq_test.py
```diff
@@ -154,3 +168,158 @@ def test_multiple_attr_uniqueness(topology_st):
...
+    log.info('Clean up users')
+    user1.delete()
+    user2.delete()
+    user3.delete()
+    user4.delete()
+    cont1.delete()
+    cont2.delete()
+    cont3.delete()
+    attruniq.disable()
+    attruniq.delete()
```
droideck: The deletes, probably, can be done in a loop.

Also, just to be safe, I'd do the clean up in the `finally` block (or any other similar approach)

droideck (changes requested):

droideck (approved):
LGTM

PR #6812, opened 2025-06-13
Issue 6811 - Downgrade concread to prevent regressions

### LINKED ISSUE #6811: Downgrade concread to prevent regressions
**Description**
We need to downgrade concread from 0.5.x to a stable version (2.2.21) for our upcoming 2.7.0 and 3.1.3 releases. Significant changes were introduced between concread 2.2.21 and 0.5.x, and we should not ship with 0.5.x until we can verify it doesn't cause the confirmed regressions. This downgrade should also be backported to older versions as a safety measure. Multiple downstream distributions (Fedora, SUSE, Debian/Ubuntu, Arch) depend on our releases and should not be affected by potential crashes. 

There is [a proposed fix](https://github.com/kanidm/concread/pull/140) that removes arc-swap (where the regression appears to originate), but this needs to be thoroughly tested in a reproducible environment before we consider using concread 0.5.x. Until we can confirm the fix resolves the issue without introducing new problems, we should proceed with the downgrade.

droideck:
Description: Downgrade concread from 0.5.x to 2.2.21 to prevent potential regressions. The arc-swap dependency in concread 0.5.x has been identified as the source of regressions. This change ensures stability until the proposed arc-swap removal fix can be thoroughly tested and validated.

Fixes: https://github.com/389ds/389-ds-base/issues/6811

Reviewed by: ?

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -13,34 +13,20 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


progier389 (approved):
LGTM. Better wait for the test completion before merging but so far it looks promising !

Firstyear:
The downgrade as mentioned is problematic as 0.5 has memory leak fixes you need. The issue was in arc-swap, which @droideck has already reverted, so 0.2 and 0.5 the big change is just that you are back on the mutex strategy. So I think this downgrade is premature.

Firstyear:
I have just released 0.5.6 which reverts the arcswap implementation, to the proven mutex approach for 2.2.

Issue #6813, opened 2025-06-16
Improve invalid syntax error messages


mreynolds389:
**Issue Description**

When an attribute has an invalid syntax we do not log what that invalid valid was.  

    "passwordExpirationTime: value #0 invalid per syntax"

It would be very helpful to log this information:

    "passwordExpirationTime: value #0 (ZZZZZZZZZZZ) invalid per syntax

PR #6814, opened 2025-06-16
Issue 6813 - Improve invalid syntax error messages

mreynolds389:
Description:

Log the invalid value in the log message to help improve debugging

relates: https://github.com/389ds/389-ds-base/issues/6813

### DIFF
--- a/ldap/servers/slapd/plugin_syntax.c
+++ b/ldap/servers/slapd/plugin_syntax.c
@@ -19,6 +19,7 @@
 
 #include "slap.h"
 
+#define INVALID_BUFSIZ 512
 
 struct slapdplugin *
 plugin_syntax_find(const char *nameoroid)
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/plugin_syntax.c
```diff
@@ -399,15 +399,21 @@ slapi_entry_syntax_check(
                     if ((a->a_plugin->plg_syntax_validate((struct berval *)bval)) != 0) {
                         if (syntaxlogging) {
                             slapi_log_err(SLAPI_LOG_ERR, "slapi_entry_syntax_check",
-                                          "\"%s\": (%s) value #%d invalid per syntax\n",
-                                          slapi_entry_get_dn(e), a->a_type, hint);
+                                          "\"%s\": (%s) value #%d (%s) invalid per syntax\n",
+                                          slapi_entry_get_dn(e),
+                                          a->a_type,
+                                          hint,
+                                          bval->bv_val ? bval->bv_val : "NULL");
```
tbordaz: Any risk that bval->bv_val is not null terminated ?

progier389: I agree: using directly the bval provided by the user looks unsafe.
we should take its len in account and we should probably also escape the bval

mreynolds389: Ok I think I got it worked out, please review...

droideck:
I have a bit of a concern (which can be groundless) that logging actual invalid values could expose sensitive information (passwords, API keys? etc.) if users accidentally enter them in the wrong fields.

We might add a configuration option to disable value logging (as audit can be disabled too) or, at least, we might need to document the security implications in the admin guide.

ldap/servers/slapd/plugin_syntax.c
```diff
@@ -500,16 +528,44 @@ slapi_mods_syntax_check(
...
+                                    /* strip last char/newline */
+                                    memcpy(&invalid_val,
+                                           mod->mod_bvalues[j]->bv_val,
+                                           INVALID_BUFSIZ-1);
+                                    /* Add truncation chars ... */
+                                    memcpy(&invalid_val[INVALID_BUFSIZ-4],
+                                           "...", 3);
+                                }
+                            } else {
+                                PR_snprintf(invalid_val, sizeof(invalid_val), "%s",
```
progier389, line 548: Still a risk of read overflow about bval 
should probably:
  - use bv_len as size instead of sizeof(invalid_val)
  - invalid_val[bv_len] = '\0';

ldap/servers/slapd/plugin_syntax.c
```diff
@@ -500,16 +528,44 @@ slapi_mods_syntax_check(
                 /* Loop through the values and validate each one */
                 for (j = 0; mod->mod_bvalues[j] != NULL; j++) {
                     if (syntax_plugin->plg_syntax_validate(mod->mod_bvalues[j]) != 0) {
+                        char invalid_val[INVALID_BUFSIZ] = {0};
+                        char buffer[BUFSIZ];
+                        if (mod->mod_bvalues[j]->bv_len > 0) {
+                            if (mod->mod_bvalues[j]->bv_len >= INVALID_BUFSIZ) {
+                                if (mod->mod_bvalues[j]->bv_val[mod->mod_bvalues[j]->bv_len] != '\0') {
+                                    PR_snprintf(invalid_val, INVALID_BUFSIZ - 3,
```
progier389, line 536: why -3 ?

tbordaz:
> I have a bit of a concern (which can be groundless) that logging actual invalid values could expose sensitive information (passwords, API keys? etc.) if users accidentally enter them in the wrong fields.
> 
> We might add a configuration option to disable value logging (as audit can be disabled too) or, at least, we might need to document the security implications in the admin guide.

Indeed it already exists code to prevent adding userpassword/unhashuserpassword value in audit log (add_entry_attrs_ext) or error log (entry_apply_mod). To align with this code I agree to not dump at least those values on a file (audit,error).

ldap/servers/slapd/plugin_syntax.c
mreynolds389: To leave room for "..." .  This is how we do it in other places.

Anyway this is turning into a can of worms and I'm not sure its worth it anymore.   I have other more important things that need to get done so I'm going to shelf this PR for now...

Issue #6817, opened 2025-06-17
Regression test test_repl_after_reindex missing Tasks import on 389-ds-base-2.4, 2.3, 2.2.


abeisemb:
**Issue Description**
The regression test dirsrvtests/tests/suites/replication/regression_m2_test.py::test_repl_after_reindex throws an error when running from upstream branches 389-ds-base-2.4, 389-ds-base-2.3, and 389-ds-base-2.2. The test succeeds on 389-ds-base-2.5 to 389-ds-base-3.0, as well as main; the test does not exist on 389-ds-base-2.1 and lower. 
Full error: 
```
>       tasks = Tasks(m1)
E       NameError: name 'Tasks' is not defined

dirsrvtests/tests/suites/replication/regression_m2_test.py:1013: NameError
```

**Steps to Reproduce**
1. Install latest 389-ds-base build. 
2. Run test, i.e. py.test -v --disable-warnings dirsrvtests/tests/suites/replication/regression_m2_test.py::test_repl_after_reindex. 

**Expected results**
Test runs through to completion, outputting either a success or failure. 

**Additional context**
1 line import fix, I'll submit a PR to 2.4 later today.

PR #6818, opened 2025-06-17
Issue 6817 - Add missing import to test_repl_after_reindex

abeisemb:
Description:
Fixes regression test test_repl_after_reindex that was throwing an error due to a missing Tasks import. This is a one line fix to add said import.

Resolves: https://github.com/389ds/389-ds-base/issues/6817


abeisemb:
Unable to build off of 389-ds-base-2.4 base branch, so these failures don't seem to be caused by my change to the test.

vashirov (approved):

droideck (approved):

Issue #6250, opened 2024-07-04
'entryUsn' value decrements/overflows on add/modify operation


CaptainDmitro:
**Issue Description**
Trying to add new entry or modify existing one results in decremented/overflowed 'entryUsn' value when the prior operation was adding existing entry.

**Package Version and Platform:**
 - Platform: Debian 4.19.269-1 (Buster) , Docker version 24.0.6, build ed223bc
 - Package and version: 389ds/dirsrv:2.4; 389ds/dirsrv:3.0

**Steps to Reproduce**
1. Start 389ds with USN plugin enabled and sample data: 

```
1. docker run --name 389ds_test -p 3389:3389 -v dirsrv:/data -e DS_DM_PASSWORD='TestPwd100' -e DS_SUFFIX_NAME='dc=example,dc=net' 389ds/dirsrv:2.4
2. docker exec 389ds_test dsconf localhost config replace nsslapd-dynamic-plugins=on
3. <restart container>
4. docker exec 389ds_test dsconf localhost plugin usn enable && \
      docker exec 389ds_test dsconf localhost plugin usn global on
5. docker exec 389ds_test dsconf localhost backend create --suffix="dc=example,dc=net" --be-name userRoot --create-suffix --create-entries
```

2. Try to add existing entry 'uid=demo_user,ou=people,dc=example,dc=net' for serveral times (e.g. 10 times)
4. Modify operation for entry 'uid=demo_user,ou=people,dc=example,dc=net' results in 'entryUsn' value changed from '6' to '18446744073709551611' 

**Expected results**
According to [Entry USN Design](https://www.port389.org/docs/389ds/design/entry-usn.html) 'entryUsn' value for added/updated entry should be the greatest among other entries.

**Screenshots**
![Decremented and overflowed 'entryUsn'](https://github.com/389ds/389-ds-base/assets/48180766/26e4aec7-b527-4d09-825f-6a7535c73ef0)

**Additional context**
[Python script ](https://gist.github.com/CaptainDmitro/ac6e9566af176f4388f6dd1ef1afad82) to reproduce bug.

droideck:
377a4ddef..605d30959  389-ds-base-3.0 -> 389-ds-base-3.0
   ff8393ff5..41209deb3  389-ds-base-2.7 -> 389-ds-base-2.7
   77abc92fc..c308dffcf  389-ds-base-2.6 -> 389-ds-base-2.6
[...]

PR #6821, opened 2025-06-18
Issue 6250 - Add test for entryUSN overflow on failed add operations

droideck:
Description: Add comprehensive test to reproduce the entryUSN overflow issue where failed attempts to add existing entries followed by modify operations cause entryUSN values to underflow/overflow instead of incrementing properly.

Related: https://github.com/389ds/389-ds-base/issues/6250

Reviewed by: ?


droideck:
Looks like we have some kind of race condition in entryUSN (with incrementing and decrementing in bepostop).
But I haven't checked usn.c deeper... Just the test for now.

tbordaz (approved):
LGTM

PR #6826, opened 2025-06-20
Issue 6825 - RootDN Access Control Plugin with wildcards for IP addre…

### LINKED ISSUE #6825: RootDN Access Control Plugin with wildcards for IP addresses fails with an error
**Issue Description**
RootDN Access Control Plugin with wildcards for IP addresses fails with an error "Invalid IP address"

The documentation states that it should be possible:  "--allow-ip, and --deny-ip. These are all multi-valued attributes and you can use wild cards to allow or deny IP ranges."

 - Package and version: main branch back

**Steps to Reproduce**
dsconf -D "cn=Directory Manager" ldaps://localhost:3012  -w password plugin root-dn set --allow-ip 10.10.10.*
Error: Invalid IP address (10.10.10.*) for '--allow-ip'

**Expected results**
IP address specified with wildcard should get added

jchapma:
…sses fails with an error

Bug description:
RootDN Access Control Plugin with wildcards for IP addresses fails withi an error "Invalid IP address"

socket.inet_aton() validates IPv4 IP addresses and does not support wildcards.

Fix description:
Add a regex pattern to match wildcard IP addresses, check each octet is between 0-255

Fixes: https://github.com/389ds/389-ds-base/issues/6825

Reviewed by:

### DIFF
--- a/src/lib389/lib389/cli_conf/plugins/rootdn_ac.py
+++ b/src/lib389/lib389/cli_conf/plugins/rootdn_ac.py
@@ -8,7 +8,7 @@
 
 import socket
 from lib389.plugins import RootDNAccessControlPlugin
-from lib389.utils import is_valid_hostname
+from lib389.utils import is_valid_hostname, is_valid_ip
 from lib389.cli_conf import add_generic_plugin_parsers, generic_object_edit
 from lib389.cli_base import CustomHelpFormatter
 
@@ -62,19 +62,13 @@ def validate_args(args):
 
     if args.allow_ip is not None:
         for ip in args.allow_ip:
-            if ip != "delete":
-                try:
-                    socket.inet_aton(ip)
-                except socket.error:
-                    raise ValueError(f"Invalid IP address ({ip}) for '--allow-ip'")
+            if ip != "delete" and not is_valid_ip(ip):
+                raise ValueError(f"Invalid IP address ({ip}) for '--allow-ip'")
 
     if args.deny_ip is not None and args.deny_ip != "delete":
         for ip in args.deny_ip:
-            if ip != "delete":
-                try:
-                    socket.inet_aton(ip)
-                except socket.error:
-                    raise ValueError(f"Invalid IP address ({ip}) for '--deny-ip'")
+            if ip != "delete" and not is_valid_ip(ip):
+                raise ValueError(f"Invalid IP address ({ip}) for '--deny-ip'")
 
     if args.allow_host is not None:
         for hostname in args.allow_host:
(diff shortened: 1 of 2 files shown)


droideck (changes requested):
The function needs improvement as it validates something that it shouldn't (and other way around). Checked with this:

```python
import pytest

class TestIsValidIP:
    @pytest.mark.parametrize("ip", [
        "192.168.1.1",
        "0.0.0.0",
        "255.255.255.255",
        "10.0.0.1",
        "172.16.0.1",
    ])
    def test_valid_ipv4(self, ip):
        """Test valid IPv4 addresses"""
        assert is_valid_ip(ip) is True

    @pytest.mark.parametrize("pattern", [
        "192.168.*.*",
        "*.*.*.1",
        "*.*.*.*",
        "10.0.*.1",
        "192.168.1.*",
    ])
    def test_valid_ipv4_wildcards(self, pattern):
        """Test valid IPv4 addresses with wildcards"""
        assert is_valid_ip(pattern) is True

    @pytest.mark.parametrize("ip", [
        "192.168.1.256",  # Out of range
        "192.168.1",      # Too few octets
        "192.168",        # Too few octets
        "192",            # Too few octets
        "192.168.1.1.1",  # Too many octets
        "192.168.-1.1",   # Negative number
        "192.168.a.1",    # Non-numeric
        "",               # Empty string
        "   ",            # Whitespace only
    ])
    def test_invalid_ipv4(self, ip):
        """Test invalid IPv4 addresses"""
        assert is_valid_ip(ip) is False

    @pytest.mark.parametrize("pattern", [
        "192.168.*",      # Too few octets
        "*.*.256.*",      # Out of range octet
        "192.**.1.1",     # Multiple asterisks
        "192.1*.1.1",     # Mixed content in octet
        "*192.168.1.1",   # Asterisk not alone
    ])
    def test_invalid_wildcards(self, pattern):
        """Test invalid wildcard patterns"""
        assert is_valid_ip(pattern) is False

    @pytest.mark.parametrize("ip", [
        "::1",
        "2001:db8::1",
        "fe80::1",
        "2001:0db8:0000:0000:0000:0000:0000:0001",
        "::ffff:192.168.1.1",  # IPv4-mapped IPv6
    ])
    def test_valid_ipv6(self, ip):
        """Test valid IPv6 addresses"""
        assert is_valid_ip(ip) is True

    @pytest.mark.parametrize("ip", [
        ":::",
        "2001:db8::1::2",  # Multiple ::
        "gggg::1",         # Invalid hex
        "2001:db8:1",      # Incomplete
    ])
    def test_invalid_ipv6(self, ip):
        """Test invalid IPv6 addresses"""
        assert is_valid_ip(ip) is False
```

droideck (approved):
Looks good! Ack

PR #6831, opened 2025-06-25
Bump version to 3.1.3

vashirov:


src/lib389/pyproject.toml
```diff
@@ -16,7 +16,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lib389"
-version = "3.1.2"  # Should match the main 389-ds-base version
+version = "3.1.3"  # Should match the main 389-ds-base version
```
vashirov, line 19: @droideck thanks for `validate_version.py`, I stepped right into it :D

droideck: Nice! Tested in action!:D

droideck (approved):
LGTM

PR #6832, opened 2025-06-25
Issue 6810 - Deprecate PAM PTA plugin configuration attributes in base entry

### LINKED ISSUE #6810: Deprecate PAM PTA plugin configuration attributes in base entry
**Issue Description**
Previous to 2012 you could only configure the plugin through the main plugin entry under cn=config:
      
      cn=PAM Pass Through Auth,cn=plugins,cn=config

But after 2012 we added "config" child entries under the main plugin entry:

     cn=<config>, cn=PAM Pass Through Auth,cn=plugins,cn=config

For backwards compatibility, users could use the main plugin entry although the child entry approach is
what we wanted to use. To make life easier for CLI/UI and to be consistent with other plugins, we are
deprecating the use of main plugin config entries in favour of using child config entries. (2025)

 - Package and version: Main branch back to 1.4.3

**Steps to Reproduce**
1. Create an instance, enable the PAM PTA plugin, restart instance
2. Add a child entry that excludes suffix ou=people,dc=example,dc=com
3. Add userA under ou=people,dc=example,dc=com suffix 
4. Attempt a bind with userA
5. Notice PAM PTA plugin warnings in the DS error logs

**Expected results**
userA suffix should be excluded from PAM PTA scope so should see no plugin warnings in error logs

**Additional context**
User survey in 389-users@lists.fedoraproject.org mailing list (How are you using the PAM PTA plugin?)

jchapma:
Description: For backwards compatibility, users could use the main plugin entry although the child entry approach is what we wanted to use. To make life easier for CLI/UI and to be consistent with other plugins, we are deprecating the use of main plugin config entries in favour of using child config entries.

Fix: On server startup an upgrade function checks the PTA parent entry for config entries. If it finds any it creates a child entry and populates it with discovered config attrs.

Fixes: https://github.com/389ds/389-ds-base/issues/6810

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/plugins/pam_pta_test.py
+++ b/dirsrvtests/tests/suites/plugins/pam_pta_test.py
@@ -1,0 +1,284 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import pwd
+import pytest
+import os
+import shutil
+from lib389.utils import *
+from lib389.topologies import topology_st
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.plugins import PAMPassThroughAuthPlugin, PAMPassThroughAuthConfigs
+from lib389.idm.user import UserAccounts, DEFAULT_BASEDN_RDN
+
+pytestmark = pytest.mark.tier1
+
+DEBUGGING = os.getenv('DEBUGGING', False)
+
+logging.getLogger(__name__).setLevel(logging.DEBUG)
+log = logging.getLogger(__name__)
+
+PAM_PTA_PLUGIN_DN = 'cn=PAM Pass Through Auth,cn=plugins,cn=config'
+PAM_PTA_MIGRATED_DN = "cn=default,cn=PAM Pass Through Auth,cn=plugins,cn=config"
+PLUGIN_ENABLED = 'nsslapd-pluginEnabled'
+SYSTEM_USER = 'pam_pta_user'
+SYSTEM_PSWD = 'Secret123'
+
+@pytest.fixture(scope='module')
+def system_user():
+    """Create a local system user and yield its details."""
+
+    try:
+        user_info = pwd.getpwnam(SYSTEM_USER)
+    except KeyError:
+        log.info("Create system user, set password, add permissions")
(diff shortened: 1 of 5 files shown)


jchapma:
It seems there are some issues running these tests in a container, probably related to the creation of the system user that is used for pamService system-auth. I will look into this in the morning.

dirsrvtests/tests/suites/plugins/pam_pta_test.py
```diff
@@ -0,0 +1,294 @@
...
+import pytest
+import os
+import shutil
+from lib389.utils import *
+from lib389.topologies import topology_st
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.plugins import PAMPassThroughAuthPlugin, PAMPassThroughAuthConfigs
+from lib389.idm.user import UserAccounts, DEFAULT_BASEDN_RDN
+
+pytestmark = pytest.mark.tier1
```
vashirov: This looks like a destructive test case that can potentially harm the system. So let's remove this pytest.mark, and run it only when env variable is present similar to disk monitoring tests: 
https://github.com/389ds/389-ds-base/blob/c6fc373283c0cf5957b7a82d6bebd516bee9d503/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_test.py#L24

jchapma:
rebase

ldap/servers/slapd/upgrade.c
```diff
@@ -330,6 +408,182 @@ upgrade_remove_subtree_rename(void)
...
+        slapi_log_error(SLAPI_LOG_ERR, "upgrade_pam_pta_default_config",
+                        "Base entry not found: %s\n", parent_dn_str);
+
+        slapi_sdn_free(&parent_sdn);
+        return UPGRADE_FAILURE;
+    }
+
+    /* Check if the base entry contains config attrs */
+    pb = slapi_pblock_new();
+    rc = get_pam_default_attrs(parent_dn_str, &present_attrs, &num_attrs);
```
tbordaz: Why not using the already retrieved parent_dn_str (base_entry) it prevents a second search

tbordaz (approved):
Except a minor comment LGTM

ldap/servers/slapd/upgrade.c
jchapma: Good question :)

Refactored to remove the second internal search and extract the config attrs from the base entry directly. Validated with PTA plugin using  Win AD server as auth backend, ran asan, no leaks.

jchapma:
All expected CI tests are passing, so I will merge this PR

PR #6837, opened 2025-06-30
Issue 6753 - Port ticket test 48294 & 48295

mirielka:
Description: Porting ticket 48294 test to
dirsrvtests/tests/suites/plugins/linked_attributes_test.py

Relates: #6753

Author: Lenka Doudova

Reviewed by: ???


dirsrvtests/tests/suites/plugins/linked_attributes_test.py
```diff
@@ -0,0 +1,208 @@
...
+    la_configs = LinkedAttributesConfigs(topology_st.standalone)
+    config = la_configs.create(properties={'cn': 'Manager Link',
+                                  'linkType': LINKTYPE,
+                                  'managedType': MANAGEDTYPE})
+
+    def fin():
+        log.info('Cleaning up Linked Attributes plugin')
+        config.delete()
+        linkedattr.disable()
+        linkedattr.delete()
```
droideck: Why do you need to remove `cn=Linked Attributes,cn=plugins,cn=config`?

dirsrvtests/tests/suites/plugins/linked_attributes_test.py
```diff
@@ -0,0 +1,208 @@
...
+    return employee
+
+
+@pytest.fixture(scope='function')
+def setup_linked_attributes(topology_st, request):
+    """Fixture to set up the Linked Attributes plugin."""
+    log.info('Setting up Linked Attributes plugin')
+
+    log.info('Enable Dynamic plugins and the Linked Attributes plugin')
+    topology_st.standalone.config.replace('nsslapd-dynamic-plugins', 'on')
```
droideck: IIRC, we don't support dynamic plugins, and it's mostly for debugging purposes.
I think we'd better test without it (hence, a restart is required here).

droideck (approved):
Besides a couple of minor issues, looks good!

dirsrvtests/tests/suites/plugins/linked_attributes_test.py
mirielka: Thank you, will consult a bit more to see which way is more preferable.

dirsrvtests/tests/suites/plugins/linked_attributes_test.py
mirielka: It seems appropriate to clean up everything that was created within the test. In case a new test is added to a file, the author either has to use new names or add a cleanup commands to the old test (I also have to do this sometimes when porting ticket tests to existing suites). It does not help reusability of code if the test fails right in the beginning because it tries to create an entry that already exists. This way any new test can easily use new clean setup even with reused fixture/function.

droideck: This line does nothing because we '_protect' the plugin entries. It's probably a typo you have there.

dirsrvtests/tests/suites/plugins/linked_attributes_test.py
vashirov: @droideck Dynamic plugins are supported, just not enabled by default.

Issue #6800, opened 2025-05-28
Check for minimal supported Python version


vashirov:
**Issue Description**
This issue is a follow up for https://github.com/389ds/389-ds-base/issues/5805 and https://github.com/389ds/389-ds-base/pull/6775#issuecomment-2875285603

We should add a GH action to automatically check for the minimal supported version, for example using https://pypi.org/project/vermin/

vashirov:
Initial run:
```
# uv tool run vermin  --target=3.8 ds/src/lib389
File with incompatible versions: /root/ds/src/lib389/lib389/instance/options.py
  Versions could not be combined: 2.3, !3 and !2, 3.6
File with incompatible versions: /root/ds/src/lib389/lib389/paths.py
  fstrings (requires !2, 3.6) vs. 'ConfigParser' module (requires 2.0, !3)
File with incompatible versions: /root/ds/src/lib389/lib389/tools.py
  Versions could not be combined: 2.7, !3 and !2, 3.0
File with incompatible versions: /root/ds/src/lib389/lib389/utils.py
  fstrings (requires !2, 3.6) vs. 'popen2' module (requires 2.0, !3)
Note: Some files had incompatible versions so the results might not be correct!
Tips:
- Generic or literal annotations might be in use. If so, try using: --eval-annotations
  But check the caveat section: https://github.com/netromdk/vermin#caveats
- You're using potentially backported modules: argparse, configparser, enum, statistics, typing
  If so, try using the following for better results: --backport argparse --backport configparser --backport enum --backport statistics --backport typing
- Since '# novm' or '# novermin' weren't used, a speedup can be achieved using: --no-parse-comments
(disable using: --no-tips)

Minimum required versions: 3.11
Incompatible versions:     2
Target versions not met:   3.8
```

PR #6840, opened 2025-07-01
Issue 6800 - Check for minimal supported Python version

vashirov:
Description:
Add a new check for minimal supported Python version for lib389 and dirsrvtests

Fixes:
https://github.com/389ds/389-ds-base/issues/6800

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,4 +1,4 @@
-name: Validate tests
+name: Validate tests and minimal Python version
 
 on:
   push:
@@ -25,3 +25,11 @@ jobs:
       - name: Check for duplicate IDs
         if: always()
         run: python3 dirsrvtests/check_for_duplicate_ids.py dirsrvtests/tests/suites
+
+      - name: Check for minimal Python version for lib389
+        if: always()
+        run: uv tool run vermin --target=3.8 src/lib389
+
+      - name: Check for minimal Python version for dirsrvtests
+        if: always()
+        run: uv tool run vermin --target=3.8 dirsrvtests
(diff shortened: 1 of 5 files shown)


vashirov:
`Validate` pipeline will fail until https://github.com/389ds/389-ds-base/pull/6839 is merged. It also reports a bunch of python2 vs python3 incompatibilities, we should clean them up in a separate PR.

vashirov:
Actually, there is not many of them, added another commit that drops the remaining Python 2 compatibility workarounds.

progier389:
I wonder if we should not better write an utility class for these one. with different implementation according to the python version otherwise someone may see deprecated warning when using lib389  on python3 and change thing back !

src/lib389/lib389/perftools.py
```diff
@@ -398,7 +398,8 @@ def ldclt(self, measure_name, args, nbThreads=10, nbMes=10):
         tmout = 30+10*nbMes
         print (f"Running ldclt with a timeout of {tmout} seconds ...\r")
         try:
-            result = subprocess.run(args=cmd, capture_output=True, timeout=tmout)
+            result = subprocess.run(args=cmd, stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE, timeout=tmout)
```
droideck: `capture_output=True` is available since `Python 3.7` and it's quite a modern way. Why change it?

droideck (approved):
Minor issue, otherwise, - looks good!

src/lib389/lib389/perftools.py
vashirov: I had other instances of `capture_output` that I replaced, but then reverted. This one I missed :)

vashirov:
@progier389, which ones?

PR #6842, opened 2025-07-01
Issue 6841 - Cancel Actions when PR is updated

### LINKED ISSUE #6841: Cancel Actions when PR is updated
**Issue Description**
GH Actions take some time to run, and updating PR before the previous run has completed makes the new runs wait.

Instead we should cancel the previous runs for this PR to save time and don't waste computing resources.

vashirov:
Description
GH Actions take some time to run, and updating PR before the previous run has completed makes the new runs wait.

Instead we should cancel the previous runs for this PR to save time and don't waste computing resources.

Fixes: https://github.com/389ds/389-ds-base/issues/6841


droideck (approved):
LGTM!

PR #6856, opened 2025-07-09
Issue 6843 - Add CI tests for logconv.py

### LINKED ISSUE #6843: Add CI tests for logconv.py
**Issue Description**
Add CI tests for logconv.py

 - Package and version: main branch back to 3.0

jchapma:
Description: logconv.py was added in 389-ds-base-3.0, we need some CI tests to help validate the upgrade of logconv.py to support JSON formatted access logs.

Fixes: https://github.com/389ds/389-ds-base/issues/6843

Reviewed by:

### DIFF
--- a/dirsrvtests/tests/suites/logging/logconv_test.py
+++ b/dirsrvtests/tests/suites/logging/logconv_test.py
@@ -237,7 +237,7 @@ class TestLogconv:
             2. Success
             3. Success
             4. Success
-            4. Actual stats match expected
+            5. Actual stats match expected
         """
         self.truncate_logs()
 
--- a/dirsrvtests/tests/suites/logging/logconv_test.py
+++ b/dirsrvtests/tests/suites/logging/logconv_test.py
@@ -1,0 +1,992 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import ldap
+import logging
+import os
+import pytest
+import re
+import shutil
+import subprocess
+import time
+
+from lib389._constants import *
+from lib389.idm.account import Anonymous, Accounts
+from lib389.idm.user import UserAccounts
+from lib389.topologies import topology_st
+from lib389.utils import ensure_str
+from ldap.controls.vlv import VLVRequestControl
+from ldap.controls.sss import SSSRequestControl
+from ldap.controls import SimplePagedResultsControl
+from ldap.controls.psearch import PersistentSearchControl
+from lib389.backend import DatabaseConfig
+from lib389.dbgen import dbgen_users
+from lib389._mapped_object import DSLdapObjects
+
+log = logging.getLogger(__name__)
+
+# Path to the logconv script (Change to logconv.pl for comparison).
+LOGCONV_SCRIPT = "logconv.py"
+ # If True, each test validates logconv output individually; otherwise, final summary test runs validation.
+VALIDATE_EACH_TEST = True
+# Expected summary values used for validation when VALIDATE_EACH_TEST is False.
+VALIDATE_ALL_SUMMARY = {}
+
+@pytest.fixture(scope="module", autouse=True)
+def get_access_log_path(topology_st):
+    """ Truncate logs before running tests to remove startup noise. """
+    inst = topology_st.standalone
+    path = inst.config.get_attr_val_utf8('nsslapd-accesslog')
+    inst.config.set('nsslapd-accesslog-logbuffering', 'off')
+    if not path:
+        pytest.fail("Access log path not found.")
+    with open(path, "w") as f:
+        log.info("Truncating access logs after startup.")
+        f.truncate(0)
+    yield path
+
+# Utilities
+def truncate_access_log(path):
+    """ Truncate logs if each test validation is enabled. """
+    if not path:
+        raise ValueError("Access log path not found.")
+    if VALIDATE_EACH_TEST:
+        with open(path, "w") as f:
+            f.truncate(0)
+        return path
+    else:
+        return path
+
+def get_logconv_path():
+    """ Get the location of logconv script."""
+    return shutil.which(LOGCONV_SCRIPT) or None
+
+def run_logconv(access_log_path: str, debug_output=False):
+    """ Run logconv script, if debug is enabled the output is written to file."""
+    LOGCONV_OUTPUT = 'logconv-debug.log'
+    logconv_path = get_logconv_path()
+
+    if not logconv_path:
+        raise FileNotFoundError(f"{LOGCONV_SCRIPT} not found.")
+
+    result = subprocess.run(
+        [logconv_path, access_log_path],
+        capture_output=True,
+        text=True
+    )
+    if result.returncode != 0:
+        raise RuntimeError(f"{LOGCONV_SCRIPT} failed: {result.stderr}")
+
+    summary = result.stdout
+
+    if debug_output:
+        with open(LOGCONV_OUTPUT, 'w') as f:
+            f.write(summary)
+        os.chmod(LOGCONV_OUTPUT, 0o644)
+
+    return summary
+
+def parse_summary(output: str):
+    """ Key patterns from logconv default output. """
+    patterns = {
+        # Server stats
+        "restarts": r"^Restarts:\s+(\d+)",
+
+        # Connections
+        "concurrent_conns": r"^Peak Concurrent connections:\s+(\d+)",
+        "total_connections": r"^Total connections:\s+(\d+)",
+        "ldap_conns": r"- LDAP connections:\s+(\d+)",
+        "ldapi_conns": r"- LDAPI connections:\s+(\d+)",
+        "ldaps_conns": r"- LDAPS connections:\s+(\d+)",
+
+        # Operations
+        "operations": r"^Total Operations:\s+(\d+)",
+        "results": r"^Total Results:\s+(\d+)",
+        "searches": r"^Searches:\s+(\d+)",
+        "mods": r"^Modifications:\s+(\d+)",
+        "adds": r"^Adds:\s+(\d+)",
+        "deletes": r"^Deletes:\s+(\d+)",
+        "modrdns": r"^Mod RDNs:\s+(\d+)",
+        "compares": r"Compares:\s+(\d+)",
+        "binds": r"^Binds:\s+(\d+)",
+        "unbinds": r"^Unbinds:\s+(\d+)",
+        "autobinds": r"AUTOBINDs\(LDAPI\):\s+(\d+)",
+        "sasl_binds": r"SASL Binds:\s+(\d+)",
+        "anon_binds": r"Anonymous Binds:\s+(\d+)",
+        "ssl_client_binds": r"SSL Client Binds\s+(\d+)",
+        "ssl_client_bind_failed": r"Failed SSL Client Binds:\s+(\d+)",
+
+        # Advanced operations
+        "persistent_searches": r"^Persistent Searches:\s+(\d+)",
+        "extended_ops": r"^Extended Operations:\s+(\d+)",
+        "internal_ops": r"^Internal Operations:\s+(\d+)",
+        "abandoned": r"^Abandoned Requests:\s+(\d+)",
+        "proxied_auth_ops": r"^Proxied Auth Operations:\s+(\d+)",
+        "vlv_operations": r"^VLV Operations:\s+(\d+)",
+        "vlv_unindexed_searches": r"VLV Unindexed Searches:\s+(\d+)",
+        "sort_operations": r"^SORT Operations:\s+(\d+)",
+        "paged_searches": r"^Paged Searches:\s+(\d+)",
+
+        # Errors and filter issues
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,992 @@
...
+        "unindexed_components": 0,
+        "operations": 2,
+        "results": 1
+    }
+    run_and_validate(expected, access_log, test_name="test_partially_unindexed_vlv_sort")
+
+def test_fully_unindexed_search(topology_st, get_access_log_path):
+    """Test that a fully unindexed search is correctly identified by logconv.
+
+    :id: 71f93a9f-3f9d-450e-ae98-79fabe5f21f4
```
droideck: Duplicate UUID

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,992 @@
...
+        2. StartTLS search completes without errors.
+        3. logconv returns correct updated stats.
+        4. Expected stats match logconv output.
+    """
+    access_log = truncate_access_log(get_access_log_path)
+
+    cmd = [
+        "ldapsearch",
+        "-x",
+        "-H", f"ldap://localhost:38901",
```
droideck: I think the port shouldn't be hardcoded (in case, it's changed in the future)

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,992 @@
...
+    inst = topology_st.standalone
+    access_log = truncate_access_log(get_access_log_path)
+
+    # Force an unindexed search
+    db_cfg = DatabaseConfig(inst)
+    default_idlistscanlimit = db_cfg.get_attr_vals_utf8('nsslapd-idlistscanlimit')
+    db_cfg.set([('nsslapd-idlistscanlimit', '100')])
+    users = UserAccounts(inst, DEFAULT_SUFFIX)
+    for i in range(101):
+        users.create_test_user(uid=i)
```
droideck: I think it should be done with a fixture (and cleaned up accordingly)

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,992 @@
...
+    raw_objects = DSLdapObjects(inst, basedn=DEFAULT_SUFFIX)
+    raw_objects.filter("(description=test*)")
+    db_cfg.set([('nsslapd-idlistscanlimit', default_idlistscanlimit)])
```
droideck: It'll be safer to have it in `finally` block

droideck (changes requested):
The CI fails (at least, I see it in BDB case). Probably, it's a bit flaky.

Additionally, I think we need some negative tests...
Something along these error conditions (at least to some basic degree):
- What happens when logconv.py fails?
- How are malformed log entries handled?
- Edge cases like empty or corrupted log files?

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,980 @@
...
+        _test_ldapi,
+        _test_restart,
+        _test_invalid_filter,
+        _test_partially_unindexed_vlv_sort,
+        _test_fully_unindexed_search,
+        _test_ldaps,
+        _test_starttls,
+        _test_missing_access_log,
+        _test_invalid_access_log_format
+    ]
```
droideck: what is the reason for such test suite structure?
It'll be harder to maintain, the output will be not in the expected pytest format, and it's hard to run one test without running all of them.

Pytest allows for test classes and will provide correct output:

So, I think, it should be something like this:

```python
# Each test should be independent
class TestLogConv:
    @pytest.mark.parametrize("log_format", ["default", "json"])
    def test_add_delete_operations(self, fresh_instance, log_format):
        # Test only add/delete
        
    @pytest.mark.parametrize("log_format", ["default", "json"])
    def test_modify_operations(self, fresh_instance, log_format):
        # Test only modify
```

Or just plainly without `class TestLogConv:`

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,980 @@
...
+            time.sleep(poll_interval)
+
+@pytest.fixture
+def fresh_instance():
+    topo = create_topology({ReplicaRole.STANDALONE: 1})
+    inst = topo.standalone
+    inst.start()
+    yield inst
+    inst.stop()
+    inst.delete(escapehatch="i am sure")
```
droideck: It needs a finalizer.

So, I think, it's better to reuse a usual topology approach:

```python


def topology_st(request):
    """Create DS standalone instance"""

    topology = create_topology({ReplicaRole.STANDALONE: 1})

    def fin():
        if topology.standalone.exists():
            topology.standalone.delete()
    request.addfinalizer(fin)

    return topology

File: dirsrvtests/tests/suites/logging/logconv_test.py [hunk 7]
```diff
@@ -0,0 +1,980 @@
...
+            except (FileNotFoundError, RuntimeError) as e:
+                log.error(f"{test_name}: Error running logconv: {e}")
+                raise
+
+            summary = self.parse_summary(output)
+            for key, existing_val in summary.items():
+                if key in IGNORE_KEYS:
+                    continue
+                expected_val = expected.get(key, 0)
+                if existing_val < expected_val:
```
droideck: What about `if existing_val > expected_val` case? In case of duplicates, or when we expect 0, or some other bug?

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,980 @@
...
+        test_env.inst.config.set('nsslapd-ldapifilepath', f'/var/run/slapd-{test_env.inst.serverid}.socket')
+        test_env.inst.restart()
+        ldapi_socket = test_env.inst.config.get_attr_val_utf8('nsslapd-ldapifilepath').replace('/', '%2F')
+        cmd = [
+            "ldapsearch",
+            "-H", f"ldapi://{ldapi_socket}",
+            "-Y", "EXTERNAL",
+            "-b", "dc=example,dc=com"
+        ]
+        subprocess.run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
droideck, line 679: Not capturing output makes debugging failures difficult (in case, something will happen during the search - we can see the issue right away).

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,980 @@
...
+
+def _test_add_del(test_env):
+
+    def _add_del_users(test_env):
+        users = UserAccounts(test_env.inst, DEFAULT_SUFFIX)
+        try:
+            for idx in range(3):
+                users.create_test_user(uid=str(idx), gid=str(idx))
+        except ldap.ALREADY_EXISTS as e:
+            log.warning(f"_test_add_del - User already exists: {e}")
```
droideck, line 250: Do we really need it?
I think we expect the test to be isolated...
But it's a minor issue, I don't have a strong preference.

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,980 @@
...
+        "fds_taken": 1,
+        "fds_returned": 1
+    }
+
+    test_env.truncate_logs()
+    test_env.run_operation(test_env.operation)
+
+def _test_ldapi(test_env):
+
+    def _ldapi_search(tetest_envt):
```
droideck: **tetest_envt** typo:)

droideck (changes requested):

dirsrvtests/tests/suites/logging/logconv_test.py
jchapma: > what is the reason for such test suite structure? It'll be harder to maintain, the output will be not in the expected pytest format, and it's hard to run one test without running all of them.
> 
> Pytest allows for test classes and will provide correct output:
> 
> So, I think, it should be something like this:
> 
> ```python
> # Each test should be independent
> class TestLogConv:
>     @pytest.mark.parametrize("log_format", ["default", "json"])
>     def test_add_delete_operations(self, fresh_instance, log_format):
>         # Test only add/delete
>         
>     @pytest.mark.parametrize("log_format", ["default", "json"])
>     def test_modify_operations(self, fresh_instance, log_format):
>         # Test only modify
> ```
> 
> Or just plainly without `class TestLogConv:`

I guess the main reason was because the sequence of each test was the same, truncate the logs, perform some operation, run logconv.py and compare the actual with the expected. I didn't like the idea of using a fixture for each logical step so I encapsulated all steps in a single class.  A single test_ function is used as a test runner to iterate over all the defined tests, which all have a defined structure, making it easy to extend,  well In my opinion. I did consider the option of defining each test separately but then I though when would we need to run a single test in isolation.

To be honest I didnt realise pytest has a  test class, let me look into that.

dirsrvtests/tests/suites/logging/logconv_test.py
jchapma: > It needs a finalizer.
> 
> So, I think, it's better to reuse a usual topology approach:
> 
> ```python
> @pytest.fixture(scope="function")
> def topology_st(request):
>     """Create DS standalone instance"""
> 
>     topology = create_topology({ReplicaRole.STANDALONE: 1})
> 
>     def fin():
>         if topology.standalone.exists():
>             topology.standalone.delete()
>     request.addfinalizer(fin)
> 
>     return topology
> ```

Yep, that is better

dirsrvtests/tests/suites/logging/logconv_test.py
jchapma: > What about `if existing_val > expected_val` case? In case of duplicates, or when we expect 0, or some other bug?

Initially the expected_val had to be == the expected_val for a test to pass, but with non test activity n the logs from plugins, etc made it very tricky. The best I could come up with is a form of "loose" validation where the actual matched or exceed the expected. Any suggestions on improving this ?

dirsrvtests/tests/suites/logging/logconv_test.py
jchapma: > Not capturing output makes debugging failures difficult (in case, something will happen during the search - we can see the issue right away).

Good point

dirsrvtests/tests/suites/logging/logconv_test.py
jchapma: > Do we really need it? I think we expect the test to be isolated... But it's a minor issue, I don't have a strong preference.

In a previous version all test for both formats were run on the same instance, so I added this to be sure. But in this version we do not.

dirsrvtests/tests/suites/logging/logconv_test.py
droideck: A few options I see:
- Categorize Keys by Validation Strategy (some strict, some more tolerant)
- Validate with a percentage tolerance for certain metrics
- Get baseline counts before the test and account for that

I'd say the final solution should be a combination of that.
This is the most important part of the test, so I think it's worth spending time calibrating.

jchapma:
Hi @droideck, I think I have addressed all your concerns, if you get a chance could you re-review this PR. Thanks

droideck:
> Hi @droideck, I think I have addressed all your concerns, if you get a chance could you re-review this PR. Thanks

Hi!
I noticed that some non-JSON tests are still failing:

```
FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_starttls[default]
FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_invalid_access_log_format[default]
FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_ldaps[default]
```

Are these failures expected at this stage? I held off on following up as I thought you might still be investigating these issues.
Apologies for the lack of communication on my end...

jchapma:
> > Hi @droideck, I think I have addressed all your concerns, if you get a chance could you re-review this PR. Thanks
> 
> Hi! I noticed that some non-JSON tests are still failing:
> 
> ```
> FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_starttls[default]
> FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_invalid_access_log_format[default]
> FAILED dirsrvtests/tests/suites/logging/logconv_test.py::TestLogconv::test_ldaps[default]
> ```
> 
> Are these failures expected at this stage? I held off on following up as I thought you might still be investigating these issues. Apologies for the lack of communication on my end...

Hey, ye these are expected as there is a bug in the original version. No worries. Thanks

dirsrvtests/tests/suites/logging/logconv_test.py
```diff
@@ -0,0 +1,1094 @@
...
+            "operations": 1,
+            "paged_searches": 1,
+            "results": 1
+        }
+
+        output = self.run_logconv()
+        logconv_stats = self.extract_logconv_stats(output)
+        assert self.validate_logconv_stats(expected, logconv_stats, "test_paged_search")
+
+    def test_persistant_search(self):
```
droideck:
```suggestion
    def test_persistent_search(self):
```

droideck: And more typos like this in the test (doc string and the end of the test)

droideck (approved):
LGTM!

PR #6858, opened 2025-07-09
Issue 6857 - uiduniq: allow specifying match rules in the filter

### LINKED ISSUE #6857: Allow Uniqueness plugin to search uniqueness attributes using custom matching rules
Uniqueness plugin builds an LDAP filter to search for duplicate values using list of uniqueness attributes per configuration. The search then uses existing matching rules defined for these attributes to match values captured in LDAP ADD operation.

Sometimes comparison needs to be a bit wider than the original match. For example, case-insensitive match would be needed to capture duplicates for already existing attributes with exact matches.

It would be nice to specify attributes and their custom match rules in the plugin configuration.

For example,
```
uniqueness-attribute-name: fooName:caseIgnoreMatch:
```
would allow uniqueness plugin to build a filter that includes `(|(fooName:caseIgnoreMatch:=<some value>)(..))`.

Right now this syntax is accepted without any issue but then fails to apply because the plugin in pre-add callback will attempt to iterate over the configured attributes and retrieve them from the added mods. Obviously, `fooName:caseIgnoreMatch:` attribute name will not match any attribute in the ADD mods.

abbra:
Allow uniqueness plugin to work with attributes where uniqueness should be enforced using different matching rule than the one defined for the attribute itself.

Since uniqueness plugin configuration can contain multiple attributes, add matching rule right to the attribute as it is used in the LDAP rule (e.g. 'attribute:caseIgnoreMatch:' to force 'attribute' to be searched with case-insensitive matching rule instead of the original matching rule.

Fixes: https://github.com/389ds/389-ds-base/issues/6857

### DIFF
--- a/ldap/servers/plugins/uiduniq/uid.c
+++ b/ldap/servers/plugins/uiduniq/uid.c
@@ -1030,7 +1030,14 @@ preop_add(Slapi_PBlock *pb)
     }
 
     for (i = 0; attrNames && attrNames[i]; i++) {
+        char *attr_match = strchr(attrNames[i], ':');
+        if (attr_match != NULL) {
+            attr_match[0] = '\0';
+        }
         err = slapi_entry_attr_find(e, attrNames[i], &attr);
+        if (attr_match != NULL) {
+            attr_match[0] = ':';
+        }
         if (!err) {
             /*
                  * Passed all the requirements - this is an operation we


mreynolds389:
This accomplishes one part of the issue.  What I've found is when the plugin does an internal search the matching rule is not detected and is ignored.  But an external search (ldapsearch) using the same filter does have its matching rule applied.  I'll be looking into this...

mreynolds389:
> This accomplishes one part of the issue. What I've found is when the plugin does an internal search the matching rule is not detected and is ignored. But an external search (ldapsearch) using the same filter does have its matching rule applied. I'll be looking into this...

This diff gets everything working, but I will open a new PR to add this fix separately

```
diff --git a/ldap/servers/slapd/str2filter.c b/ldap/servers/slapd/str2filter.c
index 9fdc500f7..0ed0af24d 100644
--- a/ldap/servers/slapd/str2filter.c
+++ b/ldap/servers/slapd/str2filter.c
@@ -344,6 +344,13 @@ str2simple(char *str, int unescape_filter)
                 return NULL; /* error */
             } else {
                 f->f_choice = LDAP_FILTER_EXTENDED;
+                if (f->f_mr_oid) {
+                    rc = plugin_mr_filter_create(&f->f_mr);
+                    if (rc) {
+                        slapi_filter_free(f, 1);
+                        return NULL; /* error */
+                    }
+                }
             }
         } else if (str_find_star(value) == NULL) {
             f->f_choice = LDAP_FILTER_EQUALITY;
```

abbra:
Wonderful! Thank you for the quick fix. 

There are few tests failing, are they related or known ones?

mreynolds389 (approved):
LGTM, tests pass for me.

Issue #6859, opened 2025-07-09
str2filter is not fully applying matching rules


mreynolds389:
**Issue Description**

When we have an extended filter, one with a MR applied, it is ignored during internal searches:

    "(cn:CaseExactMatch:=Value)"

For internal searches we use str2filter() and it doesn't fully apply extended search matching rules

mreynolds389:
b2e82dc3f..1deedd9a8  389-ds-base-3.0 -> 389-ds-base-3.0
119f9a7e5..41ce6414d  389-ds-base-2.7 -> 389-ds-base-2.7
3de8fe82b..ac1535cec  389-ds-base-2.6 -> 389-ds-base-2.6
4ab587655..94ff751fa  389-ds-base-2.5 -> 389-ds-base-2.5
d2a421f35..7d0cb0d9f  389-ds-base-2.4 -> 389-ds-base-2.4
a2805776d..9afb1c1fd  389-ds-base-2.3 -> 389-ds-base-2.3
daa634ee4..98303a4e9  389-ds-base-2.2 -> 389-ds-base-2.2
9611ce9aa..3c9dc4c2e  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6860, opened 2025-07-09
Issue 6859 - str2filter is not fully applying matching rules

mreynolds389:
Description:

When we have an extended filter, one with a MR applied, it is ignored during internal searches:

    "(cn:CaseExactMatch:=Value)"

For internal searches we use str2filter() and it doesn't fully apply extended search filter matching rules

Relates: https://github.com/389ds/389-ds-base/issues/6857
Relates: https://github.com/389ds/389-ds-base/issues/6859

### DIFF
--- a/dirsrvtests/tests/suites/plugins/attruniq_test.py
+++ b/dirsrvtests/tests/suites/plugins/attruniq_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2021 Red Hat, Inc.
+# Copyright (C) 2025 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/str2filter.c
```diff
@@ -344,6 +344,14 @@ str2simple(char *str, int unescape_filter)
                 return NULL; /* error */
             } else {
                 f->f_choice = LDAP_FILTER_EXTENDED;
+                if (f->f_mr_oid) {
+                    /* apply the MR indexers */
+                    rc = plugin_mr_filter_create(&f->f_mr);
```
droideck, line 349: More of a curiosity, as the area is new to me.
Where `SLAPI_PLUGIN_MR_VALUE` comes from?

IIUC, we need it in `plugin_mr_filter_create`-> `attempt_mr_filter_create`. And `_parse_ext_filter` parses only `&f->f_mr_type, &f->f_mr_oid, &f->f_mr_dnAttrs`

And I don't see it supplied anywhere.

mreynolds389: Sorry I'm not sure what you mean "where it comes from".  I'm not that familiar with it either, but it is "get" and "set" in plugin_mr.c into a mr_private struct and stored in the pblock -> SLAPI_PLUGIN_OBJECT.  Then its used in filter and other matching rule code.  I haven't walked the code to see how it's used - this is just how it appears when I grep the code.

droideck: When `plugin_mr_filter_create(&f->f_mr)` is called, `f->f_mr_value` hasn't been set yet, so `SLAPI_PLUGIN_MR_VALUE` will be uninitialized.

How I see this (maybe I miss something):

1. `_parse_ext_filter()` only parses and sets:
   - `f->f_mr_type`
   - `f->f_mr_oid` 
   - `f->f_mr_dnAttrs`

2. The actual filter value is parsed **later** in the function and stored in `f->f_avvalue` (not `f->f_mr_value`). This happens in the section after your patch where it processes the value for equality/substring/presence.

3. Looking at the filter structure, it uses a union where extended filters share storage with AVA filters.

https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/slap.h#L562-L579

The code stores the value in the AVA part (`f_avvalue`) but `plugin_mr_filter_create()` expects it in the extended part (`f_mr_value`).

Should we, before calling `plugin_mr_filter_create()`, ensure `f->f_mr_value` is properly set?

dirsrvtests/tests/suites/plugins/attruniq_test.py
```diff
@@ -324,4 +324,65 @@ def test_exclude_subtrees(topology_st):
...
+        3. Success
+        4. Success
+        5. Success
+    """
+
+    inst = topology_st.standalone
+
+    attruniq = AttributeUniquenessPlugin(inst,
+                                         dn="cn=attribute uniqueness,cn=plugins,cn=config")
+    attruniq.add_unique_attribute('cn:CaseExactMatch:')
```
droideck, line 356: Should we also add `(cn:CaseExactMatch:=Value)` test - hence, to add the `value` part?

ldap/servers/slapd/str2filter.c
tbordaz: @droideck , my understanding of your question is how 'SLAPI_PLUGIN_*' get initialized.
The point is that any call to 'slapi_pblock_s/get(pb, SLAPI_PLUGIN_...)' is done after call slapi_pblock_set(pb, SLAPI_PLUGIN, mrp) that does the magic.
'mrp' is a matching rule structure pointer. For each supported MR oid, there is a slapdplugin that is created/initialized at startup and stored into a table. Then you retrieve the MR structure through the oid. Then before using SLAPI_PLUGIN_MR_OID/SLAPI_PLUGIN_MR_VALUE..., you need to set a pblock with SLAPI_PLUGIN. Then you can access SLAPI_PLUGIN_* that have already been initialized.

dirsrvtests/tests/suites/plugins/attruniq_test.py
mreynolds389: > Should we also add `(cn:CaseExactMatch:=Value)` test - hence, to add the `value` part?

I don't think that will work as this value is supposed to be the "attribute".  The plugin adds the "value" through its config attribute: 

    uniqueness-attribute-name: krbCanonicalName:CaseIgnoreMatch:
 
This is converted to a search filter in preop_add()/preop_modify():   

    krbCanonicalName:CaseIgnoreMatch:=<value from client's add/mod>

droideck: Okay, got it. I think that's where my previous `mr_value` confusion originated. I wasn't able to find the `uniqueness-attribute-name: matching rule` info in our docs (probably, looked in the wrong place) and misunderstood it.
Sorry for the fuss:)

mreynolds389: I don't think we document anywhere about the matching rule filter option.  Attribute uniqueness plugin did not support it until Alexanders PR that was pushed a few days ago.

tbordaz (approved):
LGTM

PR #6862, opened 2025-07-11
Issue 6594 - Add test for numSubordinates replication consistency with tombstones

### LINKED ISSUE #6594: Incorrect value of the numSubordinates attribute on the new replication server
**Issue Description**
Incorrect value of the numSubordinates attribute on the new replication server if the value of the tombstoneNumSubordinates attribute is not zero

 - Platform: any
 - Package and version: 389ds 3.1.2 and earlier

**Steps to reproduce the behavior:**
To reproduce the problem, you need to: Deploy a 389ds instance and create any object in any container. Then delete this object. Deploy a second 389ds server and configure replication between them. On the second server, the value of the numSubordinates attribute in the used container will not match the value of the primary server. Presumably, the value of the numSubordinates attribute in the used container on the second server will be equal to the sum of the values ​​of the numSubordinates and tombstoneNumSubordinates attributes of the first server.

**Additional context**
As a result, if we need to delete an empty used container on the second server, this will not be possible, since the numSubordinates attribute will not be zero.

droideck:
Description: Add comprehensive test to verify that numSubordinates and tombstoneNumSubordinates attributes are correctly replicated between instances when tombstone entries are present.

Fixes: https://github.com/389ds/389-ds-base/issues/6594

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/replication/numsubordinates_replication_test.py
+++ b/dirsrvtests/tests/suites/replication/numsubordinates_replication_test.py
@@ -1,0 +1,144 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import os
+import logging
+import pytest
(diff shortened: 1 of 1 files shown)


droideck:
The test is a result of upstream triage, as I needed to verify that we don't have the reported issue.

progier389 (approved):
LGTM

Issue #6868, opened 2025-07-11
UI - schema attribute table expansion doesn't work after pagination


mreynolds389:
**Issue Description**

If you move to a new page in the table then the attribute expansion no longer works

PR #6869, opened 2025-07-11
Issue 6868 - UI - schema attribute table expansion break after moving to

mreynolds389:
a new page

Description:

Used the wrong formula to select the expanded row for Attributes

Relates: https://github.com/389ds/389-ds-base/issues/6868


droideck (approved):
Looks good!

Issue #6872, opened 2025-07-15
compressed log rotation creates files with world readable permission


mreynolds389:
**Issue Description**

when the rotation is setup with compression, the compressed files (*.gz) for all the log files get created with 644 instead of 600 permissions, allowing them to be read by anyone. If compression isn't selected permissions are correct.

mreynolds389:
1deedd9a8..94a3f48bc  389-ds-base-3.0 -> 389-ds-base-3.0
41ce6414d..d8e3ac837  389-ds-base-2.7 -> 389-ds-base-2.7
ac1535cec..c0f28fd9e  389-ds-base-2.6 -> 389-ds-base-2.6
94ff751fa..5867881cb  389-ds-base-2.5 -> 389-ds-base-2.5
7d0cb0d9f..b9a27e584  389-ds-base-2.4 -> 389-ds-base-2.4
9afb1c1fd..3e3570545  389-ds-base-2.3 -> 389-ds-base-2.3
98303a4e9..da7f3ebae  389-ds-base-2.2 -> 389-ds-base-2.2

PR #6873, opened 2025-07-15
Issue 6872 - compressed log rotation creates files with world readable permission

mreynolds389:
Description:

When compressing a log file, first create the empty file using PR_Open
so we can set the correct permissions right from the start. gzopen()
always uses permission 644 and that is not safe. So after creating it
with PR_Open(), with the correct permissions, then reopen it with gzopen()
and write the compressed content.

relates: https://github.com/389ds/389-ds-base/issues/6872

### DIFF
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -174,17 +174,28 @@ get_syslog_loglevel(int loglevel)
 }
 
 static int
-compress_log_file(char *log_name)
+compress_log_file(char *log_name, int32_t mode)
 {
     char gzip_log[BUFSIZ] = {0};
     char buf[LOG_CHUNK] = {0};
     size_t bytes_read = 0;
     gzFile outfile = NULL;
     FILE *source = NULL;
+    int fd = 0;
 
     PR_snprintf(gzip_log, sizeof(gzip_log), "%s.gz", log_name);
-    if ((outfile = gzopen(gzip_log,"wb")) == NULL) {
-        /* Failed to open new gzip file */
+
+    /*
+     * Try to open the file as we may have an incorrect path. We also need to
+     * set the permissions using open() as gzopen() creates the file with
+     * 644 permissions (world readable - bad). So we create an empty file with
+     * the correct permissions, then we pass the FD to gzdopen() to write the
+     * compressed content.
+     */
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/log.c
```diff
@@ -193,6 +193,19 @@ compress_log_file(char *log_name)
         gzclose(outfile);
         return -1;
     }
+
+    /* Close the gzip file and set the perms, then start writing to it */
+    gzclose(outfile);
+    if (chmod(gzip_log, mode) != 0) {
```
progier389: I have a concern about that one: 
  The file is still created world writable and someone may open the file before the chmod is done.
  I am afraid that the only safe way is to use umask so that gzopen creates the file with the proper mode
  but as said in the man page, umask is prone to race condition.   Maybe we should have global a 0177 umask and explicitly set the proper mode after creating the files  everywhere else (A least if we forget to change some mode it will not be any more a security issue)

progier389 (changes requested):

ldap/servers/slapd/log.c
mreynolds389: It's not world writable, only readable:

    -rw-r--r--. 1 dirsrv dirsrv    900 Jul 15 17:00 access.20250715-165309.gz

So I would think this approach is still valid.

progier389: if it is "only" world readable, the issue is indeed less serious but IMHO it is still an issue:
The access log contains sensitive data and we should not left anybody been able to read it ...
As a general security practice, we should create the file with the lesser permissions then adjust them afterwards
 Creating file with loose permission then fixing them afterwards always generate security holes ,,,

mreynolds389: > if it is "only" world readable, the issue is indeed less serious but IMHO it is still an issue: The access log contains sensitive data and we should not left anybody been able to read it ... As a general security practice, we should create the file with the lesser permissions then adjust them afterwards Creating file with loose permission then fixing them afterwards always generate security holes ,,,

The main access log uses 600 (not world readable), only the empty gz file has 644, then we convert it to 600, and then we reopen it and start writing to it.  Sorry I still don't see a risk here.

mreynolds389: @progier389 - ok changes made please review...

progier389 (approved):
LGTM

PR #6879, opened 2025-07-16
Issue 6878 - Prevent repeated disconnect logs during shutdown

### LINKED ISSUE #6878: Shutdown with CONN error log level enabled generates thousands of repeated log messages
**Issue Description**
During server shutdown, the connection table iteration function `connection_table_disconnect_all` process all pre-allocated slots, including initialized ones with invalid sockets (sd=0) and connid=0. So during shutdown, this triggers repeated calls to disconnect_server_nomutex_ext's CONN level debug messages on these non-active connections, resulting in excessive debug logs stating "Not setting conn 0 to be disconnected: socket is invalid". Although harmless, this floods the logs and significantly impacts shutdown time.

droideck:
Description: Avoid logging non-active initialized connections via CONN in disconnect_server_nomutex_ext by adding a check to skip invalid conn=0 with invalid sockets, preventing excessive repeated messages.

Update ds_logs_test.py by adding test_no_repeated_disconnect_messages to verify the fix.

Fixes: https://github.com/389ds/389-ds-base/issues/6878

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
+++ b/dirsrvtests/tests/suites/ds_logs/ds_logs_test.py
@@ -24,7 +24,7 @@ from lib389.plugins import AutoMembershipPlugin, ReferentialIntegrityPlugin, Aut
 from lib389.idm.user import UserAccounts, UserAccount
 from lib389.idm.group import Groups
 from lib389.idm.organizationalunit import OrganizationalUnits
-from lib389._constants import DEFAULT_SUFFIX, LOG_ACCESS_LEVEL, PASSWORD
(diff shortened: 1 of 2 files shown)


droideck:
Before making the PR, I've tried other approaches like filtering `connection_table_disconnect_all` calls or rewriting the loop (by using `connection_table_get_first/next_active_connection` linked list traversal).
I've decided otherwise, as it might not be so safe.

IIUC, we might be left with some "hanging" connections if we skip Connection_Table items based on specific criteria.
So, a complete `disconnect_all` might be a bit excessive, but still a safer choice, and it's totally okay for shutdown.

mreynolds389 (approved):
LGTM

Issue #6663, opened 2025-03-11
RFE - Add option to write error log in JSON


mreynolds389:
**Issue Description**

Just like our other logs we should add an option to write error log in JSON

    {
        local_time: "",
        severity: "",
        subsystem: "",
        msg: "",
    }

mreynolds389:
3e7710114..3f92fb79d  389-ds-base-3.0 -> 389-ds-base-3.0

mreynolds389:
CLI

3f92fb79d..6c2cef0d4  389-ds-base-3.0 -> 389-ds-base-3.0

PR #6883, opened 2025-07-17
Issue 6663 - Fix NULL subsystem crash in JSON error logging

aadhikar:
Relates: https://github.com/389ds/389-ds-base/issues/6663

Reviewed by: ???


mreynolds389:
Where is the logging call being made where subsystem is not set?

aadhikar:
> Where is the logging call being made where subsystem is not set?

@mreynolds389 The crash happens when the test sets error logging to JSON format using `dsconf logging error set log-format json`.After that, any server operation that logs an error without specifying a subsystem (which is normal for internal/system errors) will crash because the JSON code tries to use NULL as a string.

The fix just checks if subsystem is NULL and uses an empty string instead, so the JSON logging doesn't crash.

aadhikar:
Adding the failing test: 

`================================================= test session starts =================================================
platform linux -- Python 3.13.5, pytest-8.4.1, pluggy-1.6.0 -- /usr/sbin/python3
cachedir: .pytest_cache
metadata: {'Python': '3.13.5', 'Platform': 'Linux-6.12.7-200.fc41.aarch64-x86_64-with-glibc2.41', 'Packages': {'pytest': '8.4.1', 'pluggy': '1.6.0'}, 'Plugins': {'flaky': '3.8.1', 'base-url': '2.1.0', 'custom-exit-code': '0.3.0', 'html': '4.1.1', 'metadata': '3.1.1', 'playwright': '0.7.0', 'anyio': '4.8.0'}, 'Base URL': ''}
389-ds-base: 3.1.2-4.fc42
nss: 3.113.0-1.fc42
nspr: 4.36.0-10.fc42
openldap: 2.6.9-3.fc42
cyrus-sasl: 2.1.28-30.fc42
FIPS: disabled
rootdir: /workspace/dirsrvtests
configfile: pytest.ini
plugins: flaky-3.8.1, base-url-2.1.0, custom-exit-code-0.3.0, html-4.1.1, metadata-3.1.1, playwright-0.7.0, anyio-4.8.0
collected 1 item                                                                                                      

dirsrvtests/tests/suites/clu/dsconf_logging_test.py::test_log_settings[error] INFO:lib389.SetupDs:Starting installation ...
INFO:lib389.SetupDs:Validate installation settings ...
INFO:lib389.SetupDs:Create file system structures ...

ERROR:dirsrvtests.tests.suites.clu.dsconf_logging_test:Command failed: /usr/sbin/dsconf standalone1 -j -D cn=Directory Manager -w password logging error set time-format %D
ERROR:dirsrvtests.tests.suites.clu.dsconf_logging_test:Stderr: {
    "msgtype": 103,
    "msgid": 3,
    "result": 53,
    "desc": "Server is unwilling to perform",
    "ctrls": [],
    "info": "Unknown attribute nsslapd-errorlog-time-format will be ignored\n",
    "ldap_request": "modify_ext_s(('cn=config', [(2, 'nsslapd-errorlog-time-format', [b'%D'])]),{'serverctrls': None, 'clientctrls': None, 'escapehatch': 'i am sure'}) on instance standalone1"
}

FAILED`

mreynolds389:
> > Where is the logging call being made where subsystem is not set?
> 
> @mreynolds389 The crash happens when the test sets error logging to JSON format using `dsconf logging error set log-format json`.After that, any server operation that logs an error without specifying a subsystem (which is normal for internal/system errors) will crash because the JSON code tries to use NULL as a string.
> 
> The fix just checks if subsystem is NULL and uses an empty string instead, so the JSON logging doesn't crash.

I was wondering where in the code it was crashing (if you had a stack trace) :-)  Every call should be setting the subsystem, so I'm worried your fix is hiding a real bug.

aadhikar:
> I was wondering where in the code it was crashing (if you had a stack trace) :-) Every call should be setting the subsystem, so I'm worried your fix is hiding a real bug.

@mreynolds389 Here is the stack trace showing exactly where the NULL subsystem crash occurs:

`(gdb) bt full
#0  0x00007fc5c917ff5d in __strlen_avx2 () at target:/lib64/libc.so.6
#1  0x00007fc5c8d48d4d in json_object_new_string () at target:/lib64/libjson-c.so.5
#2  0x00007fc5c92f945c in vslapd_log_error () at target:/usr/lib64/dirsrv/libslapd.so.0
#3  0x00007fc5c92f99c8 in slapd_log_error_proc_internal () at target:/usr/lib64/dirsrv/libslapd.so.0
#4  0x00007fc5c92f9d4c in slapi_log_error () at target:/usr/lib64/dirsrv/libslapd.so.0
#5  0x00007fc5c92d8da9 in eq_loop_rel () at target:/usr/lib64/dirsrv/libslapd.so.0
#6  0x00007fc5c8d814d1 in _pt_root () at target:/lib64/libnspr4.so
#7  0x00007fc5c90bc11a in start_thread () at target:/lib64/libc.so.6
#8  0x00007fc5c912cc3c in clone3 () at target:/lib64/libc.so.6'

The NULL comes from `eq_loop_rel()` internal server events that don't belong to any specific subsystem. This is normal system logging, not a missing parameter. JSON logging just needs the same NULL check that regular logging already has.

mreynolds389 (approved):

PR #6885, opened 2025-07-19
Issue 6884 - Mask password hashes in audit logs

### LINKED ISSUE #6884: Password hashes exposed in audit logs for user operations
**Issue Description**
When creating or modifying users with passwords, the password hashes are being written to the audit log files. 

While these are hashed passwords rather than plaintext, this still represents a security weakness. Password hashes can potentially be cracked, although it's unlikely with a strong password scheme.

droideck:
Description: Fix the audit log functionality to mask password hash values for both userPassword and nsslapd-rootpw attributes in ADD and MODIFY operations. Update auditlog.c to detect password attributes and replace their values with asterisks (**********************) in both LDIF and JSON audit log formats. Add comprehensive test suite audit_password_masking_test.py to verify password masking works correctly across all log formats and operation types.

Fixes: https://github.com/389ds/389-ds-base/issues/6884

Reviewed by: ?

### DIFF
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -44,7 +44,10 @@ static int
 is_password_attribute(const char *attr_name)
 {
     return (strcasecmp(attr_name, SLAPI_USERPWD_ATTR) == 0 ||
-            strcasecmp(attr_name, CONFIG_ROOTPW_ATTRIBUTE) == 0);
+            strcasecmp(attr_name, CONFIG_ROOTPW_ATTRIBUTE) == 0 ||
+            strcasecmp(attr_name, SLAPI_MB_CREDENTIALS) == 0 ||
+            strcasecmp(attr_name, SLAPI_REP_CREDENTIALS) == 0 ||
+            strcasecmp(attr_name, SLAPI_REP_BOOTSTRAP_CREDENTIALS) == 0);
 }
 
 /* Helper function to create a masked string representation of an entry */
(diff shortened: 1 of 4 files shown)


droideck:
`Audit LDIF (default) + modify` tests fail very often because of this issue - https://github.com/389ds/389-ds-base/issues/6886

I think we should open a separate PR for that. I checked the logs manually, and the password hashes are correctly masked.

ldap/servers/slapd/auditlog.c
```diff
@@ -39,6 +39,78 @@ static void write_audit_file(Slapi_PBlock *pb, Slapi_Entry *entry, int logtype,
 
 static const char *modrdn_changes[4];
 
+/* Helper function to check if an attribute is a password that needs masking */
+static int
+is_password_attribute(const char *attr_name)
+{
+    return (strcasecmp(attr_name, SLAPI_USERPWD_ATTR) == 0 ||
+            strcasecmp(attr_name, CONFIG_ROOTPW_ATTRIBUTE) == 0);
```
mreynolds389: We should also check for:

- nsmultiplexorcredentials
- nsds5ReplicaCredentials
- nsds5ReplicaBootstrapCredentials

vashirov: Would it make sense to have this list configurable with some values present by default? I can imagine some users would like to extend this list, or remove some attributes from it for debug purposes.

droideck: Sure, I think it's a good idea!
I'd leave it for a separate PR, though...

This PR addresses CWE, which can be easily backported to older branches to close the weakness.
A new configurational parameter will be an RFE (even though it's small), which should be planned accordingly, as I think it might not be included in all releases.

mreynolds389 (approved):

mreynolds389 (approved):
Nice

Issue #6888, opened 2025-07-21
Missing access JSON logging for TLS/CLient auth


mreynolds389:
**Issue Description**

TLS/Client auth logging was not converted to JSON (auth.c got missed)

PR #6890, opened 2025-07-21
Issue 6888 - Missing access JSON logging for TLS/Client auth

mreynolds389:
Description:

TLS/Client auth logging was not converted to JSON (auth.c got missed)

Relates: https://github.com/389ds/389-ds-base/issues/6888

### DIFF
--- a/dirsrvtests/tests/suites/logging/access_json_logging_test.py
+++ b/dirsrvtests/tests/suites/logging/access_json_logging_test.py
@@ -19,6 +19,8 @@ from lib389.idm.user import UserAccounts
 from lib389.dirsrv_log import DirsrvAccessJSONLog
 from lib389.index import VLVSearch, VLVIndex
 from lib389.tasks import Tasks
+from lib389.config import CertmapLegacy
+from lib389.nss_ssl import NssSsl
 from ldap.controls.vlv import VLVRequestControl
 from ldap.controls.sss import SSSRequestControl
 from ldap.controls import SimplePagedResultsControl
@@ -67,11 +69,11 @@ def get_log_event(inst, op, key=None, val=None, key2=None, val2=None):
                     if val == str(event[key]).lower() and \
                        val2 == str(event[key2]).lower():
                         return event
-
-            elif key is not None and key in event:
-                val = str(val).lower()
-                if val == str(event[key]).lower():
-                    return event
+            elif key is not None:
+                if key in event:
+                    val = str(val).lower()
+                    if val == str(event[key]).lower():
+                        return event
             else:
                 return event
 
@@ -163,6 +165,7 @@ def test_access_json_format(topo_m2, setup_test):
         14. Test PAGED SEARCH is logged correctly
         15. Test PERSISTENT SEARCH is logged correctly
         16. Test EXTENDED OP
+        17. Test TLS_INFO is logged correctly
     :expectedresults:
         1. Success
         2. Success
@@ -180,6 +183,7 @@ def test_access_json_format(topo_m2, setup_test):
         14. Success
         15. Success
         16. Success
+        17. Success
     """
 
     inst = topo_m2.ms["supplier1"]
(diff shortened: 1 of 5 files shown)


ldap/servers/slapd/auth.c
```diff
@@ -505,18 +576,38 @@ handle_handshake_done(PRFileDesc *prfd, void *clientData)
...
-                         "conn=%" PRIu64 " %s failed to map client "
-                         "certificate to LDAP DN (%s)\n",
-                         conn->c_connid,
-                         sslversion, extraErrorMsg);
+        if (log_format != LOG_FORMAT_DEFAULT) {
+            slapd_log_pblock_init(&logpb, log_format, NULL);
+            logpb.conn_id = conn->c_connid;
+            logpb.msg = "failed to map client certificate to LDAP DN";
+            logpb.tls_version = sslversion;
+            logpb.client_dn = clientDN;
```
droideck: Minor nitpick, clientDN is NULL here, so it'll be more explicit to set it so directly. But I don't have a hard preference here... 
```c
            logpb.client_dn = NULL;

File: ldap/servers/slapd/auth.c [hunk 2]
```diff
@@ -434,47 +468,84 @@ handle_handshake_done(PRFileDesc *prfd, void *clientData)
...
+            logpb.conn_id = conn->c_connid;
+            logpb.tls_version = sslversion;
+            logpb.keysize = keySize;
+            logpb.cipher = cipher ? cipher : "NULL";
+            slapd_log_access_tls(&logpb);
+        } else {
+            slapi_log_access(LDAP_DEBUG_STATS, "conn=%" PRIu64 " %s %i-bit %s\n",
+                    conn->c_connid,
+                    sslversion, keySize, cipher ? cipher : "NULL");
+        }
```
droideck, line 498: Should we add something like `logpb.msg = "no client certificate provided";` when `clientCert == NULL` but `config_get_SSLclientAuth() != OFF`?

ldap/servers/slapd/auth.c
```diff
@@ -363,15 +363,28 @@ handle_bad_certificate(void *clientData, PRFileDesc *prfd)
...
+        slapd_log_access_tls_client_auth(&logpb);
+    } else {
+        slapi_log_access(LDAP_DEBUG_STATS,
+                        "conn=%" PRIu64 " " SLAPI_COMPONENT_NAME_NSPR " error %i (%s); unauthenticated client %s; issuer %s\n",
+                        conn->c_connid, errorCode, slapd_pr_strerror(errorCode),
+                        subject ? escape_string(subject, sbuf) : "NULL",
+                        issuer ? escape_string(issuer, ibuf) : "NULL");
+    }
     if (issuer)
         free(issuer);
```
droideck: As you already changed `free(issuer);` in `handle_handshake_done`, might worth changing here too:
```c
slapi_ch_free_string(&issuer);

File: dirsrvtests/tests/suites/logging/access_json_logging_test.py [hunk 4]
```diff
@@ -560,6 +564,88 @@ def test_access_json_format(topo_m2, setup_test):
...
+    assert 'tls_version' in event
+    assert event['msg'] == "client bound"
+
+    # Check for failed certmap error
+    with pytest.raises(ldap.INVALID_CREDENTIALS):
+        inst.open(saslmethod='EXTERNAL', connOnly=True, certdir=ssca_dir,
+                  userkey=tls_locs_wrong['key'],
+                  usercert=tls_locs_wrong['crt'])
+
+    event = get_log_event(inst, "TLS_CLIENT_INFO", "err", -185)
```
droideck, line 643: I think it would be better to describe the magic number `-185`, as it's unclear in the test what exactly the error is.

droideck (changes requested):
In `slapd_log_access_tls_client_auth` and `slapd_log_access_tls`, can any of the  `json_obj_add_str` values be NULL?

Because if it's NULL, it snprintf's "NULL" - but JSON null would be better (json_object_new_null()).

mreynolds389:
> In `slapd_log_access_tls_client_auth` and `slapd_log_access_tls`, can any of the `json_obj_add_str` values be NULL?
> 
> Because if it's NULL, it snprintf's "NULL" - but JSON null would be better (json_object_new_null()).

Well I don't want to print every possible parameter if it's not set.  This is the logic that was used throughout all the JSON logging.  Also some logging functions are multi-purpose - so again displaying NULL for a key, where the key might not make sense, is misleading.  And also for keeping disk space usage under control I don't want to print empty values.

dirsrvtests/tests/suites/logging/access_json_logging_test.py
mreynolds389: > I think it would be better to describe the magic number `-185`, as it's unclear in the test what exactly the error is.

I am only using it to find the log event I want to look at.  So in this case err "-185" = "Certificate couldn't be mapped to an ldap entry" which you can see in the assert below.

ldap/servers/slapd/auth.c
mreynolds389: > As you already changed `free(issuer);` in `handle_handshake_done`, might worth changing here too:
> 
> ```c
> slapi_ch_free_string(&issuer);
> ```

Yeah no problem, I'll make that change.

ldap/servers/slapd/auth.c
mreynolds389: > Should we add something like `logpb.msg = "no client certificate provided";` when `clientCert == NULL` but `config_get_SSLclientAuth() != OFF`?

Yeah we can improve the logging message.

ldap/servers/slapd/auth.c
mreynolds389: > Minor nitpick, clientDN is NULL here, so it'll be more explicit to set it so directly. But I don't have a hard preference here...
> 
> ```c
>             logpb.client_dn = NULL;
> ```

Yeah nice catch, the code still would have handled it correctly, but I'll clean this up.  FYI slapd_log_pblock_init() resets the logpb, so no need to explicitly set it to NULL.

dirsrvtests/tests/suites/logging/access_json_logging_test.py
droideck: Right... Yeah, I agree the assertion is descriptive enough!

droideck:
> Well I don't want to print every possible parameter if it's not set. This is the logic that was used throughout all the JSON > logging. Also some logging functions are multi-purpose - so again displaying NULL for a key, where the key might not make sense, is misleading. And also for keeping disk space usage under control I don't want to print empty values.

That sounds good!
I was actually looking at other functions and overlooked the `if`s in the functions I mentioned before... Sorry about that. My mistake!

And regarding other places where `json_obj_add_str` is used? Like `slapd_log_access_bind`, `slapd_log_access_result`, etc...
I wonder if it's possible that any of the `json_obj_add_str` arguments can be NULL there.

For that case, I was wondering if we should modify `json_obj_add_str()` so it correctly processes NULL (by using `json_object_new_null()`)

mreynolds389:
> And regarding other places where `json_obj_add_str` is used? Like `slapd_log_access_bind`, `slapd_log_access_result`, etc... I wonder if it's possible that any of the `json_obj_add_str` arguments can be NULL there.

I was very careful around these functions but I guess someone in the future could mess it up.

> 
> For that case, I was wondering if we should modify `json_obj_add_str()` so it correctly processes NULL (by using `json_object_new_null()`)

I do not want to write NULL though.  If it's not being used then it shouldn't be logged.

I think in a new PR we should create a wrapper function that just checks for NULL and doesn't write to the object in that case.  Then we don't need all those "if's".

ldap/servers/slapd/auth.c
mreynolds389: > > Should we add something like `logpb.msg = "no client certificate provided";` when `clientCert == NULL` but `config_get_SSLclientAuth() != OFF`?
> 
> Yeah we can improve the logging message.

Actually this is not an error, and I don't think we need to state that this is not a client authentication as it's just a regular TLS connection.

droideck (approved):
LGTM!

Issue #6893, opened 2025-07-22
Log user that is updated during password modify extended operation


mreynolds389:
**Issue Description**

When a user's password is updated via an extended operation (password modify plugin) we only log the bind DN and not what user was updated.  While "internal operation" logging will display the the user it should be logged by the default logging level

mreynolds389:
c29c382a4..1cfd2c5b5  389-ds-base-3.0 -> 389-ds-base-3.0
a5717c9a5..6029af4ad  389-ds-base-2.7 -> 389-ds-base-2.7
ddc814977..d471322e4  389-ds-base-2.6 -> 389-ds-base-2.6
c88f33de9..9088ce640  389-ds-base-2.5 -> 389-ds-base-2.5
652f129be..ee3e4305d  389-ds-base-2.4 -> 389-ds-base-2.4
298340f8f..d399c022a  389-ds-base-2.3 -> 389-ds-base-2.3
eddf0d4a2..32ba717b8  389-ds-base-2.2 -> 389-ds-base-2.2
05f0bd78d..835dff111  389-ds-base-2.1 -> 389-ds-base-2.1
f0f3e7df8..a80149655  389-ds-base-2.0 -> 389-ds-base-2.0
42be6f395..7883b5e6f  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #6894, opened 2025-07-22
Issue 6893 - Log user that is updated during password modify extended operation

mreynolds389:
Description:

When a user's password is updated via an extended operation (password modify plugin) we only log the bind DN and not what user was updated. While "internal operation" logging will display the the user it should be logged by the default logging level.

Add access logging using "EXT_INFO" for the old logging format, and "EXTENDED_OP_INFO" for json logging where we display the bind dn, target dn, and message.

Relates: https://github.com/389ds/389-ds-base/issues/6893

### DIFF
--- a/ldap/servers/slapd/passwd_extop.c
+++ b/ldap/servers/slapd/passwd_extop.c
@@ -465,12 +465,14 @@ passwd_modify_extop(Slapi_PBlock *pb)
     BerElement *response_ber = NULL;
     Slapi_Entry *targetEntry = NULL;
     Connection *conn = NULL;
+    Operation *pb_op = NULL;
     LDAPControl **req_controls = NULL;
     LDAPControl **resp_controls = NULL;
     passwdPolicy *pwpolicy = NULL;
     Slapi_DN *target_sdn = NULL;
     Slapi_Entry *referrals = NULL;
-    /* Slapi_DN sdn; */
+    Slapi_Backend *be = NULL;
+    int32_t log_format = config_get_accesslog_log_format();
 
     slapi_log_err(SLAPI_LOG_TRACE, "passwd_modify_extop", "=>\n");
 
@@ -647,7 +649,7 @@ parse_req_done:
     }
     dn = slapi_sdn_get_ndn(target_sdn);
     if (dn == NULL || *dn == '\0') {
-        /* Refuse the operation because they're bound anonymously */
+        /* Invalid DN - refuse the operation */
         errMesg = "Invalid dn.";
         rc = LDAP_INVALID_DN_SYNTAX;
         goto free_and_return;
@@ -724,14 +726,19 @@ parse_req_done:
         ber_free(response_ber, 1);
     }
 
-    slapi_pblock_set(pb, SLAPI_ORIGINAL_TARGET, (void *)dn);
+    slapi_pblock_get(pb, SLAPI_OPERATION, &pb_op);
+    if (pb_op == NULL) {
+        slapi_log_err(SLAPI_LOG_ERR, "passwd_modify_extop", "pb_op is NULL\n");
+        goto free_and_return;
+    }
 
+    slapi_pblock_set(pb, SLAPI_ORIGINAL_TARGET, (void *)dn);
     /* Now we have the DN, look for the entry */
     ret = passwd_modify_getEntry(dn, &targetEntry);
     /* If we can't find the entry, then that's an error */
     if (ret) {
         /* Couldn't find the entry, fail */
-        errMesg = "No such Entry exists.";
+        errMesg = "No such entry exists.";
         rc = LDAP_NO_SUCH_OBJECT;
         goto free_and_return;
     }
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/passwd_extop.c
```diff
@@ -848,6 +840,26 @@ passwd_modify_extop(Slapi_PBlock *pb)
...
+    if (dn) {
+        /* Log the target ndn (if we have a target ndn) */
+        if (log_format != LOG_FORMAT_DEFAULT) {
+            /* JSON logging */
+            slapd_log_pblock logpb = {0};
+            slapd_log_pblock_init(&logpb, log_format, pb);
+            logpb.name = "passwd_modify_plugin";
+            logpb.target_dn = dn;
+            logpb.bind_dn = bindDN;
+            logpb.msg = errMesg ? errMesg : "success";
```
droideck, line 852: Does it make sense to also log `rc` here, in your opinion? (along with errMesg)
Like:

```c
logpb.err = rc;
logpb.msg = errMesg ? errMesg : "success";
```

droideck (approved):
Besides the minor question, looks good!

One more thing, though, just to be on the safe side, I think we can add a test case for insufficient privileges (and possibly for an anonymous bind).

ldap/servers/slapd/passwd_extop.c
```diff
@@ -742,30 +749,18 @@ passwd_modify_extop(Slapi_PBlock *pb)
...
-            rc = LDAP_OPERATIONS_ERROR;
-            goto free_and_return;
-        }
-        slapi_pblock_set(pb, SLAPI_BACKEND, be);
+    /* In order to perform the access control check, we need to select a backend (even though
+     * we don't actually need it otherwise). */
+    be = slapi_mapping_tree_find_backend_for_sdn(slapi_entry_get_sdn(targetEntry));
+    if (NULL == be) {
+        errMesg = "Failed to find backend for target entry";
+        rc = LDAP_OPERATIONS_ERROR;
```
tbordaz: May be NOT_SUCH_OBJECT ?

ldap/servers/slapd/passwd_extop.c
```diff
@@ -848,6 +840,26 @@ passwd_modify_extop(Slapi_PBlock *pb)
...
+            logpb.name = "passwd_modify_plugin";
+            logpb.target_dn = dn;
+            logpb.bind_dn = bindDN;
+            logpb.msg = errMesg ? errMesg : "success";
+            slapd_log_access_extop_info(&logpb);
+        } else {
+            slapi_log_access(LDAP_DEBUG_STATS,
+                    "conn=%" PRIu64 " op=%d EXT_INFO name=\"passwd_modify_plugin\" bind_dn=\"%s\" target_dn=\"%s\" msg=\"%s\"\n",
+                    conn ? conn->c_connid : -1, pb_op ? pb_op->o_opid : -1,
+                    bindDN ? bindDN : "", dn,
```
tbordaz, line 859: The description said we should log the 'target dn' of the extop, but here we are logging bindDN (ie. 'cn=directory manager') rather than the 'dn'

tbordaz (commented):
minor questions

ldap/servers/slapd/passwd_extop.c
mreynolds389: I left the original code as is because I didn't want to change the behavior.  I wasn't really looking at the old code, I was just focusing on adding the informative logging

ldap/servers/slapd/passwd_extop.c
mreynolds389: It logs both the bind DN and target DN for completeness

ldap/servers/slapd/passwd_extop.c
mreynolds389: It seemed redundant to include the err code since the RESULT is logged next.  I know I log the err message though, but I didn't want to add too much redundant info.  But I don't have too strong an opinion on this one though.

mreynolds389:
> Besides the minor question, looks good!
> 
> One more thing, though, just to be on the safe side, I think we can add a test case for insufficient privileges (and possibly for an anonymous bind).

Sure I can do that

ldap/servers/slapd/passwd_extop.c
mreynolds389: Also, "dn" is the normalized dn of the target_dn - I think that's where the confusion is :-)

ldap/servers/slapd/passwd_extop.c
mreynolds389: But yeah this makes sense so I changed the error code

ldap/servers/slapd/passwd_extop.c
mreynolds389: I added the err code :-)

droideck (approved):

ldap/servers/slapd/passwd_extop.c
tbordaz: Sorry the confusion was on my side, I erroneously thought the patch logged BindDN in place of 'dn'. Sorry for the noise

tbordaz (approved):
LGTM as well

Issue #6895, opened 2025-07-23
Crash if repl keep alive entry can not be created


mreynolds389:
**Issue Description**

Heap use after free when logging that the replicaton keep-alive entry can not be created:

```
==1894806==ERROR: AddressSanitizer: heap-use-after-free on address 0x510000509e50 at pc 0x7fb066e4a806 bp 0x7fab5aef2340 sp 0x7fab5aef2338
READ of size 8 at 0x510000509e50 thread T6
    #0 0x7fb066e4a805 in slapi_sdn_get_dn (/usr/lib64/dirsrv/libslapd.so.0+0x4a805) (BuildId: f099de2a8aa9e8e503e13eff09dae4deafcc4ad6)
    #1 0x7fb05ff0d0dc in replica_subentry_create ldap/servers/plugins/replication/repl5_replica.c:468
    #2 0x7fb05ff0d0dc in replica_subentry_check ldap/servers/plugins/replication/repl5_replica.c:505
    #3 0x7fb05ff0d420 in replica_subentry_update ldap/servers/plugins/replication/repl5_replica.c:543
    #4 0x7fb066e73b1f in eq_call_all_rel ldap/servers/slapd/eventq.c:278
    #5 0x7fb066e73b1f in eq_loop_rel ldap/servers/slapd/eventq.c:324
    #6 0x7fb067b71376 in _pt_root (/lib64/libnspr4.so+0x23376) (BuildId: fa7dc258da2c207de5d157c34e1023780f382d96)
    #7 0x7fb067428d45 in asan_thread_start(void*) (/lib64/libasan.so.8+0x28d45) (BuildId: 89c230c891879ee538159d2e56f56784c84db409)
    #8 0x7fb066c7d723 in start_thread (/lib64/libc.so.6+0x71723) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)
    #9 0x7fb066d0180b in __clone3 (/lib64/libc.so.6+0xf580b) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)

0x510000509e50 is located 16 bytes inside of 184-byte region [0x510000509e40,0x510000509ef8)
freed by thread T6 here:
    #0 0x7fb0674c1518 in free.part.0 (/lib64/libasan.so.8+0xc1518) (BuildId: 89c230c891879ee538159d2e56f56784c84db409)
[...]

PR #6896, opened 2025-07-23
Issue 6895 - Crash if repl keep alive entry can not be created

mreynolds389:
Description:

Heap use after free when logging that the replicaton keep-alive entry can not be created. slapi_add_internal_pb() frees the slapi entry on error, then we try and get the dn from the entry and we get a use-after-free crash.

Relates: https://github.com/389ds/389-ds-base/issues/6895


mreynolds389:
Still need to write a CI test (it's a complicated set up to trigger the crash)

droideck (approved):
The fix looks good to me!

PR #6903, opened 2025-07-25
Issue 6800 - Rerun the check in verbose mode on failure

vashirov:
Bug Description:
By default vermin doesn't report line numbers where the check failed, it's available in verbose mode. But verbose mode is too noisy.

Fix Description:
Rerun vermin again in the verbose mode, only when the first run exited with error.

Relates: https://github.com/389ds/389-ds-base/issues/6800


progier389 (approved):
LGTM

Issue #6906, opened 2025-07-28
Regression: dirsrv cannot be restarted when nsslapd-accesslog-mode is set


kippels-hhu:
**Issue Description**
I am currently testing migrating from 2.4.5 to 3.1.2 and I have noticed that it is no longer possible to restart the instance when nsslapd-accesslog-mode is set. Restarting the service with the option set produces  the follogwing output:
```
[28/Jul/2025:11:59:20.569827713 +0200] - ERR - dse_read_one_file - The entry cn=config in file /etc/dirsrv/instance-name/dse.ldif (lineno: 10) is invalid, error code 53 (Server is unwilling to perform) - nsslapd-accesslog-mode: Failed to chmod access log file to 644: errno 1 (Operation not permitted)
[28/Jul/2025:11:59:20.584514911 +0200] - ERR - init_dse_file - Could not load config file [dse.ldif]
[28/Jul/2025:11:59:20.585625253 +0200] - ERR - setup_internal_backends - Please edit the file to correct the reported problems and then restart the server.
```

**Package Version and Platform:**
 - Platform: Debian Trixie
 - Package and version: 389-ds-base (3.1.2+dfsg1-1)

**Steps to Reproduce**
1. `dsconf -D "cn=Directory Manager" -w supersecretpassword ldap://localhost config replace nsslapd-accesslog-mode=644`
2. Check access rights in /var/log/dirsrv/slapd-instance-name/
3. `dsctl instance-name restart`

**Expected results**
Server restarts without producing error messages.

**Additional context**
I have used this setting successfully on 2.3.1 on Debian Bookworm and on 2.4.5 on Ubuntu Noble before. It only chashes now on 3.1.2.

vashirov:
Hi @kippels-hhu,
could you please edit `dirsrv@.service` and add `CAP_FOWNER` to `CapabilityBoundingSet`?
`CapabilityBoundingSet` was introduced in d7b56a1ec60dac5da4959cb11b9d9977f40e431a to harden systemd unit, some capabilities were added back in 188874460af83f962f729e50a63750c64b6e4d8f, but not `CAP_FOWNER`.

kippels-hhu:
Just as a sanity check… is there anything I have to add when I install a new certificate authority? Because now that this works I see that replication is also broken now with an error message about the issuer certificate.

```
[...]

PR #6908, opened 2025-07-28

vashirov:
Description:
Expand CapabilityBoundingSet to include CAP_FOWNER

Relates: https://github.com/389ds/389-ds-base/issues/6181
Relates: https://github.com/389ds/389-ds-base/issues/6906


progier389 (approved):
LGYM

Issue #6910, opened 2025-07-28
Coverity issues


mreynolds389:
**Issue Description**

Latest coverity warnings:

```
_____________________________________________________________________________________________
*** CID 1612606:           (RESOURCE_LEAK)
/ldap/servers/slapd/log.c: 205             in compress_log_file()
199             return -1;
200         }
201     
202         if ((source = fopen(log_name, "r")) == NULL) {
203             /* Failed to open log file */
204             gzclose(outfile);
>>>     CID 1612606:           (RESOURCE_LEAK)
>>>     Handle variable "fd" going out of scope leaks the handle.
205             return -1;
206         }
207     
208         bytes_read = fread(buf, 1, LOG_CHUNK, source);
209         while (bytes_read > 0) {
210             int bytes_written = gzwrite(outfile, buf, bytes_read);
/ldap/servers/slapd/log.c: 215             in compress_log_file()
209         while (bytes_read > 0) {
210             int bytes_written = gzwrite(outfile, buf, bytes_read);
211             if (bytes_written == 0)
212             {
213                 fclose(source);
214                 gzclose(outfile);
>>>     CID 1612606:           (RESOURCE_LEAK)
>>>     Handle variable "fd" going out of scope leaks the handle.
215                 return -1;
216             }
217             bytes_read = fread(buf, 1, LOG_CHUNK, source);
218         }
219         gzclose(outfile);
220         fclose(source);
/ldap/servers/slapd/log.c: 223             in compress_log_file()
217             bytes_read = fread(buf, 1, LOG_CHUNK, source);
218         }
219         gzclose(outfile);
220         fclose(source);
221         PR_Delete(log_name); /* remove the old uncompressed log */
222     
>>>     CID 1612606:           (RESOURCE_LEAK)
>>>     Handle variable "fd" going out of scope leaks the handle.
223         return 0;
224     }
225     
226     static int
227     slapd_log_error_proc_internal(
228         int loglevel,


_____________________________________________________________________________________________
*** CID 1618837:           (OVERRUN)
/ldap/servers/slapd/back-ldbm/db-bdb/bdb_bdbreader_glue.c: 438             in dbc_get()
432                 rc = store_val(rc, dbc, key, data);
433                 break;
434             case DB_SET_RECNO:
435                 break;
436         }
437     
>>>     CID 1618837:           (OVERRUN)
[...]

PR #6912, opened 2025-07-29
Issue 6910 - Fix latest coverity issues

mreynolds389:
Description:

Fix various coverity/ASAN warnings

Relates: http://github.com/389ds/389-ds-base/issues/6910

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_bdbreader_glue.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_bdbreader_glue.c
@@ -1,5 +1,5 @@
 /** BEGIN COPYRIGHT BLOCK
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2025 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -19,7 +19,7 @@ int db_close(DB *, u_int32_t);
 #undef slapi_log_err
 
 #define LOGK(k)    (((k)&&(k)->data && (k)->size) ? (k)->data : "<NULL>")
-#define LOGN(v, names)    (((v)> sizeof names/sizeof names[0]) ? "Unexpected value" : names[v])
+#define LOGN(v, names)    (((v)>= sizeof names/sizeof names[0]) ? "Unexpected value" : names[v])
 
 int db_cursor(DB *db, DB_TXN *txnid, DBC **cursorp, u_int32_t flags);
 int dbc_close(DBC *dbc);
@@ -92,7 +92,7 @@ char *db_version(int *major, int *minor, int *patch)
     return version;
 }
 
-int nothing()
+int nothing(void)
 {
     return DB_SUCCESS;
 }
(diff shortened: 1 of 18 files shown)


ldap/servers/slapd/back-ldbm/db-bdb/bdb_bdbreader_glue.c
```diff
@@ -471,7 +471,7 @@ int db_cursor(DB *db, DB_TXN *txnid, DBC **cursorp, u_int32_t flags)
     dbc->impl = bdbreader_cur_open(db->impl);
 
     slapi_log_err(SLAPI_LOG_INFO, "bdb_ro", "%s: dbc=%p dbc->impl=%p\n", __FUNCTION__, dbc, dbc->impl);
-    return (db->impl == NULL) ? DB_OSERROR : DB_SUCCESS;
+    return (dbc->impl == NULL) ? DB_OSERROR : DB_SUCCESS;
```
mreynolds389: I think this is correct @progier389   You said about moving the NULL check up, but I think it supposed to be "dbc" not  "db"

droideck (approved):
LGTM!

A minor comment - only three files bumped to 2025; others still show older years.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_bdbreader_glue.c
mreynolds389: > I think this is correct @progier389 You said about moving the NULL check up, but I think it supposed to be "dbc" not "db"

@progier389 can you double check this?

progier389: In practice bdbreader_cur_open  cannot return NULL  because it use slapi_ch_malloc
but if db->impl is null, a null database pointer is set in the cursor leading to a crash 
Note: We can  check dbc instead of db as you propose (logically it makes sense)
But we must test if db->impl is NULL before calling  bdbreader_cur_open
and also we need to free dbc and set *cursorp to NULL in error cases (for both tests) to avoid leaks

mreynolds389: > In practice bdbreader_cur_open cannot return NULL because it use slapi_ch_malloc but if db->impl is null, a null database pointer is set in the cursor leading to a crash Note: We can check dbc instead of db as you propose (logically it makes sense) But we must test if db->impl is NULL before calling bdbreader_cur_open and also we need to free dbc and set *cursorp to NULL in error cases (for both tests) to avoid leaks

I see, ok changes made.  Please review...

progier389 (approved):
LGTM

ldap/servers/slapd/log.c
```diff
@@ -202,6 +202,7 @@ compress_log_file(char *log_name, int32_t mode)
     if ((source = fopen(log_name, "r")) == NULL) {
         /* Failed to open log file */
         gzclose(outfile);
+        close(fd);
```
progier389: fd is closed by gzclose so you should  not close(fd) despite covscan warning ...

progier389: Better mark it as a known false/positive error ...

ldap/servers/slapd/log.c
```diff
@@ -212,12 +213,14 @@ compress_log_file(char *log_name, int32_t mode)
...
             fclose(source);
             gzclose(outfile);
+            close(fd);
             return -1;
         }
         bytes_read = fread(buf, 1, LOG_CHUNK, source);
     }
     gzclose(outfile);
     fclose(source);
+    close(fd);
```
progier389: same here

Issue #6913, opened 2025-07-30
ACI Parser Accepts Invalid Syntax


droideck:
**Issue Description**
The ACI parser incorrectly accepts these malformed ACIs that should be rejected:

1. **Extremely long userdn**: `userdn="ldap:///123123123..."` (900+ characters)

2. **Invalid keyword `userdns`**: `deny(write)userdns="ldap:///anyone"`

3. **Malformed LDAP URL**: `userdn="ldap:////////anyone"`

These should all raise `INVALID_SYNTAX` errors but are currently being accepted by the parser.

The tests are available at https://github.com/389ds/389-ds-base/blob/main/dirsrvtests/tests/suites/acl/syntax_test.py#L172-L211

Issue #6886, opened 2025-07-19
Audit log parsing fails when displaying binary attributes like nsState


droideck:
**Issue Description**
When enabling full audit logging of attributes via nsslapd-auditlog-display-attrs: *, the audit log includes binary data from attributes like nsState in cn=uniqueid generator,cn=config. This attribute stores the state of the unique ID generator (including potential continuation bytes or non-UTF-8 encodable content, e.g., byte 0x80), which causes decoding failures when processing the log.

**Package Version and Platform:**
Current main branch.

**Steps to Reproduce**
1. Configure audit logging to display all attributes: inst.config.replace('nsslapd-auditlog-display-attrs', '*')
2. Attempt to parse the audit log (like with `open('audit', 'rb')` or `open('audit', 'r')`)

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 214: invalid start byte
```

**Expected results**
Audit logs should handle binary attributes safely, perhaps by base64-encoding them in the display section or skipping/escaping non-text data to prevent decoding errors.

**Additional context**
This PR's tests can reproduce the issue https://github.com/389ds/389-ds-base/pull/6885

PR #6914, opened 2025-07-30
Issues 6913, 6886, 6250 - Adjust xfail marks

droideck:
Description: Some of the ACI invalid syntax issues were fixed, so we need to remove xfail marks.
Disk space issue should have a 'skipif' mark.
Display all attrs (nsslapd-auditlog-display-attrs: *) fails because of a bug. EntryUSN inconsistency and overflow bugs were exposed with the tests.

Related: https://github.com/389ds/389-ds-base/issues/6913
Related: https://github.com/389ds/389-ds-base/issues/6886
Related: https://github.com/389ds/389-ds-base/issues/6250

Reviewed by: ?


vashirov (approved):
LGTM!

PR #6923, opened 2025-08-04
Issue 6922 - AddressSanitizer: leaks found by acl test suite

### LINKED ISSUE #6922: AddressSanitizer: leaks found by `acl` test suite
**Issue Description**
There are multiple leaks reported when running `acl` test suite with ASAN build.

vashirov:
Description:
Fix multiple memory leaks in ACL plugin:

* aclgroup.c: Fix memory leak in `aclgroup_free()` by properly cleaning up all user groups before destroying the rwlock and freeing the main structure
* aclscan.l & acltext.y: use `PERM_FREE` macro instead of `free()`
* Add cleanups in aclparse.c, aclinit.c, acllist.c, acllas.c

Fixes: https://github.com/389ds/389-ds-base/issues/6922

### DIFF
--- a/lib/libaccess/aclscan.l
+++ b/lib/libaccess/aclscan.l
@@ -64,11 +64,18 @@ variable	[\*a-zA-Z0-9\.\-\_][\*a-zA-Z0-9\.\-\_]*
 
 <<EOF>>		{	
 			yylval.string = NULL;
-			last_string = yylval.string;
+			if (last_string) {
+				PERM_FREE(last_string);
+				last_string = NULL;
+			}
 			return(0);
 		}
 
 {qstring}	{ 
+			if (last_string) {
+				PERM_FREE(last_string);
+				last_string = NULL;
+			}
 			yylval.string = PERM_STRDUP( yytext+1 );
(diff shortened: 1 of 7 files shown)


lib/libaccess/aclscan.l
```diff
@@ -64,13 +64,19 @@ variable	[\*a-zA-Z0-9\.\-\_][\*a-zA-Z0-9\.\-\_]*
...
 			return(0);
 		}
 
 {qstring}	{ 
+			if (last_string) {
+				PERM_FREE(last_string);
+				last_string = NULL;
+			}
 			yylval.string = PERM_STRDUP( yytext+1 );
-			last_string = yylval.string;
```
progier389, line 73: Should we really remove that line ?
Is it normal that laststring is now NULL instead of PERM_STRDUP( yytext+1 ); in the old code ?

vashirov: We shouldn't, I was chasing another leak and removed it here and in `{variable}`.

progier389 (approved):
LGTM

PR #6924, opened 2025-08-04
Migrate from PR_Poll to epoll and timerfd.

llile:
Now that the multi-listener, turbo and threading changes have been merged, I wanted to brush off the epoll patch again.  The prior patch had unexplained performance issues and since 389-ds-base:main has changed so much, I started over from scratch.

This version of the epoll change is laced directly into the existing code.  This will allow any further updates to the multi-listener changes to be applied directly to both the epoll and non-epoll versions for comparison testing.  Idle connection checks have been reimplemented using timerfd using epoll as well, eliminating the need for the check_idletimeout function.  I have seen no indication of mysterious performance issues with this patch.

I don’t believe that epoll will provide any significant performance benefit to single connection or even small numbers of connections.  I anticipate that the benefits will start to show with thousands of long-lived connections, like you would see when using 389ds for core name services in a Unix environment, and hopefully at c10k levels.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -396,7 +396,7 @@ fi
 AM_CONDITIONAL(enable_posix_winsync,test "$enable_posix_winsync" = "yes")
 
 if test -z "$enable_epoll" ; then
-  enable_epoll=yes
+  enable_epoll=no
 fi
 
 AC_MSG_CHECKING(for --enable-epoll)
--- a/configure.ac
+++ b/configure.ac
@@ -33,7 +33,7 @@ LT_INIT
 # Checks for header files.
 AC_HEADER_DIRENT
 AC_HEADER_SYS_WAIT
-AC_CHECK_HEADERS([arpa/inet.h errno.h fcntl.h malloc.h netdb.h netinet/in.h stdlib.h string.h strings.h sys/file.h sys/socket.h sys/time.h syslog.h unistd.h mntent.h sys/sysinfo.h sys/endian.h endian.h])
+AC_CHECK_HEADERS([arpa/inet.h errno.h fcntl.h malloc.h netdb.h netinet/in.h stdlib.h string.h strings.h sys/file.h sys/socket.h sys/time.h syslog.h unistd.h mntent.h sys/sysinfo.h sys/endian.h endian.h sys/epoll.h sys/timerfd.h])
 # These are *required* headers without option.
 AC_CHECK_HEADERS([inttypes.h], [], AC_MSG_ERROR([unable to locate required header inttypes.h]))
 AC_CHECK_HEADERS([crack.h], [], AC_MSG_ERROR([unable to locate required header crack.h]))
@@ -394,6 +394,23 @@ else
   AC_MSG_RESULT(no)
 fi
 AM_CONDITIONAL(enable_posix_winsync,test "$enable_posix_winsync" = "yes")
+
+if test -z "$enable_epoll" ; then
+  enable_epoll=yes
+fi
+
+AC_MSG_CHECKING(for --enable-epoll)
+AC_ARG_ENABLE(epoll,
+  AS_HELP_STRING([--enable-epoll],
+                 [enable support for epoll (default: yes)]))
+if test "$enable_epoll" = yes ; then
+  AC_MSG_RESULT(yes)
+  AC_DEFINE([ENABLE_EPOLL], [1], [enable support for epoll])
+else
+  AC_MSG_RESULT(no)
+fi
+
+AM_CONDITIONAL(enable_epoll, test "$enable_epoll" = "yes")
 
 # the default prefix - override with --prefix or --with-fhs
 AC_PREFIX_DEFAULT([/opt/$PACKAGE_NAME])
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -190,7 +190,7 @@ connection_cleanup(Connection *conn)
         PR_Close(conn->c_prfd);
     }
 #ifdef ENABLE_EPOLL
-    if (conn->c_idle_tfd >= 0) {
+    if (conn->c_idle_tfd != -1) {
         /* Close the idle timer */
         timerfd_settime(conn->c_idle_tfd, 0, NULL, NULL);
         close(conn->c_idle_tfd);
@@ -1914,7 +1914,7 @@ connection_threadmain(void *arg)
                not idle */
             conn->c_idlesince = curtime;
 #ifdef ENABLE_EPOLL
-            if (conn->c_idle_tfd >= 0) {
+            if (conn->c_idle_tfd != -1) {
                 /* Reset the idle timer */
                 slapi_log_err(SLAPI_LOG_DEBUG,
                               "handle_pr_read_ready", "resetting idle timer for connection %d to %d\n", conn->c_ci, conn->c_idletimeout);
@@ -1959,7 +1959,7 @@ connection_threadmain(void *arg)
                      */
                     conn->c_idlesince = curtime;
 #ifdef ENABLE_EPOLL
-                    if (conn->c_idle_tfd >= 0) {
+                    if (conn->c_idle_tfd != -1) {
                         /* Reset the idle timer */
                         slapi_log_err(SLAPI_LOG_DEBUG,
                                       "handle_pr_read_ready", "resetting idle timer for connection %d to %d\n", conn->c_ci, conn->c_idletimeout);
@@ -2536,16 +2536,16 @@ disconnect_server_nomutex_ext(Connection *conn, PRUint64 opconnid, int opid, PRE
         slapi_log_err(SLAPI_LOG_DEBUG, "disconnect_server_nomutex_ext", "Removing idle timer fd %d for connection %p (descriptor %d, table %d, conn %d)\n",
                   conn->c_idle_tfd, conn, conn->c_sd, conn->c_ct_list, conn->c_ci);
         /* Remove the idle timer if it exists */
-        if (conn->c_idle_event) {
+        if (conn->c_idle_tfd != -1) {
             if (epoll_ctl(conn->c_ct->epoll_fd[conn->c_ct_list], EPOLL_CTL_DEL, conn->c_idle_tfd, NULL) == -1) {
                 slapi_log_err(SLAPI_LOG_ERR, "disconnect_server_nomutex_ext", "Failed to remove idle timer fd %d for connection %d - %s\n",
                               conn->c_idle_tfd, conn->c_sd, strerror(errno));
             }
-        }
-        if (conn->c_idle_tfd >= 0) {
             /* Close the idle timer */
             close(conn->c_idle_tfd);
             conn->c_idle_tfd = -1;
+
+            conn->c_idle_event = NULL;
         }
 #endif /* ENABLE_EPOLL */
         g_decrement_current_conn_count();
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -192,6 +192,7 @@ connection_cleanup(Connection *conn)
 #ifdef ENABLE_EPOLL
     if (conn->c_idle_tfd != -1) {
         /* Close the idle timer */
+        epoll_ctl(conn->c_ct->epoll_fd[conn->c_ct_list], EPOLL_CTL_DEL, conn->c_idle_tfd, NULL);
         timerfd_settime(conn->c_idle_tfd, 0, NULL, NULL);
         close(conn->c_idle_tfd);
         conn->c_idle_tfd = -1;
@@ -2544,8 +2545,6 @@ disconnect_server_nomutex_ext(Connection *conn, PRUint64 opconnid, int opid, PRE
             /* Close the idle timer */
             close(conn->c_idle_tfd);
             conn->c_idle_tfd = -1;
-
-            conn->c_idle_event = NULL;
         }
 #endif /* ENABLE_EPOLL */
         g_decrement_current_conn_count();
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -189,7 +189,7 @@ connection_cleanup(Connection *conn)
     if (conn->c_prfd) {
         PR_Close(conn->c_prfd);
     }
-#ifdef ENABLE_EPOLL_NOTYET
+#ifdef ENABLE_EPOLL
     if (conn->c_idle_tfd >= 0) {
         /* Close the idle timer */
         timerfd_settime(conn->c_idle_tfd, 0, NULL, NULL);
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -189,6 +189,14 @@ connection_cleanup(Connection *conn)
     if (conn->c_prfd) {
         PR_Close(conn->c_prfd);
     }
+#ifdef ENABLE_EPOLL_NOTYET
+    if (conn->c_idle_tfd >= 0) {
+        /* Close the idle timer */
+        timerfd_settime(conn->c_idle_tfd, 0, NULL, NULL);
+        close(conn->c_idle_tfd);
+        conn->c_idle_tfd = -1;
+    }
+#endif /* ENABLE_EPOLL */
 
     conn->c_sd = SLAPD_INVALID_SOCKET;
     conn->c_ldapversion = 0;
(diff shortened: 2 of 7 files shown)


ldap/servers/slapd/daemon.c
```diff
@@ -775,22 +794,88 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
...
+    /* If no flags were set, return a specific message */
+    if (len == 0 || (len == 1 && buf[0] == '\0')) {
+        snprintf(buf, sizeof(buf), "No flags");
+    }
+    return buf;
+}
+
 static void
+#ifdef ENABLE_EPOLL
+handle_listeners(struct epoll_event *events, int event_count)
```
Firstyear, line 845: This function is small enough that you may as well just duplicate it here so that there is a epoll vs old versions of the function. It'll make it a bit clearer to follow.

ldap/servers/slapd/daemon.c
```diff
@@ -775,22 +794,88 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
...
         if (listenfd) {
+#ifdef ENABLE_EPOLL
+            if (events[idx].events & EPOLLIN) {
+#else /* !ENABLE_EPOLL */
             PR_ASSERT(fds != NULL);
             PR_ASSERT(listenfd == fds[fdidx].fd);
             if (SLAPD_POLL_LISTEN_READY(fds[fdidx].out_flags)) {
-                /* accept() the new connection, put it on the active list for handle_pr_read_ready */
+#endif /* ENABLE_EPOLL */
+                /* Question: If the listenfd is non-blocking, shouldn't we loop here until
```
Firstyear: the return here can also be -1 if PR_Accept fails, so we don't want to loop here if that already happened.

Generally if we get to this situation we really are at an edge case where we have filled the connection table, so erroring faster is better than looping, because if we loop then we also cause other connections to be delayed in finding out they can't be accepted.

ldap/servers/slapd/daemon.c
```diff
@@ -826,7 +916,19 @@ accept_thread(void *vports)
     i_unix = ports->i_socket;
 #endif /* ENABLE_LDAPI */
 
+#ifdef ENABLE_EPOLL
```
Firstyear: Similar to the above, I think shortterm duplicating the accept_threads() fn here is better than trying to ifdef so much which makes it harder to reader.

ldap/servers/slapd/daemon.c
```diff
@@ -2030,6 +2262,48 @@ handle_new_connection(Connection_Table *ct, int tcps, PRFileDesc *listenfd, int
...
+    if (epoll_ctl(the_connection_table->epoll_fd[conn->c_ct_list], EPOLL_CTL_ADD, PR_FileDesc2NativeHandle(conn->c_prfd), conn->c_event) == -1) {
+        slapi_log_err(SLAPI_LOG_ERR, "handle_new_connection", "epoll_ctl() failed: %s\n",
+                      strerror(errno));
+        /* XXX cleanup how? */
+    }
+
+    if ((conn->c_idle_tfd = timerfd_create(CLOCK_MONOTONIC, TFD_NONBLOCK)) < 0) {
+        slapi_log_err(SLAPI_LOG_ERR, "handle_new_connection", "timerfd_create() failed: %s\n",
+                      strerror(errno));
+        /* XXX cleanup how? */
```
Firstyear: If we can't create this, or add to epoll, I think we go straight into disconnect_server_nomutex() so that we can disconnect the connection and log an appropriate error. Generally both of these will error if we are out of fd's or some other resource limit so I think disconnecting and making noise is correct here.

ldap/servers/slapd/daemon.c
```diff
@@ -2620,6 +2894,13 @@ get_connection_table_size(void)
         size = (FDS_PROCESS_MAX - resrvdesc);
     }
 
+#ifdef ENABLE_EPOLL
+    /* epoll and timerfd use 2 descriptors per connection
```
Firstyear: I think the way this is written is a bit unclear, but also we actually need more file descriptors than this. 

We have an epoll fds in the accept threads, and we also have epoll fds in all the handlers too. So I actually think this shourld be:

```
size = (size / 2) - num_listeners - 3 ;
size = size - ((size / 2) % 2);
```

I think that covers it?

Firstyear (commented):
This looks really good so far! I look forward to giving it a test run :)

ldap/servers/slapd/daemon.c
llile: After reading your suggestion, does this comment make clearer the calculation?  I have also updated the calculation to hopefully be more accurate, I was originally trying to modulus to the multiple of the number of connection table lists and failed to use the actual count.

    /* We need to account for the additional timer file descriptor per connection,
     * which is used for the idle timeout, the listener file descriptors, the epoll
     * file descriptors in the accept threads and connection table.
     * 
     * Then round the size down to the nearest multiple of the number of connection
     * table lists, so that we can use the same number of connections per list.
     */
    size = size - (config_get_num_listeners() * 2); /* One listener per table, one epoll fd per table */
    size = (size / 2) - ((size / 2) % config_get_num_listeners()); /* One fd for the connection, one for the timerfd */

ldap/servers/slapd/daemon.c
llile: No problem, it's a fine line between staying tightly coupled to the original code and being readable.

ldap/servers/slapd/daemon.c
llile: That makes sense, thanks for the clarification.

ldap/servers/slapd/daemon.c
llile: Accepted as above.

ldap/servers/slapd/daemon.c
llile: Excellent, thanks for pointing me to disconnect_server_nomutex here.  I used SLAPD_DISCONNECT_ABORT, ECANCELED as they seemed the most appropriate to the situation.

llile: I'm not sure disconnect_server_nomutex is the right path now that I've written the code.

This will fall through to the mutex unlock (fine) but then increments the connection count and hands off the connection to the connection table.

Will this be properrly cleaned up?

ldap/servers/slapd/connection.c
```diff
@@ -189,6 +189,14 @@ connection_cleanup(Connection *conn)
     if (conn->c_prfd) {
         PR_Close(conn->c_prfd);
     }
+#ifdef ENABLE_EPOLL_NOTYET
```
Firstyear: Just curious why this is _NOTYET? What was your concern about it?

ldap/servers/slapd/connection.c
```diff
@@ -1938,6 +1958,18 @@ connection_threadmain(void *arg)
                      * are bypassing both of those, we set idlesince here
                      */
                     conn->c_idlesince = curtime;
+#ifdef ENABLE_EPOLL
+                    if (conn->c_idle_tfd >= 0) {
+                        /* Reset the idle timer */
+                        slapi_log_err(SLAPI_LOG_DEBUG,
+                                      "handle_pr_read_ready", "resetting idle timer for connection %d to %d\n", conn->c_ci, conn->c_idletimeout);
+                        timerfd_settime(conn->c_idle_tfd, 0,
```
Firstyear, line 1967: If the connection is in CONN_FLAG_CLOSING, then won't this timerfd_settime fail as the event was removed?

ldap/servers/slapd/daemon.c
```diff
@@ -775,6 +794,83 @@ disk_monitoring_thread(void *nothing __attribute__((unused)))
...
+
+#ifdef ENABLE_EPOLL
+static void
+handle_listeners(struct epoll_event *events, int event_count)
+{
+    Connection_Table *ct = the_connection_table;
+    size_t idx;
+    int ctlist = 0;
+    struct listener_info *listener = NULL;
+    for (idx = 0; idx < (size_t)event_count; idx++) {
```
Firstyear: minor, `for (size_t idx = 0; ... )` There are some compiler optimisations that only work if you put it into the for loop like this, and it's a nice defence against using the idx outside the loop generally.

ldap/servers/slapd/daemon.c
Firstyear: I think you're right. Up until the connection is added to the active list, it seems there is a bit of an odd limbo state here where we don't have a clear way to return this to the freelist on the conntable. I suspect we need to break `connection_table_move_connection_out_of_active_list`. Maybe the code after https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/conntable.c#L467 needs to become a separate function where we can return the connection from these early error cases where disconnect server mutex isn't the right answer, and before we are on the active list.

Really, the active list and the concept of it isn't needed once we have epoll, but that's a bigger cleanup and rework.

ldap/servers/slapd/connection.c
llile: No concern.  I just failed to revert that when diagnosing an issue.

ldap/servers/slapd/daemon.c
llile: Old habit from when you couldn't do that.

ldap/servers/slapd/connection.c
llile: The itimerspec is attached to the actual fd, not the epoll event struct.  It is reset here to avoid firing while closing.

ldap/servers/slapd/daemon.c
llile: I will look at this code again and see if I can move it above the pthread_mutex_lock, so that a failure can clean up immediately and return -1.

llile: I've moved that section of code above the mutex.  It will perform the needed incremental cleanup steps and return -1 if there is a failure.  It's a bit messy but as you mentioned it should be collected into a larger refactoring.

ldap/servers/slapd/connection.c
llile: I've removed the _NOTYET as this code should not have been disabled.

Firstyear (approved):
This looks great I'm really happy with this. I'll leave it for a day or so for other team members if they want to review, but I'm excited to see how this performs. Thanks so much for the work on this!

Firstyear:
@jchapma Since you looked at the connection code recently, maybe you want to look at this too.

jchapma:
Hi @llile, its great to see your epoll enhancement back to life. I am reviewing the code and running some performance tests in parallel, I will report my findings on Monday. Looks positive so far. Thank you for your work.

llile:
@jchapma thanks for running your performance tests on the epoll variant.  I'm sad to see it doesn't really offer any base level benefit, I'm hoping it will have greater effect under higher connection counts.  

A couple of quick questions: 
1. Which version of slamd are you using?  The Oracle 2.1.0 or Github version?  
2. Do you happen to have your slamd configuration documented anywhere I could access?
3. Do you think slapd_accept_wakeup_timer should be increased or decreased?

configure.ac
```diff
@@ -395,6 +395,23 @@ else
 fi
 AM_CONDITIONAL(enable_posix_winsync,test "$enable_posix_winsync" = "yes")
 
+if test -z "$enable_epoll" ; then
+  enable_epoll=yes
```
progier389: We should better have "no" as default for now

progier389 (commented):
As most of the code is within #ifdef.
I think that once you change the configure.ac to disable the feature by default
we can (and should) merge the PR (as @firstyear said, we can always work on the performance later on)
But we cannot enable it by default right now. Apparently it changes the server dynamics (and that is probably why some tests fails ) The tests spend more time than usual  and some tests fails because replication does not synchronize in time 

Many tests are still failing compared to the usual run (and I feel that the tests are much longer . 
apparently replication does not seems to synchronize soon enough - maybe a change of dynamic and a timeout issue)

configure.ac
llile: I have updated configure.ac in the patch as requested.

ldap/servers/slapd/connection.c
```diff
@@ -1938,6 +1958,18 @@ connection_threadmain(void *arg)
...
                      */
                     conn->c_idlesince = curtime;
+#ifdef ENABLE_EPOLL
+                    if (conn->c_idle_tfd != -1) {
+                        /* Reset the idle timer */
+                        slapi_log_err(SLAPI_LOG_DEBUG,
+                                      "handle_pr_read_ready", "resetting idle timer for connection %d to %d\n", conn->c_ci, conn->c_idletimeout);
+                        timerfd_settime(conn->c_idle_tfd, 0,
+                                        &(struct itimerspec) {
+                                            .it_value = { .tv_sec = conn->c_idletimeout, .tv_nsec = 0 },
```
tbordaz, line 1969: As we are going to read this connection (not polling it), shouldn't the timer value/interval set to 0 ?

ldap/servers/slapd/connection.c
```diff
@@ -189,6 +189,14 @@ connection_cleanup(Connection *conn)
     if (conn->c_prfd) {
         PR_Close(conn->c_prfd);
     }
+#ifdef ENABLE_EPOLL
+    if (conn->c_idle_tfd != -1) {
+        /* Close the idle timer */
+        timerfd_settime(conn->c_idle_tfd, 0, NULL, NULL);
```
tbordaz, line 196: does it worth to remove the c_idle_tfd from the list of fd ?

`epoll_ctl(conn->c_ct->epoll_fd[conn->c_ct_list], EPOLL_CTL_DEL, conn->c_idle_tfd, NULL) `

ldap/servers/slapd/connection.c
```diff
@@ -2493,6 +2525,29 @@ disconnect_server_nomutex_ext(Connection *conn, PRUint64 opconnid, int opid, PRE
...
+        if (conn->c_idle_tfd != -1) {
+            if (epoll_ctl(conn->c_ct->epoll_fd[conn->c_ct_list], EPOLL_CTL_DEL, conn->c_idle_tfd, NULL) == -1) {
+                slapi_log_err(SLAPI_LOG_ERR, "disconnect_server_nomutex_ext", "Failed to remove idle timer fd %d for connection %d - %s\n",
+                              conn->c_idle_tfd, conn->c_sd, strerror(errno));
+            }
+            /* Close the idle timer */
+            close(conn->c_idle_tfd);
+            conn->c_idle_tfd = -1;
+
+            conn->c_idle_event = NULL;
```
tbordaz: Are you sure ? it should be kept for the next time the connection is used.

ldap/servers/slapd/daemon.c
```diff
@@ -1689,12 +1959,25 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
...
                                               SLAPD_DISCONNECT_POLL, EPIPE);
+#endif /* ENABLE_EPOLL */
                 } else if (readready) {
                     /* read activity */
                     slapi_log_err(SLAPI_LOG_CONNS,
                                   "handle_pr_read_ready", "read activity on %d\n", c->c_ci);
                     c->c_idlesince = curtime;
-
+#ifdef ENABLE_EPOLL
+                    if (c->c_idle_tfd != -1 && c->c_idle_tfd >= 0) {
```
tbordaz: Why testing '>=0' , as long as it is not -1 it should be a valid fd.

tbordaz (approved):
Looks very nice to me as well. Few remarks. I agree it can be pushed upstream now and improved later.

ldap/servers/slapd/daemon.c
llile: I think code completion got me there, I'll fix that.

ldap/servers/slapd/connection.c
llile: The idle timer is a count down timer, when it expires it will in c_idletimeout seconds, it will send a signal which will cause the fd to become ready to read, with the number of expirations as the data.  This resets the count down to the full idle timeout value, with a single interval of the full idle timeout.  So I belive this code is correct.

ldap/servers/slapd/connection.c
llile: Yes, I think you're correct and I believe I misunderstood your comment on my first reading.

progier389 (approved):
As excepted, with the feature disabled the tests results are back to the expected ones. 
Waiting for the changes related to @tbordaz comments before merging the PR

llile:
Should I squash the commit?

jchapma (approved):

Firstyear:
With 4 acks, and consensus we want to merge this, I've gone ahead and done so. Thank you so much @llile for this, we look forward to continuing to improve this feature and it's performance!

llile:
Thanks everyone for the assistance with the patch, I hope it will be valuable going forward.

PR #6930, opened 2025-08-11
Issue 6929 - Compilation failure with rust-1.89 on Fedora ELN

### LINKED ISSUE #6929: Compilation failure with rust-1.89 on Fedora ELN
**Issue Description**
```
error: hiding a lifetime that's elided elsewhere is confusing
  --> slapi_r_plugin/src/value.rs:64:17
   |
64 |     pub fn iter(&self) -> ValueArrayRefIter {
   |                 ^^^^^     ----------------- the same lifetime is hidden here
   |                 |
   |                 the lifetime is elided here
   |
   = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
note: the lint level is defined here
  --> slapi_r_plugin/src/lib.rs:1:9
   |
1  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(mismatched_lifetime_syntaxes)]` implied by `#[deny(warnings)]`
help: use `'_` for type paths
   |
64 |     pub fn iter(&self) -> ValueArrayRefIter<'_> {
   |                                            ++++

error: could not compile `slapi_r_plugin` (lib) due to 1 previous error

```

vashirov:
Bug Description:
The `ValueArrayRefIter` struct has a lifetime parameter `'a`. But in the `iter` method  the return type doesn't specify the lifetime parameter.

Fix Description:
Make the lifetime explicit.

Fixes: https://github.com/389ds/389-ds-base/issues/6929

### DIFF
--- a/src/slapi_r_plugin/src/value.rs
+++ b/src/slapi_r_plugin/src/value.rs
@@ -61,7 +61,7 @@ impl ValueArrayRef {
         ValueArrayRef { raw_slapi_val }
     }
 
-    pub fn iter(&self) -> ValueArrayRefIter {
+    pub fn iter(&self) -> ValueArrayRefIter<'_> {
         ValueArrayRefIter {
             idx: 0,
             va_ref: &self,


droideck (approved):
LGTM

PR #6938, opened 2025-08-14
Issue 6753 - Removing ticket 47676 test and porting to DSLdapObject

aadhikar:
Description:
The old ticket47676_test.py had compatibility issues. This ports the functionality to schema_update_policy_test.py using modern DSLdapObject methods.

Relates: https://github.com/389ds/389-ds-base/issues/6753

Reviewed by: ???


vashirov (approved):
LGTM

vashirov (dismissed):
Actually, there is a test validation failure:
```
* Docstring should have at least id token(s)
```

aadhikar:
> > And there is a failure on BDB: https://github.com/389ds/389-ds-base/actions/runs/16965769739/job/48090208310?pr=6938 I couldn't reproduce it on my VM though.
> 
> @vashirov The test was failing due to a race condition where the schema replication policy wasn't fully loaded after server restart. Added proper synchronization to wait for the policy configuration to be active before proceeding with the test. Let me know what you think about it.

BTW I have tried using `waitForEntry` but this won't work because: `waitForEntry` can only wait for an attribute to exist, but cannot wait for an attribute to be absent or have a specific value.

droideck (approved):
LGTM

PR #6941, opened 2025-08-15
Issue 6940 - dsconf monitor server fails with ldapi:// due to absent server ID

### LINKED ISSUE #6940: dsconf monitor server fails with ldapi:// due to absent server ID
**Description**
The `dsconf monitor server` command fails when using `ldapi://` protocol because the server ID is not set, preventing PID retrieval from `defaults.inf`. This causes the Web console (Directory Server console) to fail displaying the "Server Version" field under Server → Server Settings → General Settings.
But it might cause more issues with the CLI and WebUI.

**Steps**:
1. Login to web console
2. Navigate to Server → Server Settings → General Settings
3. Observe the "Server Version" field

**Expected Result**: "Server Version" field displays the server version

**Actual Result**: "Server Version" field is empty/not displayed

droideck:
Description: The dsconf monitor server command fails when using ldapi://
    protocol because the server ID is not set, preventing PID retrieval from
    defaults.inf. This causes the Web console to fail displaying the "Server
    Version" field and potentially other CLI/WebUI issues.

    The fix attempts to derive the server ID from the LDAPI socket path when
    not explicitly provided. This covers the common case where the socket name
    contains the instance name (e.g., slapd-instance.socket).
    If that's not possible, it also attempts to derive the server ID from the
    nsslapd-instancedir configuration attribute. The derived server ID
    is validated against actual system instances to ensure it exists.
    Note that socket names can vary and nsslapd-instancedir can be changed.
    This is a best-effort approach for the common naming pattern.

    Also fixes the LDAPI socket path extraction which was incorrectly using
    offset 9 instead of 8 for ldapi:// URIs.

    The monitor command now handles missing PIDs gracefully, returning zero
    values for process-specific stats instead of failing completely.

Fixes: https://github.com/389ds/389-ds-base/issues/6940

Reviewed by: ?

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -17,7 +17,7 @@
 
 import sys
 import os
-from urllib.parse import urlparse
+from urllib.parse import urlparse, unquote
 import stat
 import pwd
 import grp
@@ -67,7 +67,8 @@ from lib389.utils import (
     get_default_db_lib,
     selinux_present,
     selinux_label_port,
-    get_user_is_root)
+    get_user_is_root,
+    get_instance_list)
 from lib389.paths import Paths
 from lib389.nss_ssl import NssSsl
 from lib389.tasks import BackupTask, RestoreTask, Task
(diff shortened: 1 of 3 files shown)


vashirov (changes requested):
Works for ldapi, but doesn't work for other URI schemes (ldap, ldaps):

```
# dsconf -D "cn=Directory Manager" -w password ldap://localhost monitor server
Error: process PID not found (pid=0)

dsconf -D "cn=Directory Manager" -w password ldaps://localhost:636 monitor server
Error: process PID not found (pid=0)
```
Instead of relying on socket name (which can be changed), maybe it's better extract it from `nsslapd-instancedir`?

src/lib389/lib389/__init__.py
```diff
@@ -528,6 +580,15 @@ def allocate(self, args):
             self.ldapi_autobind = args.get(SER_LDAPI_AUTOBIND, 'off')
             self.isLocal = True
             self.log.debug("Allocate %s with %s", self.__class__, self.ldapi_socket)
+        elif self.ldapuri is not None and isinstance(self.ldapuri, str) and self.ldapuri.startswith('ldapi://'):
+            # Using LDAPI via URI implies local instance. Try to learn serverid.
```
vashirov: This is not entirely true. I have 2 containers running, their sockets are symlinked on the host to `/tmp/sockets/container1.socket` and `/tmp/sockets/container2.socket`. I can talk to these instances using `dsconf` or `ldapsearch`, but they are not local in a sense that `dsconf` can discover them on the filesystem.

BTW, I found another issue, which is not related to this PR, it's happening on 3.1.3, but was working on 2.4.x:
```
# dsconf ldapi://%2Ftmp%2Fsockets%2Fcontainer1.socket monitor server
Error: Inappropriate authentication(48) - SASL EXTERNAL bind requires an SSL connection
```
The same command works under non-root user.

vashirov (approved):
Besides minor issue with the comment, LGTM!

mreynolds389 (approved):
Not sure if all of Viktor's comments have been addressed, but the rest LGTM

Issue #6947, opened 2025-08-19
Revise time skew check in healthcheck tool


mreynolds389:
**Issue Description**

The current check reports critical warning is time skew is greater than 1 day.  Even if "nsslapd-ignore-time-skew" is set to "on" in DS.

We should consider changing the severity level based on how large the skew is, and if "ignore time skew" is set.

PR #6949, opened 2025-08-20
Issue 6947 - Revise time skew check in healthcheck tool

mreynolds389:
Description:

The current check reports a critical warning if time skew is greater than 1 day - even if "nsslapd-ignore-time-skew" is set to "on". If we are ignoring time skew we should still report a warning if it's very significant like 30 days.

Relates: https://github.com/389ds/389-ds-base/issues/6947

The 30 day check is debatable.   It could be set higher like 6 months...  Thoughts?

### DIFF
--- a/src/lib389/lib389/dseldif.py
+++ b/src/lib389/lib389/dseldif.py
@@ -23,7 +23,8 @@ from lib389.lint import (
     DSPERMLE0002,
     DSSKEWLE0001,
     DSSKEWLE0002,
-    DSSKEWLE0003
+    DSSKEWLE0003,
+    DSSKEWLE0004
 )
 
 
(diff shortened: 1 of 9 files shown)


src/lib389/lib389/dseldif.py
```diff
@@ -68,26 +69,49 @@ def lint_uid(cls):
...
+        skew_low = 21600  # 6 hours
+
+        ignore_skew = self.get("cn=config", "nsslapd-ignore-time-skew")
+        if ignore_skew is not None and ignore_skew[0].lower() == "on":
+            # If we are ignoring time skew only report a warning if the skew
+            # is significant
+            ignoring_skew = True
+            skew_high = 86400 * 30  # Report a warning for skew over 30 days
+            skew_medium = 99999999999
+            skew_low = 99999999999
```
droideck, line 88: Minor nitpick, I think we can use something like `float('inf')` or `sys.maxsize` instead here. It'll be a bit more elegant:)

droideck (commented):
LGTM.

As my 2 cents, I think it can be even set to `14 days` by default for 'ignore: on'.

But also, does it make sense to calculate some kind of replication retention window?
We can base it on nsslapd-changelogmaxage or/and nsDS5ReplicaPurgeDelay.

And then the healthcheck message may refer to that as a "very significant" skew that's worthy of the warning.

src/lib389/lib389/dseldif.py
mreynolds389: > Minor nitpick, I think we can use something like `float('inf')` or `sys.maxsize` instead here. It'll be a bit more elegant:)

Yeah :-)

droideck (approved):

vashirov (approved):

vashirov:
@mreynolds389, sounds good. I've noticed `test_healthcheck_low_disk_space` failed both in BDB and LMDB, but it seems to be unrelated. I retriggered the tests.

PR #6950, opened 2025-08-22
Lint rust plugins

bakanovskii:
No features added or removed, mostly lint changes from clippy
I might add a few modules for `slapi_r_plugin` later on, so as a first step to get familiar with your workflow and contrubiton I opened this PR

I used https://www.port389.org/docs/389ds/contributing.html for reference, sorry if i did something wrong

## Summary by Sourcery

Apply lint and style improvements across Rust plugins, modernizing FFI declarations, pointer handling, and general Rust idioms while maintaining existing functionality.

Enhancements:
- Modernize FFI declarations to unsafe extern "C" and simplify pointer casts with .cast(), const fn constructors, and explicit type annotations
- Refine Rust idioms by using Self in enum matches, range syntax for validation, let Ok(...) patterns, vec![] and vec![0; len], and format string interpolation in log messages
- Improve const correctness by marking constructors and pointer-returning methods as const and deriving Eq for FilterType

Build:
- Modernize build.rs scripts with inline variable interpolation for cargo directives and file paths

Tests:
- Update PBKDF2 tests to remove Result wrappers, simplify assertions, and adjust get_pbkdf2_rounds return type

### DIFF
--- a/src/librnsslapd/build.rs
+++ b/src/librnsslapd/build.rs
@@ -10,7 +10,7 @@ fn main() {
                 .with_crate(crate_dir)
                 .generate()
                 .expect("Unable to generate bindings")
-                .write_to_file(format!("{}/rust-nsslapd-private.h", out_dir));
+                .write_to_file(format!("{out_dir}/rust-nsslapd-private.h"));
         }
     }
 }
--- a/src/librslapd/build.rs
+++ b/src/librslapd/build.rs
@@ -10,7 +10,7 @@ fn main() {
                 .with_crate(crate_dir)
                 .generate()
                 .expect("Unable to generate bindings")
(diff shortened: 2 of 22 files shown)


vashirov (changes requested):
entryuuid and pwdchan related tests are failing with this PR.

vashirov:
In this case ensure that you don't have 389-ds packages installed from the distro, and use binaries installed in the prefix.

droideck:
Thanks for the contribution and for your patience.
This PR has grown into a broad cleanup containing many unrelated style, API, and FFI changes, which makes it difficult to review and maintain.
Closing it for now.

Please feel free to reopen it with a minimal, focused set of changes that avoids unrelated refactoring. Smaller follow-up PRs for separate cleanup categories would also be easier to review. Thanks again for working on this.

Issue #6951, opened 2025-08-25
On line server certificate refresh


progier389:
** Goal
Problem Statement Support updating/renewing TLS certificate without restarting slapd
Presently we require a service restart for the update: https://docs.redhat.com/en-us/documentation/red_hat_directory_server/12/pdf/securing_red_hat_directory_server/Red_Hat_Directory_Server-12-Securing_Red_Hat_Directory_Server-en-US.pdf
Sections 1.3 and 1.4

** What would you like
I'm hoping however that it would be possible to support some method of reloading the TLS certificate/key without stopping and restarting the service. If not via something simple like SIGHUP, perhaps a task entry like reloading the schema could trigger slapd to reload the TLS key & certificate and then use it for new client connections.

(This would enable something using a ACME renewal process to call a helper script to easily update the certificate and make it active without needing to coordinate the process across the pool of load-balanced replicas and wait to drain connections from the instance.)

** Acceptance criteria

    - Renew certificate
    - Have alternatives for admins other than the currently expected, # dsctl instance_name restart

**Additional context**
See original demand in https://issues.redhat.com/browse/RHEL-86320

PR #6952, opened 2025-08-25
Issue 6951 - On line server certificate refresh

progier389:
Goal: be able to change the server certificate without restarting the server
Implementation:
Add a  new dsconf config refresh-certs subcommand  that set a new boolean attribute in cn=config entry 
(nsslapd-refresh-certificates)  then poll it until the server reset it after the certificate get refreshed.
The flag triggers a synchronisation mechanism that block the accept thread and the listener thread until all these threads are blocked then it reset the security system by clearing some SSL cache, poping SSL layer from the listening NSPR I/O stack (i.e: PR_Filedesc) and redo the security initialization from scratch.
Once everything thing is done, the  nsslapd-refresh-certificates is reset and the blocked threads get unblock
Usage: 
  Administrator may change/renew the server certificate with the NSS dtabase (cert9.db) using dsctl commands or certutil . Administrator may also change some of the security parameter (related to the certificate handling) then run the   'dsconf  instance config refresh-certs' command.
Beware that the connection using the old CA certificate will be broken  

Issue: #6951 

Reviewed by: ?

### DIFF
--- a/dirsrvtests/tests/suites/tls/ecdsa_test.py
+++ b/dirsrvtests/tests/suites/tls/ecdsa_test.py
@@ -173,7 +173,7 @@ class ECDSA_Certificate:
         self.dir = outdir
         self.prefix = prefix
         self.args = { **ECDSA_Certificate.VDEF, **kwargs }
-        if not 'organizationName' in kwargs:
+        if 'organizationName' not in kwargs:
             self.args['organizationName'] = f'{self.args["organizationName"]}-{prefix}'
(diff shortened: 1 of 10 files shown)


progier389:
testcase fails when running as root server fails to retrieve the private key
 maybe a pin password issue ...

progier389:
Closing this PR for now:
 The tests shows that there is no reliable way to reload the nss db
 We should rather process differently by managing the nss db  from within ns-slapd process.
There is a working rfe in progress for that one

Issue #6956, opened 2025-08-28
Fix various issues in UI


mreynolds389:
**Issue Description**

- Not all Radio buttons were updated for PF5 so the handler functions
need updating.
- Fix some html tags/localization issues.
- Disable Close button during instance creation
- Fix capitalization in MO plugin field description
- Add a toggle button for displaying the Last Modified Time in
  different formats

PR #6969, opened 2025-09-04
Issue 6956 - More UI fixes

mreynolds389:
Description:

-  Cannot add group as a member to a group
-  Instance Name is not validated properly
-  memberOf plugin: Auto Add OC value not shown after being set (mixed case OC)
-  Broken table layout for removing members from a group
-  No prompt in webUI to restart instance after configuring Attribute Uniqueness plugin

Relates: https://github.com/389ds/389-ds-base/issues/6956

### DIFF
--- a/src/cockpit/389-console/src/dsModals.jsx
+++ b/src/cockpit/389-console/src/dsModals.jsx
@@ -79,7 +79,7 @@ export class CreateInstanceModal extends React.Component {
     }
 
     validInstName(name) {
-        return /^[\w@_:-]*$/.test(name);
+        return /^(?!-)[\w#%@_:-]*$/.test(name);
     }
 
     validRootDN(dn) {
@@ -118,7 +118,7 @@ export class CreateInstanceModal extends React.Component {
             } else if (!this.validInstName(this.state.createServerId)) {
                 all_good = false;
                 errObj.createServerId = true;
-                createServerIdMsg = _("Instance name can only contain letters, numbers, and these 4 characters:  - @ : _");
+                createServerIdMsg = _("Instance name can only contain letters, numbers, and these characters: % # - @ : _");
             }
         }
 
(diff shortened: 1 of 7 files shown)


aadhikar (approved):
LGTM!

src/cockpit/389-console/src/lib/plugins/attributeUniqueness.jsx
```diff
@@ -611,7 +611,8 @@ class AttributeUniqueness extends React.Component {
                     console.info("attrUniqOperation", "Result", content);
                     this.props.addNotification(
                         "success",
-                        cockpit.format(_("The $0 operation was successfully done on \"$1\" entry"), action, configName)
+                        cockpit.format("The $0 operation was successfully done on \"$1\" entry, please restart the instance for these changes to take effect.",
```
aadhikar, line 614: earlier it was in `_()`: Message is hardcoded and won't be translated, not sure if it will break anything!

mreynolds389: It only translates it if there is a corresponding mapping in ja.po.  It won't match anymore, so by removing the "function" it is easier to spot as a place where we need a new language mapping.

src/cockpit/389-console/src/dsModals.jsx
```diff
@@ -79,7 +79,7 @@ export class CreateInstanceModal extends React.Component {
     }
 
     validInstName(name) {
-        return /^[\w@_:-]*$/.test(name);
+        return /^(?!-)[\w#@_:-]*$/.test(name);
```
droideck: I noticed that this regex doesn't include `%` (but in Python we validate it)

droideck (approved):
LGTM!

Issue #6954, opened 2025-08-26
Investigate error when reiniting a read only consumer


mreynolds389:
**Issue Description**

This appears to happen when you reinitialize a consumer.  This happens when trying to delete the existing referrals (not replace them - see repl_set_mtn_state_and_referrals() ):

```
[17/Jun/2025:16:57:28.550789474 +0000] - ERR - NSMMReplicationPlugin - repl_set_mtn_state_and_referrals - Could not set referrals for replica dc=example,dc=com: 53
```

But this appears to be harmless as after the reinit everything works fine, but the message is alarming and it should be adjusted or removed.

PR #6974, opened 2025-09-04
Issue 6954 - do not delete referrals on chain_on_update backend

mreynolds389:
Description:

The server does not allow deleting referrals on a backend with "referral on update" backend state.  You can replace referrals but not delete them. Trying to delete just leads to a harmless but alarming error 53 in the errors log.

This happens when doing an online init of a read-only consumer that is setup for chain-on-update. At the start of the init the referrals are successfully replaced, but after the init it tries to delete them and this needs to be skipped.

Also changed lib389 warning about consumers and changelogs because UI expects successful operations to be in JSON but the warning is in standard text. Changing the warning to be written to stderr avoids the UI crash.

Relates: https://github.com/389ds/389-ds-base/issues/6954

## Summary by Sourcery

Prevent deletion of referrals on chain-on-update backends and convert changelog warnings to error-level logs to avoid UI JSON parsing issues, with minor whitespace and formatting cleanup in the replication CLI.

Bug Fixes:
- Skip deleting referrals for backends in referral-on-update state when chain-on-update is enabled in the replication plugin
- Change changelog support warnings for consumer replicas from info to error level to prevent UI JSON parsing errors

Enhancements:
- Clean up trailing whitespace and adjust docstring formatting in the replication CLI code

### DIFF
--- a/ldap/servers/plugins/replication/replutil.c
+++ b/ldap/servers/plugins/replication/replutil.c
@@ -668,7 +668,6 @@ repl_set_mtn_state_and_referrals(
         }
         /* We should delete referral only if we want to set the
            replica database in backend state mode */
-        /* if chain on update mode, go ahead and set the referrals anyway */
         if (strcasecmp(mtn_state, STATE_BACKEND) == 0 || chain_on_update) {
             rc = slapi_mtn_set_referral(repl_root_sdn, referrals_to_set);
             if (rc == LDAP_NO_SUCH_ATTRIBUTE) {
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -1189,10 +1189,12 @@ mapping_tree_entry_modify_callback(Slapi_PBlock *pb,
                 node->mtn_referral_entry = referral2entry(referral, subtree);
             } else if (SLAPI_IS_MOD_DELETE(mods[i]->mod_op)) {
                 /* it is not OK to delete the referrals if they are still
-                 * used
-                 */
-                if ((node->mtn_state == MTN_REFERRAL) ||
-                    (node->mtn_state == MTN_REFERRAL_ON_UPDATE)) {
+                 * used */
+                Slapi_Attr *ref_attr = NULL;
+                if (((node->mtn_state == MTN_REFERRAL) ||
+                    (node->mtn_state == MTN_REFERRAL_ON_UPDATE)) &&
+                    !slapi_entry_attr_find(entryAfter, "nsslapd-referral", &ref_attr)) {
+
                     PR_snprintf(returntext, SLAPI_DSE_RETURNTEXT_SIZE,
                                 "cannot delete referrals in this state\n");
                     *returncode = LDAP_UNWILLING_TO_PERFORM;
--- a/src/lib389/lib389/cli_conf/replication.py
+++ b/src/lib389/lib389/cli_conf/replication.py
@@ -556,7 +556,7 @@ def generate_lag_report(inst, basedn, log, args):
                 formats.append('csv')
             else:
                 log.warning(f"Ignoring unknown format: {fmt}")
-    
+
     if not formats:  # Fallback to html if no valid formats specified
         formats.append('html')
 
@@ -577,7 +577,7 @@ def generate_lag_report(inst, basedn, log, args):
         # Initialize ReplicationLogAnalyzer with enhanced options
         if not json_output_only:
             log.info("Initializing replication log analysis...")
-        
+
         repl_analyzer = ReplicationLogAnalyzer(
             log_dirs=args.log_dirs,
             suffixes=args.suffixes,
@@ -594,14 +594,14 @@ def generate_lag_report(inst, basedn, log, args):
         # Parse logs
         if not json_output_only:
             log.info("Analyzing replication logs...")
-        
+
         repl_analyzer.parse_logs()
 
         # Generate reports
         if not json_output_only:
             log.info("Generating analysis reports...")
             log.info(f"Creating reports in formats: {formats}")  # Debug message
-        
+
         generated_files = repl_analyzer.generate_report(
             output_dir=args.output_dir,
             formats=formats,
@@ -687,7 +687,7 @@ def set_per_backend_cl(inst, basedn, log, args):
     did_something = False
 
     if (is_replica_role_consumer(inst, suffix)):
-        log.info("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
+        log.error("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
 
     if args.encrypt:
         cl.replace('nsslapd-encryptionalgorithm', 'AES')
@@ -720,7 +720,7 @@ def get_per_backend_cl(inst, basedn, log, args):
     suffix = args.suffix
 
     if (is_replica_role_consumer(inst, suffix)):
-        log.info("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
+        log.error("Warning: Changelogs are not supported for consumer replicas. You may run into undefined behavior.")
 
     cl = Changelog(inst, suffix)
     if args and args.json:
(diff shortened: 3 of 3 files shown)


ldap/servers/plugins/replication/replutil.c
```diff
@@ -668,8 +668,7 @@ repl_set_mtn_state_and_referrals(
         }
         /* We should delete referral only if we want to set the
            replica database in backend state mode */
-        /* if chain on update mode, go ahead and set the referrals anyway */
-        if (strcasecmp(mtn_state, STATE_BACKEND) == 0 || chain_on_update) {
+        if (strcasecmp(mtn_state, STATE_BACKEND) == 0) {
```
tbordaz: I wonder if the problem could be that the second call to repl_set_mtn_state_and_referrals (after total update) should not be with 'referrals' (provided by the total init req) instead of 'mtnreferral' (value of mapping tree referral before the total init).
In short, in multisupplier_extop_StartNSDS50ReplicationRequest, calling 
`repl_set_mtn_state_and_referrals(repl_root_sdn, mtnstate,
                                                 NULL, NULL, referrals);`

mreynolds389: @tbordaz this is the stack where the referrals are NULL.  When referrals are NULL we try to delete the existing referrals which is explicitly denied by the mapping tree code:

```
#0  repl_set_mtn_state_and_referrals (repl_root_sdn=0x7fc2c78006c0, mtn_state=0x7fc808dc4e37 "referral on update", ruv=0x0, 
    ruv_referrals=0x0, other_referrals=0x0) at ../389-ds-base/ldap/servers/plugins/replication/replutil.c:608
#1  0x00007fc808d51465 in consumer5_set_mapping_tree_state_for_replica (r=0x7fc2ce05f340, supplierRuv=0x0)
    at ../389-ds-base/ldap/servers/plugins/replication/repl5_replica.c:1543
#2  0x00007fc808d51965 in replica_reload_ruv (r=0x7fc2ce05f340) at ../389-ds-base/ldap/servers/plugins/replication/repl5_replica.c:1713
#3  0x00007fc808d5708c in replica_enable_replication (r=0x7fc2ce05f340)
    at ../389-ds-base/ldap/servers/plugins/replication/repl5_replica.c:3946
#4  0x00007fc808d4bf3c in multisupplier_be_state_change (handle=0x7fc808d4bebc <multisupplier_be_state_change>, 
    be_name=0x7fc2c6604a00 "userroot", old_be_state=2, new_be_state=1)
    at ../389-ds-base/ldap/servers/plugins/replication/repl5_plugins.c:1451
#5  0x00007fc80ce747b3 in mtn_be_state_change (be_name=0x7fc2c6604a00 "userroot", old_state=2, new_state=1)
    at ../389-ds-base/ldap/servers/slapd/mapping_tree.c:218
#6  0x00007fc80ce7b54d in mtn_internal_be_set_state (be=0x7fc808867ac0, state=1) at ../389-ds-base/ldap/servers/slapd/mapping_tree.c:3610
#7  0x00007fc80ce7b5d1 in slapi_mtn_be_enable (be=0x7fc808867ac0) at ../389-ds-base/ldap/servers/slapd/mapping_tree.c:3660
#8  0x00007fc808465a28 in dbmdb_import_all_done (job=0x7fc2ca443000, ret=0)
    at ../389-ds-base/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c:804
#9  0x00007fc8084666be in dbmdb_public_dbmdb_import_main (arg=0x7fc2ca443000)
    at ../389-ds-base/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c:1133
#10 0x00007fc808466766 in dbmdb_import_main (arg=0x7fc2ca443000) at ../389-ds-base/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c:1169
```

Note the mapping tree code rejects the delete if the backend state  is "referral" or "referral on update"


```
mapping_tree_entry_modify_callback()

1109:
            } else if (SLAPI_IS_MOD_DELETE(mods[i]->mod_op)) {
                /* it is not OK to delete the referrals if they are still
                 * used
                 */
                if ((node->mtn_state == MTN_REFERRAL) ||
                    (node->mtn_state == MTN_REFERRAL_ON_UPDATE)) {
                    PR_snprintf(returntext, SLAPI_DSE_RETURNTEXT_SIZE,
                                "cannot delete referrals in this state\n");
                    *returncode = LDAP_UNWILLING_TO_PERFORM;
                    mtn_unlock();
                    slapi_sdn_free(&subtree);
                    slapi_ch_free_string(&plugin_fct);
                    slapi_ch_free_string(&plugin_lib);
                    return SLAPI_DSE_CALLBACK_ERROR;
                }

                mtn_free_referral_in_node(node);
```

mreynolds389: Also worth noting.  We call repl_set_mtn_state_and_referrals() 3 times when doing an init:

[1]  Start repl session (replace referrals)
[2]  End repl session (replace referrals)
[3]  Enabling replication after import (this is where RUV referrals are null and we try to delete referrals)

tbordaz: Okay, I thought it was called twice and my suggestion is about that [line](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/plugins/replication/repl_extop.c#L999) asking if changing  'mtnreferral' into referrals' could fix the issue.

progier389: slapi_mtn_set_referral is only called from repl_set_mtn_state_and_referrals
 and imho the modify should not fail in referral on update mode (we should be able to update the RUV entry)
  I think that operation flag should not be 0 but something else like:
    OP_FLAG_REPLICATED | OP_FLAG_REPL_FIXUP | OP_FLAG_TOMBSTONE_ENTRY | OP_FLAG_REPL_RUV as in replica_write_ruv
  
   what puzzles me is that slapi_mtn_set_referral seems to directly write the db ruv entry rather than the in memory ruv then calling replica_write_ruv
    but that may be normal (the in memory ruv may get updated from the db when enabling the backend)

mreynolds389: > I think that operation flag should not be 0 but something else like:
OP_FLAG_REPLICATED | OP_FLAG_REPL_FIXUP | OP_FLAG_TOMBSTONE_ENTRY | OP_FLAG_REPL_RUV as in replica_write_ruv

This is not the issue.  We check the backend state and reject it before trying the modify operation in mapping_tree_entry_modify_callback()

progier389: Ok, it makes more sense. so the problem is behind the test: https://github.com/389ds/389-ds-base/blob/bf48b85bad5d82c40f1170cae1dc4a38276d161d/ldap/servers/slapd/mapping_tree.c#L1194
IMHO the idea behind test is to prevent setting REFERRAL/REFERRAL ON UPDATE mode without any referrals
but the way the test is written is too strict and reject any value deletion.
We should better accept such modifier if  slapi_entry_attr_find(entryAfter, "nsslapd-referral", &state_attr))
  exists and avoid rejecting the modify operation

progier389 (changes requested):
mapping_tree_entry_modify_callback test
    /* it is not OK to delete the referrals if they are still
                 * used
                 */
 should be modified to reject the operation only if 
   !slapi_entry_attr_find(entryAfter, "nsslapd-referral", &state_attr))
(i.e: there are no referrals)

ldap/servers/plugins/replication/replutil.c
mreynolds389: Trying this, and returning the "|| chain_on_update" in replutil.c, causes the error to move to the supplier instead of the consumer.  So I just pushed another change and it's based solely on the comments in the code.  So we reject the delete if referrals ARE present and the backend state is a referral type.  That's what it says it's trying to do, and I have not been able to reproduce the error in any of the CI tests and my personal chain_on_update setup.

Alternatively we could also keep the code the same and just ignore error 53 in replutil to avoid the error being logged.

progier389 (approved):
LGTM

PR #6978, opened 2025-09-08
Issue 6977 - UI - Show error message when trying to use unavailable ports

### LINKED ISSUE #6977: UI - Show error message when trying to use unavailable ports
**Issue Description**
When trying to change the ports an instance is using to ports, that are already used, a red exclamation mark symbol is shown, but no message. A clear message (e.g. in a tooltip) that the ports are already used would improve user experience.

 - Browser Chrome

**Steps to Reproduce**
1. Create an instance
2. Go to Server -> Server settings
3. Try to change the ports to others, that are unavailable (e.g. used by another instance)
4. Red icon with exclamation mark is displayed with no explanation

**Expected results**
An explanation of the issue.

jchapma:
Bug description:
When trying to change the ports an instance is using to ports, that are already used, a red exclamation mark symbol is shown, but no message. A clear message (e.g. in a tooltip) that the ports are already used would improve user experience.

Fix description:
Added a Tooltip to inform the user that the selected port is in use, please choose another.

Fixes: https://github.com/389ds/389-ds-base/issues/6977

Reviewed by:

## Summary by Sourcery

Bug Fixes:
- Show an error tooltip when users select an already in-use port on LDAP and LDAPS port fields

### DIFF
--- a/src/cockpit/389-console/src/lib/server/settings.jsx
+++ b/src/cockpit/389-console/src/lib/server/settings.jsx
@@ -21,7 +21,6 @@ import {
 	Text,
 	TextContent,
 	TextVariants,
-    Tooltip,
 	ValidatedOptions
 } from '@patternfly/react-core';
 import {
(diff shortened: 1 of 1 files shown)


mreynolds389:
I think it might be be better to use the helper-text component with variant="error" in this case as tip tool is not meant for displaying errors:

https://v5-archive.patternfly.org/components/helper-text

This is the technique used in other places in our UI.

mreynolds389:
> I think it might be be better to use the helper-text component with variant="error" in this case as tip tool is not meant for displaying errors:
> 
> https://v5-archive.patternfly.org/components/helper-text
> 
> This is the technique used in other places in our UI.

Also ToolTip is displaying an empty box thing when there is no error so I really don't think it should be used in this case

Screenshot From 2025-09-08 10-27-17

mreynolds389 (approved):
Thanks Jamie!  Ack

PR #6980, opened 2025-09-08
Issue 6979 - Improve the way to detect asynchronous operations in the…

### LINKED ISSUE #6979: Improve the way to detect asynchronous operations in the access logs
**Issue Description**
Some applications using asynchronous operations can consume multiple workers. If the load remains under maxthreadperconn it can not be detected. Even if it hit that limit, it is not easy to identify which client (connection) is responsible of such requests. 
Monitoring the server is the easiest way to detect that behavior but it is impacting (slowdown the connections) and sometime impossible if the server is unresponsive.
As a result, if a server becomes unresponsive because of too many asynchronous requests it is difficult to diagnose.

The suggested fix is to log a new 'notes' character (e.g. 'a') in the access log to identify asynchronous requests

 - all versions

**Steps to Reproduce**
1. runs asynchronous requests under maxthreadperconn, the server becomes unresponsive

**Expected results**
asynchronous requests should be tagged in access logs so that a simple grep will find them

tbordaz:
… access logs

Bug description:
	Asynch operations are prone to make the server unresponsive.
	The detection of those operations is not easy.
	Access logs should contain a way to retrieve easilly the
	operations (an the connections) with async searches

Fix description:
	When dispatching a new operation, if the count of
	uncompleted operations on the connection overpass
	a threshold (2) then add a note to the 'notes='
	in the access log
	between the

fixes: #6979

Reviewed by:

## Summary by Sourcery

Introduce asynchronous operation detection and logging for operations dispatched on busy connections.

New Features:
- Detect asynchronous operations by marking operations when more than two uncompleted operations exist on a connection.

Enhancements:
- Add SLAPI_OP_NOTE_ASYNCH_OP to the operation note enum and include its mapping in the note map for access logs.

### DIFF
--- a/dirsrvtests/tests/suites/basic/basic_test.py
+++ b/dirsrvtests/tests/suites/basic/basic_test.py
@@ -493,6 +493,72 @@ def test_basic_ops(topology_st, import_example_ldif):
     check_db_sanity(topology_st)
     log.info('test_basic_ops: PASSED')
 
+def test_basic_search_asynch(topology_st, request):
+    """
+    Tests asynchronous searches generate string 'notes=B'
+    and 'notes=N' in access logs
+
+    :id: 1b761421-d2bb-487b-813e-2278123fd13c
+    :parametrized: no
+    :setup: Standalone instance, create test user to search with filter (uid=*).
+
+    :steps:
+        1. Create a test user
+        2. trigger async searches
+        3. Verify access logs contains 'notes=B' up to 10 attempts
+        4. Verify access logs contains 'notes=N' up to 10 attempts
+
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+
+    """
+
+    log.info('Running test_basic_search_asynch...')
+
+    search_filter = "(uid=*)"
+    topology_st.standalone.restart()
+    topology_st.standalone.config.set("nsslapd-accesslog-logbuffering", "off")
+    topology_st.standalone.config.set("nsslapd-maxthreadsperconn", "3")
+
+    try:
+        users = UserAccounts(topology_st.standalone, DEFAULT_SUFFIX, rdn=None)
+        user = users.create_test_user()
+    except ldap.LDAPError as e:
+        log.fatal('Failed to create test user: error ' + e.args[0]['desc'])
+        assert False
+
+    for attempt in range(10):
+        msgids = []
+        for i in range(5):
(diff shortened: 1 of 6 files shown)


progier389:
Quid about non search async operation. (no Notes for theses one ...)

progier389:
> Do you mean flagging (notes=B) the operation when c->c_threadnumber==c->c_max_thread_per_conn ?
Exactly
> This condition will prevent the connection to be polled but it is done on different threads (polling thread vs workers), depending on dispatching we may flag an operation although others workers already returned.
Not sure what is your concern aboput different thread:
   Nothing prevent us to check that before just returning the result.
      If we hit the conditiom that means that we are not polling any more the connection  and it is an interesting point
        now I wonder if it should be a Note (limited to search operation) or a new line like:
     conn=xxx info="not listened because there are too many asynchronous operations" 
        to get warned whether the async operation are search or not

dirsrvtests/tests/suites/basic/basic_test.py
```diff
@@ -493,6 +493,62 @@ def test_basic_ops(topology_st, import_example_ldif):
...
+    for attempt in range(10):
+        msgids = []
+        for i in range(5):
+            searchid = topology_st.standalone.search(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE, search_filter)
+            msgids.append(searchid)
+
+        for msgid in msgids:
+            rtype, rdata = topology_st.standalone.result(msgid)
+
+        error_lines = topology_st.standalone.ds_error_log.match('.*notes=a.*')
```
droideck: Probably, that should be:

```python
access_lines = topology_st.standalone.ds_access_log.match('.*notes=a.*')

File: ldap/servers/slapd/slap.h [hunk 2]
```diff
@@ -1722,6 +1722,7 @@ typedef struct conn
     uint64_t c_connid;               /* id of this connection for stats*/
     PRUint64 c_maxthreadscount;      /* # of times a conn hit max threads */
     PRUint64 c_maxthreadsblocked;    /* # of operations blocked by maxthreads */
+    PRBool c_flagblocked;            /* Flag the next read operation as blocked */
```
progier389: I would rather add the new flag around the end of the struct (less risk to disrupt things like performance impact because alignment changed))
Another point: these days we tend to use plain bool (and true/false) rather than PRBool

progier389 (approved):
LGTM

Issue #6891, opened 2025-07-22
JSON logging - add wrapper function that checks for NULL


mreynolds389:
**Issue Description**

The jsonc library will crash if you call json_obj_add_str() with a nULL vlaue.  So there are a lot of NULL checks in accesslog.c.  It would be easier to just create a wrapper that will check for NULL and skip adding the string to the json object.  Then we can remove all the NULL checks.

PR #6981, opened 2025-09-08
Issue 6891 - JSON logging - add wrapper function that checks for NULL

mreynolds389:
Description:

Created a wrapper function to check for NULL before eventually calling json_object_new_string(). This also makes the code much more readable

Relates: https://github.com/389ds/389-ds-base/issues/6891

## Summary by Sourcery

Introduce a null-checked wrapper for JSON string insertion and refactor the access log code to use it for safer and cleaner JSON logging.

Enhancements:
- Add json_add_str wrapper to skip NULL values when adding JSON strings and prevent crashes
- Refactor accesslog.c to replace direct json_object_new_string calls with the new wrapper for consistency and readability

### DIFF
--- a/ldap/servers/slapd/accesslog.c
+++ b/ldap/servers/slapd/accesslog.c
@@ -17,6 +17,9 @@
 #define JBUFSIZE 75
 #define KEY_SIZE 50
 
+/*
+ * Helper function to handle long lines that need truncating
+ */
 struct json_object*
 json_obj_add_str(const char *value)
 {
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/accesslog.c
```diff
@@ -34,6 +37,22 @@ json_obj_add_str(const char *value)
...
+json_add_str(json_object *obj, char *key, const char *value)
+{
+    if (value) {
+        json_object_object_add(obj, key, json_obj_add_str(value));
+    }
+}
+
+/*
+ * Build the core JSON logging object that we can later build upon
+ */
```
droideck: As part of futureproofing, I think we could also add `return 0;  // success` and `return 1;  // skipped due to NULL`.

ldap/servers/slapd/accesslog.c
```diff
@@ -34,6 +37,22 @@ json_obj_add_str(const char *value)
 
 }
 
+/*
+ * Wrapper function to check for NULL as NULL would cause
+ * json_object_new_string() to crash. If value is NULL we
+ * do not add anything to the JSON object.
+ */
+static void
+json_add_str(json_object *obj, char *key, const char *value)
```
droideck: I think `key` should be `const char *key` as per `json_object_object_add`'s signature.

droideck (approved):
LGTM except fro a couple of minor issue. Ack!

ldap/servers/slapd/accesslog.c
mreynolds389: > I think `key` should be `const char *key` as per `json_object_object_add`'s signature.

Sounds good, done!

ldap/servers/slapd/accesslog.c
mreynolds389: > As part of futureproofing, I think we could also add `return 0; // success` and `return 1; // skipped due to NULL`.

Done!

PR #6889, opened 2025-07-21
Issue 6887 - Enhance logconv.py to add support for JSON access logs

jchapma:
Description: Update logconv.py to support parsing and processing of JSON formatted access logs in addition to existing formats.

Fixes: https://github.com/389ds/389-ds-base/issues/6887

Reviewed by:

PR #6986, opened 2025-09-09
Issue 6985 - Some logconv CI tests fail with BDB

### LINKED ISSUE #6985: Some logconv CI tests fail with BDB
**Issue Description**
When BDB is the configured backend, a rootdse search is performed after a bind. These "extra" operations mean the stats reported by logconv don't match the expected values so the tests fail.

**Steps to Reproduce**
1. See logconv CI tests failing in CI framework

**Expected results**
All tests should pass

jchapma:
Description: When BDB is the configured backend, a rootdse search is performed after a restart. These "extra" operations mean the stats reported by logconv don't match the expected values so the tests fail.

Fix: Add a loose validation for BDB tests that require a restart.

Fixes: https://github.com/389ds/389-ds-base/issues/6985

Reviewed by

## Summary by Sourcery

Introduce loose validation for DN database (BDB) in logconv CI tests to accommodate extra rootDSE searches and prevent spurious failures

Bug Fixes:
- Relax logconv stats comparison to prevent CI test failures when using the BDB backend

Enhancements:
- Add LOOSE_VALIDATION_TESTS set and conditional logic to use a less strict comparison for specified tests under BDB

Tests:
- Update validate_logconv_stats to select strict or loose validation based on backend and test name

### DIFF
--- a/dirsrvtests/tests/suites/logging/logconv_test.py
+++ b/dirsrvtests/tests/suites/logging/logconv_test.py
@@ -47,6 +47,27 @@ def topology_st(request):
 
     return topology
 
+def adjust_expected_for_backend(self, expected: dict):
+    """
+    Adjust expected logconv stats based on backend specific behavior.
+
+    A BDB implementation triggers additional operations during internal
+    actions, like restarts. This helper updates the expected stats to
+    account for these backend specific extra operations.
+
+    Args:
+        expected (dict): Dictionary of expected logconv stats.
+
+    Returns:
+        dict: Updated expected stats reflecting backend behavior.
+    """
+    if self.inst.dblib == "bdb":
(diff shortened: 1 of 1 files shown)


droideck (changes requested):
Looking closely at the failures and logs, I agree with Pierre's comment here:
https://github.com/389ds/389-ds-base/pull/6889#issuecomment-3265534840

I think we can do a wrapper around the restart call so each restart accounts one additional value for BDB case checks.

jchapma:
The restart wrapper idea worked until I noticed in the ldaps test case that enable_tls() does an implicit instance restart. I thought it was cleaner to add an "adjust expected" function to call after an operation that triggers an instance restart, enable_tls() and inst.restart()

droideck (approved):
LGTM!

PR #6998, opened 2025-09-15
Issue 6997 - Logic error in get_bdb_impl_status prevents bdb2mdb exec…

### LINKED ISSUE #6997: Logic error in get_bdb_impl_status prevents bdb2mdb execution
**Issue Description**
My 389-ds backend is BDB, I would like to migrate to MDB. This is not possible.

 - Platform: Fedora 41
 - Package and version: 389-ds-base-3.1.3-2.fc41.src.rpm

**Steps to Reproduce**
1. run `dsctl my-instance dblib bdb2mdb`

**Expected results**
Database migrated from bdb to mdb

**Actual results**
```
Berkeley Database library is not available. Maybe 389-ds-base-bdb rpm should be installed.
Error: Berkeley Database library is not available
```

**Additional context**
`def get_bdb_impl_status()` has an implementation error

In my Fedora 41 install I have a bundled `libback-lmdb.so`, but no bundled `libback-bdb.so`. As a result, the implementation function returns `NONE` instead of `STANDARD` as it should.

`plgstrs == {'libback-bdb': True, 'libdb-': True}`

vashirov:
…ution

Bug Description:
On F41 and F42 migration to MDB fails:
```
dsctl localhost dblib bdb2mdb
Berkeley Database library is not available. Maybe 389-ds-base-bdb rpm should be installed.
Error: Berkeley Database library is not available
```

This happens because `BDB_IMPL_STATUS.NONE` gets returned earlier than we check for presence of the standard libdb provided by the system.

Fix Description:
Rewrite the logic to return `BDB_IMPL_STATUS.NONE` only when no working BDB implementation is found.

Fixes: https://github.com/389ds/389-ds-base/issues/6997

## Summary by Sourcery

Fix get_bdb_impl_status logic to correctly detect the system-provided BDB library and only return NONE when a bundled indicator exists without a valid plugin path

Bug Fixes:
- Prevent premature NONE status return in get_bdb_impl_status so that the standard libdb is properly detected and bdb2mdb can execute
- Restructure the implementation status checks in order: read-only, bundled, standard, none (for found strings without plugin), and unknown

### DIFF
--- a/dirsrvtests/tests/suites/clu/dsctl_dblib_test.py
+++ b/dirsrvtests/tests/suites/clu/dsctl_dblib_test.py
@@ -16,6 +16,7 @@ from lib389.backend import DatabaseConfig
 from lib389.cli_ctl.dblib import (
     FakeArgs,
     dblib_bdb2mdb,
+    dblib_mdb2bdb,
     dblib_cleanup,
     is_bdb_supported)
 from lib389.idm.user import UserAccounts
--- a/src/lib389/lib389/cli_ctl/dblib.py
+++ b/src/lib389/lib389/cli_ctl/dblib.py
@@ -146,20 +146,27 @@ def get_bdb_impl_status():
     bundledbdb_plugin = 'libback-bdb'
     libdb = 'libdb-'
     plgstrs = check_plugin_strings(backldbm, [bundledbdb_plugin, libdb])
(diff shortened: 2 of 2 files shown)


progier389 (approved):
LGTM

Issue #7008, opened 2025-09-19
Potential NULL dereference of `valid_slots` during rollback in `dbmdb_open_all_files()`


zeff-ir:
Static analysis points to a possible NULL dereference in the `error:` path of `dbmdb_open_all_files()` (mdb_instance.c): on early exit before allocating `valid_slots`, the code evaluates `if (valid_slots[i])`. Proposed minimal fix: change it to `if (!valid_slots || valid_slots[i])`, or alternatively allocate `valid_slots` before the first possible jump to `error:`.

zeff-ir:
False Positive

PR #7009, opened 2025-09-19
Issue 7008 - Prevent NULL deref of in dbmdb_open_all_files()

zeff-ir:
In `dbmdb_open_all_files()`, an early jump to the `error:` label can occur before the `valid_slots` array is allocated and initialized, which led to indexing `valid_slots[i]` and a NULL dereference in the rollback block. An early `error:` jump is possible if `START_TXN()` (or any subsequent call wrapped in `TST(...)`) fails before `valid_slots` is set up. The fix replaces `if (valid_slots[i])` with the safer `if (!valid_slots || valid_slots[i])`.

Issue: #7008 

## Summary by Sourcery

Bug Fixes:
- Guard against valid_slots being NULL when iterating in dbmdb_open_all_files to prevent invalid memory access

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
@@ -515,7 +515,7 @@ error:
         ctx->dbis_treeroot = NULL;
         for (i=0; i < ctx->startcfg.max_dbs; i++) {
             if (ctx->dbi_slots[i].dbname) {
-                if (valid_slots[i]) {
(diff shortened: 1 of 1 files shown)


zeff-ir:
Issue: #7008

progier389:
The issue description looks wrong:
  There is no goto error; nor use of the TST macro before valid_slots get initialized by  slapi_ch_calloc (that cannot return NULL)
  So although the proposed change is harmless it is also needless.

zeff-ir:
> slapi_ch_calloc (that cannot return NULL)

Oh, sorry. I didn't take that into account.

Issue #7014, opened 2025-09-19
memberOf plugin - do not use deferred_update with LMDB


mreynolds389:
**Issue Description**

With LMDB "deferred update" is not needed as it is much more efficient than BDB.  When using LMDB just ignore the setting/feature

mreynolds389:
9696f5069..96e3f3890  389-ds-base-3.0 -> 389-ds-base-3.0
0b9027ccd..4efa29561  389-ds-base-2.7 -> 389-ds-base-2.7
025fe76dc..03059e76e  389-ds-base-2.6 -> 389-ds-base-2.6
a37d7db3c..4f5230d36  389-ds-base-2.5 -> 389-ds-base-2.5

PR #7015, opened 2025-09-19
Issue 7014 - memberOf - ignored deferred updates with LMDB

mreynolds389:
Description:

When processing the memberOf plugin conifguration simply ignore the deferred update settings if LMDB is in use.  Log a message in the error log but do not reject the update because rejecting the update causes the server to not start up.

Relates: https://github.com/389ds/389-ds-base/issues/7014

## Summary by Sourcery

Ignore the memberOf plugin’s deferred_update setting when using an LMDB backend to avoid startup failures, while preserving the setting for other backends and logging a notice. Add a detection helper and update build configuration, and include tests covering both LMDB and non‐LMDB scenarios.

Bug Fixes:
- Prevent server startup failures by ignoring memberOf deferred_update on LMDB instead of rejecting the update

Enhancements:
- Add slapi_db_is_lmdb() helper to detect LMDB backend
- Link memberOf plugin against libback-ldbm in the build configuration

Tests:
- Add pytest tests verifying deferred_update is ignored with LMDB and honored with other backends

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -1184,7 +1184,7 @@ memberof_postop_start(Slapi_PBlock *pb)
     memberof_rlock_config();
     mainConfig = memberof_get_config();
     /* if the update of the members is deferred then allocate mutex/cv */
-    if (mainConfig->deferred_update) {
+    if (mainConfig->deferred_update && !mainConfig->is_lmdb) {
         MemberofDeferredList *deferred_list;
         pthread_condattr_t condAttr;
 
@@ -1367,7 +1367,7 @@ memberof_postop_del(Slapi_PBlock *pb)
         /* retrieve deferred update params that are valid until shutdown */
         memberof_rlock_config();
         mainConfig = memberof_get_config();
-        deferred_update = mainConfig->deferred_update;
+        deferred_update = mainConfig->is_lmdb ? false : mainConfig->deferred_update;
         memberof_unlock_config();
 
         if (deferred_update) {
@@ -1743,7 +1743,7 @@ memberof_postop_modrdn(Slapi_PBlock *pb)
         /* retrieve deferred update params that are valid until shutdown */
         memberof_rlock_config();
         mainConfig = memberof_get_config();
-        deferred_update = mainConfig->deferred_update;
+        deferred_update = mainConfig->is_lmdb ? false : mainConfig->deferred_update;
         memberof_unlock_config();
 
         if (deferred_update) {
@@ -2062,7 +2062,8 @@ memberof_postop_modify(Slapi_PBlock *pb)
         /* retrieve deferred update params that are valid until shutdown */
         memberof_rlock_config();
         mainConfig = memberof_get_config();
-        deferred_update = mainConfig->deferred_update;
+
+        deferred_update = mainConfig->is_lmdb ? false : mainConfig->deferred_update;
         memberof_unlock_config();
 
         if (deferred_update) {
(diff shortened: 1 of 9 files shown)


tbordaz (commented):
The change looks good to prevent setting deferred memberof (in case of LMDB) when the server is online. In case dse.ldif was updated directly I think we should also test LMDB each time we call 'mainConfig->deferred_update'.

ldap/servers/slapd/back-ldbm/dblayer.c
```diff
@@ -1487,6 +1487,27 @@ dblayer_is_lmdb(Slapi_Backend *be)
...
 }
 
+/* Return true if lmdb is the backend db library */
+bool
+slapi_db_is_lmdb(void) {
+    Slapi_Backend *be = NULL;
+    char *cookie = NULL;
+    int32_t use_lmdb = 0;
+
+    be = slapi_get_first_backend(&cookie);
```
progier389: Could return wrong value if  server is started without any db backends.
I wonder if we should not rather directly look up at the configuration value 
(i.e using internal search to check for nsslapd-backend-implement=mdb in cn=config,cn=ldbm database,cn=plugins,cn=config)

mreynolds389: > Could return wrong value if server is started without any db backends. I wonder if we should not rather directly look up at the configuration value (i.e using internal search to check for nsslapd-backend-implement=mdb in cn=config,cn=ldbm database,cn=plugins,cn=config)

Yeah I was thinking about "an internal search" over the weekend verses linking libback-ldbm into the plugin, and trying to add it to libslapd doesn't work because of a circular import issue.  So the internal search might be the best approach.

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -1367,7 +1367,7 @@ memberof_postop_del(Slapi_PBlock *pb)
         /* retrieve deferred update params that are valid until shutdown */
         memberof_rlock_config();
         mainConfig = memberof_get_config();
-        deferred_update = mainConfig->deferred_update;
+        deferred_update = mainConfig->deferred_update && !mainConfig->is_lmdb ? mainConfig->deferred_update : false;
```
progier389: Seems a bit over-complicated: 
  should maybe be: deferred_update = mainConfig->is_lmdb ? false : mainConfig->deferred_update;
Note:  this comment applies on all 4 occurrences

ldap/servers/plugins/memberof/memberof.h
```diff
@@ -140,6 +140,7 @@ typedef struct memberofconfig
     Slapi_Task *task;
     int need_fixup;
     PRBool launch_fixup;
+    PRBool is_lmdb;
```
progier389: maybe bool instead of  PRBool ?

ldap/servers/plugins/posix-winsync/posix-group-func.c
```diff
@@ -139,60 +139,7 @@ searchUid(const char *udn)
     return uid;
 }
 
-int
```
progier389, line 142: This unrelated cleanup should better be in a separate PR

ldap/servers/slapd/slapi-plugin.h
```diff
@@ -35,6 +35,7 @@ extern "C" {
 #include "nspr.h"
 #include <syslog.h>
 #include <plhash.h>
+#include <stdbool.h>
```
progier389: not sure we should change slapi-plugin.h (it is as change of interface that could have nasty side effect on customer plugins ...)

ldap/servers/plugins/memberof/memberof.h
mreynolds389: > maybe bool instead of PRBool ?

:-) missed this.  Cursor auto added it

ldap/servers/slapd/slapi-plugin.h
mreynolds389: > not sure we should change slapi-plugin.h (it is as change of interface that could have nasty side effect on customer plugins ...)

I can move it slapi-private.h I suppose

tbordaz (approved):
LGTM

progier389 (approved):
LGTM

Issue #7016, opened 2025-09-23
NULL dereference in `send_referrals_from_entry()` with empty `ref` attribute


zeff-ir:
Static analysis detected a NULL dereference in `entry.c`: when `attr != NULL` but `numValues == 0`, the `url` pointer remains `NULL` (memory was only allocated when `numValues > 0`), after which `url[numValues] = NULL;` is executed unconditionally, effectively dereferencing `url[0]` and causing a crash.

This occurs because `slapi_attr_get_numvalues(attr, &numValues)` legitimately returns `0` when the `ref` attribute has no values; in that case memory for `url` is not allocated, yet the array terminator is written without a check.

https://github.com/389ds/389-ds-base/blob/4a64bd51b4b9f0e3b70d124160b5f835057c5f4d/ldap/servers/slapd/entry.c#L3888-L3898

PR #7017, opened 2025-09-23
Issue 7016 - fix NULL deref in send_referrals_from_entry()

zeff-ir:
Static analysis (DEREF_OF_NULL.EX) flagged a possible null-pointer dereference at `entry.c` where `url[numValues] = NULL;` is written even when `numValues == 0`. If `attr != NULL` but the attribute has no values, `url` remains `NULL` because allocation was guarded by `if (numValues > 0)`, and writing the terminator dereferences `url[0]`. Fix by always allocating `(numValues + 1)` pointers when `attr != NULL`, ensuring `url[numValues]` is valid even for an empty list.

Edit:
Issue #7016 

## Summary by Sourcery

Bug Fixes:
- Ensure the url array is allocated unconditionally to prevent dereferencing a NULL pointer when an attribute has no values


progier389 (approved):
LGTM

PR #7019, opened 2025-09-24
Issue 7018 - BUG - prevent stack depth being hit

### LINKED ISSUE #7018: Excessive filters can cause stack depth to be hit
**Issue Description**
Some filters that are too excessive in size can cause stack depth to be hit during optimise.

Firstyear:
Bug Description: During filter optimise, the filter is a linked list which we traverse through recursion.

This can cause stack depth to be hit on filters with thousands of elements.

Fix Description: Limit the depth to which we recurse

fixes: https://github.com/389ds/389-ds-base/issues/7018

Review by: ???

## Summary by Sourcery

Limit recursion depth in filter optimisation to avoid hitting the call stack on large filter lists by refactoring the function into an inner helper with a depth bound and a public wrapper that sets the limit.

Bug Fixes:
- Prevent stack overflow in filter optimisation by adding a recursion depth limit

Enhancements:
- Refactor slapi_filter_optimise into a static inner function with a depth parameter
- Add a wrapper that initializes the recursion limit to 256

### DIFF
--- a/ldap/servers/slapd/filter.c
+++ b/ldap/servers/slapd/filter.c
@@ -1678,7 +1678,7 @@ filter_merge_subfilter(Slapi_Filter **list, Slapi_Filter **f_prev, Slapi_Filter 
     slapi_filter_free(*f_cur, 0);
 }
 
-/* slapi_filter_optimise
+/* slapi_filter_optimise_inner
  * ---------------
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM

Issue #7023, opened 2025-09-26
UI - if first instance that is loaded is stopped it breaks parts of the UI for other running instances


mreynolds389:
**Issue Description**

When checking for DB implementation or loading objectcalsses for the LDAP editor if the first instance is stopped then these settings are not loaded, and when you switch to a new instance that is running they are not reloaded.  This causes things to unexpectedly break for valid instances that are running.

mreynolds389:
746b44438..1e04f25a5  389-ds-base-3.0 -> 389-ds-base-3.0
a617678e5..539805df9  389-ds-base-2.7 -> 389-ds-base-2.7
fe4376adc..937280093  389-ds-base-2.6 -> 389-ds-base-2.6
a8074f285..fc931b828  389-ds-base-2.5 -> 389-ds-base-2.5
f6bc8ba34..fd9609e72  389-ds-base-2.4 -> 389-ds-base-2.4
344eba195..6cab442d7  389-ds-base-2.3 -> 389-ds-base-2.3
f4c04e7a2..bb5e14f33  389-ds-base-2.2 -> 389-ds-base-2.2

PR #7024, opened 2025-09-26
Issue 7023 - UI - if first instance that is loaded is stopped it brea…

mreynolds389:
…ks parts of the UI

Description:

When checking for DB implementation or loading objectclasses for the LDAP editor if the first instance is stopped then these settings are not loaded, and when you switch to a new instance that is running they are not reloaded. This causes things to unexpectedly break for valid instances that are running.

Relates: https://github.com/389ds/389-ds-base/issues/7023

## Summary by Sourcery

Reload LDAP object classes and database engine information when switching server instances to prevent stale UI data when the initial instance is stopped

Bug Fixes:
- Reload object classes in the LDAP editor component on serverId change
- Invoke database engine detection in monitor component when switching to a new instance

Enhancements:
- Replace DB detection error notification with console log
- Remove leftover debug console logs from the database component

### DIFF
--- a/src/cockpit/389-console/src/monitor.jsx
+++ b/src/cockpit/389-console/src/monitor.jsx
@@ -131,6 +131,7 @@ export class Monitor extends React.Component {
             } else {
                 if (this.props.serverId !== prevProps.serverId) {
                     this.loadSuffixTree(false);
+                    this.getDBEngine();
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


jchapma (approved):
LGTM

src/cockpit/389-console/src/monitor.jsx
```diff
@@ -643,10 +644,7 @@ export class Monitor extends React.Component {
                 })
                 .fail(err => {
                     const errMsg = JSON.parse(err);
-                    this.props.addNotification(
-                        "error",
-                        cockpit.format("Error detecting DB implementation type - $0", errMsg.desc)
-                    );
+                    console.log("getDBEngine - Error detecting DB implementation type - $0", errMsg.desc);
```
droideck: `$0` won't work with `console.log`, so it should be this, probably:

```suggestion
                    console.log("getDBEngine - Error detecting DB implementation type -", errMsg.desc);
```

droideck (approved):
Besides the minor issue/typo - looks good! Ack

PR #7030, opened 2025-10-02
Issue 7029 - Add test case to measure ndn cache performance impact

### LINKED ISSUE #7029: Add test case to measure ndn cache performance impact
Add a pytest tester that measure the overall ndn cache impact  on some scenarios

progier389:
Add test case to measure ndn cache performance impact

Issue: #7029 

Reviewed by: @droideck , @tbordaz  (Thanks!)

## Summary by Sourcery

Add a performance test suite to measure the impact of the NDN cache on directory operations

Tests:
- Introduce ndncache_test.py to benchmark various search and modify scenarios with NDN cache toggled on and off and record statistical results
- Include a 5Kusers.ldif fixture file to populate test data for performance measurements

### DIFF
--- a/dirsrvtests/tests/perf/ndncache_test.py
+++ b/dirsrvtests/tests/perf/ndncache_test.py
@@ -10,14 +10,30 @@ import pytest
 import re
 from abc import ABC, abstractmethod
 from ldap.controls.sss import SSSRequestControl
-from lib389._constants import DN_DM, DEFAULT_SUFFIX, PASSWORD, DEFAULT_BENAME
 from lib389.backend import Backends
+from lib389.config import BDB_LDBMConfig, LMDB_LDBMConfig
 from lib389.config import Config
 from lib389.index import Indexes
+from lib389._mapped_object import DSLdapObject
 from lib389.properties import TASK_WAIT
 from lib389.topologies import topology_st as topo
-from lib389.utils import ldap, os, time, logging, ds_is_older, get_ldapurl_from_serverid
 from statistics import fmean, stdev
+
+from lib389._constants import (
+    DEFAULT_BENAME,
+    DEFAULT_SUFFIX,
+    DN_DM,
+    DN_USERROOT_LDBM,
+    PASSWORD,
+)
+
+from lib389.utils import (
+    ldap, os, time, logging,
+    ds_is_older,
+    ensure_bytes,
+    get_default_db_lib,
+    get_ldapurl_from_serverid,
+)
 
 pytestmark = pytest.mark.tier3
 
@@ -41,8 +57,17 @@ class Scenario:
         self._desc = None
         self._name = None
 
+    def preop(self):
+        # Run before measured operation
+        pass
+
     @abstractmethod
-    def run(self):
+    def op(self):
+        # Operation to measure
+        pass
+
+    def postop(self):
+        # Run after measured operation
         pass
 
     def __str__(self):
@@ -51,6 +76,44 @@ class Scenario:
     def description(self):
         return self._desc
 
+    @staticmethod
+    def average(data):
+        data.sort()
+        log.info(f'RAW DATA: len(data)={len(data)} {data}')
+        assert len(data) >= NB_MEANINGFULL_MEASURES
+        for i in range(len(data)-1, NB_MEANINGFULL_MEASURES-1, -1):
+            del data[i]
+        log.info(f'COOKED DATA: len(data)={len(data)} {data}')
+        m = fmean(data)
+        v = stdev(data, m)
+        v = v/m * 100
+        return ( m, v, str(data) )
+
+
+    def measure(self, inst, conn):
+        name = str(self)
+        log.info(f'Perform {NB_MEASURES} of {name}')
+        data = []
+        self.ldc = conn
+        try:
+            for _ in range(NB_MEASURES):
+                self.preop()
+                with  open(f'{inst.ds_paths.log_dir}/access', 'a+') as logfd:
+                    pos = os.fstat(logfd.fileno()).st_size
+                    self.op()
+                    logfd.seek(pos)
+                    for line in iter(logfd.readline, ''):
+                        res = re.match(r'.*optime=(\d+\.?\d*).*', line)
+                        if res:
+                            data.append(float(res.group(1)))
+                    self.postop()
+            result = Scenario.average(data)
+        except ldap.LDAPError as exc:
+            log.error(f'Scenario {scen} failed because of {exc}')
+            result = ( 0, 0, [] )
+        log.info(f'result (average, normalized standard deviation, data) of {name} is {result}')
+        return result
+
 
 class Scen1(Scenario):
     def __init__(self):
@@ -58,7 +121,7 @@ class Scen1(Scenario):
         self._name = 'scen1'
         self._desc = 'Subtree search'
 
-    def run(self):
+    def op(self):
         basedn = f'ou=Technology,ou=people,{DEFAULT_SUFFIX}'
         result = self.ldc.search_ext_s(
             base=basedn,
@@ -75,7 +138,7 @@ class Scen2(Scenario):
         self._name = 'scen2'
         self._desc = 'Server Side Sorted Subtree search'
 
-    def run(self):
+    def op(self):
         basedn = f'ou=Technology,ou=people,{DEFAULT_SUFFIX}'
         result = self.ldc.search_ext_s(
             base=basedn,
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,503 @@
...
+        data.sort()
+        log.info(f'RAW DATA: len(data)={len(data)} {data}')
+        assert len(data) >= NB_MEANINGFULL_MEASURES
+        for i in range(len(data)-1, NB_MEANINGFULL_MEASURES-1, -1):
+            del data[i]
+        log.info(f'COOKED DATA: len(data)={len(data)} {data}')
+        m = fmean(data)
+        v = stdev(data, m)
+        v = v/m * 100
+        return ( m, v, str(data) )
```
droideck: Do I understand correctly that you, basically, keeping only 20 fastest measurements?
So if we have two cases:
1. High hit rate (90%)
2. Low hit rate (30%)

They will score very similarly with the approach. So we, basically, will never see a lot of cache misses unless things are really bad, won't we?

But maybe that's okay for our NDN cache effectiveness case... I need to think, but I decided to start a discussion anyway, so I can clarify. :)

droideck (commented):
I don't think it'll provide much benefit, but should we set `nsslapd-accesslog-logbuffering: off` just for a bit more precision?

And do you think it's worth also to collect `normalizeddncache*` stats?

I also wonder if it's worth to test cache eviction scenario, but that's just a thought.
What you already have is hugely impressive!

dirsrvtests/tests/perf/ndncache_test.py
progier389: You are right on both points: only the 20% fastest measure are taken in account and as the ndn cache is large enough to store all dn ans the operation is replayed again and again, in the fastest case there is no cache miss (when ndn cache is enabled)
Not really an issue when trying to evaluate  the impact  of enabling or not the ndn cache

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,566 @@
...
+    index.ensure_attr_state( { 'nsIndexType': ['eq', 'pres'] } )
+
+    index = indexes.create(properties={
+        'cn': 'modifiersName',
+        'nsSystemIndex': 'false',
+        'nsIndexType': ['eq']
+        })
+
+    index = indexes.get('member')
+    index.delete()
```
tbordaz, line 327: Humm, weird. I was expecting the creation of  substring index for 'uniqueMember' (scenario 4 and 5).

tbordaz (commented):
Large and very nice test. A minor question

dirsrvtests/tests/perf/ndncache_test.py
progier389: You are right, that is uniqueMember that should be indexed 
member should not be indexed (for scen 6)

tbordaz (approved):
LGTM

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+    scens[0] = ''
+
+    nbfiles = len(res1)
+    nbscen = len(gains)
+    nbdata = len(res1[1][1])
+
+    nbrows = nbfiles * nbscen + 1
+    nbcols = nbdata + 2 * nbscen + 4
+
+    idx_grqph_data = nbdata + 2
```
droideck: Minor typo:)

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+            scope=ldap.SCOPE_BASE,
+            filterstr='(objectclass=*)',
+            attrlist=list(STAT_ATTRS))
+        log.debug(f'getndncache_stats res={res}')
+        res = res[0][1]
+        return { a: ensure_str(res[a][0]) for a in STAT_ATTRS }
+
+
+
+    def measure(self, inst, conn):
```
droideck, line 134: This function takes forever to execute on a VM. I understand that we might focus on a faster machine execution, but if we can optimize it, it won't hurt (I can't realistically test it on any moderate VM now)

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+        name = str(self)
+        log.info(f'Perform {NB_MEASURES} of {name}')
+        data = []
+        self.ldc = conn
+        aggregated_stats = { a: 0 for a in STAT_ATTRS }
+        try:
+            for _ in range(NB_MEASURES):
+                self.preop()
+                pre_stats = self.getndncache_stats()
+                with  open(f'{inst.ds_paths.log_dir}/access', 'a+') as logfd:
```
droideck: I don't think we need to open the access log file on every iteration.
It can be done before  `for _ in range(NB_MEASURES):` like 

```python
log_file = f'{inst.ds_paths.log_dir}/access'
with open(log_file, 'a+') as logfd:
    for _ in range(NB_MEASURES):
        self.preop()
        ...
```

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+                    pos = os.fstat(logfd.fileno()).st_size
+                    self.op()
+                    logfd.seek(pos)
+                    for line in iter(logfd.readline, ''):
+                        res = re.match(r'.*optime=(\d+\.?\d*).*', line)
```
droideck: Might be worth to precompile the regex.

dirsrvtests/tests/perf/ndncache_test.py
droideck: Also, we can just open it in Binary mode, I think ('rb' instead of 'a+'). We don't need any encoding/decoding here, AFAICS.

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+                        if res:
+                            data.append(float(res.group(1)))
+                post_stats = self.getndncache_stats()
+                for k,v in post_stats.items():
+                    aggregated_stats[k] += float(v) - float(pre_stats[k])
+                self.postop()
+            result = Scenario.average(data)
+        except ldap.LDAPError as exc:
+            log.error(f'Scenario {scen} failed because of {exc}')
```
droideck:
```suggestion
            log.error(f'Scenario {name} failed because of {exc}')
```

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,577 @@
...
+        self._name = 'scen4'
+        self._desc = 'Modify member of large group with substring index'
+
+    def op(self):
+        basedn = f'cn=all_users,ou=groups,{DEFAULT_SUFFIX}'
+        value = b'uid=pwynn,ou=Information Technology,ou=people,dc=example,dc=com'
+        mods = [ ( ldap.MOD_DELETE, 'uniqueMember', [value,] ), ]
+        result = self.ldc.modify_s(basedn, mods)
+        mods = [ ( ldap.MOD_ADD, 'uniqueMember', [value,] ), ]
+        result = self.ldc.modify_s(basedn, mods)
```
droideck, line 241: Scen4 takes a lot of time when I execute in `batch` mode...
Large group + substring index expensive rebuild on the large group + 200 index rebuilds for 100 measures. Do I understand it correctly?

I don't have a hard opinion here and it might be okay and expected to wait for the powerful test case.
I wonder if it'll be okay to make a custom number of measures or a smaller group here. WDYT?

droideck (changes requested):

dirsrvtests/tests/perf/ndncache_test.py
progier389: Originally the file was open with a+ to be positioned at the end and position was read with ftell 
but I changed my mind and ended to use fstat so yes we can factorise

dirsrvtests/tests/perf/ndncache_test.py
```diff
@@ -0,0 +1,627 @@
...
+if __name__ == '__main__':
+    CURRENT_FILE = os.path.realpath(__file__)
+    args = parser.parse_args()
+    assert args.nbloops > 0
+    assert args.measure >= 10
+    NB_MEASURES = args.measure
+    NB_MEANINGFULL_MEASURES = int(NB_MEASURES/5)
+    if args.csv:
+        # Generate csv file
+        generate_csv(CSV_FILE)
```
droideck, line 616: It fails here if I run with '-c':

```
# python3 dirsrvtests/tests/perf/ndncache_test.py -b -c
  File "/mnt/tests/rhds/tests/upstream/ds/dirsrvtests/tests/perf/ndncache_test.py", line 612, in <module>
    generate_csv(CSV_FILE)
    ~~~~~~~~~~~~^^^^^^^^^^
  File "/mnt/tests/rhds/tests/upstream/ds/dirsrvtests/tests/perf/ndncache_test.py", line 523, in generate_csv
    scens = [ data[0] for data in res1[0] ]
                                  ~~~~^^^
IndexError: list index out of range

```

droideck (approved):
Besides the minor issue - looks great!
Very impressive performance tool.

Issue #7035, opened 2025-10-07
RFE - memberOf - adding scoping for specific groups


mreynolds389:
**Description**

Add ability to include/exclude specific groups that memberOf will act upon

mreynolds389:
The CLI was adjusted in:  https://github.com/389ds/389-ds-base/issues/7041

PR #7037, opened 2025-10-08
Issue 7035 - RFE - memberOf - adding scoping for specific groups

mreynolds389:
Description:

Add ability to include/exclude specific groups that memberOf will act upon

Design doc: https://www.port389.org/docs/389ds/design/memberof-specific-group-scoping-design.html

Relates: https://github.com/389ds/389-ds-base/issues/7035

## Summary by Sourcery

Introduce scoping for specific groups in the MemberOf plugin by allowing users to explicitly include or exclude particular group DNs and restrict by object classes.

New Features:
- Add memberOfSpecificGroup, memberOfExcludeSpecificGroup and memberOfSpecificGroupOC attributes to plugin configuration
- Expose new "Specific Group", "Exclude Specific Group" and "Specific Group OC" fields in the cockpit UI

Enhancements:
- Extend core plugin scoping logic to respect include/exclude specific group lists and object class filters
- Update configuration validation and application to handle new scoping attributes
- Refactor various plugin functions to pass entry context for scoping checks

Tests:
- Add dsconf and plugin unit tests covering specific group inclusion, exclusion and objectClass-based scenarios

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -107,6 +107,13 @@ typedef struct _task_data
     char *filter_str;
 } task_data;
 
+typedef struct _MemberofEntryInfo
+{
+    Slapi_Entry *e;
+    Slapi_DN *sdn;
+    bool group;
+} MemberofEntryInfo;
+
 /*** function prototypes ***/
 
 /* exported functions */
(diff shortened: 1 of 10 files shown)


ldap/servers/plugins/memberof/memberof.c
```diff
@@ -348,6 +356,30 @@ memberof_internal_postop_init(Slapi_PBlock *pb)
...
+
+    if (!config) {
+        /* No config means to "always" treat this entry as a group */
+        entry_info->group = true;
+        return;
+    }
+
+    if (config->specificGroupFilter || config->excludeSpecificGroupFilter) {
+        /* Only set entry info if we have specific group filters defined */
+        for (size_t i = 0; config->specificGroupOC && config->specificGroupOC[i]; i++) {
```
tbordaz, line 374: Couldn't we rely on the specificGroupFilter (excludeSpecificGroupFilter) for that and get rid of 'memberOfSpecificGroupOC'
For example if we want memberof to apply on 'groupOfNames' groups whose owner is 'cn=admin,dc=suffix'. Instead of defining
```
memberOfSpecificGroupOC: groupOfNames
 memberOfSpecificGroupFilter: (owner = "cn=admin,dc=suffix")
```

The config could be
`memberOfSpecificGroupFilter: (&(objectclass=groupOfNames)(owner = "cn=admin,dc=suffix")`

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -471,8 +506,10 @@ deferred_modrdn_func(MemberofDeferredModrdnTask *task)
     memberof_unlock_config();
 
     /* Need to check both the pre/post entries */
-    if ((pre_sdn && !memberof_entry_in_scope(&configCopy, pre_sdn)) &&
-        (post_sdn && !memberof_entry_in_scope(&configCopy, post_sdn)))
+    memberof_set_entry_info(pre_e, NULL, &pre_entry_info);
+    memberof_set_entry_info(post_e, NULL, &post_entry_info);
+    if ((pre_sdn && !memberof_entry_in_scope(&configCopy, &pre_entry_info)) &&
+        (post_sdn && !memberof_entry_in_scope(&configCopy, &post_entry_info)))
```
tbordaz, line 512: Should we care about the post entry ?
If the initial entry did not match the filter then it looks to me enough to do nothing.

ldap/servers/plugins/memberof/memberof.c
```diff
@@ -2446,21 +2524,53 @@ memberof_oktodo(Slapi_PBlock *pb)
...
+ *
+ * Test if the entry matches any of the specific group filters
+ * return 1 if the entry matches any of the specific group filters
+ * return 0 if the entry does not match any of the specific group filters
+ */
+static int
+memberof_test_specific_filters(MemberOfConfig *config, MemberofEntryInfo *entry_info)
+{
+    if (entry_info->e && entry_info->group) {
+        for (size_t i = 0; config->excludeSpecificGroupFilter && config->excludeSpecificGroupFilter[i]; i++) {
```
tbordaz, line 2538: @Firstyear raise this point. Should we support exclude filter ?
Clever specificGroupFilter could do include/exclude at the same time.

tbordaz (commented):
The patch looks good. Few questions

ldap/servers/plugins/memberof/memberof.c
mreynolds389: The purpose of memberOfSpecificGroupOC is mostly needed by group membership checks (users verses groups).  Here is the scenario that causes the issues.  We have specific filter that matches all the groups we want to manage.  But when you add "users" to that group they do NOT match the "specific" group filter and they are all skipped.  So we need to know when we want to check the member against the filters or not.  For users we do NOT want to check the filtering rules, but for nested groups we do.  So in order to distinguish between what is a "user" and what is a "group" we need to define these objectclasses.  This is where the entry_info struct is used.  I think this is described better in the design doc (maybe not?)

ldap/servers/plugins/memberof/memberof.c
mreynolds389: Hmm I was just following the existing scoping logic used in this function, but if we rename a group to match an exclusion group then we shouldn't update it.  So we do need to look at both IMO.

ldap/servers/plugins/memberof/memberof.c
mreynolds389: > @Firstyear raise this point. Should we support exclude filter ? Clever specificGroupFilter could do include/exclude at the same time.

The problem is that you would only be allowed to have one filter (one single valued config attribute), and that single filter would have to cover all the cases (because including/excluding requires different logic when separate filters are being used).  So it could be a really large complex search filter.  I was trying to make it easy for an admin to implement what they want without having to be an LDAP expert.  I hear what you saying, but it is harder to implement for Admins.  I was just looking at this from a usability perspective, not from a super efficient LDAP developer perspective :-)  I can rework the entire PR again if that's what everybody wants (no problem just let me know).

tbordaz: Yes that makes sense. Anyway if an admin wants to code everything (include/exclude) he can do it. I agree that include/exclude filter offers flexibility.
Thanks

ldap/servers/plugins/memberof/memberof.c
tbordaz: Okay for me

ldap/servers/plugins/memberof/memberof.c
tbordaz: Ok, it is looking like a shortcut to avoid evaluating a filter on users. That makes sense. Thanks

ldap/servers/plugins/memberof/memberof.c
mreynolds389: > Yes that makes sense. Anyway if an admin wants to code everything (include/exclude) he can do it. I agree that include/exclude filter offers flexibility. Thanks

What I'll do is update the design doc stating that a single filter can handle it all, and provide some examples.

progier389 (approved):
LGTM

PR #7043, opened 2025-10-14
Issue 7042 - Enable global_backend_lock when memberofallbackend is enabled

### LINKED ISSUE #7042: Enable global_backend_lock when memberofallbackend is enabled
**Issue Description**
When the memberOf plugin is configured with the memberOfAllBackends=on option, concurrent updates to group memberships across multiple backends can lead to deadlock.

**Steps to Reproduce**
1. Configure a standalone Directory Server instance.
2. Create two separate backends
3. Under each backend, create ou=people and ou=groups containers.
4. Create several test users under each backend
5. Create one test group under each backend
6. Enable the memberOf plugin and set its scope to cover all backends:
7. Start concurrent threads that repeatedly add and remove users from groups across backends
8. Observe that some server threads may deadlock

**Expected results**
All concurrent add/remove operations should complete successfully,  without deadlocks.

**Additional context**
The fix for this is to enable nsslapd-global-backend-lock=on when  memberOfAllBackends=on. This will likely have performance implications and should be documented.

jchapma:
Description: When the memberOf plugin is configured with memberOfAllBackends=on option, concurrent updates to group memberships across multiple backends can lead to deadlock.

Fix: The plugin now automatically updates the nsslapd-global-backend-lock config to match the state of memberOfAllBackends, ensuring global backend locking.

Fixes: https://github.com/389ds/389-ds-base/issues/7042

Reviewed by:

## Summary by Sourcery

Automatically align nsslapd-global-backend-lock with the memberOfAllBackends setting to prevent cross-backend deadlocks and add a concurrency test to verify deadlock-free behavior.

New Features:
- Automatically enable or disable the global backend lock when memberOfAllBackends is toggled

Tests:
- Add a multi-backend concurrency test with worker and watchdog threads to validate no deadlocks under memberOfAllBackends=on

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_config_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_config_test.py
@@ -320,6 +320,66 @@ def test_healthcheck_MO_plugin_substring_index(topology_st):
     standalone.restart()
 
 
+def test_healthcheck_MO_plugin_global_lock(topology_st):
+    """Verify HealthCheck returns DSMOLE0003 when memberOfAllBackends=on
+    and nsslapd-global-backend-lock=off
+
+    :id: 2a93f824-4bb7-4958-adcb-4e8ba17bfcc0
+    :setup: Standalone instance
+    :steps:
+        1. Create DS instance
+        2. Configure the instance with MO Plugin
+        3. Set memberOfAllBackends and nsslapd-global-backend-lock to off
+        4. Use HealthCheck without --json option
+        5. Use HealthCheck with --json option
+        6. Set memberOfAllBackends to on and nsslapd-global-backend-lock to off
+        7. Use HealthCheck without --json option
+        8. Use HealthCheck with --json option
+
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Healthcheck reports no issue found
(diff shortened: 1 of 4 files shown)


dirsrvtests/tests/suites/plugins/memberof_test.py
```diff
@@ -2897,6 +3016,140 @@ def test_memberof_auto_add_oc(topology_st):
     log.info('Test complete.')
 
 
+def test_memberof_all_backends(topology_st):
+    """Test cross backend, concurrent member updates with "all backends" scope.
+
+    :id: f8d8f578-17da-4122-909a-43fb26c6adbe
```
tbordaz: This test looks good. I think you may also create a test that triggers a deadlock when global_backend_lock is disable.
For example I would expect a risk of deadlock with

- users backend / groups backend
- memberof update of users when groups are update
- referential integrity update when users are deleted
- A load updating groups (memberof updates users)
- A load deleting users (referential integrity updates groups)

jchapma:
Updated PR to add a healthcheck, we also log a warning to the error logs if allbackends=on and nsslapd-global-backend-lock=off

progier389 (approved):
LGTM, 
Note: healthcheck test is failing but it is not related to this change

PR #7050, opened 2025-10-16
Issue 7049 - RetroCL plugin generates invalid LDIF

### LINKED ISSUE #7049: RetroCL plugin generates invalid LDIF
**Issue Description**
`changes` attribute contains invalid LDIF:

```
replace: modifiersName
modifiersName: cn=MemberOf Plugin,cn=plugins,cn=config
-
modifyTimestamp: 20250903092211Z
-
```
Line `replace: modifyTimestamp` is missing.

 - Platform: RHEL, Fedora
 - Package and version: main

**Steps to Reproduce**
Happens in a replicated environment at a customer site.

vashirov:
Bug Description:
When a replicated modification marked with LDAP_MOD_IGNORE is logged,
`changes` attribute contains invalid LDIF:

```
replace: modifiersName
modifiersName: cn=MemberOf Plugin,cn=plugins,cn=config
-
modifyTimestamp: 20250903092211Z
-
```
Line `replace: modifyTimestamp` is missing.

A similar issue is present in audit log:
```
time: 20251031064114
dn: ou=tuser,dc=example,dc=com
result: 0
changetype: modify
add: objectClass
objectClass: nsMemberOf
-
replace: modifiersName
modifiersName: cn=MemberOf Plugin,cn=plugins,cn=config
-
-
```
Dash separator is logged, while the operation is not.
This issue is not present wheh JSON format is used.

Fix Description:
* retrocl_po.c: add a default case to skip the entire modification if it
  has LDAP_MOD_IGNORE flag.
* auditlog.c: write the dash separator only if operation type is not
  LDAP_MOD_IGNORE

Fixes: https://github.com/389ds/389-ds-base/issues/7049

## Summary by Sourcery

Bug Fixes:
- Mask LDAP_MOD_IGNORE flag in mod_op evaluation alongside LDAP_MOD_BVALUES to ensure correct operation detection and valid LDIF output

### DIFF
--- a/ldap/servers/plugins/retrocl/retrocl_po.c
+++ b/ldap/servers/plugins/retrocl/retrocl_po.c
@@ -99,6 +99,9 @@ make_changes_string(LDAPMod **ldm, const char **includeattrs)
             addlenstr(l, ldm[i]->mod_type);
             addlenstr(l, "\n");
             break;
+        default:
+            /* LDAP_MOD_IGNORE or unknown - skip this mod entirely */
+            continue;
         }
         for (j = 0; ldm[i]->mod_bvalues != NULL &&
                     ldm[i]->mod_bvalues[j] != NULL;
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -828,8 +828,8 @@ write_audit_file(
                         slapi_ch_free((void **)&buf);
                     }
                 }
+                addlenstr(l, "-\n");
             }
-            addlenstr(l, "-\n");
         }
         break;
 


ldap/servers/plugins/retrocl/retrocl_po.c
```diff
@@ -83,7 +83,8 @@ make_changes_string(LDAPMod **ldm, const char **includeattrs)
                 continue;
             }
         }
-        switch (ldm[i]->mod_op & ~LDAP_MOD_BVALUES) {
+
+        switch (ldm[i]->mod_op & ~(LDAP_MOD_BVALUES | LDAP_MOD_IGNORE)) {
```
progier389: Unless mistaken, LDAP_MOD_IGNORE flag is set when a value get removed from modifiers list) 
  so the modifier should be ignored and should not appear in the retrocl ldif
==> Should continue the for loop if LDAP_MOD_IGNORE is set

progier389 (changes requested):
IMHO The fix should be done the other way:
 (The problem is not that the replace line is missing but that the value lines is added while it should be ignored)

ldap/servers/plugins/retrocl/retrocl_po.c
vashirov: What's surprising to me is why only `modifyTimestamp` had this flag in this change.

progier389: Without any reproducer, I am not sure we will be able to answer that question.
My guess is that the original modify request contained a modifier to add/replace modifyTimestamp
that was ignored as this attribute cannot be explicitly modified

vashirov: I was able to reproduce it with the customer's script, but it's too complex. I wasn't able to trim it down. But after reading the code, I think this is what happens:

1. User is created without the objectClass containing memberOf
2. Adding user to group on Server1 (memberOf plugin adds objectClass)
3. Simultaneously adding user to group on Server2 (memberOf plugin adds objectClass)
4. Server2's change replicates to Server1
5. Server1 detects conflict (objectClass already exists)
6. Backend marks modification with `LDAP_MOD_IGNORE` https://github.com/389ds/389-ds-base/blob/06e1268aff69268a826f233f0e86b3869cb5d3db/ldap/servers/slapd/back-ldbm/index.c#L627-L628
7. RetroCL and audit log contain lines that shouldn't be present. 

I've updated the description and the code.

progier389 (approved):
LGTM

PR #7068, opened 2025-10-23
Issue 7065 - A search filter containing a non normalized DN assertion…

### LINKED ISSUE #7065: A search filter containing a non normalized DN  assertion does not return matching entries
**Issue Description**
This bug is a regression introduced with #6172. The issue is getting worse as now sorted valueset is enabled even with a single value. The problem is that assertion value is not normalized before comparing the values stored in the entry (that are normalized).

It applies to all recent versions built with https://github.com/389ds/389-ds-base/issues/6172 (1.4.3 and later)

**Steps to Reproduce**
will create a testcase

**Expected results**
The server should return all the matching entries even if the filter contains a provided DN  not normalized

tbordaz:
… does not return matching entries

Bug description:
	This bug is a regression introduced with #6172.
	The issue is getting worse as now sorted valueset is enabled even with a single value.
	The problem is that assertion value is not normalized before comparing with the values
	stored in the entry (that are normalized).

Fix description:
	Normalize the DN value provided by the filter before the
	comparison

fixes: #7065

Reviewed by:

## Summary by Sourcery

Bug Fixes:
- Normalize the DN value in the filter before comparing to ensure matching entries with normalized DNs.

### DIFF
--- a/dirsrvtests/tests/suites/filter/basic_filter_test.py
+++ b/dirsrvtests/tests/suites/filter/basic_filter_test.py
@@ -6,12 +6,14 @@
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ----
 
+import ldap
 import pytest, os
 
 from lib389._constants import DEFAULT_SUFFIX, PW_DM
 from lib389.topologies import topology_st as topo
 
 from lib389.idm.user import UserAccount, UserAccounts
+from lib389.idm.group import UniqueGroups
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):

progier389 (commented):
Looks good. It will be even better with a CI test ! 😉

progier389 (approved):
LGTM

Issue #7071, opened 2025-10-28
(&(cn:dn:=groups)) no longer returns results


mreynolds389:
**Issue Description**

In RHEL 8/9/10, a specific filter has stopped returning results:

     (&(cn:dn:=groups))

This happens when an "and" filter is used, but it only contains one filter component

PR #7072, opened 2025-10-28
Issue 7071 - search filter (&(cn:dn:=groups)) no longer returns results

mreynolds389:
Description:

When processing an "and" filter and it only contains one filter component then the logic in the code breaks down and the filter is seen as not matching.

The logic breaks down because we are not setting "nomatch" after the access check is successful. If there are two components then it works fine because we do the access check on the first filter component and set that the access check was done(access_check_done), but "nomatch" is not set yet. So when the next filter component is checked for access we see that the access check was done and then we set "nomatch".

To recap we always need to set "nomatch" when the access check is successful in order to handle the case where an "and" fitler only has one component.

Relates: https://github.com/389ds/389-ds-base/issues/7071

## Summary by Sourcery

Ensure DN-based filters return correct results by setting nomatch after successful access checks in single-component AND filters and augment filter tests with new positive and negative cases

Bug Fixes:
- Set nomatch when access check succeeds in single-component AND filters to restore correct DN filter matching

Tests:
- Add positive and negative tests for DN filters in complex AND filter scenarios


droideck (approved):
LGTM

Issue #7078, opened 2025-11-03
audit json logging does not encode binary values


mreynolds389:
**Issue Description**

We need to base64 encode binary values like nsState.  One side effect is that the UI fails to read the log once it encounters this value.

PR #7079, opened 2025-11-04
Issue 7078 - audit json logging does not encode binary values

mreynolds389:
Description:

Audit log does encode binary values, and this breaks the UI when it tries displaying the log contents. When the value is not "printable" base64 encode it.

Relates: https://github.com/389ds/389-ds-base/issues/7078

## Summary by Sourcery

Enable Base64 encoding of non-printable binary values in JSON audit logs, add a test for this behavior, and enhance console UI log viewers with error handling and formatting cleanup.

New Features:
- JSON audit logging now Base64 encodes non-printable binary attribute values
- Add test verifying binary values are Base64 encoded and printable values remain unchanged

Enhancements:
- Improve error handling in 389-console log monitor components by displaying cockpit.error on log retrieval failures
- Clean up trailing whitespace and standardize button JSX formatting in multiple log viewer components

Tests:
- Add audit_json_logging_with_binary_value test to validate proper encoding of binary values in audit logs

### DIFF
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -12,6 +12,7 @@
 #endif
 
 #include "slap.h"
+#include <plbase64.h>
 
 /*
  * JCM - The audit log might be better implemented as a post-op plugin.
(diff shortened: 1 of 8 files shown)


ldap/servers/slapd/auditlog.c
```diff
@@ -607,17 +608,38 @@ write_audit_file_json(Slapi_PBlock *pb, Slapi_Entry *entry, int logtype,
...
+                            }
                         }
                     }
-                    json_object_object_add(mod, "values", val_list);
+                    if (json_object_array_length(val_list) > 0) {
+                        json_object_object_add(mod, "values", val_list);
+                    }
+                    if (json_object_array_length(val_encoded_list) > 0) {
+                        json_object_object_add(mod, "values_encoded", val_encoded_list);
+                    }
```
droideck, line 648: If any of the `val_list` and `val_encoded_list` is empty, should we free them?

droideck (commented):
Mostly looks good, I'm wondering if it might be valuable to add a few more test cases to ensure robustness across edge scenarios? Some ideas:
* Very large binary values - to ensure we handle them gracefully:
```python   # Test with ~2MB binary data
   large_binary_data = bytes([(i % 256) for i in range(2 * 1024 * 1024)])
```  
* Perform a modify with multiple attributes where nsState attribute, for example,  contains binary data, and we'll also modify description (text) in the same operation.
* I also wonder how attribute with a b'{"malicious": "json"}' value will look like, but it's probably out of scope of this PR. Just got me wondering :)

ldap/servers/slapd/auditlog.c
mreynolds389: ASAN did not complain

droideck: Sure)
I guess it's just to be on a safe side (and I suspect coverity scan may complain here)
Also, I see that in `add_entry_attrs_ext` there's a code like this:

```c
    if (json_object_array_length(id_list) > 0) {
        json_object_object_add(json_log, "id_list", id_list);
    } else {
        /* free empty list */
        json_object_put(id_list);
    }
```

mreynolds389: Ok there was a leak.  I was trying to debug a different issue with a new ASAN build and then it started reporting those leaks.  Not sure why the first ASAN build did not "work".  Anyway I'll be fixing it

droideck (approved):
LGTM!

progier389 (approved):
LGTM

Issue #7084, opened 2025-11-06
UI - schema - after sorting attributes the expanded row is broken


mreynolds389:
**Issue Description**

When sorting attributes the expanded row is not properly set and it crashes the browser when trying to see it

mreynolds389:
f4c30dba1..316ec594d  389-ds-base-3.0 -> 389-ds-base-3.0
8f0c27516..939a4246c  389-ds-base-2.7 -> 389-ds-base-2.7
c56e5b57a..bc0a338e9  389-ds-base-2.6 -> 389-ds-base-2.6
c350e51dc..b26fc6b66  389-ds-base-2.5 -> 389-ds-base-2.5
473b2125d..cb0e3daef  389-ds-base-2.4 -> 389-ds-base-2.4

PR #7085, opened 2025-11-06
Issue 7084 - UI - schema - sorting attributes breaks expanded row

mreynolds389:
Description:

When sorting attributes the expanded row is not properly set and it crashes the browser when trying to see it. The problem is that during sorting we are not transfering the "attribute data" to the new sorted row.

Relates: https://github.com/389ds/389-ds-base/issues/7084

Reviewed by: ?

## Summary by Sourcery

Bug Fixes:
- Transfer originalData to sorted rows and their expanded rows to avoid missing attribute data after sorting


droideck (approved):
Looks good! Ack

PR #7095, opened 2025-11-13
Issue 1793 - RFE - Implement dynamic lists

mreynolds389:
Description:

Implement a new preoperation plugin to build dynamic content based of LDAP URI's. Configuration includes an identifying objectclass to mark an entry as a dynamic content entry. Another setting for the attribute that contains the LDAP URI, and an attribute for storing the dynamic content. Attributes specified in the LDAP URI override the content attribute and instead write that attribute's value into the dynamic content entry.

Design doc: https://www.port389.org/docs/389ds/design/dynamic-lists-design.html

Relates: https://github.com/389ds/389-ds-base/issues/1793

ASAN tests pass

## Summary by Sourcery

Implement the Dynamic Lists feature by adding a new preoperation plugin that builds entry attributes from LDAP URLs, integrate it into the server upgrade path, CLI, and 389-console UI, and provide comprehensive build support and tests.

New Features:
- Introduce Dynamic Lists preoperation plugin to dynamically populate entry content from LDAP URLs
- Support CLI management of the Dynamic Lists plugin via new dsconf `dynamic-lists` and `dynamic-lists config-entry` commands
- Add dynamic lists configuration UI in 389-console for plugin settings and shared config entries

Enhancements:
- Extend lib389 with DynamicListsPlugin and DynamicListsConfig classes for automated plugin and config entry management
- Add server upgrade path to create the default Dynamic Lists plugin entry if missing
- Integrate the build rules for libdynamic-lists-plugin into Makefile.am

Build:
- Add libdynamic-lists-plugin to the serverplugin libraries in Makefile.am

Tests:
- Add dsconf_dynamic_lists_test suite for CLI operations (status, show, set, config-entry)
- Add functional integration tests under dirsrvtests to verify dynamic list population based on memberURL

Chores:
- Include `dynamic-lists` parser in lib389 CLI subcommands
- Adjust existing plugin JS code to support attribute normalization for dynamic lists

### DIFF
--- a/dirsrvtests/tests/suites/plugins/dynamic_lists_test.py
+++ b/dirsrvtests/tests/suites/plugins/dynamic_lists_test.py
@@ -1,0 +1,422 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+import ldap
+from lib389._constants import *
+from lib389.topologies import topology_st as topo
+from lib389.plugins import DynamicListsPlugin
+from lib389.idm.user import UserAccounts
+from lib389.idm.group import Groups, Group
+from lib389.idm.organizationalunit import OrganizationalUnits
+
+pytestmark = pytest.mark.tier1
+
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="function")
+def dynamic_lists_setup(topo, request):
+    log.info("Creating test users")
+    users = UserAccounts(topo.standalone, DEFAULT_SUFFIX)
+    user_dns = []
+    num_users = 5
+
+    for i in range(1, num_users + 1):
+        user = users.create(properties={
+            'uid': f'testuser{i}',
+            'cn': f'testuser{i}',
+            'sn': f'User{i}',
+            'uidNumber': f'{1000 + i}',
+            'gidNumber': f'{1000 + i}',
+            'homeDirectory': f'/home/testuser{i}',
+        })
+        user_dns.append(user.dn)
+        log.info(f"Created user: {user.dn}")
+
+    def fin():
+        for user in users.list():
+            user.delete()
+    request.addfinalizer(fin)
+
+    return user_dns
+
+
+def test_dynamic_lists_memberurl_all_users(topo, dynamic_lists_setup):
+    """Test dynamic lists plugin with memberURL filter (uid=testuser*) to list all users
+
+    :id: 6913fc98-fb22-4f21-817d-4cc19be1f1cd
+    :setup: Standalone Instance
+    :steps:
+        1. Enable the dynamic lists plugin
+        2. Configure the plugin with objectclass, url_attr, and list_attr
+        3. Create a group with memberURL attribute containing LDAP URL with filter (uid=testuser*)
+        4. Search for the group and verify all users are listed as members
+    :expectedresults:
+        1. Plugin should be enabled successfully
+        2. Plugin should be configured successfully
+        3. Group should be created with memberURL successfully
+        4. All users should be listed as members of the group
+    """
+
+    inst = topo.standalone
+
+    user_dns = dynamic_lists_setup
+
+    # Step 1: Enable the dynamic lists plugin
+    log.info("Enabling dynamic lists plugin")
+    plugin = DynamicListsPlugin(inst)
+    plugin.enable()
+
+    # Step 2: Configure the plugin
+    log.info("Configuring dynamic lists plugin")
+    plugin.set_dynamiclistobjectclass('groupOfUrls')
+    plugin.set_dynamiclisturlattr('memberURL')
+    plugin.set_dynamiclistattr('member')
+
+    # Restart the server to apply plugin configuration
+    topo.standalone.restart()
+    assert plugin.status(), "Plugin should be enabled"
+
+    # Step 3: Create a group with memberURL attribute
+    log.info("Creating group with memberURL")
+    groups = Groups(topo.standalone, DEFAULT_SUFFIX)
+    group = groups.create(properties={
+        'cn': 'test_dynamic_group',
+        'objectClass': 'groupOfUrls',
+        'memberURL': f'ldap:///{DEFAULT_SUFFIX}??sub?(uid=testuser*)',
+    })
+    group_dn = group.dn
+    log.info(f"Created group: {group_dn} with memberURL: ldap:///{DEFAULT_SUFFIX}??sub?(uid=*)")
+
+    print("MARK attached gdb")
+    import time
+    time.sleep(25)
+    # Step 4: Search for the group to trigger the plugin and verify all users
+    # are listed as members. The plugin processes memberURL during search
+    # operations and dynamically adds member attributes
+    log.info("Searching for group to trigger dynamic lists plugin")
+    members = group.list_members()
+    log.info(f"Found {len(members)} members in group")
+    log.info(f"Members: {members}")
+
+    # Verify all users are members
+    assert len(members) == 5, f"Expected {5} members, got {len(members)}. Members: {members}"
+    for user_dn in user_dns:
+        assert user_dn in members, f"User {user_dn} should be a member of the group. Members: {members}"
+
+    log.info("Test completed successfully - all users are listed as members")
+
+
+def test_dynamic_lists_non_dn_attribute(topo, dynamic_lists_setup):
+    """Test dynamic lists plugin with memberURL filter (uid=testuser*) to list
+    all users homeDirectory attribute
+
+    :id: c44ebc24-4fd5-45f7-9699-b481527821f0
+    :setup: Standalone Instance
+    :steps:
+        1. Enable the dynamic lists plugin
+        2. Configure the plugin with objectclass, url_attr, and list_attr
+        3. Create a group with memberURL attribute containing LDAP URL with
+           filter (uid=testuser*) and requested attribute (homeDirectory)
+        4. Search for the group and verify all users homeDirectory attribute
+           are listed in the entry
+    :expectedresults:
+        1. Plugin should be enabled successfully
+        2. Plugin should be configured successfully
+        3. Group should be created with memberURL successfully
+        4. All users homeDirectory attribute should be listed in the entry
+    """
+
+    inst = topo.standalone
+
+    # Step 1: Enable the dynamic lists plugin
+    log.info("Enabling dynamic lists plugin")
+    plugin = DynamicListsPlugin(inst)
+    plugin.enable()
+
+    # Step 2: Configure the plugin
+    log.info("Configuring dynamic lists plugin")
+    plugin.set_dynamiclistobjectclass('groupOfUrls')
+    plugin.set_dynamiclisturlattr('memberURL')
+    plugin.set_dynamiclistattr('member')
+
+    # Restart the server to apply plugin configuration
+    topo.standalone.restart()
+    assert plugin.status(), "Plugin should be enabled"
+
+    # Step 3: Create a group with memberURL attribute
+    log.info("Creating group with memberURL")
+    groups = Groups(topo.standalone, DEFAULT_SUFFIX)
+    group = groups.create(properties={
+        'cn': 'test_dynamic_group_2',
+        'objectClass': 'groupOfUrls',
+        'memberURL': f'ldap:///{DEFAULT_SUFFIX}?homeDirectory?sub?(uid=testuser*)',
+    })
+    group_dn = group.dn
+    log.info(f"Created group: {group_dn} with memberURL: ldap:///{DEFAULT_SUFFIX}?"
+             "homeDirectory?sub?(uid=testuser*)")
+
+    # Step 4: Search for the group to trigger the plugin and verify all
+    # homeDirectories are listed in the entry
+    values = group.get_attr_vals_utf8_l('homeDirectory')
+    assert len(values) == 5, f"Expected {5} values, got {len(values)}. Values: {values}"
+    for value in ['/home/testuser1', '/home/testuser2', '/home/testuser3',
+                  '/home/testuser4', '/home/testuser5']:
+        assert value in values, f"Value {value} should be listed in the group. Values: {values}"
+
+
+def test_dynamic_lists_shared_config_entry(topo, dynamic_lists_setup):
+    """Test dynamic lists plugin with shared config entry
+
+    :id: f3a6f01b-a25c-4435-a85c-394acd2c489e
+    :setup: Standalone Instance
+    :steps:
+        1. Enable the dynamic lists plugin
+        2. Configure the plugin with shared config entry with different settings
+           than the plugin configuration
+        3. Create a group with memberURL attribute containing LDAP URL with filter (uid=testuser*)
+        4. Search for the group and verify all users are listed as members
+        5. Unconfigure the shared config entry
+    :expectedresults:
+        1. Plugin should be enabled successfully
+        2. Plugin should be configured successfully
+        3. Group should be created with memberURL successfully
+        4. All users should be listed as members of the group
+        5. Plugin should be configured with the plugin configuration
+    """
+
+    inst = topo.standalone
+    user_dns = dynamic_lists_setup
+
+    # Step 1: Enable the dynamic lists plugin
+    log.info("Enabling dynamic lists plugin")
+    plugin = DynamicListsPlugin(inst)
+    # Set values that should be ignored since we are going to use a shared
+    # config entry
+    plugin.set_dynamiclistobjectclass('applicationProcess')
+    plugin.set_dynamiclisturlattr('cn')
+    plugin.set_dynamiclistattr('targetDN')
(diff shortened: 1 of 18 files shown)


ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,298 @@
...
+        list_attr = ludp->lud_attrs[0];
+    }
+
+    /*
+     * Check if the client has permission to read the list attribute
+     */
+    if (dynamic_lists_check_access(pb, list_entry, list_attr) == false) {
+        goto done;
+    }
+
```
droideck: If list_attr is NULL at line 223, it gets set to listConfig.attr at line 253, but the ACL check happens before this assignment. This means we're checking access for a NULL attribute (so the listConfig.attr at line 253 is actually is not checked.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,298 @@
...
+                      "%s - Internal search based on LDAP url (%s) failed: err=%d\n",
+                      slapi_entry_get_dn_const(list_entry), url, rc);
+        goto done;
+    } else {
+        Slapi_ValueSet *list_set = slapi_valueset_new();
+        bool use_dn = false;
+
+        if (list_attr == NULL) {
+            /* Requested a special attribute to populate so we won't be using the dn */
+            use_dn = true;
```
droideck: You say `so we won't be using the dn` but then it says `use_dn = true;`. Is it a mistake or do I miss something?

dirsrvtests/tests/suites/plugins/dynamic_lists_test.py
```diff
@@ -0,0 +1,317 @@
...
+        'cn': 'test_dynamic_group',
+        'objectClass': 'groupOfUrls',
+        'memberURL': f'ldap:///{DEFAULT_SUFFIX}??sub?(uid=testuser*)',
+    })
+    group_dn = group.dn
+    log.info(f"Created group: {group_dn} with memberURL: ldap:///{DEFAULT_SUFFIX}??sub?(uid=*)")
+
+    print("MARK attached gdb")
+    import time
+    time.sleep(25)
```
droideck, line 101: Looks like gdb is still attached here. Would you mind dis-attach-it? 😄

src/lib389/lib389/cli_conf/plugins/dynamic_lists.py
```diff
@@ -0,0 +1,109 @@
...
+import ldap
+from lib389.plugins import DynamicListsPlugin, DynamicListsConfig
+from lib389.cli_conf import add_generic_plugin_parsers, generic_object_edit, generic_object_add
+from lib389.cli_base import CustomHelpFormatter
+
+arg_to_attr = {
+    'objectclass': 'dynamicListObjectclass',
+    'url_attr': 'dynamicListUrlAttr',
+    'list_attr': 'dynamicListAttr'
+}
```
droideck, line 18: I think we also need to add a config entry attr here.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,298 @@
...
+    /*
+     * Check if the entry has the configured dynamic list objectclass
+     */
+    oc_value = slapi_value_new_string(listConfig.oc);
+    if (slapi_entry_attr_has_syntax_value(list_entry, "objectclass", oc_value) == 0) {
+        /* Entry does not have the groupOfUrls objectclass */
+        goto done;
+    }
+
+    if ((url = slapi_entry_attr_get_ref(list_entry, listConfig.URLAttr)) == NULL) {
```
droideck: Is it possible to have a multiple of `memberUrl`? I think a schema definition allows it.
Also, it's multi-valued, then I think it'll be nice to have a test for that too.

droideck (commented):
A bit of the review I did today...
I'll go through the UI tomorrow.

dirsrvtests/tests/suites/plugins/dynamic_lists_test.py
mreynolds389: > Looks like gdb is still attached here. Would you mind dis-attach-it? 😄

Hmm, I guess I can do that :-)

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: > Is it possible to have a multiple of `memberUrl`? I think a schema definition allows it. Also, it's multi-valued, then I think it'll be nice to have a test for that too.

Yeah it's multivalued, and we should support that.  Also you can set what ever attribute you want - it could technically be "cn", etc.  So yeah this adds some complexity, but it's not that big of a deal.   I can see a use case for multiple urls, even though it's probably rare.  Okayyyy I'll work on that... :-)

progier389:
I think that we may also want to add  some filter rewriter (maybe in a second PR ?):
There is a fair chance that the dynamic group will be used as a plain group (i.e having application doing 
  search -s base -b groupdn member=some_user_dn to determine if an user is member of the group
Such filter will be a performance killer if the group is large and it is better to directly check if some_user_dn entry matchs the memberurl

progier389:
> Right now we add the member values as we are returning the entry (after the filter is already evaluated). So in this case it wouldn't return the group at all.

I suspect this will seriously limit the usability of the new plugin. But anyway it is worth to go ahead as we can always improve thing later on  if the inability to search the dynamic list is a real issue.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+
+        /*
+        * Check if the client has permission to read the list attribute
+        */
+        if (dynamic_lists_check_access(pb, list_entry,
+                                       list_attr ? list_attr : listConfig.attr) == false)
+        {
+            /* Access denied, so we can't proceed */
+            continue;
+        }
```
droideck, line 239: I think we might have a memory leak here, if the URL is invalid and we `continue` while not freeing the invalid `ludp`.
Also, I think we need to set `ludp = NULL` in the beginning of the loop.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+                                     NULL, 0, NULL, NULL, dynamic_lists_get_plugin_id(), 0);
+        slapi_search_internal_pb(search_pb);
+        slapi_pblock_get(search_pb, SLAPI_PLUGIN_INTOP_RESULT, &rc);
+        if (LDAP_SUCCESS != rc) {
+            /* log an error and use the plugin entry for the config */
+            slapi_log_err(SLAPI_LOG_ERR, DYNAMIC_LISTS_PLUGIN_SUBSYSTEM,
+                          "dynamic_lists_pre_entry - "
+                          "%s - Internal search based on LDAP url (%s) failed: err=%d\n",
+                          slapi_entry_get_dn_const(list_entry), url, rc);
+            continue;
```
droideck: Also here, do we need to call `slapi_pblock_destroy(search_pb);` before continue?

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+                      "dynamic_lists_pre_entry - %s (url: %s) - Performing internal search\n",
+                      slapi_entry_get_dn_const(list_entry), url);
+
+        /*
+        * Get our DN's groupOfUrls attribute
+        */
+        search_pb = slapi_pblock_new();
+        slapi_search_internal_set_pb(search_pb, ludp->lud_dn, ludp->lud_scope, ludp->lud_filter,
+                                     NULL, 0, NULL, NULL, dynamic_lists_get_plugin_id(), 0);
+        slapi_search_internal_pb(search_pb);
```
droideck, line 251: Is it possible to DoS here if a user crafts a very large search?

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+            for (size_t i = 0; entries && entries[i]; i++) {
+                if (list_attr == NULL) {
+                    /* No list attribute, so use the DN of the entry */
+                    const char *dn = slapi_entry_get_dn_const(entries[i]);
+                    slapi_valueset_add_value_ext(list_set, slapi_value_new_string(dn),
+                                                 SLAPI_VALUE_FLAG_PASSIN);
+                } else {
+                    /* Use an attribute/value from within the entry */
+                    const char *entry_list_attr = slapi_entry_attr_get_ref(entries[i],
+                                                                           list_attr ? list_attr : listConfig.attr);
```
droideck: I think we don't need the ternary operator `list_attr ? list_attr : listConfig.attr` here as we're already in the `else` block where `list_attr != NULL`.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
droideck, line 251: Also, it always runs an internal search as the plugin identity (effectively `cn=Directory Manager`?) and copies the first value of any attribute named in the LDAP URL straight into the entry being returned to the client. So anybody who can read a dynamic list entry can exfiltrate otherwise protected attributes (e.g. userPassword, homePhone) simply by pointing memberURL at them.
Do I understand the logic correctly or do I miss something?

droideck (commented):
I think, it'll be nice to have an additional `memberURL` validation in the UI.
It's possible to make a typo there and it'll be nice to have the feedback right away (instead of error log).

Also, maybe, we can "block" the dynamic attribute in the Edit wizard, so it's not possible to remove/edit it (it complains at the edit/delete operation via LDAP error, but may be we can mark the attribute as dynamic in the UI).
But it's just nice to have. Feel free to postpone or ignore the suggestion :)

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+        */
+        if ((rc = slapi_ldap_url_parse(url, &ludp, 0, &secure)) != 0) {
+            /* Failed to parse the memberUrl attribute */
+            slapi_log_err(SLAPI_LOG_ERR, DYNAMIC_LISTS_PLUGIN_SUBSYSTEM,
+                          "dynamic_lists_pre_entry - %s - failed to parse the LDAP url: %s\n",
+                          slapi_entry_get_dn_const(list_entry), url);
+            continue;
+        }
+
+        if (ludp->lud_filter == NULL) {
```
tbordaz, line 212: If the entry contains an invalid memberURL (it should not), could ludp be NULL ?.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+        if (ludp->lud_dn == NULL) {
+            slapi_log_err(SLAPI_LOG_ERR, DYNAMIC_LISTS_PLUGIN_SUBSYSTEM,
+                          "dynamic_lists_pre_entry - %s - Missing base dn in LDAP URL %s\n",
+                          slapi_entry_get_dn_const(list_entry), url);
+            continue;
+        }
+
+        if (ludp->lud_attrs != NULL && ludp->lud_attrs[0] != NULL) {
+            /* If an attribute is specified use it */
+            list_attr = ludp->lud_attrs[0];
```
tbordaz, line 228: This is something not clear to me. This plugin is not a deref plugin, it does not return the values of attributes from the entries in the URL but only the entries DN.
I think we should ignore those attributes and only take into account the one in the config.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        /*
+        * Check if the client has permission to read the list attribute
+        */
+        if (dynamic_lists_check_access(pb, list_entry,
```
tbordaz, line 234: I would expect the check_access on each entries found by the internal search below.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: > > Is it possible to have a multiple of `memberUrl`? I think a schema definition allows it. Also, it's multi-valued, then I think it'll be nice to have a test for that too.
> 
> Yeah it's multivalued, and we should support that. Also you can set what ever attribute you want - it could technically be "cn", etc. So yeah this adds some complexity, but it's not that big of a deal. I can see a use case for multiple urls, even though it's probably rare. Okayyyy I'll work on that... :-)

Fixed

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: We check access control on the dynamic attribute before doing the search

mreynolds389: > Is it possible to DoS here if a user crafts a very large search?

The search is not control by clients, but by whom ever set the memberUrl attribute.  We can't always stop an Admin from shooting themselves in the foot.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: > Also here, do we need to call `slapi_pblock_destroy(search_pb);` before continue?

Dang it, I meant to clean up the "continue" behavior in this loop when I added the multiple url support.  I'll fix this.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: > I think we might have a memory leak here, if the URL is invalid and we `continue` while not freeing the invalid `ludp`. Also, I think we need to set `ludp = NULL` in the beginning of the loop.

Yeah I just mentioned below that this loop needs to be cleaned up.  I'm on it.

ldap/servers/plugins/dynamic_lists/dynamic_lists.c
mreynolds389: > This is something not clear to me. This plugin is not a deref plugin, it does not return the values of attributes from the entries in the URL but only the entries DN. I think we should ignore those attributes and only take into account the one in the config.

If you define an attribute in the URL it DOES get that attribute value from the entry and not the DN.  Basically it overrides the dynamic list attribute in the config.  This is described in the design doc.

mreynolds389:
> I think that we may also want to add some filter rewriter (maybe in a second PR ?): There is a fair chance that the dynamic group will be used as a plain group (i.e having application doing search -s base -b groupdn member=some_user_dn to determine if an user is member of the group Such filter will be a performance killer if the group is large and it is better to directly check if some_user_dn entry matchs the memberurl

@progier indeed openldap does allow filtering on the dynamic content:

```
$ ldapsearch -xLLL -H ldap://localhost:1389 -D "cn=manager,dc=example,dc=com" -w PASSWORD -b "dc=example,dc=com" cn=demo_group
dn: cn=demo_group,ou=groups,dc=example,dc=com
objectClass: top
objectClass: groupOfURLs
cn: demo_group
memberURL: ldap:///dc=example,dc=com??sub?(uid=*)
member: uid=demo_user,ou=people,dc=example,dc=com

$ ldapsearch -xLLL -H ldap://localhost:1389 -D "cn=manager,dc=example,dc=com" -w PASSWORD -b "dc=example,dc=com" member=uid=demo_user,ou=people,dc=example,dc=com
dn: cn=demo_group,ou=groups,dc=example,dc=com
objectClass: top
objectClass: groupOfURLs
cn: demo_group
memberURL: ldap:///dc=example,dc=com??sub?(uid=*)
member: uid=demo_user,ou=people,dc=example,dc=com
```

I'm not sure how to efficiently pull this off as dynamic content is obviously not indexed and not present in id2entry (thinking of subtree search scenario).  I'll have to think about this...

mreynolds389:
> FYI: A idea about how to do it:
> 
> A place we can tweak the result is in the SLAPI_PLUGIN_POST_SEARCH_FN post op callback. We could add the dynamic entries in the search result set
> 
> Finding the entry to build should not be too complex:
> 
> if filter contains an attr=value subfilter with attr having a syntax dn Look for entry matching the dynamic list object class and the dynamic attribute url being present if value matches the attribute url value then add the dynamic attribute to the dynamic list entry), check that it still matches the original search filter and if yes add it to the search result set
> 
> This will not handle the very complex filters (for example those having substrings or ranges on the dynamic attribute) but it will handle the common case and should scale as far as there are not too many dynamic entries (For the very complex filter, I guess that the only way is to check all dynamic entries which is not very efficient ...)
> 
> Note: with the very simple filter (having only one attribute having dn syntax in the filter) we can even bypass rebuilding the whole attribute valueset but just add attr: value (to check the filter)

This is interesting and I will look into it tomorrow.  But one small snag is that we do dynamic "content" for other attributes that are not DN syntax.  We also need to process a filter that finds dynamic and static content, so we can not skip the candidate lookup.

We could still keep this PR as is (dynamic content), and then do only "dynamic groups" as a separate feature where we tweak the result set.  Ok well let me dig into this more tomorrow.  Thanks for the help and ideas!

PR #7101, opened 2025-11-18
Issue 6753 - Port ticket 548 test

mirielka:
Description:
Port ticket 548 test into
dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py

Relates: #6753
Author: Lenka Doudova
Assisted by: Cursor
Reviewer: ???

## Summary by Sourcery

Port ticket 548 shadowAccount tests into the password policy test suite and remove the standalone ticket548_test file

New Features:
- Add shadowUser fixture for creating a shadowAccount user
- Implement helper functions to convert days to seconds, configure/modify password policies, and validate shadowAccount attributes
- Add tests for shadowAccount attribute behavior under no policy, and global, subtree, and user-level policies

Enhancements:
- Skip shadowAccount tests on directory server versions older than 1.3.6

Chores:
- Remove obsolete ticket548_test.py


dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +363,397 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+    return days * 86400
+
+
+def check_shadow_attr_value(inst, user_dn, attr_type, expected, dn):
+    """ Check that shadowAccount attribute has expected value """
+    dm = DirectoryManager(inst)
+    dm.rebind()
+    user = UserAccount(inst, user_dn)
+    if user.present(attr_type):
+        actual = int(user.get_attr_val_utf8(attr_type))
```
droideck: You can use `get_attr_val_int` here

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +363,397 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+        if actual == expected:
+            log.info(f'{attr_type} of entry {dn} has expected value {actual}')
+            assert True
```
droideck: I'm not sure I understand a purpose for this line...

dirsrvtests/tests/tickets/ticket548_test.py
```diff
@@ -1,408 +0,0 @@
...
-                              'userpassword': USER_PW})))
-    except ldap.LDAPError as e:
-        log.fatal('test_ticket548: Failed to add user' + USER3_DN + ': error ' + e.message['desc'])
-        assert False
-
-    log.info('Search entry %s' % USER3_DN)
-    entry0 = topology_st.standalone.getEntry(USER3_DN, ldap.SCOPE_BASE, "(objectclass=*)")
-
-    log.info('Expecting shadowLastChange 0 since passwordMustChange is on')
-    check_shadow_attr_value(entry0, 'shadowLastChange', 0, USER3_DN)
```
droideck, line 339: Both subtree and user policy tests now expect shadowLastChange to equal today’s date immediately after the policy is created, even though both policies set passwordMustChange: 'on'.  Previously, it was explicitly asserted that shadowLastChange stays 0 until the user changes their password so we know that "must change" is enforced. 

I'm not that familiar with the area, could you please clarify if we changed it at some point? (as you changed the test). I wonder if it might be a very old regression if we asserted different things previously...

dirsrvtests/tests/tickets/ticket548_test.py
```diff
@@ -1,408 +0,0 @@
...
-    check_shadow_attr_value(entry0, 'shadowWarning', 6, USER3_DN)
-
-    log.info("Bind as %s" % USER3_DN)
-    topology_st.standalone.simple_bind_s(USER3_DN, USER_PW)
-
-    log.info('Search entry %s' % USER3_DN)
-    try:
-        entry1 = topology_st.standalone.getEntry(USER3_DN, ldap.SCOPE_BASE, "(objectclass=*)")
-    except ldap.UNWILLING_TO_PERFORM:
-        log.info('test_ticket548: Search by' + USER3_DN + ' failed by UNWILLING_TO_PERFORM as expected')
```
droideck, line 357: The original test verified that when passwordMustChange is on, searching as the user (who must change password) fails. I think it's missing too... Or do we fail now at all? (similar concern to my previous passwordMustChange point)

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +363,397 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+    edate = int(time.time() / (60 * 60 * 24))
+
+    log.info(f"Bind as {shadowUser.dn}")
+    shadowUser.bind('password')
+    check_shadow_attr_value(topology_st.standalone, shadowUser.dn,
+                            'shadowLastChange', edate, shadowUser.dn)
+
+
+@pytest.mark.skipif(ds_is_older('1.3.6'), reason="Not implemented")
+def test_shadowaccount_global_policy(topology_st, shadowUser):
```
droideck: The original test_ticket548_test_global_policy tested shadow attributes on two different users (USER1 and USER2) to verify global policy applies correctly to multiple entries. The ported version only uses one user.

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +363,397 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+                            'shadowMax', 10, shadowUser.dn)
+    check_shadow_attr_value(topology_st.standalone, shadowUser.dn,
+                            'shadowWarning', 3, shadowUser.dn)
+
+    log.info("Change global password policy")
+    modify_pwp(topology_st.standalone, pwp_mgr, 'global')
+
+    log.info("Change shadowUser password")
+    shadowUser.bind('password')
+    shadowUser.replace('userpassword', 'password2')
```
droideck: Do you want to change the password as a `shadowUser` here? Because currently it changes the password as Directory Manager.
It happens because `Account.bind()` returns a brand-new connection clone; it does not change the binding on topology_st.standalone.

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +363,397 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+            assert True
+        else:
+            log.fatal(f'{attr_type} {actual} of entry {dn} does not have expected value {expected}')
+            assert False
+    else:
+        log.fatal(f'entry {dn} does not have {attr_type} attr')
+        assert False
+
+
+def setup_pwp(inst, pwp_mgr, policy, dn=None, policy_props=None):
```
droideck, line 408: After calling `setup_pwp(..., 'global')` and `modify_pwp(..., 'global')`, the test finishes without restoring the configuration, yet pwdPolicy_attribute_test.py contains many other password-policy tests that assume the default global settings. I wonder if it might affect the results.

droideck (changes requested):

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -9,12 +9,13 @@
 import pytest
 from lib389.tasks import *
 from lib389.utils import *
-import pdb
 from lib389.topologies import topology_st
 from lib389.pwpolicy import PwPolicyManager
 from lib389.idm.user import UserAccount, UserAccounts, TEST_USER_PROPERTIES
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389._constants import (DEFAULT_SUFFIX, DN_DM, PASSWORD)
+from lib389.pwpolicy import PwPolicyManager
```
droideck: It's a duplicate import

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +362,463 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+
+    return shadowUser
+
+
+def days_to_secs(days):
+    """ Convert days to seconds """
+    return days * 86400
+
+
+def check_shadow_attr_value(inst, user_dn, attr_type, expected, dn):
```
droideck: The function takes both `user_dn` and `dn`, but they're always the same in every call...
I think it can just be a user_dn.

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -361,6 +362,463 @@ def test_pwdpolicysubentry(topology_st, password_policy):
...
+                            'shadowWarning', 9, shadowUser2.dn)
+
+    log.info('Clean up - delete second user and reset global policy')
+    dm = DirectoryManager(inst)
+    dm.rebind()
+    shadowUser2.delete()
+    inst.config.replace('passwordMinAge', '0')
+    inst.config.replace('passwordMaxAge', '8640000')
+    inst.config.replace('passwordWarning', '86400')
+    inst.config.replace('passwordExp', 'off')
```
droideck: Might be worth to put the clean up into a finalizer

droideck (approved):
A few minor issues, otherwise, looks good!

droideck (approved):

Issue #7007, opened 2025-09-18
Fix paged result search locking


mreynolds389:
**Issue Description**

We need to hold the connection's main paged result "list" lock while attempting to lock an individual paged result search lock from that list.  Otherwise it's possible the paged result search lock could be removed while accessing it

PR #7104, opened 2025-11-18
Issue 7007 - Improve paged result search locking

mreynolds389:
Description:

Hold the paged result connection hash mutex while acquiring the global connection paged result lock. Otherwise there is a window where the mutex could be removed and lead to a crash

Relates: https://github.com/389ds/389-ds-base/issues/7007

### DIFF
--- a/ldap/servers/slapd/pagedresults.c
+++ b/ldap/servers/slapd/pagedresults.c
@@ -804,6 +804,7 @@ pagedresults_cleanup(Connection *conn, int needlock)
         prp->pr_current_be = NULL;
         if (prp->pr_mutex) {
             PR_DestroyLock(prp->pr_mutex);
+            prp->pr_mutex = NULL;
         }
         memset(prp, '\0', sizeof(PagedResults));
     }
@@ -841,6 +842,7 @@ pagedresults_cleanup_all(Connection *conn, int needlock)
         prp = conn->c_pagedresults.prl_list + i;
         if (prp->pr_mutex) {
             PR_DestroyLock(prp->pr_mutex);
+            prp->pr_mutex = NULL;
         }
         if (prp->pr_current_be && prp->pr_search_result_set &&
             prp->pr_current_be->be_search_results_release) {
@@ -1022,11 +1024,10 @@ pagedresults_lock(Connection *conn, int index)
     }
     pthread_mutex_lock(pageresult_lock_get_addr(conn));
     prp = conn->c_pagedresults.prl_list + index;
-    pthread_mutex_unlock(pageresult_lock_get_addr(conn));
     if (prp->pr_mutex) {
         PR_Lock(prp->pr_mutex);
     }
-    return;
+    pthread_mutex_unlock(pageresult_lock_get_addr(conn));
 }
 
 void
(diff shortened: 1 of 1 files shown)


tbordaz (approved):
Look good to me.
I think it worth a comment in slap.h stating that the all 'struct _paged_results'  (including pr_mutex) is protected by pageresult_lock_get_addr(conn) lock.

droideck (commented):
Is it possible that by the `pagedresults_lock/pagedresults_unlock` change of this PR, we introduce a deadlock described here: https://github.com/389ds/389-ds-base/pull/6783 ?
I think we might...

mreynolds389:
> Is it possible that by the `pagedresults_lock/pagedresults_unlock` change of this PR, we introduce a deadlock described here: #6783 ? I think we might...

Pierre and I looked at this closely when we did the first internal review, and it should be fine.  This fix has also been running in a customer environment with heavy paged result search activity for almost two months now.  Note all these paged result PR's (including 6783) are all for the same customer who was originally experiencing a crash.

Anyway here are Pierre's comments:

```
Looks good to me too.

I was also worried about deadlock
    (The fact that the conn mutex is a hash) make things more complex
     (there could be deadlock risk even if lock hierarchy is respected:
scenario about 
        Thread 1
           Acquired pr_mutex lock
           Try to acquire conn hash (to unlock pr_mutex)
    Thread 2 
            Acquired same conn hash on a second conn
            Try to lock pr_mutex lock

But we are safe because the page result is associated to a connection so different connection will access  different pr_mutex ==> no deadlock
```

droideck:
Hmm, just to clarify if I understand the pagedresults logic correctly:

Paged results belong to a connection, but is it jsut narrows the scope? (so maybe it's very hard to hit the deadlock but not impossible).

If the client sends the next page before the previous one finishes (or retransmits a request), two worker threads easily end up manipulating the same slot at once (because, I think, ops are asynchronous, right?), which can trigger the deadlock situation when pagedresults_lock() now holds the hashed lock while waiting on pr_mutex.

tbordaz:
Hi @droideck this looks a very valid concern.
My understanding is that there is a risk of deadlock is we acquire locks pageresult_lock_get_addr/pr_mutex in the opposite order. I checked the code and IIUC it always acquires pageresult_lock_get_addr before lock/free pr_mutex.
So IMHO the only scenario of hang is if a thread acquires pr_mutex (pagedresults_lock) and do not release it (pagedresults_unlock). This patch looks to me independent of such scenario.

tbordaz:
My understanding is that pageresult_lock_get_addr protects PagedResultsList and pr_mutex protects each individual PagedResults in the list. So before locking pr_mutex we should hold pageresult_lock_get_addr to not get the PagedResults modified under us. Once we hold PagedResults it is safe to release pageresult_lock_get_addr (PagedResultsList).

It is not clear to me if we can have several threads (async) accessing the same PagedResults. If it occurs I agree there is a risk of deadlock (not because of acquisition in opposite order)

- T1 acquires pageresult_lock_get_addr to acquire pr_mutex
- T1 acquires pr_mutex
- T1 release pageresult_lock_get_addr because it holds pr_mutex
- T2 acquires pageresult_lock_get_addr to acquire pr_mutex
- T2 wait for pr_mutex hold by T1
- T1 wait for pageresult_lock_get_addr  (to release pr_mutex) hold by T2

My feeling is that if we have several threads (on the same connection) accessing the same PagedResults, we need to revisit the locking strategy.
The initial deadlock is quite rare and I guess it is acceptable to wait for a fully approved fix

progier389 (approved):
We should fix the crash with this PR
Then plan to rework the page result lock strategy

Issue #7110, opened 2025-11-20
Problem with editing group entries in the LDAP browser of the Directory Server console


aadhikar:
**Issue Description**
Problem with editing group entries in the LDAP browser of the Directory Server console.
When editing members, the text and check boxes overlap, making it impossible to click the check boxes.

**Package Version and Platform:**
 - Platform: Fedora 42
 - Package and version: cockpit-389-ds-3.1.3-3.fc42.noarch 389-ds-base-3.1.3-3.fc42.x86_64
 - Browser: chrome, firefox

**Steps to Reproduce**
1. Log in to Cockpit
2. Select Directory Server
3. Select the LDAP Browser tab
4. Select some group entry from the tree
5. Use "Edit..." to open the entry editor

**Actual results**
The text and checkboxes overlap, making it impossible to click the checkboxes.

**Screenshots**

<img width="1176" height="633" alt="Image" src="https://github.com/user-attachments/assets/02fa8751-2980-4b6a-86e5-6c2fd3b05294" />

**Expected results**
The text and checkboxes do not overlap, allowing you to click the checkboxes.

aadhikar:
It's a duplicate issue.

PR #7111, opened 2025-11-20
Issue 7041 - Add WebUI test for group member management

aadhikar:
Fixed checkbox overlap in LDAP browser group member tables by setting fixed column widths and proper text wrapping constraints.

Added WebUI test for group member add/remove via checkboxes (single and multiple deletions).

Relates: https://github.com/389ds/389-ds-base/issues/7110

Reviewed by: ???

## Summary by Sourcery

Fix checkbox overlap in LDAP browser group member editor by enforcing fixed column widths, wrapping exceptions, and improving checkbox styling, and add a WebUI test to validate group member removal via checkboxes

Bug Fixes:
- Fix checkbox overlap in LDAP browser group member tables by setting fixed widths, wrapping constraints, and adjusting checkbox styling

Tests:
- Add WebUI test for group member management to verify single and multiple member deletion via checkboxes

### DIFF
--- a/dirsrvtests/tests/suites/webui/ldap_browser/ldap_browser_test.py
+++ b/dirsrvtests/tests/suites/webui/ldap_browser/ldap_browser_test.py
@@ -14,6 +14,8 @@ from lib389.cli_idm.account import *
 from lib389.tasks import *
 from lib389.utils import *
 from lib389.topologies import topology_st
+from lib389.idm.user import UserAccounts
+from lib389.idm.group import Groups
 from .. import setup_page, check_frame_assignment, setup_login, create_entry, delete_entry, load_ldap_browser_tab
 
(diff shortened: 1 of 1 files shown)


droideck (changes requested):
I don't like an idea where we directly overriding PatternFly internal classes.
Especially, if done with !important - it can be fragile and break on PatternFly upgrades.

Did you try to use built-in modifiers (i.e. `<Th width={10}>` and  `<Th modifier="wrap">`)?
If it won't work, then maybe we can define a custom .ds-group-member-table with 
https://www.patternfly.org/developer-resources/global-css-variables/

mreynolds389:
> I don't like an idea where we directly overriding PatternFly internal classes. Especially, if done with !important - it can be fragile and break on PatternFly upgrades.
> 

I had the same concern, but actually this was fixed by one of my earlier commits.  I already informed @aadhikar and he's going to remove the css changes and just keep the test

droideck (approved):
LGTM!

PR #7120, opened 2025-11-27
Issue 7119 - Harden DNA plugin locking for shared server list operations

### LINKED ISSUE #7119: Harden DNA plugin locking for shared server list operations
**Description**
A potential race condition exists in the DNA plugin's management of the global shared server list (dna_global_servers). In `dna_load_shared_servers()` the write lock was acquired and released inside the loop for each configuration entry, creating windows where concurrent config updates or post-operation hooks could access a partially rebuilt list.
And another, smaller issue, `dna_close()` called `dna_delete_global_servers()` without holding any lock, allowing teardown to race with active readers or rebuilds (so we might hit an issue on 'disable plugin' op, IIUC)

While I was unable to reproduce the exact failure scenario, a crash was reported in production FreeIPA environments where `dna_free_shared_server()` received a corrupted pointer (happened during startup.) This hardening addresses the theoretical race conditions that could lead to such memory corruption in rare concurrent scenarios.

droideck:
Description: Hold dna_server_write_lock() across both the global list teardown and full rebuild in dna_load_shared_servers(), eliminating rare races where concurrent config updates could free or append entries while the list is being reconstructed.

Guard dna_delete_global_servers() at shutdown behind the same lock, ensuring teardown doesn't run in parallel with active readers.

Add comprehensive multi-supplier DNA test suite covering basic allocation, uniqueness, shared config replication, restart recovery, and range exhaustion scenarios.

Fixes: https://github.com/389ds/389-ds-base/issues/7119

Reviewed by: ?

## Summary by Sourcery

Harden DNA plugin server list locking and add regression coverage for multi-supplier and range-exhaustion scenarios.

Bug Fixes:
- Protect global DNA shared server list teardown and rebuild with the server write lock to prevent races with concurrent configuration updates and shutdown.
- Ensure dna_delete_global_servers() is only called while holding the appropriate write lock, including during plugin shutdown.

Tests:
- Add an automated multi-supplier DNA test suite validating basic allocation ranges, uniqueness across suppliers, shared configuration replication, post-restart allocation behavior, and small-range exhaustion/transfer semantics.

### DIFF
--- a/ldap/servers/plugins/dna/dna.c
+++ b/ldap/servers/plugins/dna/dna.c
@@ -696,7 +696,13 @@ dna_close(Slapi_PBlock *pb __attribute__((unused)))
     slapi_destroy_rwlock(g_dna_cache_lock);
     g_dna_cache_lock = NULL;
 
-    dna_delete_global_servers();
+    if (g_dna_cache_server_lock) {
+        dna_server_write_lock();
+        dna_delete_global_servers();
+        dna_server_unlock();
+    } else {
+        dna_delete_global_servers();
+    }
     slapi_destroy_rwlock(g_dna_cache_server_lock);
     g_dna_cache_server_lock = NULL;
 
@@ -769,7 +775,8 @@ out:
 }
 
 /*
- * Free the global linkedl ist of shared servers
+ * Free the global linked list of shared servers
+ * Caller must hold dna_server_write_lock() before invoking this helper.
  */
 static void
 dna_delete_global_servers(void)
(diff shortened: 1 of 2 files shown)


ldap/servers/plugins/dna/dna.c
```diff
@@ -812,34 +823,31 @@ dna_load_shared_servers(void)
...
-                dna_delete_global_servers();
-                freed_servers = 1;
-            }
             if (shared_list) {
                 server_list = PR_LIST_HEAD(shared_list);
                 while (server_list != shared_list) {
                     server = (struct dnaServer *)server_list;
-                    if (global_servers == NULL) {
-                        dna_global_servers = global_servers = server;
+                    if (dna_global_servers == NULL) {
```
tbordaz, line 835: Your changes looks good to me but to be honest I do not undersand the purpose of this loop !
It looks to me that the only result is to set dna_global_servers = PR_LIST_HEAD(shared_list).

tbordaz (approved):
It looks good to me.

ldap/servers/plugins/dna/dna.c
droideck: From what I understood, `dna_global_servers` serves a different purpose than the PRCList `shared_list`.

This loop in dna_load_shared_servers() is doing more than just the final assignment - it walks the circular PRCList that dna_get_shared_servers() builds and converts it into linear `dna_global_servers` chain while preserving the "most remaining IDs first" order that the PRC list encodes.

The `PRCList` is a temporary thing that makes it easy to keep the per-range results ordered by "remaining" values. While building the list in dna_get_shared_servers() we insert each candidate before or after existing nodes via `PR_INSERT_*`, which is much simpler with the circular/sentinel API than with a manual singly linked list (no special cases for head/tail, no extra bookkeeping).

And `dna_global_servers` (straight next chain) is the long-lived cache the rest of the plugin expects. All consumers (e.g. dna_get_shared_config_attr_val, range-request fallback) walk a simple server->next chain; they don’t know about sentinel nodes or NSPR macros.

tbordaz (approved):
LGTM as well

PR #7127, opened 2025-12-02
Issue 7126 - WARN - keys2idl - received NULL idl from index_read_ext_allids

### LINKED ISSUE #7126: WARN - keys2idl - received NULL idl from index_read_ext_allids
**Issue Description**
During testing https://github.com/389ds/389-ds-base/issues/7124 I often see these errors:
```
[29/Nov/2025:22:59:13.930441639 +0000] - WARN - keys2idl - received NULL idl from index_read_ext_allids, treating as empty set
[29/Nov/2025:22:59:13.936543122 +0000] - WARN - keys2idl - this is probably a bug that should be reported
[29/Nov/2025:22:59:13.947131944 +0000] - ERR - build_candidate_list - Database error -12795
```

 - Platform: Fedora, RHEL
 - Package and version: all versions with BDB backend

**Steps to Reproduce**
1. Reproducer from https://github.com/389ds/389-ds-base/issues/7124

**Expected results**
No keys2idl warnings

vashirov:
Bug Description:
Under high concurrency, occasional NULL IDL warnings were logged:
```
[29/Nov/2025:22:59:13.930441639 +0000] - WARN - keys2idl - received NULL idl from index_read_ext_allids, treating as empty set 
[29/Nov/2025:22:59:13.936543122 +0000] - WARN - keys2idl - this is probably a bug that should be reported 
[29/Nov/2025:22:59:13.947131944 +0000] - ERR - build_candidate_list - Database error -12795
```
Fix Description:
Ensure idl returned by `index_read_ext_allids()` is never NULL.

Fixes: https://github.com/389ds/389-ds-base/issues/7126

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -163,7 +163,7 @@ typedef unsigned short u_int16_t;
 #define SUBLEN 3
 #define LDBM_CACHE_RETRY_COUNT 1000        /* Number of times we re-try a cache operation */
 #define RETRY_CACHE_LOCK       2           /* error code to signal a retry of the cache lock */
-#define IDL_FETCH_RETRY_COUNT  5           /* Number of times we re-try idl_fetch if it returns deadlock */
+#define IDL_FETCH_RETRY_COUNT  10          /* Number of times we re-try idl_fetch if it returns deadlock */
 #define IMPORT_SUBCOUNT_HASHTABLE_SIZE 500 /* Number of buckets in hash used to accumulate subcount for broody parents */
 
 /* minimum max ids that a single index entry can map to in ldbm */
--- a/ldap/servers/slapd/back-ldbm/index.c
+++ b/ldap/servers/slapd/back-ldbm/index.c
@@ -990,7 +990,7 @@ index_read_ext_allids(
     prefix = index_index2prefix(indextype);
     if (prefix == NULL) {
         slapi_log_err(SLAPI_LOG_ERR, "index_read_ext_allids", "NULL prefix\n");
-        return NULL;
+        return idl_alloc(0);
     }
     if (slapi_is_loglevel_set(LDAP_DEBUG_TRACE)) {
         slapi_log_err(SLAPI_LOG_TRACE, "index_read_ext_allids", "=> ( \"%s\" %s \"%s\" )\n",
@@ -1006,7 +1006,7 @@ index_read_ext_allids(
     if (ai == NULL) {
         index_free_prefix(prefix);
         slapi_ch_free_string(&basetmp);
-        return NULL;
+        return idl_alloc(0);
     }
 
     slapi_log_err(SLAPI_LOG_ARGS, "index_read_ext_allids", "indextype: \"%s\" indexmask: 0x%x\n",
@@ -1025,7 +1025,7 @@ index_read_ext_allids(
         slapi_ch_free_string(&basetmp);
         if (NULL == val || NULL == val->bv_val) {
             /* entrydn value was not given */
-            return NULL;
+            return idl_alloc(0);
         }
         slapi_sdn_init_dn_byval(&sdn, val->bv_val);
         rc = entryrdn_index_read(be, &sdn, &id, txn);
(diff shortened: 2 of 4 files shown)


ldap/servers/slapd/back-ldbm/index.c
```diff
@@ -1110,6 +1114,14 @@ index_read_ext_allids(
         ber_bvfree(encrypted_val);
     }
 
+    /* Ensure we never return NULL - always return valid IDL */
+    if (idl == NULL) {
+        slapi_log_err(SLAPI_LOG_WARNING, "index_read_ext_allids",
+                      "Returning empty IDL for %s after error %d\n",
+                      basetype, *err);
```
droideck, line 1169: I'm worried that we might have a use-after-free in `basetype` here because on 941 line:

```
    if ((basetmp = slapi_attr_basetype(type, typebuf, sizeof(typebuf))) != NULL) {
        basetype = basetmp;
    }
 ```
 
 and on 1103, before the code, we:
 
 ```
     slapi_ch_free_string(&basetmp);
 ```

**Review by droideck — COMMENTED** [review]
Also, there are a few more `return NULL` exists beside the main code path.
I don't have a strong opinion here though... So if you checked it under a high contention and the errors are gone, - I'm okay with that
Yet, it'd be nice to clean up the logic a bit as some of the function's callers don't do any NULL checks.

**Review by tbordaz — APPROVED** [review]
Agree with @droideck basetmp comments.
Except that LGTM

File: ldap/servers/slapd/back-ldbm/index.c (hunk 1 above)
Comment (vashirov) [reply ldap/servers/slapd/back-ldbm/index.c:1169]: Nice catch, moved `slapi_ch_free_string(&basetmp);`.

**Review by droideck — APPROVED** [review]
LGTM!

**Review by droideck — COMMENTED** [review]
Just a note, the new commit message has a typo - `idl_alloc(1)` instead of `idl_alloc(0)`.

Also a dead code in https://github.com/389ds/389-ds-base/blob/8fe7bfe6ba2a21bcee34c6a383f080b02ecc4683/ldap/servers/slapd/back-ldbm/filterindex.c#L1160-L1164

**Comment by droideck** [comment]
> > Also, there are a few more `return NULL` exists beside the main code path. I don't have a strong opinion here though... So if you checked it under a high contention and the errors are gone, - I'm okay with that Yet, it'd be nice to clean up the logic a bit as some of the function's callers don't do any NULL checks.
> 
> I agree, `index_read_ext_allids` has `IDList` type, so we should always return a valid IDL. I cleaned up other early exits to return empty IDL instead of NULL.
> 
> There are some checks for NULL IDL in
> 
> https://github.com/389ds/389-ds-base/blob/8fe7bfe6ba2a21bcee34c6a383f080b02ecc4683/ldap/servers/slapd/back-ldbm/db-mdb/mdb_ldif2db.c#L384
> 
> and
> 
> https://github.com/389ds/389-ds-base/blob/8fe7bfe6ba2a21bcee34c6a383f080b02ecc4683/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c#L493
> 
> Since we changed the return type, I think we should adjust these as well, correct?

I'm not fully sure...
I tend to lean toward the option to check like this:

```c
if (*err == MDB_NOTFOUND) {

} else if (*err != 0) {

} else if (idl == NULL) {

```

IIUC, an empty ID list is perfectly valid for these cases and we can treat an empty-but-valid IDL (zero count, *err == 0) as a normal result.

@progier389 @tbordaz WDYT?

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -161,7 +161,7 @@ typedef unsigned short u_int16_t;
 #define SUBLEN 3
 #define LDBM_CACHE_RETRY_COUNT 1000        /* Number of times we re-try a cache operation */
 #define RETRY_CACHE_LOCK       2           /* error code to signal a retry of the cache lock */
-#define IDL_FETCH_RETRY_COUNT  5           /* Number of times we re-try idl_fetch if it returns deadlock */
+#define IDL_FETCH_RETRY_COUNT  20          /* Number of times we re-try idl_fetch if it returns deadlock */
```
progier389: Isn't IDL_FETCH_RETRY_COUNT too high ?
   Now the operation is aborted after about 3.5 seconds while it was after 0.5 second originally.
IDL_FETCH_RETRY_COUNT 10 seems more reasonable: (around 1.5 second)

vashirov: Yes, it might be too high. I set it during testing, wasn't able to reach more than 7 retries.

progier389 (approved):
LGTM

Issue #7132, opened 2025-12-03
Keep alive entry updated too soon after an offline import


progier389:
**Issue Description**
Keep alive is updated 30 seconds after starting the server. 
This is too fast: 
  maximum backoff timer value is by default 5 minutes so it is possible that 
   the the keep alive entry get updated before the other replica have replayed the local 
   change toward the just reinitialized server.
  If that is the case there are missing CSN error and some changes are never replayed toward the 
   reinitialized replica.

**Package Version and Platform:**
 - Platform: [e.g. Fedora] RHEL, Fedora
 - Package and version: All versions are impacted

**Steps to Reproduce**
1. Stop all the replicas
2. Reinitialize first replica from an older ldid  
3. Wait 40 seconds
4. Start the other replicas 
5. perform change on one of the other replica
6. wait until change is replicated everywhere
7. Check for missing updates on first replica

**Additional context**
https://issues.redhat.com/browse/RHEL-129675

progier389:
49ed4ad368..8fe7bfe6ba  main -> main
   295ed6bca4..833103fbfa  389-ds-base-3.0 -> 389-ds-base-3.0
   da5aaeef6..9e48fb684  389-ds-base-2.7 -> 389-ds-base-2.7
   b96eef05e1..bd26504808  389-ds-base-2.6 -> 389-ds-base-2.6
   d9ba753e7..7473ac492  389-ds-base-2.5 -> 389-ds-base-2.5

PR #7133, opened 2025-12-03
Issue 7132 - Keep alive entry updated too soon after an offline import

progier389:
Problem: first keep alive update is done 30 seconds after restarting the server which may be before the other replica have the time to replicate local changes after a re-initialization.
Solution:  unify the timer management so that it starts after the keep alive interval (which is configurable)  in the 3 following cases:
 - server starts 
 - replica is enabled (i.e after bulk import)
 - keep alive interval is changed (to avoid having to restart the server after configuration change)
 
 Also logs a warning if the keep alive interval is smaller than the maximum backoff  timer value

Issue: #7132 

Reviewed by: @tbordaz , @droideck  (Thanks!)

## Summary by Sourcery

Adjust replication keep-alive scheduling to avoid premature updates after offline imports and ensure timers honor the configured interval across key lifecycle events.

New Features:
- Add a regression test covering offline import scenarios to verify keep-alive updates do not occur before replicas have time to resynchronize.

Bug Fixes:
- Ensure replica keep-alive updates are scheduled based on the configured interval rather than a hard-coded delay, and are correctly reset when replicas are created, enabled, or reconfigured.
- Prevent premature keep-alive updates after bulk imports or restores that could interfere with replica resynchronization.

Enhancements:
- Centralize keep-alive timer management into a shared helper for consistent behavior across replica lifecycle operations.
- Log a warning when the configured keep-alive interval is smaller than the maximum replication backoff timeout to highlight unsafe configurations.

Tests:
- Extend the replication regression test suite with a multi-supplier scenario that validates correct behavior of keep-alive updates following offline export/import and topology changes.

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m3_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m3_test.py
@@ -19,7 +19,6 @@ from lib389.dseldif import *
 from lib389.topologies import topology_m3 as topo_m3
 from lib389.agreement import Agreements
 from lib389.replica import ReplicationManager, Replicas
-from lib389.cli_ctl.dblib import DbscanHelper
 
 
 pytestmark = pytest.mark.tier1
@@ -191,7 +190,7 @@ def perform_updates(inst, replicas):
     repl = ReplicationManager(DEFAULT_SUFFIX)
     dn = 'uid=demo_user,ou=people,dc=example,dc=com'
     user = UserAccount(inst, dn)
-    for i in range(4):
+    for _ in range(4):
         val = f'Test {count}'
         user.add('description', val)
         count += 1
@@ -209,7 +208,7 @@ def check_updates(inst):
         assert f'Test {i}'.encode() in vals
 
 
-def test_offline_import(topo_m3):
+def test_offline_import(topo_m3, request):
     """Test that keep alive is not updated too soon after an offline import
 
     :id: 493a4810-cc5b-11f0-8c70-c85309d5c3e3
@@ -254,15 +253,21 @@ def test_offline_import(topo_m3):
     S2 = topo_m3.ms["supplier2"]
     S3 = topo_m3.ms["supplier3"]
 
+    def fin():
+        get_agmt(S1, "003").resume()
+        get_agmt(S2, "001").resume()
+        get_agmt(S3, "002").resume()
+
+    request.addfinalizer(fin)
     r1 = Replicas(S1).get(DEFAULT_SUFFIX)
     # 60 is the smallest accepted value for the interval time
     r1.replace('nsds5ReplicaKeepAliveUpdateInterval', '60')
     # Need to restart to reset repeated timer
     S1.restart()
 
-    get_agmt(S1, "003").delete()
-    get_agmt(S2, "001").delete()
-    get_agmt(S3, "002").delete()
+    get_agmt(S1, "003").pause()
+    get_agmt(S2, "001").pause()
+    get_agmt(S3, "002").pause()
     perform_updates(S1, (S2, S3))
     S3.stop()
     ldif_file = '%s/supplier3.ldif' % S3.get_ldif_dir()
--- a/dirsrvtests/tests/suites/replication/regression_m3_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m3_test.py
@@ -11,12 +11,15 @@ import time
 import logging
 import ldap
 import pytest
-from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts
+from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts, UserAccount
 from lib389.utils import *
 from lib389._constants import *
 from lib389.replica import Changelog5
 from lib389.dseldif import *
 from lib389.topologies import topology_m3 as topo_m3
+from lib389.agreement import Agreements
+from lib389.replica import ReplicationManager, Replicas
+from lib389.cli_ctl.dblib import DbscanHelper
 
 
 pytestmark = pytest.mark.tier1
@@ -37,6 +40,7 @@ else:
     logging.getLogger(__name__).setLevel(logging.INFO)
 log = logging.getLogger(__name__)
 
+count = 0
 
 def test_cleanallruv_repl(topo_m3):
     """Test that cleanallruv could not break replication if anchor csn in ruv originated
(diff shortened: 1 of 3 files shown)


ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -146,6 +146,28 @@ replica_new(const Slapi_DN *root)
     return r;
 }
 
+void
+reset_keepalive_timer(Replica *r)
+{
+    if (r->repl_eqcxt_ka_update != NULL) {
```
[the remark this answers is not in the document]
progier389: The mutex is acquired by reset_keepalive_timer callers (with the excption of replica_new_from_entry (but it is not a problem  as the replica is being created and not yet available at this time ==> no possible race condition)

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -146,6 +146,28 @@ replica_new(const Slapi_DN *root)
...
+    }
+    if (replica_get_type(r) == REPLICA_TYPE_UPDATABLE) {
+        int64_t interval = replica_get_keepalive_update_interval(r);
+        int64_t bomax = slapi_counter_get_value(r->backoff_max);
+        if (interval <= bomax) {
+            slapi_log_err(SLAPI_LOG_WARNING, repl_plugin_name, "Replica %s "
+                          "should have a keep alive interval greater than the "
+                          "maximum backoff timeout\n", r->repl_name);
+        }
+        r->repl_eqcxt_ka_update = slapi_eq_repeat_rel(replica_subentry_update, r,
```
tbordaz, line 164: An option is to accept interval < bomax only for the period. For the 'when' (first time) we can force max(interval,bomax).
Here it is the time of the first update that trigger the issue.

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -146,6 +146,28 @@ replica_new(const Slapi_DN *root)
     return r;
 }
 
+void
```
tbordaz: could be a static function

tbordaz (commented):
Looks good to me. A minor question.

dirsrvtests/tests/suites/replication/regression_m3_test.py
```diff
@@ -11,12 +11,15 @@
...
-from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts
+from lib389.idm.user import TEST_USER_PROPERTIES, UserAccounts, UserAccount
 from lib389.utils import *
 from lib389._constants import *
 from lib389.replica import Changelog5
 from lib389.dseldif import *
 from lib389.topologies import topology_m3 as topo_m3
+from lib389.agreement import Agreements
+from lib389.replica import ReplicationManager, Replicas
+from lib389.cli_ctl.dblib import DbscanHelper
```
droideck: It's not used, AFAICS

dirsrvtests/tests/suites/replication/regression_m3_test.py
```diff
@@ -165,6 +169,119 @@ def test_cleanallruv_repl(topo_m3):
...
+
+
+def test_offline_import(topo_m3):
+    """Test that keep alive is not updated too soon after an offline import
+
+    :id: 493a4810-cc5b-11f0-8c70-c85309d5c3e3
+    :setup: 3 Suppliers
+    :steps:
+        1. Configure Topology in single loop S1 --> S2 --> S3 --> S1
+        2. Set keep alive interrval to 60 seconds
```
droideck, line 218:
```suggestion
        2. Set keep alive interval to 60 seconds
```

dirsrvtests/tests/suites/replication/regression_m3_test.py
```diff
@@ -165,6 +169,119 @@ def test_cleanallruv_repl(topo_m3):
...
+
+    r1 = Replicas(S1).get(DEFAULT_SUFFIX)
+    # 60 is the smallest accepted value for the interval time
+    r1.replace('nsds5ReplicaKeepAliveUpdateInterval', '60')
+    # Need to restart to reset repeated timer
+    S1.restart()
+
+    get_agmt(S1, "003").delete()
+    get_agmt(S2, "001").delete()
+    get_agmt(S3, "002").delete()
```
droideck: I'm a bit afraid it could break if we ever change the name. I think, it'll be better to use another way (`get_agmt` can get it by port, may be? the one you extract with `S3.port`, `S1.port`, etc)

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -3743,6 +3756,7 @@ replica_set_keepalive_update_interval(Replica *r, int64_t interval)
 {
     replica_lock(r->repl_lock);
     r->keepalive_update_interval = interval;
+    reset_keepalive_timer(r);
```
droideck, line 3759: Please, fix me if the logic if wrong somewhere and I missed somthing:

`replica_set_keepalive_update_interval()` grabs `r->repl_lock`, updates the interval, and immediately calls `reset_keepalive_timer()` while still holding that lock.

but then, in `reset_keepalive_timer()` there's a `replica_subentry_update` callback, which has `replica_get_rid()` which has `repl_lock`.

Does it mean we can deadlock when changing `nsds5ReplicaKeepAliveUpdateInterval`? (in a very rare cases)

dirsrvtests/tests/suites/replication/regression_m3_test.py
```diff
@@ -165,6 +169,119 @@ def test_cleanallruv_repl(topo_m3):
...
+def check_updates(inst):
+    global count
+    dn = 'uid=demo_user,ou=people,dc=example,dc=com'
+    user = UserAccount(inst, dn)
+    vals = user.get_attr_vals('description')
+    for i in range(count):
+        assert f'Test {i}'.encode() in vals
+
+
+def test_offline_import(topo_m3):
```
droideck: After this test, topology is unusable by other tests. I think we need either indicate it, or use an isolated topology (possible with a `function` scope here).

droideck (changes requested):
Mostly minor issues:) And a question/concern.

dirsrvtests/tests/suites/replication/regression_m3_test.py
progier389: Nice catch! 
  I should rather disable the agreement and have a finaliser that restore them

ldap/servers/plugins/replication/repl5_replica.c
progier389: No:  slapi_eq_repeat callback is not called sequentially in the current thread 
It is called in another thread when the timer expires

ldap/servers/plugins/replication/repl5_replica.c
progier389: I would rather let the admin knows that there is a configuration issue rather than silently fix it 
especially since interval == bomax may still be an issue as it may not left enough time to replay the changes

tbordaz: okay thanks

tbordaz (approved):
LGTM (server part)

droideck (approved):
Looks good! Ack

dirsrvtests/tests/suites/replication/regression_m3_test.py
progier389: Good idea. I changed get_agmt to rely on the target instance

PR #7140, opened 2025-12-08
Issue 6951 - Dynamic Certificate refresh phase 2 - Add/Modify/Delete support

progier389:
Second phase of [Dynamic Certificate Refresh](https://www.port389.org/docs/389ds/design/online-certificate-refresh.html)

 cn=dynamiccertificates backend now supports the following operations:
-   add
- modify
- modrdn
- delete

Allowing to store and modify dynamically the nss database

issue: https://github.com/389ds/389-ds-base/issues/6951

Reviewed by:  @tbordaz  (Thanks!)

## Summary by Sourcery

Implement dynamic certificate management operations in the dynamic certificates backend and add ECDSA-based TLS tests.

New Features:
- Support adding, modifying, renaming, and deleting certificates through the dynamic certificates backend using LDAP operations.
- Add ECDSA certificate generation and management utilities for tests, including online and offline certificate installation and nickname handling.
- Introduce dynamic certificate-based TLS search and certificate lifecycle tests, including dynamic add/rename/delete flows.

Enhancements:
- Refine dynamic certificates backend DN handling by caching the suffix DN and improving nickname parsing and validation.
- Improve certificate import and trust management logic, including verification, trust decoding, and NSS database interactions.

Tests:
- Replace the shell-based ECDSA TLS test with a Python-based test suite that generates ECDSA certificates programmatically and validates TLS connectivity.
- Add tests that verify dynamic certificate add/modify/rename/delete operations via the cn=dynamiccertificates backend.

### DIFF
--- a/ldap/servers/slapd/dyncerts.c
+++ b/ldap/servers/slapd/dyncerts.c
@@ -61,7 +61,7 @@ static const struct trust_flags_mask trust_flags[] = {
 };
 
 /* Some forward definitions */
-static DCSS *get_entry_list(const Slapi_DN *basesn, int scope, char **attrs);
+static DCSS *get_entry_list(const Slapi_DN *basedn, int scope, char **attrs);
 int dyncerts_apply_cb(const char *nickname, dyc_action_cb_t cb, void *arg, char *errmsg);
 static void dyncert_nickname_free(Nickname_t *n);
 
@@ -187,7 +187,6 @@ dyncerts_find_entry(const Slapi_DN *basedn, int scope, char **attrs, DCSS **be_s
     Slapi_Entry *e = NULL;
     DCSS *ss = NULL;
 
-    read_config_info();
     if (slapi_sdn_compare(basedn, &pdyncerts.suffix_sdn) != 0) {
         scope = LDAP_SCOPE_SUBTREE;
     }
@@ -260,6 +259,7 @@ dyncerts_search(Slapi_PBlock *pb)
     }
 
     /* Let first build the list of all entries */
+    read_config_info();
     e = dyncerts_find_entry(basesdn, scope, attrs, &be_ss);
     ss = ss_new();
     ss->pdscc = be_ss;
@@ -1027,7 +1027,7 @@ dyncerts_import_cert_and_key(CertCtx_t *ctx, bool verifyOnly)
     e = dyncerts_find_entry(ctx->sdn, LDAP_SCOPE_BASE, NULL, &ss);
     ss_destroy(&ss);
     if (!e) {
-        slapi_log_err(SLAPI_LOG_ERR, "dyncerts_add",
+        slapi_log_err(SLAPI_LOG_ERR, "dyncerts_import_cert_and_key",
                       "Failed to add certificate %s (entry not found after import).\n",
                       ctx->n.fullnickname);
         ERRMSG(ctx, LDAP_UNWILLING_TO_PERFORM,
@@ -1270,6 +1270,7 @@ dyncerts_add(Slapi_PBlock *pb)
     /*
      * Get the database, the dn and the entry to add
      */
+    read_config_info();
     if (slapi_pblock_get(pb, SLAPI_PLUGIN_PRIVATE, &pdcerts) < 0 ||
         slapi_pblock_get(pb, SLAPI_ADD_TARGET_SDN, &sdn) < 0 ||
         slapi_pblock_get(pb, SLAPI_ADD_ENTRY, &e) < 0 || (NULL == pdcerts)) {
@@ -1304,8 +1305,8 @@ done:
                       dn, rc, returntext);
     }
     ss_destroy(&ss);
-    free_config_info();
     dyncert_nickname_free(&n);
+    free_config_info();
     slapi_send_ldap_result(pb, rc, NULL, returntext[0] ? returntext : NULL, 0, NULL);
     /* The frontend does not free the added entry, so we should do it now */
     if (e) {
@@ -1395,6 +1396,9 @@ dyncerts_modify_cert(Slapi_PBlock *pb, Slapi_Entry *e, LDAPMod **mods, DCSS *ss,
     int rc = dyncerts_mods2entry(pb, sdn, mods, &n, errmsg, &newe);
     const char *nickname = n.nickname;
     PR_ASSERT(n.nickname);
+    if (rc != LDAP_SUCCESS) {
+        goto done;
+    }
     if (slapi_entry_attr_get_charptr(newe, DYCATTR_CERTDER)) {
         /* need to change the certificate ==> import the entry */
         rc = dyncerts_import_entry(newe, errmsg, true);
@@ -1500,6 +1504,7 @@ dyncerts_modify(Slapi_PBlock *pb)
     /*
      * Get the database, the dn and the modifiers
      */
+    read_config_info();
     PR_ASSERT(pb);
     if (slapi_pblock_get(pb, SLAPI_PLUGIN_PRIVATE, &pdcerts) < 0 ||
         slapi_pblock_get(pb, SLAPI_MODIFY_TARGET_SDN, &sdn) < 0 ||
@@ -1595,7 +1600,7 @@ dyncerts_list_cert_cb(CERTCertificate *cert, SECItem *sitem, void *arg)
 
 /* Generate the parent search set */
 static DCSS *
-get_entry_list(const Slapi_DN *basesn, int scope, char **attrs)
+get_entry_list(const Slapi_DN *basedn, int scope, char **attrs)
 {
     DCSS *ss = ss_new();
     int str2entry_flags = SLAPI_STR2ENTRY_EXPAND_OBJECTCLASSES |
@@ -1618,7 +1623,7 @@ get_entry_list(const Slapi_DN *basesn, int scope, char **attrs)
     if (LDAP_SCOPE_BASE == scope) {
         /* Bypass getting cert list if only looking for the container */
         Slapi_DN sdn = {0};
-        if (slapi_sdn_compare(&pdyncerts.suffix_sdn, slapi_entry_get_sdn_const(e)) == 0) {
+        if (slapi_sdn_compare(&pdyncerts.suffix_sdn, basedn) == 0) {
             return ss;
         }
         slapi_sdn_done(&sdn);
@@ -1647,6 +1652,7 @@ dyncerts_delete(Slapi_PBlock *pb)
     /*
      * Get the backend context and the dn
      */
+    read_config_info();
     if (slapi_pblock_get(pb, SLAPI_PLUGIN_PRIVATE, &pdcerts) < 0 ||
         slapi_pblock_get(pb, SLAPI_DELETE_TARGET_SDN, &sdn) < 0 ||
         (pdcerts == NULL)) {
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/dyncerts.c
```diff
@@ -839,9 +1618,7 @@ get_entry_list(const Slapi_DN *basesn, int scope, char **attrs)
     if (LDAP_SCOPE_BASE == scope) {
```
tbordaz, line 1623: It looks  basedn parameter is not used

ldap/servers/slapd/dyncerts.h
```diff
@@ -69,16 +71,23 @@ typedef enum {
     DYC_BOTH
 } dyc_et_t;
 
+/* Storage for nickname */
+typedef struct {
+    char buf[TRUST_SIZE*2];
+    char *data;
+    char *fullnickname;
```
tbordaz: May be add a comment that those fields should not be freed

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+    if (ctx->dercert.data) {
+         slapi_ch_free((void**)&ctx->dercert.data);
+         ctx->dercert.data = NULL;
+    }
+    if (ctx->derpkey.data) {
+        slapi_ch_free((void**)&ctx->derpkey.data);
+        ctx->derpkey.data = NULL;
+    }
+    dyncert_nickname_free(&ctx->n);
+    slapi_ch_free_string(&ctx->trust);
```
tbordaz: This fields is sometime set with static string (e.g. "CT,,") we should have a flag to free it or not

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+    int rc = LDAP_SUCCESS;
+    CertCtx_t ctx = {0};
+
+    dyncert_nickname_from_full_nickname(&ctx.n, slapi_entry_attr_get_ref(e, DYCATTR_NICKNAME));
+    ctx.dercert = slapi_entry_attr_get_secitem(e, DYCATTR_CERTDER);
+    ctx.derpkey = slapi_entry_attr_get_secitem(e, DYCATTR_PKEYDER);
+    ctx.trust = slapi_entry_attr_get_charptr(e, DYCATTR_TRUST);
+    ctx.force = (slapi_entry_attr_get_bool(e, DYCATTR_FORCE) == PR_TRUE);
+    ctx.errmsg = errmsg;
+    ctx.sdn = (Slapi_DN*) slapi_entry_get_sdn_const(e);
```
tbordaz, line 1250: It is a reference in an entry that is later freed. We should duplicate the sdn and free it in dyncerts_cert_ctx_done

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+    }
+    dyncert_nickname_free(&ctx->n);
+    slapi_ch_free_string(&ctx->trust);
+    ctx->force = false;
+    if (ctx->slot) {
+        PK11_FreeSlot(ctx->slot);
+        ctx->slot = NULL;
+    }
+    dyncerts_free_nss_add_cert_and_key(ctx);
+    ctx->ldaprc = 0;
```
tbordaz, line 839: Should also free 'errmsg'

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+        rv = PK11_Authenticate(ctx->slot, PR_TRUE, pw);
+        CHECK_ERR("Failed to authenticate to token")
+    }
+    cert = CERT_DecodeCertFromPackage((char *)ctx->dercert.data, ctx->dercert.len);
+    ctx->cert = cert;
+    if (cert == NULL) {
+        errmsg = "Failed to decode certificate";
+        goto done;
+    }
+    if (cert->isperm && !(verifyOnly && strcasecmp(ctx->n.nickname, cert->nickname) == 0)) {
```
tbordaz, line 909: are we sure that cert->nickname is always set ?

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+         * by itself
+         */
+        rv = SECFailure;
+        ctx->ldaprc = LDAP_UNWILLING_TO_PERFORM;
+        PR_snprintf((ctx)->errmsg, SLAPI_DSE_RETURNTEXT_SIZE, "Confict with certificate %s",
+                    cert->nickname);
+        goto done2;
+    }
+    ctx->primary = is_servercert_int(ctx->n.token, ctx->n.nickname);
+    if (!ctx->trust) {
```
tbordaz, line 921: It does not reset the trust string. But if the cert/key defining changes, may be the trust will also change ?

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+    if (rc) {
+        if (!error) {
+            slapi_pblock_set(pb, SLAPI_PLUGIN_OPRETURN, &rc);
+        }
+        slapi_log_err(SLAPI_LOG_DYC, "dyncerts_add",
+                      "Add operation of entry %s failed: rc=%d : %s\n",
+                      dn, rc, returntext);
+    }
+    ss_destroy(&ss);
+    free_config_info();
```
tbordaz: If we raise this point before calling dyncerts_import_entry (goto done), the mutex was not acquire yet.

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+    }
+done:
+    if (rc) {
+        slapi_pblock_set(pb, SLAPI_PLUGIN_OPRETURN, &rc);
+        slapi_log_err(SLAPI_LOG_DYC, "dyncerts_modify",
+                      "Modify operation of entry %s failed: rc=%d : %s\n",
+                      dn, rc, returntext);
+    }
+    ss_destroy(&ss);
+    free_config_info();
```
tbordaz, line 1535: If we raise this point without calling dyncert_find_entry (goto) then mutex is not acquired

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+}
+
+/* modify a certificate entry */
+int
+dyncerts_modify_cert(Slapi_PBlock *pb, Slapi_Entry *e, LDAPMod **mods, DCSS *ss, char *errmsg)
+{
+    Slapi_DN *sdn = (Slapi_DN *) slapi_entry_get_sdn_const(e);
+    Slapi_Entry *newe = NULL;
+    Nickname_t n = {0};
+    int rc = dyncerts_mods2entry(pb, sdn, mods, &n, errmsg, &newe);
```
tbordaz, line 1396: If rc not SUCCESS we should go to done

tbordaz (commented):
Some comments

ldap/servers/slapd/dyncerts.c
progier389: No. errmsg is not allocated but rather set to some local buffer

ldap/servers/slapd/dyncerts.c
progier389: Yes: cert->isperm means that the certificate has been found within nss db (so it has a nickname since it is required to add the cert in the db)

ldap/servers/slapd/dyncerts.c
progier389: I do not understand what you mean by 'change' in this context:
We are in the function that add a certificate in the db  either trust is provided and used or some default trusts are set (according the certificate type).

ldap/servers/slapd/dyncerts.c
progier389: Not an issue because dyncerts_cert_ctx_done is called before the entry get freed.
   (in other words: the reference lifetime is shorter than the entry lifetime)
Allocating and freeing the sdn would be a needless overhead

ldap/servers/slapd/dyncerts.c
progier389: I think that you means before dyncerts_find_entry , but is is a valid point

ldap/servers/slapd/dyncerts.c
tbordaz: Okay, so if the trust is provided (!ctx->trust) we use it else we "compute" it.
What surprised me is that "computed" trust is for example set to "CTu,u,u" if ctx->primary.
That means that the provided trust can differ from ""CTu,u,u"" even if ctx->primary

ldap/servers/slapd/dyncerts.c
```diff
@@ -798,6 +797,786 @@ dyncerts_cert2entry(CERTCertificate *cert)
...
+
+/* dyncerts_apply_cb callback to change trust flags */
+void
+dyncert_set_trust_cb(CertCtx_t *ctx)
+{
+    Slapi_Entry *e = ctx->arg;
+    CERTCertTrust trust = {0};
+    int rv = 0;
+
+    ctx->trust = slapi_entry_attr_get_charptr(e, DYCATTR_TRUST);
```
tbordaz, line 1362: Shouldn't free ctx->trust before setting it. Also if we return UNWILLING_TO_PERFORM, revert the original value.

ldap/servers/slapd/dyncerts.c
progier389: Yes. The idea is to provide a way to let the administrator to control the trust flags (as it does when using certutil -T)  
It the flags are very wrongs then the verification may fail (because cert is not trusted)

BTW the usual trust for primary cert is u,u,u  CT,u,u is for the rare case (both primary vert and self signed CA)

ldap/servers/slapd/dyncerts.c
progier389: ctx->trust is NULL when entering dyncert_set_trust_cb 
because it is called by dyncerts_apply_cb which free the ctx between each callback call
No need to revert the value. it will be freed anyway by dyncerts_apply_cb

ldap/servers/slapd/dyncerts.c
progier389: Good point. Now fixed by moving read_config_info outside of find_entry 
and calling it first thing for each operation

ldap/servers/slapd/dyncerts.c
progier389: Fixed. In fact one of the slapi_sdn_compare test was wrong (Comparing root entry dn with suffix dn (which is always true) instead of the search basedn)

ldap/servers/slapd/dyncerts.c
progier389: fixed rc is not tested

ldap/servers/slapd/dyncerts.c
progier389: Fixed, now the mutex is taken first thing add/del/modify/modrdn operations

ldap/servers/slapd/dyncerts.c
tbordaz: okay

ldap/servers/slapd/dyncerts.c
tbordaz: okay thanks

tbordaz (approved):
Thanks for the explanations. LGTM

PR #7141, opened 2025-12-08
Issue 1793 - RFE - Implement dynamic lists

mreynolds389:
Implement a backend feature to build dynamic content based of LDAP URI's. Configuration includes an identifying objectclass to mark an entry as a dynamic content entry. Another setting for the attribute that contains the LDAP URI, and an attribute for storing the dynamic content. Attributes specified in the LDAP URI override the content attribute and instead write that attribute's value into the dynamic content entry.

Design doc: https://www.port389.org/docs/389ds/design/dynamic-lists-design.html

Relates: https://github.com/389ds/389-ds-base/issues/1793

## Summary by Sourcery

Add backend support for dynamic LDAP list entries that are materialized on search using LDAP URLs and are never cached persistently.

New Features:
- Introduce dynamic list handling in the ldbm backend, populating dynamic attributes based on LDAP URL searches and marking such entries as transient.
- Add configurable backend settings to enable dynamic lists and define the marker object class, URL attribute, and target list attribute.

Enhancements:
- Ensure dynamic entries are excluded from the entry cache so that their content is rebuilt on each access, and integrate dynamic list evaluation into candidate list building for relevant searches.

Tests:
- Add an automated test suite for dynamic lists covering basic member URL expansion, non-DN attribute population, invalid configuration handling, and multiple URL aggregation behavior.

### DIFF
--- a/dirsrvtests/tests/suites/dynamic_lists/dynamic_lists_test.py
+++ b/dirsrvtests/tests/suites/dynamic_lists/dynamic_lists_test.py
@@ -1,0 +1,312 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2025 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+import ldap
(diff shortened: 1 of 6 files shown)


mreynolds389:
I will be adding the CLI and UI portions of this in a future PR

progier389 (approved):
LGTM

PR #7142, opened 2025-12-08
Issue 7119 - Add DNA stress test

droideck:
Description: Add multi-supplier DNA stress test with 34 concurrent threads targeting race conditions in dna_load_shared_servers(). Test runs user creation, shared config modifications, MemberOf fixup, and group operations across 4-supplier topology with 10K pre-imported users.

Successfully reproduces undefined behavior with ASAN builds, exposing races where concurrent config updates access the global shared server list during reconstruction without proper locking.

Relates: https://github.com/389ds/389-ds-base/issues/7119

Reviewed by: ?

## Summary by Sourcery

Tests:
- Introduce a long-running stress test that drives concurrent user creation, DNA shared configuration churn, MemberOf fixup tasks, and group membership/group lifecycle changes across a four-supplier topology to detect dna_load_shared_servers() race conditions.

### DIFF
--- a/src/lib389/lib389/topologies.py
+++ b/src/lib389/lib389/topologies.py
@@ -470,6 +470,17 @@ def topology_i3(request):
     topology.logcap = LogCapture()
     return topology
 
+
+@pytest.fixture(scope="module")
+def topology_i4(request):
+    """Create three instance DS deployment"""
+
+    topology = create_topology({ReplicaRole.STANDALONE: 4}, request=request)
+
+    topology.logcap = LogCapture()
+    return topology
+
+
 @pytest.fixture(scope="module")
 def topology_m1(request):
     """Create Replication Deployment with one supplier and one consumer"""
--- a/src/lib389/lib389/topologies.py
+++ b/src/lib389/lib389/topologies.py
@@ -473,7 +473,7 @@ def topology_i3(request):
 
 @pytest.fixture(scope="module")
 def topology_i4(request):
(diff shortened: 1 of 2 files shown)


tbordaz:
Nice catch !
My understanding is that the stress test, leading to asan violation, is based only on updates (create users and group membership) on a single suffix. Is that correct  ?
What are type of asan violation (double free, use after free,..) ? Does it create a asan abort ?

progier389 (approved):
LGTM

src/lib389/lib389/topologies.py
```diff
@@ -470,6 +470,17 @@ def topology_i3(request):
     topology.logcap = LogCapture()
     return topology
 
+
+@pytest.fixture(scope="module")
+def topology_i4(request):
+    """Create three instance DS deployment"""
```
tbordaz: Three or four ?

tbordaz (commented):
The test coverage looks very impressive.
My understanding is that the fix protects the loading of shared config/servers (dna_config_check_post_op/dna_load_shared_servers) during direct updates. Hammering the instances with ADD/MOD/DEL looks good.
It looks that memberof plugin is configured/used during the tests. I do not see the benefit of this plugin to reproduce the issue. IMHO it should be disabled.
Except that the fix looks good.

src/lib389/lib389/topologies.py
droideck: Four. Copy-paste typo :)

Issue #7147, opened 2025-12-10
entrycache_eviction_test CI test is failing


progier389:
**Issue Description**
dirsrvtests/tests/suites/features/entrycache_eviction_test.py::test_entry_cache_eviction FAILED [ 50%]

**Package Version and Platform:**
 - Platform: [e.g. Fedora]
 - Package and version: main branch

**Steps to Reproduce**
1. Look at CI tests results 

**Expected results**
Test should pass

droideck:
Follow-up CI sweep found this is still active on recent `main` runs, so I did not open a duplicate issue.

Recent signal:
- LMDB `features` failures still hit `dirsrvtests/tests/suites/features/entrycache_eviction_test.py::test_entry_cache_eviction`.
- Representative run: https://github.com/389ds/389-ds-base/actions/runs/26734200121, job `78784950161`.
- Representative assertion: `assert 0 == 5` at `entrycache_eviction_test.py:233`.

This looks separate from the new Quay/image-pull failures because pytest starts and produces a normal failure summary.

progier389:
Although I found a few occurrences of the assert 0 == 5  issue, most recent failures of test_entry_cache_eviction test case are due to core dumps which leads to ldap.SERVER_DOWN errors
I will first focus on these crashes

progier389:
#0  __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex-base.S:81
#1  0x00007f2b9f3693df in entrycache_remove_int (cache=cache@entry=0x7f2ba23e8890, e=e@entry=0x7f2b79e1f5e0) at ldap/servers/slapd/back-ldbm/cache.c:1167
#2  0x00007f2b9f369b5f in entrycache_flush (cache=cache@entry=0x7f2ba23e8890) at ldap/servers/slapd/back-ldbm/cache.c:842
#3  0x00007f2b9f36b316 in entrycache_add_int (cache=0x7f2ba23e8890, e=0x7f2b74e3bdc0, state=state@entry=0, alt=alt@entry=0x7f2b79df8970)
    at ldap/servers/slapd/back-ldbm/cache.c:1999
#4  0x00007f2b9f36b75a in cache_add (cache=cache@entry=0x7f2ba23e8890, ptr=<optimized out>, alt=alt@entry=0x7f2b79df8970)
    at ldap/servers/slapd/back-ldbm/cache.c:2034
[...]

PR #7148, opened 2025-12-10
Issue: 7147 - entrycache_eviction_test is failing

progier389:
Several reason explain the test failure:

- log buffereing is not disabled
- a race condition causing double free in slapi_re_exec_nt when several thread uses the same compiled regex
- The searched done during the test were silently unindexed so some entries were unexpectedly added  to the entry cache

Issue: #7147

## Summary by Sourcery

Stabilise entry cache eviction behaviour and debug logging by tightening test searches and making regex-based cache debugging thread-safe.

Bug Fixes:
- Prevent concurrent use of the shared cache debug regex from causing race conditions by serialising calls to slapi_re_exec_nt with a mutex.

Enhancements:
- Ensure cache debugging output is consistently captured by disabling error log buffering for the test.
- Avoid interference from the automember plugin during entry cache eviction testing by disabling it for the test instance.

Tests:
- Update entry cache eviction tests to search only for relevant group and person object classes so entries remain indexed and eviction behaviour is deterministic.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -1061,7 +1061,12 @@ static bool
 debug_pattern_matches(struct cache *cache, const char *dn)
 {
     if (cache->c_inst->cache_debug_re && dn) {
-        if (slapi_re_exec_nt(cache->c_inst->cache_debug_re, dn)) {
+        static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+        int rc = 0;
+        pthread_mutex_lock(&mutex);
+        rc = slapi_re_exec_nt(cache->c_inst->cache_debug_re, dn);
+        pthread_mutex_unlock(&mutex);
+        if (rc) {
             return true;
         }
     }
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1061,7 +1061,12 @@ static bool
 debug_pattern_matches(struct cache *cache, const char *dn)
 {
     if (cache->c_inst->cache_debug_re && dn) {
-        if (slapi_re_exec_nt(cache->c_inst->cache_debug_re, dn)) {
+        static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
droideck, line 1064: More of a nitpick, does it make sense to add the mutex as `cache->c_inst->cache_debug_re_mutex`? So different backends won't contend on the same mutex unnecessarily?

But it's probably a very minor optimization...

droideck (approved):
LGTM!

tbordaz (approved):
LGTM as well

ldap/servers/slapd/back-ldbm/cache.c
progier389: Not really concerned about contention in this code. 
Anyway in production cache->c_inst->cache_debug_re is NULL 
(This debug feature is mostly intended for the CI test)

PR #7151, opened 2025-12-11
Issue 7150 - Compressed access log rotations skipped, accesslog-list out of sync

### LINKED ISSUE #7150: Compressed access log rotations skipped, leaving accesslog-list out of sync
**Description**
Compressed access/audit/error log files (*.gz) are being skipped when rebuilding the server’s internal rotation list. That could leave old compressed logs undeleted and the nsslapd-accesslog-list out of sync with what's on disk.

droideck:
Description: Accept `.gz`-suffixed rotated log filenames when rebuilding rotation info and checking previous logs, preventing compressed rotations from being dropped from the internal list.

Add regression tests to stress log rotation with compression, verify `nsslapd-accesslog-list` stays in sync, and guard against crashes when flushing buffered logs during rotation. Minor doc fix in test.

Fixes: https://github.com/389ds/389-ds-base/issues/7150

Reviewed by: ?

## Summary by Sourcery

Ensure access log rotation correctly tracks compressed and uncompressed files and keep the server’s internal access log list in sync with on-disk logs.

Bug Fixes:
- Include compressed (.gz) rotated access logs when rebuilding rotation metadata so they are not dropped from the internal list.
- Handle previous log discovery for compressed access logs by correctly parsing timestamps and extensions to avoid list/file mismatches.
- Prevent potential crashes and inconsistencies when flushing buffered logs during access log rotation under heavy load.

Enhancements:
- Add stress tests that exercise access log rotation with compression, verify maxlogsperdir enforcement, and assert nsslapd-accesslog-list matches actual files.
- Clarify and slightly adjust an existing log flush and rotation crash test, including a minor doc string fix and ordering note.

### DIFF
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -137,6 +137,7 @@ static void vslapd_log_emergency_error(LOGFD fp, const char *msg, int locked);
 static int get_syslog_loglevel(int loglevel);
 static void log_external_libs_debug_openldap_print(char *buffer);
 static int log__fix_rotationinfof(char *pathname);
+static int log__validate_rotated_logname(const char *timestamp_str, PRBool *is_compressed);
 
 static int
 get_syslog_loglevel(int loglevel)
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/log.c
```diff
@@ -4923,13 +4923,17 @@ log__fix_rotationinfof(char *pathname)
                     ignoreit = 1;
                 }
             }
-            if (ignoreit || (q - p != 15)) {
+            /*
+             * Accept timestamp length of 15 (YYYYMMDD-HHMMSS) for uncompressed,
+             * or 18 (YYYYMMDD-HHMMSS.gz) for compressed files.
+             */
+            if (ignoreit || ((q - p != 15) && (q - p != 18))) {
```
progier389: Should also ignore if q-p == 18 && strcmp(p + 15, ".gz"))

Maybe we should rather simplify the code and use a regex ?

progier389 (approved):
LGTM

PR #7153, opened 2025-12-15
Issue 6753 - Port ticket47953 test to acl/misc_test.py using DSLdapOb…

aadhikar:
…ject

Description: The old ticket47953_test.py had compatibility issues. This ports the functionality to acl/misc_test.py::test_delete_aci_with_invalid_syntax using modern DSLdapObject methods.

Relates: https://github.com/389ds/389-ds-base/issues/6753

Reviewed by: ???

## Summary by Sourcery

Port the legacy ticket47953 ACI deletion test into the ACL misc test suite using the modern testing APIs and remove the old ticket-specific test artifacts.

New Features:
- Add a misc ACL test that verifies an invalid ACI can be imported and subsequently deleted without errors.

Tests:
- Add test_delete_aci_with_invalid_syntax to validate handling and deletion of syntactically invalid ACIs using ImportTask and Domain helpers.
- Remove the obsolete ticket47953-specific LDIF test data and its standalone test file.


dirsrvtests/tests/suites/acl/misc_test.py
```diff
@@ -499,6 +504,54 @@ def finofaci():
...
+
+    def fin():
+        if os.path.exists(ldif_file):
+            os.remove(ldif_file)
+    request.addfinalizer(fin)
+
+    log.info("Import LDIF with invalid ACI")
+    import_task = ImportTask(inst)
+    import_task.import_suffix_from_ldif(ldiffile=ldif_file, suffix=DEFAULT_SUFFIX)
+    import_task.wait()
```
droideck, line 558: I think we need to check the import task exit status here - `import_task.get_exit_code()`

droideck (changes requested):
The test suite is a bit destructive as it re-imports the entire DEFAULT_SUFFIX via without restoring prior tree/ACIs.
We either need to limit the fixture scope to function or restore the previous state. Easier is to use an instance that will be destroyed after the test, IMO.

droideck (approved):
LGTM!

PR #7154, opened 2025-12-16
Issue 6951 - Dynamic Certificate refresh phase 4 - NssSsl PKCS#12 support

jchapma:
Description:
Dynamic certs provide a virtual backend that exposes the in memory NSS DB.

If the virtual backend is offline, lib389 can fall back to using the on disk NSS DB. These changes form the basis for a unified interface so lib389 can handle PKCS#12 imports consistently, regardless of backend.

PKCS#12 files contain both the cert and its private key, enabling online server cert switching. PEM/DER formats usually separate the cert and key, complicating matters.

Fix:
Extend NssSsl.add_cert to accept PKCS#12 files with password, text or path. Added PKCS#12 support to cert verification (static) Add CI tests for PEM, DER, and PKCS#12 cert imports

Relates: https://github.com/389ds/389-ds-base/issues/6951

Reviewed by:

## Summary by Sourcery

Add unified NSS SSL certificate import support for PEM, DER, and PKCS#12 formats with stricter validation and duplicate detection, and extend tests to cover the new behaviors.

New Features:
- Support importing PKCS#12 server certificates into the NSS DB with configurable password handling and automatic nickname resolution.
- Allow importing PEM CA bundles with multiple certificates while enforcing constraints for server certificate bundles.

Bug Fixes:
- Normalize error reporting for unsupported certificate inputs and invalid imports in TLS-related commands and NSS SSL operations.

Enhancements:
- Extend NssSsl.add_cert to handle PEM, DER, and PKCS#12 inputs with nickname requirements, CA/server validation, and flexible trust flag handling.
- Add certificate duplicate detection based on DER-encoded hash comparisons across the NSS database.
- Provide an option to override CA verification via a do_it flag when importing potentially invalid CA certificates.
- Expose raw DER-encoded certificate bytes from the NSS DB via get_cert_details to support binary-level comparisons.
- Update CA bundle splitting to use predictable .pem filenames for easier cleanup in tests.

Tests:
- Add comprehensive tests for importing PEM, DER, and PKCS#12 certificates into a temporary NSS database, including success and failure cases such as missing nicknames, duplicate nicknames, missing files, bundles, and do_it overrides.
- Add helper utilities and fixtures in dsctl_tls_test to generate temporary certificates in multiple formats and inspect NSS database contents.

### DIFF
--- a/src/lib389/lib389/nss_ssl.py
+++ b/src/lib389/lib389/nss_ssl.py
@@ -10,6 +10,7 @@
 """
 
 import copy
+import hashlib
 import os
 import re
 import socket
@@ -24,6 +25,8 @@ from lib389.passwd import password_generate
 from lib389._mapped_object_lint import DSLint
 from lib389.lint import DSCERTLE0001, DSCERTLE0002
 from lib389.utils import ensure_str, format_cmd_list, DSVersion, cert_is_ca
+from cryptography import x509
+from cryptography.hazmat.primitives import serialization
 
 
 KEYBITS = 4096
@@ -1096,19 +1099,54 @@ only.
 
         return ensure_str(result)
 
-    def get_cert_details(self, nickname):
-        """Get the trust flags, subject DN, issuer, and expiration date
+    def get_cert_details(self, nickname, raw_bytes=False):
+        """
+        Get cert details or raw cert bytes from the NSS DB.
 
-        return a list:
-            0 - nickname
-            1 - subject
-            2 - issuer
-            3 - expire date
-            4 - trust_flags
+        Parameters:
+            nickname (str): The NSS certificate nickname to look up.
+            raw_bytes (bool): If True, return the certs DER encoded
+                bytes rather than parsed metadata. Used for detecting
+                duplicate certs by hashing the binary DER file.
+
+        Returns:
+            If raw_bytes=True:
+                bytes — the DER-encoded certificate.
+
+            If raw_bytes=False:
+                list containing:
+                    0 - nickname (str)
+                    1 - subject DN (str)
+                    2 - issuer DN (str)
+                    3 - expiration date (str)
+                    4 - trust flags (str)
+
+        Raises:
+            ValueError: If the certificate is not found, or if certutil fails.
         """
         all_certs = self._rsa_cert_list()
         for cert in all_certs:
             if cert[0] == nickname:
+                if raw_bytes:
+                    # Export the certificate in PEM and then convert to DER binary format
+                    # NSS's internal storage might differ from the input format (PEM/DER),
+                    # so the cert is first exported in PEM format (certutil -L -a), then
+                    # converted to DER
+                    cmd = [
+                        "/usr/bin/certutil",
+                        "-L",
+                        "-d", self._certdb,
+                        "-n", nickname,
+                        "-a"
+                    ]
+                    try:
+                        output = check_output(cmd, stderr=subprocess.STDOUT)
+                        pem_data = output.decode("utf-8")
+                        cert_obj = x509.load_pem_x509_certificate(pem_data.encode("utf-8"))
+                        return cert_obj.public_bytes(serialization.Encoding.DER)
+                    except subprocess.CalledProcessError as e:
+                        raise ValueError(e.output.decode("utf-8").rstrip())
+
                 trust_flags = cert[1]
                 certdetails = self.display_cert_details(nickname)
                 end_date_str = certdetails.split("Not After : ")[1].split("\n")[0]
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/nss_ssl.py
```diff
@@ -1096,19 +1099,54 @@ def display_cert_details(self, nickname):
...
+                    # converted to DER
+                    cmd = [
+                        "/usr/bin/certutil",
+                        "-L",
+                        "-d", self._certdb,
+                        "-n", nickname,
+                        "-a"
+                    ]
+                    try:
+                        output = check_output(cmd, stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
progier389: Not really an issue in that case (unprivileged CLI ) 
command injection provides no more benefit than directly typing the command in the shell prompt instead of dsctl/dsconf command.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
-        else:
-            self._assert_not_chain(input_file)
+        if not self._db_exists(even_partial=True):
+            log.info('Security database does not exist. Creating a new one in {}.'.format(self.dirsrv.get_cert_dir()))
+            self.tlsdb.reinit()
+
+        # To really check for duplicates we compare cert hash
+        with open(input_file, "rb") as f:
+            input_bytes = f.read()
+        input_hash = hashlib.sha256(input_bytes).hexdigest()
```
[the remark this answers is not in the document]
progier389: sourcery is right:

the input_file certificate should be converted in DER before being hashed.
Adding a cert_to_der(function ?)
can be done easily by using openssl cryptography module
FYI: you can find example of python code extracting DER from PKCS12 in
online_install function in dirsrvtests/tests/suites/tls/ecdsa_test.py

And IMHO once you have the DER you can also use it to check if it is a CA or not:
def is_ca(der):
      cert = x509.load_der_x509_certificate(der, default_backend())
      try:
          basic_constraint=cert.extensions.get_extension_for_oid(ExtensionOID.BASIC_CONSTRAINTS)
          return basic_constraints.value.ca
      except x509.ExtensionNotFound:
          return False

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
+            if pkcs12_pin_path:
+                p12_pin = pkcs12_pin_path
+                p12_flag = "-w"
+            elif pkcs12_pin_text:
+                p12_pin = pkcs12_pin_text
+                p12_flag = "-W"
+            elif pkcs12_pin_stdin:
+                p12_pin = "-"
+                p12_flag = "-W"
+            else:
```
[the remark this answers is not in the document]
progier389: probably not a good idea of providing pkc12 to cert_is_ca function use rather directly the der

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
             raise ValueError("The certificate file ({}) does not exist".format(input_file))
 
-        pem_file = True
-        if not input_file.lower().endswith(".pem"):
-            pem_file = False
-        else:
-            self._assert_not_chain(input_file)
+        if not self._db_exists(even_partial=True):
+            log.info('Security database does not exist. Creating a new one in {}.'.format(self.dirsrv.get_cert_dir()))
+            self.tlsdb.reinit()
```
droideck, line 1236: I think something is missing as the `self.tlsdb` attribute is undefined.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
         if not os.path.exists(input_file):
             raise ValueError("The certificate file ({}) does not exist".format(input_file))
 
-        pem_file = True
-        if not input_file.lower().endswith(".pem"):
-            pem_file = False
-        else:
-            self._assert_not_chain(input_file)
+        if not self._db_exists(even_partial=True):
+            log.info('Security database does not exist. Creating a new one in {}.'.format(self.dirsrv.get_cert_dir()))
```
droideck, line 1235: If `self.dirsrv` is None, `self.dirsrv.get_cert_dir()` will raise an exception.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
+
+        # To really check for duplicates we compare cert hash
+        with open(input_file, "rb") as f:
+            input_bytes = f.read()
+        input_hash = hashlib.sha256(input_bytes).hexdigest()
+
+        for nick, _ in self._rsa_cert_list():
+            try:
+                existing_bytes = self.get_cert_details(nick, raw_bytes=True)
+                if hashlib.sha256(existing_bytes).hexdigest() == input_hash:
```
droideck, line 1246: Do I understand correctly that the input file could be PEM, but `get_cert_details(raw_bytes=True)` returns DER? So their hashes will never match.

Also, PKCS#12 files contain encrypted data + private keys -- their hash will never match the extracted certificate's DER bytes.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,45 +1208,184 @@ def list_client_ca_certs(self):
...
+                if not certs:
+                    raise ValueError("PKCS#12 import succeeded but no certificates were found in the DB")
+                # NSS lists them in import order, so use the last one
+                nickname = certs[-1][0]
+
+            # Apply trust flags
+            if ca:
+                try:
+                    # Check if the certificate is a CA
+                    if not cert_is_ca(input_file, pkcs12_password=p12_pin) and not do_it:
```
droideck, line 1303: I'm not sure I fully understand the `pkcs12_pin_path -> p12_pin` logic. `p12_pin` can be a path at this point, right? but `cert_is_ca` expects a password. Also I agree with Pierre's comments.

droideck (changes requested):
Also, `Validate tests and minimal Python version / validate (pull_request)` is failed for dirsrvtests/tests/suites/clu/dsctl_tls_test.py

src/lib389/lib389/nss_ssl.py
progier389: IMHO self._certdb should rather be used (and that is quite a general comment:
 use self._certdb instead of self.dirsrv.get_cert_dir() everywhere except in __init__

src/lib389/lib389/nss_ssl.py
progier389: Yes, that was also what sourcery complained about.
IMHO we should better systematically convert the certificate(s) in DER from the start (and probably not by checking the suffix by directly by checking the file contents (cryptolib provide the functions to do that)

jchapma:
The intention was to have smaller commits to make code reviews more manageable. However, when I started working on the actual DynamicCerts backend code, I ended up including NssSsl changes as well. This commit is not really fit for purpose, perhaps I didn’t fully understand the big picture. I am therefore closing this PR and will open a fresh one containing all the changes.

I apologise for wasting your review time.

PR #7156, opened 2025-12-17
Issue 7155 - build_candidate_list - Database error 11 with range search

### LINKED ISSUE #7155: build_candidate_list - Database error 11 with range search
**Issue Description**
Range search that hit both nsslapd-rangelookthroughlimit and filter AND shortcut accelerator 
 returns OPERATION_ERROR after displaying an error message:
build_candidate_list - Database error 11 
  

 - Platform: main
 - Package and version: all versions up to the main

**Steps to Reproduce**
Test case will be provided in PR

**Expected results**
Search should be successful  with having to increase nsslapd-rangelookthroughlimit

**Additional context**
Increasing nsslapd-rangelookthroughlimit is a work-around

progier389:
Range search behavior discrepancy when hitting  nsslapd-rangelookthroughlimit whether filter_candidates_ext returns a short or long list of candidates:
  Search is successful if the candidate list is long but fail if it is short
 Error should be ignored in both case 
 
 Issue: #7155 
 
## Summary by Sourcery

Handle range search candidate list errors consistently and validate correct behavior with a new regression test.

Bug Fixes:
- Avoid returning spurious database error codes from subtree candidate searches when a non-empty candidate list is available.

Tests:
- Add a regression test and supporting LDAP objects/fixtures to verify range search behavior when nsslapd-rangelookthroughlimit is hit.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_search.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_search.c
@@ -1553,6 +1553,8 @@ subtree_candidates(
             idl_free(&tmp);
             idl_free(&descendants);
         }
+    } else if (candidates != NULL) {
+        *err = LDAP_SUCCESS;
     }
 
     return (candidates);
--- a/dirsrvtests/tests/suites/indexes/regression_test.py
+++ b/dirsrvtests/tests/suites/indexes/regression_test.py
@@ -12,22 +12,29 @@ import pytest
 import ldap
 import logging
 import glob
+import random
 import re
-from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
+from contextlib import suppress
 from lib389.backend import Backend, Backends, DatabaseConfig
-from lib389.cos import  CosClassicDefinition, CosClassicDefinitions, CosTemplate
+from lib389.cli_base import FakeArgs
+from lib389.cli_ctl.dbgen import dbgen_create_groups
 from lib389.cli_ctl.dblib import DbscanHelper
+from lib389.config import LDBMConfig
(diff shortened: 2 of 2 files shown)


ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -1557,6 +1553,8 @@ subtree_candidates(
             idl_free(&tmp);
             idl_free(&descendants);
         }
+    } else if (candidates != NULL) {
```
tbordaz, line 1556: Why putting this in the 'else'. Shouldn't we assume that whatever paths we took, if candidate is not NULL then we should not report an error.
Also LDAP_SUCCESS looks better than 0.

progier389: subree_candidate logic was:   
    candidates = filter_candidates_ext(..., err, ...);
    if (candidates && num(candidates) >= SOME_THRESHOLD) {
        *err = ldbm_ancestorid_read_ext(...)
    }
So  err set by filter_candidates_ext was overwritten if we read ancestors but not if we do not read it.
I could have set err to 0 BEFORE the if or in the ELSE (I just choose the second solution) 
but we should not do it after the if and without else because it override the ldbm_ancestorid_read_ext  error and I am not sure that we should ignore it (but we could ignore filter_candidates_ext error because that is what is already done in the usual case)

ok for the cosmetic change about LDAP_SUCCESS

tbordaz: Yes I agree

tbordaz (approved):
LGTM

PR #7163, opened 2025-12-23
Issue 7162 - Fix shutdown crash in entry cache destruction

### LINKED ISSUE #7162: [3.1.4][IPA] crash on service shutdown
**Issue Description**

As from `3.1.4` `ns-slapd` (integrated with IPA) randomly (?) crashes on shutdown.
```
systemd[1]: Stopping dirsrv@IPA-TEST.service - 389 Directory Server IPA-TEST....
ns-slapd[9480]: [19/Dec/2025:11:14:25.899829997 +0000] - INFO - ct_thread_cleanup - slapd shutting down - signaling connection table threads
ns-slapd[9480]: [19/Dec/2025:11:14:25.901296491 +0000] - INFO - op_thread_cleanup - slapd shutting down - signaling operation threads - op stack size 9 max work q size 6 max work q stack size 6
ns-slapd[9480]: [19/Dec/2025:11:14:25.904731353 +0000] - INFO - slapd_daemon - slapd shutting down - waiting for 2 threads to terminate
ns-slapd[9480]: [19/Dec/2025:11:14:26.143285859 +0000] - INFO - slapd_daemon - slapd shutting down - accept_thread
ns-slapd[9480]: [19/Dec/2025:11:14:26.143801946 +0000] - INFO - slapd_daemon - slapd shutting down - closing down internal subsystems and plugins
systemd[1]: dirsrv@IPA-TEST.service: Main process exited, code=dumped, status=11/SEGV
systemd[1]: dirsrv@IPA-TEST.service: Failed with result 'core-dump'.
systemd[1]: Stopped dirsrv@IPA-TEST.service - 389 Directory Server IPA-TEST..
```

stacktrace:
```
Core was generated by `/usr/sbin/ns-slapd -D /etc/dirsrv/slapd-IPA-TEST -i /run/dirsrv/slapd-IPA-TEST.'.
#0  entrycache_flush (cache=cache@entry=0x7fdc492d5790)
    at ldap/servers/slapd/back-ldbm/cache.c:776
776	        e->ep_refcnt++;

Thread 1 (Thread 0x7fdc49d1c480 (LWP 9480)):
#0  entrycache_flush (cache=cache@entry=0x7fdc492d5790) at ldap/servers/slapd/back-ldbm/cache.c:776
        e = 0x0
#1  0x00007fdc46173924 in entrycache_clear_int (cache=0x7fdc492d5790) at ldap/servers/slapd/back-ldbm/cache.c:802
        eflush = 0x0
        eflushtemp = 0x0
        size = 738197504
#2  0x00007fdc461739fd in erase_cache (cache=0x7fdc492d5790, type=<optimized out>) at ldap/servers/slapd/back-ldbm/cache.c:837
#3  cache_destroy_please (cache=0x7fdc492d5790, type=<optimized out>) at ldap/servers/slapd/back-ldbm/cache.c:852
#4  0x00007fdc4618a997 in ldbm_instance_stop_cache (be=<optimized out>) at ldap/servers/slapd/back-ldbm/instance.c:334
        inst = 0x7fdc492d5700
#5  0x00007fdc4618a9f1 in ldbm_instance_stopall_caches (li=0x7fdc492c9500) at ldap/servers/slapd/back-ldbm/instance.c:395
        inst_obj = warning: could not convert 'object' from the host encoding (UTF-8) to UTF-32.
This normally should not happen, please file a bug report.
0x7fdc4684d680
        inst = <optimized out>
#6  0x00007fdc46179170 in ldbm_back_close (pb=0x7fd97ec008a0) at ldap/servers/slapd/back-ldbm/close.c:35
        li = 0x7fdc492c9500
#7  0x00007fdc4afffcd5 in plugin_call_func (list=0x7fdc4664a600, operation=operation@entry=210, pb=pb@entry=0x7fd97ec008a0, call_one=call_one@entry=1) at ldap/servers/slapd/plugin.c:1996
        n = <optimized out>
        func = 0x7fdc461790f0 <ldbm_back_close>
        rc = <optimized out>
        return_value = 0
        count = 0
#8  0x00007fdc4b000272 in plugin_call_one (list=<optimized out>, operation=210, pb=0x7fd97ec008a0) at ldap/servers/slapd/plugin.c:1945
#9  plugin_dependency_closeall () at ldap/servers/slapd/plugin.c:1839
        pb = 0x7fd97ec008a0
        plugins_closed = 46
        index = 56
#10 0x00007fdc4b0002b9 in plugin_closeall (close_backends=close_backends@entry=1, close_globals=close_globals@entry=1) at ldap/servers/slapd/plugin.c:1932
#11 0x000055cdc4077789 in slapd_daemon (ports=0x7ffffd7d3a90) at ldap/servers/slapd/daemon.c:1460
        n_tcps = <optimized out>
        s_tcps = 0x7fdc467a8ff0
        i_unix = 0x7fdc465a8340
        fdesp = <optimized out>
        threads = 0
        in_referral_mode = 0
        connection_table_size = <optimized out>
#12 0x000055cdc40685f0 in main (argc=5, argv=0x7ffffd7d3cd8) at ldap/servers/slapd/main.c:1104
        return_value = 0
        mcfg = {extraname = 0x0, slapd_exemode = 1, n_port = 389, i_port = 1, s_port = 636, myname = 0x7fdc49836050 "ns-slapd", ldif_file = 0x0, ldif_files = 0, cmd_line_instance_name = 0x0, cmd_line_instance_names = 0x0, skip_db_protect_check = 0, db2ldif_include = 0x0, db2ldif_exclude = 0x0, ldif2db_removedupvals = 1, ldif2db_noattrindexes = 0, db2index_attrs = 0x0, ldif_printkey = 5, archive_name = 0x0, db2ldif_dump_replica = 0, db2ldif_dump_uniqueid = 1, ldif_include_changelog = 0, ldif2db_generate_uniqueid = 1, ldif2db_namespaceid = 0x0, importexport_encrypt = 0, upgradedb_flags = 0, upgradednformat_dryrun = 0, is_quiet = 0, backuptools_verbose = 0, dbverify_verbose = 0, dbverify_dbdir = 0x0}
        rc = <optimized out>
        sdn = 0x0
        thp_disable = <optimized out>
        slapdFrontendConfig = <optimized out>
        ports_info = {n_port = 389, s_port = 636, n_listenaddr = 0x0, s_listenaddr = 0x0, n_socket = 0x0, i_listenaddr = 0x0, i_port = 1, i_socket = 0x0, s_socket = 0x0}
        m = <optimized out>
        notify = <optimized out>
```

 - Platform: ALTLinux
 - Package and version: 389-ds-base 3.1.4

**Steps to Reproduce**
Happens randomly on service shutdown.

droideck:
Description: Prevent shutdown SIGSEGV by locking cache destruction and fixing pinned list pointers.

During shutdown cache_destroy_please() cleared caches without holding cache->c_mutex, allowing entrycache_flush() to race with other cache activity and dereference a NULL LRU tail.

Hold the cache lock while erasing caches.
Also fix pinned_remove() to clear the new head/tail prev/next links to avoid stale pointers.

Fixes: https://github.com/389ds/389-ds-base/issues/7162

Reviewed by: ?

## Summary by Sourcery

Prevent crashes during cache shutdown by synchronizing cache destruction and correcting pinned entry list maintenance.

Bug Fixes:
- Hold the cache mutex while destroying caches to prevent races that could dereference invalid LRU pointers during shutdown.
- Reset head and tail neighbor pointers when removing pinned entries to avoid stale list links and potential SIGSEGVs.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -458,11 +458,13 @@ static void
 lru_delete(struct cache *cache, void *ptr)
 {
     struct backcommon *e;
+
     if (NULL == ptr) {
         LOG("=> lru_delete\n<= lru_delete (null entry)\n");
         return;
     }
     e = (struct backcommon *)ptr;
+
 #ifdef LDAP_CACHE_DEBUG_LRU
     pinned_verify(cache, __LINE__);
     lru_verify(cache, e, 1);
@@ -475,8 +477,9 @@ lru_delete(struct cache *cache, void *ptr)
         e->ep_lrunext->ep_lruprev = e->ep_lruprev;
     else
         cache->c_lrutail = e->ep_lruprev;
-#ifdef LDAP_CACHE_DEBUG_LRU
+    /* Always clear pointers after removal to prevent stale pointer issues */
     e->ep_lrunext = e->ep_lruprev = NULL;
+#ifdef LDAP_CACHE_DEBUG_LRU
     lru_verify(cache, e, 0);
 #endif
 }
@@ -633,9 +636,14 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                 if (entry->ep_refcnt == 0) {
                     entry->ep_refcnt++;
                     if (entry->ep_state & ENTRY_STATE_PINNED) {
+                        /* Entry is in pinned list, not LRU - remove from pinned only.
+                         * pinned_remove clears lru pointers and won't add to LRU since refcnt > 0.
+                         */
                         pinned_remove(cache, laste);
+                    } else {
+                        /* Entry is in LRU list - remove from LRU */
+                        lru_delete(cache, laste);
                     }
-                    lru_delete(cache, laste);
                     if (type == ENTRY_CACHE) {
                         entrycache_remove_int(cache, laste);
                         entrycache_return(cache, (struct backentry **)&laste, PR_TRUE);
@@ -679,9 +687,14 @@ flush_hash(struct cache *cache, struct timespec *start_time, int32_t type)
                     if (entry->ep_refcnt == 0) {
                         entry->ep_refcnt++;
                         if (entry->ep_state & ENTRY_STATE_PINNED) {
+                            /* Entry is in pinned list, not LRU - remove from pinned only.
+                             * pinned_remove clears lru pointers and won't add to LRU since refcnt > 0.
+                             */
                             pinned_remove(cache, laste);
+                        } else {
+                            /* Entry is in LRU list - remove from LRU */
+                            lru_delete(cache, laste);
                         }
-                        lru_delete(cache, laste);
                         entrycache_remove_int(cache, laste);
                         entrycache_return(cache, (struct backentry **)&laste, PR_TRUE);
                     } else {
@@ -771,6 +784,11 @@ entrycache_flush(struct cache *cache)
             e = CACHE_LRU_TAIL(cache, struct backentry *);
         } else {
             e = BACK_LRU_PREV(e, struct backentry *);
+        }
+        if (e == NULL) {
+            slapi_log_err(SLAPI_LOG_WARNING, "entrycache_flush",
+                          "Unexpected NULL entry while flushing cache - LRU list may be corrupted\n");
+            break;
         }
         ASSERT(e->ep_refcnt == 0);
         e->ep_refcnt++;
@@ -1160,6 +1178,7 @@ pinned_remove(struct cache *cache, void *ptr)
 {
     struct backentry *e = (struct backentry *)ptr;
     ASSERT(e->ep_state & ENTRY_STATE_PINNED);
+
     cache->c_pinned_ctx->npinned--;
     cache->c_pinned_ctx->size -= e->ep_size;
     e->ep_state &= ~ENTRY_STATE_PINNED;
@@ -1172,13 +1191,23 @@ pinned_remove(struct cache *cache, void *ptr)
             cache->c_pinned_ctx->head = cache->c_pinned_ctx->tail = NULL;
         } else {
             cache->c_pinned_ctx->head = BACK_LRU_NEXT(e, struct backentry *);
+            /* Update new head's prev pointer to NULL */
+            if (cache->c_pinned_ctx->head) {
+                cache->c_pinned_ctx->head->ep_lruprev = NULL;
+            }
         }
     } else if (cache->c_pinned_ctx->tail == e) {
         cache->c_pinned_ctx->tail = BACK_LRU_PREV(e, struct backentry *);
+        /* Update new tail's next pointer to NULL */
+        if (cache->c_pinned_ctx->tail) {
+            cache->c_pinned_ctx->tail->ep_lrunext = NULL;
+        }
     } else {
+        /* Middle of list: update both neighbors to point to each other */
         BACK_LRU_PREV(e, struct backentry *)->ep_lrunext = BACK_LRU_NEXT(e, struct backcommon *);
         BACK_LRU_NEXT(e, struct backentry *)->ep_lruprev = BACK_LRU_PREV(e, struct backcommon *);
     }
+    /* Clear the removed entry's pointers */
     e->ep_lrunext = e->ep_lruprev = NULL;
     if (e->ep_refcnt == 0) {
         lru_add(cache, ptr);
@@ -1245,6 +1274,7 @@ pinned_add(struct cache *cache, void *ptr)
             return false;
     }
     /* Now it is time to insert the entry in the pinned list */
+
     cache->c_pinned_ctx->npinned++;
     cache->c_pinned_ctx->size += e->ep_size;
     e->ep_state |= ENTRY_STATE_PINNED;
@@ -1754,7 +1784,7 @@ entrycache_add_int(struct cache *cache, struct backentry *e, int state, struct b
                  * 3) ep_state: 0 && state: 0
                  *    ==> increase the refcnt
                  */
-                if (e->ep_refcnt == 0)
+                if (e->ep_refcnt == 0 && (e->ep_state & ENTRY_STATE_PINNED) == 0)
                     lru_delete(cache, (void *)e);
                 e->ep_refcnt++;
                 e->ep_state &= ~ENTRY_STATE_UNAVAILABLE;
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -849,8 +849,10 @@ erase_cache(struct cache *cache, int type)
 void
 cache_destroy_please(struct cache *cache, int type)
 {
+    cache_lock(cache);
```
progier389: I do not think that we need to lock the cache here because no other threads is supposed to access the cache at this point. (otherwise there will be a race condition between the thread acceding the cache and    PR_DestroyMonitor(cache->c_mutex);
In practice the backend is down so there should not be any pending operation on it ...  )

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1172,9 +1174,11 @@ pinned_remove(struct cache *cache, void *ptr)
             cache->c_pinned_ctx->head = cache->c_pinned_ctx->tail = NULL;
         } else {
             cache->c_pinned_ctx->head = BACK_LRU_NEXT(e, struct backentry *);
+            cache->c_pinned_ctx->head->ep_lruprev = NULL;
```
progier389: Redundant with  e->ep_lrunext = e->ep_lruprev = NULL;

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1172,9 +1174,11 @@ pinned_remove(struct cache *cache, void *ptr)
...
         }
     } else if (cache->c_pinned_ctx->tail == e) {
         cache->c_pinned_ctx->tail = BACK_LRU_PREV(e, struct backentry *);
+        cache->c_pinned_ctx->tail->ep_lrunext = NULL;
```
progier389: Redundant with  e->ep_lrunext = e->ep_lruprev = NULL;

droideck: IIUC, e->ep_lrunext = e->ep_lruprev = NULL; - clears the pointers of e (the entry being removed)

My fix - updates the pointers of the **neighboring** entry (the new head or new tail that remains in the list)...

ldap/servers/slapd/back-ldbm/cache.c
droideck: I reproduced and bisected the issue down to this commit https://github.com/389ds/389-ds-base/commit/15c63802ce78e2f1182c67fb648ce46920ede310 and I think the crash occurs because of this:

- The LRU flush algorithm traverses backwards via BACK_LRU_PREV() - that's how it's done in the commit. When reaching the head of the list, BACK_LRU_PREV() returns NULL (since head->ep_lruprev is NULL), but the code dereferences this pointer before checking.

- The erase_cache() function (I'm moving the lock there, BTW), called during shutdown via cache_destroy_please(), allowed the race in the intensive FreeIPA env at shutdown. My hypotheses:

   - Thread A (shutdown): Enters entrycache_flush(), gets entry E from LRU tail, checks "E == c_lruhead" which is FALSE
   - Thread B (worker): Performs cache_find_id(), calls lru_delete() on the head entry, setting c_lruhead to the next entry (or NULL if it was the only entry)
   - Thread A: Continues to next iteration, calls BACK_LRU_PREV(E) which returns the actual head entry H
   - Thread A: Processes H, checks "H == c_lruhead" - but c_lruhead was modified by Thread B, so the check fails
   - Thread A: Next iteration calls BACK_LRU_PREV(H) which returns  NULL (H is the real head), then crashes on NULL dereference

So adding cache_lock()/cache_unlock() to erase_cache() now prevents concurrent access during shutdown.

progier389: Adding lock does not fully prevent the crash risk is the above scenario:
  thread A may destroy the cache monitor (in cache_destroy_please)
  and thread B use it when calling cache_unlock
So there is still a race condition
   
The big question is why we do have a worker thread attempting to access the cache we are trying to destroy ?
The worker threads are supposed to be completed when plugin_closeall is called.

IMHO  there a problem with the active_threads count (i.e a thread is somehow not accounted. 
   probably not a worker thread but some plugin thread maybe ... )

ldap/servers/slapd/back-ldbm/cache.c
progier389: You are right: my mistake ! :face_holding_back_tears:

ldap/servers/slapd/back-ldbm/cache.c
droideck: Good point! You're right about the lock not fully solving the race with `PR_DestroyMonitor()`. I actually removed it - the other callers already hold the lock anyway.

droideck:
So the crash was at shutdown in `entrycache_flush()` but the corruption happened earlier... I hunted down the orphaned entries with debug messages and found a couple of issues:

First - when returning an existing entry via `alt`, the code checked `e->ep_refcnt` but called `lru_delete()` on `*alt`. Different entries! So it was removing entries that weren't even in the LRU possibly. There was also another call path in `entrycache_add_int()` missing the pinned state check.

Second - `flush_hash()` called both `pinned_remove()` and `lru_delete()` on pinned entries. `pinned_remove()` clears LRU pointers and won't add to LRU since refcnt > 0 in this place in `flush_hash()`. Also, I noticed `pinned_remove()` also wasn't updating the neighbor's pointers when removing head or tail entries from the pinned list.

These bugs sometimes (in a race manner) created orphaned entries with stale NULL pointers during normal operation. Then at shutdown, when `entrycache_flush()` walked backwards through the list, it hit one of these corrupted entries and crashed.

The NULL check is a safety net for shutdown - added the same to `dncache_flush()` as well. I also moved the pointer clearing in `lru_delete()` out of the debug-only block so it always runs (is it okay? I'm not sure why it was hidden in the debug block).

@progier389 please, review!

progier389 (approved):
Looks good.

I initially was tricked because pinned_remove could put the entry in the lru
 That was why both pinned_remove and lru_delete were called .
But this assumption is wrong because e->refcnt > 0) so the change is good

(But probably not the crash cause since there is no pinned entry by default)
The confusion between e and my_alt may be the reak root cause

vashirov:
I hit this crash again recently in https://github.com/389ds/389-ds-base/issues/7177, will check if this PR fixes it.

vashirov (approved):
Test passed, LGTM!

PR #7164, opened 2025-12-23
Backport issue 7012 to 389-ds-base-2.6.

abeisemb:
* Issue 7012 - improve dscrl dbverify result when backend does not exists

Improve error message when running dsctl instance dbverify wrongBeName To tell explicitly that the backend does not exists in bdb case To tell explicitly that dbverify is useless and always successfull in lmdb case

(cherry picked from commit 2c4dc8ecb2652452edcd47ed2295e67220ba2629. patch applies cleanly)

## Summary by Sourcery

Backport improved dbverify behavior and messaging for non-existent backends and LMDB backends to the 2.6 branch.

Bug Fixes:
- Ensure dbverify exits with an error and logs a clear message when run against a non-existent BDB backend.
- Clarify that dbverify is a no-op that always succeeds when using LMDB backends and log this behavior.
- Prevent test log handler leakage by removing the file handler during test cleanup.

Tests:
- Add a regression test that validates dbverify output and return codes for non-existent backends under BDB and LMDB configurations.

### DIFF
--- a/dirsrvtests/tests/suites/clu/dbverify_test.py
+++ b/dirsrvtests/tests/suites/clu/dbverify_test.py
@@ -31,6 +31,7 @@ def set_log_file(request):
 
     def fin():
         log.info('Delete log file')
+        log.removeHandler(fh)
         os.remove(LOG_FILE)
 
     request.addfinalizer(fin)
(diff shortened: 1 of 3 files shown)


tbordaz (approved):
LGTM
The typo suggested by sourcery is correct but I prefer the fix being the same in 2.x

Issue #7166, opened 2026-01-05
db_config_set asserts because of dynamic list


progier389:
**Issue Description**

dynamic list bring a regression in lib389 and cause CI Test failure:
dirsrvtests/tests/suites/config/config_test.py::test_lmdb_config FAILED  [ 82%]
...
>       set_and_check(inst, db_config, 'mdb_max_size', 'nsslapd-mdb-max-size', parse_size('2G'))
...    
    db_config_set(inst, db_config.dn, log, args)
...
>       if args.enable_dynamic_lists and args.disable_dynamic_lists:
           ^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'FakeArgs' object has no attribute 'enable_dynamic_lists'

IMHO db_config_set  should catch the AttributeError exception in that test.

**Package Version and Platform: **
 - Platform: Fedora
 - Package and version:main branch

**Steps to Reproduce**
1. See CI test results "config"

**Expected results**
No error  in db_config_set

mreynolds389:
This needs to get cherry picked to the 3.1 branch, thanks!

PR #7167, opened 2026-01-05
Issue 7166 - db_config_set asserts because of dynamic list

progier389:
Avoid assertion in db_config_set when args does not contains dynamic list attributes

Issue: #7166 

## Summary by Sourcery

Bug Fixes:
- Prevent db_config_set from asserting when dynamic list enable/disable options are not provided by treating them as optional arguments.


tbordaz (approved):
LGTM

PR #7171, opened 2026-01-07
Issue 6951 - Dynamic Certificate refresh phase 4 - Update lib389 and dsconf

jchapma:
Desciption:
Add the CertManager abstraction layer and DynamicCerts backend module. Enhance the NssSsl backend to support importing PKCS#12 containers, containing cert and private key. Update dsconf to support new PKCS#12-related args, allowing users to supply passwords via text, stdin, or file.

Fix:
- Introduce CertManager abstraction layer for uniform cert management.
- Implement DynamicCerts backend with add/list/delete operations.
- Extend NssSsl.add_cert to handle PKCS#12 files with passwords via text, stdin, or file.
- Update dsconf CLI to accept PKCS#12 arguments (--pkcs12-pin-text, --pkcs12-pin-stdin, --pkcs12-pin-path).
- Add tests for PEM, DER and PKCS#12 certificates covering both backends.

Relates: https://github.com/389ds/389-ds-base/issues/6951

Reviewed by:

## Summary by Sourcery

Introduce a certificate management abstraction that can use either the DynamicCerts LDAP backend or the NSS database, and extend CLI and SSL tooling to support PKCS#12 certificate imports.

New Features:
- Add CertManager abstraction to route certificate operations to DynamicCerts or NSS based on availability.
- Implement DynamicCerts backend to manage certificates via LDAP entries under the dynamic certificates subtree.
- Extend NssSsl certificate handling to support importing PKCS#12 containers with optional passwords and CA trust configuration.
- Enhance dsconf security certificate subcommands to work through CertManager and accept PKCS#12 inputs with multiple password sources (text, stdin, file).

Bug Fixes:
- Fix gentime_to_datetime to use the correct datetime namespace when parsing generalized time values.

Enhancements:
- Improve certificate existence checks, error handling, and logging for add, list, get, and delete operations in the certificate CLI.
- Add RSACertificate helper for generating and saving test certificates in multiple formats (PEM, DER, PKCS#12) for test scenarios.

Tests:
- Add integration tests covering PEM, DER, and PKCS#12 certificate add/list/delete flows for both DynamicCerts and NSS backends via CertManager and the dsconf CLI.


src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,304 @@
...
+    def _ensure_ldapi_connection(self):
+        """
+        Ensure LDAP connection is initialized and bound using SASL EXTERNAL.
+
+        Returns:
+            None
+        """
+        if self.conn is not None:
+            return
+
```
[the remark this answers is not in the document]
progier389: No. the above sourcery proposal is dangerous:
 Falling back to NssSsl when the server is on line is a bad idea that  will desynchronize the certificate db from the NSS caches in the server instance  (and there is no way to resynchronize except restarting the server which defeat the feature goal)
So in case of error the  ldapError exception should be propagated as initially proposed .

src/lib389/lib389/cert_manager.py
```diff
@@ -0,0 +1,206 @@
...
+
+class CertManager:
+    """
+    Certificate manager for 389 Directory Server.
+
+    Automatically selects DynamicCert backend if available via LDAPI,
+    otherwise falls back to NSS DB.
+    """
+
+    def __init__(self, dirsrv, backend=None):
```
progier389: backend is confusing should maybe use another term like: handler
And IMHO it should not be a parameter but directly initialized in __init__:
self.handler = DynamicCerts(dirsrv) if dirsrv.status() else NssSsl(dirsrv)

src/lib389/lib389/cli_conf/security.py
```diff
@@ -216,30 +233,35 @@ def security_disable_plaintext_port(inst, basedn, log, args, warn=True):
...
-        RSA(inst).set('nsSSLPersonalitySSL', args.name)
+    # Use CertManager to offload the request to DynamicCerts or NSS
+    certmgr = CertManager(dirsrv=inst)
 
-    # Add the cert
-    tlsdb.add_cert(args.name, args.file)
+    cert_exists = certmgr.get_cert_details(args.name)
+    if cert_exists:
+        log.info(f"Certificate '{args.name}' already exists, skipping")
+        return
```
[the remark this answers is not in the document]
progier389: Change look good as we we to ensure command idempotence as far as possible

src/lib389/lib389/cert_manager.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        Get certificate details by nickname/CN.
+
+        Returns:
+            dict containing certificate details (CN, subject, issuer, notBefore, notAfter, trustFlags)
+            None if certificate does not exist
+        """
+        try:
+            return self.backend.get_cert_details(nickname=nickname)
+        except ValueError as e:
```
progier389: IMHO sourcery is right. 
ValueError should not be handled at this level but directly at the handler level

src/lib389/lib389/cert_manager.py
```diff
@@ -0,0 +1,206 @@
...
+        Returns:
+            None
+        """
+        if not os.path.isfile(file_path):
+            raise ValueError(f"Certificate file not found: {file_path}")
+
+        if not nickname:
+            raise ValueError("Certificate nickname must not be empty")
+
+        if isinstance(self.backend, DynamicCerts):
```
progier389: IMHO you should not test the instance but rather have a sub function both in NssSsl and DynamicCerts in which you put the dedicated code

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        try:
+            self._ensure_ldapi_connection()
+            self.conn.search_s(DYNCERT_SUFFIX, ldap.SCOPE_ONELEVEL,
+                               "(objectClass=*)", attrlist=[DYCATTR_CN])
+            return True
+        except ldap.NO_SUCH_OBJECT:
+            log.warning(f"DynamicCert base DN {DYNCERT_SUFFIX} does not exist")
+            return False
+        except ldap.LDAPError as e:
```
progier389, line 93: Should probably raise an exception  if ldap.NO_SUCH_OBJECT or LDAPError  (after logging an error message) If the server is online we do not want to switch to NssSsl (which is dangerous)

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,359 @@
...
+DYCATTR_HAS_PKEY    = DYCATTR_PREFIX + "hasprivatekey"
+DYCATTR_TRUST       = DYCATTR_PREFIX + "trustflags"
+DYCATTR_TYPE        = DYCATTR_PREFIX + "type"
+DYCATTR_TOKEN       = DYCATTR_PREFIX + "tokenname"
+DYCATTR_KALGO       = DYCATTR_PREFIX + "keyalgorithm"
+DYCATTR_NOTBEFORE   = DYCATTR_PREFIX + "notbefore"
+DYCATTR_NOTAFTER    = DYCATTR_PREFIX + "notafter"
+DYCATTR_FORCE	    = DYCATTR_PREFIX + "Force"
+
+class DynamicCerts:
```
progier389: I wonder if we should not rather have:
class  DynamicCert(DSLdapObject) :
and class  DynamicCerts(DSLdapObjects) :
 to reuse the standard entry framework
   to list/retrieve a certificate (acceding attribute with entry.get_attr_val_utf8(attr) )

And extend them if needed.

progier389 (changes requested):
Some changes are needed around CertManager and Dyncerts

src/lib389/lib389/cli_conf/security.py
```diff
@@ -504,11 +549,19 @@ def create_parser(subparsers):
     cert_add_parser = certs_sub.add_parser('add', help='Add a server certificate', description=(
         'Add a server certificate to the NSS database'))
     cert_add_parser.add_argument('--file', required=True,
-        help='Sets the file name of the certificate')
+        help='Sets the file pkcs12 name of the certificate')
+    cert_add_parser.add_argument('--pkcs12-pin-text', required=False,
+        help='The pkcs12 password')
+    cert_add_parser.add_argument('--pkcs12-pin-stdin', required=False,
+        help='The pkcs12 password')
```
mreynolds389: Can you elaborate on the help text here?  They probably shouldn't both be the same :-)

src/lib389/lib389/dyncerts.py
jchapma: I agree with you, let me look into this

jchapma:
Ok, so I was having issues with CI tests so reverted them for now, I validate with dsconf for all cert operations. I will add CI tests soon, but as other work is dependant on this work I pushed the non CI test changes.

src/lib389/lib389/cert_manager.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __init__(self, instance):
+        """
+        Initialise a CertManager object, which selects the appropriate
+        certificate handler (DynamicCert or NSS).
+
+        :param instance: DirSrv instance
+        :raises ValueError: If instance is None
+        """
+        if instance is None:
```
progier389: should better be: 
if not isinstance(instance, DirSrv):

src/lib389/lib389/cert_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        try:
+            dyncerts = DynamicCerts(instance=self.dirsrv)
+            if dyncerts.is_online():
+                self.cert_handler = dyncerts
+                log.debug("Using DynamicCert backend")
+            else:
+                raise RuntimeError("DynamicCert backend not accessible")
+        except Exception as e:
+            log.debug(f"DynamicCert backend unavailable ({e}), falling back to NSS")
+            self.cert_handler = NssSsl(dirsrv=self.dirsrv)
```
progier389, line 43: NO: CERTIFICATE DB SHOULD NOT BE CHANGED WHEN THE INSTANCE IS UP

 NssSsl should not be called in case of problem.
 It should be called only if the instance is down:
    
if self.dirsrv.status():
    self.cert_handler = DynamicCerts(instance=self.dirsrv)
else:
    self.cert_handler = NssSsl(dirsrv=self.dirsrv)

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,395 @@
...
+            return raw
+
+    def is_ca(self):
+        """
+        Determine if this certificate is a CA.
+
+        :return: True if certificate is a CA, else False
+        """
+        # Check trust flags first
+        trust_flags = self.get_attr_val_utf8(DYCATTR_TRUST)
```
progier389: not sure that trust flag are the most reliable to check for CA.
(maybe using the cert_is_ca function in util.py)

```

File: src/lib389/lib389/dyncerts.py [hunk 12]
```diff
@@ -0,0 +1,395 @@
...
+        :param instance: DirSrv instance
+        :param dn: Entry distinguished name (DN)
+        """
+        super(DynamicCert, self).__init__(instance, dn)
+        self._rdn_attribute = DYCATTR_CN
+        self._create_objectclasses = ["top", "extensibleObject"]
+        self._protected = False
+        self._basedn = DYNCERT_SUFFIX
+
+    def _normalise_timestamp(self, raw: str):
```
progier389, line 61: typo: normalize  (normalise is the french word :wink: )

jchapma: > typo: normalize (normalise is the french word 😉 )

We use that here too :)

src/lib389/lib389/cert_manager.py
jchapma: Ahh, so as long as the instance is running we are good to go ? Then we dont need the is_online check ?

src/lib389/lib389/dyncerts.py
jchapma: Yes I agree, its not ideal. I added DER blob support to cert_is_ca(), which is fine when we are adding a cert as we have the cert file in the NSS case or the cert DER blob in the dyncert case. But when we list certs we dont, we just get a list of child entries, no cert file or cert blob. This is the reason we use a certs trust flags, like NssSsl.

progier389: It is relatively easy to get the der:
 offline:  certutil -L -d certdir -n nickname -r
 online:  get dsDynamicCertificateDer value from the object in the list
 maybe we should have a cert_and_der_list that return a list of couples (nickname, der)

src/lib389/lib389/cert_manager.py
progier389: Exactly. Anyway as the code is aligned to lib389, it should always work and if it does not for some reason, it seems better to abort the command and let the user knows that there is a problem

src/lib389/lib389/dyncerts.py
jchapma: Hmm, when i try access the cert object during listing, I get 

**INFO: JC list_certs cert cn=SERVER_CERT_PEM,cn=dynamiccertificates der length: 0**

[snip]
        cert_objects = self.list()
        certs = []
        for cert in cert_objects:
            #JC test get cert der
            if cert._dn != self._basedn:
                der_list = cert.get_attr_vals_bytes(DYCATTR_CERTDER)
                der_bytes = der_list[0] if der_list else None
                **log.info(f"JC list_certs cert:{cert._dn} der length: {len(der_bytes) if der_bytes else 0}")**

[19/Jan/2026:13:31:11.634987751 +0000] conn=115 op=5 SRCH base="cn=SERVER_CERT_PEM,cn=dynamiccertificates" scope=0 filter="(objectClass=*)" attrs="dsdynamiccertificateder"
[19/Jan/2026:13:31:11.643515349 +0000] conn=115 op=5 RESULT err=0 tag=101 nentries=1 wtime=0.000175585 optime=0.008529738 etime=0.008704572
[snip]

It looks like i am adding the cert correctly:

[19/Jan/2026:13:31:11.617178499 +0000] conn=115 op=4 ADD dn="cn=SERVER_CERT_PEM,cn=dynamiccertificates"
[19/Jan/2026:13:31:11.634527343 +0000] conn=115 op=4 RESULT err=0 tag=105 nentries=0 wtime=0.000166210 optime=0.017355309 etime=0.017520674

        dn = f"cn={nickname},{self._basedn}"
        attrs = {
            "cn": [nickname.encode()],
            "objectClass": [b"top", b"extensibleObject"],
            DYCATTR_CERTDER: [ der_cert, ]
        }
[snip]
        cert_obj = self.get_cert_obj(nickname)
        if not cert_obj:
            cert_obj = DynamicCert(self._instance, dn)
            cert_obj.create(properties=attrs)
        else:
            attrs_list = [(attr, vals) for attr, vals in attrs.items()]
            cert_obj.replace_many(*attrs_list)

progier389: You are right, there is bug in dyncert code:
```
ldapsearch -Q -LLL -Y EXTERNAL -H ldapi://%2fhome%2fprogier%2fsb%2f389%2ftst%2fci-install%2fvar%2frun%2fslapd-supplier1.socket -b cn=dynamiccertificates dsdynamiccertificateder
dn: cn=dynamiccertificates

dn: cn=Self-Signed-CA,cn=dynamiccertificates
dsdynamiccertificateder:

dn: cn=Server-Cert,cn=dynamiccertificates
dsdynamiccertificateder:
```
Will open an issue to fix it.

jchapma: Ok, I will make the suggested changes here.

jchapma: That bug exposed a bug I had in cert_is_ca, working now

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,398 @@
...
+        if not nickname:
+            raise ValueError("Certificate CN cannot be empty")
+        try:
+            cert = self.get(nickname)
+        except ldap.NO_SUCH_OBJECT:
+            return None
+
+        return cert
+
+    def get_cert_details(self, nickname: str):
```
droideck, line 414: Not all `get_cert_details()` usages are fixes. For example `_lint_certificate_expiration` will break.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1142,26 +1141,33 @@ def get_cert_details(self, nickname):
...
+                    "issuer": issuer,
+                    "expires": str(end_date),
+                    "trust_flags": trust_flags,
+                }
 
-        # Did not find cert with that name
-        raise ValueError("Certificate '{}' not found in NSS database".format(nickname))
+        return None
 
     def list_certs(self, ca=False):
```
droideck, line 1154: IIUC, `security_enable()` will break because it uses old format -  `RSA(inst).set('nsSSLPersonalitySSL', certs[0][0])`

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +273,93 @@ def cert_add(inst, basedn, log, args):
...
-        log.info('Security database does not exist. Creating a new one in {}.'.format(inst.get_cert_dir()))
-        tls.reinit()
-
-    tls.add_ca_cert_bundle(args.file, args.name)
+    # Does it make sense to add a CA cert from p12 container ?
+    if args.file.lower().endswith((".p12", ".pfx")):
+        raise ValueError("PKCS#12 CA certificates not supported. Use PEM or DER file")
 
+    certmgr = CertManager(instance=inst)
+    certmgr.add_ca_cert(args.file, args.name, args.force)
```
droideck: Do we still support bundles? I think this function doesn't unless I miss something

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +273,93 @@ def cert_add(inst, basedn, log, args):
...
-    if args.json:
-        log.info(json.dumps(cert_list, indent=4))
+    certmgr = CertManager(instance=inst)
+    certs = certmgr.list_certs()
+    if not certs:
+        log.info("No certificates found.")
+        return
 
+    for cert in certs:
+        _dump_cert(cert, json_output=args.json)
```
droideck: I wonder is UI will function correctly here (or if we have a regression) -- for cert_list/cacert_list -- `_dump_cert` prints a single "certificate" object and is called once per cert. With --json, this yields multiple top-level JSON objects, not a single JSON list as before.

src/lib389/lib389/utils.py
```diff
@@ -1993,28 +1996,93 @@ def check_cert_info(cert_file_name, search_text):
...
+            with open(cert_file, "rb") as f:
+                data = f.read()
+        except ValueError as ve:
+            raise ValueError(f"Unable to load certificate '{cert_data}': {ve}")
+
+    try:
+        # p12
+        if cert_file and cert_file.lower().endswith((".p12", ".pfx")):
+            if pkcs12_password is None:
+                raise ValueError("Password required for PKCS#12 file")
```
droideck: Hmm, this prevents importing PKCS#12 files with empty passwords (which are valid, I think). The `dyncerts.py:202` and `nss_ssl.py:1219-1220` handle this correctly by defaulting to empty string though.

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,41 +1176,107 @@ def list_client_ca_certs(self):
...
+                    raise ValueError(f"Failed to set CA trust for {nickname} {e}")
+
+            return
+
+        pem_file = cert_file.lower().endswith(".pem")
+        if pem_file:
+            self._assert_not_chain(cert_file)
+
+        is_ca_cert = cert_is_ca(cert_file)
+        if ca and not is_ca_cert:
```
droideck: It's a duplicate -- previously in the same function there's already the check -- `nss_ssl.py:1195-1210`

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,398 @@
...
+
+        :return: A list of certificate dictionaries for each certificate
+        """
+        cert_objects = self.list()
+        certs = []
+        for cert in cert_objects:
+            if cert._dn == self._basedn:
+                continue
+            der_cert = cert.get_attr_vals_bytes(DYCATTR_CERTDER)[0]
+            log.info(f"JC der_cert len:{len(der_cert)}")
```
droideck: debug statement:)

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,41 +1176,107 @@ def list_client_ca_certs(self):
             if self._rsa_cert_is_caclienttrust(cert)
         ]
 
-    def add_cert(self, nickname, input_file, ca=False):
+    def add_cert(self, nickname: str, cert_file: str, ca=False,
+                 pkcs12_password=None, force: bool = False):
```
droideck: Both - this function and `add_ca_cert()` - they use `force` but actually it's not used. It's for a future?

droideck (changes requested):

src/lib389/lib389/dyncerts.py
jchapma: I missed that one, thanks

src/lib389/lib389/cli_conf/security.py
jchapma: We decided not to support CA cert bundles for dyncerts, but, it is something we might add in the future.

src/lib389/lib389/cli_conf/security.py
jchapma: You are right, the UI expects a list of cert objects, fixed

src/lib389/lib389/utils.py
jchapma: Yes are correct, this check is too strict. I will use the dyncerts.py method of import a p12 file, with or without password.

src/lib389/lib389/nss_ssl.py
jchapma: In the case where the dyncerts instance is offline we use NssSsl, but IIUC there is no force option for importing a cert into the NSS DB.

progier389 (commented):
Looks good but there are still some issues:
   - dsconf supplier1 security ca-certificate list
   - dsconf supplier1 security certificate list
 lists twice the certificates:

dsconf supplier1 security ca-certificate list
Certificate Name: Self-Signed-CA
Subject DN: CN=ssca.389ds.example.com,O=testing,L=389ds,ST=Queensland,C=AU
Issuer DN: CN=ssca.389ds.example.com,O=testing,L=389ds,ST=Queensland,C=AU
Expires: 2028-01-29 14:03:05
Trust Flags: CT,,

Certificate Name: Self-Signed-CA
Subject DN: CN=ssca.389ds.example.com,O=testing,L=389ds,ST=Queensland,C=AU
Issuer DN: CN=ssca.389ds.example.com,O=testing,L=389ds,ST=Queensland,C=AU
Expires: 2028-01-29 14:03:05
Trust Flags: CT,,

Certificate Name: NewCA
Subject DN: C=FR,O=test-ECDSA-CA,CN=ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com
Issuer DN: C=FR,O=test-ECDSA-CA,CN=ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com
Expires: 2036-01-27 14:49:25
Trust Flags: CT,,

Certificate Name: NewCA
Subject DN: C=FR,O=test-ECDSA-CA,CN=ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com
Issuer DN: C=FR,O=test-ECDSA-CA,CN=ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com
Expires: 2036-01-27 14:49:25
Trust Flags: CT,,

progier389 (approved):
LGTM

src/lib389/lib389/nss_ssl.py
```diff
@@ -1170,41 +1176,105 @@ def list_client_ca_certs(self):
...
+                "pk12util",
+                "-v",
+                "-n", nickname,
+                "-i", cert_file,
+                "-d", self._certdb,
+                "-k", f"{self._certdb}/{PWD_TXT}",
+                "-W", pkcs12_password,
+            ]
+
+            self.log.info(f"nss import p12 cmd: {format_cmd_list(cmd)}", )
```
droideck: I think we should mask the password here (now it's printed plain text in the logs).

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +270,107 @@ def cert_add(inst, basedn, log, args):
...
-        )
-    else:
-        log.info('Certificate Name: {}'.format(details[0]))
-        log.info('Subject DN: {}'.format(details[1]))
-        log.info('Issuer DN: {}'.format(details[2]))
-        log.info('Expires: {}'.format(details[3]))
-        log.info('Trust Flags: {}'.format(details[4]))
+        output = _dump_cert(cert, json_output=args.json)
+        print(json.dumps(output, indent=4))
+    _dump_cert(cert, json_output=args.json)
```
droideck: When it's JSON, it'll call it twice (same in `cacert_get`)

Also I think it'll be silent in non-JSON mode (no `log.info`). Is it intentional?

src/lib389/lib389/cli_conf/security.py
```diff
@@ -529,19 +563,20 @@ def create_parser(subparsers):
...
     # CA certificate management
     cacerts = security_sub.add_parser('ca-certificate', help='Manage TLS certificate authorities', formatter_class=CustomHelpFormatter)
     cacerts_sub = cacerts.add_subparsers(help='ca-certificate')
     cacert_add_parser = cacerts_sub.add_parser('add', help='Add a Certificate Authority', description=(
-        'Add a Certificate Authority to the NSS database'))
+        'Add a CA certificate (PEM or DER only) to the NSS database or DynamicCerts backend.'))
     cacert_add_parser.add_argument('--file', required=True,
-        help='Sets the file name of the CA certificate')
-    cacert_add_parser.add_argument('--name', nargs='+', required=True,
         help='Sets the name/nickname of the CA certificate, if adding a PEM bundle then specify multiple names one for '
```
droideck: I think it's a wrong docstring typo as it says `Sets the name/nickname` while it's under `--file`.

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +270,107 @@ def cert_add(inst, basedn, log, args):
...
-            )
-        )
-    else:
-        log.info('Certificate Name: {}'.format(details[0]))
-        log.info('Subject DN: {}'.format(details[1]))
-        log.info('Issuer DN: {}'.format(details[2]))
-        log.info('Expires: {}'.format(details[3]))
-        log.info('Trust Flags: {}'.format(details[4]))
+        output = _dump_cert(cert, json_output=args.json)
+        print(json.dumps(output, indent=4))
```
droideck: Why it was changed to `print`? I think we should use `log.info`.

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,455 @@
...
+        if ca:
+            attrs[DYCATTR_TRUST] = [b"CT,,"]
+
+        if force:
+            attrs[DYCATTR_FORCE] = [b"TRUE"]
+
+        if not der_cert:
+            raise ValueError(f"Failed to extract DER bytes from {cert_file}")
+
+        dn = f"cn={nickname},{self._basedn}"
```
droideck: I think we should use `ldap.dn.escape_dn_chars()` here for safety.

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,455 @@
...
+    def _normalise_timestamp(self, raw: str):
+        """
+        Convert DynamicCert timestamp to NSS like format.
+
+        :param raw: Raw DynamicCert timestamp (e.g. 20260109181934Z)
+        :return: Formatted timestamp as "YYYY-MM-DD HH:MM:SS", or original string
+        """
+        try:
+            year   = int(raw[0:4])
+            month  = int(raw[4:6]) + 1  # PRExplodedTime.tm_month is 0–11
```
droideck, line 70: One of the places where I think we'll need to have test coverage. I worry that there can be cases where the dates are shifted by 1 month.

src/lib389/lib389/utils.py
```diff
@@ -1993,28 +1996,92 @@ def check_cert_info(cert_file_name, search_text):
...
+    cert_file = None
+    if isinstance(cert_data, (bytes, bytearray)):
+        data = bytes(cert_data)
+    else:
+        cert_file = cert_data
+        try:
+            with open(cert_file, "rb") as f:
+                data = f.read()
+        except ValueError as ve:
+            raise ValueError(f"Unable to load certificate '{cert_data}': {ve}")
```
droideck: IIRC, `open()` raises FileNotFoundError/IOError, not ValueError (also, OSError?)

src/lib389/lib389/cli_conf/security.py
```diff
@@ -412,18 +440,20 @@ def csr_del(inst, basedn, log, args):
...
-    tlsdb = NssSsl(dirsrv=inst)
-    tlsdb.del_cert(args.name)
-    log.info(f"Successfully deleted certificate")
-
+    certmgr = CertManager(instance=inst)
+    try:
+        certmgr.del_cert(args.name)
+        log.info(f"Successfully deleted certificate")
+    except ValueError as e:
+        log.info(f"Failed to delete certificate '{args.name}': {e}")
```
droideck: I think this should be a `log.error()`.

droideck (changes requested):
Additional test coverage would be helpful - a few areas caught my eye that might benefit from closer review, though it's hard to say for certain without tests to explore them further.
Since this touches a sensitive security area, I think we should consider expanding the test suite before releasing this feature. Of course, as a separate PR. 

Nice to cover areas, IMO:
  - Invalid/malformed certificate files (corrupted PEM, truncated DER, invalid PKCS#12)
  - PKCS#12 with wrong password, missing password, empty password
  - Overwriting existing certificates (with and without --force)
  - Backend switching and consistency behavior (DynamicCerts vs NssSsl)
  - JSON vs text output consistency across all list/get commands
  - DynamicCerts timestamp normalization

src/lib389/lib389/dyncerts.py
jchapma: dyncerts timestamps use 0 based months:
dsdynamiccertificatenotbefore: 20260006181934Z
dsdynamiccertificatenotafter: 20270006181934Z

This is why we month  +1 in _normalise_timestamp, without it, January would be month 0, which is invalid for datetime.

src/lib389/lib389/cli_conf/security.py
```diff
@@ -140,6 +142,54 @@ def add_parser(action, value):
...
+            f"Certificate Name: {cert['cn']}\n"
+            f"Subject DN: {cert['subject']}\n"
+            f"Issuer DN: {cert['issuer']}\n"
+            f"Expires: {cert['expires']}\n"
+            f"Trust Flags: {cert['trust_flags']}\n"
+        )
+        if log:
+            log.info(msg)
+        else:
+            print(msg)
```
droideck: I am not sure I understand the purpose here...

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +272,109 @@ def cert_add(inst, basedn, log, args):
...
-        log.info(json.dumps(cert_list, indent=4))
+    certmgr = CertManager(instance=inst)
+    certs = certmgr.list_certs()
+    if not certs:
+        log.info("No certificates found.")
+        return
 
+    if args.json:
+        output = [_dump_cert(cert, json_output=True) for cert in certs]
+        print(json.dumps(output, indent=4))
```
droideck: I think it should be `log.info()` as it was before.

src/lib389/lib389/cli_conf/security.py
```diff
@@ -222,127 +272,109 @@ def cert_add(inst, basedn, log, args):
...
-        log.info(json.dumps(cert_list, indent=4))
+    certmgr = CertManager(instance=inst)
+    ca_certs = certmgr.list_ca_certs()
+    if not ca_certs:
+        log.info("No CA certificates found.")
+        return
 
+    if args.json:
+        output = [_dump_cert(cert, json_output=True) for cert in ca_certs]
+        print(json.dumps(output, indent=4))
```
droideck: Same here.

src/lib389/lib389/cli_conf/security.py
```diff
@@ -509,6 +543,10 @@ def create_parser(subparsers):
         help='Sets the name/nickname of the certificate')
     cert_add_parser.add_argument('--primary-cert', action='store_true',
                                  help="Sets this certificate as the server's certificate")
+    cert_add_parser.add_argument('--pkcs12-pin-text', help='The PKCS#12 password as plain text')
```
droideck: I think it's worth to add an additional warning here about the fact that the password will be displayed in `ps aux` and shell history. So the warning should advise to use stdin and file options which are more secure.

src/lib389/lib389/dyncerts.py
```diff
@@ -0,0 +1,458 @@
...
+
+            if cert is None:
+                raise ValueError("PKCS#12 file contains no certificate")
+
+            der_cert = cert.public_bytes(serialization.Encoding.DER)
+            if privkey is not None:
+                der_privkey = privkey.private_bytes(
+                    encoding=serialization.Encoding.DER,
+                    format=serialization.PrivateFormat.PKCS8,
+                    encryption_algorithm=serialization.NoEncryption()
```
droideck, line 189: We might need to document that the private keys are stored unencrypted in LDAP. So anyone with read access to cn=dynamiccertificates can extract private keys (so it should be protected by ACI always). Or/and we can add encryption. Otherwise, I'll leave it up to you to decide the approach here. :)

src/lib389/lib389/utils.py
```diff
@@ -1993,28 +1996,92 @@ def check_cert_info(cert_file_name, search_text):
...
+                data = f.read()
+        except OSError as e:
+            raise ValueError(f"Unable to load certificate '{cert_data}': {e}")
+
+    try:
+        # p12
+        if cert_file and cert_file.lower().endswith((".p12", ".pfx")):
+            try:
+                privkey, cert, _ = pkcs12.load_key_and_certificates(
+                    data, pkcs12_password.encode() if pkcs12_password else None,
```
droideck: The function's type hint says `bytes` but `encode()` will fail if called on `bytes`. Consider using `ensure_bytes()` here.

droideck (approved):
Besides a few small issues - LGTM!

src/lib389/lib389/cli_conf/security.py
jchapma: True, good catch.

src/lib389/lib389/dyncerts.py
jchapma: The entries under cn=dynamiccertificates store only certificate metadata (subject, issuer, validity, trust flags, DER encoded cert) and indicate whether a private key exists. The actual private keys are stored in the NSS DB, protected by the token/passwd, so we dont expose unencrypted private keys. Access to these entries should still be protected via ACI. Do you feel some documentation to reflect this is required ?

src/lib389/lib389/utils.py
jchapma: Nice catch

src/lib389/lib389/cli_conf/security.py
jchapma: I think this was used in case we don't pass in a logger, but we do in each case so its redundant.

src/lib389/lib389/cli_conf/security.py
jchapma: It should :)

src/lib389/lib389/dyncerts.py
droideck: Sounds good! Yes, IMO, a brief docstring in dyncerts.py explaining this architecture would help prevent confusion -- the Python code appears to store private keys via DYCATTR_PKEYDER, but understanding that it's a transport mechanism (with NSS DB as the actual secure storage) isn't directly obvious... A 2-3 line module docstring noting that private keys remain in NSS and only metadata lives in LDAP would be sufficient, IMO.

And yeah, the metadata still can be sensitive and should be protected with ACI, true.

src/lib389/lib389/cli_conf/security.py
droideck: > I think this was used in case we don't pass in a logger, but we do in each case so its redundant.

Okay, I initially thought that the `log` parameter was about `if set -- we log the info, else -- we don't`.

For out CLI and general lib389 purposes, I see no reason to use `print()` at all, and it breaks logging and output processing for some cases.

mreynolds389:
There doesn't seem to be any validation for the file path either

```
# dsconf -j ldapi://%2fvar%2frun%2fslapd-localhost.socket security certificate add --name=test --file=/etc/dirsrv/slapd-localhost/replica.pem --pkcs12-pin-path=/does/not/exist
Successfully added certificate
```

src/lib389/lib389/dyncerts.py
jchapma: > Sounds good! Yes, IMO, a brief docstring in dyncerts.py explaining this architecture would help prevent confusion -- the Python code appears to store private keys via DYCATTR_PKEYDER, but understanding that it's a transport mechanism (with NSS DB as the actual secure storage) isn't directly obvious... A 2-3 line module docstring noting that private keys remain in NSS and only metadata lives in LDAP would be sufficient, IMO.
> 
> And yeah, the metadata still can be sensitive and should be protected with ACI, true.

Great, I will add the docstring the next time I need to update code in that area, ok ?

progier389: IMHO it will be pretty soon: :wink: 
  We have a problem with ML-DSA certificate in PKCS12 file because python cryptography module does not yet support them.  I am looking if there is a way to work around the issue by using openssl command

PR #7180, opened 2026-01-10
Issue 7172 - (2nd) Index ordering mismatch after upgrade

### LINKED ISSUE #7172: Index ordering mismatch after upgrade
**Issue Description**
Commit daf731f55071d45eaf403a52b63d35f4e699ff28 introduced a regression. After upgrading to a version that adds `integerOrderingMatch` matching rule to `parentid` and `ancestorid` indexes, searches may return empty or incorrect results.

This happens because the existing index data was created with lexicographic ordering, but the new compare function expects integer ordering. Index lookups fail because the compare function doesn't match the data ordering.
The root cause is that `ldbm_instance_create_default_indexes()` calls `attr_index_config()` unconditionally for `parentid` and `ancestorid` indexes, which triggers `ainfo_dup()` to overwrite `ai_key_cmp_fn` on existing indexes. This breaks indexes that were created without the `integerOrderingMatch` matching rule.

 - Platform: Fedora/RHEL
 - Package and version: all versions containing fix https://github.com/389ds/389-ds-base/issues/6928

**Steps to Reproduce**
Steps to reproduce the behavior: https://bugzilla.redhat.com/show_bug.cgi?id=2424132#c0

**Expected results**
No index ordering mismatch after upgrade. Instructions to reindex should be provided to the user, but the server should use the existing index ordering.

vashirov:
Commit 742c12e0247ab64e87da000a4de2f3e5c99044ab introduced a regression where the check to skip creating parentid/ancestorid indexes if they already exist was incorrect.
The `ainfo_get()` function falls back to returning LDBM_PSEUDO_ATTR_DEFAULT attrinfo when the requested attribute is not found.
Since LDBM_PSEUDO_ATTR_DEFAULT is created before the ancestorid check, `ainfo_get()` returns LDBM_PSEUDO_ATTR_DEFAULT instead of NULL, causing the ancestorid index creation to be skipped entirely.

When operations later try to use the ancestorid index, they fall back to LDBM_PSEUDO_ATTR_DEFAULT, and attempting to open the .default dbi mid-transaction fails with MDB_NOTFOUND (-30798).

Fix Description:
Instead of just checking if `ainfo_get()` returns non-NULL, verify that the returned attrinfo is actually for the requested attribute.

Fixes: https://github.com/389ds/389-ds-base/issues/7172

## Summary by Sourcery

Bug Fixes:
- Correct the detection of existing parentid and ancestorid attrinfo so that fallback .default attrinfo does not incorrectly suppress index creation.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/instance.c
+++ b/ldap/servers/slapd/back-ldbm/instance.c
@@ -190,7 +190,7 @@ ldbm_instance_create_default_indexes(backend *be)
     char *ancestorid_indexes_limit = NULL;
     char *parentid_indexes_limit = NULL;
     struct attrinfo *ai = NULL;
-    struct attrinfo *index_already_configured = NULL;
+    int index_already_configured = 0;
     struct index_idlistsizeinfo *iter;
     int cookie;
     int limit;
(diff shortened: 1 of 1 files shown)


tbordaz (approved):
Shame on me. The initial PR was correct to check the type matching and my suggestion introduce this failure.
The fix looks good to me

Issue #6758, opened 2025-04-29
Fix failing webUI tests


mirielka:
**Issue Description**
Fix webUI tests that are failing due to recent changes.

```
FAILED dirsrvtests/tests/suites/webui/database/database_test.py::test_dictionary_check_checkbox[firefox] - playwright._impl._errors.TimeoutError: Locator.click: Timeout 65000ms exceeded.
FAILED dirsrvtests/tests/suites/webui/ldap_browser/ldap_browser_test.py::test_create_and_delete_user[firefox] - playwright._impl._errors.TimeoutError: Locator.click: Timeout 65000ms exceeded.
FAILED dirsrvtests/tests/suites/webui/ldap_browser/ldap_browser_test.py::test_create_and_delete_custom_entry[firefox] - playwright._impl._errors.TimeoutError: Locator.click: Timeout 65000ms exceeded.
FAILED dirsrvtests/tests/suites/webui/monitoring/monitoring_test.py::test_server_statistics_visibility[firefox] - playwright._impl._errors.Error: Locator.is_visible: Error: strict mode violation: get_by_role("button", name="Refresh") res...
FAILED dirsrvtests/tests/suites/webui/monitoring/monitoring_test.py::test_logging_visibility[firefox] - playwright._impl._errors.TimeoutError: Locator.wait_for: Timeout 65000ms exceeded.
```

PR #7182, opened 2026-01-12
Issue 6758 - Fix WebUI plugin visibility tests

aadhikar:
Description:
Use full accessible text with exact=True for plugin nav item selection to ensure consistent behavior across browsers.

Relates: #6758

## Summary by Sourcery

Align WebUI plugin navigation items with stable OUIA identifiers and update visibility tests to target these identifiers for consistent cross-browser behavior.

Enhancements:
- Add OUIA IDs to WebUI plugin navigation items to provide stable selectors for automated tooling.

Tests:
- Update WebUI plugin visibility tests to select plugins via OUIA-based locators instead of visible text.


droideck (approved):
Looks good!

One minor thing, do you know why the rest of the tests in the test suite are writted with the strange `get_by_text` calls like `frame.get_by_text('Plugin is enabledManaged Entries')`?

Can we also rewrite them similarly? (if original code's `id`s are inconsistent - we can fix them too, I think)

Issue #7184, opened 2026-01-12
argparse.HelpFormatter _format_actions_usage() is deprecated


mreynolds389:
**Issue Description**
 
_format_actions_usage() has been removed in python 3.15

PR #7185, opened 2026-01-12
Issue 7184 - argparse.HelpFormatter _format_actions_usage() is deprecated

mreynolds389:
Description:

_format_actions_usage() was removed in python 3.15. Instead we can use _get_actions_usage_parts() but it also behaves differently between python 3.14 and 3.15 so we need special handling.

Relates: https://github.com/389ds/389-ds-base/issues/7184

## Summary by Sourcery

Bug Fixes:
- Handle argparse.HelpFormatter usage formatting correctly across Python 3.14 and 3.15 by switching from the deprecated _format_actions_usage() to _get_actions_usage_parts().

### DIFF
--- a/src/lib389/lib389/cli_base/__init__.py
+++ b/src/lib389/lib389/cli_base/__init__.py
@@ -413,7 +413,20 @@ class CustomHelpFormatter(argparse.HelpFormatter):
 
     def _format_usage(self, usage, actions, groups, prefix):
         usage = super(CustomHelpFormatter, self)._format_usage(usage, actions, groups, prefix)
-        formatted_options = self._format_actions_usage(parent_arguments, [])
+
+        if sys.version_info < (3, 13):
+            # Use _format_actions_usage() for Python 3.12 and earlier
+            formatted_options = self._format_actions_usage(parent_arguments, [])
+        else:
+            # Use _get_actions_usage_parts() for Python 3.13 and later
+            action_parts = self._get_actions_usage_parts(parent_arguments, [])
+            if sys.version_info >= (3, 15):
+                # Python 3.15 returns a tuple (list of actions, count of actions)
+                formatted_options = ' '.join(action_parts[0])
+            else:
+                # Python 3.13 and 3.14 return a list of actions
+                formatted_options = ' '.join(action_parts)
+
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/cli_base/__init__.py
```diff
@@ -413,7 +413,14 @@ def add_arguments(self, actions):
 
     def _format_usage(self, usage, actions, groups, prefix):
         usage = super(CustomHelpFormatter, self)._format_usage(usage, actions, groups, prefix)
-        formatted_options = self._format_actions_usage(parent_arguments, [])
+        action_parts = self._get_actions_usage_parts(parent_arguments, [])
```
droideck: As of now, we support `Python >=3.8`. But with this change we'll limit out support with `Python >=3.13` as `_get_actions_usage_parts` is not in these early version.

Should we process something like this?

Python 3.8-3.12: Use _format_actions_usage() (still available)
Python 3.13: Use _get_actions_usage_parts() returning a list
Python 3.15+: Use _get_actions_usage_parts() returning a tuple

src/lib389/lib389/cli_base/__init__.py
```diff
@@ -413,7 +413,20 @@ def add_arguments(self, actions):
...
+            # Use _format_actions_usage() for Python 3.12 and earlier
+            formatted_options = self._format_actions_usage(parent_arguments, [])
+        else:
+            # Use _get_actions_usage_parts() for Python 3.13 and later
+            action_parts = self._get_actions_usage_parts(parent_arguments, [])
+            if sys.version_info >= (3, 15):
+                # Python 3.15 returns a tuple (list of actions, count of actions)
+                formatted_options = ' '.join(action_parts[0])
+            else:
+                # Python 3.14 and earlier returns a list of actions
```
droideck: Very minor nitpick :D 

```suggestion
                # Python 3.13 and 3.14 return a list of actions
```

droideck (approved):
LGTM!

src/lib389/lib389/cli_base/__init__.py
mreynolds389: I'll add this in and merge, thanks!

PR #7192, opened 2026-01-16
Issue 7189 - DSBLE0007 generates incorrect remediation commands for s…

### LINKED ISSUE #7189: DSBLE0007 generates incorrect remediation commands for scan limits
**Issue Description**
```
[1] DS Lint Error: DSBLE0007
--------------------------------------------------------------------------------
Severity: HIGH
Check: backends:changelog:system_indexes
Affects:
 -- cn=changelog

Details:
-----------
System indexes are essential for proper directory server operation. Missing or
incorrectly configured system indexes can lead to poor search performance, replication
issues, and other operational problems.

The following system indexes should be present with correct configuration:
- entryrdn: index type 'subtree'
- parentid: index type 'eq' with matching rule 'integerOrderingMatch'
- ancestorid: index type 'eq' with matching rule 'integerOrderingMatch'
- objectClass: index type 'eq'
- aci: index type 'pres'
- nscpEntryDN: index type 'eq'
- nsUniqueId: index type 'eq'
- nsds5ReplConflict: index types 'eq', 'pres'
- nsCertSubjectDN: index type 'eq'
- numsubordinates: index type 'pres'
- nsTombstoneCSN: index type 'eq'
- targetuniqueid: index type 'eq'
- changeNumber: index type 'eq' with matching rule 'integerOrderingMatch'
- entryusn: index type 'eq' with matching rule 'integerOrderingMatch'

Current discrepancies:
- Index parentid missing matching rule: integerOrderingMatch
- Index parentid missing fine grain definition of IDs limit: integerOrderingMatch


Resolution:
-----------
Add the missing system indexes or fix the incorrect configurations using dsconf:

    # dsconf EXAMPLE-TEST backend index set changelog --attr parentid --add-mr integerOrderingMatch
    # dsconf EXAMPLE-TEST backend index set changelog --attr parentid --add-mr integerOrderingMatch --add-scanlimit limit=5000 type=eq flags=AND

After adding or modifying indexes, you may need to reindex the affected attributes:

    # dsconf EXAMPLE-TEST backend index reindex changelog --attr parentid

WARNING: Reindexing can be resource-intensive and may impact server performance on a live system.
Consider scheduling reindexing during maintenance windows or periods of low activity. For production
systems, you may want to reindex offline or use the --wait option to monitor task completion.
```

This line should display missing scan limits, but instead it shows MR:
```
- Index parentid missing fine grain definition of IDs limit: integerOrderingMatch
```

And the second remediation step should not contain `--add-mr integerOrderingMatch` and scan limits must be in quotes.

vashirov:
…can limits

Bug Description:

The generated dsconf commands for fixing missing system indexes had two issues:

1. The --add-scanlimit value was not quoted, causing the shell to interpret "limit=5000 type=eq flags=AND" as multiple arguments instead of a single value, resulting in "unrecognized arguments: type=eq flags=AND" error.

2. When both matching rule and scanlimit were missing, two separate commands were generated where the second would fail because the matching rule was already added by the first command.

Fix Description:

1. Quote the scanlimit value in all remediation commands

2. Combine matching rule and scanlimit fixes into a single command when both are missing for the same index instead of expected_scanlimit)

Fixes: https://github.com/389ds/389-ds-base/issues/7189

## Summary by Sourcery

Fix healthcheck remediation for missing system index configuration and add regression tests.

Bug Fixes:
- Ensure generated dsconf remediation commands always quote scanlimit values to avoid shell argument parsing errors.
- Generate a single combined remediation command when both matching rule and scanlimit are missing for the same index instead of separate conflicting commands.

Tests:
- Add coverage for missing scanlimit remediation to verify quoted scanlimit values in generated commands.
- Add coverage for scenarios where both matching rule and scanlimit are missing to ensure a single combined remediation command is produced.

### DIFF
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -712,7 +712,7 @@ class Backend(DSLdapObject):
                             missing_mr = True
 
                     missing_scanlimit = False
-                    if (attr_name.lower() == "parentid") and expected_scanlimit and (len(actual_scanlimit) == 0):
+                    if expected_scanlimit and (len(actual_scanlimit) == 0):
                         discrepancies.append(f"Index {attr_name} missing fine grain definition of IDs limit: {expected_scanlimit}")
                         missing_scanlimit = True
 
--- a/src/lib389/lib389/backend.py
+++ b/src/lib389/lib389/backend.py
@@ -678,7 +678,7 @@ class Backend(DSLdapObject):
                     if expected_config.get('matching_rule'):
                         cmd += f" --matching-rule {expected_config['matching_rule']}"
                     if expected_config.get('scanlimit'):
-                        cmd += f" --add-scanlimit {expected_config['scanlimit']}"
+                        cmd += f" --add-scanlimit \"{expected_config['scanlimit']}\""
                     remediation_commands.append(cmd)
                     reindex_attrs.add(attr_name)  # New index needs reindexing
                 else:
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM

src/lib389/lib389/backend.py
```diff
@@ -700,28 +700,31 @@ def _lint_system_indexes(self):
...
-                            # Add the missing matching rule
-                            cmd = f"dsconf YOUR_INSTANCE backend index set {bename} --attr {attr_name} --add-mr {expected_mr}"
-                            remediation_commands.append(cmd)
-                            reindex_attrs.add(attr_name)
+                            missing_mr = True
 
-                    # Check fine grain definitions for parentid ONLY
-                    expected_scanlimit = expected_config.get('scanlimit')
+                    missing_scanlimit = False
                     if (attr_name.lower() == "parentid") and expected_scanlimit and (len(actual_scanlimit) == 0):
```
droideck: I think we also miss `ancestorid` here.
But if we `expected_scanlimit`, then it covers the both cases so the hardcoded name check `attr_name.lower() == "parentid")` is probably unnecessary.

droideck (approved):
Besides that, looks good!

PR #7204, opened 2026-01-22
Issue 7201 - Syscall overhead in LMDB import writer thread

### LINKED ISSUE #7201: Syscall overhead in LMDB import writer thread
**Issue Description**
I noticed that offline and online import with LMDB are slower than with BDB:
```
[21/Jan/2026:14:12:00.578654329 -0500] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 292000 entries -- average rate 3518.1/sec, recent rate 3259.5/sec,

```
vs.
```
[21/Jan/2026:13:00:19.552791877 -0500] - INFO - bdb_import_monitor_threads - import userroot: Processed 2097996 entries -- average rate 4540.9/sec, recent rate 4303.9/sec, hit ratio 100%
```

`perf` shows that we spend too much time calling `clock_gettime()`:
```
Samples: 210K of event 'task-clock:ppp', Event count (approx.): 52735250000
Overhead  Command   Shared Object         Symbol
  16.90%  ns-slapd  [vdso]                [.] __cvdso_clock_gettime_data.constprop.0
   4.60%  ns-slapd  [kernel.kallsyms]     [k] _raw_spin_unlock_irqrestore
   4.47%  ns-slapd  [kernel.kallsyms]     [k] syscall_enter_from_user_mode
   3.86%  ns-slapd  liblmdb.so.0.0.0      [.] mdb_mid2l_search
   2.57%  ns-slapd  libc.so.6             [.] _int_free
   2.51%  ns-slapd  libc.so.6             [.] _int_malloc
   2.16%  ns-slapd  libc.so.6             [.] __GI___lll_lock_wake
   2.14%  ns-slapd  libc.so.6             [.] pthread_rwlock_unlock@@GLIBC_2.34
   1.99%  ns-slapd  libc.so.6             [.] __GI___lll_lock_wait
   1.60%  ns-slapd  libc.so.6             [.] pthread_rwlock_rdlock@GLIBC_2.2.5
...
```

```
# perf stat -e 'syscalls:sys_enter_clock_gettime' -p $(pgrep -f 'ns-slapd.*ldif2db') -- sleep 30

 Performance counter stats for process id '4800':

          18970701      syscalls:sys_enter_clock_gettime

      30.001978211 seconds time elapsed

```
`clock_gettime()` is called by MDB_STAT_STEP macro to collect perf stats, but it's called on every operation inside the writer loop:
https://github.com/389ds/389-ds-base/blob/6ce33b1bedd2cd13d7e6544692354715f6e613b8/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c#L3933-L3944

It's 3 syscalls per write or 6000 syscalls per transaction (with 2000 as the default batch size).

If I move the stats collection to the transaction boundary, the number of syscalls drops dramatically:

```
# perf stat -e 'syscalls:sys_enter_clock_gettime' -p $(pgrep -f 'ns-slapd.*ldif2db') -- sleep 30

 Performance counter stats for process id '5234':

             30491      syscalls:sys_enter_clock_gettime

      30.001895860 seconds time elapsed
```

And offline import rate is on par with BDB:
```
[21/Jan/2026:17:00:36.572466952 -0500] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 2460000 entries -- average rate 4685.7/sec, recent rate 4425.0/sec,
```

And the online import rate is even better:
```
[21/Jan/2026:17:09:20.814987400 -0500] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 2507728 entries -- average rate 8924.3/sec, recent rate 8599.5/sec,
[21/Jan/2026:17:09:25.839091929 -0500] - INFO - dbmdb_import_monitor_threads - import userroot: Import writer thread usage: run: 87.77% read: 9.54% write: 0.32% pause: 0.20% txnbegin: 0.00% txncommit: 2.17%
```

And if I switch CLOCK_THREAD_CPUTIME_ID to CLOCK_MONOTONIC in stats collection, we have 0 syscalls, since it uses VDSO:
```
# perf stat -e 'syscalls:sys_enter_clock_gettime' -p $(pgrep -f 'ns-slapd.*ldif2db') -- sleep 30

 Performance counter stats for process id '5939':

                 0      syscalls:sys_enter_clock_gettime

      30.001239923 seconds time elapsed

```

We can also make MDB perf stats optional and disabled by default, since they create such big perf overhead.

vashirov:
Bug Description:
ldif2db import is slower with LMDB than with BDB (~3500 vs ~4500 entries/s) due to 2 issues:
1. The MDB_STAT_STEP macro calls `clock_gettime()` to collect performance statistics. This was called on every single operation inside the writer loop, resulting in ~3 syscalls per write or ~6000 syscalls per transaction (with 2000 as the default batch size).

2. In `dbmdb_import_workerq_push()`, after copying work to a worker slot, the condition variable was never signaled. This caused workers to spend up to 100ms in `safe_cond_wait()` before checking for new work, severely limiting import throughput.

Fix Description:
1. Add a new config parameter `nsslapd-mdb-import-stats` (default: off) to control whether performance statistics collection is enabled.
3. Add `pthread_cond_broadcast()` immediately after copying data to wake the workers.

After applying these fixes ldif2db import rate is about ~10000 entries/s.

Fixes: https://github.com/389ds/389-ds-base/issues/7201

## Summary by Sourcery

Optimize LMDB import writer throughput by reducing per-operation stats collection and ensuring worker threads are promptly notified of new work.

Bug Fixes:
- Signal/broadcast the worker queue condition variable after enqueuing work items so worker threads wake immediately to process new tasks.
- Move LMDB import statistics accounting from each individual write operation to transaction boundaries to avoid excessive syscall overhead.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
@@ -59,9 +59,9 @@
 
 
 /* import thread usage statistics */
-#define MDB_STAT_INIT(stats)    { mdb_stat_collect(&stats, MDB_STAT_RUN, 1); }
-#define MDB_STAT_END(stats)     { mdb_stat_collect(&stats, MDB_STAT_RUN, 0); }
-#define MDB_STAT_STEP(stats, step)    { mdb_stat_collect(&stats, (step), 0); }
+#define MDB_STAT_INIT(stats, enabled)    { if (enabled) mdb_stat_collect(&stats, MDB_STAT_RUN, 1); }
+#define MDB_STAT_END(stats, enabled)     { if (enabled) mdb_stat_collect(&stats, MDB_STAT_RUN, 0); }
+#define MDB_STAT_STEP(stats, step, enabled)    { if (enabled) mdb_stat_collect(&stats, (step), 0); }
 
 typedef enum {
     MDB_STAT_RUN,
@@ -424,6 +424,7 @@ dbmdb_import_workerq_push(ImportQueue_t *q, WorkerQueueData_t *data)
         return -1;
     }
     dbmdb_dup_worker_slot(q, data, slot);
+    pthread_cond_broadcast(&q->cv);
     pthread_mutex_unlock(&q->mutex);
     return 0;
 }
@@ -3914,14 +3915,15 @@ dbmdb_import_writer(void*param)
     int count = 0;
     int rc = 0;
     mdb_stat_info_t stats = {0};
+    int stats_enabled = ctx->ctx->dsecfg.import_stats;
 
-    MDB_STAT_INIT(stats);
+    MDB_STAT_INIT(stats, stats_enabled);
     while (!rc && !info_is_finished(info)) {
-        MDB_STAT_STEP(stats, MDB_STAT_PAUSE);
+        MDB_STAT_STEP(stats, MDB_STAT_PAUSE, stats_enabled);
         wait_for_starting(info);
-        MDB_STAT_STEP(stats, MDB_STAT_READ);
+        MDB_STAT_STEP(stats, MDB_STAT_READ, stats_enabled);
         slot = dbmdb_import_q_getall(&ctx->writerq);
-        MDB_STAT_STEP(stats, MDB_STAT_RUN);
+        MDB_STAT_STEP(stats, MDB_STAT_RUN, stats_enabled);
         if (info_is_finished(info)) {
             dbmdb_import_q_flush(&ctx->writerq);
             break;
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -3936,13 +3937,12 @@ dbmdb_import_writer(void*param)
                 rc = TXN_BEGIN(ctx->ctx->env, NULL, 0, &txn);
             }
             if (!rc) {
-                MDB_STAT_STEP(stats, MDB_STAT_WRITE);
```
progier389, line 3939: Sorry but this defeat the interest of having stats:
  the stats logic work by increments (to determine how much time in each steps spend but now there is a single state that englobe all 
So IMHO at this point we have better to find a way to active them conditionally and do not provide stats by defaults
(Maybe by checking if SLAPI_LOG_TIMING log level is set ? )

vashirov: Fair enough. I was thinking of adding another parameter under `cn=mdb,cn=config,cn=ldbm database,cn=plugins,cn=config` to enable/disable import stats with off by default. SLAPI_LOG_TIMING can also work.

progier389: Adding new parameter has its merit too: 
 Could add some other later on (like controlling  on lmdb debug traces (( rather than having to rebuild after changing the defines )

vashirov: Ok, added `nsslapd-mdb-import-stats`

progier389 (approved):
LGTM

PR #7205, opened 2026-01-22
Issue 7096 - (2nd) During replication online total init the function …

### LINKED ISSUE #7096: During replication online total init the function idl_id_is_in_idlist is not scaling with large database
**Issue Description**
Since https://github.com/389ds/389-ds-base/issues/1815 the oneline init creates the candidate list with a range search '(parentid>=1)'. Before adding an entry to the candidate list it checks that the parent entry is already added, so that total init will send parents before children.
The checking is done with idl_id_is_in_idlist that does not scale. Looking like the total init never start.

 - 1.3 and after

**Steps to Reproduce**
will be described later

**Expected results**
Total init should not take too long to create the candidate list

vashirov:
…idl_id_is_in_idlist is not scaling with large database

Bug Description:
The fix for #7096 optimized the BDB backend's `idl_new_range_fetch()` function to use ID ranges instead of checking the full ID list during online total initialization. However, the LMDB backend's `idl_lmdb_range_fetch()` function and its callback `idl_range_add_id_cb()` were not updated and still use the non-scaling `idl_id_is_in_idlist()` function.

Fix Description:
Apply the same optimization to the LMDB backend.

Fixes: https://github.com/389ds/389-ds-base/issues/7096

## Summary by Sourcery

Optimize LMDB ID range fetching to scale with large databases by aligning it with the range-based approach used in the BDB backend.

Bug Fixes:
- Avoid non-scaling parent-ID lookups during LMDB online total initialization by replacing full ID list scans with range-based membership checks.

Enhancements:
- Track ID ranges alongside ID lists during LMDB range fetch operations and clean up associated range state after processing leftovers.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/idl_new.c
+++ b/ldap/servers/slapd/back-ldbm/idl_new.c
@@ -66,6 +66,7 @@ typedef struct {
     size_t leftoverlen;
     size_t leftovercnt;
     IDList *idl;
+    IdRange_t *idrange_list;
     int flag_err;
     ID lastid;
     ID suffix;
@@ -716,7 +717,6 @@ static int
 idl_range_add_id_cb(dbi_val_t *key, dbi_val_t *data, void *ctx)
 {
     idl_range_ctx_t *rctx = ctx;
-    int idl_rc = 0;
     ID id = 0;
 
     if (key->data == NULL) {
@@ -779,10 +779,12 @@ idl_range_add_id_cb(dbi_val_t *key, dbi_val_t *data, void *ctx)
              * found entry is the one from the suffix
              */
             rctx->suffix = keyval;
-            idl_rc = idl_append_extend(&rctx->idl, id);
-        } else if ((keyval == rctx->suffix) || idl_id_is_in_idlist(rctx->idl, keyval)) {
+            idl_append_extend(&rctx->idl, id);
+            idrange_add_id(&rctx->idrange_list, id);
+        } else if ((keyval == rctx->suffix) || idl_id_is_in_idlist_ranges(rctx->idl, rctx->idrange_list, keyval)) {
             /* the parent is the suffix or already in idl. */
-            idl_rc = idl_append_extend(&rctx->idl, id);
+            idl_append_extend(&rctx->idl, id);
+            idrange_add_id(&rctx->idrange_list, id);
         } else {
             /* Otherwise, keep the {keyval,id} in leftover array */
             if (!rctx->leftover) {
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -966,21 +960,17 @@ idl_lmdb_range_fetch(
...
-                    }
+                    idl_append_extend(&idl_range_ctx.idl, idl_range_ctx.leftover[i].id);
+                    idrange_add_id(&idl_range_ctx.idrange_list, idl_range_ctx.leftover[i].id);
                     idl_range_ctx.leftover[i].key = 0;
                     remaining--;
                 }
             }
         }
         slapi_ch_free((void **)&idl_range_ctx.leftover);
+        idrange_free(&idl_range_ctx.idrange_list);
```
tbordaz: I know that idl_range_ctx.idrange_list is only allocated/used when SLAPI_OP_RANGE_NO_IDL_SORT so the call here is valid.
However, should we make the call systematically (whatever the flag), and move it outside of the `if (operator&SLAPI_OP_RANGE_NO_IDL_SORT) {`

ldap/servers/slapd/back-ldbm/idl_new.c
```diff
@@ -841,7 +836,6 @@ idl_lmdb_range_fetch(
 {
     int ret = 0;
     int ret2 = 0;
-    int idl_rc = 0;
     dbi_cursor_t cursor = {0};
     back_txn s_txn;
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
```
tbordaz, line 841: What about setting 'idl_range_ctx.idrange_list = NULL' in the initialization of idl_range_ctx.

tbordaz (approved):
Thanks for fixing this path. LGTM.
Added comments.

ldap/servers/slapd/back-ldbm/idl_new.c
vashirov: We already initialize the struct with
```
idl_range_ctx_t idl_range_ctx = {0};
```
This should set `idl_range_ctx.idrange_list` to NULL. I can add it for clarity.

ldap/servers/slapd/back-ldbm/idl_new.c
vashirov: I added it here for consistency with the previous `slapi_ch_free` call. Since both handle NULL, we can move them both outside of `if`. Let me know how you want to proceed.

droideck (commented):
Hmm, you removed `idl_append_extend()`’s return handling (e.g., idl_rc checks) in idl_new.c. Could you please give details a bit on that point?
IIUC, `idl_append_extend()` can still return ENOMEM, and without checks it would be silent and leave a partial IDL, isn't?

droideck (approved):

ldap/servers/slapd/back-ldbm/idl_new.c
tbordaz: I tend to prefer that the slapi_ch_free is independent of the SLAPI_OP_RANGE_NO_IDL_SORT.

ldap/servers/slapd/back-ldbm/idl_new.c
tbordaz: I agree it is useless but the initialization is setting all fields of callback context, including setting them to NULL. So yes I also tend to prefer to add it just for clarity.

Issue #7206, opened 2026-01-22
Should log whether TLS key is PQC or not


progier389:
**Is your feature request related to a problem? Please describe.**
Access log shows the cipher, key name and eventually the certificates issuer names but it does not provide any hint about whether PQC algorithm are used or not.
We should log a clear hint in access log that PQC is used on a TLS connection.

**Describe the solution you'd like**
My idea is to append [PQC] to the cipher name if keaGroup is one of KEM groups and keaType one of the hybrid  types.:
[22/Jan/2026:18:53:36.395363959 +0100] conn=1 TLS1.3 128-bit AES-GCM[PQC]; client C=FR,O=client-test-ML-DSA,CN=localhost; issuer C=FR,O=test-ML-DSA-CA,CN=localhost

**Describe alternatives you've considered**
I thought about adding a new value. (And it may perhaps still be worth to do that for json log
but for legacy log format, adding new values may break some tools 

I also thought about logging directly the group name but that would be yet another obtuse value
Having a simple PQC flag seems more significant for most users

PR #7207, opened 2026-01-23
Issue 7206 - Should log whether TLS key is PQC or not

progier389:
Append [PQC] to the cipher name when logging SSL/TLS if the Key Exchange is one of the KEM group
If connection debug level is enabled, logs in error log the keaType and keaGroup (as integer) 

Issue: #7206 

Reviewed by:  @tbordaz, @droideck  (Thanks!)

## Summary by Sourcery

Log TLS key exchange PQC usage and tag PQC cipher suites in connection logging.

New Features:
- Append a [PQC] marker to the logged cipher name when the TLS key exchange uses a PQC KEM group.
- Log TLS key exchange keaType and keaGroup values at connection debug level for TLS handshakes.

Enhancements:
- Add helper to detect PQC-capable key exchanges based on NSS SSL channel info.
- Use the correct unsigned type for NSS algorithm policy flags when configuring PQC algorithm policy.

Tests:
- Extend the MLDSA TLS test to enable detailed error log debugging for PQC-related logging.

### DIFF
--- a/ldap/servers/slapd/auth.c
+++ b/ldap/servers/slapd/auth.c
@@ -411,7 +411,7 @@ check_pqc(uint64_t connid, SSLChannelInfo *sci)
     slapi_log_err(SLAPI_LOG_CONNS, "check_pqc", "conn=%" PRIu64 " TLS keaType=%d keaGroup=%d\n",
                   connid, sci->keaType, sci->keaGroup);
 #ifdef MAX_ML_DSA_PRIVATE_KEY_LEN
-    /* NSS supports PQC (because of the ifdef). No let check if the connection use it */
+    /* NSS supports PQC (because of the ifdef). Now lets check if the connection uses it */
     /* Check that PQC KeaType is hybrid */
     switch (sci->keaType) {
         case ssl_kea_ecdh_hybrid:
--- a/ldap/servers/slapd/auth.c
+++ b/ldap/servers/slapd/auth.c
@@ -396,14 +396,14 @@ handle_bad_certificate(void *clientData, PRFileDesc *prfd)
 
 
 /*
- * Determine if the connection key exchange is Post Quantum Cryptology aware.
+ * Determine if the connection key exchange is Post Quantum Cryptography aware.
  * This function may need to evolve with NSS as more PQC methods get supported.
  */
 static bool
-check_pqc(int connid, SSLChannelInfo *sci)
+check_pqc(uint64_t connid, SSLChannelInfo *sci)
 {
     /*
-     * FYI: To interprets the values:
+     * FYI: To interpret the values:
      * KeaType and KeaGroup values are defined in
      * https://github.com/nss-dev/nss/blob/master/lib/ssl/sslt.h
      * Respectively in SSLKEAType and SSLNamedGroup enums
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/auth.c
```diff
@@ -454,7 +492,12 @@ handle_handshake_done(PRFileDesc *prfd, void *clientData)
     }
 
     keySize = cipherInfo.effectiveKeyBits;
-    cipher = slapi_ch_strdup(cipherInfo.symCipherName);
+    pqc = check_pqc(conn->c_connid, &channelInfo);
+    if (pqc) {
+        cipher = slapi_ch_smprintf("%s[PQC]", cipherInfo.symCipherName);
```
tbordaz, line 497: Should check  cipherInfo.symCipherName != NULL ?

tbordaz (approved):
LGTM. a question

ldap/servers/slapd/auth.c
progier389: According to SSL_GetCipherSuiteInfo NSS code, it cannot be null if slapd_ssl_getCipherSuiteInfo succeed
IMHO that was why the old code was not testing it either.

ldap/servers/slapd/auth.c
```diff
@@ -395,6 +395,43 @@ handle_bad_certificate(void *clientData, PRFileDesc *prfd)
...
+    /*
+     * FYI: To interpret the values:
+     * KeaType and KeaGroup values are defined in
+     * https://github.com/nss-dev/nss/blob/master/lib/ssl/sslt.h
+     * Respectively in SSLKEAType and SSLNamedGroup enums
+     */
+    slapi_log_err(SLAPI_LOG_CONNS, "check_pqc", "conn=%" PRIu64 " TLS keaType=%d keaGroup=%d\n",
+                  connid, sci->keaType, sci->keaGroup);
+#ifdef MAX_ML_DSA_PRIVATE_KEY_LEN
+    /* NSS supports PQC (because of the ifdef). No let check if the connection use it */
```
droideck:
```suggestion
    /* NSS supports PQC (because of the ifdef). Now lets check if the connection uses it */
```

ldap/servers/slapd/auth.c
```diff
@@ -395,6 +395,43 @@ handle_bad_certificate(void *clientData, PRFileDesc *prfd)
 }
 
 
+/*
+ * Determine if the connection key exchange is Post Quantum Cryptography aware.
+ * This function may need to evolve with NSS as more PQC methods get supported.
+ */
+static bool
+check_pqc(uint64_t connid, SSLChannelInfo *sci)
```
droideck, line 403: `check_pqc` is compiled only if MAX_ML_DSA_PRIVATE_KEY_LEN is defined.
It's probably unlikely, but If at some point in the future, NSS supports PQC KEM groups but not ML‑DSA, this will silently disable PQC tagging (false negatives). Should we have a macro tied to PQC KEM support or a more direct feature test.

droideck (approved):
Looks good!
Besides the mentioned points, -- could you please add a test that checks that `[PQC]` is in the logs?

ldap/servers/slapd/auth.c
progier389: I would rather keep the MAX_ML_DSA_PRIVATE_KEY_LEN to be consistent with the other PQC changes,
the main goal of these #ifdef is to be able to build on older versions (with NSS that does not supports PQC)
It is also used to protect the ML-DSA key types and policies so the code will probably need to be modified when ML-DSA support will be dropped ( Anyway at this time there will not be any more NSS version that does not support PQC and we could remove these ifdef

PR #7220, opened 2026-02-03
Issue 7076 - Fix revert_cache() never called in modrdn

aadhikar:
Description: The postentry check in PR #7077 was broken - postentry is always NULL at that point, fixed by removing the check.

Relates: #7076

Reviewed by: @vashirov

## Summary by Sourcery

Bug Fixes:
- Fix revert_cache() not being called during parent modrdn operations when BETXN callbacks fail due to an incorrect postentry check.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
@@ -102,6 +102,7 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
     Connection *pb_conn = NULL;
     int32_t parent_op = 0;
     int32_t betxn_callback_fails = 0; /* if a BETXN fails we need to revert entry cache */
+    int32_t cache_mod_phase = 0; /* set when we reach the cache modification phase */
     struct timespec parent_time;
     Slapi_Mods *smods_add_rdn = NULL;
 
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1363,12 +1363,8 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
     }
 
-    /* Revert the caches if this is the parent operation and cache modifications were made.
-     * Cache modifications (via modify_switch_entries) only happen after BETXN PRE plugins succeed,
-     * so we should only revert if we got past that point (i.e., BETXN POST plugin failures).
-     * For BETXN PRE failures, no cache modifications were made to parent/newparent entries.
-     */
-    if (parent_op && betxn_callback_fails && postentry) {
+    /* Revert the caches if this is the parent operation */
+    if (parent_op && betxn_callback_fails) {
```
mreynolds389: This change makes me nervous because the comments specifically state that you only need to revert the cache if you hit the postop phase (where postentry is not NULL).  Did you verify if a postop plugin fails that postentry is still NULL?

aadhikar: @mreynolds389 Valid concern - fixed. Added `cache_mod_phase` flag so `revert_cache()` only runs after we enter the cache modification phase.

@vashirov  Your `postentry` concern handled using the flag instead.

mreynolds389 (approved):

droideck (approved):

vashirov (approved):
LGTM!

PR #7225, opened 2026-02-05
Issue 7224 - CI Test - Simplify test_reserve_descriptor_validation

### LINKED ISSUE #7224: CI Test - Simplify test_reserve_descriptor_validation
**Issue Description**
Previously, the test_reserve_descriptor_validation CI test calculated the expected number of file descriptors based on backends, indexes, SSL/FIPS mode, and compared it to the value returned by the server. This approach is fragile, especially in FIPS mode.

The test has been updated to simply verify that the server corrects the configured nsslapd-reservedescriptors value if it is set too low, instead of calculating the expected total.

 - Package and version: [main branch back]

**Steps to Reproduce**
1. Enable FIPS mode
2. sudo pytest -s -v dirsrvtests/tests/suites/resource_limits/fdlimits_test.py::test_reserve_descriptor_validatio
3. Notice test failures

**Expected results**
Test should pass in FIPS and non FIPS mode

**Additional context**
The underlying method that validates nsslapd-reservedescriptors will be improved in a later commit.

jchapma:
Description:
Previously, the test_reserve_descriptor_validation CItest calculated the expected number of file descriptors based on backends, indexes, SSL/FIPS mode, and compared it to the value returned by the server. This approach is fragile, especially in FIPS mode.

Fix:
The test has been updated to simply verify that the server corrects the configured nsslapd-reservedescriptors value if it is set too low, instead of calculating the expected total.

Fixes: https://github.com/389ds/389-ds-base/issues/7224

Reviewed by:

## Summary by Sourcery

Tests:
- Adjust test_reserve_descriptor_validation to set a low reserved descriptor value, restart the instance, and assert the corrected value is higher rather than matching a computed constant.

### DIFF
--- a/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
+++ b/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
@@ -27,7 +27,7 @@ RESRV_FD_ATTR = "nsslapd-reservedescriptors"
 GLOBAL_LIMIT = resource.getrlimit(resource.RLIMIT_NOFILE)[1]
 SYSTEMD_LIMIT = ensure_str(check_output("systemctl show -p LimitNOFILE dirsrv@standalone1".split(" ")).strip()).split('=')[1]
 CUSTOM_VAL = str(int(SYSTEMD_LIMIT) - 10)
-RESRV_DESC_VAL = str(10)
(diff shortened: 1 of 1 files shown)


bsimonova (approved):
LGTM

PR #7234, opened 2026-02-09
Issue 7233 - test_produce_division_by_zero fails with IsADirectoryErr…

### LINKED ISSUE #7233: test_produce_division_by_zero fails with IsADirectoryError in conftest.py log collection
**Problem:**
Running disk_monitoring tests with HTML report fails with:
`IsADirectoryError: [Errno 21] Is a directory: '/var/log/dirsrv/slapd-standalone1/tmp'`

**Cause:**
`conftest.py` `pytest_runtest_makereport` hook uses `glob('/*/*')` which matches both files and directories, then tries to `open()` them.

**Reproducer:**
DISK_MONITORING_ACK=1 pytest-3 -v --html=report.html dirsrvtests/tests/suites/disk_monitoring/

aadhikar:
…or in conftest.py

Description: glob('/*/*') matches directories causing open() to fail.

Fixes: #7233

Reviewed by: ???

## Summary by Sourcery

Guard test log collection and cleanup to prevent directory-related errors during disk monitoring tests.

Bug Fixes:
- Skip non-regular files when iterating over log glob matches to avoid IsADirectoryError when opening paths.
- Ensure dummy mount temporary directory is removed after unmounting to avoid leftover directories interfering with tests.

### DIFF
--- a/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
+++ b/dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
@@ -31,15 +31,16 @@ def create_dummy_mount(topology_st, request):
     log.info('Create dummy mount')
     for cmd in cmds:
         log.info('Command used : %s' % cmd)
-        subprocess.Popen(cmd, shell=True)
+        subprocess.run(cmd, shell=True)
 
     def fin():
         cmds = ['umount /var/log/dirsrv/slapd-{}/tmp'.format(topology_st.standalone.serverid),
+                'rmdir /var/log/dirsrv/slapd-{}/tmp'.format(topology_st.standalone.serverid),
                 'setenforce 1']
 
         for cmd in cmds:
-            log.info('Command used : %s' % cmds)
-            subprocess.Popen(cmd, shell=True)
+            log.info('Command used : %s' % cmd)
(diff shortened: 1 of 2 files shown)


droideck (changes requested):
Just to be on a safe side, could you please change both (34, 43 lines) `subprocess.Popen(cmd, shell=True)` calls with `subprocess.run(cmd, shell=True)`?

`Popen` spawns the process and returns immediately without waiting for it to finish (unless you command it to `.wait()`). `rmdir` and `umount` on the same dir may introduce a race condition.

dirsrvtests/tests/suites/disk_monitoring/disk_monitoring_divide_test.py
```diff
@@ -31,15 +31,16 @@ def create_dummy_mount(topology_st, request):
     log.info('Create dummy mount')
     for cmd in cmds:
         log.info('Command used : %s' % cmd)
-        subprocess.Popen(cmd, shell=True)
+        subprocess.run(cmd, shell=True, check=True)
```
droideck: That would fail on SELinux-permissive systems because of the `check=True`.

droideck (approved):
Otherwise, LGTM!

droideck (approved):

PR #7239, opened 2026-02-11
Issue 6753 - Port ticket 49039 test

mirielka:
Description:
Port ticket 49039 test into
dirsrvtests/tests/suites/password/pwp_test.py

Relates: #6753
Author: Lenka Doudova
Assisted by: Cursor
Reviewer: ???

## Summary by Sourcery

Port the password policy regression test for ticket 49039 into the main password test suite and remove the old ticket-specific test file.

Tests:
- Add a test ensuring passwordMinAge does not prevent a user from changing a reset password, including TLS and ldappasswd coverage.
- Remove the legacy ticket49039-specific test file now covered by the password test suite.


dirsrvtests/tests/suites/password/pwp_test.py
```diff
@@ -512,6 +515,84 @@ def test_passwordlockout(topo, _fix_password):
...
+        # Reset again as Directory Manager
+        dm.rebind(PASSWORD)
+        user.replace('userpassword', PASSWORD)
+        time.sleep(1)
+
+        # Run ldappasswd as user to change password to password2
+        env = os.environ.copy()
+        env['LDAPTLS_CACERTDIR'] = topo.standalone.get_cert_dir()
+        cmd = [
+            'ldappasswd', #topo.standalone.host,
```
bsimonova: Remove the commented out topo.standalone.host

dirsrvtests/tests/suites/password/pwp_test.py
```diff
@@ -512,6 +515,84 @@ def test_passwordlockout(topo, _fix_password):
...
+        # Reset password as user (must succeed; min age must not block after reset)
+        user.bind(PASSWORD)
+        user.replace('userpassword', PASSWORD)
+
+        # Reset again as Directory Manager
+        dm.rebind(PASSWORD)
+        user.replace('userpassword', PASSWORD)
+        time.sleep(1)
+
+        # Run ldappasswd as user to change password to password2
```
bsimonova: Just a nitpick to add more info to the comment that we specifically test the ldappasswd tool like in original test: 
Change password through ldappasswd as user to ensure functionality

bsimonova (approved):
Minor changes, otherwise LGTM.

dirsrvtests/tests/suites/password/pwp_test.py
```diff
@@ -512,6 +515,83 @@ def test_passwordlockout(topo, _fix_password):
...
+    dm.bind()
+
+    user = _create_user(topo, 'user', 'Test User', '1002', PASSWORD)
+    try:
+        # Reset password as Directory Manager
+        user.replace('userpassword', PASSWORD)
+        time.sleep(1)
+
+        # Reset password as user (must succeed; min age must not block after reset)
+        user.bind(PASSWORD)
```
vashirov: `rebind()` does a simple bind on the existing connection, without doing additional searches for version, etc. 
```suggestion
        user.rebind(PASSWORD)
```

dirsrvtests/tests/suites/password/pwp_test.py
```diff
@@ -512,6 +515,83 @@ def test_passwordlockout(topo, _fix_password):
...
+            '-D', user.dn, '-w', PASSWORD,
+            '-a', PASSWORD, '-s', 'password2',
+            user.dn,
+        ]
+        result = subprocess.run(cmd, env=env, capture_output=True, text=True)
+        assert result.returncode == 0, f'ldappasswd failed: {result.stderr}'
+
+        # Bind as user with new password
+        user.bind('password2')
+    finally:
```
vashirov, line 591: In case of test failure, we need to ensure we can delete the user (it might be still bound as a regular user). 
```suggestion
    finally:
    dm.rebind(PASSWORD)
```

vashirov (commented):
> It references {self.serverid} - shouldn't that be {self._serverid} ?

Yes, it should. 

> but the user account does not have the permission to read this value?

User is in a state where only password change operation can be completed, not search for the vendor version. So the search operation fails.

vashirov (approved):

PR #6934, opened 2025-08-13
Issue 6928 - The parentId attribute is indexed with improper matching…

vashirov:
… rule

Bug Description:
`parentId` attribute contains integer values and needs to be indexed with `integerOrderingMatch` matching rule. This attribute is a system attribute and the configuration entry for this attribute is created when a backend is created. The bug is that the per backend configuration entry does not contain `nsMatchingRule: integerOrderingMatch`.

Fix Description:
* Update `ldbm_instance_create_default_indexes` to support matching rules and update default system index configuration for `parentId` to include `integerOrderingMatch` matching rule.
* Add healthcheck linter for default system indexes and indexes created by RetroCL and USN plugins.

Fixes: https://github.com/389ds/389-ds-base/issues/6928
Fixes: https://github.com/389ds/389-ds-base/issues/6915

PR #7240, opened 2026-02-11
Fix system index configuration issues (parentid/ancestorid) - backport to 1.4.3

vashirov:
## Summary by Sourcery

Adjust system index configuration and introduce tooling to detect and remediate parentid/ancestorid index issues.

New Features:
- Add a dbtasks index-check CLI command to detect and optionally fix system index ordering and configuration problems using dbscan and dse.ldif inspection.
- Run an early server upgrade hook at startup to clean up problematic index settings and log guidance for missing integerOrderingMatch on parentid indexes.

Bug Fixes:
- Remove legacy nsIndexIDListScanLimit configuration paths and related CLI options to prevent misconfiguration of system indexes.
- Ensure ancestorid index configuration entries are removed both on upgrade and in default index templates so ancestorid remains an internal system index.

Enhancements:
- Simplify default system index creation by removing scanlimit handling and ensuring parentid/ancestorid always use integerOrderingMatch.
- Refine backend system index linting to no longer manage scanlimit settings and to focus on required matching rules.
- Add runtime checks on backend startup to detect mismatches between configured and on-disk index ordering for parentid and ancestorid and to advise reindexing if needed.
- Extend DSEldif utilities to support global substitution, case-insensitive attribute lookup, and backend discovery.

Tests:
- Update healthcheck and configuration tests to reflect removal of scanlimit from system indexes and to cover upgrade cleanup of parentid/ancestorid configuration.
- Add an extensive test suite for the new index-check command, including basic usage, backend selection, mismatch detection, and automatic fixing of multiple index issues.
- Simplify paging and require-index tests now that ancestorid-specific scanlimit tuning is no longer needed.

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
@@ -451,6 +451,58 @@ def test_multiple_missing_indexes(topology_st, log_buffering_enabled):
     run_healthcheck_and_flush_log(topology_st, standalone, json=True, searched_code=JSON_OUTPUT)
 
 
+def test_upgrade_removes_parentid_scanlimit(topology_st):
+    """Check if upgrade function removes nsIndexIDListScanLimit from parentid index
+
+    :id: 2808886e-c1c1-441d-b3a3-299c4ef1ab4a
+    :setup: Standalone instance
+    :steps:
+        1. Create DS instance
+        2. Stop the server
+        3. Use DSEldif to add nsIndexIDListScanLimit to parentid index
+        4. Start the server (triggers upgrade)
+        5. Verify nsIndexIDListScanLimit is removed from parentid index
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. nsIndexIDListScanLimit is no longer present
+    """
+    from lib389.dseldif import DSEldif
+
+    standalone = topology_st.standalone
+    PARENTID_DN = "cn=parentid,cn=index,cn=userroot,cn=ldbm database,cn=plugins,cn=config"
+    SCANLIMIT_VALUE = "limit=5000 type=eq flags=AND"
+
+    log.info("Stop the server")
+    standalone.stop()
+
+    log.info("Add nsIndexIDListScanLimit to parentid index using DSEldif")
+    dse_ldif = DSEldif(standalone)
+    dse_ldif.add(PARENTID_DN, "nsIndexIDListScanLimit", SCANLIMIT_VALUE)
+
+    # Verify it was added
+    scanlimit = dse_ldif.get(PARENTID_DN, "nsIndexIDListScanLimit")
+    assert scanlimit is not None, "Failed to add nsIndexIDListScanLimit"
+    log.info(f"Added nsIndexIDListScanLimit: {scanlimit}")
+
+    log.info("Start the server (triggers upgrade)")
+    standalone.start()
+
+    log.info("Verify nsIndexIDListScanLimit was removed by upgrade")
+    # Check via LDAP - the upgrade should have removed it
+    parentid_index = Index(standalone, PARENTID_DN)
+    scanlimit_after = parentid_index.get_attr_vals_utf8("nsIndexIDListScanLimit")
+    log.info(f"nsIndexIDListScanLimit after upgrade: {scanlimit_after}")
+
+    # The upgrade function should have removed nsIndexIDListScanLimit
+    assert not scanlimit_after, \
+        f"nsIndexIDListScanLimit should have been removed but found: {scanlimit_after}"
+
+    log.info("Upgrade successfully removed nsIndexIDListScanLimit from parentid index")
+
+
 if __name__ == "__main__":
     # Run isolated
     # -s for DEBUG mode
--- a/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
@@ -172,8 +172,7 @@ def test_missing_parentid(topology_st, log_buffering_enabled):
 
     log.info("Re-add the parentId index")
     backend = Backends(standalone).get("userRoot")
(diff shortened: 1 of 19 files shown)


dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
```diff
@@ -406,178 +404,781 @@ def test_retrocl_plugin_missing_matching_rule(topology_st, retrocl_plugin_enable
...
+    dse_ldif.add(PARENTID_DN, "nsIndexIDListScanLimit", SCANLIMIT_VALUE)
 
-    log.info("Remove both integerOrderingMatch and nsIndexIDListScanLimit from parentId index")
+    # Verify it was added
+    scanlimit = dse_ldif.get(PARENTID_DN, "nsIndexIDListScanLimit")
+    assert scanlimit is not None, "Failed to add nsIndexIDListScanLimit"
+    log.info(f"Added nsIndexIDListScanLimit: {scanlimit}")
+
+    log.info("Start the server (triggers upgrade)")
+    standalone.start()
```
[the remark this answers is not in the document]
progier389: we do not care because there is only a single backend involved in that test

ldap/servers/slapd/back-ldbm/instance.c
```diff
@@ -308,23 +239,281 @@ ldbm_instance_create_default_indexes(backend *be)
...
          */
-        ainfo_get(be, (char *)LDBM_ANCESTORID_STR, &ai);
-        /* Check if the attrinfo is actually for ancestorid, not a fallback to .default */
-        index_already_configured = (ai != NULL && strcmp(ai->ai_type, LDBM_ANCESTORID_STR) == 0);
-        if (!index_already_configured) {
-            e = ldbm_instance_init_config_entry(LDBM_ANCESTORID_STR, "eq", 0, 0, 0, "integerOrderingMatch", ancestorid_indexes_limit);
-            ldbm_instance_config_add_index_entry(inst, e, flags);
-            attr_index_config(be, "ldbm index init", 0, e, 1, 0, NULL);
-            slapi_entry_free(e);
+        e = ldbm_instance_init_config_entry(LDBM_ANCESTORID_STR, "eq", 0, 0, 0, "integerOrderingMatch");
```
tbordaz: Indexing ancestorid in 'integerOrdering' ? I thought we wanted to introduce this only in 10.3 .

ldap/servers/slapd/back-ldbm/instance.c
```diff
@@ -308,23 +239,281 @@ ldbm_instance_create_default_indexes(backend *be)
...
+            inst->inst_name, index_name);
+
+    /* Check if integerOrderingMatch is configured in DSE */
+    config_has_int_order = ldbm_instance_index_has_int_order_in_dse(inst->inst_name, index_name);
+
+    /* Get attrinfo for the index */
+    ainfo_get(be, (char *)index_name, &ai);
+    if (ai == NULL || strcmp(ai->ai_type, index_name) != 0) {
+        /* No index config found */
+        slapi_log_err(SLAPI_LOG_DEBUG, "ldbm_instance_check_index_config",
```
tbordaz, line 358: What about writing it in ERR level ?

ldap/servers/slapd/back-ldbm/instance.c
```diff
@@ -308,23 +239,281 @@ ldbm_instance_create_default_indexes(backend *be)
...
+                inst->inst_name, index_name);
+        return;
+    }
+
+    /* Open the index file */
+    ret = dblayer_get_index_file(be, ai, &db, 0);
+    if (ret != 0 || db == NULL) {
+        /* Index file doesn't exist or can't be opened - this is fine for new instances */
+        slapi_log_err(SLAPI_LOG_DEBUG, "ldbm_instance_check_index_config",
+                "Backend '%s': could not open %s index file (ret=%d), skipping order check\n",
```
tbordaz, line 369: Writing it in ERR level ?

src/lib389/lib389/backend.py
```diff
@@ -569,10 +569,11 @@ def _lint_system_indexes(self):
         indexes = self.get_indexes()
 
         # Default system indexes taken from ldap/servers/slapd/back-ldbm/instance.c
+        # Note: entryrdn and ancestorid are internal system indexes that are not
+        # exposed in cn=config - they are managed internally by the server.
+        # Only parentid has a DSE config entry (for the integerOrderingMatch rule).
         expected_system_indexes = {
-            'entryrdn': {'types': ['subtree'], 'matching_rule': None},
```
tbordaz, line 573: I agree with this change (entryrdn) but I do not recall it was added during the upgrade index issue ?
Is this change doing a old code cleanup or reverting something added recently ?

tbordaz (commented):
The code looks good.
Few questions

ldap/servers/slapd/back-ldbm/instance.c
vashirov: You're right, and the same (lexicographical ordering) should be in main. With the current fix in main we again have a discrepancy between the config (integer) and disk (lexicographic). I will fix it here and submit another PR to fix it in main.

ldap/servers/slapd/back-ldbm/instance.c
vashirov: Initially it was NOTICE, but on instances with a lot of backends it would flood the logs on every startup, so I moved it to DEBUG. Only when a discrepancy is found we log it in ERR.

src/lib389/lib389/backend.py
vashirov: expected_system_indexes was added as part of https://github.com/389ds/389-ds-base/pull/6934#discussion_r2273854469 for de17154efc4fa40393f7560ec6ffc26af8560f84

tbordaz: okay thanks

ldap/servers/slapd/back-ldbm/instance.c
tbordaz: That is is right. thanks

tbordaz (approved):
LGTM

PR #7242, opened 2026-02-11
Issue 7241 - Drop dateutil

### LINKED ISSUE #7241: Drop dateutil
**Issue Description**
From https://fedoraproject.org/wiki/Changes/DeprecatePython-dateutil

> Python-dateutil is an extension to the standard Python datetime features, widely used by a lot of other projects. However, upstream is unmaintained and possibly with unaddressed security issues.

We have only few places using dateutil, shouldn't be hard to get rid of it.

vashirov:
Bug Description:
python-dateutil is unmaintained upstream and is marked for deprecation.

Fix Description:
* Replace `dateutil.tz.tzoffset` with `datetime.timezone(datetime.timedelta())`.
* Replace `dateutil.parser.parse` with standard `datetime` calls.
* Import `datetime` as `dt` to avoid confusion between module and class.

Fixes: https://github.com/389ds/389-ds-base/issues/7241

## Summary by Sourcery

Replace use of python-dateutil with standard library datetime handling and remove the deprecated dependency.

Bug Fixes:
- Ensure log timestamp and password expiration time parsing uses robust standard library datetime APIs instead of the deprecated python-dateutil package.

Enhancements:
- Refine log timestamp parsing to explicitly handle month and timezone offset components using datetime.timezone and datetime.timedelta.
- Standardize datetime imports in tests and code by aliasing the module as dt to avoid ambiguity.

Build:
- Remove python-dateutil from pyproject dependency configuration and lib389 requirements.

Tests:
- Update log parsing and password policy tests to assert against datetime objects built with standard library timezone and parsing APIs instead of python-dateutil helpers.

### DIFF
--- a/src/lib389/lib389/tests/dirsrv_log_test.py
+++ b/src/lib389/lib389/tests/dirsrv_log_test.py
@@ -12,8 +12,7 @@ from lib389 import DirSrv, Entry
 import pytest
 import time
 import shutil
-import datetime
-from dateutil.tz import tzoffset
+import datetime as dt
 
 INSTANCE_PORT = 54321
 INSTANCE_SERVERID = 'standalone'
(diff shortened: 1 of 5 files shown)


jchapma (approved):
LGTM

src/lib389/lib389/tests/dirsrv_log_test.py
```diff
@@ -74,29 +73,29 @@ def test_access_log(topology):
         topology.standalone.ds_access_log.parse_line('[27/Apr/2016:12:49:49.726093186 +1000] conn=1 fd=64 slot=64 connection from ::1 to ::1') ==
         {
             'slot': '64', 'remote': '::1', 'action': 'CONNECT', 'timestamp': '[27/Apr/2016:12:49:49.726093186 +1000]', 'fd': '64', 'conn': '1', 'local': '::1',
-            'datetime': datetime.datetime(2016, 4, 27, 12, 0, 0, 726093, tzinfo=tzoffset(None, 36000))
+            'datetime': dt.datetime(2016, 4, 27, 12, 0, 0, 726093, tzinfo=dt.timezone(dt.timedelta(seconds=36000)))
```
droideck: We, probably, should assert minutes and seconds too here. But also, the tests will fail much early as these `lib389/tests` are broken...

droideck (approved):
Looks good!

Issue #7066, opened 2025-10-23
Subtree and user password policy history count cannot be set to 0


mirielka:
**Issue Description**
Subtree and user password policies cannot have passwordInHistory attribute set to 0. This setting allows for checking only current password, but no previous password. Global password policies can have passwordInHistory set to 0.

**Package Version:**
389-ds-base-3.1.3-5.el10_1.x86_64

**Steps to Reproduce**
1. Create a subtree/user password policy
2. Attempt setting passwordInHistory to 0
Fails with
`passwordInHistory: invalid value "0"`

**Expected results**
Should accept 0 value

Issue #7052, opened 2025-10-16
Setting password history count to 0 does not flush history


mirielka:
Setting password history count to 0 should flush all passwords stored in the history

**Package version:**
389-ds-base-3.1.3-5.el10_1.x86_64

**Steps to reproduce**
1.     Enable global password history
2.     Set password history count to e. g. 2,
3.     Fill a user's password history with values, e.g. password1, password2
4.     Set password history count to 0 - only current password is blocked (e. g. password3)
5.     Set password history count to 2 again
6.     Check that any of the older passwords (password1, password2) are still usable

**Expected results**
It should be possible to set both password1 and password2, but not password3 (current password)

**Actual results**
Both password1 and password2 are blocked, the history is still populated by original values before changing password history count to 0

**Additional info**
Only applies to setting password history count to 0. Other cases (e.g. 3 -> 2) work properly.

mreynolds389:
0b8a67b45..11b941349  389-ds-base-3.1 -> 389-ds-base-3.1
c9c203f51..fb0cf4a4b  389-ds-base-3.0 -> 389-ds-base-3.0
75b1213df..7809104b9  389-ds-base-2.8 -> 389-ds-base-2.8
5c3f2af34..faae4ca8a  389-ds-base-2.7 -> 389-ds-base-2.7
ff4d93a27..b8d578066  389-ds-base-2.6 -> 389-ds-base-2.6

PR #7245, opened 2026-02-11
Issue 7066/7052 - allow password history to be set to zero and remove history

mreynolds389:
Description:

For local password policies the server was incorrectly rejecting updates that set the value to zero.  When password history is set to zero the old passwords in the entry history are not cleaned as expected.

relates: https://github.com/389ds/389-ds-base/issues/7052
relates: https://github.com/389ds/389-ds-base/issues/7066

## Summary by Sourcery

Allow password history policies to accept a value of zero and ensure existing history entries are cleared when history is disabled.

Bug Fixes:
- Fix rejection of password policy updates that set pw_inhistory to zero.
- Clear stored passwordHistory values when password history enforcement is disabled for an entry.

Enhancements:
- Enable zero as a valid value for pw_inhistory in password policy configuration validation.

Tests:
- Unskip and adjust password policy history tests to cover zero-history behavior and admin password resets.

### DIFF
--- a/ldap/servers/slapd/pw.c
+++ b/ldap/servers/slapd/pw.c
@@ -1535,7 +1535,18 @@ update_pw_history(Slapi_PBlock *pb, const Slapi_DN *sdn, char *old_pw)
     pwpolicy = new_passwdPolicy(pb, dn);
 
     if (pwpolicy->pw_inhistory == 0){
-        /* We are only enforcing the current password, just return */
+        /* We are only enforcing the current password, just return but first
+         * cleanup any old passwords in the history */
+        attribute.mod_type = "passwordHistory";
+        attribute.mod_op = LDAP_MOD_REPLACE;
+        attribute.mod_values = NULL;
+        list_of_mods[0] = &attribute;
+        list_of_mods[1] = NULL;
+        mod_pb = slapi_pblock_new();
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/pw.c
```diff
@@ -1535,7 +1535,18 @@ update_pw_history(Slapi_PBlock *pb, const Slapi_DN *sdn, char *old_pw)
     pwpolicy = new_passwdPolicy(pb, dn);
 
     if (pwpolicy->pw_inhistory == 0){
-        /* We are only enforcing the current password, just return */
+        /* We are only enforcing the current password, just return but first
+         * cleanup any old passwords in the history */
+        attribute.mod_type = "passwordHistory";
+        attribute.mod_op = LDAP_MOD_DELETE;
```
progier389: Should not it be LDAP_MOD_REPLACE ? (to avoid error if the attribute does not exists)

mreynolds389: That's why I choose to ignore errors on this mod op.  I never tried a mod replace with null values - I didn't think that would work.  I'll try it...

mreynolds389: > That's why I choose to ignore errors on this mod op. I never tried a mod replace with null values - I didn't think that would work. I'll try it...

@progier389 yeah it works fine, please review...

progier389 (approved):
LGTM

Issue #7250, opened 2026-02-12
CLI - dsctl db2index not working correctly in mdb


mreynolds389:
**Issue Description**

If you don't supply a backend name to dsctl db2index then all attributes for all backends are reindexed.  This is done by calling the "upgradedb" command, but "upgradedb" doesn't work with mdb.  In the mdb case we need a different approach

PR #7251, opened 2026-02-12
Issue - CLI - dsctl db2index needs some hardening with MBD

mreynolds389:
Description:

The usage for dsctl db2index was confusing. The way the attr options and backend name were displayed it looks like the backend name could come after the attributes, but instead the backend name was treated as an attribute.

Instead make the backend name required, and change the attribute naming to require individual options instead of a list of values.

Relates: https://github.com/389ds/389-ds-base/issues/7250

## Summary by Sourcery

Require an explicit backend when running db2index and simplify db task logging integration.

New Features:
- Make the db2index CLI require a backend argument and accept repeated attribute options instead of a positional list.

Bug Fixes:
- Ensure db2index treats the backend name as a backend rather than as an attribute and harden its invocation semantics.

Enhancements:
- Propagate the CLI logger into db task instance methods for db2index, backups, LDIF operations, and verification to centralize logging.
- Clarify db2index help text to describe the new attribute flag usage and backend requirement.
- Document that the db2index upgradedb path only works with BDB backends.

Tests:
- Update the import test to call db2index with an explicit backend to match the new CLI requirements.

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -2953,7 +2953,7 @@ class DirSrv(SimpleLDAPObject, object):
 
         return True
 
-    def db2index(self, bename=None, suffixes=None, attrs=None, vlvTag=None):
+    def db2index(self, bename, suffixes=None, attrs=None, vlvTag=None):
         """
         @param bename - The backend name to reindex
         @param suffixes - List/tuple of suffixes to reindex, currently unused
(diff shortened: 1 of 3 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -2986,6 +2986,7 @@ def db2index(self, bename=None, suffixes=None, attrs=None, vlvTag=None):
                         cmd.append('-t')
                         cmd.append(idx)
         else:
+            # only works in BDB
```
progier389: we may perhaps want to check the db implementation 
 and if mdb:
   walk all backend and index them

mreynolds389: > we may perhaps want to check the db implementation and if mdb: walk all backend and index them

Well using the CLI we will never hit this condition as a backend name is now required.  But it can happen if you run "ns-slapd db2index" directly.  Best to be safe so I'll tweak this, thanks!

mreynolds389: @progier389 so I just removed the upgradedb part and used the same approach for all db types.  The upgradedb part was a just short cut from the old perl scripts and it was brought over when we first implemented the new python tools.

progier389 (approved):
LGTM

Issue #7252, opened 2026-02-12
PQC - Should refer ML-DSDA-* oids by tags


progier389:
**Issue Description**
Should not directly use         
       SEC_OID_ML_DSA_44,
        SEC_OID_ML_DSA_65,
        SEC_OID_ML_DSA_87,
(because the value may/will change ) but rather iterate of the tag to dynamically find them.

**Package Version and Platform:**
 - Platform: [e.g. Fedora]
 - Package and version: main

**Additional context**
Current code generate build break on RHEL with nss 3.112

PR #7254, opened 2026-02-12
Issue 7252 - PQC - Need to iterate on SECOidTag instead of using OID

progier389:
Need to dynamically iterate on SECOidTag instead of using SEC_OID_ML_DSA_*  OIDs to avoid issue with upcoming nss versions and fix a RHEL build break with nss 3.112

Issue: #7252 

Reviewed by: @mreynolds389, @droideck, @vashirov 

## Summary by Sourcery

Enhancements:
- Iterate over SECOidTag entries to detect ML-DSA algorithms by description and apply SSL/signature usage policy when not already enabled and not in FIPS mode.

### DIFF
--- a/ldap/servers/slapd/ssl.c
+++ b/ldap/servers/slapd/ssl.c
@@ -732,31 +732,25 @@ SSLPLCY_Install(void)
 {
 
     SECStatus s = 0;
-#ifdef MAX_ML_DSA_PRIVATE_KEY_LEN
     int flags = NSS_USE_ALG_IN_SIGNATURE | NSS_USE_ALG_IN_SSL;
-    static const SECOidTag oids[] = {
-        SEC_OID_ML_DSA_44,
-        SEC_OID_ML_DSA_65,
-        SEC_OID_ML_DSA_87,
-    };
-#endif
+    SECOidData *oid = NULL;
 
     s = NSS_SetDomesticPolicy();
 
-#ifdef MAX_ML_DSA_PRIVATE_KEY_LEN
     /* Should rely on the crypto module policy in FIPS mode */
     if (!slapd_pk11_isFIPS()) {
         /* Set explicitly PQC algorithm policy if it is not set by default */
-        for (size_t i=0; s == SECSuccess && i < PR_ARRAY_SIZE(oids); i++) {
-            PRUint32 oflags = 0;
-            (void) NSS_GetAlgorithmPolicy(oids[i], &oflags);
-            if ((oflags & flags) != flags) {
-                s = NSS_SetAlgorithmPolicy(oids[i], flags, 0);
+        for (SECOidTag tag = 1; s == 0 && (oid = SECOID_FindOIDByTag(tag)) != NULL; tag++) {
+            if (oid->mechanism != CKM_INVALID_MECHANISM &&
+                PL_strncasecmp(oid->desc, "ML-DSA-", 7) == 0) {
+                PRUint32 oflags = 0;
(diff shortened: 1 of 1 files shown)


ldap/servers/slapd/ssl.c
```diff
@@ -732,31 +732,25 @@ SSLPLCY_Install(void)
...
     if (!slapd_pk11_isFIPS()) {
         /* Set explicitly PQC algorithm policy if it is not set by default */
-        for (size_t i=0; s == SECSuccess && i < PR_ARRAY_SIZE(oids); i++) {
-            PRUint32 oflags = 0;
-            (void) NSS_GetAlgorithmPolicy(oids[i], &oflags);
-            if ((oflags & flags) != flags) {
-                s = NSS_SetAlgorithmPolicy(oids[i], flags, 0);
+        for (SECOidTag tag = 1; s == 0 && (oid = SECOID_FindOIDByTag(tag)) != NULL; tag++) {
+            if (oid->mechanism != CKM_INVALID_MECHANISM &&
+                PL_strncasecmp(oid->desc, "ML-DSA-", 7) == 0) {
```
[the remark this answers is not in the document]
progier389: oid->desc cannot be NULL -  NULL entry in OID table are skipped as they have CKM_INVALID_MECHANISM

mreynolds389 (approved):

ldap/servers/slapd/ssl.c
droideck, line 745: If I understand correctly, `oid->desc` can be NULL if build with `SECOID_NO_STRINGS`. So even though it's highly unlikely, IMO, it'll be safer to have a guard here.

ldap/servers/slapd/ssl.c
```diff
@@ -732,31 +732,25 @@ SSLPLCY_Install(void)
...
-#ifdef MAX_ML_DSA_PRIVATE_KEY_LEN
     /* Should rely on the crypto module policy in FIPS mode */
     if (!slapd_pk11_isFIPS()) {
         /* Set explicitly PQC algorithm policy if it is not set by default */
-        for (size_t i=0; s == SECSuccess && i < PR_ARRAY_SIZE(oids); i++) {
-            PRUint32 oflags = 0;
-            (void) NSS_GetAlgorithmPolicy(oids[i], &oflags);
-            if ((oflags & flags) != flags) {
-                s = NSS_SetAlgorithmPolicy(oids[i], flags, 0);
+        for (SECOidTag tag = 1; s == 0 && (oid = SECOID_FindOIDByTag(tag)) != NULL; tag++) {
```
droideck:
```suggestion
        for (SECOidTag tag = 1; s == SECSuccess && (oid = SECOID_FindOIDByTag(tag)) != NULL; tag++) {
```

Minor nitpick, but I think it's a bit cleaner.:)

droideck (approved):
Couple of minor comments, otherwise, looks good!

vashirov (approved):
Build on RHEL 10.2 works, test `dirsrvtests/tests/suites/tls/mldsa_test.py::test_mldsa` passed with this build with NSS 3.112 (I had to remove skipif).

ldap/servers/slapd/ssl.c
progier389: On the other side if  build with  SECOID_NO_STRINGS, the whole logic of this code is broken 
  as we cannot compare the desc. 
IMHO it is then better to get a clean crash to warn that something is amiss rather than silently disabling PQC policy.
Another point:   The SECOID_FindOIDTagFromDescripton function in NSS code does not check for NULL either

Issue #7174, opened 2026-01-08
Add github spec-kit to the repo


mreynolds389:
Add the basic github spec-kit into the source tree for easier use of spec-driven development tasks

https://github.com/github/spec-kit/

It will add ".specify" directory for the github's spec-kit.  There are additional IDE configurations that can be added for cursor, co-pilot, claude-code, gemini, etc.

PR #7255, opened 2026-02-12
Issue 7174 - Add spec-kit constitution file

mreynolds389:
Description:

Using a constitution file is the foundation of using spec-driven development. For now just add this file so the entire team and can use it if they deceide they want to use a spec-kit for development.

relates: https://github.com/389ds/389-ds-base/issues/7174

## Summary by Sourcery

Documentation:
- Add a versioned constitution document outlining reliability, performance, security, standards compliance, and development workflow requirements for the project.

### DIFF
--- a/constitution.md
+++ b/constitution.md
@@ -1,0 +1,110 @@
+<!--
+Sync Impact Report:
+Version change: N/A → 1.0.0 (initial constitution)
+Modified principles: N/A (all new)
+Added sections: Reliability, Performance & Scalability, Standards Compliance, Enterprise Features, Development Workflow
+Removed sections: N/A
+Templates requiring updates:
+  ✅ plan-template.md (Constitution Check section exists, no changes needed)
+  ✅ spec-template.md (no constitution-specific references)
+  ✅ tasks-template.md (no constitution-specific references)
+  ✅ All command files (no constitution-specific references found)
+Follow-up TODOs: None
+-->
+
+# 389 Directory Server Constitution
+
+## Core Principles
+
+### I. Reliability (NON-NEGOTIABLE)
+The 389 Directory Server MUST be an open source, real-world, hardened directory service. All code MUST be extensively tested with sanitization tools before integration. The server MUST provide a rich feature set of fail-over and backup technologies to give administrators confidence their accounts are safe. Reliability is critical for enterprise deployments handling identity and organizational data.
+
+### II. Performance & Scalability (NON-NEGOTIABLE)
+The server MUST be a high-performance LDAP server capable of handling thousands of operations per second and supporting hundreds of thousands of accounts. Database size MUST only be restricted by disk space. The server MUST support high throughput performance and multi-supplier replication for horizontal scaling to meet the needs of the most demanding environments - from small business to cloud deployments.
+
+### III. Standards Compliance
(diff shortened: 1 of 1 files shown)


constitution.md
```diff
@@ -0,0 +1,110 @@
...
+All LDAP protocol implementations MUST comply with relevant RFC specifications. The server MUST support:
+- Core LDAPv3 protocols (RFC 2251, RFC 3377)
+- Attribute syntax and schema definitions (RFC 2252, RFC 2256, RFC 1274)
+- Distinguished name representations (RFC 2253, RFC 1779, RFC 2247)
+- Search filters and LDAP URLs (RFC 2254, RFC 2255, RFC 1558)
+- Authentication and security (RFC 2829, RFC 2830, RFC 2222)
+- LDIF format (RFC 2849)
+- Operational attributes and controls (RFC 3673, RFC 4527)
+- Network Information Service (RFC 2307)
+- Other applicable RFCs as specified
```
droideck, line 68: Outdated RFCs (mostly), RFC 4510 obsolete a lot of them and specified which are the new ones, and RFC 4422 is the new for SASL.

constitution.md
```diff
@@ -0,0 +1,110 @@
...
+New features MUST include:
+- User-facing documentation
+- Configuration documentation
+- API documentation (for plugins/APIs)
+- Migration guides (for breaking changes)
+- Standards compliance documentation (for protocol features)
+
+## Governance
+
+This constitution supersedes all other development practices and guidelines. All pull requests and code reviews MUST verify compliance with these principles. Amendments to this constitution require:
```
droideck, line 102: "Supersedes all other development practices" is pretty strong, IMO. I'd soften that to something like "complements existing project policies" with maintainer decision as the tiebreaker when things conflict. Otherwise this technically overrides release engineering, security processes, etc. which I don't think is the intent.
So to sum up -- that's good, we just should be careful with such strong assertions as it may create a worse result.

constitution.md
```diff
@@ -0,0 +1,110 @@
...
+
+# 389 Directory Server Constitution
+
+## Core Principles
+
+### I. Reliability (NON-NEGOTIABLE)
+The 389 Directory Server MUST be an open source, real-world, hardened directory service. All code MUST be extensively tested with sanitization tools before integration. The server MUST provide a rich feature set of fail-over and backup technologies to give administrators confidence their accounts are safe. Reliability is critical for enterprise deployments handling identity and organizational data.
+
+### II. Performance & Scalability (NON-NEGOTIABLE)
+The server MUST be a high-performance LDAP server capable of handling thousands of operations per second and supporting hundreds of thousands of accounts. Database size MUST only be restricted by disk space. The server MUST support high throughput performance and multi-supplier replication for horizontal scaling to meet the needs of the most demanding environments - from small business to cloud deployments.
```
droideck, line 23: IMO, the absolute performance MUSTs are a bit tricky. "MUST support thousands of operations per second" and "MUST support hundreds of thousands of accounts" -- these are good but they're not testable at the PR level. And they are too broad.
Also, you suggest below -- "All pull requests and code reviews MUST verify compliance with these principles.". And these MUSTs are not testable.

Something like "for performance-affecting changes, validate no regression against established benchmarks" would be more practical and still provide a nice direction for the dev.

Issue #7223, opened 2026-02-05
Fix system index configuration issues (parentid/ancestorid)


vashirov:
**Issue Description**
Several issues exist with system index configuration for parentid and ancestorid indexes after recent updates.

1. parentid index missing integerOrderingMatch
The parentid index should use integer ordering for correct subtree searches, but older installations may be missing this matching rule, causing incorrect search results. Was attempted to address in #6928 and #7172. 

2. ancestorid has incorrect DSE config entry
The ancestorid index is special (no corresponding attribute type) and should not have a DSE configuration entry. However, it was incorrectly added to `cn=default indexes` in the previous changes.

3. `nsIndexIDListScanLimit` on system indexes
Previous changes added scan limits to parentid/ancestorid which can cause performance issues and should be removed.

4. Config/disk ordering mismatch
If the DSE config has integerOrderingMatch but the on-disk index uses lexicographic ordering (or vice versa), searches return incorrect results.

vashirov:
f7298a5ef..a40336390  389-ds-base-2.2 -> 389-ds-base-2.2
b97efcbb5..6d14ff36c  389-ds-base-2.4 -> 389-ds-base-2.4
119483e66..47d77326c  389-ds-base-2.6 -> 389-ds-base-2.6
b36d425bb..d346cdaa7  389-ds-base-2.7 -> 389-ds-base-2.7
4209ff653..09b6f770f  389-ds-base-2.8 -> 389-ds-base-2.8
fb0cf4a4b..6d2d9740f  389-ds-base-3.0 -> 389-ds-base-3.0
12125e237..273e0a92d  389-ds-base-3.1 -> 389-ds-base-3.1
b216b86c5..6ce19a9a3  main -> main

Firstyear:
@vashirov Hey mate, we had some people report issues moving from 2.5 (without this patch) to 2.7 (with this patch). Did you do upgrade testing? Are there any steps that need to be taken? Seeing people end up with queries like `(cn=*)` return 0 results.

vashirov:
@Firstyear can you tell me precise versions of 2.5 and 2.7 packages? There were multiple fixes around parentid/ancestorid indexes and scan limits.
We do upgrade testing, there shouldn't be any manual steps required, unless it's an edge case like #7518.

Firstyear:
2.5.3 - 95b15d57c
2.7.0 - f597a91d8

[...]

PR #7256, opened 2026-02-13
Issue 7223 - Use lexicographical order for ancestorid

vashirov:
Description:
`ldbm_instance_create_default_indexes()` configured ancestorid with integerOrderingMatch in the in-memory attrinfo, but ancestorid on disk might be using lexicographic ordering (data before the upgrade or after ldif2db import).

Relates: https://github.com/389ds/389-ds-base/issues/7223

## Summary by Sourcery

Bug Fixes:
- Prevent mismatches between in-memory and on-disk ancestorid index ordering that could occur when ancestorid was configured with integerOrderingMatch in memory while stored lexicographically on disk.


tbordaz (approved):
LGTM

Issue #7213, opened 2026-01-30
MDB_BAD_VALSIZE error while handling VLV


progier389:
**Issue Description**
With LMDB: Operation fail with MDB_BAD_VALSIZE error 
if trying to add an entry with an attribute indexed by VLV if the attribute value is too long

**Package Version and Platform:**
 - Platform: [e.g. Fedora] RHEL
 - Package and version:  389-ds-base-3.0.6-13.el10_0.x86_64

**Steps to Reproduce**

Run the following script then check M2 error log

```
#!/bin/bash
SUFFIX="dc=example,dc=com"
DM_DN="cn=Directory Manager"
DM_PASSWD="password"
REPL_MANAGER_DN="cn=replication manager,cn=config"
REPL_MANAGER_PASSWD="replpassword"

# Remove M1 and M2
dsctl M1 remove --do-it
dsctl M2 remove --do-it

# Create M1 with BDB
cat > /tmp/m1.inf << EOF
[general]
config_version = 2
full_machine_name = localhost

[slapd]
instance_name = M1
port = 38901
secure_port = 63601
root_dn = cn=Directory Manager
root_password = password
db_lib = bdb

[backend-userroot]
suffix = dc=example,dc=com
sample_entries = yes
EOF
dscreate from-file /tmp/m1.inf

# Create M2 with MDB
cat > /tmp/m2.inf << EOF
[general]
config_version = 2
full_machine_name = localhost

[slapd]
instance_name = M2
port = 38902
secure_port = 63602
root_dn = cn=Directory Manager
root_password = password
db_lib = mdb

[backend-userroot]
suffix = dc=example,dc=com
sample_entries = yes
EOF
dscreate from-file /tmp/m2.inf

echo "Creating VLV index on description attribute..."

# Create VLV search and index on M1
dsconf M1 backend vlv-index add-search userRoot \
  --name "vlvDescSearch" \
  --search-base "$SUFFIX" \
  --search-scope 2 \
  --search-filter "(objectClass=inetOrgPerson)"

dsconf M1 backend vlv-index add-index userRoot \
  --parent-name "vlvDescSearch" \
  --index-name "vlvDescIndex" \
  --sort "description" \
  --index-it

# Create same VLV search and index on M2 (so it builds during bulk import)
dsconf M2 backend vlv-index add-search userRoot \
  --name "vlvDescSearch" \
  --search-base "$SUFFIX" \
  --search-scope 2 \
[...]

PR #7258, opened 2026-02-13
Issue 7213 - (2nd) MDB_BAD_VALSIZE error while handling VLV

vashirov:
Decription:
Disable test_vlv_long_attribute_value on BDB as it hangs sometimes in CI, blocking other pipelines.

Relates: https://github.com/389ds/389-ds-base/issues/7213

## Summary by Sourcery

Tests:
- Mark the VLV long attribute value test as skipped on BDB backends due to lack of support and intermittent hangs in CI.


progier389 (approved):
LGTM (but IMHO we should better also add a comment telling that this test hangs on bdb)

Issue #7265, opened 2026-02-18
changelog maxage validation is not strict enough


mreynolds389:
**Issue Description**

Currently you can specify a changelog maxage without a duration unit, like seconds, minutes, hours, and weeks.  This can cause issues with replication if the value is too small.  We need to enforce a proper duration unit is set.

PR #7266, opened 2026-02-18
Issue 7265 - changelog maxage validation is not strict enough

mreynolds389:
Description:

We need to enforce a duration unit is set and the first digit is not zero when setting maxage for the replication and retro changelogs.

relates: https://github.com/389ds/389-ds-base/issues/7265

## Summary by Sourcery

Enforce stricter validation of changelog maxage values for replication and retro changelog configuration and adjust tests accordingly.

Bug Fixes:
- Reject changelog maxage values without a duration unit, with non-digit characters in the numeric portion, or starting with zero for replication and retro changelog configuration.

Enhancements:
- Introduce a strict duration validation helper used by changelog and retro changelog maxage handling.
- Clarify the changelog maxage configuration error message to describe the valid range and required duration unit.

Tests:
- Expand replication and retro changelog tests to cover accepted and rejected maxage formats, including unit presence, case handling, and leading-zero and invalid-unit cases.

### DIFF
--- a/dirsrvtests/tests/suites/replication/changelog_test.py
+++ b/dirsrvtests/tests/suites/replication/changelog_test.py
@@ -570,8 +570,10 @@ def test_changelog_maxage(topo, changelog_init):
             '/var/lib/dirsrv/slapd-supplier1/changelog' and
             set cn=Retro Changelog Plugin,cn=plugins,cn=config to 'on'
     :steps:
-        1. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values - '12345','10s','30M','12h','2D','4w'
-        2. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values - '-123','xyz'
+        1. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values:
+           '100s','100S','30m','30M','12h','12H','2d','2D','4w','4W'
+        2. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values:
+           '12345', 'd', '-123', '-123d', '345345xyz', '0d', 'xyz'
 
     :expectedresults:
         1. Operation should be successful
@@ -583,13 +585,22 @@ def test_changelog_maxage(topo, changelog_init):
     topo.ms["supplier1"].log.info("Bind as %s" % DN_DM)
     topo.ms["supplier1"].simple_bind_s(DN_DM, PASSWORD)
 
-    add_and_check(topo, CHANGELOG, MAXAGE, '12345', True)
-    add_and_check(topo, CHANGELOG, MAXAGE, '10s', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '100s', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '100S', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '30m', True)
     add_and_check(topo, CHANGELOG, MAXAGE, '30M', True)
     add_and_check(topo, CHANGELOG, MAXAGE, '12h', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '12H', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '2d', True)
     add_and_check(topo, CHANGELOG, MAXAGE, '2D', True)
     add_and_check(topo, CHANGELOG, MAXAGE, '4w', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '4W', True)
+    add_and_check(topo, CHANGELOG, MAXAGE, '12345', False)
+    add_and_check(topo, CHANGELOG, MAXAGE, 'd', False)
     add_and_check(topo, CHANGELOG, MAXAGE, '-123', False)
+    add_and_check(topo, CHANGELOG, MAXAGE, '-123d', False)
+    add_and_check(topo, CHANGELOG, MAXAGE, '345345xyz', False)
+    add_and_check(topo, CHANGELOG, MAXAGE, '0d', False)
     add_and_check(topo, CHANGELOG, MAXAGE, 'xyz', False)
 
 
@@ -668,9 +679,9 @@ def test_retrochangelog_maxage(topo, changelog_init):
             set cn=Retro Changelog Plugin,cn=plugins,cn=config to 'on'
     :steps:
         1. Set nsslapd-changelogmaxage in cn=Retro Changelog Plugin,cn=plugins,cn=config to values -
-           '12345','10s','30M','12h','2D','4w'
+           '100s','100S','30m','30M','12h','12H','2d','2D','4w','4W'
         2. Set nsslapd-changelogmaxage in cn=Retro Changelog Plugin,cn=plugins,cn=config to values -
-           '-123','xyz'
+           '12345', 'd', '-123', '-123d', '345345xyz', '0d', 'xyz'
 
     :expectedresults:
         1. Operation should be successful
@@ -682,13 +693,22 @@ def test_retrochangelog_maxage(topo, changelog_init):
     topo.ms["supplier1"].log.info("Bind as %s" % DN_DM)
     topo.ms["supplier1"].simple_bind_s(DN_DM, PASSWORD)
 
-    add_and_check(topo, RETROCHANGELOG, MAXAGE, '12345', True)
-    add_and_check(topo, RETROCHANGELOG, MAXAGE, '10s', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '100s', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '100S', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '30m', True)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, '30M', True)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, '12h', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '12H', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '2d', True)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, '2D', True)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, '4w', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '4W', True)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '12345', False)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, 'd', False)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, '-123', False)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '-123d', False)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '345345xyz', False)
+    add_and_check(topo, RETROCHANGELOG, MAXAGE, '0d', False)
     add_and_check(topo, RETROCHANGELOG, MAXAGE, 'xyz', False)
 
     topo.ms["supplier1"].log.info("ticket47669 was successfully verified.")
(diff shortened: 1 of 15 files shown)


dirsrvtests/tests/suites/replication/changelog_test.py
vashirov: Test docstring should be updated to reflect new tested values.

src/cockpit/389-console/src/lib/plugins/retroChangelog.jsx
vashirov, line 73: I can decrement below zero using down key in the input field.

src/cockpit/389-console/src/lib/replication/replChangelog.jsx
vashirov, line 46: Shouldn't this be 1? onChange handler uses it to check min value.

vashirov (changes requested):
1. Other tests that set changelog maxage without units need to be updated, otherwise they fail:
dirsrvtests/tests/suites/replication/changelog_trimming_test.py
dirsrvtests/tests/suites/replication/regression_m2c1_test.py
dirsrvtests/tests/suites/replication/regression_m3_test.py
dirsrvtests/tests/suites/healthcheck/healthcheck_test.py


2. If an instance had nsslapd-changelogmaxage configured without units, server logs a message and sets the default value:
```
[19/Feb/2026:09:16:47.409827169 +0000] - NOTICE - NSMMReplicationPlugin - changelog program - changelog5_extract_config - nsslapd-changelogmaxage: invalid value "(null)", ignoring the change.
```

And for RetroCL trimming gets completely disabled:
```
[19/Feb/2026:09:20:28.272738111 +0000] - ERR - DSRetroclPlugin - retrocl_init_trimming: ignoring invalid nsslapd-changelogmaxage value 3600; not trimming retro changelog.
```

Instead of ignoring perhaps these should be converted to values with units? I.e. instead of 3600 set 3600s.

3. `dsconf` still accepts unit-less values, help does't show that units are required:
```
# dsconf localhost replication set-changelog --help  | grep max-age
                                                           [--max-age MAX_AGE]
  --max-age MAX_AGE     Set the maximum age of a replication changelog entry

# dsconf localhost plugin retro-changelog set --help | grep max-age -A2
       [--directory DIRECTORY] [--max-age MAX_AGE]
       [--trim-interval TRIM_INTERVAL] [--exclude-suffix [EXCLUDE_SUFFIX ...]]
       [--exclude-attrs [EXCLUDE_ATTRS ...]]
--
  --max-age MAX_AGE     Specifies the maximum age of any entry in the
                        changelog. Used to trim the changelog (nsslapd-
                        changelogmaxage)
```

IMHO, we should harden dsconf instead of enforcing on the server side.

mreynolds389:
> 1. Other tests that set changelog maxage without units need to be updated, otherwise they fail:
>    dirsrvtests/tests/suites/replication/changelog_trimming_test.py
>    dirsrvtests/tests/suites/replication/regression_m2c1_test.py
>    dirsrvtests/tests/suites/replication/regression_m3_test.py
>    dirsrvtests/tests/suites/healthcheck/healthcheck_test.py
> 2. If an instance had nsslapd-changelogmaxage configured without units, server logs a message and sets the default value:
> 
> ```
> [19/Feb/2026:09:16:47.409827169 +0000] - NOTICE - NSMMReplicationPlugin - changelog program - changelog5_extract_config - nsslapd-changelogmaxage: invalid value "(null)", ignoring the change.
> ```
> 
> And for RetroCL trimming gets completely disabled:
> 
> ```
> [19/Feb/2026:09:20:28.272738111 +0000] - ERR - DSRetroclPlugin - retrocl_init_trimming: ignoring invalid nsslapd-changelogmaxage value 3600; not trimming retro changelog.
> ```
> 
> Instead of ignoring perhaps these should be converted to values with units? I.e. instead of 3600 set 3600s.
> 
> 3. `dsconf` still accepts unit-less values, help does't show that units are required:
> 
> ```
> # dsconf localhost replication set-changelog --help  | grep max-age
>                                                            [--max-age MAX_AGE]
>   --max-age MAX_AGE     Set the maximum age of a replication changelog entry

    # dsconf localhost replication set-changelog --max-age 10 --suffix dc=example,dc=com
    Error: Server is unwilling to perform(53) - nsslapd-changelogmaxage: invalid value "10", nsslapd-changelogmaxage must range from 1 to 9223372036854775807 and end with a duration unit[sSmMhHdDwW]

But retro changelog does accept it.

> 
> IMHO, we should harden dsconf instead of enforcing on the server side.

Best to have it on both since people still use ldapmodify, but lib389 will more involved since we use generic args_to_attrs for most of these settings.

Anyway, I'll work on these issues you pointed out. Thanks!

mreynolds389:
> > > But retro changelog does accept it.
> > 
> > 
> > This is weird because the CI tests for invalid max-age "passes" for the retro changelog. I'll be looking into this as well.
> 
> I installed an old build, configured maxage, then upgraded the instance. This error was during startup.

dsconf still works though:

    # dsconf localhost plugin retro-changelog set  --max-age 12
    Successfully changed the cn=Retro Changelog Plugin,cn=plugins,cn=config

src/cockpit/389-console/src/lib/replication/replChangelog.jsx
vashirov: If I have a value without units in the config, UI cuts off last character, i.e. I have 3650 in dse.ldif, UI shows 365. The same applies to RetroCL.

dirsrvtests/tests/suites/replication/regression_m2c1_test.py
vashirov:
```
ValueError: invalid literal for int() with base 10: '5s'
```

vashirov (commented):
Added few more comments.

I still think that we shouldn't discard previously configured values:
```
[19/Feb/2026:20:13:28.898552173 +0000] - NOTICE - NSMMReplicationPlugin - changelog program - changelog5_extract_config - nsslapd-changelogmaxage: invalid value "3600", trimming will not be done based on age.
[19/Feb/2026:20:13:28.901298333 +0000] - ERR - DSRetroclPlugin - retrocl_init_trimming: ignoring invalid nsslapd-changelogmaxage value 12345; not trimming retro changelog.
```
If a user had maxage configured without units, changelog trimming based on age won't be performed after upgrade. We need to have an upgrade path.


In the UI RetroCL allows to set 0 and -1 as maxage, which fails CLI validation:
```
Danger alert:
Failed to update Retro Changelog Plugin - Invalid max age value: 0
```

ldap/servers/plugins/retrocl/retrocl_trim.c
```diff
@@ -431,15 +431,15 @@ retrocl_init_trimming(void)
     int trim_interval = DEFAULT_CHANGELOGDB_TRIM_INTERVAL;
 
     cl_maxage = retrocl_get_config_str(CONFIG_CHANGELOG_MAXAGE_ATTRIBUTE);
-    if (cl_maxage) {
-        if (slapi_is_duration_valid(cl_maxage)) {
+    if (cl_maxage && strcmp(cl_maxage, "0") != 0) {
+        if (slapi_is_duration_valid_strict(cl_maxage)) {
```
vashirov, line 435: One remaining issue is upgrading from an instance that uses legacy unitless values. Currently they are rejected and basically turn off trimming based on age, which can lead to changelog size growth.
Instead of rejecting, we can fallback to `slapi_is_duration_valid()` and treat values as seconds, logging a WARN that the value was accepted as seconds.

vashirov (commented):
@mreynolds389, that's fine, there is no rush. Thanks for addressing dsconf issues.

ldap/servers/plugins/retrocl/retrocl_trim.c
mreynolds389: > One remaining issue is upgrading from an instance that uses legacy unitless values. Currently they are rejected and basically turn off trimming based on age, which can lead to changelog size growth. Instead of rejecting, we can fallback to `slapi_is_duration_valid()` and treat values as seconds, logging a WARN that the value was accepted as seconds.

Oh yeah I'll get this in.

mreynolds389: @vashirov Ok at server startup we accept missing duration unit, assume seconds,and log a message.  Also updated UI to handle missing duration unit and display proper values/units.

vashirov (approved):

src/lib389/lib389/cli_conf/plugins/retrochangelog.py
```diff
@@ -46,16 +52,17 @@ def _add_parser_args(parser):
                         help='Specifies the name of the directory in which the changelog database '
                              'is created the first time the plug-in is run')
     parser.add_argument('--max-age',
-                        help='Specifies the maximum age of any entry in the changelog.  Used to trim the '
-                            'changelog (nsslapd-changelogmaxage)')
+                        help='Specifies the maximum age of any entry in the changelog. '
+                             'The value must be a number followed by a duration unit [sSmMhHwW]. '
```
droideck: The helper text misses `dD` in `[sSmMhHwW]`, IIUC.

src/lib389/lib389/cli_conf/replication.py
```diff
@@ -1525,7 +1530,9 @@ def create_parser(subparsers):
     repl_set_per_backend_cl.set_defaults(func=set_per_backend_cl)
     repl_set_per_backend_cl.add_argument('--suffix', required=True, help='Sets the suffix that uses the changelog')
     repl_set_per_backend_cl.add_argument('--max-entries', help="Sets the maximum number of entries to get in the replication changelog")
-    repl_set_per_backend_cl.add_argument('--max-age', help="Set the maximum age of a replication changelog entry")
+    repl_set_per_backend_cl.add_argument('--max-age',
+                                         help='Specifies the maximum age of any entry in the changelog. '
+                                              'The value must be a number followed by a duration unit [sSmMhHwW].')
```
droideck: And here.

dirsrvtests/tests/suites/replication/changelog_test.py
```diff
@@ -570,8 +570,10 @@ def test_changelog_maxage(topo, changelog_init):
             '/var/lib/dirsrv/slapd-supplier1/changelog' and
             set cn=Retro Changelog Plugin,cn=plugins,cn=config to 'on'
     :steps:
-        1. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values - '12345','10s','30M','12h','2D','4w'
-        2. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values - '-123','xyz'
+        1. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values:
+           '100s','100S','30m','30M','12h','12H','2d','2D','4w','4W'
+        2. Set nsslapd-changelogmaxage in cn=changelog5,cn=config to values:
+           '12345', 'd', '-123', '-123d', '345345xyz', '0d', 'xyz'
```
droideck, line 576: Minor thing, but the test also tests  '12345' and 'd' which are not mentioned here.

droideck (approved):
LGTM

src/lib389/lib389/cli_conf/plugins/retrochangelog.py
mreynolds389: fixed

dirsrvtests/tests/suites/replication/changelog_test.py
mreynolds389: fixed

mreynolds389:
> > > Another thing...
> > > I don't have a strong opinion here, but should we also add a SLAPI_OPERATION_MODIFY DSE callback for the retro changelog plugin to validate nsslapd-changelogmaxage at the LDAP level, similar to what cldb_config_modify does for the replication changelog?
> > 
> > 
> > I hear what you are saying, but it still get validated on a modify operation. Since this PR is already getting large I'd prefer to do it in another PR or just skip it as it's not needed.
> 
> I don't think it's validated for modify operation for the retro changelog. CLI and UI -- yes, but nothing for direct LDAP call, unless I miss something. But the replication changelog has it.

Oh...  Yeah we "validate" it at a later stage, but we allow the modify to occur.  Yeah lets do that in a separate PR, but I'll work on it right away.

PR #7268, opened 2026-02-19
Issue 7267 - MDB_BAD_VALSIZE error when updating index

### LINKED ISSUE #7267: MDB_BAD_VALSIZE error when updating index
**Issue Description**

With lmdb: Add/Modify/Import operation fails if an indexed attributes has lot of relatively long values
(for example 500 values, each value containing 512 bytes.) and error log shows some MDB_BAD_VALSIZE errors:

[19/Feb/2026:17:11:05.441692295 +0100] - ERR - dbmdb_map_error - dbmdb_public_db_op failed with db error -30781 : MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size
 

 - Platform: Fedora

**Steps to Reproduce**
1. Try to add an entry with an heavily multi valued having long values
2. See error (operation fails and MDB_BAD_VALSIZE in error logs)

**Additional context**
https://issues.redhat.com/browse/RHEL-133085

progier389:
Problem with the key prefix handling when key is too long and must be hashed. 
The issue is that the # that is prepended is not reset when iterating over the valueset values (Ending up with very long prefix)

Also refactored the code to avoid duplicate the code that prepare the key from the attribute value (used when updating the index or retrieving a value from an index)

Issue: #7267 

Reviewed by: @tbordaz , @vashirov (Thanks!)

## Summary by Sourcery

Fix handling of oversized index keys by centralizing key preparation and improve diagnostics for MDB import and indexing failures.

Bug Fixes:
- Correct index key prefix handling when long attribute values require hashing to prevent MDB_BAD_VALSIZE and malformed keys.
- Ensure large index keys are hashed based on full key length including prefix rather than value length alone.

Enhancements:
- Introduce a shared helper to prepare index keys (including hashing and encryption) for both index updates and reads, reducing duplication and inconsistencies.
- Add detailed logging and hex dumps for MDB import write, transaction, and queue element failures to aid debugging.
- Expose a reusable hex-dump logging utility and replace backend-specific implementations with the shared version.

Tests:
- Add a regression test that creates and validates a user with 512 large sn attribute values to cover large multi-valued index scenarios.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -104,6 +104,8 @@ typedef unsigned short u_int16_t;
  */
 #define BE_CHANGELOG_FILE     "replication_changelog"
 
+#define INDEX_KEY_LENGHT(lenval,lenprefix)  (lenval+lenprefix+2)
+
 #define BDB_IMPL              "bdb"
 #define BDB_BACKEND           "libback-ldbm" /* This backend plugin */
 #define BDB_NEWIDL            "newidl"       /* new idl format */
--- a/ldap/servers/slapd/back-ldbm/back-ldbm.h
+++ b/ldap/servers/slapd/back-ldbm/back-ldbm.h
@@ -104,7 +104,7 @@ typedef unsigned short u_int16_t;
  */
 #define BE_CHANGELOG_FILE     "replication_changelog"
 
-#define INDEX_KEY_LENGHT(lenval,lenprefix)  (lenval+lenprefix+2)
+#define INDEX_KEY_LENGTH(lenval,lenprefix)  (lenval+lenprefix+2)
 
 #define BDB_IMPL              "bdb"
 #define BDB_BACKEND           "libback-ldbm" /* This backend plugin */
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
@@ -1123,6 +1123,21 @@ dbmdb_import_entry_info_by_backentry(mdb_privdb_t *db, BulkQueueData_t *bqdata, 
     return dnrc;
 }
 
+/* Log wqelmt details */
+void
+log_wqelmt(int loglvl, char *fname, WorkerQueueData_t *wqelmt)
+{
+    if (wqelmt->dn) {
+        slapi_log_err(loglvl, fname, "log_wqelmt: dn=%s\n", wqelmt->dn);
+    }
+    if (wqelmt->filename && wqelmt->lineno) {
+        slapi_log_err(loglvl, fname, "log_wqelmt: ldif=%s[%d]\n", wqelmt->filename, wqelmt->lineno);
+    }
+    if (wqelmt->data) {
+        size_t len = wqelmt->datalen ? wqelmt->datalen : strlen(wqelmt->data);
+        slapi_log_hexadump(loglvl, "log_wqelmt:data", wqelmt->data, len);
+    }
+}
 
 /* producer thread for ldif import case:
  * read through the given file list, parsing entries (str2entry), assigning
@@ -1255,6 +1270,7 @@ dbmdb_import_producer(void *param)
                 import_log_notice(job, SLAPI_LOG_ERR, "dbmdb_import_producer",
                                   "ns_slapd software error: unexpected dbmdb_import_entry_info return code: %d.",
                                   wqelmt.dnrc);
+                log_wqelmt(SLAPI_LOG_ERR, "dbmdb_import_producer", &wqelmt);
                 abort();
             case DNRC_OK:
             case DNRC_SUFFIX:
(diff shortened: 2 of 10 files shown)


ldap/servers/slapd/back-ldbm/index.c
progier389, line 835: infinite loop [CWE-835]  warning is a false/positive because ptin increase at each loops 
(because the switch condition is always 1, 2 or 3

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -104,6 +104,8 @@ typedef unsigned short u_int16_t;
  */
 #define BE_CHANGELOG_FILE     "replication_changelog"
```
progier389, line 105: Next warning is true but harmless (and code was not modified by this PR)
dblayer_open_file  filename parameter should better be "const char *"

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -3965,6 +3999,10 @@ dbmdb_import_writer(void*param)
         MDB_STAT_STEP(stats, MDB_STAT_RUN, stats_enabled);
         if (!rc) {
             txn = NULL;
+        } else {
```
tbordaz, line 4002: This log is similar to the previous added one. Could it be ambiguous which TXN_COMMIT failed ?

ldap/servers/slapd/back-ldbm/index.c
```diff
@@ -881,6 +881,67 @@ index_read(
...
+    if (hashed_bvp) {
+        prefix = slapi_ch_smprintf("%c%s",HASH_PREFIX, prefix);
+        plen++;
+    }
+    if (buf && buflen) {
+        if (plen+bvp->bv_len+1 > *buflen) {
+            *buflen = plen+bvp->bv_len+1;
+            *buf = slapi_ch_realloc(*buf, *buflen);
+        }
+        dblayer_value_concat(be, key, *buf, *buflen, prefix, plen, bvp->bv_val, bvp->bv_len, "", 1);
```
tbordaz, line 927: It looks to me that 'prefix' buffer is concat to 'buf'. Shouldn't it be the opposite, 'buf' concat to 'prefix' ?

tbordaz (commented):
LGTM
couple of additional questions

ldap/servers/slapd/back-ldbm/back-ldbm.h
```diff
@@ -104,6 +104,8 @@ typedef unsigned short u_int16_t;
...
 
+#define INDEX_KEY_LENGHT(lenval,lenprefix)  (lenval+lenprefix+2)
```
vashirov: Typo, should be INDEX_KEY_LENGTH
```suggestion
#define INDEX_KEY_LENGTH(lenval,lenprefix)  (lenval+lenprefix+2)
```

ldap/servers/slapd/back-ldbm/index.c
```diff
@@ -881,6 +881,67 @@ index_read(
...
+{
+    /* Key format is [Hash?] [prefix] [val] [\0] */
+    struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
+    size_t plen = strlen(prefix);
+    struct berval *hashed_bvp = NULL;
+    struct berval *encrypted_bvp = NULL;
+    int rc = 0;
+
+    /* Hash large index key if necessary */
+    if (INDEX_KEY_LENGHT(bvp->bv_len,plen) >=  li->li_max_key_len) {
```
vashirov:
```suggestion
    if (INDEX_KEY_LENGTH(bvp->bv_len,plen) >=  li->li_max_key_len) {
```

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt.c
```diff
@@ -1079,15 +1079,15 @@ attrcrypt_decrypt_index_key(backend *be,
...
  */
 int
-attrcrypt_hash_large_index_key(backend *be, char **prefix, struct attrinfo *ai, const struct berval *in, struct berval **out)
+attrcrypt_hash_large_index_key(backend *be, const char *prefix, struct attrinfo *ai, const struct berval *in, struct berval **out)
 {
     int ret = 0;
     struct berval *out_berval = NULL;
     struct ldbminfo *li = (struct ldbminfo *)be->be_database->plg_private;
-    char *new_prefix;
+    size_t final_key_len = INDEX_KEY_LENGHT(in->bv_len, strlen(prefix));
```
vashirov:
```suggestion
    size_t final_key_len = INDEX_KEY_LENGTH(in->bv_len, strlen(prefix));
```

vashirov (approved):
Besides small typo, LGTM

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389: it is, but I do not think it really matters, the important point is that a txn commit fails.
Meaning that there is a serious system error, the exact location is not really meaningful as it is pretty random

ldap/servers/slapd/back-ldbm/index.c
progier389: Your assumption about dblayer_value_concat is wrong.
 It concat the 3 buffers and set the result in buf.
It does not append the result after buf existing content

```
/* Concat all data/size pairs into a dbi_val_t
 * value is either set from buffer (if it is large enough) or it is alloced
 * (Cannot use varargs here because it leads to a crash on Suze)
 */
void dblayer_value_concat(Slapi_Backend *be, dbi_val_t *data,
    void *buf, size_t buflen, const char *str1, size_t len1,
    const char *str2, size_t len2, const char *str3, size_t len3)
```

tbordaz (approved):
Thanks for the feeback. LGTM

Issue #7271, opened 2026-02-23
plugins that create threads need to update active thread count


mreynolds389:
**Issue Description**

 Plugins that create threads need to up to the global active thread count.  Otherwise when the server is being stopped the plugin's close function gets called while these threads are still running and still using the plugin configuration.  This can lead to crashes.

vashirov:
I've got another hang in `dirsrvtests/tests/suites/vlv/regression_test.py`, now in cos_cache and roles_cache:
[pstack-20260304-121353.txt](https://github.com/user-attachments/files/25740501/pstack-20260304-121353.txt)

Previously test would fail, but now it passes, although the supplier1 or supplier2 gets killed after systemd timeout.

mreynolds389:
> [@mreynolds389](https://github.com/mreynolds389), yes, I'm sure. I'm using 389-ds-base: 3.2.0.202603041940git761452f79-1.fc43 from the nightly copr repo. Test doesn't fail, but it takes about 36 mins to complete. Server gets killed with SIGABRT by timeout. It should be visible in `coredumpctl list` output.

I do see sigaborts in that output from the last run, but that test ran in under 2 minutes.  It was normal.  So I don't trust that output as a way to verify if the fix is working.  Anyway I'll keep trying to reproduce it and watch the run times.  I wonder if that sigabort is the PR_ASSERT that my recent PR fixes.  Was your copr build a debug build?

PR #7272, opened 2026-02-23
Issue 7271 - plugins that create threads need to update active thread count

mreynolds389:
… count

Description:

Plugins that create threads need to up to the global active thread count. Otherwise when the server is being stopped the plugin's close function gets called while these threads are still running and still using the plugin configuration. This can lead to crashes.

relates: https://github.com/389ds/389-ds-base/issues/7271

## Summary by Sourcery

Track background plugin threads in the global active thread count to prevent shutdown races with still-running plugin operations.

Bug Fixes:
- Ensure sync plugin result-sending threads update the global active thread count while running.
- Ensure replication protocol worker threads are accounted for in the global active thread count.
- Ensure retro changelog trimming and replica tombstone reaper threads are tracked in the global active thread count to avoid plugin-use-after-close crashes.

Enhancements:
- Perform minor whitespace and formatting cleanups in the sync plugin code.

### DIFF
--- a/ldap/servers/plugins/retrocl/retrocl_trim.c
+++ b/ldap/servers/plugins/retrocl/retrocl_trim.c
@@ -243,6 +243,8 @@ trim_changelog(void)
 
     now_interval = slapi_current_rel_time_t(); /* monotonic time for interval */
 
+    g_incr_active_threadcnt();
+
     PR_Lock(ts.ts_s_trim_mutex);
     max_age = ts.ts_c_max_age;
     trim_interval = ts.ts_c_trim_interval;
@@ -257,7 +259,7 @@ trim_changelog(void)
          */
         done = 0;
         now_maxage = slapi_current_utc_time(); /* real time for trim candidates */
-        while (!done && retrocl_trimming == 1) {
+        while (!done && retrocl_trimming == 1 && !slapi_is_shutting_down()) {
             int did_delete;
 
             did_delete = 0;
@@ -309,6 +311,9 @@ trim_changelog(void)
                       "trim_changelog: removed %d change records\n",
                       num_deleted);
     }
+
+    g_decr_active_threadcnt();
+
     return rc;
 }
 
--- a/ldap/servers/plugins/replication/repl5_protocol.c
+++ b/ldap/servers/plugins/replication/repl5_protocol.c
@@ -23,6 +23,7 @@
 
 #include "repl5.h"
 #include "repl5_prot_private.h"
+#include "slap.h"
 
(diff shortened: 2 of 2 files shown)


ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -3246,6 +3246,8 @@ _replica_reap_tombstones(void *arg)
                   "_replica_reap_tombstones - Beginning tombstone reap for replica %s.\n",
                   replica_name ? replica_name : "(null)");
 
+    g_incr_active_threadcnt();
```
progier389: FYI: probably not really needed here because the agmt threads are explicitly stopped when stopping the backend.
But anyway increasing the thread count does not harm

progier389 (approved):
LGTM

ldap/servers/plugins/replication/repl5_replica.c
mreynolds389: I was just being overly paranoid :-)

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -1091,6 +1095,8 @@ sync_send_results(void *arg)
     }
     slapi_ch_free((void **)&req);
     thread_count--;
+
```
tbordaz: similar remark as the one done by @progier389 . sync_persist_terminate_all is handling the sync_persist threads count when the sync_plugin is stopped. So it is probably not necessary but there is no harm

ldap/servers/plugins/retrocl/retrocl_trim.c
```diff
@@ -243,6 +243,8 @@ trim_changelog(void)
 
     now_interval = slapi_current_rel_time_t(); /* monotonic time for interval */
 
+    g_incr_active_threadcnt();
```
tbordaz, line 246: I wonder if it is enough. Trim_changelog is not testing shutdown and continue until it completed its job.
We may test shutdown in the loop else systemd may kill us abruptly.

Also I wonder if we can not use retrocl_active_threads in the shutdown callback rather than using the global thread count.

mreynolds389: Currently retrocl_active_threads does nothing.  Comments say it was added for debugging but it literally serves no purpose at the moment.  What is nice about the global thread count is that it prevents the plugin close functions from being called.  Otherwise we need to add that duplicate logic to the close function for it to work.

But yes we should check if slapd is shutting down to exit out of the retro trim loop.  I just pushed this change

Issue #7273, opened 2026-02-23
In a chaining environment binding as remote user causes an invalid error in the logs


mreynolds389:
**Issue Description**

In a database link/chaining environment you can bind as a remote user, and this triggers an error when trying to "upgrade_on_bind" as the user does not locally have a userpassword since it's remote.  There is no strong case to log an error in this situation.

PR #7274, opened 2026-02-23
Issue 7273 - In a chaining environment binding as remote user causes an invalid error in the logs

mreynolds389:
Description:

In a database link/chaining environment you can bind as a remote user, and this triggers an error when trying to "upgrade_on_bind" as the user does not locally have a userpassword since it's remote. There is no strong case to log an error in this situation.

relates: http://github.com/389ds/389-ds-base/issues/7273

## Summary by Sourcery

Bug Fixes:
- Avoid emitting spurious warning logs when binding as a remote user without a local userpassword attribute in chaining/database link environments.


vashirov (approved):
LGTM

Issue #7281, opened 2026-02-24
RFE - add support to managing additional encryption modules


mreynolds389:
**Issue Description**

Right now there is no way to add/manage additional encryption modules through the CLI/UI.  We need this for allowing multiple server certificates.

PR #7282, opened 2026-02-24
Issue 7281 - RFE - CLI - add support to managing additional encryption modules

mreynolds389:
Description:

Right now there is no way to add/manage additional encryption modules through the CLI/UI. We need this for allowing multiple server certificates.

relates: https://github.com/389ds/389-ds-base/issues/7281

## Summary by Sourcery

Add CLI and backend support for managing additional TLS encryption modules and multiple server certificates.

New Features:
- Introduce security encryption-module CLI commands to add, edit, delete, list, and get encryption modules.
- Add EncryptionModule and EncryptionModules configuration objects to manage nsEncryptionModule entries under cn=encryption,cn=config.
- Allow specifying a custom certificate name when importing server key/cert pairs and when creating or importing RSA keys and certificates.

Enhancements:
- Extend TLS/NSS helper methods to accept configurable certificate nicknames instead of using a single hardcoded name.

### DIFF
--- a/src/lib389/lib389/cli_conf/security.py
+++ b/src/lib389/lib389/cli_conf/security.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2023 Red Hat, Inc.
+# Copyright (C) 2026 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
@@ -8,9 +8,10 @@
 
 from collections import OrderedDict, namedtuple
 import json
+import ldap
 import os
 import sys
-from lib389.config import Config, Encryption, RSA
+from lib389.config import Config, Encryption, RSA, EncryptionModules
 from lib389.nss_ssl import NssSsl
 from lib389.cert_manager import CertManager
 from lib389.cli_base import _warn, CustomHelpFormatter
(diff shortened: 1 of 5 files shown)


src/lib389/lib389/cli_conf/security.py
```diff
@@ -619,6 +730,49 @@ def create_parser(subparsers):
...
+        description=('Delete an encryption module from the encryption module list.'))
+    encryption_module_delete_parser.add_argument('name', help='The name of the encryption module')
+    encryption_module_delete_parser.set_defaults(func=encryption_module_delete)
+    # edit an encryption module
+    encryption_module_edit_parser = encryption_module_sub.add_parser('edit', help='Edit an encryption module',
+        description=('Edit an encryption module in the encryption module list.'))
+    encryption_module_edit_parser.add_argument('name', help='The name of the encryption module')
+    encryption_module_edit_parser.add_argument('--cert-nickname', help='The personality or nickname of the server certificate')
+    encryption_module_edit_parser.add_argument('--activate', action='store_true', help='The activate the encryption module')
+    encryption_module_edit_parser.add_argument('--deactivate', action='store_true', help='The deactivate the encryption module')
```
vashirov:
```suggestion
    encryption_module_edit_parser.add_argument('--activate', action='store_true', help='Activate the encryption module')
    encryption_module_edit_parser.add_argument('--deactivate', action='store_true', help='Deactivate the encryption module')
```

src/lib389/lib389/cli_conf/security.py
```diff
@@ -514,6 +514,117 @@ def export_cert(inst, basedn, log, args):
...
+    encryption_modules = EncryptionModules(instance=inst)
+    encryption_modules.create(properties=properties)
+    log.info("Successfully added encryption module")
+
+
+def encryption_module_delete(inst, basedn, log, args):
+    """Delete an encryption module
+    """
+    if args.name.lower() == "rsa":
+        log.error("Cannot delete RSA encryption module")
```
vashirov, line 540: In the future we should be able to delete RSA and keep only ECDSA or MLDSA modules. But, IIUC, we have some hardcoded DNs for attribute encryption and SSL initialization that require RSA to be present.

src/lib389/lib389/cli_ctl/tls.py
vashirov, line 76: import_server_cert always imports it as Server-Cert

vashirov: CSR is always generated for Server-Cert

vashirov: When I import server certs with different names, `tls show-server-cert` still expects it as Server-Cert, and I can't override it:
```
# dsctl localhost tls show-server-cert
Error: certutil: Could not find cert: Server-Cert
: PR_FILE_NOT_FOUND_ERROR: File not found
```

vashirov (changes requested):

src/lib389/lib389/cli_ctl/tls.py
```diff
@@ -13,7 +13,10 @@
 
 def show_servercert(inst, log, args):
     tls = NssSsl(dirsrv=inst)
-    log.info(tls.display_cert_details(CERT_NAME))
+    nickname = args.nickname
```
progier389: I would rather use:
    nickname = getattr(args, 'nickname', CERT_NAME)
to avoid exception if some test call directly show_servercert without nickname

Same remark for list_client_cas, generate_key_csr, import_server_cert

src/lib389/lib389/cli_ctl/tls.py
```diff
@@ -59,21 +60,26 @@ def import_key_cert_pair(inst, log, args):
     tls = NssSsl(dirsrv=inst)
     key_path = args.key_path
     cert_path = args.cert_path
-    tls.add_server_key_and_cert(key_path, cert_path)
+    cert_name = args.cert_name
```
droideck: I think that should be `args.nickname`?

src/lib389/lib389/cli_conf/security.py
```diff
@@ -514,6 +514,117 @@ def export_cert(inst, basedn, log, args):
...
+        return
+
+    encryption_module = EncryptionModules(instance=inst).get(args.name)
+    if not encryption_module:
```
droideck: I don't think that could ever happen.
DSLdapObjects.get() at _mapped_object.py raises ldap.NO_SUCH_OBJECT when no entry matches -- it never returns a falsy value. So the if not encryption_module block is dead code and the friendly error message will never be shown. Instead, users will see a raw LDAP exception traceback.

Also, the same for encryption_module_edit, and encryption_module_get

src/lib389/lib389/config.py
```diff
@@ -429,6 +429,55 @@ def create(self, rdn=None, properties={'cn': 'RSA', 'nsSSLPersonalitySSL': 'Serv
...
+    """
+
+    def __init__(self, instance):
+        super(EncryptionModules, self).__init__(instance=instance)
+        self._objectclasses = ['nsEncryptionModule']
+        self._filterattrs = ['cn', 'nsSSLPersonalitySSL']
+        self._childobject = EncryptionModule
+        self._basedn = 'cn=encryption,cn=config'
+
+    def create(self, properties={}):
```
droideck: More of a correctness issue than real bug (as we always pass a new `properties` to .create())
But I think the anti-pattern still should be fixed.

The empty dict is shared across all calls. If create() were ever called without arguments, subsequent calls would see the token already set. So the dict carries state between calls which is wrong.
Should be:

```python
 def create(self, properties=None):
     if properties is None:
         properties = {}
```

src/lib389/lib389/cli_conf/security.py
mreynolds389: Good catch, fixed!

src/lib389/lib389/config.py
mreynolds389: Fixed

src/lib389/lib389/cli_ctl/tls.py
mreynolds389: Fixed

src/lib389/lib389/cli_ctl/tls.py
mreynolds389: Fixed

droideck (approved):

src/lib389/lib389/cli_conf/security.py
```diff
@@ -514,6 +515,122 @@ def export_cert(inst, basedn, log, args):
...
+        return
+
+    encryption_module.delete()
+    log.info("Successfully deleted encryption module")
+
+
+def encryption_module_edit(inst, basedn, log, args):
+    """Edit an encryption module
+    """
+    encryption_module = EncryptionModules(instance=inst).get(args.name)
```
vashirov: This should be removed (already wrapped in try/except below):
```suggestion
```

src/lib389/lib389/cli_conf/security.py
```diff
@@ -619,6 +736,50 @@ def create_parser(subparsers):
...
 
+    # Encryption module management (Should replace RSA at some point)
+    encryption_module = security_sub.add_parser('encryption-module', help='Manage encryption modules',
+                                                formatter_class=CustomHelpFormatter)
+    encryption_module_sub = encryption_module.add_subparsers(help='encryption-module')
+    # Add an encryption module
+    encryption_module_add_parser = encryption_module_sub.add_parser('add', help='Add an encryption module',
+            description=('Add a new encryption module to the encryption module list.'))
+    encryption_module_add_parser.add_argument('name', help='The name of the encryption module')
+    encryption_module_add_parser.add_argument('--cert-nickname', help='The personality or nickname of the server certificate')
```
vashirov: I think this should be required, otherwise I get ldap.UNWILLING_TO_PERFORM:
```
# dsconf localhost security encryption-module add ECDSA
Error: Attribute nsSSLPersonalitySSL must not be None
```

mreynolds389: @vashirov originally I had this required, but when I was looking at the UI work for this, and the schema definition, I decided to remove the requirement :-)

vashirov (approved):
LGTM

PR #7283, opened 2026-02-25
Issue 7093 - A password policy can be created even when an identical policy already exists

### LINKED ISSUE #7093: A password policy can be created even when an identical policy already exists.
**Issue Description**
Currently, during password policy creation, if an existing policy entry is found, a MOD_REPLACE operation is silently performed instead of alerting the user. This behaviour makes the operation appear successful but hides the fact that the policy already exists.

**Steps to Reproduce**
1. Setup an instance
2. Create a local password policy
3. Create the same policy again
4. A success message is returned to the user

**Expected results**
The operation should fail.

jchapma:
Description:
Currently, during password policy creation, if an existing policy entry is found, a MOD_REPLACE operation is silently performed instead of alerting the user. This behaviour makes the operation appear successful but hides the fact that the policy already exists.

Fix:
Introduce a strict mode in ensure_state that returns the actual outcome of the operation. Update the UI to handle ensure_state responses and add CI test to verify the behaviour.

Fixes: https://github.com/389ds/389-ds-base/issues/7093

Reviewed by:

## Summary by Sourcery

Improve password policy creation to distinguish between new, updated, and unchanged policies and surface this outcome through CLI, API, and UI.

New Features:
- Expose ensure_state outcome (added, updated, unchanged) on directory objects for consumers to inspect and act upon.

Bug Fixes:
- Prevent silent overwrites when creating a password policy that already exists by treating duplicate creations as no-op updates and reporting them as unchanged.

Enhancements:
- Add a strict ensure_state mode that only applies modifications when policy attributes differ from the existing entry and tracks the resulting status.
- Update password policy CLI to return structured status and messages, and update the Cockpit UI to display accurate notifications for created, updated, and unchanged policies.

Tests:
- Add an integration test to verify duplicate subtree and user password policies are not duplicated, are reported as up to date, and that updates correctly modify existing entries.

### DIFF
--- a/dirsrvtests/tests/suites/password/password_policy_test.py
+++ b/dirsrvtests/tests/suites/password/password_policy_test.py
@@ -15,7 +15,7 @@ import pytest
 import time
 from lib389.backend import Backend
 from lib389.config import Config
-from lib389.cli_conf.pwpolicy import list_policies
+from lib389.cli_conf.pwpolicy import list_policies, create_subtree_policy, create_user_policy, del_local_policy
 from lib389.cli_base import FakeArgs
 from lib389.topologies import topology_st as topo
 from lib389.topologies import topology_m1
@@ -1598,6 +1598,122 @@ def test_pwpolicy_list(topo, request):
             backend_entry.delete()
         except Exception:
             pass
+
+    request.addfinalizer(fin)
+
+def test_duplicate_pwpolicy(topo, request):
+    """Verify that duplicate local password policies are not created
+    or listed for subtree and user password policies, and that updates
+    correctly modify existing policy entries.
+
+    :id: 42847a30-5d86-489c-af70-f6cc361a288b
+    :setup: Standalone
+    :steps:
+        1. Ensure OU exists
+        2. Create a subtree pwp on ou=people
+        3. Attempt to create the same subtree pwp again
+        4. Update the existing subtree pwp with new attributes
+        5. List all local pwps and verify no duplicates exist
+        6. Delete subtree pwp
+        7. Create a test user under ou=people
+        8. Create a user pwp for the test user
+        9. Attempt to create the same user pwp again
+        10. Update the existing user pwp with new attributes
+        11. List all local pwps and verify no duplicates exist
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Duplicate subtree pwp returns a no change message
+        4. Subtree pwp entry is updated
+        5. No duplicate pwp entries
+        6. Success
+        7. Success
+        8. Success
+        9. Duplicate user pwp returns a no change message
+        10. User pwp entry is updated
+        11. No duplicate pwp entries
+    """
+    inst = topo.standalone
+
+    def _cli_args(dn, **extraargs):
+        args = FakeArgs()
+        args.DN = dn
+        args.suffix = False
+        args.json = False
+        args.verbose = False
+        args.pwdchange = None
+        args.passwordlockout = None
+        for key, value in extraargs.items():
+            setattr(args, key, value)
+        return args
+
+    def _assert_no_duplicate_in_logs(logcap):
+        logged_output = []
+        for rec in logcap.outputs:
+            msg = rec.getMessage()
+            for line in msg.splitlines():
+                if line.strip():
+                    logged_output.append(line)
+        assert len(logged_output) == len(set(logged_output))
+
+    # Ensure OU exists
+    ous = OrganizationalUnits(topo.standalone, DEFAULT_SUFFIX)
+    ous.ensure_state(properties={'ou': 'people'})
+    people = ous.get("People")
+    assert people.exists()
+
+    # Add subtree pwp
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn]))
+    assert topo.logcap.contains("Successfully created new password policy")
+
+    # Try to add the same pwp again
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn]))
+    assert topo.logcap.contains("Password policy is already up to date")
+
+    # Update an existing subtree pwp
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn], pwdchange='on', passwordlockout='on'))
+    assert topo.logcap.contains("Password policy successfully updated")
+
+    # Verify there are no duplicates listed
+    topo.logcap.flush()
+    list_policies(inst, None, topo.logcap.log, _cli_args(None))
+    _assert_no_duplicate_in_logs(topo.logcap)
+
+    # Delete subtree pwp
+    topo.logcap.flush()
+    del_local_policy(inst, None, topo.logcap.log, _cli_args([people.dn]))
+    assert topo.logcap.contains("Successfully deleted subtree policy")
+
+    # Create test user
+    test_user = create_user(inst, 'testuser', 'test user', 'User', 'Test', 'Secret', '1234', "ou=people")
+
+    # Add user pwp
+    topo.logcap.flush()
+    create_user_policy(inst, None, topo.logcap.log, _cli_args([test_user.dn]))
+    assert topo.logcap.contains("Successfully created new password policy")
+
+    # Try add the same pwp again
+    topo.logcap.flush()
+    create_user_policy(inst, None, topo.logcap.log, _cli_args([test_user.dn]))
+    assert topo.logcap.contains("Password policy is already up to date")
+
+    # Update an existing user pwp
+    topo.logcap.flush()
+    create_user_policy(inst, None, topo.logcap.log, _cli_args([test_user.dn], pwdchange='on', passwordlockout='on'))
+    assert topo.logcap.contains("Password policy successfully updated")
+
+    # Verify there are no duplicates listed
+    topo.logcap.flush()
+    list_policies(inst, None, topo.logcap.log, _cli_args(None))
+    _assert_no_duplicate_in_logs(topo.logcap)
+
+    # Cleanup
+    def fin():
+            del_local_policy(inst, None, topo.logcap.log, _cli_args([test_user.dn]))
+            test_user.delete()
 
     request.addfinalizer(fin)
 
--- a/dirsrvtests/tests/suites/password/password_policy_test.py
+++ b/dirsrvtests/tests/suites/password/password_policy_test.py
@@ -25,7 +25,7 @@ from lib389.idm.user import UserAccounts, UserAccount
 from lib389._constants import DEFAULT_SUFFIX
 from lib389.pwpolicy import PwPolicyManager, PwPolicyEntries
 from lib389.idm.account import Account
-from lib389.idm.nscontainer import nsContainers
+from lib389.idm.nscontainer import nsContainers, nsContainer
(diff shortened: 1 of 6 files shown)


src/lib389/lib389/pwpolicy.py
droideck, line 237: You only report one status.
If a CoS template gets deleted or corrupted while the pwp_entry is intact. Running create_subtree_policy again:
  - pwp_entry already exists, same attributes -> ENSURE_UNCHANGED
  - cos_template is missing -> gets created (ENSURE_ADDED)
  - cos_pointer_def has wrong cosTemplateDn -> gets fixed (ENSURE_UPDATED)
  - User sees: "Password policy is already up to date" -- misleading

But that's nice that it is repaired!

dirsrvtests/tests/suites/password/password_policy_test.py
```diff
@@ -1601,6 +1601,122 @@ def fin():
...
+
+    # Try to add the same pwp again
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn]))
+    assert topo.logcap.contains("Password policy is already up to date")
+
+    # Update an existing subtree pwp
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn], pwdchange='on', passwordlockout='on'))
+    assert topo.logcap.contains("Password policy successfully updated")
```
droideck: Shouldn't that be `Successfully updated password policy`?

dirsrvtests/tests/suites/password/password_policy_test.py
```diff
@@ -1601,6 +1601,122 @@ def fin():
...
+    # Ensure OU exists
+    ous = OrganizationalUnits(topo.standalone, DEFAULT_SUFFIX)
+    ous.ensure_state(properties={'ou': 'people'})
+    people = ous.get("People")
+    assert people.exists()
+
+    # Add subtree pwp
+    topo.logcap.flush()
+    create_subtree_policy(inst, None, topo.logcap.log, _cli_args([people.dn]))
+    assert topo.logcap.contains("Successfully created new password policy")
```
droideck, line 1711: Also, when I run the tests on my VM, this assertion fails.

src/lib389/lib389/_mapped_object.py
```diff
@@ -1064,11 +1074,85 @@ def _create(self, rdn=None, properties=None, basedn=None, ensure=False):
...
+                    return self
+
+                if mods:
+                    _modify_ext_s(self._instance, self._dn, mods,
+                                serverctrls=self._server_controls,
+                                clientctrls=self._client_controls,
+                                escapehatch='i am sure')
+                    self._ensure_status = self.ENSURE_UPDATED
+                else:
+                    self._ensure_status = self.ENSURE_UNCHANGED
```
droideck: I'm not sure this line can ever happen as we've already checked `if matches:` above, hence mods is not empty.

src/lib389/lib389/_mapped_object.py
```diff
@@ -1064,11 +1074,85 @@ def _create(self, rdn=None, properties=None, basedn=None, ensure=False):
...
+                    else:
+                        values = v
+
+                    normalised = []
+                    for x in values:
+                        # Handle DN valued attributes
+                        if is_a_dn(x):
+                            x = ldap.dn.dn2str(ldap.dn.str2dn(x))
+                        normalised.append(ensure_str(x))
+                    current_attrs[key] = normalised
```
droideck: Duplicate code.
It probably can be extracted into a helper method like `_normalize_attrs(attrs_dict)`.

dirsrvtests/tests/suites/password/password_policy_test.py
```diff
@@ -1601,6 +1601,122 @@ def fin():
...
+    inst = topo.standalone
+
+    def _cli_args(dn, **extraargs):
+        args = FakeArgs()
+        args.DN = dn
+        args.suffix = False
+        args.json = False
+        args.verbose = False
+        args.pwdchange = None
+        args.passwordlockout = None
```
droideck: I think it's:


```suggestion
        args.pwdlockout = None
```

src/lib389/lib389/_mapped_object.py
```diff
@@ -121,6 +121,11 @@ class DSLdapObject(DSLogging, DSLint):
     :type dn: str
     """
 
+    # Outcome results for ensure_state operations
+    ENSURE_ADDED = "added"
+    ENSURE_UPDATED = "updated"
+    ENSURE_UNCHANGED= "unchanged"
```
droideck, line 145:
```suggestion
    ENSURE_UNCHANGED = "unchanged"
```

src/lib389/lib389/pwpolicy.py
```diff
@@ -150,12 +150,12 @@ def create_user_policy(self, dn, properties):
...
+        pwp_container = pwp_containers.ensure_state(properties={'cn': 'nsPwPolicyContainer'}, strict=True)
 
         # Create or update the policy entry
         properties['cn'] = 'cn=nsPwPolicyEntry_user,%s' % dn
         pwp_entries = PwPolicyEntries(self._instance, pwp_container.dn)
-        pwp_entry = pwp_entries.ensure_state(properties=properties)
+        pwp_entry = pwp_entries.ensure_state(properties=properties, strict=True)
         try:
             # Add policy to the entry
             user_entry.replace('pwdpolicysubentry', pwp_entry.dn)
```
droideck: This modify is unconditional. Minor thing, but it'll be nice not to run it if not needed.

dirsrvtests/tests/suites/password/password_policy_test.py
```diff
@@ -1601,6 +1601,122 @@ def fin():
...
+    assert topo.logcap.contains("Password policy successfully updated")
+
+    # Verify there are no duplicates listed
+    topo.logcap.flush()
+    list_policies(inst, None, topo.logcap.log, _cli_args(None))
+    _assert_no_duplicate_in_logs(topo.logcap)
+
+    # Cleanup
+    def fin():
+            del_local_policy(inst, None, topo.logcap.log, _cli_args([test_user.dn]))
```
droideck, line 1759: Double indentation:)

src/lib389/lib389/_mapped_object.py
```diff
@@ -1064,11 +1074,85 @@ def _create(self, rdn=None, properties=None, basedn=None, ensure=False):
...
+                self._ensure_status = self.ENSURE_UPDATED
+                return self
+            # Strict mode, compare and update only the differences
+            else:
+                # Get existing attributes
+                current_attrs_raw = self.get_all_attrs_utf8()
+                current_attrs = {}
+                for k, v in current_attrs_raw.items():
+                    key = k.lower()
+                    if not isinstance(v, list):
```
droideck: The check is redundant as get_all_attrs_utf8 returns lists (via ensure_list_str) and valid_props are lists (via ensure_list_bytes).

droideck (changes requested):
Overall, looks really great!

Besides a few comments below, could you please add another test to cover the CoS repair scenario? Something like:
  1. Creates a subtree policy 
  2. Manually deletes or corrupts a CoS entry (e.g., cos_template.delete())
  3. Calls create_subtree_policy again
  4. Verifies the CoS entry was recreated/repaired
  5. Verifies the status correctly reports the repair

dirsrvtests/tests/suites/password/password_policy_test.py
jchapma: hmm, after making the other changes I can run this test fine

src/lib389/lib389/_mapped_object.py
jchapma: Ah yes

src/lib389/lib389/_mapped_object.py
jchapma: Yep, its dead code

src/lib389/lib389/pwpolicy.py
jchapma: Good point

src/lib389/lib389/pwpolicy.py
jchapma: Would a "policy repaired" message be nicer :)

dirsrvtests/tests/suites/password/password_policy_test.py
jchapma: I was only running new test in isolation, when I run them all I get a failure too. A previous test didnt cleanup and when doing a subtree pwp add we got an ensure_statue outcome of updated instead of new.

src/lib389/lib389/pwpolicy.py
jchapma: > You only report one status. If a CoS template gets deleted or corrupted while the pwp_entry is intact. Running create_subtree_policy again:
> 
> * pwp_entry already exists, same attributes -> ENSURE_UNCHANGED
> * cos_template is missing -> gets created (ENSURE_ADDED)
> * cos_pointer_def has wrong cosTemplateDn -> gets fixed (ENSURE_UPDATED)
> * User sees: "Password policy is already up to date" -- misleading
> 
> But that's nice that it is repaired!

Fixed aggregate subtree pwp status 
 * pwp_entry already exists, same attributes -> ENSURE_UNCHANGED
 * cos_template is missing -> gets created (ENSURE_ADDED)
 * cos_pointer_def has wrong cosTemplateDn -> gets fixed (ENSURE_UPDATED)
 * User sees: "Password policy is already up to date" --> ENSURE_UPDATED

src/lib389/lib389/pwpolicy.py
```diff
@@ -157,8 +157,10 @@ def create_user_policy(self, dn, properties):
         pwp_entries = PwPolicyEntries(self._instance, pwp_container.dn)
         pwp_entry = pwp_entries.ensure_state(properties=properties, strict=True)
         try:
-            # Add policy to the entry
-            user_entry.replace('pwdpolicysubentry', pwp_entry.dn)
+            # Add policy to the entry if needed
+            user_pwp_dn = user_entry.get_attr_val_utf8('pwdpolicysubentry', [])
```
droideck: get_attr_val_utf8 signature is (self, key, use_json=False). So, I think, `[]` doesn't make sense here.

src/lib389/lib389/pwpolicy.py
```diff
@@ -157,8 +157,10 @@ def create_user_policy(self, dn, properties):
...
+            if user_pwp_dn != [pwp_entry.dn]:
```
droideck: This comparison always evaluates to True --- get_attr_val_utf8 returns either None or a str. Comparing a string/None to a list is always not-equal (in Python).

src/lib389/lib389/pwpolicy.py
droideck: Nice.
Additionally, we need the same for `create_user_policy`, I think.

jchapma: IMHO I dont think we need aggregate status for create_user_policy() as it only uses the policy entry, there is no CoS chain involved.

droideck: There's still the container (nsPwPolicyContainer) but I think that's fine.
The only issue I see -- if the container is brand new but the policy entry somehow already existed (unlikely edge case), the status would say "unchanged" which is still accurate from the user's perspective — the policy itself didn't change. So yeah, closing this comments as resolved. Thanks!

droideck (approved):
LGTM!

PR #7288, opened 2026-02-27
Issue 7281 - UI - Add encryption module management

mreynolds389:
Description:

Implement the Security page Encryption Modules tab with list/create/edit/enable-disable/delete workflows.

relates: https://github.com/389ds/389-ds-base/issues/7281

Assisted by: Cursor

## Summary by Sourcery

Add a new Security page tab for managing encryption modules with full CRUD and activation workflows backed by dsconf operations.

New Features:
- Introduce an Encryption Modules tab in the Security UI to list, create, edit, enable/disable, and delete encryption modules.
- Add a reusable encryption module modal and table component with validation and certificate selection that prevents reusing the same certificate across modules.

Enhancements:
- Extend the shared TypeaheadSelect component to support disabled options and additional option metadata used by the encryption module certificate selector.

Tests:
- Add a web UI test that verifies the Encryption Modules tab, its primary actions, and table headers are visible in the Security page.

### DIFF
--- a/dirsrvtests/tests/suites/webui/server/server_test.py
+++ b/dirsrvtests/tests/suites/webui/server/server_test.py
@@ -344,6 +344,39 @@ def test_security_log_availability(topology_st, page, browser_name):
     assert frame.get_by_text('Log Archive Exceeds (in MB)').is_visible()
 
 
+def test_security_encryption_modules_availability(topology_st, page, browser_name):
+    """ Test Encryption Modules tab visibility and controls
+
+    :id: b8a36f2e-381a-49ce-bf3a-4fe25a83f02a
+    :setup: Standalone instance
+    :steps:
+         1. Click on Security button on the side panel and check if Security Configuration tab is visible.
+         2. Click on Encryption Modules tab and check if Add Encryption Module button is visible.
+         3. Check if table headers Module Name and Server Certificate Name are visible.
+    :expectedresults:
+         1. Element is visible.
+         2. Element is visible.
+         3. Elements are visible.
+    """
+    setup_login(page)
+    time.sleep(1)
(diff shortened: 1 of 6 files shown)


jchapma (commented):
Works fine for me, some overall questions.

Regarding the "Extract File" components in the encryption module. On my fresh install I found that providing names for server cert and server key didnt result in the files being extracted to disk. Only later did I discover that extract-pemfiles needs to be enabled before extraction happens.  Is it worth mentioning this in the encryption module or should the user know this already.

When adding a new encryption module, should we check it exists before the server adds it ?

jchapma:
> Yeah we could add that check, although it would be rare because adding new security modules is not common and there would not be a lot of them. But I can add it since I think all that information is already available in the component.

Only if you think its worth it

jchapma (approved):
I played around with a bit and it works well for me.

I know no one would try to delete the primary cert profile, but when I try to delete the RSA profile I get a success message from the UI, but the profile still exists. When I try it on the CLI it fails with "cannot delete" message. It seems the dsconf exit code for a failed deletion is 0, so its treated as success. I created and  deleted several other profiles with no issue.

I know this is an extreme corner case so I will leave it up to you to decide what to do.

mreynolds389:
> I played around with a bit and it works well for me.
> 
> I know no one would try to delete the primary cert profile, but when I try to delete the RSA profile I get a success message from the UI, but the profile still exists. When I try it on the CLI it fails with "cannot delete" message. It seems the dsconf exit code for a failed deletion is 0, so its treated as success. I created and deleted several other profiles with no issue.
> 
> I know this is an extreme corner case so I will leave it up to you to decide what to do.

It is intentionally restricted and the UI should complain.  Should be simple to fix - I'll work on it...

PR #7290, opened 2026-02-28
Issue 7289 - RFE - Container First Run Improvements

### LINKED ISSUE #7289: dscontainer first-run improvements
To assist with container deployments, we should handle some cases better:

* When certificates are imported, we shouldn't generate self-signed certificates on first run
* If a DS_SUFFIX is provided, then we automatically create it on first run

Firstyear:
Bug Description: On first run of a container we always are creating the self-signed-cert even if it's not used. Additionally setting the DS_SUFFIX_NAME only adjusted dsrc, it didn't create the suffix

Fix Description:

* If a TLS certificate is being imported, don't generate self-signed certificates on first run.
* If a DS_SUFFIX_NAME is set, create the suffix on first run

fixes: https://github.com/389ds/389-ds-base/issues/7289

Review by: ???

## Summary by Sourcery

Adjust container first-run behavior to avoid unnecessary certificate generation and ensure directory suffix creation when configured.

Bug Fixes:
- Skip generating self-signed TLS certificates on first container run when an external certificate is being imported.
- Create the directory suffix on first container run when DS_SUFFIX_NAME is set instead of only updating configuration.

### DIFF
--- a/src/lib389/cli/dscontainer
+++ b/src/lib389/cli/dscontainer
@@ -108,6 +108,10 @@ def _begin_environment_config():
             ldbmconfig = LDBMConfig(inst)
             ldbmconfig.set("nsslapd-cache-autosize", str(autotune_pct))
 
+    # If the user imported TLS certs, ensure that the TLS listener is enabled.
+    if os.path.exists(CONTAINER_TLS_SERVER_KEY):
+        inst.config.set('nsslapd-security', 'on')
+
     inst.close()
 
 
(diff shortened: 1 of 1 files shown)


Firstyear:
I intend to backport this to 2.7 as well if that's okay :)

vashirov (approved):
Coincidentally, I've been working on something similar, but also supporting multiple backends, sample entries and ldif import as part of the initial run, as it's something our customers have been asking for. This change doesn't interfere with it, LGTM.

Firstyear:
Sure I'll do 2.8 as well.

PR #7294, opened 2026-03-02
Issue 7271 - implement a pre-close plugin function

mreynolds389:
Description:

replication protocol could benefit from being notified the shutdown process has started before calling the "close" plugin function. This would allow the replication protocol to wake up and adjust its active thread count to prevent a hang during shutdown.

relates: https://github.com/389ds/389-ds-base/issues/7271

## Summary by Sourcery

Introduce a pre-close lifecycle hook for plugins and use it to quiesce replication before shutdown.

New Features:
- Add a SLAPI pre-close plugin callback type and expose it via the plugin parameter block API.
- Invoke pre-close callbacks for all plugins during server shutdown before waiting for active threads to terminate.
- Register a pre-close handler in the multi-supplier replication plugin to shut down replication agreements ahead of final plugin close.

Enhancements:
- Extend the plugin descriptor structure to store a pre-close function pointer.

### DIFF
--- a/ldap/servers/plugins/replication/repl5_init.c
+++ b/ldap/servers/plugins/replication/repl5_init.c
@@ -1,6 +1,6 @@
 /** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
- * Copyright (C) 2005 Red Hat, Inc.
+ * Copyright (C) 2026 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -121,6 +121,8 @@ static int multisupplier_started_flag = 0;
 static PRUintn thread_private_agmtname; /* thread private index for logging*/
 static PRUintn thread_private_cache;
 static PRUintn thread_primary_csn;
+
+static int multisupplier_pre_stop(Slapi_PBlock *pb __attribute__((unused)));
 
 char *
 get_thread_private_agmtname()
@@ -836,10 +838,6 @@ multisupplier_stop(Slapi_PBlock *pb __attribute__((unused)))
     int rc = 0; /* OK */
 
     if (!multisupplier_stopped_flag) {
-        if (!is_ldif_dump) {
-            /* Shut down replication agreements */
-            agmtlist_shutdown();
-        }
         /* if we are cleaning a ruv, stop */
         stop_ruv_cleaning();
         /* unregister backend state change notification */
@@ -853,6 +851,16 @@ multisupplier_stop(Slapi_PBlock *pb __attribute__((unused)))
     return rc;
 }
 
+static int
+multisupplier_pre_stop(Slapi_PBlock *pb __attribute__((unused)))
+{
+    if (!multisupplier_stopped_flag) {
+        /* Shut down replication agreements which will stop all the
+         * replication protocol threads */
+        agmtlist_shutdown();
+    }
+    return 0;
+}
 
 PRBool
 multisupplier_started()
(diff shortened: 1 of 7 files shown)


progier389 (approved):
LGTM

ldap/servers/plugins/replication/repl5_init.c
```diff
@@ -853,6 +855,18 @@ multisupplier_stop(Slapi_PBlock *pb __attribute__((unused)))
...
 }
 
+static int
+multisupplier_pre_stop(Slapi_PBlock *pb __attribute__((unused)))
+{
+    if (!multisupplier_stopped_flag) {
+        if (!is_ldif_dump) {
+            /* Shut down replication agreements which will stop all the
+             * replication protocol threads */
+            agmtlist_shutdown();
```
droideck: If we added it here to `multisupplier_pre_stop` should we remove it from `multisupplier_stop`? Unless I miss something, there's no reason to have the second call.

ldap/servers/slapd/pblock.c
```diff
@@ -1,6 +1,6 @@
-/** BEGIN COPYRIGHT BLOCK
+ /** BEGIN COPYRIGHT BLOCK
```
droideck, line 1:
```suggestion
/** BEGIN COPYRIGHT BLOCK
```

ldap/servers/slapd/plugin.c
```diff
@@ -1847,6 +1847,25 @@ plugin_dependency_closeall(void)
     }
 }
 
+/* Call the pre close functions of all the plugins */
+void
+plugin_pre_closeall(void)
+{
+    for (size_t type = 0; type < PLUGIN_LIST_GLOBAL_MAX; type++) {
+        struct slapdplugin *plugin = global_plugin_list[type];
```
droideck: Just to future proof a bit, should we iterate through `global_plugin_shutdown_order[]` similarly to https://github.com/mreynolds389/389-ds-base/blob/main/ldap/servers/slapd/plugin.c#L1829 ?

mreynolds389: > Just to future proof a bit, should we iterate through `global_plugin_shutdown_order[]` similarly to https://github.com/mreynolds389/389-ds-base/blob/main/ldap/servers/slapd/plugin.c#L1829 ?

Doesn't hurt, why not.

ldap/servers/plugins/replication/repl5_init.c
mreynolds389: We don't call agmtlist_shutdown or the PRE_CLOSE function during a "db2ldif -r" so yes it can be removed

ldap/servers/plugins/replication/repl5_init.c
```diff
@@ -853,6 +851,16 @@ multisupplier_stop(Slapi_PBlock *pb __attribute__((unused)))
     return rc;
 }
 
+static int
+multisupplier_pre_stop(Slapi_PBlock *pb __attribute__((unused)))
+{
+    if (!multisupplier_stopped_flag) {
```
droideck, line 857: Previously we also had `if (!is_ldif_dump) {` guard.

Looks like it's a small safety guard which mirrors the init section:
https://github.com/389ds/389-ds-base/blob/main/ldap/servers/plugins/replication/repl5_init.c#L816-L821

Was it intentionally not included here?

droideck (approved):
Besides the small question, looks good!

vashirov (approved):
Test no longer hangs on my VM, LGTM

ldap/servers/plugins/replication/repl5_init.c
mreynolds389: > Previously we also had `if (!is_ldif_dump) {` guard.
> 
> Looks like it's a small safety guard which mirrors the init section: https://github.com/389ds/389-ds-base/blob/main/ldap/servers/plugins/replication/repl5_init.c#L816-L821
> 
> Was it intentionally not included here?

Yes it was removed intentionally because it's not needed now that the pre-close function exists.  I've tried all scenarios of import/export & offline/online repl inits.  It's correct.

PR #7301, opened 2026-03-04
Issue 7300 - Add OS-level thread names to all server threads

### LINKED ISSUE #7300: Add OS-level thread names to all server threads
**Description**
Right now, looking at pstack, top -H, or perf output from ns-slapd gives you a wall of anonymous threads. You end up cross-referencing thread IDs against internal structures just to figure out what each thread is doing. Not great for debugging.

We need to name our threads using `pthread_setname_np`.

droideck:
Description: Add slapi_set_thread_name() API wrapping pthread_setname_np and apply it to all server threads.
Workers appear as worker-0..N, the listener as "listener", replication threads as "repl-prot", "repl-inc-res", etc.

For functions that serve as both thread entry points and direct calls (db2ldif_skip_all, dbmdb_recno_cache_build), thin wrapper functions are used to keep thread naming separate from the core logic.

Fixes: https://github.com/389ds/389-ds-base/issues/7300

Reviewed by: ?

## Summary by Sourcery

Introduce a common API for setting OS-level thread names and apply consistent naming across core server, backend, and plugin threads for improved observability and diagnostics.

New Features:
- Add slapi_set_thread_name() API to expose OS-level thread naming to the server and plugins, and use it to assign descriptive names to key threads (listener, workers, replication, tasks, caches, and maintenance).

Enhancements:
- Refactor shared thread entry functions to use thin wrapper thread routines where needed to keep thread naming separate from core logic.
- Include config.h in select plugins to ensure proper feature detection for the new threading support.

Tests:
- Add an integration test that inspects /proc for the server process to verify that key threads (listener, worker-*, ct-list-*, event-q, housekeep) are correctly named at the OS level.

### DIFF
--- a/dirsrvtests/tests/suites/threads/thread_naming_test.py
+++ b/dirsrvtests/tests/suites/threads/thread_naming_test.py
@@ -43,12 +43,12 @@ def test_thread_names_present(topology_st):
     :setup: Standalone instance
     :steps:
         1. Get the server PID
-        2. Read thread names from /proc/<pid>/task/*/comm
+        2. Read thread names from /proc/{pid}/task/{tid}/comm
         3. Verify expected thread names are present
     :expectedresults:
         1. PID is obtained
         2. Thread names are readable
-        3. Core thread names (listener, worker-*, ct-list-*, event-q,
+        3. Core thread names (listener, worker, ct-list, event-q,
            housekeep) are found
     """
     inst = topology_st.standalone
--- a/dirsrvtests/tests/suites/threads/thread_naming_test.py
+++ b/dirsrvtests/tests/suites/threads/thread_naming_test.py
@@ -1,0 +1,81 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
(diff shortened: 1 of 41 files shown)


vashirov (approved):
LGTM

jchapma:
This is a great idea, test runs fine. Just a small observation, it seems the current thread names are functional, would we improve the visibility with a hierarchical naming scheme that groups related threads. It might be easier to troubleshoot if the threads could be grouped by subsystem when viewed by system tools.

ps aux | grep "srv-" shows all core server threads
ps aux | grep "repl-" shows all replication activity


Examples
```
listener - srv-listener
worker-N - srv-worker-N
ct-list-N - srv-conn-mgr-N
housekeep - srv-housekeep

repl-prot - repl-protocol
repl-inc-res - repl-recv-inc
repl-tot-res - repl-recv-tot
clean-ruv - repl-cleanruv
clean-ruv-ext - repl-cleanruv-ext
abort-ruv - repl-abortruv
cl-trim - repl-log-trim
cl5-test - repl-log-test

memberof-fix - plug-memberof-fix
memberof-def - plug-memberof-def
automem-abort - plug-automem-abort
automem-exprt - plug-automem-export
automem-map - plug-automem-map
```

mreynolds389 (approved):

Issue #7304, opened 2026-03-04
retro changelog should not cache dn


mreynolds389:
**Issue Description**

When adding an entry to the retro changelog we pass a flag to not cache the entry, but we still cache the DN which leads to some unexpected memory growth.

mreynolds389:
887994563..41564e997  389-ds-base-3.1 -> 389-ds-base-3.1
d9f85e5ee..e31095da1  389-ds-base-3.0 -> 389-ds-base-3.0
b1de2fe7b..dc439f3e9  389-ds-base-2.8 -> 389-ds-base-2.8
2c7cc3522..2d678ed96  389-ds-base-2.7 -> 389-ds-base-2.7
ce3d921d3..68859012b  389-ds-base-2.6 -> 389-ds-base-2.6
68dd1354c..ba678900c  389-ds-base-2.4 -> 389-ds-base-2.4
7c31db40e..bad79e694  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7306, opened 2026-03-04
Issue 7304 - retrocl should not cache DN

mreynolds389:
Description:

When adding a record to the retro changelog we pass in the flag SLAPI_OP_FLAG_NEVER_CACHE to prevent the entry from being cached, but we still update the DN cache leading to unexpected memory growth.

relates: https://github.com/389ds/389-ds-base/issues/7304

## Summary by Sourcery

Bug Fixes:
- Avoid updating or keeping DN cache entries when operations are marked to not use the cache, preventing memory leaks in the DN cache.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_add.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_add.c
@@ -1397,8 +1397,11 @@ common_return:
                 struct backdn *bdn = dncache_find_id(&inst->inst_dncache,
                                                      addingentry->ep_id);
                 if (bdn) { /* already in the dncache */
+                    if (is_remove_from_cache) {
+                        CACHE_REMOVE(&inst->inst_dncache, bdn);
+                    }
                     CACHE_RETURN(&inst->inst_dncache, &bdn);
-                } else { /* not in the dncache yet */
+                } else if (!is_remove_from_cache) { /* not in the dncache yet */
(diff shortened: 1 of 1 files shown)


tbordaz (commented):
Just a dummy question, how SLAPI_OP_FLAG_NEVER_CACHE is it used ? looking at ldbm_back_add it looks to me that the 'addingentry' is systematically added in entrycache.

mreynolds389:
> Just a dummy question, how SLAPI_OP_FLAG_NEVER_CACHE is it used ? looking at ldbm_back_add it looks to me that the 'addingentry' is systematically added in entrycache.

@tbordaz Well we only set that flag in retrocl plugin.  I'm not sure when this "feature" was added to retrocl (maybe it was always there).  Well whenever we touch an entry it's has to get added to the cache until we are done with the update.  In this code path it's already in the cache before we even enter this area of the code (common_return).   I'm just making sure it's removed from both caches as it doesn't make sense to only remove it from the entry cache and not the dn cache.  

Sorry I'm not sure if I'm answering your question completely, if not please let me know.

tbordaz (approved):
Actually SLAPI_OP_FLAG_NEVER_CACHE is used in plugins and translated into OP_FLAG_NEVER_CACHE in the core server. I missed that and the flag is_remove_from_cache stores it.
The code looks good to me.

PR #7308, opened 2026-03-05
Issue 7307 - Expose work queue and worker utilization metrics

### LINKED ISSUE #7307: Expose work queue and worker utilization metrics on cn=monitor
**Description**
The cn=monitor entry has connection counts and op counters, but nothing about the internal work queue or how busy worker threads are. This makes it hard to detect thread pool saturation with external monitoring tools (PCP, Grafana, scripts).

Add four new read-only attributes to cn=monitor:
  - currentworkqueue - operations currently waiting in the queue
  - maxworkqueue - high-water mark since startup
  - currentbusyworkers - workers currently processing an operation
  - maxbusyworkers - peak busy workers since startup

Saturation signal: currentbusyworkers == nsslapd-threadnumber && currentworkqueue > 0.

droideck:
Description: Add four new read-only attributes to cn=monitor: currentworkqueue, maxworkqueue, currentbusyworkers, and maxbusyworkers. These let administrators detect thread pool saturation using external monitoring tools (PCP, Grafana, custom scripts).
Worker counts use lock-free atomics with defensive clamping to guarantee non-negative values.
Add tests for basic metrics validation and concurrent busy worker tracking.
Add Cockpit UI fields to Monitor - Server Statistics.

Fixes: https://github.com/389ds/389-ds-base/issues/7307

Reviewed by: ?

## Summary by Sourcery

Expose work queue and worker utilization metrics through cn=monitor and surface them in both the Python monitoring API and Cockpit UI.

New Features:
- Add cn=monitor attributes for current work queue size, maximum work queue size, current busy workers, and maximum busy workers.
- Expose the new work queue and worker metrics via lib389 Monitor helpers and the Cockpit Server Monitor page.

Enhancements:
- Track current and peak busy worker counts and work queue depth using atomic counters for low‑overhead monitoring.

Tests:
- Add monitor tests validating the new work queue and busy worker metrics, including concurrent worker activity behavior.

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -1741,7 +1741,7 @@ connection_threadmain(void *arg)
     int doshutdown = 0;
     int maxthreads = 0;
     long bypasspollcnt = 0;
-    int is_busy = 0;
+    bool is_busy = false;
 
 #if defined(hpux)
     /* Arrange to ignore SIGPIPE signals. */
@@ -1769,7 +1769,7 @@ connection_threadmain(void *arg)
 
             /* Mark this worker as idle before blocking on the work queue */
             if (is_busy) {
-                is_busy = 0;
+                is_busy = false;
                 slapi_atomic_decr_32(&current_busy_workers, __ATOMIC_ACQ_REL);
             }
 
@@ -1870,14 +1870,14 @@ connection_threadmain(void *arg)
         }
         /* Once we're here we have a pb - mark this worker as busy */
         if (!is_busy) {
-            is_busy = 1;
+            is_busy = true;
             int32_t val = slapi_atomic_incr_32(&current_busy_workers, __ATOMIC_ACQ_REL);
             /*
              * Best-effort high-water mark: without a CAS primitive two threads
              * could race and briefly regress the value, but it self-corrects on
              * the next higher peak.  Acceptable for a monitoring-only metric.
              */
-            if (val > slapi_atomic_load_32(&max_busy_workers, __ATOMIC_RELAXED)) {
+            while (val > slapi_atomic_load_32(&max_busy_workers, __ATOMIC_RELAXED)) {
                 slapi_atomic_store_32(&max_busy_workers, val, __ATOMIC_RELAXED);
             }
         }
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/connection.c
```diff
@@ -1739,6 +1741,7 @@ connection_threadmain(void *arg)
     int doshutdown = 0;
     int maxthreads = 0;
     long bypasspollcnt = 0;
+    int is_busy = 0;
```
progier389: should better be:  
    bool is_busy = false;

ldap/servers/slapd/connection.c
```diff
@@ -78,6 +78,8 @@ static PRStack *work_q_stack;         /* stack of work_q structs so we don't hav
 static PRInt32 work_q_stack_size;     /* size of work_q_stack */
 static PRInt32 work_q_stack_size_max; /* max size of work_q_stack */
 static PRInt32 op_shutdown = 0;       /* if non-zero, server is shutting down */
+static int32_t current_busy_workers = 0;   /* workers currently processing ops */
```
progier389, line 81: I wonder if you should not better store all the counter in a struct
It will make things easier to put them in mmap later on

ldap/servers/slapd/connection.c
```diff
@@ -1850,11 +1868,26 @@ connection_threadmain(void *arg)
...
+        /* Once we're here we have a pb - mark this worker as busy */
+        if (!is_busy) {
+            is_busy = 1;
+            int32_t val = slapi_atomic_incr_32(&current_busy_workers, __ATOMIC_ACQ_REL);
+            /*
+             * Best-effort high-water mark: without a CAS primitive two threads
+             * could race and briefly regress the value, but it self-corrects on
+             * the next higher peak.  Acceptable for a monitoring-only metric.
+             */
+            if (val > slapi_atomic_load_32(&max_busy_workers, __ATOMIC_RELAXED)) {
```
progier389: I wonder if a while would not be better than an if here.
 IMHO it would slightly decrease the race condition window

progier389 (commented):
Looks mostly good.

Note that the tests seems to hangs.  Maybe you have not rebased and get latest fix from mark

mreynolds389 (approved):
Besides Pierres comments this LGTM.  I was concerned about the use of atomics as they are expensive, but I don't think these calls are in performance critical areas.

ldap/servers/slapd/connection.c
droideck: I think it'll stretch this PR too much...
The existing work_q_size/work_q_size_max are PRInt32 using NSPR atomics, while the new counters are int32_t with C11 builtins. Wrapping them in a struct now means either mixing the two styles or converting the NSPR ones, both of which are a stretch to this ticket. The getters already isolate all access, so when mmap PR lands we can introduce thread_pool_mmap_t with the full mmap layout (version field, per-worker slots, backtrace reservations) and migrate the counters into it in one shot -- the struct will reflect the actual mmap requirements rather than a guess.

progier389 (approved):
LGTM

PR #7311, opened 2026-03-05
Issue 7310 - On LMDB an export loops indefinitely if the suffix entry…

### LINKED ISSUE #7310: On LMDB an export loops indefinitely if the suffix entry has not the entryid=1
**Issue Description**
On LMDB there is a strong assumption that the suffix entry is the entry number 1.
If it is not then there are several possible symptoms

- high CPU during export
- abort of the consumer bulk_producer during an online import

The bug exists in all version but has no visible impact on BDB.

**Steps to Reproduce**
Will provide test script

**Expected results**
We should not see high CPU or abort on export or import

tbordaz:
… has not the entryid=1

Bug description:
	On LMDB there is a strong requirement that the suffix entry has the entry-id=1.
	In the case replication is enabled on a backend before the suffix entry exists
	it creates the RUV as entry-id=1 and so the suffix becomes entry-id=2.
	One of the symptom is an infinite loop when exporting the suffix
		#0  slapi_ch_array_add_ext
```
		#1  0x00007f24e56a36eb in slapi_rdn_add_rdn_to_all_rdns
		#2  0x00007f24e14ea99e in entryrdn_lookup_dn
		#3  0x00007f24e154b3a6 in dbmdb_db2ldif
		#4  0x00007f24e150294c in ldbm_back_ldbm2ldif
		#5  0x00005587335ff56f in slapd_exemode_db2ldif
		#6  main
```

Fix description:
	Hardening of _replica_configure_ruv that checks that the suffix
	entry exists before attempting to create the RUV.

fixes: https://github.com/389ds/389-ds-base/issues/7310

Reviewed by:

## Summary by Sourcery

Validate replica suffix entries before configuring RUV to prevent invalid LMDB layouts and export loops.

Bug Fixes:
- Prevent RUV configuration when the replica suffix entry is missing, returning an appropriate LDAP error instead of looping indefinitely during export.

Enhancements:
- Add an internal suffix-entry existence check used by replication configuration and RUV setup paths.
- Propagate detailed error text from RUV configuration failures when creating replicas from configuration entries.

### DIFF
--- a/dirsrvtests/tests/suites/backups/backup_test.py
+++ b/dirsrvtests/tests/suites/backups/backup_test.py
@@ -247,8 +247,10 @@ def test_replication(topo_m2):
         # Step 6: Start S1
         S1.start()
         # Step 7: Restore S1 from backup
+        # and restart the instance to take into account the restored backend
         rc = S1.tasks.bak2db(backup_dir=f'{backup_dir}', args={TASK_WAIT: True})
         assert rc == 0
+        S1.restart()
         # Step 8: Wait until replication is in sync
         # Must replicate first from S2 to S1 to resync S1
         repl.wait_for_replication(S2, S1)
--- a/ldap/servers/plugins/replication/repl5_replica.c
+++ b/ldap/servers/plugins/replication/repl5_replica.c
@@ -87,7 +87,7 @@ static int _replica_check_validity(const Replica *r);
 static int _replica_init_from_config(Replica *r, Slapi_Entry *e, char *errortext);
 static int _replica_update_entry(Replica *r, Slapi_Entry *e, char *errortext);
 static int _replica_config_changelog(Replica *r);
-static int _replica_configure_ruv(Replica *r, PRBool isLocked);
+static int _replica_configure_ruv(Replica *r, PRBool isLocked, char *errortext);
 static char *_replica_get_config_dn(const Slapi_DN *root);
 static char *_replica_type_as_string(const Replica *r);
 /* DBDB, I think this is probably bogus : */
@@ -222,9 +222,11 @@ replica_new_from_entry(Slapi_Entry *e, char *errortext, PRBool is_add_operation,
     }
 
     /* configure ruv */
-    rc = _replica_configure_ruv(r, PR_FALSE);
+    rc = _replica_configure_ruv(r, PR_FALSE, errortext);
     if (rc != 0) {
-        rc = LDAP_OTHER;
+        if (rc != LDAP_UNWILLING_TO_PERFORM) {
+                rc = LDAP_OTHER;
+        }
         goto done;
     } else {
         rc = LDAP_SUCCESS;
@@ -1624,7 +1626,7 @@ replica_reload_ruv(Replica *r)
 
     r->repl_ruv = NULL;
 
-    rc = _replica_configure_ruv(r, PR_TRUE);
+    rc = _replica_configure_ruv(r, PR_TRUE, NULL);
 
     replica_unlock(r->repl_lock);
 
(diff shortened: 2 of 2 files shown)


ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2577,6 +2579,53 @@ _replica_config_changelog(Replica *replica)
...
+		slapi_pblock_get(pb, SLAPI_PLUGIN_INTOP_SEARCH_ENTRIES, &entries);
+		if (NULL == entries || NULL == entries[0]) {
+			slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name,
+					      "_suffix_entry_check - suffix entry %s not retrieved\n",
+						  slapi_sdn_get_dn(r->repl_root));
+			goto done;
+		}
+		return_value = 0;
+	} else if (rc == LDAP_NO_SUCH_OBJECT) {
+		slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name,
```
progier389: I wonder if that one should not rather be a Warning rather than an error.

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2598,6 +2647,15 @@ _replica_configure_ruv(Replica *r, PRBool isLocked __attribute__((unused)))
     CSN *csn = NULL;
     ReplicaId rid = 0;
 
+    if (_suffix_entry_check(r)) {
+	slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name,
```
progier389: lots of frightening error messages at _suffix_entry_check then at _suffix_entry_check
maybe should not log the _suffix_entry_check message if entry is not found 
  (if success but no entry you can log a message as this case should not happen)
 and log a warning about being unable to write the ruv because suffix entry does not exists

ldap/servers/plugins/replication/repl5_replica.c
mreynolds389: Or maybe even REPL logging...

ldap/servers/plugins/replication/repl5_replica.c
mreynolds389: Need to indent the logging function

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2611,7 +2669,7 @@ _replica_configure_ruv(Replica *r, PRBool isLocked __attribute__((unused)))
     slapi_search_internal_set_pb(
         pb,
         slapi_sdn_get_dn(r->repl_root),
-        LDAP_SCOPE_BASE,
+		LDAP_SCOPE_ONELEVEL,
```
progier389: LDAP_SCOPE_BASE should be fine.  Better keep it.
      Since ns-uniqueid is provided  it is used rather the dn and we are returning here a single entry

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2577,6 +2579,50 @@ _replica_config_changelog(Replica *replica)
...
+{
+    Slapi_PBlock *pb = NULL;
+    Slapi_Entry **entries = NULL;
+    int return_value = -1;
+    int rc;
+
+    pb = slapi_pblock_new();
+    if (!pb) {
+         slapi_log_err(SLAPI_LOG_ERR, repl_plugin_name,
+                       "_suffix_entry_found - Out of memory\n");
```
droideck, line 2594:
```suggestion
                       "_suffix_entry_check - Out of memory\n");
```

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -222,9 +222,11 @@ replica_new_from_entry(Slapi_Entry *e, char *errortext, PRBool is_add_operation,
     }
 
     /* configure ruv */
-    rc = _replica_configure_ruv(r, PR_FALSE);
+    rc = _replica_configure_ruv(r, PR_FALSE, errortext);
     if (rc != 0) {
-        rc = LDAP_OTHER;
+        if (rc != LDAP_UNWILLING_TO_PERFORM) {
+                rc = LDAP_OTHER;
```
droideck, line 228: Indentation

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2586,7 +2632,7 @@ _replica_config_changelog(Replica *replica)
  * Returns 0 on success, -1 on failure. If 0 is returned, the RUV is present in the replica.
```
droideck, line 2632: `Returns 0 on success, -1 on failure` -- probably, it should be extended as we also return LDAP_UNWILLING_TO_PERFORM now

ldap/servers/plugins/replication/repl5_replica.c
```diff
@@ -2577,6 +2579,50 @@ _replica_config_changelog(Replica *replica)
...
+                          slapi_sdn_get_dn(r->repl_root));
+            goto done;
+        }
+        return_value = 0;
+    } else if (rc == LDAP_NO_SUCH_OBJECT) {
+        goto done;
+    }
+
+
+done:
```
droideck, line 2618: Double blank line before `done:` :)

droideck (approved):
Very minor issues. Otherwise, LGTM!

dirsrvtests/tests/suites/backups/backup_test.py
```diff
@@ -247,8 +247,10 @@ def test_replication(topo_m2):
         # Step 6: Start S1
         S1.start()
         # Step 7: Restore S1 from backup
+        # and restart the instance to take into account the restored backend
         rc = S1.tasks.bak2db(backup_dir=f'{backup_dir}', args={TASK_WAIT: True})
         assert rc == 0
+        S1.restart()
```
progier389, line 253: That is not normal that you have to restart the server .
 as bak2db already restart the backend. 
IMHO if you really need to restart that means that something is broken somewhere

tbordaz: Ok. I will investigate deeper

tbordaz: It is an online restore. During startup, it fails to retrieve the suffix and check the RUV because data.mdb was remove. So the backend is offline/broken before the restore that should contribute to the backend not restarted after restore. continuing

tbordaz: Indeed, the in-memory replica is built when parsing the dse.ldif config (replica_config_add) that is done at startup or when the replica is created. This step failed because data.mb was removed and the replica did not exist. When the backend is restored, it is started but not the replica. This because replica_config_add is not called again.
An option would be to detect that the suffix is missing, log an error, skip the creation of the RUV and return silently a success. Worth a try.

progier389: I wonder if we have not the same issue after online import
Maybe we should add a new step when bringing up backend on line 
 to configure the ruv if it is not done and if the suffix entry exists

tbordaz: Yes there is the same issue with online import. May be we could use framework plg_mmr_betxn_preop

progier389: About plg_mmr_betxn_preop:  the preop is not called in case of restore/import 
  and I am not sure we want to do the test for every operation.  (But we may want to do it when adding the suffix entry (in bepostop?). ( BTW how replication is bootstrapped  when creating empty backend then adding suffix entry ?)
IMHO, for import/restore case, you should add a new pre_enable_be function (a bit like what Mark has done with https://github.com/389ds/389-ds-base/pull/7294 for the pre_close function)

progier389: Well in fact the callback already exists:
 multisupplier_be_state_change

Issue #7312, opened 2026-03-05
UI - Database Maximum Size cannot be easily set by typing


jchapma:
**Issue Description**
Setting Database Maximum size is not easy to set by typing. The field enforces the minimum value to be written in any time a user tries to overwrite the value.

**Package Version and Platform:**
 - Package and version: main branch

**Steps to Reproduce**
1. Create an instance
2. WebUI -> Database -> Global Database Configuration -> Database Size
3. Select the value in the Database Maximum Size field
4. Try to type new value

**Actual results**
Minimum value of 100 is enforced and no lower number is allowed to be entered, i.e. when trying to set 12345, user selects original value, starts typing the new value, but upon hitting the first digit, which is of course less than 100, the value 100 appears in the field and has to be overwritten in a way that does not make the value in the field >100, as that also triggers the minimal value enforcement.

**Expected results**
User should be able to comfortably type new value without having minimal value enforced. Minimal value can be checked on the go and a proper error message shown under the field (similar to changing ports in server menu)

**Additional context**
Reported by @mirielka

PR #7313, opened 2026-03-05
Issue 7312 - UI - Database Maximum Size cannot be easily set by typing

jchapma:
Description:
The mdb max size field was clamping user input while typing. With a min value of 100, when users tried to type values, the first digit would clamp to 100, making it impossible to enter valid values.

Fix:
Used an existing config change function that allows free typing, validation is now done in a blur function.

Tidied up the conversion from bytes to MB, converting only at the dsconf boundaries.

Fixes:
https://github.com/389ds/389-ds-base/issues/7312

Reviewed by:

## Summary by Sourcery

Adjust MDB max size configuration UI to allow free typing in megabytes with proper validation and byte conversion at the configuration boundaries.

Bug Fixes:
- Fix MDB max size input clamping that prevented users from typing valid values below the minimum while editing.
- Ensure special-value handling in range blur validation only applies when a special value is defined.

Enhancements:
- Change MDB max size increment/decrement controls to operate in megabytes within the available database size range.
- Convert MDB max size between bytes and megabytes only when reading from and writing to the underlying configuration attributes.

### DIFF
--- a/src/cockpit/389-console/src/database.jsx
+++ b/src/cockpit/389-console/src/database.jsx
@@ -427,7 +427,7 @@ export class Database extends React.Component {
                                         pagelooklimit: attrs['nsslapd-pagedlookthroughlimit'][0],
                                         pagescanlimit: attrs['nsslapd-pagedidlistscanlimit'][0],
                                         rangelooklimit: attrs['nsslapd-rangelookthroughlimit'][0],
-                                        mdbmaxsize: attrs['nsslapd-mdb-max-size'][0],
+                                        mdbmaxsize: Math.floor(attrs['nsslapd-mdb-max-size'][0] / (1024 * 1024)),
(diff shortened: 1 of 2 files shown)


mreynolds389 (approved):
Nice, Ack!

droideck (commented):
If I set -1 to `Page Search Look Through Limit`, for example. And then I reload the page -- I am able to press -2 and set `Page Search Look Through Limit` to that value (if I change any other field).

Also, no error is displayed. I think `state.error` is never written to.

droideck (commented):
When I try to change non-numeric fields like `Dynamic List Objectclass`, it doesn't change Save Config button state (so the fields can't be easily edited)

droideck (approved):
LGTM!

PR #7318, opened 2026-03-09
Issue 6942 - Crash in `slapi_sdn_get_ndn()`

### LINKED ISSUE #6942: Crash in `slapi_sdn_get_ndn`
**Issue Description**
I noticed a crash during `dirsrvtests/tests/suites/replication/conflict_resolve_test.py::TestTwoSuppliers::test_complex_add_modify_modrdn_delete` on LMDB:
```
Thread 1 (Thread 0x7efb6cfff6c0 (LWP 722)):
#0  slapi_sdn_get_ndn (sdn=0x7365743d6e632c30) at ldap/servers/slapd/dn.c:2365
#1  0x00007eff5347bbed in entry_same_dn (e=<optimized out>, k=0x7efb69213460) at ldap/servers/slapd/back-ldbm/cache.c:149
#2  0x00007eff5347bde5 in add_hash (ht=0x7eff538e2000, key=key@entry=0x7efb69213460, keylen=<optimized out>, entry=entry@entry=0x7efb692134d0, alt=alt@entry=0x7efb6cff8788) at ldap/servers/slapd/back-ldbm/cache.c:201
#3  0x00007eff5347d72b in entrycache_add_int (cache=0x7eff56300b10, e=0x7efb692134d0, state=state@entry=0, alt=alt@entry=0x7efb6cff8840) at ldap/servers/slapd/back-ldbm/cache.c:1502
#4  0x00007eff5347da3a in cache_add (cache=cache@entry=0x7eff56300b10, ptr=<optimized out>, alt=alt@entry=0x7efb6cff8840) at ldap/servers/slapd/back-ldbm/cache.c:1675
#5  0x00007eff53488342 in id2entry (be=0x7eff53862dc0, id=48, txn=<optimized out>, err=0x7efb6cff8924) at ldap/servers/slapd/back-ldbm/id2entry.c:441
#6  0x00007eff534c157d in ldbm_back_next_search_entry (pb=0x7efb69200000) at ldap/servers/slapd/back-ldbm/ldbm_search.c:1745
#7  0x00007eff580025e2 in iterate (send_result=1, pb=0x7efb69200000, be=0x7eff53862dc0, pnentries=0x7efb6cff8bf8, pagesize=-1, pr_statp=0x7efb6cff8b60) at ldap/servers/slapd/opshared.c:1254
#8  send_results_ext (pb=pb@entry=0x7efb69200000, nentries=nentries@entry=0x7efb6cff8bf8, pagesize=-1, pr_stat=pr_stat@entry=0x7efb6cff8b60, send_result=<optimized out>) at ldap/servers/slapd/opshared.c:1607
#9  0x00007eff5800484c in op_shared_search (pb=pb@entry=0x7efb69200000, send_result=send_result@entry=1) at ldap/servers/slapd/opshared.c:943
#10 0x00005620025598f4 in do_search (pb=0x7efb69200000) at ldap/servers/slapd/search.c:411
#11 connection_dispatch_operation (conn=<optimized out>, op=<optimized out>, pb=pb@entry=0x7efb69200000) at ldap/servers/slapd/connection.c:737
#12 0x000056200255c1e2 in connection_threadmain (arg=<optimized out>) at ldap/servers/slapd/connection.c:1995
#13 0x00007eff575ad843 in _pt_root () at /lib64/libnspr4.so
#14 0x00007eff57dec154 in start_thread () at /lib64/libc.so.6
#15 0x00007eff57e6ecac in __clone3 () at /lib64/libc.so.6
(gdb) p sdn
$3 = (const Slapi_DN *) 0x7365743d6e632c30
(gdb) p *sdn
Cannot access memory at address 0x7365743d6e632c30
```

`0x7365743d6e632c30` is `0,cn=tes`, so I suspect a memory corruption.
I rerun it 10 times with ASAN and without, didn't get the crash.

vashirov:
Bug Description:
In `id2entry_add_ext()` we have a workaround that normalizes entry DN against the parent's cached DN to fix case difference (bz628300). But `cache_add_tentative()` has already inserted the entry into the `c_dntable` hash table under the old normalized DN (slot A). When the workaround for bz628300 changes the DN in-place, the subsequent `CACHE_ADD()` inserts the entry into `c_dntable` under the new normalized DN (slot B). The entry now exists in 2 hash slots at the same time. And when the entry is later freed or evicted, only slot B is cleaned up. Slot A still contains a dangling pointer to the freed memory. Any further operation that walks the slot A hash chain (search via `id2entry` or delete via `cache_find_dn`), calls `entry_same_dn()`, which dereferences the dangling pointer, leading to SIGSEGV.

Fix Description:

1. In `id2entry_add_ext()` call `cache_remove_dn_hash()` under CACHE_LOCK immediately before the bz628300 DN change to remove the entry from its current hash slot.

2. In `entrycache_add_int()` when confirming a tentative add, call `remove_hash_entry()` to find and remove the entry by pointer. This is needed as a safety net because a betxn pre-add plugin may have already changed the DN before the bz628300 code runs.

Fixes: https://github.com/389ds/389-ds-base/issues/6942

## Summary by Sourcery

Prevent DN hash table corruption and crashes when an entry's DN changes after tentative cache insertion.

Bug Fixes:
- Ensure entries are removed from the DN cache under their old normalized DN before in-place DN changes in id2entry_add_ext to avoid duplicate hash slots and dangling pointers.
- Add a safety-net removal of cache entries by pointer when confirming tentative adds so DN changes by betxn pre-add plugins cannot leave stale hash entries.

Enhancements:
- Introduce helper functions to remove cache entries from DN and generic hash tables to keep cache state consistent.

Tests:
- Add a stress test with a custom betxn pre-add plugin that mutates DNs under load to reproduce the DN hash corruption scenario and verify servers do not crash.

### DIFF
--- a/dirsrvtests/tests/suites/betxns/entrycache_dn_hash_corruption_test.py
+++ b/dirsrvtests/tests/suites/betxns/entrycache_dn_hash_corruption_test.py
@@ -1,0 +1,443 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import os
+import time
+import shutil
+import logging
+import subprocess
+import threading
+import pytest
+import ldap
+
+from lib389.topologies import topology_m2
+from lib389._mapped_object import DSLdapObject
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.idm.organizationalunit import OrganizationalUnits
+from lib389.idm.user import UserAccounts
+from lib389.replica import ReplicationManager
+from lib389.utils import get_default_db_lib
+from lib389.config import BDB_LDBMConfig, LMDB_LDBMConfig
+
+DEBUGGING = os.getenv('DEBUGGING', default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+PLUGIN_SOURCE = r"""
+/*
+ * dn-hash-corruption plugin
+ *
+ *   For every ADD under the configured container (default ou=test),
+ *   prepends 'x' to the uid RDN value:
+ *   uid=w1u42,ou=test,dc=example,dc=com -> uid=xw1u42,ou=test,dc=example,dc=com
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+#include <stdlib.h>
+#include "slapi-plugin.h"
+
+#define PLUGIN_NAME "dn-hash-corruption"
+
+/* TRIGGER_PARENT_RDN is the parent RDN prefix that selects which
+ * entries the plugin modifies.
+ */
+#ifndef TRIGGER_PARENT_RDN
+#define TRIGGER_PARENT_RDN "ou=test"
+#endif
+#define TRIGGER_PARENT_RDN_LEN (sizeof(TRIGGER_PARENT_RDN) - 1)
+
+static Slapi_PluginDesc plugin_desc = {
+    PLUGIN_NAME, "reproducer", "1.0",
+    "Triggers entry cache DN hash corruption"
+};
+static Slapi_ComponentId *plugin_id = NULL;
+
+/*
+ * betxn pre-add callback.
+ * Runs after `cache_add_tentative()` inserted the entry into `c_dntable`
+ * under the original DN, but before `CACHE_ADD()` re-inserts it.
+ * Changing the DN here puts the entry in two different hash slots.
+ */
+static int dnhashcorr_betxn_pre_add(Slapi_PBlock *pb)
+{
+    Slapi_Entry *entry = NULL;
+    const char *dn = NULL;
+    char *parent_dn = NULL;
+    char *old_uid = NULL;
+    char *new_uid = NULL;
+    char *new_dn = NULL;
+    Slapi_DN *sdn = NULL;
+
+    slapi_pblock_get(pb, SLAPI_ADD_ENTRY, &entry);
+    if (!entry) return 0;
+
+    dn = slapi_entry_get_dn_const(entry);
+    if (!dn) return 0;
+
+    /* Only target entries under the configured container */
+    parent_dn = slapi_dn_parent(dn);
+    if (!parent_dn) return 0;
+    if (strncasecmp(parent_dn, TRIGGER_PARENT_RDN, TRIGGER_PARENT_RDN_LEN) != 0) {
+        slapi_ch_free_string(&parent_dn);
+        return 0;
+    }
+
+    /* Get the current uid attribute value */
+    old_uid = slapi_entry_attr_get_charptr(entry, "uid");
+    if (!old_uid) {
+        slapi_ch_free_string(&parent_dn);
+        return 0;
+    }
+
+    /*
+     * Build new DN by prepending 'x' to the uid value.
+     * This changes the NDN hash -> entry ends up in two hash slots.
+     */
+    new_dn = slapi_ch_smprintf("uid=x%s,%s", old_uid, parent_dn);
+    slapi_ch_free_string(&parent_dn);
+    if (!new_dn) {
+        slapi_ch_free_string(&old_uid);
+        return 0;
+    }
+
+    /*
+     * Replace the entry's SDN in-place.
+     * The entry is still linked in c_dntable under the old NDN hash (slot A).
+     * After this CACHE_ADD will insert it under the new NDN hash (slot B).
+     */
+    sdn = slapi_entry_get_sdn(entry);
+    slapi_sdn_set_dn_passin(sdn, new_dn);
+    slapi_sdn_get_ndn(sdn);
+
+    /* Update uid attribute to match new RDN to avoid naming violation */
+    new_uid = slapi_ch_smprintf("x%s", old_uid);
+    slapi_entry_attr_set_charptr(entry, "uid", new_uid);
+    slapi_ch_free_string(&new_uid);
+    slapi_ch_free_string(&old_uid);
+
+    return 0;
+}
+
+static int dnhashcorr_start(Slapi_PBlock *pb)
+{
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -278,6 +278,55 @@ remove_hash(Hashtable *ht, const void *key, uint32_t keylen)
...
+ *
+ * Returns 1 if the entry was found and removed, 0 otherwise.
+ */
+static int
+remove_hash_entry(Hashtable *ht, void *entry)
+{
+    u_long i;
+    void *e, *laste;
+
+    for (i = 0; i < ht->size; i++) {
```
tbordaz: What about storing in the entry the hash slot where the entry is stored. It would prevent to go through all the slots. As there are several hash tables, we could use a callback (e.g. ((*ht->get_slot_fn)) (e)) specific to each table to retrieve the hash slot in the entry structure.

This can be done with another ticket.

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1762,6 +1811,23 @@ entrycache_add_int(struct cache *cache, struct backentry *e, int state, struct b
...
+     * ends up in two hash slots and the old slot becomes a dangling pointer
+     * after the entry is freed -> use-after-free crash in entry_same_dn().
+     */
+    if ((e->ep_state & ENTRY_STATE_CREATING) && (state == 0)) {
+        if (remove_hash_entry(cache->c_dntable, (void *)e)) {
+            already_in = 1;
+        }
+    }
+
     if (!add_hash(cache->c_dntable, (void *)ndn, strlen(ndn), e,
```
tbordaz, line 1842: Could we use 'already_in' in the test ?
`if (already_in || !add_hash(cache->c_dntable, ...)) {`

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1762,6 +1811,23 @@ entrycache_add_int(struct cache *cache, struct backentry *e, int state, struct b
...
+     * was changed between cache_add_tentative() and this CACHE_ADD() call
+     * (e.g. by the bz628300 parent-DN adjustment in id2entry_add_ext, or
+     * by a betxn pre-add plugin), we must remove the stale old-DN hash
+     * reference BEFORE inserting under the new DN. Otherwise the entry
+     * ends up in two hash slots and the old slot becomes a dangling pointer
+     * after the entry is freed -> use-after-free crash in entry_same_dn().
+     */
+    if ((e->ep_state & ENTRY_STATE_CREATING) && (state == 0)) {
+        if (remove_hash_entry(cache->c_dntable, (void *)e)) {
+            already_in = 1;
```
tbordaz: Should not it be the opposite already_in=0 ?

ldap/servers/slapd/back-ldbm/cache.c
vashirov: Yes, I'm testing a similar approach right now (store NDN instead of index). I'd like to do it as part of this PR to avoid unnecessary performance penalty.

ldap/servers/slapd/back-ldbm/cache.c
vashirov: I'll add a comment to clarify, but basically setting this flag to 1 here means we found the entry in the hash table under the old DN, removed it, and now we're going to re-add it under the new DN. If we set the flag to 0, then the confirmation later will break when [`add_hash(cache->c_idtable, ...)`](https://github.com/389ds/389-ds-base/blob/7d5de9965e6945e995ce5d9c06c7913b5b7e4154/ldap/servers/slapd/back-ldbm/cache.c#L1837-L1858) fails (the ID is already in `c_idtable` from the tentative add).

ldap/servers/slapd/back-ldbm/cache.c
vashirov: If we use it here, then we would skip the entire `add_hash` block, `my_alt` would be uninitialized. Even if we initialize `my_alt = e`, the entry wouldn't be re-inserted into `c_dntable` under the new DN.

ldap/servers/slapd/back-ldbm/cache.c
progier389, line 1982: I wonder if c_stats.size should not take in account e->ep_dn_hash_ndn
to avoid that the dn cache memory footprint increase because of the change

Anyway whatever will do this change will have an impact  (Either on memory footprint or on eviction rate)

dirsrvtests/tests/suites/betxns/entrycache_dn_hash_corruption_test.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+def test_entrycache_dn_hash_corruption(setup_plugin):
+    """Test that entry cache DN hash corruption via in-place DN change
+    does not crash the server.
+
+    :id: 7d1c0af5-1f0f-454a-8114-ea419a93a410
+    :setup: Two supplier replication topology with reproducer plugin
+    :steps:
+        1. Run concurrent add/search/delete workload for 120 seconds
```
droideck: Looks like it's 300 by default... As per `WORKLOAD_DURATION`.
Maybe it's worth not to mention the exact seconds at all.

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1762,6 +1809,36 @@ entrycache_add_int(struct cache *cache, struct backentry *e, int state, struct b
...
+     * after the entry is freed -> use-after-free crash in entry_same_dn().
+     */
+    if ((e->ep_state & ENTRY_STATE_CREATING) && (state == 0)) {
+        if (e->ep_dn_hash_ndn) {
+            if (remove_hash_entry(cache->c_dntable,
+                    e->ep_dn_hash_ndn,
+                    strlen(e->ep_dn_hash_ndn), e)) {
+                /*
+                 * Set `already_in = 1` because the entry *is* already in the
+                 * cache (it was tentatively added). This flag is checked
```
droideck: The comment is a bit confusing, IMO. 

`"because the entry *is* already in the cache (it was tentatively added)"`: 
but `cache_add_tentative()` only inserts into c_dntable -- the `add_hash(c_idtable, ...)` call is gated on `if (state == 0)`  which is not true for `cache_add_tentative ` which has the explicit `entrycache_add_int(cache, e, ENTRY_STATE_CREATING, alt);` call (ENTRY_STATE_CREATING != 0)

Maybe it can be reworded to something like:
```
The tentative add already counted this entry in c_stats.
Set already_in so the stats block below does not double-count
when we re-insert under the new DN.
```

droideck (approved):
LGTM!
Besides the couple of very minor nits.

PR #7321, opened 2026-03-09
Issue 7265 - CI - fix retro changelog maxage validation test

mreynolds389:
Description:

Previously the retro changelog allowed invalid values even though they are ignored, but not they are rejected and the CI test needs to be updated

relates: https://github.com/389ds/389-ds-base/issues/7265

## Summary by Sourcery

Tests:
- Adjust changelog_test to consistently treat the presence of certain attribute values as unexpected, removing the special-case handling for the CHANGELOG plugin.


droideck (approved):
LGTM

PR #7324, opened 2026-03-11
Issue 6753 - Port ticket 47976 test

mirielka:
Description:
Port ticket 47976 test into dirsrvtests/tests/suites/plugins/managed_entry_test.py using lib389 API.

Relates: #6753
Author: Lenka Doudova
Assisted by: Cursor
Reviewer: ???

## Summary by Sourcery

Port the ticket 47976 regression scenario into the managed entry plugin test suite using the lib389 API and retire the old standalone ticket test.

Tests:
- Add a managed entry regression test that verifies user deletions do not hang after DB reimport with large BDB page size.
- Remove the legacy ticket47976 standalone test file now that the scenario is covered in the managed_entry_test suite.


aadhikar (approved):
Nice port, one question:

You're `using _set_dbsizes(inst, 64 * 1024, 80960)` but the original test used `128 * 1024` for page size. Was this change intentional?

Otherwise LGTM!

Issue #7327, opened 2026-03-12
dsctl healthcheck DSMOLE0001 inaccurate recommendations with multiple LDAP backends


jchapma:
**Issue Description**
The dsctl healthcheck tool is generating inaccurate recommendations for the MemberOf plugin when there are more than one LDAP backend, particularly in IPA context with the ipaca backend. 

**Steps to Reproduce**
1. Setup DS with multiple backends
```
cn=changelog (changelog)
dc=idm,dc=example,dc=test (userroot)
o=ipaca (ipaca)
```
2. Verify the MO plugin has useroot backend n scope
```
dn: cn=MemberOf Plugin,cn=plugins,cn=config
cn: MemberOf Plugin
memberofattr: memberOf
memberofentryscope: dc=idm,dc=example,dc=test
```
3. Run healthcheck on your instance
4. See HC error:
```
[1] DS Lint Error: DSMOLE0001
--------------------------------------------------------------------------------
Severity: HIGH
Check: memberof:attr_indexes
Affects:
– cn=memberof plugin,cn=plugins,cn=config
– o=ipaca
– memberuser
```

**Actual result**
dsctl healthcheck recommends adding indexes for attributes such as memberUser to the ipaca backend, even though the MO plugin configuration applies only to the useroot backend

**Expected results**
Evaluate MO plugin configuration in relation to the backend defined by memberofentryscope

**Additional context**
Reported by @Marc Sauton

vashirov:
b512f6a57..084cab379  389-ds-base-2.8 -> 389-ds-base-2.8
f34eff848..af0216be8  389-ds-base-2.9 -> 389-ds-base-2.9
bcfcf7964..09208f27c  389-ds-base-3.0 -> 389-ds-base-3.0
97e5e5ee8..f7f78983f  389-ds-base-3.1 -> 389-ds-base-3.1

PR #7328, opened 2026-03-12
Issue 7327 - dsctl healthcheck DSMOLE0001 inaccurate recommendations with multiple backends

jchapma:
Description:
The dsctl healthcheck tool generates incorrect recommendations for the MO plugin when multiple backends are present. This commonly occurs in IPA environments where both userroot and ipaca backends exist. Healthcheck incorrectly suggests indexing attributes for backends that are not within the MemberOf plugin scope.

Fix:
Determine the MO plugin scope while iterating over backends and only generate recommendations for backends that fall within that scope.

Removed references to nsslapd-plugincontainerscope as we dont use it.

Fixes:
https://github.com/389ds/389-ds-base/issues/7327

Reviewed by:

## Summary by Sourcery

Update MemberOf plugin healthcheck logic to respect the plugin’s configured entry scopes when checking backend indexes.

Bug Fixes:
- Ensure dsctl healthcheck only recommends MemberOf-related indexes for backends that fall within the plugin’s configured entry scopes instead of all backends.
- Remove use of the deprecated/unused nsslapd-plugincontainerscope attribute in MemberOf healthcheck validations.

### DIFF
--- a/src/lib389/lib389/plugins.py
+++ b/src/lib389/lib389/plugins.py
@@ -794,16 +794,21 @@ class MemberOfPlugin(Plugin):
             from lib389.backend import Backends
             backends = Backends(self._instance).list()
             attrs = self.get_attr_vals_utf8_l("memberofgroupattr")
-            container = self.get_attr_val_utf8_l("nsslapd-plugincontainerscope")
+            scopes = self.get_attr_vals_utf8_l("memberofentryscope")
             for backend in backends:
                 suffix = backend.get_attr_val_utf8_l('nsslapd-suffix')
                 if suffix == "cn=changelog":
                     # Always skip retro changelog
                     continue
-                if container is not None:
-                    # Check if this backend is in the scope
-                    if not container.endswith(suffix):
-                        # skip this backend that is not in the scope
+                if scopes:
+                    # Is this backend suffix in scope
+                    in_scope = False
+                    for scope in scopes:
+                        if scope.endswith(suffix):
+                            in_scope = True
+                            break
+
+                    if not in_scope:
                         continue
                 indexes = backend.get_indexes()
                 for attr in attrs:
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/plugins.py
```diff
@@ -794,16 +794,21 @@ def _lint_member_attr_indexes(self):
...
                     continue
-                if container is not None:
-                    # Check if this backend is in the scope
-                    if not container.endswith(suffix):
-                        # skip this backend that is not in the scope
+                if scopes:
+                    # Is this backend suffix in scope
+                    in_scope = False
+                    for scope in scopes:
+                        if scope.endswith(suffix):
```
droideck: It's preexistent but as you rewrite, it might worth to fix it.
I don't think something like happens often but we might have sub suffix backends (ou=people,dc=example,dc=com) and if memberOfEntryScope is dc=example,dc=com, that will incorrectly skip the backend.

droideck (commented):
I think we also need a test with two backends where one is out of scope.

src/lib389/lib389/plugins.py
jchapma: Yes, you are correct, the check looks backward to me. Nice catch

jchapma: > Yes, you are correct, the check looks backward to me. Nice catch

I added a test for sub suffix on multiple backends, it triggered a bug in the plugin healthcheck code where we do the same thing,  check the scope is in a suffix rather than the other way around. I will include the plugin change in this PR, its a one liner. Thanks

src/lib389/lib389/plugins.py
```diff
@@ -794,16 +794,21 @@ def _lint_member_attr_indexes(self):
...
                     continue
-                if container is not None:
-                    # Check if this backend is in the scope
-                    if not container.endswith(suffix):
-                        # skip this backend that is not in the scope
+                if scopes:
+                    # Is this backend suffix in scope
+                    in_scope = False
+                    for scope in scopes:
+                        if suffix.endswith(scope):
```
droideck: I think previously it was partially correct though...
The purpose is to determine whether a backend is relevant for a given scope. A backend is relevant when the scope lives within that backend's tree -- i.e., the scope DN ends with the backend suffix.

So it, IMO, should be `if scope.endswith(suffix) or suffix.endswith(scope):`. Or do I miss something?

droideck (approved):
Besides the small issue, LGTM!

src/lib389/lib389/plugins.py
jchapma: You are correct, we only  consider one direction of the hierarchy, missing the case where the scope is more specific than the suffix.

Issue #7331, opened 2026-03-12
UI - improve LDIF initialization table


mreynolds389:
**Issue Description**

Under a suffix you can initialize the database using a table of available LDIF files.  There is an action column with just one option (to import it).  This should just be a single button since there are no other options available.

mreynolds389:
1c67860c0..e7ba48e57  389-ds-base-3.1 -> 389-ds-base-3.1
40daa691b..c15cf21f6  389-ds-base-3.0 -> 389-ds-base-3.0
f92e424a9..4700338a2  389-ds-base-2.8 -> 389-ds-base-2.8
44947cfaa..dc5d02460  389-ds-base-2.7 -> 389-ds-base-2.7

PR #7332, opened 2026-03-12
Issue - UI - Improve suffix import LDIF table

mreynolds389:
Description:

Under a suffix you can initialize the database using a table of available LDIF files. There is an action column with just one option (to import it). This should just be a single button since there are no other options available.

relates: https://github.com/389ds/389-ds-base/issues/7331

## Summary by Sourcery

Enhancements:
- Replace the actions dropdown in the LDIF table with a single inline import button for each row and adjust the column layout accordingly.


droideck (approved):
LGTM

Issue #7322, opened 2026-03-09
Do not allow creating a replication agreement that points to itself


mreynolds389:
**Issue Description**

There is nothing that stops you from adding a replication agreement that points to itself.  This will break replication.  We need to check and reject this.

PR #7336, opened 2026-03-12
Issue 7322 - Reject adding a replication agreement that points to itself

mreynolds389:
Description:

There is nothing that stops you from adding a replication agreement that points to itself. This will break replication. We need to check and reject this.

relates: https://github.com/389ds/389-ds-base/issues/7322

## Summary by Sourcery

Reject replication agreements that point to the local server and surface clear errors in both backend and Cockpit UI while tightening related replica metadata and tests.

Bug Fixes:
- Prevent creation of replication agreements whose target host and port match the local replica, returning an appropriate LDAP error and message.
- Fix Cockpit replication agreement create/update flows to correctly handle structured error responses instead of raw CLI output.
- Correct usage of the bootstrap bind password field when creating replication agreements in the Cockpit UI.

Enhancements:
- Track replica hostname and ports on the replica object and expose accessors for use by replication logic.
- Improve Cockpit wizard spacing by adjusting progress stepper margins across database, replication, and security views.

Tests:
- Add an acceptance test ensuring self-referential replication agreements are rejected with LDAP_UNWILLING_TO_PERFORM.

### DIFF
--- a/dirsrvtests/tests/suites/replication/acceptance_test.py
+++ b/dirsrvtests/tests/suites/replication/acceptance_test.py
@@ -693,6 +693,22 @@ def test_invalid_agmt(topo_m4):
         assert False
 
 
+def test_reject_self_referential_agmt(topo_m4):
+    """Test that creating a replication agreement pointing to self is rejected
+
+    :id: 7df2f91a-878b-4a7e-86f7-4f72ef5726c6
+    :setup: MMR with four suppliers
+    :steps:
+        1. Try to add an agreement on supplier1 that points to supplier1 host/port
+    :expectedresults:
+        1. Agreement creation should fail with UNWILLING_TO_PERFORM
+    """
+    m1 = topo_m4.ms["supplier1"]
+
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        m1.agreement.create(suffix=DEFAULT_SUFFIX, host=m1.host, port=m1.port)
+
+
 def test_warning_for_invalid_replica(topo_m4):
     """Testing logs to indicate the inconsistency when configuration is performed.
 
--- a/ldap/servers/plugins/replication/repl5_agmtlist.c
+++ b/ldap/servers/plugins/replication/repl5_agmtlist.c
@@ -26,6 +26,8 @@
 #define CONFIG_FILTER "(objectclass=nsds5replicationagreement)"
 #define WINDOWS_CONFIG_FILTER "(objectclass=nsdsWindowsreplicationagreement)"
 #define GLOBAL_CONFIG_FILTER "(|" CONFIG_FILTER WINDOWS_CONFIG_FILTER " )"
+#define AGMT_OK 0
+#define AGMT_ERROR -1
 
 
 PRCallOnceType once = {0};
@@ -123,24 +125,29 @@ agmtlist_agmt_exists(const Repl_Agmt *ra)
     return exists;
 }
 
-
 /*
  * Note: when we add the new object, we have a reference to it. We hold
  * on to this reference until the agreement is deleted (or until the
  * server is shut down).
+ * Return AGMT_OK on success
+ * Return AGMT_ERROR on generic error
  */
-int
-add_new_agreement(Slapi_Entry *e)
+static int
+add_new_agreement(Slapi_Entry *e, char **err_buffer)
 {
-    int rc = 0;
+    int rc = AGMT_OK;
     Repl_Agmt *ra = agmt_new_from_entry(e);
     Slapi_DN *replarea_sdn = NULL;
     Replica *replica = NULL;
     Object *ro = NULL;
+    char *agmt_hostname = NULL;
+    char *err_str = NULL;
+    char *server_err_str = NULL;
+    int agmt_port = 0;
 
     /* tell search result handler callback this entry was not sent */
     if (ra == NULL)
-        return 1;
+        return AGMT_ERROR;
 
     ro = object_new((void *)ra, agmt_delete);
     objset_add_obj(agmt_set, ro);
@@ -149,13 +156,43 @@ add_new_agreement(Slapi_Entry *e)
     /* get the replica for this agreement */
     replarea_sdn = agmt_get_replarea(ra);
     if (!replarea_sdn) {
-        return 1;
+        *err_buffer = slapi_ch_smprintf("problem getting the replica suffix for the agreement");
+        return AGMT_ERROR;
     }
     replica = replica_get_replica_from_dn(replarea_sdn);
     slapi_sdn_free(&replarea_sdn);
 
-    rc = replica_start_agreement(replica, ra);
+    agmt_hostname = agmt_get_hostname(ra);
+    agmt_port = agmt_get_port(ra);
 
+    if (replica && slapi_is_local_host(agmt_hostname, &err_str, &server_err_str) &&
+        (agmt_port == replica_get_port(replica) || agmt_port == replica_get_secure_port(replica)) )
+    {
+        /* You can not create an agreement that points to the local server */
+        *err_buffer = slapi_ch_smprintf("agreement hostname points to the local server instance: %s",
+                                        agmt_hostname);
+        slapi_ch_free_string(&agmt_hostname);
+        return AGMT_ERROR;
+    }
+
+    /* Log warnings if there is an issue checking the agreement or local hostname */
+    if (err_str) {
+        slapi_log_err(SLAPI_LOG_WARNING, repl_plugin_name, "add_new_agreement - "
+                      "(%s) problem with the agreement hostname (%s) - %s",
+                      slapi_entry_get_dn(e), agmt_hostname, err_str);
+    }
+    if (server_err_str) {
+        slapi_log_err(SLAPI_LOG_WARNING, repl_plugin_name, "add_new_agreement - "
+                      "(%s) problem checking the server's hostname - %s",
+                      slapi_entry_get_dn(e), server_err_str);
+    }
+    slapi_ch_free_string(&agmt_hostname);
+
+    /* All good, start the agreement */
+    rc = replica_start_agreement(replica, ra);
+    if (rc != 0) {
+        *err_buffer = slapi_ch_strdup("Problem starting the agreement");
+    }
     return rc;
 }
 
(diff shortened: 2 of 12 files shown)


ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -154,6 +159,18 @@ add_new_agreement(Slapi_Entry *e)
     replica = replica_get_replica_from_dn(replarea_sdn);
     slapi_sdn_free(&replarea_sdn);
 
+    agmt_hostname = agmt_get_hostname(ra);
```
droideck, line 165: IIUC, we need to free `agmt_hostname` after the comparison.

dirsrvtests/tests/suites/replication/acceptance_test.py
```diff
@@ -693,6 +693,22 @@ def test_invalid_agmt(topo_m4):
...
+    :setup: MMR with four suppliers
+    :steps:
+        1. Try to add an agreement on supplier1 that points to supplier1 host/port
+    :expectedresults:
+        1. Agreement creation should fail with UNWILLING_TO_PERFORM
+    """
+    m1 = topo_m4.ms["supplier1"]
+
+    with pytest.raises(ldap.UNWILLING_TO_PERFORM):
+        m1.agreement.create(suffix=DEFAULT_SUFFIX, host=m1.host, port=m1.port)
```
droideck, line 709: It fails on a clean VM for me (the exception is not raised). `m1.host` is different from `m1.config.get_attr_val_utf8('nsslapd-localhost')` in this `topo_m4`.

droideck (changes requested):

dirsrvtests/tests/suites/replication/acceptance_test.py
mreynolds389: Interesting and something I was worried about, but it works for me.  Do you know what the two values are?

ldap/servers/plugins/replication/repl5_agmtlist.c
mreynolds389: > IIUC, we need to free `agmt_hostname` after the comparison.

Indeed, fixed.

dirsrvtests/tests/suites/replication/acceptance_test.py
droideck: m1.host resolves to FQDN and `nsslapd-localhost` to `localhost.localdomain`

I wonder if we need to handle hostname resolution here (e.g., resolve both via getaddrinfo() and compare IPs), or at minimum add explicit loopback address checks? Agreement creation is rare so the DNS overhead should be fine...

mreynolds389: Yeah I was worried this might be needed, okay let me dive into this one.  Hopefully it's not a can of worms... :-)

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -154,6 +159,20 @@ add_new_agreement(Slapi_Entry *e)
...
+    agmt_port = agmt_get_port(ra);
+
+    /* does this match the current replica? */
+    replica_hostname = replica_get_hostname(replica);
+    if (replica_hostname && agmt_hostname &&
+        strcasecmp(agmt_hostname, replica_hostname) == 0 &&
+        (agmt_port == replica_get_port(replica) || agmt_port == replica_get_secure_port(replica)) ) {
+        /* You can not create an agreement that points to the local server */
+        slapi_ch_free_string(&agmt_hostname);
+        return 2;
```
progier389: we should probably better use a #define instead of 2
 #define SAME_REPLICA_ERROR 2

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -154,6 +159,20 @@ add_new_agreement(Slapi_Entry *e)
     replica = replica_get_replica_from_dn(replarea_sdn);
     slapi_sdn_free(&replarea_sdn);
 
+    agmt_hostname = agmt_get_hostname(ra);
+    agmt_port = agmt_get_port(ra);
+
+    /* does this match the current replica? */
+    replica_hostname = replica_get_hostname(replica);
+    if (replica_hostname && agmt_hostname &&
+        strcasecmp(agmt_hostname, replica_hostname) == 0 &&
```
progier389: Simply comparing the host name is not enough to ensure we are not targeting the same host.
(typically if there are several IP interfaces they could have different name)
IMHO you should not really bother about the replica host name but resolve the agmt_hostname
and check if one of the associated ip address is one of the local interface 
(by using getaddrinfo and getifaddrs ?)

progier389 (changes requested):

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -149,11 +161,37 @@ add_new_agreement(Slapi_Entry *e)
...
+    agmt_port = agmt_get_port(ra);
+
+    /* We need to check that the agmt is not pointing to this replica */
+    getaddrinfo(agmt_hostname, NULL, NULL, &info);
+    get_ip_str(info->ai_addr, agmt_ip_str, sizeof(agmt_ip_str));
+    getifaddrs(&ifap);
+
+    /* Loop over all the interfaces looking for a match */
+    for (struct ifaddrs *ifa = ifap; ifa != NULL; ifa = ifa->ifa_next) {
+        get_ip_str(info->ai_addr, replica_ip_str, sizeof(replica_ip_str));
```
droideck: Should that be?
```suggestion
        get_ip_str(ifa->ifa_addr, replica_ip_str, sizeof(replica_ip_str));
```

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -149,11 +161,37 @@ add_new_agreement(Slapi_Entry *e)
...
+        return AGMT_ERROR;
     }
     replica = replica_get_replica_from_dn(replarea_sdn);
     slapi_sdn_free(&replarea_sdn);
 
+    agmt_hostname = agmt_get_hostname(ra);
+    agmt_port = agmt_get_port(ra);
+
+    /* We need to check that the agmt is not pointing to this replica */
+    getaddrinfo(agmt_hostname, NULL, NULL, &info);
```
droideck: I think we need to check for errors here as it can fail because of DNS or something else.
If it returns non-zero, `info` stays NULL, and info->ai_addr is a crash.

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -149,11 +161,37 @@ add_new_agreement(Slapi_Entry *e)
...
+    get_ip_str(info->ai_addr, agmt_ip_str, sizeof(agmt_ip_str));
+    getifaddrs(&ifap);
```
droideck: Looking at the manpage `ifa_addr` may be NULL... So I think we need to check for NULL for ifa->ifa_addr in the loop below.

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -149,11 +161,37 @@ add_new_agreement(Slapi_Entry *e)
...
+
+    /* Loop over all the interfaces looking for a match */
+    for (struct ifaddrs *ifa = ifap; ifa != NULL; ifa = ifa->ifa_next) {
+        get_ip_str(info->ai_addr, replica_ip_str, sizeof(replica_ip_str));
+        if (strcasecmp(agmt_ip_str, replica_ip_str) == 0 &&
+            (agmt_port == replica_get_port(replica) || agmt_port == replica_get_secure_port(replica)) )
```
droideck: I think we might also need null checks here.
Previously, we had a "safety net" in `replica_start_agreement` with `replica`'s NULL-check. But now we use it earlier. So just to be safe...

droideck (changes requested):

ldap/servers/plugins/replication/repl5_agmtlist.c
progier389: Minor point:
I wonder if we should not better move that part in a separate function (in libslapd, maybe util.c)
bool slapi_is_local_host(const char *addr, bool *error)  
both for code readability and re-usability

mreynolds389: Sure that sounds good

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -149,11 +157,24 @@ add_new_agreement(Slapi_Entry *e)
...
+
+    if (replica && slapi_is_local_host(agmt_hostname, &err) &&
+        (agmt_port == replica_get_port(replica) || agmt_port == replica_get_secure_port(replica)) )
+    {
+        /* You can not create an agreement that points to the local server */
+        slapi_ch_free_string(&agmt_hostname);
+        return AGMT_SELF_POINTING;
+    }
+    slapi_ch_free_string(&agmt_hostname);
+
```
progier389: Should probably return AGMT_ERROR if err != 0

mreynolds389: Well we only get errors for DNS issues.  I can see a scenario where you add a disabled agreement for a host  that it not provisioned yet.  I don't think we should reject the agreement in this case.  This PR is just to reject an agreement that points to itself so I'm not sure if we should be validating the host name beyond that criteria.  Thoughts?

progier389: I feel the opposite: :wink: 
IMHO the user will have a better experience if we check for DNS errors
  -  No real benefit to adding an agreement on an host that cannot be resolved it just and add unnecessary burden to the server and trigger messages in the error log ==> better add the agmt once the target server is on line  
  - Checking that hostname may be resolved will detect typos in host name at an early stage
  - Risk of missing a agreement that points to itself if there is a transient issue with the DNS

droideck (approved):
LGTM.

One small nit -- with your `getaddrinfo` call, only first DNS result is checked (but the function can return multiple addresses for a hostname (e.g. both IPv4 and IPv6)).
Rarely we can hit an issue here, though.

progier389 (approved):
LGTM

ldap/servers/plugins/replication/repl5_agmtlist.c
mreynolds389: Ok well we have one CI test regression from this:

```
tests/suites/gssapi_repl/gssapi_repl_test.py::test_gssapi_repl

CRITICAL lib389:agreement.py:1049 Failed to add replication agreement: {'msgtype': 105, 'msgid': 10, 'result': 53, 'desc': 'Server is unwilling to perform', 'ctrls': [], 'info': 'problem with the agreement hostname (ldapkdc2.example.com) - No address associated with hostname'}
```

Now I'm worried about IPA, and others.  Maybe we should log a warning instead of rejecting the update?  But first, I'll try and test this PR against IPA...

ldap/servers/slapd/util.c
```diff
@@ -1997,3 +1998,69 @@ slapi_db_is_lmdb(void)
...
+    char local_ip_str[64] = {0};
+    int32_t rc = 0;
+    bool is_localhost = false;
+
+    rc = getaddrinfo(hostname, NULL, NULL, &info);
+    if (rc != 0) {
+        *err_str = (char *)gai_strerror(rc);
+        return false;
+    }
+    get_ip_str(info->ai_addr, ip_str, sizeof(ip_str));
```
vashirov: `getaddrinfo` can return one or more addresses (A, AAAA, round-robin DNS), but the code only checks for the first result.

vashirov (commented):
I have doubts this will work in HA setup, where 2 suppliers are running on different hosts, but share a common virtual hostname (DNS alias or load-balancer VIP). I want to test it before the merge.

ldap/servers/plugins/replication/repl5_agmtlist.c
```diff
@@ -704,19 +751,21 @@ static int
...
-    if (0 == rc) {
+    rc = add_new_agreement(e, &err_buff);
+    if (rc == AGMT_OK) {
         (*agmtcount)++;
     } else {
         slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name, "handle_agmt_search - "
-                                                        "The replication agreement named \"%s\" could not be correctly parsed. No "
-                                                        "replication will occur with this replica.\n",
-                      slapi_sdn_get_dn(slapi_entry_get_sdn(e)));
+                      "The replication agreement named \"%s\" could not be correctly parsed. No "
```
vashirov, line 765: Do I understand correctly that previously created agreements that no longer pass the checks will be rejected?

vashirov (approved):
LGTM

ldap/servers/plugins/replication/repl5_agmtlist.c
mreynolds389: > Do I understand correctly that previously created agreements that no longer pass the checks will be rejected?

The only check that prevents the agreement from starting is if it thinks it points to the local host using the same port.

vashirov (approved):
LGTM

Issue #7339, opened 2026-03-12
Offline export unexpectedly converts DN to lowercase


mreynolds389:
**Issue Description**

Online and offline exports behave differently.  Offline (db2ldif) reconstructs the lower portion of the DN components to lower case (the RDN stays intact though):

    dn: uid=User,ou=People,dc=EXAMPLE,dc=COM

^^^ gets converted to:

    dn: uid=User,ou=people,dc=example,dc=com

The export should preserve the case during online or offline database export.

tbordaz:
Should we back port it to the LMDB Tech Preview relevant 389-ds-base-2.x branches?

PR #7341, opened 2026-03-16
Issue 7339 - Return the exact DN during export

mreynolds389:
Description:

During an export if the entry is not in the DN cache we rebuild the DN from scratch using the entryrdn index, but this resets the case of the DN.

If we want to use the original case of the DN, and the dsEntryDN attribute is present then use that DN for exported ldif.

relates: https://github.com/389ds/389-ds-base/issues/7339

## Summary by Sourcery

Preserve original distinguished name (DN) casing when exporting entries to LDIF using db2ldif, leveraging stored dsEntryDN values when configured.

New Features:
- Add support for using the stored dsEntryDN attribute during db2ldif export to retain the original DN case when the corresponding configuration option is enabled.

Bug Fixes:
- Ensure offline db2ldif exports no longer normalize or alter DN case when a dsEntryDN value is available, preventing case changes in exported LDIF.

Tests:
- Add a regression test that verifies db2ldif exports preserve the exact DN case in the generated LDIF file.

### DIFF
--- a/dirsrvtests/tests/suites/export/export_test.py
+++ b/dirsrvtests/tests/suites/export/export_test.py
@@ -9,6 +9,8 @@
 import os
 import pytest
 import subprocess
+import ldap
+from lib389.idm.organizationalunit import OrganizationalUnit
 from lib389.topologies import topology_st as topo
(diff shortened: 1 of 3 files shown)


progier389 (approved):
LGTM

Issue #7053, opened 2025-10-17
Replication. MemberOf plugin generating duplicate changes


vectinx:
**Issue Description**  
The *memberOf* plugin generates duplicate modifications when users are deleted.

Original discussion: https://lists.fedoraproject.org/archives/list/389-users@lists.fedoraproject.org/thread/VGCAUFFNS7DNL65XI5FHPJWV3R3IC7JQ/

**Package Version and Platform:**  
- **Platform:** AlmaLinux 9  
- **Package and version:** 389-ds-base.x86_64 2.6.1-11.el9_6

**Steps to Reproduce**  
1. Install two FreeIPA servers and configure replication between them.
2. Add a user on one server:
   ```bash
   ipa user-add test --first test --last test
   ```
3. Delete the previously created user:
   ```bash
   ipa user-del test
   ```
4. Check the replication changelog:
   ```bash
   dbscan -f /var/lib/dirsrv/slapd-TEST-LOC/db/userRoot/replication_changelog.db
   ```

The changelog shows duplicate modifications made by the *memberOf* plugin — the user is removed from a group once on one replica, and again on the other.
Example:
```
dbid: 68edda57000000040000
        encrypted: no
        replgen: 1760418390 Tue Oct 14 08:06:30 2025
        csn: 68edda57000000040000
        uniqueid: 316c0981-a8bb11f0-b7c78566-a52e6fbb
        dn: uid=test2,cn=users,cn=accounts,dc=test,dc=loc
        operation: delete

dbid: 68edda57000100040000
        encrypted: no
        replgen: 1760418390 Tue Oct 14 08:06:30 2025
        csn: 68edda57000100040000
        uniqueid: 316c0983-a8bb11f0-b7c78566-a52e6fbb
        dn: cn=test,cn=groups,cn=accounts,dc=test,dc=loc
        operation: modify
                member: uid=test2,cn=users,cn=accounts,dc=test,dc=loc
                modifiersname: cn=referential integrity postoperation,cn=plugins,cn=config
                modifytimestamp: 20251014050630Z
                entryusn: 2913

dbid: 68edda57000200040000
        encrypted: no
        replgen: 1760418390 Tue Oct 14 08:06:30 2025
        csn: 68edda57000200040000
        uniqueid: 02ac4fb5-a83711f0-a3868566-a52e6fbb
        dn: cn=ipausers,cn=groups,cn=accounts,dc=test,dc=loc
        operation: modify
[...]

PR #7345, opened 2026-03-17
Issue 7053 - Add tests for duplicate member operations

aadhikar:
Description: Validates MemberOf plugin no longer creates duplicate changelog entries when users are deleted or renamed in replicated topology.

Relates: #7053

Reviewed by: ???

## Summary by Sourcery

Add regression tests to verify that MemberOf and ReferentialIntegrity plugins do not generate duplicate group membership changes in replicated topologies when users are deleted or renamed.

Tests:
- Add replicated-topology test covering user deletion to ensure groups and changelogs do not contain duplicate member removal operations.
- Add replicated-topology test covering user rename to ensure groups and changelogs correctly reflect membership changes without duplicates.

### DIFF
--- a/dirsrvtests/tests/suites/referint_plugin/memberof_duplicate_test.py
+++ b/dirsrvtests/tests/suites/referint_plugin/memberof_duplicate_test.py
@@ -1,0 +1,314 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


tbordaz (commented):
It looks very nice.
Just a question regarding the number of updates in the changelog related to the deletion of the user from each group. You are testing 'deleted_count_s1 <= 1'.
IIUC it should be strictly equal to 0, am I correct ?

tbordaz (approved):
LGTM

PR #7347, opened 2026-03-17
Issue 7346 - DS does not handle escape char in bind user

### LINKED ISSUE #4383: group rdn with leading space char and add fails error 21 invalid syntax and delete fails error 32

### LINKED ISSUE #7346: DS  does not handle escape char in bind user
**Issue Description**
When normalizing a DN, the server keeps the escaped spaces in the assertion values.
This was introduced with [389-ds-base/issues/4383](https://github.com/389ds/389-ds-base/issues/4383)
 [rfc4514](https://datatracker.ietf.org/doc/html/rfc4514#section-3) requires to keep them only for the heading/leading spaces. Spaces in the middle of the assertion should be normalized as SPACE.

 since 1.4.3

**Steps to Reproduce**
Create an entry like 'uid=my uid,dc=com' then bind with 'uid=my\\20uid,dc=com'

**Expected results**
escaped spaces in the middle  assertion values of RDNs should be normalize as ' '

tbordaz:
Bug description:
	When normalization assertion value in a DN, the fix #4383
        keeps the escaped spaces (i.e '\20') whatever their position
	in the value.
	The RFC requires this for heading spaces but not for the others.
	This prevent to authenticate with a DN containing escaped spaces
	in the middle of assertion value
Fix description:
	Apply the fix #4383 only for the 1rst value of the assertion.

fixes: #7346

Reviewed by:

## Summary by Sourcery

Handle escaped spaces in DNs correctly during normalization to allow binds with non-leading escaped spaces.

Bug Fixes:
- Fix DN normalization so that escaped spaces (\20) are only preserved for the first attribute value, allowing authentication with DNs containing escaped spaces in the middle of the value.

Tests:
- Add a regression test that creates a user with a space in the RDN and verifies binding and searching using both escaped and unescaped spaces in the DN and search filter.

### DIFF
--- a/ldap/servers/slapd/dn.c
+++ b/ldap/servers/slapd/dn.c
@@ -872,7 +872,16 @@ slapi_dn_normalize_ext(char *src, size_t src_len, char **dest, size_t *dest_len)
                     if (n == 0) { /* don't change \00 */
                         *d++ = *++s;
                         *d++ = *++s;
-                    } else if (n == 32) { /* leave \20 (space) intact */
+                    } else if ((state == INVALUE1ST) &&
+                               ((n == 32) || (n ==35))) {
+                        /*
+                         * RFC 4514
+                         * ; The following characters are to be escaped when they appear
+                         * ; in the value to be encoded: ESC, one of <escaped>, leading
+                         * ; SHARP or SPACE, trailing SPACE, and NULL.
+                         *
+                         * Leave intact the heading SPACE and SHARP
+                         */
                         *d++ = *s;
                         *d++ = *++s;
                         *d++ = *++s;
--- a/dirsrvtests/tests/suites/syntax/acceptance_test.py
+++ b/dirsrvtests/tests/suites/syntax/acceptance_test.py
@@ -13,7 +13,7 @@ from lib389.schema import Schema
 from lib389.config import Config
 from lib389.idm.user import UserAccounts
 from lib389.idm.group import Group, Groups
-from lib389._constants import DEFAULT_SUFFIX
+from lib389._constants import DEFAULT_SUFFIX, PW_DM
 from lib389.topologies import log, topology_st as topo
 
 pytestmark = pytest.mark.tier0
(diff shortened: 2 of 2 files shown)


ldap/servers/slapd/dn.c
```diff
@@ -872,7 +872,8 @@ slapi_dn_normalize_ext(char *src, size_t src_len, char **dest, size_t *dest_len)
                     if (n == 0) { /* don't change \00 */
                         *d++ = *++s;
                         *d++ = *++s;
-                    } else if (n == 32) { /* leave \20 (space) intact */
+                    } else if ((state == INVALUE1ST) && (n == 32)) {
+			/* leave \20 (space) intact for heading space */
```
mreynolds389: Indentation is off :-)

Also what about other approved escaped characters like quotes, and equal signs?

tbordaz: The RFC says regarding the attributeValue
<pre>
; The following characters are to be escaped when they appear
; in the value to be encoded: ESC, one of <escaped>, leading
; SHARP or SPACE, trailing SPACE, and NULL.
</pre>
So IIUC for this specific leading character we should preserve leading SHARP and SPACE.
I will had it

mreynolds389 (approved):

Issue #6724, opened 2025-04-02
Log fine grained details of operation timing


mreynolds389:
**Issue Description**

It might be useful to add another keyword to an access log RESULT that reports how long we spent evaluating ACL's for a given operation.  something like:   acltime=##.###

I'm not sure if this can be easily accomplished by just tracking the time spent in slapi_access_allowed(), but it might be worth looking into.

progier389:
I think that we should think more globally about fine detail operation timing.
Other times are useful depending of what we track:
-  filter candidate list evaluation
-  filter verifying (the check done on each candidates)
-  acl evaluation.
-  plugin operation handling (i.e preop/bepreop/bepostop/postop ...) ideally per plugin
- time to read/send data over the network

But that is a lot of data so we probably want to have a kind of operation-timing-detail parameter to enable/disable these timing.

tbordaz:
One way to log it (per operation) would be to use the existing framework https://www.port389.org/docs/389ds/design/log-operation-stats.html

mreynolds389:
> One way to log it (per operation) would be to use the existing framework https://www.port389.org/docs/389ds/design/log-operation-stats.html

I like this as we can add log levels to it, and have the fine grain stats that @progier389 mentioned.

Issue #6326, opened 2024-09-16
high wtime may not be related to worker starvation


tbordaz:
**Issue Description**
In access logs, high wtime is a sign of worker starvation and often leads to increase nsslapd-threadnumber.
It may occur that high wtime is related to reading the client request rather then waiting for a worker. (see [here](https://github.com/389ds/389-ds-base/pull/6285#issuecomment-2270642760))

The discussion kind of agree to keep wtime to time spent by a new request in the wait queue (waiting for a worker) while a new report 'iotime' would report the time spent by the selected worker to read the full request.

**Package Version and Platform:**
all release

**Steps to Reproduce**
unknown. Possibly having a client sending request (large?) slowly

**Expected results**
wtime should stick to time to get a worker
iotime should report the time spent, by the worker, to read the request

progier389:
I agree: we could easily test this one by writing a python test  that use directly the socket API to send a trivial ldap request but split the ldap ber in at least two socket.send()  waiting a few seconds between sending the PDUs) 

The hardest thing to check is the time to read the results (er could also wait but the network buffer will probaly makes things more complex (maybe the result should be long enough to fill up the buffers ...)

PR #7350, opened 2026-03-19
Issue 6724 - Log fine grained details of operation timing

progier389:
- Implement 2 new timing measure:
   wqtime (time spent to wait for working threads)
   write time (time spend to write results over the network (included poll time if needed)
- Allow to select set of timing measure among
wqtime, wtime, writetime, optime, etime

Design document is https://www.port389.org/docs/389ds/design/fine-grain-operation-timing.html

Issues: #6724 and #6326 

Reviewed by: @droideck . @mreynolds389, @tbordaz (Thanks !)

## Summary by Sourcery

Add configurable fine-grained operation timing and expose it in access logs.

New Features:
- Introduce fine-grained operation timing counters for different phases of request processing, including work queue, wait, processing, write, and total times.
- Add a configurable ds-fine-grain-operation-timing setting to control which timing metrics are collected and logged, with sensible defaults.

Enhancements:
- Include fine-grained timing data in both text and JSON access log formats, replacing legacy etime/wtime/optime fields with a unified, extensible representation.
- Track network write time for both regular and SASL I/O paths to better understand response delivery latency.

Tests:
- Add integration tests to verify fine-grained timing configuration combinations and to ensure work-queue wait time is reflected correctly in access logs.

### DIFF
--- a/ldap/admin/src/logconv.py
+++ b/ldap/admin/src/logconv.py
@@ -881,7 +881,7 @@ class logAnalyser:
                     return False
 
                 # Read Fine Grain Operation Timing
-                pattern = rf'.*\s([a-z]+time)=(\S+)\s'
+                pattern = rf'\s([a-z]+time)=([0-9.]+)'
                 for key,val in re.findall(pattern, line):
                     groups[key] = val
 
--- a/ldap/admin/src/logconv.py
+++ b/ldap/admin/src/logconv.py
@@ -503,7 +503,7 @@ class logAnalyser:
                 \serr=(?P<err>\d+)                                  # err=int
                 \stag=(?P<tag>\d+)                                  # tag=int
                 \snentries=(?P<nentries>\d+)                        # nentries=int
-                (?:\s[^=]+time=[^ ]+)*                              # fine grain operation timing
+                (?:\s[a-z]+time=[^ ]+)*                             # fine grain operation timing
                 (?:\sdn="(?P<dn>[^"]*)")?                           # Optional: dn="", dn="strings"
                 (?:,\s+(?P<sasl_msg>SASL\s+bind\s+in\s+progress))?  # Optional: SASL bind in progress
                 (?:\s+notes=(?P<notes>[A-Z]))?                      # Optional: notes[A-Z]
@@ -881,7 +881,7 @@ class logAnalyser:
                     return False
 
                 # Read Fine Grain Operation Timing
-                pattern = rf'.*\s([^=]+time)=(\S+)\s'
+                pattern = rf'.*\s([a-z]+time)=(\S+)\s'
                 for key,val in re.findall(pattern, line):
                     groups[key] = val
 
@@ -3395,7 +3395,7 @@ def main():
         wqtimes = sorted(db.result.wqtime_duration, reverse=True)
         num_wqtimes = len(wqtimes)
         if num_wqtimes > 0:
-            print(f"\n----- Top {db.size_limit} Longest wtimes (wait times) -----\n")
+            print(f"\n----- Top {db.size_limit} Longest wqtimes (work queue times) -----\n")
             for num, wqtime in enumerate(wqtimes):
                 if num >= db.size_limit:
                     break
@@ -3405,7 +3405,7 @@ def main():
         writetimes = sorted(db.result.writetime_duration, reverse=True)
         num_writetimes = len(writetimes)
         if num_writetimes > 0:
-            print(f"\n----- Top {db.size_limit} Longest writetimes (wait times) -----\n")
+            print(f"\n----- Top {db.size_limit} Longest writetimes (write I/O times) -----\n")
             for num, writetime in enumerate(writetimes):
                 if num >= db.size_limit:
                     break
@@ -3526,7 +3526,7 @@ def main():
                 rec_count += 1
 
             if round(avg_writetime, 9) > 0.5:
-                print(f"\n {rec_count}. Your average writetime is {avg_wtime:.9f}. Maybe the application is not reading the results in a timely way.\n")
+                print(f"\n {rec_count}. Your average writetime is {avg_writetime:.9f}. Maybe the application is not reading the results in a timely way.\n")
                 rec_count += 1
 
             if round(avg_optime, 9) > 0:
--- a/ldap/admin/src/logconv.py
+++ b/ldap/admin/src/logconv.py
@@ -307,11 +307,15 @@ class ResultData:
 
     total_etime: float = 0.0
     total_wtime: float = 0.0
+    total_wqtime: float = 0.0
+    total_writetime: float = 0.0
     total_optime: float = 0.0
     etime_stat: float = 0.0
 
     etime_duration: List[float] = field(default_factory=list)
     wtime_duration: List[float] = field(default_factory=list)
+    wqtime_duration: List[float] = field(default_factory=list)
+    writetime_duration: List[float] = field(default_factory=list)
     optime_duration: List[float] = field(default_factory=list)
 
     nentries_num: List[int] = field(default_factory=list)
@@ -499,9 +503,7 @@ class logAnalyser:
                 \serr=(?P<err>\d+)                                  # err=int
                 \stag=(?P<tag>\d+)                                  # tag=int
                 \snentries=(?P<nentries>\d+)                        # nentries=int
-                \swtime=(?P<wtime>\d+\.\d+)                         # wtime=float
-                \soptime=(?P<optime>\d+\.\d+)                       # optime=float
-                \setime=(?P<etime>\d+\.\d+)                         # etime=float
+                (?:\s[^=]+time=[^ ]+)*                              # fine grain operation timing
                 (?:\sdn="(?P<dn>[^"]*)")?                           # Optional: dn="", dn="strings"
                 (?:,\s+(?P<sasl_msg>SASL\s+bind\s+in\s+progress))?  # Optional: SASL bind in progress
                 (?:\s+notes=(?P<notes>[A-Z]))?                      # Optional: notes[A-Z]
@@ -878,6 +880,11 @@ class logAnalyser:
                     self.logger.error(f"Failed to get group from match line: {line} - {e}.")
                     return False
 
+                # Read Fine Grain Operation Timing
+                pattern = rf'.*\s([^=]+time)=(\S+)\s'
+                for key,val in re.findall(pattern, line):
+                    groups[key] = val
+
                 timestamp_raw = groups.get('timestamp')
                 if timestamp_raw:
                     try:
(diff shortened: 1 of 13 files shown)


ldap/servers/slapd/sasl_io.c
```diff
@@ -556,6 +556,7 @@ sasl_io_send(PRFileDesc *fd, const void *buf, PRInt32 amount, PRIntn flags, PRIn
     PRInt32 ret = 0;
     sasl_io_private *sp = sasl_get_io_private(fd);
     Connection *c = sp->conn;
+    Operation *op = c->c_ops; /* In theory during bind there is only a single active operation */
 
     slapi_log_err(SLAPI_LOG_CONNS,
                   "sasl_io_send", "Writing %d bytes\n", amount);
```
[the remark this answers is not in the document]
progier389: sasl_io_send is called during sasl bind operation ==> op cannot be NULL

ldap/servers/slapd/operation.c
```diff
@@ -780,3 +781,37 @@ operation_get_result_controls(const Operation *o)
 {
     return o->o_results.result_controls;
 }
+
+/* Add t1-t2 in cumulative fine grain operation timing slot */
+void
+fgot_compute(struct op *op, fgot_id_t fgot_id, struct timespec *t1, struct timespec *t2)
+{
+    struct timespec elapsed;
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
```
mreynolds389: Doesn't look like we use slapdFrontendConfig

ldap/servers/slapd/operation.c
```diff
@@ -780,3 +781,37 @@ operation_get_result_controls(const Operation *o)
...
+    if (config_check_fgot(fgot_id) || op->o_fgots[fgot_id].enabled) {
+        op->o_fgots[fgot_id].enabled = true;
+        slapi_timespec_diff(t1, t2, &elapsed);
+        slapi_timespec_add(&op->o_fgots[fgot_id].c, &elapsed);
```
droideck, line 808: Do I understand correctly that `fgot_compute` will **accumulate** because of `slapi_timespec_add`?
Won't it corrupt etime/optime values for chaining backend operations?

ldap/servers/slapd/libglobs.c
```diff
@@ -10256,6 +10266,125 @@ config_get_malloc_mmap_threshold()
...
+
+int
+config_set_fgot(const char *attrname, char *value, char *errorbuf, int apply)
+{
+    int retVal = LDAP_SUCCESS;
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+    uint64_t flags = 0;
+
+    if (value != NULL) {
+        char *pt = slapi_ch_strdup(value);
```
droideck, line 10352: Do we need to free `pt`?

ldap/admin/src/logconv.py
```diff
@@ -878,6 +880,11 @@ def _match_line(self, line: str, bytes_read: int):
                     self.logger.error(f"Failed to get group from match line: {line} - {e}.")
                     return False
 
+                # Read Fine Grain Operation Timing
+                pattern = rf'.*\s([^=]+time)=(\S+)\s'
```
droideck: `.*` prefix is greedy... It might backfire and capture only the last.

May be something like that? `r'\b((?:wq|w|op|e|write)time)=(\S+)'`

ldap/admin/src/logconv.py
```diff
@@ -3466,6 +3521,14 @@ def main():
                 print(f"\n {rec_count}. Your average wtime is {avg_wtime:.9f}. You may need to increase the number of worker threads (nsslapd-threadnumber).\n")
                 rec_count += 1
 
+            if round(avg_wqtime, 9) > 0.5:
+                print(f"\n {rec_count}. Your average wqtime is {avg_wqtime:.9f}. You may need to increase the number of worker threads (nsslapd-threadnumber).\n")
+                rec_count += 1
+
+            if round(avg_writetime, 9) > 0.5:
+                print(f"\n {rec_count}. Your average writetime is {avg_wtime:.9f}. Maybe the application is not reading the results in a timely way.\n")
```
droideck: `writetime is {avg_wtime:.9f}.`

The recommendation prints wrong variable.

ldap/servers/slapd/slapi-private.h
```diff
@@ -1663,6 +1680,9 @@ int32_t slapd_log_access_extop_info(slapd_log_pblock *logpb);
 int32_t slapd_log_access_sort(slapd_log_pblock *logpb);
 int32_t slapd_log_access_tls(slapd_log_pblock *logpb);
 int32_t slapd_log_access_tls_client_auth(slapd_log_pblock *logpb);
+void slapi_log_fgot_json(struct op *op, slapd_log_pblock *logpb);
+void slapi_log_fgot_text(struct op *op, char *buff, size_t buflen);
+const char*fgot_get_name(fgot_id_t id);
```
droideck, line 1685:
```suggestion
const char *fgot_get_name(fgot_id_t id);
```

ldap/servers/slapd/slap.h
```diff
@@ -249,6 +251,7 @@ typedef void (*VFPV)(); /* takes undefined arguments */
 #define SLAPD_INVALID_SOCKET_INDEX (-1)
 
 #define ETIME_BUFSIZ 42 /* room for struct timespec */
+#define FGOT_BUFSIZ 200 /* room for fine grtain operation timing */
```
droideck, line 254:
```suggestion
#define FGOT_BUFSIZ 200 /* room for fine grain operation timing */
```

ldap/admin/src/logconv.py
```diff
@@ -3356,6 +3391,26 @@ def main():
                     break
                 print(f"wtime={wtime:<12}")
 
+        # Longest work queue times
+        wqtimes = sorted(db.result.wqtime_duration, reverse=True)
+        num_wqtimes = len(wqtimes)
+        if num_wqtimes > 0:
+            print(f"\n----- Top {db.size_limit} Longest wtimes (wait times) -----\n")
```
droideck: It says `Longest wtime` but the section is about `wqtimes`

droideck (changes requested):

ldap/servers/slapd/operation.c
progier389: Indeed fgot_compute accumulate the time
But I am surprised by your remark. 
For etime and optime fgot_compute is called while logging the RESULT line and as far as I know for a given op struct this occur only once:  
In chaining case the operation is replayed to the target ds but with another connection (and op struct )

ldap/servers/slapd/operation.c
```diff
@@ -780,3 +781,37 @@ operation_get_result_controls(const Operation *o)
...
     return o->o_results.result_controls;
 }
+
+/* Add t1-t2 in cumulative fine grain operation timing slot */
+void
+fgot_compute(struct op *op, fgot_id_t fgot_id, struct timespec *t1, struct timespec *t2)
+{
+    struct timespec elapsed;
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+    if (config_check_fgot(fgot_id) || op->o_fgots[fgot_id].enabled) {
```
tbordaz: If fgot_id config is enabled during an operation, is it possible that o_fgots[fgot_id] counters ('s', 'c') are not set ? so the logged time are invalid ?
we could set 'enabled' only at the operation_new.

ldap/admin/src/logconv.py
progier389: I  would prefer having something more generic and avoid having to impact logconv if a new timing measure is added. 
Maybe rf'.*\s([a-z]+time)=(\S+)\s' to ensure that we do not take strange character as part of the keyword

tbordaz (commented):
LGTM
One question though

ldap/servers/slapd/operation.c
progier389: Good idea. Even for a performance point of view it seems better (so we do not have to retrieve the config flags from another module for every measure

ldap/servers/slapd/operation.c
droideck: IIUC, we have couple of situations where we call `slapi_operation_time_elapsed` on the same op...

---

Chaining (conditional — only when timelimit is set):
  - First call: cb_search.c:159 -- slapi_operation_time_elapsed(op, &elapsed) -- only when timelimit != -1 && timelimit != 0 (line 153)
  - Second call: result.c:2323 via op_shared_search:1069 → send_ldap_result -> send_ldap_result_ext:773 -> log_result

  Both operate on the same op from the pblock. The etime would be inflated by the amount of elapsed time at the first call.

----

  Abandon (paged results only):
  - First call: abandon.c:129 -- slapi_operation_time_elapsed(o, &o_copy.hr_time_end) on the target op
  - Second call: result.c:2323 via opshared.c:979 -> send_ldap_result -> send_ldap_result_ext:773 -> log_result

  Same op struct o. The etime would be accumulated.

ldap/servers/slapd/accesslog.c
```diff
@@ -614,6 +614,47 @@ slapd_log_access_modrdn(slapd_log_pblock *logpb)
...
+    for (fgot_id_t id=0; id<FGOT_MAX; id++) {
+        if (buflen < ETIME_BUFSIZ) {
+            break;
+        }
+        if (op->o_fgots[id].enabled) {
+            struct timespec *t = &op->o_fgots[id].c;
+            size_t len;
+            snprintf(buff, ETIME_BUFSIZ, "%" PRId64 ".%.09" PRId64 "",
+                     (int64_t)(t->tv_sec), (int64_t)(t->tv_nsec));
+            logpb->fgot[id] = buff;
```
droideck, line 629: Do I understand correctly that we store the pointers into stack-allocated buff_fgot[FGOT_BUFSIZ] (result.c:2296)? And as it's logpb->fgot[id] -- it will be freed later via slapi_ch_free_string in slapd_log_access_result? I think it might be incorrect (free() on a stack address), unless I miss something.

ldap/admin/src/logconv.py
```diff
@@ -878,6 +880,11 @@ def _match_line(self, line: str, bytes_read: int):
                     self.logger.error(f"Failed to get group from match line: {line} - {e}.")
                     return False
 
+                # Read Fine Grain Operation Timing
+                pattern = rf'.*\s([a-z]+time)=(\S+)\s'
```
droideck: It still shows be `[('etime', '0.004000000')]` on `[25/Mar/2026:10:00:00.123456789 -0400] conn=1 op=2 RESULT err=0 tag=48 nentries=0 wqtime=0.001000000 wtime=0.002000000 optime=0.003000000 etime=0.004000000 notes=A`.
I think `.*` is too greedy for that cause.

ldap/servers/slapd/libglobs.c
```diff
@@ -10256,6 +10266,125 @@ config_get_malloc_mmap_threshold()
...
+    { "e", FGOT_ETIME },
+    { "wtime", FGOT_W },
+    { "w", FGOT_W },
+    { 0 }
+};
+
+const char *
+fgot_allowed_values()
+{
+    static char names[20];
```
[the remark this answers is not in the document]
droideck: I agree that we truncate here. It will fit only something like `"wqtime,wq,writetime," (19 chars + NUL)`

ldap/servers/slapd/libglobs.c
```diff
@@ -10256,6 +10266,119 @@ config_get_malloc_mmap_threshold()
...
```
droideck: The buffer is still too small. 
I think I've seen it was changed at some point, but something went wrong and it's back to old number.

droideck (approved):
Besides the small issue, LGTM!

PR #7356, opened 2026-03-20
Issue 7271 - Add test for retrocl trimming shutdown crash

aadhikar:
Description: Validates server does not crash when shutdown occurs while retro changelog trim thread is actively running.

Relates: #7271

Reviewed by: ???

## Summary by Sourcery

Tests:
- Add a retro changelog test that performs rapid restart cycles during active trimming to detect shutdown crashes.


progier389 (approved):
LGTM

PR #7373, opened 2026-03-30
Issue 7372 - Reindex adds tombstones to ancestorid causing export failures

### LINKED ISSUE #7372: Reindex adds tombstones to ancestorid causing export failures
**Issue Description**
During import/reindex, tombstone entries are added to the ancestorid index. When those tombstones are later purged, the purge thread skips updates to ancestorid index. This leaves stale entry IDs in the ancestorid index referencing entries in id2entry that are no longer there.

 - Platform: RHEL/Fedora
 - Package and version: 3.x

**Steps to Reproduce**
1. Add entries on S1
2. Delete entries on S1
3. Perform additional modifications to advance the RUV
4. Reindex S1
5. Configure aggressive tombstone purging and wait for tombstones to be purged on S1
6. Export userRoot with -s ou=people,dc=example,dc=com

```
[root@fedora43 ds]# dsconf supplier1 backend export userroot -s ou=people,dc=example,dc=com
Error: Export task failed
-------------------------
Beginning export of 'userRoot'
Backend userRoot: Failed to read entry 21, err -30798
userRoot: Processed 2 entries (100%).
backend 'userRoot' export failed (-1)
Export failed.
[root@fedora43 ds]# dbscan -f /var/lib/dirsrv/slapd-supplier1/db/userroot/ancestorid.db -r | grep -w 21
        2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25
        7 21 22 23 24 25
[root@fedora43 ds]# dbscan -f /var/lib/dirsrv/slapd-supplier1/db/userroot/id2entry.db -K 21
Can't set cursor to returned item: Database operation error: Key not found (or no more keys)
```

vashirov:
Bug Description:
During import/reindex, tombstone entries are added to the ancestorid index. When those tombstones are later purged, the purge thread skips updates to ancestorid index. This leaves stale entry IDs in the ancestorid index referencing entries in id2entry that are no longer there.

Fix Description:
Skip tombstone entries when building the ancestorid index during import/reindex.

Fixes: https://github.com/389ds/389-ds-base/issues/7372

## Summary by Sourcery

Prevent ancestorid index from including tombstone entries during import and reindex operations.

Bug Fixes:
- Avoid adding tombstone entries to the ancestorid index during db2index and import, preventing stale ancestorid references that break subsequent exports.

Tests:
- Add an export regression test that verifies exporting a subtree succeeds after reindexing and aggressive tombstone purging in a replicated topology.

### DIFF
--- a/dirsrvtests/tests/suites/export/export_reindex_tombstone_test.py
+++ b/dirsrvtests/tests/suites/export/export_reindex_tombstone_test.py
@@ -1,0 +1,107 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import os
+import time
+import pytest
+import ldap
+from lib389.idm.user import UserAccounts
+from lib389.idm.domain import Domain
+from lib389.tombstone import Tombstones
+from lib389.topologies import topology_m2 as topo_m2
+from lib389._constants import DEFAULT_SUFFIX, DEFAULT_BENAME, ErrorLog
+from lib389.utils import *
+from lib389.backend import Backends
+from lib389.replica import Replicas, ReplicationManager
+
+pytestmark = pytest.mark.tier1
+
+
+def test_export_after_reindex_and_tombstone_purge(topo_m2):
+    """Test that export with -s works after reindex and tombstone purge.
+
+    :id: 8c5cb603-1f43-46ad-9935-5f518d6d7fe0
+    :setup: Two supplier replication topology
+    :steps:
+        1. Add entries under ou=people on S1, wait for replication to S2
+        2. Delete the entries on S1, wait for replication
+        3. Perform additional modifications to advance the RUV
+        4. Reindex S1
+        5. Configure aggressive tombstone purging and wait for tombstones to be purged on S1
+        6. Export ou=people,dc=example,dc=com with -s
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+        6. Export completes successfully
+    """
+    S1 = topo_m2.ms["supplier1"]
+    S2 = topo_m2.ms["supplier2"]
+    PEOPLE = f"ou=people,{DEFAULT_SUFFIX}"
+
+    users = UserAccounts(S1, DEFAULT_SUFFIX, rdn="ou=people")
+    test_users = []
+    for i in range(5):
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
progier389, line 3069: IMHO we should not call prepare_ids and dbmdb_import_writeq_push in tombstone case

vashirov: Isn't parentid required to maintain tombstonenumsubordinates?

progier389: The parentid operational attribute within the tombstone entry should indeed exists to compute tombstonenumsubordinates
But I am not sure its associated key/value should also be in the parentid index file.

I guess the big point is whether or not tombstone purging left pending ids in parentid index or not.
 (If it does not that is ok but if it does we have to prevent db2index to add tombstone in parentid index)

dirsrvtests/tests/suites/export/export_reindex_tombstone_test.py
```diff
@@ -0,0 +1,108 @@
...
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import os
+import time
+import pytest
+import ldap
+from lib389.idm.organizationalunit import OrganizationalUnits
```
droideck: OrganizationalUnits is unused

droideck (approved):
LGTM

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
tbordaz: During tombstone purging why not deleting 'tombstoneID' value from parentid/ancestorid whether it exists or not ?

vashirov: @progier389, when an entry is converted to a tombstone, parentid is re-added to the parentid index: https://github.com/389ds/389-ds-base/blob/03be3fb31da58c0edf5fcad158f2e50803bf2917/ldap/servers/slapd/back-ldbm/ldbm_delete.c#L958-L966
When tombstones are purged, `index_addordel_entry()` iterates through all attributes including parentid and properly removes them. So no stale parentid entries are left in the index after the purge.

@tbordaz, since tombstones are never supposed to be in the ancestorid index in the first place, with this fix I think cleaning them up would be a no-op.

PR #7376, opened 2026-04-01
Issue 7375 - CI - Fix clu/dsconf_tasks_test

### LINKED ISSUE #7375: CI - Fix clu/dsconf_tasks_test
Issue Description
The dsconf_tasks_test fails due to checking for a string that does not always appear in the command output

Steps to Reproduce

    suites/clu/dsconf_tasks_test.py::test_dsconf_task_watch_output
    See error

AssertionError: 'Copying' should be in the output

Expected results
Test should pass

mirielka:
Description:
The dsconf_tasks_test fails because of checking for a string that is not always included in the command output. Modifying the test so that is checks for a string that is included in the command output with --watch option, but not without it.

Fixes: #7375
Author: Lenka Doudova
Reviewer: ???

## Summary by Sourcery

Bug Fixes:
- Update dsconf backup create/restore watch output checks to look for stable, mode-specific messages rather than an inconsistent string.


progier389 (approved):
LGTM

Issue #7378, opened 2026-04-01
Make sure suffix entry always gets assigned ID 1


mreynolds389:
**Issue Description**

You can enable replication on an empty database.  It will add the RUV tombstone entry to the database as the first entry, but the entry ID should not be 1 as LMDB expects the suffix to always have  ID 1.  This can lead to odd initialization issues.

Instead we need to assign it ID 2 so that ID 1 is reserved for the suffix root node entry

PR #7379, opened 2026-04-01
Issue 7378 - Make sure suffix entry always gets assigned ID 1

mreynolds389:
Description:

When enabling replication and we add the database RUV tombstone entry check if this is the first entry being added.  If it is, then set the ID to 2, and set a flag indicating that after we add the suffix entry to bump the next ID count past the RUV entry id.

And at server startup when we get the "next id from disk" also check if there if only the RUV entry.  If it is, then set the next ID to 1 (for the suffix) and set the ruv flag that indicating that the entry ID count needs to be bumped past the RUV entry when the suffix is finally added.

For some reason dbscan returns error 3 even on success.  So a parameter was added to lib389's dbscan interface to skip checking the return code.

relates: https://github.com/389ds/389-ds-base/issues/7378

CI test assisted by: Cursor

## Summary by Sourcery

Ensure correct assignment and recovery of entry IDs when the RUV tombstone is created before the suffix entry, and relax dbscan return-code checking to support new tests.

Bug Fixes:
- Guarantee the RUV tombstone entry never consumes entry ID 1, reserving ID 1 for the suffix entry even when replication is enabled before the suffix exists.
- Handle server startup when only the RUV entry is present so that the next assigned ID is set correctly for the suffix and subsequent entries.

Enhancements:
- Add a flag to track when the RUV entry was inserted before the suffix so that the next-id counter can be adjusted appropriately.
- Extend lib389's dbscan wrapper with an option to skip checking the subprocess return code when dbscan reports nonstandard success values.

Tests:
- Add a replication test verifying objectclass index keys map to the expected entry IDs when the suffix entry is created after replication is enabled.

### DIFF
--- a/dirsrvtests/tests/suites/replication/ruv_before_suffix_entryid_test.py
+++ b/dirsrvtests/tests/suites/replication/ruv_before_suffix_entryid_test.py
@@ -1,0 +1,133 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+"""Exercise objectclass index entry IDs when the suffix is created after replication."""
+
+import logging
+import os
+import re
+import pytest
+
+from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
+from lib389.idm.domain import Domain
+from lib389.idm.organizationalunit import OrganizationalUnits
+from lib389.replica import Replicas
+from lib389.topologies import topology_no_sample
+
+pytestmark = pytest.mark.tier2
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+
+def _assert_oc_index_missing_key(inst, key):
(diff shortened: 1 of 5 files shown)


progier389 (approved):
LGTM

tbordaz (approved):
LGTM as well

dirsrvtests/tests/suites/replication/ruv_before_suffix_entryid_test.py
```diff
@@ -0,0 +1,129 @@
...
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+
+def _assert_oc_index_missing_key(inst, key):
+    """Dump the objectclass index with dbscan -r; equality key ``key`` must not appear."""
+    inst.stop()
+    try:
+        path = os.path.join(inst.dbdir, DEFAULT_BENAME.lower(), "objectclass")
+        dblib = inst.get_db_lib()
```
droideck: `dblib` is never used

dirsrvtests/tests/suites/replication/ruv_before_suffix_entryid_test.py
```diff
@@ -0,0 +1,129 @@
...
+                f"objectclass index unexpectedly contains equality key {key!r}:\n{out}"
+            )
+
+
+def _assert_oc_index_key_contains_id(inst, key, eid):
+    """Run dbscan -r on the objectclass index; require key and entry ID (see dbscan idl_format)."""
+    inst.stop()
+    try:
+        path = os.path.join(inst.dbdir, DEFAULT_BENAME.lower(), "objectclass")
+        dblib = inst.get_db_lib()
```
droideck: And here

droideck (approved):
The rest LGTM!

PR #7393, opened 2026-04-09
Issue 6753 - Port ticket 48266 test

mirielka:
Description:
Port ticket 48266 test into dirsrvtests/tests/suites/fractional/fractional_test.py using lib389 API.

Relates: #6753
Author: Lenka Doudova
Assisted by: Cursor
Reviewer: ???

## Summary by Sourcery

Port the ticket 48266 fractional replication test into the main fractional test suite using the lib389 API and two-supplier topology_m2 fixtures.

Tests:
- Add new two-supplier fractional replication setup fixture configuring agreements, fractional attributes, and logging for topology_m2.
- Add tests verifying basic replication, non-fractional attribute replication, and CSN evaluation/skip behavior under fractional replication using staging user accounts.
- Remove the legacy ticket48266_test.py test file now that its coverage is provided in the fractional test suite.


dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,279 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+        topology_m2=topology_m2,
+        supplier1=supplier1,
+        supplier2=supplier2,
+    )
+
+
+def _extract_csn(found_csns, supplier):
+    found_csn = (supplier.ds_access_log.parse_line(found_csns[-1]))['rem']
+    found_csn_dict = {i.split("=")[0]: i.split("=")[1] for i in found_csn.split(" ")}
+    return found_csn_dict['csn']
```
droideck, line 478: I think `_extract_csn()` performs redundant fragile parsing of the rem field when `parse_line(` already extracts csn directly into the result dict.
IMO, the old code correctly used found_csn['csn'] directly.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -10,17 +10,26 @@
...
+from lib389.idm.group import Groups
+from lib389.idm.user import UserAccount, UserAccounts
+from lib389.plugins import MemberOfPlugin
+from lib389.replica import ReplicationManager, Replicas
+from lib389.topologies import topology_m2, topology_m2c2
+from lib389.utils import ensure_bytes, ensure_str
+
+log = logging.getLogger(__name__)
 
 pytestmark = pytest.mark.tier1
```
droideck, line 34: Just a note, the test porting has promoted the tests tier2 -> tier1. They might be a bit slow, I think. So double-checking if it was intentional...

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,279 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+    dn5 = f"uid={FRACTIONAL_M2_STAGING_ACCOUNT}5,ou=people,{DEFAULT_SUFFIX}"
+    last_csn = _fractional_m2_get_unreplicated_csn(supplier1, dn5)
+
+    max_loop = 10
+    cnt = 0
+    current_no_more_update = _fractional_m2_count_full_session(supplier1)
+    while current_no_more_update == no_more_update_cnt:
+        cnt = cnt + 1
+        if cnt > max_loop:
+            assert False, f"Timeout waiting for replication sessions to complete after {cnt} loops"
```
droideck, line 624: The old code used break for all three loops (now only one `break`s and other two fail with assertions). This inconsistency changes the test's failure semantics -- slow-but-working replication will now fail the test on the 2nd/3rd loop when it would have passed before.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,279 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+    with open(supplier1.errlog, "r") as file_obj:
+        while True:
+            line = file_obj.readline()
+            if regex.search(line):
+                found_ruv = line
+                break
+            if line == "":
+                break
+        if found_ruv:
+            log.info(f"last operation was found at {file_obj.tell()}")
```
droideck, line 662: Missing assertion that `found_ruv` was actually found. If the RUV line for last_csn isn't in the error log (e.g., log rotated, buffering), found_ruv is None, and the second search loop at line 669 reads from wherever the first loop stopped -- the final assert not found_skip could pass trivially because it never reached the relevant section of the log.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,279 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+
+    supplier1.restart()
+    supplier2.restart()
+
+    repl = ReplicationManager(DEFAULT_SUFFIX)
+    repl.ensure_agreement(supplier1, supplier2)
+    repl.test_replication(supplier1, supplier2)
+
+    def fin():
+        for user in UserAccounts(supplier1, DEFAULT_SUFFIX).list():
```
droideck, line 458: The finalizer deletes all users, not just the 20 it created. I think it can lead to issues for some other test development. Minor thing but can cost some dev time in the future.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,279 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+    user.set("description", description)
+
+    loop = 0
+    while loop <= 10:
+        ent = topology_m2.ms["supplier2"].getEntry(user.dn, ldap.SCOPE_BASE, "(objectclass=*)")
+        if ent.hasAttr("description") and ent.getValue("description") == ensure_bytes(description):
+            break
+        time.sleep(1)
+        loop += 1
+    assert loop <= 10
```
droideck, line 565: The loop can be replaced with `ReplicationManager.wait_for_replication()`

droideck (changes requested):

dirsrvtests/tests/suites/fractional/fractional_test.py
mirielka: Yes, I consulted with Viktor previously, one of the points of porting is to get the tests to suites and tier1, so that they would be executed.

dirsrvtests/tests/suites/fractional/fractional_test.py
mirielka: Unfortunately the old code failed exactly because the parse_line() did not work as expected. The 'found_csn' needs to be processed further to extract only the CSN value. I changed the code to get rid of the extra function though.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,261 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+        regex_skip = re.compile(pattern_skip)
+        found_skip = False
+        while True:
+            line = file_obj.readline()
+            if regex_skip.search(line):
+                found_skip = True
+                break
+            if line == "":
+                break
+        assert found_skip
```
droideck: I'm not sure I fully understand what happens here...

Just above the comment says: `log.info(f"Now check the we can not find the first csn {first_csn} in the log")`

But here you `assert found_skip`.
Shouldn't that be `assert not found_skip` as the original test's intention was?

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,261 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+    found_ops = supplier1.ds_access_log.match(f".*MOD dn=\"{dn}\".*")
+    assert len(found_ops) > 0
+    found_op = supplier1.ds_access_log.parse_line(found_ops[-1])
+
+    found_csns = supplier1.ds_access_log.match(
+        f".*conn={found_op['conn']} op={found_op['op']} RESULT.*")
+    assert len(found_csns) > 0
+
+    found_csn = supplier1.ds_access_log.parse_line(found_csns[-1])
+    csn = found_csn['rem'].split()[-1].split('=')[1]
```
droideck: That might be a bit brittle. Any new field added will break this unexpectedly ( like `wtime/optime` that were added recently)

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -396,6 +405,261 @@ def test_implicit_replication_of_password_policy(_create_entries):
...
+
+def _wait_for_replication_sessions_to_complete(supplier1, no_more_update_cnt):
+    max_loop = 10
+    cnt = 0
+    current_no_more_update = _fractional_m2_count_full_session(supplier1)
+    while current_no_more_update == no_more_update_cnt:
+        cnt = cnt + 1
+        if cnt > max_loop:
+            log.info(f"Timeout waiting for replication sessions to complete after {cnt} loops")
+            break
```
droideck: I have doubts here. It silently times out and logs an info line and returns the unchanged count. The caller has no way to distinguish "replication finished" from "we gave up waiting."

dirsrvtests/tests/suites/fractional/fractional_test.py
mirielka: Thanks for the catch, that was a log message I forgot to modify. This actually works different than originally - originally this message was supposed not to be there because it signified there was an attempt to send changes that should not be sent. Now this message only occurs when changes are not sent because they should not be. I fixed the log message.

dirsrvtests/tests/suites/fractional/fractional_test.py
mirielka: Would it be preferable to have the test fail whenever replication timed out?

dirsrvtests/tests/suites/fractional/fractional_test.py
droideck: The "Skipping update operation" log meaning hasn't actually changed since it was introduced in 2012 (Ticket 317, commit c72f6ba1e991f6e1a4fef5fac04ee474ec846692) -- it has always fired whenever fractional rules strip an op to empty. What did change is that Issue 3903 (commit 55cb7b50cd65f99594f0f97980707848a711f348 + a couple commits int the same ticket) removed the >100-skips threshold that triggered the keep-alive, so the test's "102 MODs -> keep-alive advances RUV -> first_csn no longer re-skipped" premise no longer holds -- so yeah, we might still want to make sure `assert not found_skip` happens but the condition when it happens has changed.

dirsrvtests/tests/suites/fractional/fractional_test.py
droideck: Yes, right now a replication timeout silently falls through to a downstream assert found_ruv failure that gives no hint the real cause was incomplete replication, so failing (or surfacing the timeout to the caller) would make CI failures actionable, IMO.

dirsrvtests/tests/suites/fractional/fractional_test.py
droideck: If I understand correctly the current situation, since 3903 removed the threshold trigger, the only way to actually exercise the keep-alive path inside the test window is to drop `nsds5ReplicaKeepAliveUpdateInterval` to 60 (the enforced min) in the fixture, sleep ~65s after the 102 MODs, then `wait_for_replication` so the update reaches supplier2. Then `assert not found_skip` hopefully works again. 

Currently, as-is we're just asserting that with the default 1h interval the timer never fires inside our window, which isn't really a regression test for anything, IMO.

dirsrvtests/tests/suites/fractional/fractional_test.py
```diff
@@ -10,17 +10,26 @@
...
 from lib389._constants import DEFAULT_SUFFIX
 from lib389.idm.user import UserAccounts, UserAccount
-from lib389.replica import ReplicationManager
+from lib389.replica import ReplicationManager, Replicas
 from lib389.agreement import Agreements
 from lib389.plugins import MemberOfPlugin
 from lib389.idm.group import Groups
 from lib389.config import Config
-import ldap
+from lib389.topologies import topology_m2, topology_m2c2
```
droideck: Probably, should be removed.

droideck (approved):
Besides the import, LGTM!

PR #7395, opened 2026-04-09
Issue 7394 - UI - Manual typing of ports can leave out digits

### LINKED ISSUE #7394: UI - Manual typing of ports can leave out digits
**Issue Description**
When typing to change ports in an existing instance, digits are written into the field with a delay. If the user is typing very quickly and uses Tab to jump between fields it can happen that the last digit(s) are not written into the field and user may overlook that the value set is not the value typed.

**Steps to Reproduce**
1. Create an instance
2. WebUI -> Server -> LDAP Port, LDAPS Port (applicable to both)
3. QUICKLY insert a value and jump to the next field using Tab
4. The last digit typed appears for a fraction of a second, then disappears and comes back with a tiny delay

**Expected results**
Digits should be written without delay

**Additional context**
Reported by Lenka Doudova

jchapma:
Description:
The onChange event triggered port validation on every keystroke, resulting in calls to check for used ports.

Fix:
Moved port validation from onChange to onBlur event.

Fixes: https://github.com/389ds/389-ds-base/issues/7394

Reviewed by:

## Summary by Sourcery

Adjust server settings UI port validation to run on blur instead of on change to avoid premature validation during manual entry.

Bug Fixes:
- Prevent port fields from being incorrectly validated and disabled while users are still typing by deferring port-in-use checks until the input loses focus.

Enhancements:
- Introduce a dedicated on-blur port validation path that updates error state and save-button enablement based on port-in-use checks.

### DIFF
--- a/src/cockpit/389-console/src/lib/server/settings.jsx
+++ b/src/cockpit/389-console/src/lib/server/settings.jsx
@@ -175,6 +175,7 @@ export class ServerSettings extends React.Component {
         this.handleSaveAdvanced = this.handleSaveAdvanced.bind(this);
         this.reloadAdvanced = this.reloadAdvanced.bind(this);
         this.validateSaveBtn = this.validateSaveBtn.bind(this);
+        this.validatePortOnBlur = this.validatePortOnBlur.bind(this);
 
         this.onMinusConfig = (id, nav_tab) => {
             this.setState({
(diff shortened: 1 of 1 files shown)


mirielka (approved):
Works great, looks great, approved!
Thanks!

jchapma:
4108a96ce..f97bab39f  389-ds-base-3.1 -> 389-ds-base-3.1
e7c569bb8..413966066  389-ds-base-3.0 -> 389-ds-base-3.0
19361beb7..827185b74  389-ds-base-2.8 -> 389-ds-base-2.8
88e356ba7..b64674f86  389-ds-base-2.7 -> 389-ds-base-2.7
1f85dac62..602b2e3f4  389-ds-base-2.6 -> 389-ds-base-2.6
1f52decb7..470192f20  389-ds-base-2.4 -> 389-ds-base-2.4

PR #7401, opened 2026-04-10
Issue 7400 - 389-ds handles " *" as "*" in ldapsearch command and ign…

### LINKED ISSUE #7400: 389-ds handles " *" as "*" in ldapsearch command and ignores space character
**Issue Description**
389-ds handles " *" as "*" in ldapsearch command and ignores space character
substring filter containing initial matching ignores spaces and may return entries that does not match the filter

 All releases

**Steps to Reproduce**
Will provide a testcase

**Expected results**
Returned entries to a substring filter should match the filter

tbordaz:
…ores space character

Bug description:
	A initial substring search containing trailing spaces may return
	improper matching entries.
	The RC is that the normalization of values trim leading and trailing
	spaces.

Fix description:
	When trimming (spaces) is requested the caller should specify if
	leading/trailing/both spaces should be trimmed

fixes: #7400

Reviewed by:

## Summary by Sourcery

Make whitespace normalization configurable so callers can control trimming of leading and trailing blanks.

Bug Fixes:
- Prevent substring and value comparisons from incorrectly removing trailing spaces that affect LDAP search matching.

Enhancements:
- Introduce trim mask flags for leading and trailing blanks and propagate them through string and value normalization helpers to support finer-grained control.

### DIFF
--- a/ldap/servers/plugins/syntaxes/cis.c
+++ b/ldap/servers/plugins/syntaxes/cis.c
@@ -1291,8 +1291,15 @@ cis_normalize(
     char **alt)
 {
     int trim_mask = 0;
-    if (trim_spaces) {
-	    trim_mask = TRIM_LEADING_BLANK | TRIM_TRAILING_BLANK;
+    if (trim_spaces == COMPATIBLE_NOT_TRIM_SPACES) {
+        /* value 0x00 */
+        trim_mask = NO_TRIM_SHRINK_BLANK;
+    } else if (trim_spaces == COMPATIBLE_TRIM_SPACES) {
+        /* value 0x01 */
+        trim_mask = COMPATIBLE_TRIM_MASK;
+    } else {
+        /* value 0x10-0x80 */
+        trim_mask = trim_spaces;
     }
     value_normalize_ext(s, SYNTAX_CIS, trim_mask, alt);
     return;
--- a/ldap/servers/plugins/syntaxes/cis.c
+++ b/ldap/servers/plugins/syntaxes/cis.c
@@ -1290,6 +1290,10 @@ cis_normalize(
     int trim_spaces,
     char **alt)
 {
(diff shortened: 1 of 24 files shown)


ldap/servers/plugins/syntaxes/cis.c
```diff
@@ -1290,6 +1290,17 @@ cis_normalize(
     int trim_spaces,
     char **alt)
 {
-    value_normalize_ext(s, SYNTAX_CIS, trim_spaces, alt);
+    int trim_mask = 0;
```
progier389, line 1293: Should probably factorize that code by having a 
value_trim_and_normalize(s, syntax, trim_spaces, alt) function that have this code
then call it in all syntaxes: This will limit the change impact if we need to modify the algorithm

ldap/servers/slapd/back-ldbm/ldbm_search.c
```diff
@@ -355,6 +355,96 @@ ldbm_back_search_cleanup(Slapi_PBlock *pb,
...
+    char *head;
+    char *s;
+    char *end;
+    int ends_with_space = 0;
+
+    if (val == NULL) {
+        return NULL;
+    }
+
+    head = slapi_ch_strdup(val);
```
progier389, line 374: In most case there is nothing to do, so IMHO we should avoid to do systematic strdup 
IMHO it is probably faster to check whether the string must be changed (keeping the place where the first change must be done) then duplicate the string

progier389 (changes requested):
Although the logic looks good there are two points that IMHO need be changed:
 - Factorize the code that handle the trimming compatibility
 - Optimize ldbm_search_substring_shrink_blanks to avoid systematic strdup

progier389:
This PR  is replaced by PR #7727 ( That starts from current state and fix a few things )

Issue #7402, opened 2026-04-10
CLI - allow healthcheck to work when server is stopped


mreynolds389:
**Issue Description**

Currently healthcheck only works when the server is started because many of the tests are doing searches on the database/config.

We can still perform many of these checks by using the DSEldif python class, so we should allow some of the functionality to work when the server is stopped.

PR #7403, opened 2026-04-10
Issue 7402 - CLI - allow healthcheck to work when server is stopped

mreynolds389:
Description:

Use and extend the DSEldif class to handle most of the lint checks that query cn=config. If a check must be skipped a message is displayed to the user.

relates: https://github.com/389ds/389-ds-base/issues/7402

Assisted by: Cursor

## Summary by Sourcery

Enable healthcheck lint checks to run in offline mode using dse.ldif when the directory server is stopped, while skipping checks that require a running instance, and update tests accordingly.

New Features:
- Allow healthcheck/DS lint to operate in an offline mode that reads configuration and index data directly from dse.ldif when the server is down.
- Expose helper utilities to read instance configuration and attributes from dse.ldif, including backend, mapping tree, replica, and database config data.

Enhancements:
- Refactor lint attribute access through helper functions that transparently use LDAP online or dse.ldif offline.
- Extend backend, replica, plugin, config, and tunables lint logic to work consistently whether the server is online or offline, including system index and changelog trimming checks.
- Improve healthcheck CLI JSON output to include metadata about skipped checks and host-level errors while preserving backward compatibility.
- Factor dse.ldif-based LDAP URL resolution into a dedicated dseutils module and update callers.
- Tighten referint plugin shutdown logic and update lint error text for backend mapping trees.

Tests:
- Expand healthcheck, configuration, replication, skew, tunables, and system index tests to cover offline healthcheck runs with the server stopped, and adjust expectations for new checks being listed and reported.

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/healthcheck_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/healthcheck_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2023 Red Hat, Inc.
+# Copyright (C) 2026 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
@@ -34,21 +34,21 @@ def run_healthcheck_and_flush_log(logcap, instance, searched_code=None,
                                   json=False, searched_code2=None,
                                   list_checks=False, list_errors=False,
                                   check=None, searched_list=None):
+    # If we are using BDB as a backend, we will get error DSBLE0006 on new
+    # versions
+    if ds_is_newer("3.0.0") and instance.get_db_lib() == 'bdb' and \
+       (searched_code is CMD_OUTPUT or searched_code is JSON_OUTPUT):
+        searched_code = 'DSBLE0006'
+
     args = FakeArgs()
     args.instance = instance.serverid
     args.verbose = instance.verbose
     args.list_errors = list_errors
     args.list_checks = list_checks
     args.check = check
-    args.exclude_check = []
+    args.exclude_check = ['tunables:thp']
     args.dry_run = False
     args.json = json
-
-    # If we are using BDB as a backend, we will get error DSBLE0006 on new
-    # versions
-    if ds_is_newer("3.0.0") and instance.get_db_lib() == 'bdb' and \
-       (searched_code is CMD_OUTPUT or searched_code is JSON_OUTPUT):
-        searched_code = 'DSBLE0006'
 
     log.info('Use healthcheck with --json == {} option'.format(json))
     health_check_run(instance, logcap.log, args)
@@ -62,12 +62,13 @@ def run_healthcheck_and_flush_log(logcap, instance, searched_code=None,
         log.info('Healthcheck returned searched code: %s', searched_code)
 
     if searched_code2 is not None:
+        code2 = searched_code2
         if ds_is_newer("3.0.0") and instance.get_db_lib() == 'bdb' and \
            (searched_code2 is CMD_OUTPUT or searched_code2 is JSON_OUTPUT):
-            searched_code = 'DSBLE0006'
+            code2 = 'DSBLE0006'
 
-        assert logcap.contains(searched_code2)
-        log.info('Healthcheck returned searched code: %s', searched_code2)
+        assert logcap.contains(code2)
+        log.info('Healthcheck returned searched code: %s', code2)
 
     log.info('Clear the log')
     logcap.flush()
@@ -125,8 +126,10 @@ def test_healthcheck_disabled_suffix(topology_st):
     mt.replace("nsslapd-state", "disabled")
     topology_st.standalone.config.set("nsslapd-accesslog-logbuffering", "on")
 
-    run_healthcheck_and_flush_log(topology_st.logcap, topology_st.standalone, RET_CODE, json=False)
-    run_healthcheck_and_flush_log(topology_st.logcap, topology_st.standalone, RET_CODE, json=True)
+    standalone = topology_st.standalone
+
+    run_healthcheck_and_flush_log(topology_st.logcap, standalone, RET_CODE, json=False)
+    run_healthcheck_and_flush_log(topology_st.logcap, standalone, RET_CODE, json=True)
 
     # reset the suffix state
     mt.replace("nsslapd-state", "backend")
@@ -149,7 +152,15 @@ def test_healthcheck_standalone(topology_st):
     """
 
     standalone = topology_st.standalone
+    standalone.config.set("nsslapd-accesslog-logbuffering", "on")
 
+    standalone.stop()
+    try:
+        run_healthcheck_and_flush_log(topology_st.logcap, standalone, CMD_OUTPUT, json=False)
+        run_healthcheck_and_flush_log(topology_st.logcap, standalone, JSON_OUTPUT, json=True)
+    finally:
+        if not standalone.status():
+            standalone.start()
     run_healthcheck_and_flush_log(topology_st.logcap, standalone, CMD_OUTPUT, json=False)
     run_healthcheck_and_flush_log(topology_st.logcap, standalone, JSON_OUTPUT, json=True)
 
@@ -171,25 +182,42 @@ def test_healthcheck_list_checks(topology_st):
     """
 
     output_list = ['config:passwordscheme',
+                   'config:auditlog_buffering',
+                   # 'config:accesslog_buffering',  Skip test access log buffering is disabled
+                   'config:securitylog_buffering',
+                   'config:unauth_binds',
                    'backends:userroot:cl_trimming',
                    'backends:userroot:mappingtree',
                    'backends:userroot:search',
+                   'backends:userroot:system_indexes',
                    'backends:userroot:virt_attrs',
+                   'backends:userroot:backend_implementation',
+                   'backends:userroot:backend_implementation_cleanup_needed',
+                   'backends:userroot:backend_implementation_config_attributes',
                    'encryption:check_tls_version',
                    'fschecks:file_perms',
                    'refint:attr_indexes',
                    'refint:update_delay',
+                   'memberof:member_attr_indexes',
+                   'memberof:member_globalbackend_lock',
+                   'memberof:member_substring_index',
                    'monitor-disk-space:disk_space',
                    'replication:agmts_status',
                    'replication:conflicts',
+                   'replication:no_ruv',
                    'dseldif:nsstate',
                    'tls:certificate_expiration',
                    'logs:notes',
-                   'tunables:thp',
-                   ]
+                   'tunables:thp']
 
     standalone = topology_st.standalone
 
+    standalone.stop()
+    try:
+        run_healthcheck_and_flush_log(topology_st.logcap, standalone, json=False, list_checks=True, searched_list=output_list)
+    finally:
+        if not standalone.status():
+            standalone.start()
     run_healthcheck_and_flush_log(topology_st.logcap, standalone, json=False, list_checks=True, searched_list=output_list)
 
 
@@ -243,6 +271,12 @@ def test_healthcheck_list_errors(topology_st):
 
     standalone = topology_st.standalone
 
+    standalone.stop()
+    try:
+        run_healthcheck_and_flush_log(topology_st.logcap, standalone, json=False, list_errors=True, searched_list=output_list)
+    finally:
+        if not standalone.status():
+            standalone.start()
     run_healthcheck_and_flush_log(topology_st.logcap, standalone, json=False, list_errors=True, searched_list=output_list)
 
 
@@ -263,18 +297,25 @@ def test_healthcheck_check_option(topology_st):
     """
 
     output_list = ['config:passwordscheme',
+                   'config:auditlog_buffering',
                    # 'config:accesslog_buffering',  Skip test access log buffering is disabled
                    'config:securitylog_buffering',
                    'config:unauth_binds',
                    'backends:userroot:cl_trimming',
                    'backends:userroot:mappingtree',
                    'backends:userroot:search',
+                   'backends:userroot:system_indexes',
                    'backends:userroot:virt_attrs',
+                   'backends:userroot:backend_implementation',
+                   'backends:userroot:backend_implementation_cleanup_needed',
+                   'backends:userroot:backend_implementation_config_attributes',
                    'encryption:check_tls_version',
                    'fschecks:file_perms',
                    'refint:attr_indexes',
                    'refint:update_delay',
                    'memberof:member_attr_indexes',
+                   'memberof:member_globalbackend_lock',
+                   'memberof:member_substring_index',
                    'monitor-disk-space:disk_space',
                    'replication:agmts_status',
                    'replication:conflicts',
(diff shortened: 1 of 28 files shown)


src/lib389/lib389/backend.py
```diff
@@ -529,11 +556,24 @@ def _lint_mappingtree(self):
...
         try:
-            mt = self._mts.get(suffix)
-            if mt.get_attr_val_utf8_l('nsslapd-backend') != bename.lower() and mt.get_attr_val_utf8('nsslapd-state') != 'backend':
+            if server_running:
+                mt = self._mts.get(suffix)
+            else:
+                dse = DSEldif(self._instance)
+                mapping_trees = dse.get_mapping_trees()
+                for mapping_tree in mapping_trees:
+                    suffixes = lint_get_attr_val_utf8_l(mapping_tree, 'nsslapd-suffix')
```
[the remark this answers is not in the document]
droideck: Looks like `sourcery-ai` is not wrong here.
Probably, `get_mapping_trees` can be rewritten to return DSLdapObject instances

src/lib389/lib389/backend.py
```diff
@@ -453,16 +468,24 @@ def __init__(self, instance, dn=None):
...
+
         # First check nsrole
         try:
-            indexes.get('nsrole')
+            if server_running:
+                indexes.get('nsrole')
+            else:
+                index = DSLdapObject(self._instance,
+                                     dn=f'cn=nsrole,cn=index,cn={bename},cn=ldbm database,cn=plugins,cn=config')
+                lint_get_attr_vals_utf8_l(index, "objectclass")
```
droideck: I think you're missing import for this one.

src/lib389/lib389/tunables.py
```diff
@@ -43,23 +45,32 @@ def systemwide_thp_enabled() -> bool:
...
+                            thp_line = line
+                            break
+                    if thp_line is not None:
+                        thp_value = int(thp_line.split()[1])
+                        return bool(thp_value)
+            else:
+                # Get systemctl data for THP from systemd
+                systemctl_data = subprocess.check_output(
+                    ['systemctl', 'show', '-P', 'Environment',
+                     'dirsrv@%s' % self._instance.serverid, 'THP_enabled'])
```
droideck, line 66: `THP_enabled`  is misleading here as `systemctl` thinks it's another instance. `THP_DISABLE=1` is displayed anyway without it.

ldap/servers/plugins/referint/referint.c
```diff
@@ -1370,11 +1370,14 @@ int
 referint_postop_close(Slapi_PBlock *pb __attribute__((unused)))
 {
     /* signal the batch thread to exit */
-    if (referint_get_delay() > 0) {
+    if (keeprunning) {
+        slapi_log_err(SLAPI_LOG_ERR, REFERINT_PLUGIN_SUBSYSTEM, "referint_postop_close - MARK signalling mutex\n");
```
droideck: Some debug messages were MARKed! :)

ldap/servers/plugins/referint/referint.c
```diff
@@ -1370,11 +1370,14 @@ int
...
         pthread_mutex_lock(&keeprunning_mutex);
         keeprunning = 0;
         pthread_cond_signal(&keeprunning_cv);
         pthread_mutex_unlock(&keeprunning_mutex);
+    } else {
+        slapi_log_err(SLAPI_LOG_ERR, REFERINT_PLUGIN_SUBSYSTEM, "referint_postop_close - MARK no singal\n");
```
droideck: Also, `singal` -> `signal` typo.

src/lib389/lib389/dseldif.py
```diff
@@ -232,6 +274,39 @@ def get_backends(self):
...
+        return replicas
+
+    def suffix_replicated(self, suffix):
+        """
+        Check if the suffix is replicated in the DSE.
+        """
+        for entry in self._contents:
+            if fnmatch.fnmatch(entry, "*cn=replica,cn=*,cn=mapping tree,cn=config*"):
+                dn = entry.replace("dn: ", "").strip()
+                value = self.get(dn, "nsDS5ReplicaRoot")[0]
```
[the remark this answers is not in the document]
droideck: I agree with sorcery here. Worth to have a None check here.

src/lib389/lib389/backend.py
```diff
@@ -620,15 +662,22 @@ def _lint_backend_implementation_config_attributes(self):
...
+                dse = DSEldif(self._instance)
+                replicas = dse.get_replicas()
+                if len(replicas) == 0 or not dse.suffix_replicated(suffix):
+                    return
+                cl = DSLdapObject(self._instance, dn=f'cn=changelog,cn={bename},cn=ldbm database,cn=plugins,cn=config')
+            if lint_get_attr_val_utf8(cl, 'nsslapd-changelogmaxentries') is None and \
+               lint_get_attr_val_utf8(cl, 'nsslapd-changelogmaxage') is None:
                 report = copy.deepcopy(DSCLLE0001)
                 report['fix'] = report['fix'].replace('YOUR_INSTANCE', self._instance.serverid)
                 report['check'] = f'backends:{bename}::cl_trimming'
```
droideck: Is this `::` double colon a typo?

dirsrvtests/tests/suites/healthcheck/healthcheck_test.py
```diff
@@ -590,6 +714,11 @@ def test_lint_backend_implementation_wrong_files(topology_st):
     run_healthcheck_and_flush_log(topology_st.logcap, inst, RET_CODE, json=False)
     run_healthcheck_and_flush_log(topology_st.logcap, inst, RET_CODE, json=True)
 
+    inst.stop()
```
droideck, line 717: Should we also have a `try-finally` here and start the server?

src/lib389/lib389/plugins.py
```diff
@@ -483,12 +493,20 @@ def _lint_attr_indexes(self):
                     if not container.endswith(suffix):
                         # skip this backend that is not in the scope
                         continue
+                #if server_running:
```
droideck: Leftover comment

droideck (changes requested):
Overall, looks well rounded and great!

A few points below + What do you think about checking `monitor-disk-space:disk_space` also? It currently reads a `runtime-generated entry` but we probably can reimplement this with os.statvfs() on the known paths (db_dir, config_dir, log_dir), for example.

ldap/servers/plugins/referint/referint.c
mreynolds389: HMmm I don't see this debug logging in my branch, maybe I didn't push an update?  :-)

ldap/servers/plugins/referint/referint.c
mreynolds389: Yeah this is no longer present and was only for debugging

src/lib389/lib389/backend.py
mreynolds389: Indeed it does

src/lib389/lib389/backend.py
mreynolds389: HMm not seeing this.  I think this is outdated.  Refreshing PR...

mreynolds389:
> Overall, looks well rounded and great!
> 
> A few points below + What do you think about checking `monitor-disk-space:disk_space` also? It currently reads a `runtime-generated entry` but we probably can reimplement this with os.statvfs() on the known paths (db_dir, config_dir, log_dir), for example.

It crossed my mind, but the PR was already getting large.  I'll add it though....

mreynolds389:
> > Overall, looks well rounded and great!
> > A few points below + What do you think about checking `monitor-disk-space:disk_space` also? It currently reads a `runtime-generated entry` but we probably can reimplement this with os.statvfs() on the known paths (db_dir, config_dir, log_dir), for example.
> 
> It crossed my mind, but the PR was already getting large. I'll add it though....

Ok added it via a new commit, but some of the tests are failing (unrelated to my commit).  I'll look into them tomorrow...

src/lib389/lib389/backend.py
```diff
@@ -529,11 +557,24 @@ def _lint_mappingtree(self):
...
         try:
-            mt = self._mts.get(suffix)
-            if mt.get_attr_val_utf8_l('nsslapd-backend') != bename.lower() and mt.get_attr_val_utf8('nsslapd-state') != 'backend':
+            if server_running:
+                mt = self._mts.get(suffix)
+            else:
+                dse = DSEldif(self._instance)
+                mapping_trees = dse.get_mapping_trees()
+                for mapping_tree in mapping_trees:
+                    suffixes = lint_get_attr_val_utf8_l(mapping_tree, 'nsslapd-suffix')
```
droideck: I think there's still the issue and sorcery incorrectly "resolved" it.
dse.get_mapping_trees() returns raw strings from self._contents, but the line passes them to lint_get_attr_val_utf8_l() which expects a DSLdapObject with ._instance and ._dn. This causes AttributeError, silently caught by except Exception: continue in health.py. The  offline mappingtree lint check silently does nothing.
Fix: construct DSLdapObject instances from the DNs, like Backends._list_from_dse() does.

src/lib389/lib389/dseldif.py
droideck: Another moment where sorcery incorrectly marked "resolved". 
self.get(dn, "nsDS5ReplicaRoot")[0] will TypeError if get() returns None (which happens when the attribute is missing). Fix: check for None before indexing.

src/lib389/lib389/utils.py
```diff
@@ -2162,3 +2126,24 @@ def validate_max_age(value, ignore_value=None):
...
+    path = os.path.realpath(os.path.abspath(dir))
+    while not os.path.ismount(path):
+        path = os.path.dirname(path)
+    return path
+
+
+def get_disk_space(path):
+    block_size = os.statvfs(path).f_frsize
+    total_size = os.statvfs(path).f_blocks * block_size
+    available_size = os.statvfs(path).f_bavail * block_size
```
droideck: Calling it three times separately instead of once can lead to inconsistency if filesystem state changes between calls (even though unlikely).

src/lib389/lib389/dseldif.py
mreynolds389: I think I marked these are resolved, anyway changes made...

src/lib389/lib389/replica.py
```diff
@@ -1814,6 +1816,17 @@ def create(self, rdn=None, properties=None):
 
         return replica
 
+    def _list_from_dse(self):
+        dse = DSEldif(self._instance)
+        replicas = dse.get_replicas()
+        return replicas
```
droideck: `Replicas._list_from_dse` returns raw strings ("dn: cn=replica,…"), but `Replicas.list()` is contracted to return Replica objects.

I think:

```python
def _list_from_dse(self):
    dse = DSEldif(self._instance)
    replicas = dse.get_replicas()   # list[str]
    return replicas                 # should be list[Replica]
```

src/lib389/lib389/utils.py
```diff
@@ -2162,3 +2126,25 @@ def validate_max_age(value, ignore_value=None):
...
+    return path
+
+
+def get_disk_space(path):
+    stats = os.statvfs(path)
+    block_size = stats.f_frsize
+    total_size = stats.f_blocks * block_size
+    available_size = stats.f_bavail * block_size
+    used_size = total_size - available_size
+    used_percent = (used_size / total_size) * 100
```
droideck, line 2152: I think it's possible `total_size == 0` on bind mounts, tmpfs variants, cgroup mounts. So just in case, I'd clean it here (but we swallow Exception later, so it's not a big issue)

src/lib389/lib389/tunables.py
droideck: `systemctl show -P Environment UNIT1 UNIT2` treats `THP_enabled` as a second unit name...

mreynolds389: > `systemctl show -P Environment UNIT1 UNIT2` treats `THP_enabled` as a second unit name...

Should be fixed now

src/lib389/lib389/replica.py
mreynolds389: > `Replicas._list_from_dse` returns raw strings ("dn: cn=replica,…"), but `Replicas.list()` is contracted to return Replica objects.

Should be fixed now

src/lib389/lib389/utils.py
mreynolds389: > I think it's possible `total_size == 0` on bind mounts, tmpfs variants, cgroup mounts. So just in case, I'd clean it here (but we swallow Exception later, so it's not a big issue)

Should be fixed...

mreynolds389:
@droideck - ok everything should be addressed now, please review...

src/lib389/lib389/cli_ctl/health.py
```diff
@@ -111,9 +127,25 @@ def _run(inst, log, args, checks, exclude_checks):
...
-        except:
+        except ValueError as e:
+            if offline and 'not ONLINE' in str(e):
+                reason = 'instance is not connected (offline); check needs LDAP'
+                skipped.append({'check': s[0], 'reason': reason})
+                if not args.json:
+                    log.info(f" - skipped {s[0]} - requires running server: {reason}")
+            else:
+                raise
+        except Exception:
```
droideck: I know this is pre-existing, but should we at least add a log.debug(f"check {s[0]} failed: {e}") here?

I'm a bit worried that silently swallowing all exceptions could mask real bugs and make them harder to track down later.

src/lib389/lib389/dseldif.py
```diff
@@ -232,6 +274,41 @@ def get_backends(self):
 
         return list(set(backends))
 
+    def get_mapping_trees(self):
+        """Return a list of mapping tree names from DSE.
+        """
+        mapping_trees = []
+        for entry in self._contents:
+            if fnmatch.fnmatch(entry, "*cn=*,cn=mapping tree,cn=config*"):
```
droideck: IMO, the fnmatch patterns here are a bit too loose -- that leading * means they'll also match joined attribute lines, not just DN lines, so a value that happens to contain a mapping-tree-shaped substring could sneak through and give us wrong answers. It's probably unlikely but worth cleaning, IMO.

Could we tighten them to something like `entry.startswith("dn: cn=")` and `",cn=mapping tree,cn=config"` in entry? That matches the pattern get_backends() already uses just above.

src/lib389/lib389/dseldif.py
```diff
@@ -232,6 +274,41 @@ def get_backends(self):
...
+                replicas.append(entry.strip()[4:].strip())
+
+        return replicas
+
+    def suffix_replicated(self, suffix):
+        """
+        Check if the suffix is replicated in the DSE.
+        """
+        for entry in self._contents:
+            if fnmatch.fnmatch(entry, "*cn=replica,cn=*,cn=mapping tree,cn=config*"):
```
droideck: And here.

droideck (approved):
Besides a couple of small issues -- LGTM!

Sorry it took a bit long to walk through the PR!

src/lib389/lib389/cli_ctl/health.py
mreynolds389: Added logging

mreynolds389:
@droideck - ok _now_ I think we are good to go

mreynolds389:
The healthcheck tests (bdb/lmdb wrong files) are consistently failing in github, but they work fine for me.  Might have to add debugging and rerun tests in github...

src/lib389/lib389/backend.py
vashirov:
```suggestion
        if self._instance.ds_paths.db_home_dir is not None and self._instance.ds_paths.db_home_dir != self._instance.ds_paths.db_dir and os.path.isdir(self._instance.ds_paths.db_home_dir):
```


When the server is stopped, `db_home_dir` comes from `defaults.inf` `/dev/shm/slapd-{instance_name}`. But when the server stops, `/dev/shm` gets cleaned up. So `os.listdir(self._instance.ds_paths.db_home_dir)` throws FileNotFoundError. The exception gets swallowed silently.
We should add a check `os.path.isdir()` before listing files there.

Issue #7407, opened 2026-04-13
dbscan -k option lists non-matching entries


mreynolds389:
**Issue Description**

With dbscan you can provide an option to display a single entry that matches a certain key.  However, it uses that key as a starting pint and then display every entry after that key.

This loop will also always report a non-zero result code as it reports the return code from after it hits the end of the db:

    dblayer_cursor_op(&cursor, DBI_OP_NEXT,  &key, &data);

mreynolds389:
557673486..cbd63d5d6  389-ds-base-3.1 -> 389-ds-base-3.1
9f212e532..50d804788  389-ds-base-3.0 -> 389-ds-base-3.0
50d804788..8781c2f3d  389-ds-base-3.0 -> 389-ds-base-3.0 (cherry pick issue)
827185b74..d7cf938c1  389-ds-base-2.8 -> 389-ds-base-2.8
b64674f86..46b2cc149  389-ds-base-2.7 -> 389-ds-base-2.7
602b2e3f4..b35722b54  389-ds-base-2.6 -> 389-ds-base-2.6

PR #7408, opened 2026-04-13
Issue 7407 - dbscan -k option display entries that do not match the specified key

mreynolds389:
Description:

dbscan starts with locating the requested key, but then it continues to display every entry after that one until it hits the end of the db.  As a side effect it also always returns a non-zero result code.

relates: https://github.com/389ds/389-ds-base/issues/7407

### DIFF
--- a/ldap/servers/slapd/tools/dbscan.c
+++ b/ldap/servers/slapd/tools/dbscan.c
@@ -1351,6 +1351,7 @@ main(int argc, char **argv)
     int longopt_idx = 0;
     int c = 0;
     char optstring[2*COUNTOF(options)+1] = {0};
+    char *orig_key = NULL;
 
     if (defdbimpl) {
         if (strcasecmp(defdbimpl, "bdb") == 0) {
@@ -1559,6 +1560,7 @@ main(int argc, char **argv)
     }
 
     if (find_key) {
+        orig_key = slapi_ch_strdup(find_key);
         /* Position cursor at the matching key */
         dblayer_value_set_buffer(be, &key, find_key, strlen(find_key) + 1);
         dblayer_value_free(be, &data);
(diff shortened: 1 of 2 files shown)


progier389 (changes requested):
Beware several records may have the same key (in indexes)
So after seeking the first key you still need to loop on the records using DBI_OP_NEXT_DATA

mreynolds389:
> Beware several records may have the same key (in indexes) So after seeking the first key you still need to loop on the records using DBI_OP_NEXT_DATA

Ah ok, like continuation keys.  I'll rework it

ldap/servers/slapd/tools/dbscan.c
```diff
@@ -1584,8 +1586,12 @@ main(int argc, char **argv)
                 goto done;
             }
             do {
-                display_item(&cursor, &key, &data);
-                ret = dblayer_cursor_op(&cursor, DBI_OP_NEXT,  &key, &data);
+                if (strcmp((char *)key.data, orig_key) == 0) {
```
progier389: I am not sure that key.data is always null terminated
Maybe should rather check key length and use memcmp

ldap/servers/slapd/tools/dbscan.c
```diff
@@ -1584,9 +1586,17 @@ main(int argc, char **argv)
                 goto done;
             }
             do {
-                display_item(&cursor, &key, &data);
-                ret = dblayer_cursor_op(&cursor, DBI_OP_NEXT,  &key, &data);
+                if (memcmp((char *)key.data, orig_key, key.size) == 0) {
```
progier389: should check that (key.size <= strlen(orig_key)+1) && memcmp ... 
otherwise there will be a read buffer overflow on orig_key buffer

mreynolds389: > should check that (key.size <= strlen(orig_key)+1) && memcmp ... otherwise there will be a read buffer overflow on orig_key buffer

Ok I did look at comparing the size, and key was always one larger than origin_key, but I wasn't sure if that was always the case so I abandoned it as it seemed unreliable  :-)

progier389 (approved):
LGTM

PR #7409, opened 2026-04-15
Issue 6641 - MODRDN fails when referint-tracked attr is MUST and SING…

aadhikar:
…LE-VALUE

Description: Fall back to LDAP_MOD_REPLACE when DELETE of a required attribute triggers OBJECT_CLASS_VIOLATION.

Relates: #6641

Reviewed by: ???

## Summary by Sourcery

Ensure referential integrity updates succeed when modifying required single-valued attributes during MODRDN operations.

New Features:
- Add a regression test covering MODRDN of entries with a referential-integrity-tracked attribute that is MUST and SINGLE-VALUE in the schema.

Bug Fixes:
- Handle OBJECT_CLASS_VIOLATION errors in the referential integrity plugin by retrying a DELETE+ADD sequence as a REPLACE when updating required attributes during internal modifies.

Tests:
- Introduce an automated test that defines custom schema, enables referential integrity on a DN-valued attribute, and verifies MODRDN correctly updates the referencing attribute without failing.

### DIFF
--- a/dirsrvtests/tests/suites/plugins/modrdn_test.py
+++ b/dirsrvtests/tests/suites/plugins/modrdn_test.py
@@ -6,9 +6,11 @@
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
 #
+import logging
 import pytest
 from lib389.topologies import topology_st
 from lib389._constants import DEFAULT_SUFFIX
+from lib389._mapped_object import DSLdapObject
 from lib389.idm.group import Groups
 from lib389.idm.user import nsUserAccounts
 from lib389.plugins import (
@@ -18,6 +20,9 @@ from lib389.plugins import (
     MemberOfPlugin,
     ReferentialIntegrityPlugin,
 )
+from lib389.schema import Schema
+
+log = logging.getLogger(__name__)
 
 pytestmark = pytest.mark.tier1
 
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/plugins/modrdn_test.py
```diff
@@ -172,3 +177,83 @@ def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
...
+    owner.rename(new_rdn='uid=testowner_renamed')
+    owner_new_dn = owner.dn
+
+    log.info('Verify the rename succeeded and testOwner was updated')
+    assert owner.get_attr_val_utf8('uid') == 'testowner_renamed'
+
+    updated_owner = owned.get_attr_val_utf8('testOwner')
+    assert updated_owner is not None, "testOwner attribute is missing"
+    assert updated_owner.lower() == owner_new_dn.lower(), (
+        f"testOwner should be '{owner_new_dn}', got '{updated_owner}'"
```
droideck: I think these assertions are not enough.
It doesn't store the original DN and doesn't assert it's gone.
It also doesn't assert exactly one value exists, so a hypothetical scenario where REPLACE adds a second value wouldn't be caught.

IMO, it should:

```python
# store before rename
# rename
# get with get_attr_vals_utf8 (Vals)
# check exactly 1 val
# compare with owner_orig_dn
```

dirsrvtests/tests/suites/plugins/modrdn_test.py
```diff
@@ -172,3 +177,83 @@ def test_modrdn_of_a_member_of_2_automember_groups(topology_st):
     # Check groups dont contain original username
     assert not groupA.is_member(user_orig_dn)
     assert not groupZ.is_member(user_orig_dn)
+
+
+def test_modrdn_with_required_singlevalued_referint_attr(topology_st):
```
droideck: The test adds a lot things, so I think it should have a clean up.

ldap/servers/plugins/referint/referint.c
```diff
@@ -713,21 +713,49 @@ _do_modify(Slapi_PBlock *mod_pb, Slapi_DN *entrySDN, LDAPMod **mods)
...
-        if (allow_repl) {
-            rc = slapi_single_modify_internal_override(mod_pb, entrySDN, mod,
-                                                        referint_plugin_identity, OP_FLAG_REPLICATED);
-        } else {
-            rc = slapi_single_modify_internal_override(mod_pb, entrySDN, mod,
-                                                        referint_plugin_identity, 0);
+        /*
+         * If DELETE failed with OBJECT_CLASS_VIOLATION and the next mod is
+         * an ADD on the same attribute, the attribute is likely required
+         * (MUST) and possibly single-valued.  Removing it first violates
```
droideck: `possibly single-valued` - but I think it also applies to multi-valued for the MUST cases?
Plus, I think it would be mice to test:
- MUST + multi-valued attribute where only one value remains
- non-MUST + SINGLE-VALUE

droideck (changes requested):
Overall, very good!

Besides the points below, I think it's still broken for DELETE of the MUST attributes. I think it should be fixed here as it's basically the same issue.

ldap/servers/plugins/referint/referint.c
```diff
@@ -713,21 +713,49 @@ _do_modify(Slapi_PBlock *mod_pb, Slapi_DN *entrySDN, LDAPMod **mods)
...
+         * the schema.  Fall back to REPLACE which updates the value
+         * atomically without an intermediate empty state.
+         */
+        if (rc == LDAP_OBJECT_CLASS_VIOLATION &&
```
tbordaz, line 735: I think the problem comes from the fix #6641 where an atomic update of all the mods was split into several individual mod that indeed can conflict with the schema.

@jchapma do you remember why we had to replace slapi_modify_internal_set_pb_ext with slapi_single_modify_internal_override ?

JFR slapi_entry_dup, slapi_entry_apply_mods, slapi_entry_schema_check on the duplicated entry are tasks done by atomic update of all mods

tbordaz (commented):
The code looks nice. However I wonder if referint.c could be done another way. Asking @jchapma

ldap/servers/plugins/referint/referint.c
jchapma: > @jchapma do you remember why we had to replace slapi_modify_internal_set_pb_ext with slapi_single_modify_internal_override ?

IIRC, the original code done a single internal modify using the full mods array. The issue was that if any individual mod failed, the entire operation failed. We switched to slapi_single_modify_internal_override so each mod is processed individually, allowing us to apply override logic per mod rather than bundling them together. With this change, we can safely ignore expected errors like “already exists” and “no such value”.

droideck: +1 to Thierry. I think the cleanest fix is to restore atomicity where it actually matters, not to catch OCV after the fact.

So, IIUC, splitting the batched modify into per-mod calls moved schema_check from "once on the final state" to "once per mod", which is exactly why a DEL on a MUST+SINGLE-VALUE attr fails with OCV before the ADD that would have fixed it runs.

Instead of retrying as REPLACE, we can group consecutive DEL+ADD on the same attribute and submit them as one atomic slapi_modify_internal_set_pb_ext call. schema_check will see the final state -> no OCV. Standalone mods still go through slapi_single_modify_internal_override, which preserves the idempotency tolerance that #6643 added (TYPE_OR_VALUE_EXISTS / NO_SUCH_ATTRIBUTE).

WDYT?

ldap/servers/plugins/referint/referint.c
```diff
@@ -713,21 +713,83 @@ _do_modify(Slapi_PBlock *mod_pb, Slapi_DN *entrySDN, LDAPMod **mods)
...
+
+            slapi_log_err(SLAPI_LOG_PLUGIN, REFERINT_PLUGIN_SUBSYSTEM,
+                          "_do_modify - DELETE of required attr \"%s\" on \"%s\" "
+                          "failed (%d), deleting dependent entry\n",
+                          mods[i]->mod_type, slapi_sdn_get_dn(entrySDN), rc);
+
+            slapi_delete_internal_set_pb(del_pb, slapi_sdn_get_dn(entrySDN),
+                                         NULL, NULL, referint_plugin_identity,
+                                         op_flags);
+            slapi_delete_internal_pb(del_pb);
```
droideck, line 781: Do I understand correctly that we "silently" delete the entry? It feels wrong for referint who usually just cleans up references (remove the DN from a member list) and didn't delete whole entries...

ldap/servers/plugins/referint/referint.c
jchapma: You're spot on,  the big issue is schema check timing, moving from "final state" to "per-mod" validation broke legit DEL+ADD sequences on MUST attributes. Your suggestion of grouping consecutive DEL+ADD on the same attribute for atomic processing solves this issue, while preserving the idempotency. Let me get back with a fix.

jchapma:
So i have a patch for the original issue that separated a multiple mod into individual mods, but which breaks the schema check  when the attr value is MUST+SINGLE-VALUE (https://github.com/389ds/389-ds-base/issues/6641). @aadhikar are you ok with merging my fix to the inital commit with this PR ?

aadhikar:
> So i have a patch for the original issue that separated a multiple mod into individual mods, but which breaks the schema check when the attr value is MUST+SINGLE-VALUE (#6641). @aadhikar are you ok with merging my fix to the inital commit with this PR ?

Thanks @jchapma! I have implemented the atomic DEL+ADD grouping per @droideck’s review, and all tests are passing. Could you please take a look and let me know if there is anything I might have missed or anything you would like me to add? Happy to incorporate your feedback 😊

jchapma:
> Thanks @jchapma! I have implemented the atomic DEL+ADD grouping per @droideck’s review, and all tests are passing. Could you please take a look and let me know if there is anything I might have missed or anything you would like me to add? Happy to incorporate your feedback 😊

Perfect, sure I can review.

ldap/servers/plugins/referint/referint.c
```diff
@@ -712,25 +712,61 @@ static int
...
+         * as a single atomic modify so that schema_check sees the final
+         * state (value replaced) rather than the intermediate state
+         * (required attribute removed).  This avoids OBJECT_CLASS_VIOLATION
+         * on MUST / SINGLE-VALUE attributes during MODRDN.
+         *
+         * Standalone mods go through slapi_single_modify_internal_override
+         * which tolerates TYPE_OR_VALUE_EXISTS and NO_SUCH_ATTRIBUTE,
+         * preserving the idempotency behaviour for replicated operations.
+         */
+        if ((mods[i]->mod_op & LDAP_MOD_OP) == LDAP_MOD_DELETE &&
```
jchapma: IMO. It would be cleaner to contain the logic of detecting a consecutive DEL + ADD on the same attr to a separate function. Then we could use it again in the future if needed and makes the code a little cleaner. Its not a major thing, just a nice to have.

ldap/servers/plugins/referint/referint.c
```diff
@@ -708,29 +708,73 @@ isFatalSearchError(int search_result)
...
+/*
+ * Check whether mods[i] and mods[i+1] form a consecutive DEL+ADD pair
+ * on the same attribute.  Such pairs must be submitted as a single
+ * atomic modify so that schema_check sees the final state (value
+ * replaced) rather than the intermediate state (required attribute
+ * removed).  This avoids OBJECT_CLASS_VIOLATION on MUST / SINGLE-VALUE
+ * attributes during MODRDN.
+ */
+static int
+_is_del_add_pair(LDAPMod **mods, size_t i)
```
jchapma, line 720: IIUC _do_modify() can be called with a single mod. If mods[i] == NULL then we will crash, maybe and a check before we access mods, something like:
    if (mods == NULL || mods[i] == NULL || mods[i + 1] == NULL) {
        return 0;
    }

aadhikar:
@jchapma @tbordaz @droideck can you guys please look into this when free!

jchapma (approved):
Tests run fine, code changes look good, nice

droideck (approved):
LGTM!

tbordaz (approved):
LGTM as well

PR #7414, opened 2026-04-20
Issue 6753 - Port ticket 48194 test

mirielka:
Description:
Porting ticket 48194 test to dirsrvtests/tests/suites/tls/cipher_test.py

Relates: #6753
Author: Lenka Doudova
Reviewer: ???

## Summary by Sourcery

Port TLS cipher policy tests from the ticket-specific suite into the shared TLS cipher_test suite and remove the old ticket48194 test file.

New Features:
- Add parametrized TLS cipher policy tests validating weak/strong cipher handshake behavior under various nsSSL3Ciphers and allowWeakCipher configurations.
- Introduce helper and fixture logic to drive openssl-based TLS connections against the standalone instance for cipher validation.

Tests:
- Add multiple TLS cipher policy integration tests covering default, custom, and invalid cipher suite configurations, including failure cases when all ciphers are disabled or unsupported ciphers are requested.
- Remove the legacy ticket-specific test file ticket48194_test.py now that its coverage is provided in the tls cipher_test suite.


mirielka:
This needs to be reviewed if the tests are even valid, also there are some failed assertions on handshake and I am not certain whether it's desired change in behavior or a bug.

dirsrvtests/tests/suites/tls/cipher_test.py
```diff
@@ -53,6 +67,329 @@ def test_long_cipher_list(topo):
...
+    """
+    log.info(f'Testing {cipher} -- expect to handshake {"successfully" if expect else "failed"}')
+
+    myurl = f'localhost:{LDAPSPORT}'
+    cmdline = ['/usr/bin/openssl', 's_client', '-connect', myurl, '-cipher', cipher]
+
+    strcmdline = " ".join(cmdline)
+    log.info(f"Running cmdline: {strcmdline}")
+
+    try:
```
progier389: You should simplify this by using context manager and plain assert. Something like:
```
with subprocess.Popen(cmdline, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.STDOUT) as proc:
    for line in proc.stdout:
        if line == b"":
            break
        if b'Cipher is' in line:
            log.info(f"Found: {line}")
            if expect:
                assert b'(NONE)' not in line
            else:
                assert b'(NONE)' in line
```

jchapma:
Some tests are failing:
FAILED dirsrvtests/tests/suites/tls/cipher_test.py::test_cipher_policy_2[AES256-SHA256-False] - assert False
FAILED dirsrvtests/tests/suites/tls/cipher_test.py::test_cipher_policy_6[AES256-SHA256-False] - assert False

It seems to me that server cipher restriction is one issue, topo.standalone.encryption.change_ciphers('+', ['rsa_aes_128_sha','rsa_aes_256_sha']) adds these to the existing ciphers but the server still accepts AES256-SHA256

It seems some ciphers are not accepted by the server:
WARNING:Encryption:Cipher rsa_aes_128_sha is not supported.
WARNING:Encryption:Cipher rsa_aes_256_sha is not supported.

mirielka:
Test was redone after consult since some functionalities are no longer supported with TLS1.3. The test currently mainly verifies that setting attributes works as expected.

jchapma (approved):
Tests all pass, LGTM

Issue #7088, opened 2025-11-10
clcache_load_buffer - Can't locate CSN 691226b2000400300000 in the changelog


arjones852:
**Issue Description**
[10/Nov/2025:11:46:20.032740958 -0600] - ERR - agmt="cn=meTofreeipa1.us.ep.corp.local" (freeipa1:389) - clcache_load_buffer - Can't locate CSN 6912262c000100300000 in the changelog (DB rc=-12797). If replication stops, the consumer may need to be reinitialized.
[10/Nov/2025:11:48:12.022925249 -0600] - ERR - agmt="cn=meTofreeipa1.us.ep.corp.local" (freeipa1:389) - clcache_load_buffer - Can't locate CSN 6912269c000000300000 in the changelog (DB rc=-12797). If replication stops, the consumer may need to be reinitialized.
[10/Nov/2025:11:48:34.126837044 -0600] - ERR - agmt="cn=meTofreeipa1.us.ep.corp.local" (freeipa1:389) - clcache_load_buffer - Can't locate CSN 691226b2000400300000 in the changelog (DB rc=-12797). If replication stops, the consumer may need to be reinitialized.

**Package Version and Platform:**
 - Platform: Rocky 10
 - Package and version: 389-ds-base-3.0.6-12.el10_0.x86_64

**Steps to Reproduce**
After upgrading a 4 node replicated FreeIPA cluster (demote/remove/replace replicas) from Rocky 8 to Rocky 10, I am now getting random replication errors like above.

I have gone as far as removing all replicas except for 1 and then adding them back. After adding just 1 back, errors began again. I have been tearing my hair out trying to figure out why and then came across this mailing list post which describes my issue perfectly: https://lists.fedorahosted.org/archives/list/389-users@lists.fedoraproject.org/thread/Z6LSV2RRCNID7VN3XUCS2JMT7WLJXNH7/

**Expected results**
Replication works

I am open to assist however I can with resolving this issue.

arjones852:
Thanks!

Typically the errors were showing about every 5 to 10 minutes on all 4 servers, but ever since enabling debug on them I am unable to find any obvious error messages, and the string "Can't locate CSN" doesn't show up anymore at all. I've been running debug for about half an hour now.

What would an error message look like with debug enabled so that I know what I am looking for?

arjones852:
This is the contents of the CSN it's complaining about:

```
changetype: modify
replgen: 5e29f3d2000000040000
[...]

PR #7416, opened 2026-04-20
Issue 7088 - Change log level for "Can't locate CSN" error message

mreynolds389:
Description:

There are legitimate reasons for a CSN temporarily not found in the changelog, and in recent cases it always resolves itself.  Instead of logging a scary message by default move the log level to "replication".

relates: https://github.com/389ds/389-ds-base/issues/7088

## Summary by Sourcery

Enhancements:
- Downgrade the log level of the "Can't locate CSN in the changelog" message from error to replication-specific logging.

### DIFF
--- a/ldap/servers/plugins/replication/cl5_clcache.c
+++ b/ldap/servers/plugins/replication/cl5_clcache.c
@@ -242,7 +242,7 @@ clcache_get_buffer(Replica *replica, CLC_Buffer **buf, dbi_db_t *db, ReplicaId c
             if (0 == clcache_enqueue_busy_list(replica, db, *buf)) {
                 Slapi_Backend *be = (*buf)->buf_busy_list->bl_be;
                 /* buf_busy_list is now set, and we can get the backend. So lets initialize the dbimpl buffers */
-                
+
                 dblayer_bulk_set_buffer(be, &(*buf)->buf_bulk, &(*buf)->buf_bulkdata,
                         WORK_CLC_BUFFER_PAGE_SIZE, DBI_VF_BULK_RECORD);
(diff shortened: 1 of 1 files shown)


tbordaz (commented):
This error being most of the time transient it looks good to change the log level.

If the failure occurs  several times for the same CSN do you thinnk it worth logging an alarming message, that could also help for diagnose.
What about calling clcache_load_buffer with a cookie (nb try, csn) so that if it fails too many time for the same CSN we can log an error. The csn is provided by clcache_load_buffer upon failure.

mreynolds389:
> This error being most of the time transient it looks good to change the log level.
> 
> If the failure occurs several times for the same CSN do you think it worth logging an alarming message, that could also help for diagnose. What about calling clcache_load_buffer with a cookie (nb try, csn) so that if it fails too many time for the same CSN we can log an error. The csn is provided by clcache_load_buffer upon failure.

Hmm I don't think there is a loop to properly handle this approach.  The initial loop in clcache_get_next_change() exits as soon as clcache_load_buffer() returns an error.  I'm not sure if higher up in the code path there is another loop where I can track these retries.  It loops in send_updates() but I don't recall if it gets fixed in the same or separate replication session.  If it's not fixed in a same session then we are still going to be incorrectly logging the error.  I've asked the original reporter to try and get a fresh debug log so we can see when it recovers and that should hopefully tell us if your idea is possible.

progier389 (approved):
LGTM

tbordaz (approved):
You are correct, we can only detect we are iterating on the same starting csn at the beginning of the session. Here it is too deep in CL code.
LGTM

Issue #7423, opened 2026-04-21
Cleanup pblock after freeing post/pre entries


mreynolds389:
**Issue Description**

There are a few places where we don't set the pre/post entry to NULL in the pblock after freeing these entries.  Best to be safe and make sure they are set to NULL when freeing them.

mreynolds389:
0c319c9f2..64add750a  389-ds-base-3.1 -> 389-ds-base-3.1
589bcdeea..883819a49  389-ds-base-3.0 -> 389-ds-base-3.0
0c5b917a8..956023037  389-ds-base-2.8 -> 389-ds-base-2.8
d84b2ac5d..57b1df8ea  389-ds-base-2.7 -> 389-ds-base-2.7
62d069e20..aec91a304  389-ds-base-2.6 -> 389-ds-base-2.6
8678c5dfc..39e8ef725  389-ds-base-2.4 -> 389-ds-base-2.4
6187848c1..de69bd48f  389-ds-base-2.2 -> 389-ds-base-2.2
47de56a1e..cbc6a1bfa  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7424, opened 2026-04-21
Issue 7423 - cleanup pblock after freeing pre/post entries

mreynolds389:
Description:

There are few places where we don't see the pre/post entry in the pblock to NULL after freeing them. Just being overly cautious as some of these changes might not be needed, but it's still good to set the code example and also future-proof these areas.

relates: https://github.com/389ds/389-ds-base/issues/7423

## Summary by Sourcery

Ensure operation pblock entries are cleared after their associated resources are freed to avoid stale pointers.

Bug Fixes:
- Clear pre/post operation entries in the pblock after freeing them in modrdn and add operations to prevent use of stale entry pointers.
- Reset modify operation mods in the pblock after freeing them to avoid dangling references.
- Null out pre-op entry pointers in the generation helper after freeing the associated entries.

### DIFF
--- a/ldap/servers/slapd/add.c
+++ b/ldap/servers/slapd/add.c
@@ -848,6 +848,7 @@ done:
     if (be)
         slapi_be_Unlock(be);
     slapi_pblock_get(pb, SLAPI_ENTRY_POST_OP, &pse);
+    slapi_pblock_set(pb, SLAPI_ENTRY_POST_OP, NULL);
     slapi_entry_free(pse);
     slapi_ch_free((void **)&operation->o_params.p.p_add.parentuniqueid);
     slapi_entry_free(e);
--- a/ldap/servers/slapd/generation.c
+++ b/ldap/servers/slapd/generation.c
@@ -78,6 +78,7 @@ set_database_dataversion(const char *dn, const char *dataversion)
     if (NULL != pb) {
         Slapi_Entry *e;
         slapi_pblock_get(pb, SLAPI_ENTRY_PRE_OP, &e);
+        slapi_pblock_set(pb, SLAPI_ENTRY_PRE_OP, NULL);
         slapi_entry_free(e);
     }
     slapi_pblock_destroy(pb);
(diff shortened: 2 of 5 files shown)


vashirov:
LGTM!
I think we should also clear mods in `do_modify()`
```diff
diff --git a/ldap/servers/slapd/modify.c b/ldap/servers/slapd/modify.c
index 0ecce9bc8..29edc515e 100644
--- a/ldap/servers/slapd/modify.c
+++ b/ldap/servers/slapd/modify.c
@@ -378,6 +378,7 @@ do_modify(Slapi_PBlock *pb)

     slapi_pblock_get(pb, SLAPI_MODIFY_MODS, &normalized_mods);
     ldap_mods_free(normalized_mods, 1 /* Free the Array and the Elements */);
+    slapi_pblock_set(pb, SLAPI_MODIFY_MODS, NULL);

 free_and_return:;
     slapi_ch_free_string(&old_pw);
```

and `_mtn_update_config_param()`
```diff
diff --git a/ldap/servers/slapd/mapping_tree.c b/ldap/servers/slapd/mapping_tree.c
index 7c276f06b..c052bd4da 100644
--- a/ldap/servers/slapd/mapping_tree.c
+++ b/ldap/servers/slapd/mapping_tree.c
@@ -3842,6 +3842,7 @@ _mtn_update_config_param(int op, char *type, char *strvalue)
      * since the internal modify could realloced mods. */
     slapi_pblock_get(confpb, SLAPI_MODIFY_MODS, &mods);
     ldap_mods_free(mods, 1 /* Free the Array and the Elements */);
+    slapi_pblock_set(confpb, SLAPI_MODIFY_MODS, NULL);
     slapi_pblock_destroy(confpb);

     return rc;
```

mreynolds389:
> LGTM! I think we should also clear mods in `do_modify()`
> 

Yeah when I cleared the first mods I was wondering if I should comb through the code for other "mods" but I was worried it might turn into a can of worms.  So yes I'll definitely add this change.

vashirov (approved):

PR #7427, opened 2026-04-22
Issue 7426 - logconv.py is out of sync with server-emitted note codes

### LINKED ISSUE #7426: logconv.py is out of sync with server-emitted note codes
**Description**
notes=N (asynchronous) and notes=B (blocked) are not reported anywhere.
notes=M (MFA) entry doesn't function correctly -- logconv silently fails to produce results, with no useful error to point at the cause.

droideck:
Description: notes=N (asynchronous) and notes=B (blocked) were not tracked or reported, and notes=M (MFA) silently failed to produce any per-operation output because the class-level dict shadowed the ResultData field.

Add counters and detail dicts for notes=N and notes=B in ResultData, remove the shadowing assignments in logAnalyser, and log a debug message (instead of crashing) when an unrecognized note code is encountered so future codes do not break analysis.

Decrease redundancy with a shared helper.
Add a test that injects SRCH/RESULT pairs for every known note code plus an unknown one and asserts both summary counters and verbose detail blocks.

Fixes: https://github.com/389ds/389-ds-base/issues/7426

Reviewed by: ?

## Summary by Sourcery

Track and report all server-emitted note codes in logconv, including new asynchronous and blocked operation codes, while making verbose note reporting more robust and less error-prone.

New Features:
- Expose counters and verbose detail sections for asynchronous (N) and blocked (B) operations in logconv output.

Bug Fixes:
- Ensure multi-factor authentication (M) notes are correctly recorded and reported instead of being shadowed and silently ignored.
- Handle unknown note codes gracefully by logging a debug message instead of crashing during log analysis.

Enhancements:
- Refactor per-note verbose reporting into a shared helper to reduce duplication and standardize output formatting.
- Extend ResultData structures to maintain per-operation details for all supported note codes, including M, N, and B.

Tests:
- Add a tier-1 test that injects log entries for each known note code plus an unknown one to verify summary counters and verbose detail blocks in both legacy and JSON log formats.

### DIFF
--- a/dirsrvtests/tests/suites/logging/logconv_test.py
+++ b/dirsrvtests/tests/suites/logging/logconv_test.py
@@ -5,6 +5,7 @@
 # License: GPL (version 3 or any later version).
 # See LICENSE for details.
 # --- END COPYRIGHT BLOCK ---
+import json
 import ldap
 import logging
 import os
@@ -29,6 +30,8 @@ from lib389._mapped_object import DSLdapObjects
 from lib389.config import CertmapLegacy
 from lib389.nss_ssl import NssSsl
 from lib389.topologies import create_topology
+
+pytestmark = pytest.mark.tier1
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


progier389 (approved):
LGTM

PR #7428, opened 2026-04-22
Issue 6753 - Removing ticket49412_test and porting to DSLdapObject

aadhikar:
Description:
The old ticket49412_test.py had compatibility issues.
This ports the functionality to changelog_test.py
using modern Changelog5 class from lib389.

Relates: #6753

Reviewed by: ???
Assisted by: Claude

## Summary by Sourcery

Migrate legacy changelog type validation coverage into the replication changelog test suite and remove the obsolete ticket-specific test file.

Tests:
- Add a replication changelog test that validates string-based maxage and trim interval configuration using the Changelog5 helper.
- Remove the deprecated ticket49412_test.py now that its coverage is provided in the changelog test suite.


dirsrvtests/tests/suites/replication/changelog_test.py
```diff
@@ -786,6 +786,49 @@ def test_changelog_pagesize(topo):
...
+    :expectedresults:
+        1. Changelog object retrieved successfully
+        2. Valid string values should be accepted
+        3. Configuration should be updated with correct values
+    """
+
+    supplier = topo.ms["supplier1"]
+
+    log.info('Creating Changelog5 object for supplier1')
+    if ds_supports_new_changelog():
```
bsimonova, line 819: Since it's just a skip, the changelog check could be added to pytestmark.skipif before the test

aadhikar:
Note: changelog_test.py is not part of the CI test matrix. test_changelog_type_validation was verified locally and passes on both LMDB and BDB backends.

dirsrvtests/tests/suites/replication/changelog_test.py
```diff
@@ -786,6 +798,50 @@ def test_changelog_pagesize(topo):
...
+
+    log.info('Reading current changelog max age configuration')
+    current_maxage = cl.get_attr_val_utf8(MAXAGE)
+    log.info(f'Current {MAXAGE}: {current_maxage}')
+
+    log.info('Setting changelog maxage to 60s')
+    cl.replace(MAXAGE, '60s')
+
+    log.info('Setting changelog trim interval to 10s')
+    cl.replace(TRIMINTERVAL, '10s')
```
mreynolds389, line 832: I don't think the time unit "s" is supported.  The trim interval is always in seconds.  So just "10" is sufficient.

aadhikar: @mreynolds389 
I believe 's' is supported, see is_valid_duration_unit(): https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/time.c#L716-L734

But happy to change it to just '10' if you would prefer keeping it without the unit.

mreynolds389: Indeed it's allowed, ok leave it as it.

mreynolds389 (approved):

Issue #7431, opened 2026-04-22
password policy - passwordBadWords not working for local polciies


mreynolds389:
**Issue Description**

When processing local password policies we only initialize the  "bad words" setting from the global config.  So the local policy setting for passwordBadWords is always skipped/ignored.

mreynolds389:
4817a02e3..c366230e8  389-ds-base-3.1 -> 389-ds-base-3.1
892a3f387..07d67c92d  389-ds-base-3.0 -> 389-ds-base-3.0
ec81f9a48..5bc5cbc92  389-ds-base-2.8 -> 389-ds-base-2.8
4094934d6..dcab68d74  389-ds-base-2.7 -> 389-ds-base-2.7
18589efc9..a48ad7e98  389-ds-base-2.6 -> 389-ds-base-2.6
6d42ed176..981f38a7e  389-ds-base-2.4 -> 389-ds-base-2.4
a561bca21..919efac76  389-ds-base-2.2 -> 389-ds-base-2.2

PR #7432, opened 2026-04-22
Issue 7431 - password policy - passwordBadWords is ignored in local p…

mreynolds389:
…olicies

Description:

When processing local password policies we only initialize the "bad words" setting from the global config. So the local policy setting for passwordBadWords is always skipped/ignored.

In fact we do initialize the password policy config, but when we do the actual syntax check we directly pull from the global config via: config_get_pw_bad_words_array()

relates: https://github.com/389ds/389-ds-base/issues/7431

CI test assisted by: Cursor

## Summary by Sourcery

Ensure password bad-words checks honor subtree/local password policies instead of always using the global configuration.

Bug Fixes:
- Fix password syntax checking to use the effective password policy's bad-words list rather than always reading from the global configuration.

Tests:
- Extend password policy syntax tests to cover global vs subtree passwordBadWords behavior and validation for comma- and space-separated lists.

### DIFF
--- a/ldap/servers/slapd/pw.c
+++ b/ldap/servers/slapd/pw.c
@@ -1,6 +1,6 @@
 /** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
- * Copyright (C) 2021 Red Hat, Inc.
+ * Copyright (C) 2026 Red Hat, Inc.
  * Copyright (C) 2009 Hewlett-Packard Development Company, L.P.
  * All rights reserved.
  *
@@ -1066,7 +1066,7 @@ check_pw_syntax_ext(Slapi_PBlock *pb, const Slapi_DN *sdn, Slapi_Value **vals, c
 {
     Slapi_Attr *attr;
     Slapi_Value **va = NULL;
-    int i, pwresponse_req = 0;
+    int pwresponse_req = 0;
     int is_replication = 0;
     int internal_op = 0;
     char *dn = (char *)slapi_sdn_get_ndn(sdn); /* jcm - Had to cast away const */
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/pw.c
```diff
@@ -1185,16 +1184,13 @@ check_pw_syntax_ext(Slapi_PBlock *pb, const Slapi_DN *sdn, Slapi_Value **vals, c
             }
 
             /* Check for bad words */
-            bad_words_array = config_get_pw_bad_words_array();
-            if (bad_words_array) {
-                for (size_t b = 0; bad_words_array && bad_words_array[b]; b++) {
-                    if (strcasestr(pwd, bad_words_array[b])) {
+            if (pwpolicy->pw_bad_words_array) {
```
progier389, line 1187: Should we not still use config_get_pw_bad_words_array() if pwpolicy->pw_bad_words_array == NULL ?

mreynolds389: > Should we not still use config_get_pw_bad_words_array() if pwpolicy->pw_bad_words_array == NULL ?

We would only use that if we are inheriting the global policy and the local policy has syntax checking disabled  - this is done in new_passwdPolicy().  However, we are not "inheriting" this setting (and possibly others).  I will update the PR for this.

progier389: Ok. after seeing your explanation, I dont think that we should "inherit" the global policy if the local policy has syntax checking disabled 
IMHO the local policy is supposed to replace the global one (to allow having some specific entries with syntax checking disabled while most entries have it enabled)
So LGTM

progier389 (approved):
LGTM

ldap/servers/slapd/pw.c
mreynolds389: @progier389 Actually there is a setting for inheriting the global policy and that is where I added the missing syntax settings.  By default we don't inherit the global policy and that was not changed in this PR.

So I made more changes that need review.

mreynolds389: @progier I made more changes after your approval.  Can you quick look over it again?

progier389 (approved):
Still looks good.
LGTM

mreynolds389:
> Still looks good. LGTM

Thanks!  I didn't want to blindly just push an unreviewed change :-)

PR #7434, opened 2026-04-23
Issue 6753 - Port ticket 48270 test

mirielka:
Description:
Port ticket 48270 test into dirsrvtests/tests/suites/indexes/regression_test.py

Relates: #6753
Author: Lenka Doudova
Assisted by: Cursor
Reviewer: ???

## Summary by Sourcery

Port the homeDirectory index regression coverage for ticket 48270 into the shared indexes regression test suite and remove the standalone ticket-specific test file.

Tests:
- Add a reusable homeDirectory_setup fixture and matching rule helper for homeDirectory index tests.
- Introduce regression tests validating reindexing of homeDirectory with dual IA5 matching rules and ensuring no invalid matching rule errors are logged.
- Add a regression test verifying equality and extensible homeDirectory searches with mixed-case values behave correctly under dual IA5 matching rules.
- Remove the legacy ticket48270_test.py file now that its scenarios are covered in the regression suite.


dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -1014,6 +1014,164 @@ def test_concurrent_modifications_during_indexing(topo, homeDirectory_index_clea
...
+        1. Reindex task completes with return code 0; no such error line appears
+        2. Assertion holds (no match found)
+    """
+
+    user_list, index = homeDirectory_setup
+
+    setup_matching_rule(topo, index)
+
+    log.info("Scan error log for bad matching rule")
+    regex = re.compile("unknown or invalid matching rule")
```
progier389: I think we should rather directly use match(regex) function of DirsrvErrorLog class from lib389
(Eventually fixing it if is broken) rather than rewriting it

mirielka: Thanks for the suggestion, fixed!

progier389 (approved):
LGTM

PR #7461, opened 2026-04-29
Issue 7460 - MOD_REPLACE on groups/link attributes modifies overlap targets

### LINKED ISSUE #7460: MOD_REPLACE on groups/link attributes modifies overlap targets
**Description**
When a group's `member` list is changed with `MOD_REPLACE`, members that are in both the old and the new list still get modified — their `entryUSN` advances even though they were members before and remain members after. Only members actually added or removed should be touched. The Linked Attributes plugin has the same behaviour for `linkType` MOD_REPLACE on overlap targets.

droideck:
Description: memberof_replace_list and linked_attrs_replace_backpointers sort pre/post with slapi_attr_value_cmp_ext, which returns 0 on match and a negative value or LDAP error code on non-match — not a strict weak ordering. The merge loop's cmp > 0 branch is never taken, so overlap members are deleted and re-added on every MOD_REPLACE, bumping entryUSN.

Switch both comparators to slapi_utf8casecmp on raw bv_val. For that to work, normalize DN-syntax values in str2entry_fast (which did not normalize at all — the old code sat under an unset OBSOLETE_DN_SYNTAX_CHECK) and route str2entry_dupcheck through the same helper. Skip the upgradedn path (SLAPI_STR2ENTRY_USE_OBSOLETE_DNFORMAT) in both, preserving raw bytes for the LMDB import. Fix a stray missing comma in the dupcheck strict-fail slapi_log_err. Drop the now-unused qsortConfig plugin global.

Add Python regression tests for both memberOf and linkedAttrs cases.

Fixes: https://github.com/389ds/389-ds-base/issues/7460

Reviewed by: ?

## Summary by Sourcery

Ensure DN-valued attributes are normalized consistently so memberOf and linked attributes handle MOD_REPLACE diffs correctly without unnecessarily re-touching overlapping members.

Bug Fixes:
- Fix memberOf group MOD_REPLACE so overlapping members are not removed and re-added, avoiding unnecessary entryUSN bumps.
- Fix linked attributes MOD_REPLACE handling so overlap targets are not spuriously re-modified.
- Correct DN syntax error logging in str2entry_dupcheck to avoid memory leaks when strict DN checking fails.

Enhancements:
- Normalize DN-syntax attribute values during entry parsing and duplicate-checking while preserving raw bytes for legacy upgradedn imports.
- Align memberOf and linked attributes comparator logic to use a strict weak ordering based on UTF-8 case-insensitive comparison of normalized DNs.
- Remove the now-unnecessary memberOf plugin global used to pass config into qsort comparators.

Tests:
- Add regression tests exercising memberOf behavior across overlapping, empty, denormalized, and no-op group MOD_REPLACE scenarios with entryUSN checks.
- Add regression tests verifying DN normalization of imported denormalized member values and subsequent memberOf fixup and MOD_REPLACE behavior.
- Add a linked attributes regression test ensuring MOD_REPLACE does not spuriously update overlap targets’ entryUSN values.

### DIFF
--- a/dirsrvtests/tests/suites/memberof_plugin/regression_test.py
+++ b/dirsrvtests/tests/suites/memberof_plugin/regression_test.py
@@ -18,7 +18,7 @@ from random import sample
 from lib389.utils import ds_is_older, ensure_list_bytes, ensure_bytes, ensure_str
 from test389.topologies import topology_m1h1c1 as topo, topology_st, topology_m2 as topo_m2
 from lib389._constants import *
-from lib389.plugins import MemberOfPlugin, RetroChangelogPlugin
+from lib389.plugins import MemberOfPlugin, RetroChangelogPlugin, USNPlugin
 from lib389.backend import Backends
 from lib389.mappingTree import MappingTrees
 from lib389 import Entry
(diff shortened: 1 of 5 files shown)


progier389 (approved):
LGTM

PR #7463, opened 2026-04-29
Issue 7462 - Fix SERIAL LOCK and entry cache lock ordering on BDB dea…

### LINKED ISSUE #7462: Freeze under heavy write load with BDB backend
**Issue Description**
In IPA context, server can freeze under heavy write load (add/mod/del users and hosts, deferred memberof enabled).

I found 2 issues:
1. In deadlock retry path, a thread grabs SERIAL LOCK, during deadlock sleeps still holding the SERIAL LOCK. Every other writer thread waits for that sleep to finish.
2. After fixing the first issue, I encountered lock ordering issues with SERIAL LOCK and entry locks. On retry Thread A holds entry lock from the previous attempt, tries to grab SERIAL LOCK. Thread B holds SERIAL LOCK, wants the same entry lock. Neither can proceed, resulting in a deadlock.

vashirov:
…dlock retry

Bug Description:
In IPA context, server can freeze under heavy write load (add/mod/del users and hosts, deferred memberof enabled).

I found 2 issues:
1. In deadlock retry path, a thread grabs SERIAL LOCK, during deadlock sleeps still holding the SERIAL LOCK. Every other writer thread waits for that sleep to finish.
2. After fixing the first issue, I encountered lock ordering issues with SERIAL LOCK and entry locks. On retry Thread A holds entry lock from the previous attempt, tries to grab SERIAL LOCK. Thread B holds SERIAL LOCK, wants the same entry lock. Neither can proceed, resulting in a deadlock.

Fix Description:
1. Release SERIAL LOCK before backoff sleep on BDB deadlock retry and cache lock retry paths.
2. Fix AB-BA deadlock between SERIAL LOCK and entry cache locks by releasing/re-acquiring entry locks around retry.

Fixes: https://github.com/389ds/389-ds-base/issues/7462

## Summary by Sourcery

Adjust BDB transaction retry handling to avoid deadlocks involving the global SERIAL LOCK and entry cache locks during high-write workloads.

Bug Fixes:
- Release the SERIAL LOCK before backoff sleeps on BDB deadlock and cache lock retry paths so other writers are not blocked behind a sleeping thread.
- Resolve AB-BA deadlocks between the SERIAL LOCK and entry cache locks by unlocking entries before aborting on retry and re-locking them after re-establishing the transaction.
- Prevent double-unlock and inconsistent cache state on error paths by tracking whether entry and parent cache locks are held before unlocking or returning entries to the cache.

Enhancements:
- Standardize transaction retry behavior in add, delete, modify, and modrdn operations to always acquire the SERIAL LOCK consistently before (re)locking entry cache objects.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_add.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_add.c
@@ -100,6 +100,7 @@ ldbm_back_add(Slapi_PBlock *pb)
     int result_sent = 0;
     int32_t parent_op = 0;
     int32_t betxn_callback_fails = 0; /* if a BETXN fails we need to revert entry cache */
+    int32_t parent_locked = 0;        /* non-zero when parent entry cache lock is held */
     struct timespec parent_time;
 
     if (slapi_pblock_get(pb, SLAPI_CONN_ID, &conn_id) < 0) {
@@ -219,8 +220,19 @@ ldbm_back_add(Slapi_PBlock *pb)
     txn.back_txn_txn = NULL; /* ready to create the child transaction */
     for (retry_count = 0; retry_count < RETRY_TIMES; retry_count++) {
         if (txn.back_txn_txn && (txn.back_txn_txn != parent_txn)) {
-            /* Don't release SERIAL LOCK */
-            dblayer_txn_abort_ext(li, &txn, PR_FALSE);
+            /*
+             * Unlock entry cache locks before releasing SERIAL LOCK
+             * to maintain lock ordering (SERIAL LOCK -> entry locks)
+             * and prevent AB-BA deadlock on retry.
+             * Entries stay refcounted in cache, so pointers remain valid.
+             */
+            if (parent_locked && parent_modify_c.old_entry) {
+                cache_unlock_entry(&inst->inst_cache, parent_modify_c.old_entry);
+                parent_locked = 0;
+            }
+
+            /* Release SERIAL LOCK */
+            dblayer_txn_abort(be, &txn);
             noabort = 1;
             slapi_pblock_set(pb, SLAPI_TXN, parent_txn);
             /* must duplicate addingentry before returning it to cache,
@@ -267,11 +279,19 @@ ldbm_back_add(Slapi_PBlock *pb)
         }
         /* dblayer_txn_begin holds SERIAL lock,
          * which should be outside of locking the entry (find_entry2modify) */
-        if (0 == retry_count) {
-            /* First time, hold SERIAL LOCK */
-            retval = dblayer_txn_begin(be, parent_txn, &txn);
-            noabort = 0;
+        retval = dblayer_txn_begin(be, parent_txn, &txn);
+        if (0 != retval) {
+            if (LDBM_OS_ERR_IS_DISKFULL(retval)) {
+                disk_full = 1;
+                ldap_result_code = LDAP_OPERATIONS_ERROR;
+                goto diskfull_return;
+            }
+            ldap_result_code = LDAP_OPERATIONS_ERROR;
+            goto error_return;
+        }
+        noabort = 0;
 
+        if (0 == retry_count) {
             if (!is_tombstone_operation) {
                 rc = slapi_setbit_int(rc, SLAPI_RTN_BIT_FETCH_EXISTING_DN_ENTRY);
             }
@@ -391,23 +411,26 @@ ldbm_back_add(Slapi_PBlock *pb)
                  */
                 /* JCMREPL - Warning: A Plugin could cause an infinite loop by always returning a result code that requires some action. */
             }
-        } else {
-            /* Otherwise, no SERIAL LOCK */
-            retval = dblayer_txn_begin_ext(li, parent_txn, &txn, PR_FALSE);
         }
-        if (0 != retval) {
-            if (LDBM_OS_ERR_IS_DISKFULL(retval)) {
-                disk_full = 1;
-                ldap_result_code = LDAP_OPERATIONS_ERROR;
-                goto diskfull_return;
-            }
-            ldap_result_code = LDAP_OPERATIONS_ERROR;
-            goto error_return;
-        }
-        noabort = 0;
 
         /* stash the transaction for plugins */
         slapi_pblock_set(pb, SLAPI_TXN, txn.back_txn_txn);
+
+        /* Re-lock parent after SERIAL LOCK to maintain lock ordering */
+        if (retry_count > 0 && parent_found && parent_modify_c.old_entry) {
+            int lock_rc = cache_lock_entry(&inst->inst_cache,
+                                           parent_modify_c.old_entry);
+            if (lock_rc != 0) {
+                slapi_log_err(SLAPI_LOG_ERR, "ldbm_back_add",
+                              "conn=%" PRIu64 " op=%d Failed to re-lock parent "
+                              "entry after deadlock retry (rc=%d)\n",
+                              conn_id, op_id, lock_rc);
+                ldap_result_code = LDAP_OPERATIONS_ERROR;
+                retval = -1;
+                goto error_return;
+            }
+            parent_locked = 1;
+        }
 
         if (0 == retry_count) { /* execute just once */
             /* Nothing in this if crause modifies persistent store.
@@ -439,6 +462,7 @@ ldbm_back_add(Slapi_PBlock *pb)
                     goto error_return;
                 }
                 modify_init(&parent_modify_c, parententry);
+                parent_locked = 1;
             }
 
             /* Check if the entry we have been asked to add already exists */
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/back-ldbm/ldbm_delete.c
```diff
@@ -82,6 +82,8 @@ int delete_tombstone_entry = 0; /* We must remove the given tombstone entry from
     Connection *pb_conn;
     int32_t parent_op = 0;
     int32_t betxn_callback_fails = 0; /* if a BETXN fails we need to revert entry cache */
+    int32_t e_locked = 0;             /* non-zero when entry 'e' cache lock is held */
```
progier389, line 85: Should better use bool and true/false

progier389 (approved):
Looks good except a minor point (use bool for the flags instead of int32_t)

ldap/servers/slapd/back-ldbm/ldbm_modify.c
```diff
@@ -620,15 +621,27 @@ ldbm_back_modify(Slapi_PBlock *pb)
             ldap_result_code = -1;
             goto error_return; /* error result sent by find_entry2modify() */
         }
+        e_locked = 1;
```
droideck, line 624: Do I understand correctly that for `MANAGE_ENTRY_BEFORE_DBLOCK` case on iter 1 we acquire E first and then SERIAL LOCK? But then on iter 2 -- it's another way around (SERIAL LOCK first and then E)? 

IIUC, even though very rarely (and especially because MANAGE_ENTRY_BEFORE_DBLOCK is not a default mode but also experimental mode) but we still might have the AB-BA deadlock here.

droideck (commented):
Overall, looks great!

Regarding testing, do you think it's possible to reproduce it via stress test?
Something that spawns N writer threads doing add/mod/del under deferred-memberof probably would exercise the retry path. But that's a stretch and it's not a blocker for the PR merge.

ldap/servers/slapd/back-ldbm/ldbm_add.c
```diff
@@ -219,8 +220,19 @@ ldbm_back_add(Slapi_PBlock *pb)
     txn.back_txn_txn = NULL; /* ready to create the child transaction */
     for (retry_count = 0; retry_count < RETRY_TIMES; retry_count++) {
         if (txn.back_txn_txn && (txn.back_txn_txn != parent_txn)) {
-            /* Don't release SERIAL LOCK */
-            dblayer_txn_abort_ext(li, &txn, PR_FALSE);
+            /*
```
tbordaz, line 223: I wonder if there is not a confusion between backend lock and db_env lock.
Initially the backend lock was held during all lmdb_add and db_env release during retry/sleep.
With the move of plugins in betxn (#351) it was changed from dblayer_txn_abort (release the db_env) to dblayer_txn_abort_ext(...FALSE) (to not release the db_env) but the comment was about 'Serial Lock' that in my mind means backend lock.
Of course we do not want to release the backend lock for a retry but it looks fine to me to release db_env.

What is not clear to me is the impact of releasing it. IIUC it allows others txn but on different backend. Is it the expected benefit ?

ldap/servers/slapd/back-ldbm/ldbm_modify.c
mreynolds389: Perhaps when (!MANAGE_ENTRY_BEFORE_DBLOCK) we should still lock the entry _if_ it is not locked already (near line 721).  I think that will fix the ordering when using that db opt level.

```
if  (!MANAGE_ENTRY_BEFORE_DBLOCK) {
   ...
} else if (e && !e_locked) {
    cache_lock_entry(&inst->inst_cache, e);
    e_locked = 1;
}
```

aadhikar:
Hey @vashirov, could you address the outstanding review comments and rebase? Thanks!

vashirov:
@aadhikar, it's ready to be merged, but waiting for slapi-nis PR to be merged https://codeberg.org/freeipa/slapi-nis/pulls/71 at the same time, otherwise this will break FreeIPA nightly CI

ldap/servers/slapd/back-ldbm/ldbm_add.c
```diff
@@ -391,24 +411,27 @@ ldbm_back_add(Slapi_PBlock *pb)
...
-            goto error_return;
-        }
-        noabort = 0;
 
         /* stash the transaction for plugins */
         slapi_pblock_set(pb, SLAPI_TXN, txn.back_txn_txn);
 
+        /* Re-lock parent after SERIAL LOCK to maintain lock ordering */
+        if (retry_count > 0 && parent_found && parent_modify_c.old_entry) {
+            int lock_rc = cache_lock_entry(&inst->inst_cache,
```
droideck, line 421: IIUC, if a concurrent modify touches the parent during our backoff, modify_switch_entries() goes through cache_replace(), which marks the old backentry ENTRY_STATE_DELETED. Our pointer still references that old backentry, so cache_lock_entry() here returns RETRY_CACHE_LOCK and we turn it into LDAP_OPERATIONS_ERROR.
Previously, the concurrent writer just blocked on the lock and both operations succeeded; now the retrying add fails with err=1. And since a deadlock retry implies a concurrent writer, this fires exactly on the collision path the patch is smoothing, unless I miss something.
Same shape in modify, delete, and modrdn.

Maybe, on a RETRY_CACHE_LOCK we can refetch the entry by DN and restart the attempt?
And fail only if the entry is genuinely gone (RETRY_CACHE_LOCK also covers real deletion, so the refetch distinguishes the two for free).

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -233,8 +236,27 @@ ldbm_back_modrdn(Slapi_PBlock *pb)
...
-            /* don't release SERIAL LOCK */
-            dblayer_txn_abort_ext(li, &txn, PR_FALSE);
+            /*
+             * Unlock entry cache locks before releasing SERIAL LOCK
+             * to maintain lock ordering (SERIAL LOCK -> entry locks)
+             * and prevent AB-BA deadlock on retry.
+             * Entries stay refcounted in cache, so pointers remain valid.
+             */
+            if (e_locked) {
+                cache_unlock_entry(&inst->inst_cache, e);
```
droideck, line 246: The preamble releases e, parent, and newparent, but moddn_get_children() (inside the run-once block) takes cache_lock_entry() on every cached child
https://github.com/389ds/389-ds-base/blob/65a1afcdee407b85d6d1720560836f0a5f37ebfc/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c#L2186-L2191
And those are only released in the completion paths
https://github.com/389ds/389-ds-base/blob/65a1afcdee407b85d6d1720560836f0a5f37ebfc/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c#L1301-L1305
So a subtree rename sleeps holding N child monitors and then waits for the serial lock, while a modify of one of those children holds the serial lock and waits for the child's monitor. IIUC, that is the same AB-BA this patch removes for the top entries, moved to the children.
And cached tombstone children count too, so it is not limited to big subtrees. 

Possible, we can release the child locks in the preamble as well, so nothing entry-level is held across the backoff, and re-lock (or refetch) them after the serial lock is re-acquired. What do you think?

ldap/servers/slapd/back-ldbm/ldbm_delete.c
droideck, line 228: This dup (and the SDN read at :231) runs after e was unlocked a few lines above. The refcount keeps the backentry struct alive, but not ep_entry: the URP path at :421 does slapi_entry_free(e->ep_entry) and replaces it in place, under a lock we no longer hold. A replicated delete of the same entry during our backoff makes this read freed memory. Admittedly hard to hit, but the fix is cheap.

ldap/servers/slapd/back-ldbm/ldbm_delete.c
```diff
@@ -540,17 +586,49 @@ int delete_tombstone_entry = 0; /* We must remove the given tombstone entry from
...
                      * the parent.  If we fail to lock the entry, just try again.
+                     *
+                     * Must not sleep while holding SERIAL LOCK -- abort txn
+                     * first, sleep, then re-begin (same as deadlock retry).
                      */
                     while (1) {
                         parent = id2entry(be, pid, &txn, &retval);
                         if (parent && (cache_retry = cache_lock_entry(&inst->inst_cache, parent))) {
                             /* Failed to obtain parent entry's entry lock */
                             if (cache_retry == RETRY_CACHE_LOCK &&
```
droideck, line 597: Small commit note...
This restart triggers on RETRY_CACHE_LOCK, hence this part is backend-independent, so it is live on LMDB. And it looks like an improvement! (the old code slept holding what on LMDB is the single write transaction). Worth reflecting in the commit message, IMO.:)

droideck (changes requested):

droideck:
After my dive into the DB retry stuff, I decided to dig a bit deeper into this one... (as it might help another case with memberof massive writes) And... Well, you see what came out of it...
Sorry for the "changed my mind" recheck :(

Hopefully, some of the findings are worthy though!

Issue #7478, opened 2026-05-04
Regression in FreeIPA backup-restore test


flo-renaud:
**Issue Description**
FreeIPA has detected a regression in the backup-restore test using data-only backup, when executed with teh 389ds packages from @389ds/389-ds-base-nightly.

Test scenario:
1. install ipa server with `ipa-server-install --domain ipa.test --realm IPA.TEST --setup-dns --forwarder 10.11.5.160 -a Secret123 -p Secret123 -U`
2. add a prebackup user: `echo Secret123| kinit admin; ipa user-add prebackup --first pre --last backup`
3. perform a data-only backup with `ipa-backup --data`
4. create a postbackup user: `ipa user-add postbackup --first post --last backup`
5. restore the backup that contains only prebackupuser: `ipa-restore -U -p Secret123 /var/lib/ipa/backup/ipa-data-2026-05-04-08-56-04`
6. check the existing users. We can see both prebackup and postbackup users: `ipa user-find`

**Package Version and Platform:**
 - Platform: fedora 44
 - Package and version: 389-ds-base-3.2.1.202605040820git61d6233ea-1.fc44.x86_64

**Steps to Reproduce**
See test scenario above.

**Expected results**
The command `ipa user-find` should only display users created before the ipa-backup command was run.

**Additional context**

The ipa-restore command shows some errors:
```
# ipa-restore -U -p Secret123 /var/lib/ipa/backup/ipa-data-2026-05-04-08-56-04
Preparing restore from /var/lib/ipa/backup/ipa-data-2026-05-04-08-56-04 on server.ipa.test
Performing DATA restore from DATA backup
Temporary setting umask to 022
Each master will individually need to be re-initialized or
re-created from this one. The replication agreements on
masters running IPA 3.1 or earlier will need to be manually
re-enabled. See the man page for details.
Disabling all replication.
Stopping Directory Server
Restoring from userRoot in IPA-TEST
ldif2db failed: Error: Can't pickle local object <function DirSrv.check_ldif_file_is_readable.<locals>.chown_and_test at 0x7f466b1694e0>

Restoring from ipaca in IPA-TEST
ldif2db failed: Error: Can't pickle local object <function DirSrv.check_ldif_file_is_readable.<locals>.chown_and_test at 0x7f10aeded4e0>

Starting Directory Server
Restoring umask to 18
The ipa-restore command was successful
```

[...]

Issue #6867, opened 2025-07-11
dsctl ldif2db should show error if user has no access to the file


Dakomos:
https://github.com/389ds/389-ds-base/blob/2a3c0d8e8bbc2e6bdf9207ef7754016de2653f6e/src/lib389/lib389/__init__.py#L2690

If the user does not jave read permissions on the file it should also crash and show and error. I suggest adding another if underneath the check whether or not the path is actually a file.

`
if not os.access(import_file, os.R_OK):
    self.log.error("ldif2db: No read permission for file: %s", import_file)
    return False
`

vashirov:
Hi @Dakomos,
under what user do you run `dsctl`?
Normally `dsctl` is used by root, so this user should always have access to the file. The problem might be that `ns-slapd` can't read the file, because it runs under `dirsrv` user.

progier389:
This issue cause a regression. See #7478

PR #7479, opened 2026-05-04
Issue 7478 - Regression in FreeIPA backup-restore test

progier389:
Fix a regression caused by #6867 
It seems due by the use of inner function inside a function.
 Using class static method instead

Issue: #7478 

Reviewed by:  @vashirov  (Thanks!)

## Summary by Sourcery

Enhancements:
- Refactor the file readability check from an inline nested function into a reusable DirSrv static method used by the multiprocessing Process target.

### DIFF
--- a/src/lib389/lib389/__init__.py
+++ b/src/lib389/lib389/__init__.py
@@ -2742,9 +2742,9 @@ class DirSrv(SimpleLDAPObject, object):
             online = False
         DirSrvTools.runUpgrade(self.ds_paths.prefix, online)
 
-    # Become dirsrv and check that wimport_file is still readable
+    # Become dirsrv and check that import_file is still readable
     @staticmethod
-    def chown_and_test(pw, import_file):
+    def _chown_and_test(pw, import_file):
         if os.getuid() == 0:
             os.setgroups([])
             os.setgid(pw.pw_gid)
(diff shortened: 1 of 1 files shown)


src/lib389/lib389/__init__.py
```diff
@@ -2742,20 +2742,22 @@ def upgrade(self, upgradeMode):
             online = False
         DirSrvTools.runUpgrade(self.ds_paths.prefix, online)
 
+    # Become dirsrv and check that wimport_file is still readable
```
vashirov:
```suggestion
    # Become dirsrv and check that import_file is still readable
```

src/lib389/lib389/__init__.py
```diff
@@ -2742,20 +2742,22 @@ def upgrade(self, upgradeMode):
...
+    @staticmethod
+    def chown_and_test(pw, import_file):
```
vashirov: With this change it's now a part of the public DirSrv API, is this intentional?  If not, we should underscore it.

vashirov (approved):
Beside small details, LGTM.

src/lib389/lib389/__init__.py
progier389: Good catch! Better use private function

PR #7488, opened 2026-05-11
Issue 7475 - Fix CI test failures in test_fd_limits

### LINKED ISSUE #7475: Fix CI test failures
**Issue Description**
`retrocl`, `resource_limits` and `clu` test suites are flaky and fail from time to time.

droideck:
Description: MAX_FD_VAL was computed from pytest's RLIMIT_NOFILE, but the server caps against its own rlim_max (set by systemd LimitNOFILE for dirsrv@). When pytest's limit is lower, TOO_HIGH_VAL = MAX_FD_VAL+1 stays below the server cap and the modify is accepted instead of rejected.

Compute MAX_FD_VAL from SYSTEMD_LIMIT (handling "infinity"), and use it for CUSTOM_VAL and the default assertion.

Fixes: https://github.com/389ds/389-ds-base/issues/7475

Reviewed by: ?

## Summary by Sourcery

Align file descriptor limit tests with the server's effective systemd-imposed max to prevent CI failures when pytest's RLIMIT_NOFILE is lower than the server cap.

Bug Fixes:
- Compute MAX_FD_VAL from the systemd LimitNOFILE value (handling "infinity") instead of pytest's RLIMIT_NOFILE, ensuring overly high descriptor values are correctly rejected in tests.

Tests:
- Update fd limit tests to derive custom and assertion values from the server's effective max descriptor cap and adjust expectations for max and reserved descriptor attributes accordingly.

### DIFF
--- a/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
+++ b/dirsrvtests/tests/suites/resource_limits/fdlimits_test.py
@@ -10,7 +10,6 @@ import logging
 import pytest
 import os
 import ldap
-import resource
 from lib389.backend import Backends
 from lib389._constants import *
 from test389.topologies import topology_st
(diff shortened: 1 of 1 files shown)


mirielka (approved):
LGTM!

Issue #7468, opened 2026-04-30
Add password breach checking via HIBP API


jchapma:
**Is your feature request related to a problem? Please describe.**
Customers with large user bases (e.g., ~100k users) experience spam and account compromise issues due to leaked passwords. Currently, 389 Directory Server has no mechanism to detect whether a user's password has appeared in known data breaches. Administrators have no way to proactively identify and remediate compromised credentials at the LDAP level without knowing the plaintext passwords.

**Describe the solution you'd like**
Implement password breach checking using the HIBP API with k-anonymity (only the first 5 characters of the SHA-1 hash are sent externally). Check passwords at set/change time via check_pw_syntax_ext() and at login via pre-bind hook. When a breach is detected, force a password change rather than rejecting the bind. Provide configuration for enabling the feature, custom API URLs, and timeout.

**Additional context**
HIBP API documentation: https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/#cloudflareprivacyandkanonymity

jchapma:
84d2e10c2..e4121a409  389-ds-base-3.2 -> 389-ds-base-3.2

PR #7492, opened 2026-05-12
Issue 7468 - RFE - HIBP password breach validation

jchapma:
Description:
Integrate the HIBP HTTP client into password policy. Adds config, schema, and password validation logic to check passwords against the HIBP breach database during password add and modify operations. Depends on the HIBP client PR.
- Schema: passwordBreachCheck, passwordBreachDbUrl, passwordBreachDbTimeout
- Config: Config setters in libglobs.c
- Validation: Check passwords on add/modify for both admin and non-admin users
- rootDN: Also validates nsslapd-rootpw changes against breach database

Relates: https://github.com/389ds/389-ds-base/issues/7468

Reviewed by:

## Summary by Sourcery

Integrate breached password checking into the directory server’s password policy using the HIBP client, including configuration, schema wiring, and runtime enforcement for password operations.

New Features:
- Add configurable password breach checking to the global password policy, including enable/disable flag, breach database URL, and timeout settings.
- Validate user and rootDN passwords against the HIBP breach database during add and modify operations when the feature is enabled.

Enhancements:
- Initialize and wire password breach policy settings into existing password policy handling so per-policy and global defaults are honored.
- Hook HIBP initialization into server startup so breached password checks are available when configured.


mreynolds389:
@jchapma I think this needs to be rebased.  It's not building.

mreynolds389 (approved):

ldap/servers/slapd/libglobs.c
```diff
@@ -3796,6 +3816,90 @@ config_set_pw_dict_path(const char *attrname, char *value, char *errorbuf, int a
...
+int32_t
+config_set_pw_breach_check(const char *attrname, char *value, char *errorbuf, int apply)
+{
+    int32_t retVal = LDAP_SUCCESS;
+    slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+
+#ifndef ENABLE_HIBP
+    if (apply && value && strcasecmp(value, "on") == 0) {
+        slapi_log_err(SLAPI_LOG_WARNING, "config_set_pw_breach_check",
+                      "HIBP breached password checking not enabled - passwordBreachCheck has no effect\n");
```
droideck: I think we should reject `on` with an LDAP error when HIBP support is unavailable.
An admin would see the setting as enabled and reasonably believe breached-password protection is active. I don't think logging the warning is visible enough.

ldap/servers/slapd/modify.c
```diff
@@ -1275,6 +1306,38 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
...
+#ifdef ENABLE_HIBP
+    /* Check all passwords against breach database before any other checks */
+    if (pwpolicy->pw_check_breach && mod->mod_bvalues) {
+        Slapi_Value **breach_vals = NULL;
+        valuearray_init_bervalarray(mod->mod_bvalues, &breach_vals);
+        if (breach_vals) {
+            for (size_t i = 0; breach_vals[i] != NULL; i++) {
+                const char *pwd = slapi_value_get_string(breach_vals[i]);
+                if (pwd && !slapi_is_encoded((char *)pwd)) {
+                    int breach_count = hibp_check_password(pwd, pwpolicy);
```
droideck: If we call HIBP before the ACI/password-change checks, I think, an unauthorized user can trigger outbound HIBP requests for chosen values before the server rejects the modify. The add path already checks ACI first; modify should do the same, IMO

ldap/servers/slapd/modify.c
```diff
@@ -1275,6 +1306,38 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
     pwpolicy = new_passwdPolicy(pb, (char *)slapi_sdn_get_ndn(&sdn));
     internal_op = operation_is_flag_set(operation, OP_FLAG_INTERNAL);
 
+#ifdef ENABLE_HIBP
+    /* Check all passwords against breach database before any other checks */
+    if (pwpolicy->pw_check_breach && mod->mod_bvalues) {
```
droideck: IIUC, it will be checked even for LDAP_MOD_DELETE of the userPassword attribute and that seems wrong.  We should be able to delete breached password

ldap/servers/slapd/pw.c
```diff
@@ -2365,6 +2365,21 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
...
                     }
+                } else if (!strcasecmp(attr_name, "passwordBreachCheck")) {
+                    if ((sval = attr_get_present_values(attr))) {
+                        pwdpolicy->pw_check_breach =
+                            pw_boolean_str2value(slapi_value_get_string(*sval));
+                    }
+                } else if (!strcasecmp(attr_name, "passwordBreachDbUrl")) {
+                    if ((sval = attr_get_present_values(attr))) {
+                        pwdpolicy->pw_breach_db_url =
+                            slapi_ch_strdup(slapi_value_get_string(*sval));
```
droideck: Do we free it? I'm not fully sure who owns it (I wasn't able to find)

ldap/servers/slapd/libglobs.c
```diff
@@ -3796,6 +3816,90 @@ config_set_pw_dict_path(const char *attrname, char *value, char *errorbuf, int a
...
+    if (apply && value && strlen(value) > 0) {
+        slapi_log_err(SLAPI_LOG_WARNING, "config_set_pw_breach_url",
+                      "HIBP breached password checking not enabled  - passwordBreachDbUrl has no effect\n");
+    }
+#endif
+
+    if (config_value_is_null(attrname, value, errorbuf, 0)) {
+        value = NULL;
+    }
+
```
droideck, line 3898: I wonder if we should validate `passwordBreachDbUrl` to require `https://`, normalize/require a trailing `/` and ideally also set libcurl allowed protocols to HTTPS only with CURLOPT_PROTOCOLS_STR. CURLOPT_DEFAULT_PROTOCOL can make scheme-less URLs default to HTTPS, but for this security feature I’d still reject scheme-less input rather than guess. WDYT?

droideck (changes requested):

ldap/servers/slapd/modify.c
jchapma: You are correct, we should do the HIBP check after ACI checks, thanks

ldap/servers/slapd/pw.c
jchapma: We/I do not :(

ldap/servers/slapd/libglobs.c
jchapma: Thats a good idea, currently there is no validation. Thanks

ldap/servers/slapd/pw.c
```diff
@@ -2442,6 +2457,9 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
                     pwdpolicy->pw_palindrome = g_pwdpolicy->pw_palindrome;
                     pwdpolicy->pw_check_dict = g_pwdpolicy->pw_check_dict;
                     pwdpolicy->pw_dict_path = g_pwdpolicy->pw_dict_path;
+                    pwdpolicy->pw_check_breach = g_pwdpolicy->pw_check_breach;
+                    pwdpolicy->pw_breach_db_url = g_pwdpolicy->pw_breach_db_url;
```
droideck: I think this might accidentally hand ownership of the global breach URL to the local policy. Since delete_passwdPolicy() frees pw_breach_db_url for local policies, could we duplicate this value here or otherwise avoid freeing a borrowed global pointer?

dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -1184,6 +1184,93 @@ def fin():
...
+        3. Success
+    """
+    inst = topo.standalone
+
+    # Verify passwordBreachCheck default value
+    value = inst.config.get_attr_val_utf8('passwordBreachCheck')
+    assert value.lower() == 'off'
+
+    # Verify setting passwordBreachCheck to 'on'
+    inst.config.set('passwordBreachCheck', 'on')
```
droideck: I think this may fail when the server is built without HIBP support, which is the default.

ldap/servers/slapd/modify.c
```diff
@@ -1329,6 +1360,43 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
         /* done with slapi entry e */
         slapi_search_get_entry_done(&entry_pb);
 
+#ifdef ENABLE_HIBP
+        /*
+         * Check password against breach database after ACI validation.
+         */
+        if (!SLAPI_IS_MOD_DELETE(mod->mod_op) &&
+            !pw_is_pwp_admin(pb, pwpolicy, PWP_ADMIN_OR_ROOTDN) &&
```
droideck: The add path checks all users, and the commit message says admin and non-admin password changes are validated, but this condition lets admins bypass HIBP during modify. I think we might have an inconsistancy unless I miss something...

ldap/servers/slapd/libglobs.c
```diff
@@ -583,6 +587,21 @@ static struct config_get_and_set
...
+    /* password breach database URL */
+    {CONFIG_PW_BREACH_URL_ATTRIBUTE, config_set_pw_breach_url,
+     NULL, 0,
+     (void **)&global_slapdFrontendConfig.pw_policy.pw_breach_db_url,
+     CONFIG_STRING, NULL, "", NULL},
+    /* password breach database timeout */
+    {CONFIG_PW_BREACH_TIMEOUT_ATTRIBUTE, config_set_pw_breach_timeout,
+     NULL, 0,
+     (void **)&global_slapdFrontendConfig.pw_policy.pw_breach_db_timeout,
+     CONFIG_INT, NULL, "5", NULL},
```
droideck: This default is 5... While the policy/client defaults are 10. Should they be align?

droideck (changes requested):

ldap/servers/slapd/pw.c
jchapma: Ah yes, this is a bug. The global pw_breach_db_url is asigned directly and will be freed, possibly causing a double free. Duplicating pw_breach_db_url.

dirsrvtests/tests/suites/config/config_test.py
jchapma: I just tried it and this test does fail when HIBP is not enabled, I will add a try/except to skip the test the HIBP assertions when the server returned UNWILLING_TO_PERFORM. Apologies I though I had tested this

ldap/servers/slapd/modify.c
jchapma: Good catch,  there was an inconsistency. The modify path allowed admins to bypass HIBP, but add did not. Added the admin bypass check to add.c

ldap/servers/slapd/libglobs.c
jchapma: Updated to match HIBO default

ldap/servers/slapd/modify.c
```diff
@@ -1329,6 +1360,41 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
         /* done with slapi entry e */
         slapi_search_get_entry_done(&entry_pb);
 
+#ifdef ENABLE_HIBP
+        /* Check password against breach database after ACI validation (admin bypass) */
+        if (!SLAPI_IS_MOD_DELETE(mod->mod_op) &&
```
droideck: Could we share this non-admin HIBP check with the Password Modify extended-operation path while preserving the existing PWP_ADMIN_OR_ROOTDN exemption? I may be missing another hook, but an ordinary user's passwd_s request becomes an internal Modify and does not reach this block. I think the direct Modify will reject the breached password, while passwd_s will install it without making an HIBP request. A shared helper called after determining the requester's administrator status might keep the paths consistent without changing the intentional admin bypass.

ldap/servers/slapd/modify.c
```diff
@@ -1091,6 +1094,34 @@ op_shared_modify(Slapi_PBlock *pb, int pw_change, char *old_pw)
...
+        /* Check rootpw against breach database before hashing */
+        slapdFrontendConfig_t *slapdFrontendConfig = getFrontendConfig();
+        if (slapdFrontendConfig->pw_policy.pw_check_breach) {
+            for (size_t i = 0; mods && mods[i]; i++) {
+                if (strcasecmp(mods[i]->mod_type, CONFIG_ROOTPW_ATTRIBUTE) == 0 &&
+                    mods[i]->mod_bvalues) {
+                    for (size_t j = 0; mods[i]->mod_bvalues[j]; j++) {
+                        char *val = mods[i]->mod_bvalues[j]->bv_val;
+                        if (val && !slapi_is_encoded(val)) {
+                            int breach_count = hibp_check_password(val, &slapdFrontendConfig->pw_policy);
```
droideck: I think we might have a tricky situation here. As far as I can trace, the DSE ACL check happens later in dse_modify(). An anonymous Modify with four nsslapd-rootpw values therefore might produce four delayed HTTPS requests before returning INSUFFICIENT_ACCESS. Regardless of whether an authorized rootDN should bypass HIBP, an unauthorized request should not be able to trigger this outbound work.

ldap/servers/slapd/pw.c
```diff
@@ -2359,6 +2359,21 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
                         pwdpolicy->pw_check_dict =
                             pw_boolean_str2value(slapi_value_get_string(*sval));
                     }
+                } else if (!strcasecmp(attr_name, "passwordBreachCheck")) {
+                    if ((sval = attr_get_present_values(attr))) {
+                        pwdpolicy->pw_check_breach =
```
droideck, line 2365: Local password-policy entries probably should use the same validation rules as the global setters. A local policy currently accepts an http:// endpoint, but the HIBP client permits only HTTPS. The resulting request therefore always fails, and password validation fails open. I reproduced this with an ordinary user: the local policy had breach checking enabled, but a breached password was accepted because libcurl rejected the configured protocol. Perhaps the existing HTTPS/trailing-slash validation could be shared by global and local policy Add/Modify paths.

ldap/servers/slapd/libglobs.c
```diff
@@ -3802,6 +3822,114 @@ config_set_pw_dict_path(const char *attrname, char *value, char *errorbuf, int a
...
+            slapi_create_errormsg(errorbuf, SLAPI_DSE_RETURNTEXT_SIZE,
+                                  "%s: URL must end with a trailing slash (e.g., https://api.pwnedpasswords.com/range/)",
+                                  attrname);
+            return LDAP_UNWILLING_TO_PERFORM;
+        }
+    }
+
+    if (apply) {
+        CFG_LOCK_WRITE(slapdFrontendConfig);
+        slapi_ch_free_string(&slapdFrontendConfig->pw_policy.pw_breach_db_url);
```
droideck, line 3919: I wonder if readers snapshot or duplicate the HIBP configuration under CFG_LOCK_READ before this pointer is replaced... I may be overlooking an ownership guarantee (it was some time as I worked on this part of the server), but normal policy construction shallow-copies this URL, and the root-password path also borrows it without the read lock. Hence, an online configuration update could free it while another operation is copying or using it. WDYT? Is it a valid concern?

ldap/servers/slapd/main.c
```diff
@@ -1079,6 +1082,14 @@ main(int argc, char **argv)
 
     /* pw_init() needs to be here since it uses aci function calls.  */
     pw_init();
+
+#ifdef ENABLE_HIBP
+    /* Initialise breached password checking */
+    if (hibp_init() != 0) {
```
droideck: Would it make sense to initialize HIBP/libcurl before plugin_startall() and the housekeeping/event threads? The configured libcurl floor is 7.61, and older supported versions require curl_global_init() to run before other threads exist. At this point several threads have already started. Perhaps this could run after detach() and NSS initialization, but before plugin startup.

ldap/servers/slapd/pw.c
```diff
@@ -2445,6 +2460,9 @@ new_passwdPolicy(Slapi_PBlock *pb, const char *dn)
                     pwdpolicy->pw_palindrome = g_pwdpolicy->pw_palindrome;
                     pwdpolicy->pw_check_dict = g_pwdpolicy->pw_check_dict;
                     pwdpolicy->pw_dict_path = g_pwdpolicy->pw_dict_path;
+                    pwdpolicy->pw_check_breach = g_pwdpolicy->pw_check_breach;
+                    pwdpolicy->pw_breach_db_url = slapi_ch_strdup(g_pwdpolicy->pw_breach_db_url);
```
droideck: I think we might have a leak here (BTW, have you checked with ASAN and LSAN?).
A local URL may already have been duplicated while reading the policy entry, so this assignment appears to lose that allocation whenever global syntax inheritance applies. Since policy objects are constructed during binds and password operations, this could accumulate over the server's lifetime, unless I miss something...

dirsrvtests/tests/suites/config/config_test.py
```diff
@@ -1184,6 +1184,98 @@ def fin():
...
+
+    # Verify setting passwordBreachCheck to 'on' (requires HIBP support)
+    try:
+        inst.config.set('passwordBreachCheck', 'on')
+        value = inst.config.get_attr_val_utf8('passwordBreachCheck')
+        assert value.lower() == 'on'
+        # Reset
+        inst.config.set('passwordBreachCheck', 'off')
+    except ldap.UNWILLING_TO_PERFORM:
+        pytest.skip("HIBP support not compiled in")
```
droideck: Could this no-HIBP case be split from the validation assertions below? On the standard build, pytest.skip() exits the whole test, so the invalid-value check never runs. Since the normal CI configuration also does not compile the HIBP-guarded paths, it may be useful to add one HIBP-enabled build plus functional coverage for direct Modify, Password Modify, and root-password changes.

droideck (changes requested):

ldap/servers/slapd/modify.c
jchapma: Good catch, the ext op path was indeed bypassing the HIBP check. It uses the same bypass and fail open behavior as modify.c, so both paths are now consistent. Thanks

ldap/servers/slapd/modify.c
jchapma: Good catch. I've added an isroot check before the HIBP loop,  unauthorised users will now skip the HIBP check entirely and be rejected by the DSE ACL later.

ldap/servers/slapd/pw.c
jchapma: Yes, you are correct, local pwd policies should use the same validation rules. I've added the same URL validation for local pwd policies.

ldap/servers/slapd/libglobs.c
jchapma: Hmm, yes this is a possibility.  I have added a thread safe getter that holds CFG_LOCK_READ while duplicating, and updated both new_passwdPolicy() and the rootpw HIBP check path to use it.

ldap/servers/slapd/main.c
jchapma: Yes, it does make sense. Good point. I have moved hibp_init() (and curl_global_init()) earlier in the startup sequence. This ensures libcurl is inited before any plugin threads or housekeeping threads are started.

ldap/servers/slapd/pw.c
jchapma: We do :(. I will fix this
Direct leak of 31 byte(s) in 1 object(s) allocated from:
    #0 0x7f419a0814a8 in strdup (/usr/lib64/libasan.so.8+0x814a8) (BuildId: c685e4853bc63d50ad0ad4f295435a30eafbb1f1)
    #1 0x7f4199832650 in slapi_ch_strdup ldap/servers/slapd/ch_malloc.c:211
    #2 0x7f41999a5c0a in new_passwdPolicy ldap/servers/slapd/pw.c:2381
    #3 0x7f419992993d in op_shared_allow_pw_change ldap/servers/slapd/modify.c:1317
    #4 0x7f4199922325 in do_modify ldap/servers/slapd/modify.c:338
    #5 0x49f7ec in connection_dispatch_operation ldap/servers/slapd/connection.c:719
    #6 0x4a78cf in connection_threadmain ldap/servers/slapd/connection.c:2103
    #7 0x7f4198fd5748 in _pt_root (/lib64/libnspr4.so+0x2d748) (BuildId: 7635f4fc07e42fdd171fea0f0107d7c608d22ed7)
    #8 0x7f419947d946 in start_thread (/lib64/libc.so.6+0x8c946) (BuildId: 9148cab1b932d44ef70e306e9c02ee38d06cad51)
    #9 0x7f419950396f in __clone3 (/lib64/libc.so.6+0x11296f) (BuildId: 9148cab1b932d44ef70e306e9c02ee38d06cad51)

SUMMARY: AddressSanitizer: 31 byte(s) leaked in 1 allocation(s).

ldap/servers/slapd/modify.c
```diff
@@ -1304,35 +1312,38 @@ op_shared_allow_pw_change(Slapi_PBlock *pb, LDAPMod *mod, char **old_pw, Slapi_M
...
-                                slapi_pwpolicy_make_response_control(pb, -1, -1, LDAP_PWPOLICY_INVALIDPWDSYNTAX);
+        if (!SLAPI_IS_MOD_DELETE(mod->mod_op) && pwpolicy->pw_check_breach && mod->mod_bvalues) {
+            if (pw_is_pwp_admin(pb, pwpolicy, PWP_ADMIN_OR_ROOTDN)) {
+                slapi_log_err(SLAPI_LOG_DEBUG, "op_shared_allow_pw_change",
+                    "Skipping breach check for %s - admin bypass\n", dn);
+            } else {
+                Slapi_Value **breach_vals = NULL;
+                valuearray_init_bervalarray(mod->mod_bvalues, &breach_vals);
+                if (breach_vals) {
+                    for (size_t i = 0; breach_vals[i] != NULL; i++) {
```
droideck, line 1360: I wonder if might have a security issue here... Each value triggers a blocking HTTP call, so multivalue replaces can exhaust the worker pool.
I think we need an operation-wide cap on cleartext userPassword values before the first HIBP request.

dirsrvtests/tests/suites/password/pwdPolicy_breach_test.py
```diff
@@ -0,0 +1,401 @@
...
+
+    yield inst
+
+    log.info('Disabling HIBP password breach checking')
+    inst.simple_bind_s(DN_DM, PASSWORD)
+    inst.config.set('passwordBreachCheck', 'off')
+
+
+@pytest.fixture(scope="function")
+def test_user(topology_st):
```
droideck: Testimony GH CI failure here:)

droideck (commented):
Small issue, otherwise -- looks good!

ldap/servers/slapd/modify.c
jchapma: I agree, I will add an operation  cap that rejects operations with too many cleartxt password values. Thanks

droideck (approved):
LGTM!
Very nice RFE implementation!

jchapma:
> LGTM! Very nice RFE implementation!

Thank you for your feedback, your review comments have really improved this RFE.

Issue #7493, opened 2026-05-12
RFE - Create a shadowLastChange fixup task


mreynolds389:
**Description**

ShadowAccount attributes are generated when a password is updated, but when users are imported these "shadow" attributes are not updated.  We need a fixup tasks to set ShadowLastChange if it's missing, but also to fix the value if it is stale.

mreynolds389:
173661d55..53924b67e  389-ds-base-3.1 -> 389-ds-base-3.1
314ccddb0..59bfb2bf0  389-ds-base-3.0 -> 389-ds-base-3.0

PR #7494, opened 2026-05-12
Issue 7493 - RFE - Add ShadowAccount fixup task

mreynolds389:
Description:

ShadowAccount attributes are generated when a password is updated, but when users are imported these "shadow" attributes are not updated. We need a fixup tasks to set ShadowLastChange if it's missing, but also to fix the value if it is stale.

Design doc:

https://www.port389.org/docs/389ds/design/shadow-fixup-design.html

Relates: https://github.com/389ds/389-ds-base/issues/7493

## Summary by Sourcery

Add server-side and client tools to run a ShadowAccount shadowLastChange fixup task, with corresponding CLI, UI, and test coverage.

New Features:
- Introduce a background directory server task to scan ShadowAccount entries under a suffix and create or correct shadowLastChange values, with optional force mode.
- Expose the shadowLastChange fixup as a lib389 CLI subcommand and cockpit UI workflow, including a dedicated modal and task page for running and monitoring the task.

Enhancements:
- Add configuration accessors and password policy UI layout refinements, including improved PBKDF2 iteration controls and field alignment.
- Improve cockpit task buttons with loading indicators and error/helper text for DN validation, and update various headings for consistency.

Tests:
- Add an integration test exercising the shadowLastChange fixup task across missing, correct, stale, and forced-update scenarios in password policy tests.

### DIFF
--- a/dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
+++ b/dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2016 Red Hat, Inc.
+# Copyright (C) 2026 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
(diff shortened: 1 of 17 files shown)


ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,261 @@ pw_init(void)
...
+    task_data *mytaskdata = NULL;
+    char *suffix = NULL;
+    bool force = false;
+    int rc = SLAPI_DSE_CALLBACK_OK;
+
+    *returncode = LDAP_SUCCESS;
+
+    /* get our suffix */
+    suffix = slapi_entry_attr_get_charptr(e, "suffix");
+    if (suffix == NULL) {
```
tbordaz, line 405: Should we also check that the suffix exist (slapi_get_mapping_tree_node_by_dn) ?

tbordaz (approved):
Looks good to me.
minor comment

ldap/servers/slapd/pw_mgmt.c
mreynolds389: It will fail eventually, but it would be cleaner to catch it sooner and report a nice error message to the user

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -823,6 +823,216 @@ def fin():
...
+    c_before = '20000'
+    log.info('Run fixup shadow attributes task (suffix only, no force; stale reconciliation)')
+    shadow_fixup_task = ShadowFixupTask(inst)
+    shadow_fixup_task.create(suffix=DEFAULT_SUFFIX, force=False)
+    shadow_fixup_task.wait(timeout=120)
+    assert shadow_fixup_task.get_exit_code() == 0
+
+    # User C should have ShadowLastChange updated to the current day
+    c_after_first = int(user_c.get_attr_val_utf8('shadowLastChange'))
+    assert c_after_first != c_before, (
```
droideck: Hmm, `c_before` is a string and `c_after_first` is an int, so it passes even if the value stays 20000. May be we can assert the exact expected day from passwordExpirationTime - passwordMaxAge?

ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,261 @@ pw_init(void)
...
+
+        if (!slapi_entry_attr_get_ref(entries[i], "userpassword")) {
+            /* No userpassword, then there is nothing to fixup */
+            skip_count++;
+            continue;
+        }
+        pwdExpireTime = (char *)slapi_entry_attr_get_ref(entries[i], "passwordExpirationTime");
+        lastChangeTime = slapi_entry_attr_get_longlong(entries[i], "shadowLastChange");
+
+        if (lastChangeTime && pwdExpireTime && !mytaskdata->force) {
```
droideck: I think this might mix up "missing" with "value is 0".
Do I understand correctly that `shadowLastChange: 0` is valid for must-change accounts?

src/cockpit/389-console/src/lib/database/databaseModal.jsx
```diff
@@ -502,6 +514,143 @@ class ImportModal extends React.Component {
...
+                                    />
+                                ))}
+                            </FormSelect>
+                        </GridItem>
+                    </Grid>
+                    <Grid title={_("Branch to run the fix-up task on.")}>
+                        <GridItem offset={2} span={10}>
+                            <TextInput
+                                type="text"
+                                id="fixupShadowSuffix"
```
droideck: The `id` in  TextInput and FormSelect are duplicates (fixupShadowSuffix).
I'm not fully sure if it'll conflict but it might for some verifiers...

ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,261 @@ pw_init(void)
...
+}
+
+/*
+ * ShadowLastChange fixup task fucntions
+ */
+static void
+shadow_fixup_task_destructor(Slapi_Task *task)
+{
+    if (task) {
+        task_data *mytaskdata = (task_data *)slapi_task_get_data(task);
```
droideck, line 378: I see that some other tasks use the refcount check. Should we use it here too?

fixup_tombstone_task_destructor for example:
```c
        while (slapi_task_get_refcount(task) > 0) {
            /* Yield to wait for the fixup task finishes. */
            DS_Sleep(PR_MillisecondsToInterval(100));
        }
```

mreynolds389: Fixed

src/cockpit/389-console/src/lib/database/databaseModal.jsx
mreynolds389: Fixed

ldap/servers/slapd/pw_mgmt.c
mreynolds389: Fixed

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
mreynolds389: Fixed

ldap/servers/slapd/pw_mgmt.c
mreynolds389: Fixed

ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,294 @@ pw_init(void)
...
+        } else if (lastChangeTime != -1 && !mytaskdata->force) {
+            /* ShadowLastChange time is present, skip this entry */
+            skip_count++;
+            continue;
+        } else {
+            /* Build the timestamp for the shadowLastChange attribute */
+            cur_time = slapi_current_utc_time();
+            ctime = cur_time / _SEC_PER_DAY;
+            timestr = slapi_ch_smprintf("%ld", ctime);
+        }
```
droideck: I think this `if` branch is a bit tricky regarding passwordExpirationTime... 
So if `--force` is used on a suffix that can move every existing shadowLastChange forward to the task date, even when passwordExpirationTime contains enough information to preserve the real password-change day.

From the existing design point of view, that might not be right... I think we need either explicitly mention that or, may be, recompute the value from passwordExpirationTime when it is valid, and only use today as a fallback. WDYT?

ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,294 @@ pw_init(void)
...
+                fixup_count, stale_count, fixup_count - stale_count, skip_count);
+        slapi_log_err(SLAPI_LOG_INFO, "shadow_fixup_task_thread",
+                "ShadowLastChange fixup task complete.  Updated %d entries (%d stale shadowLastChange, %d missing shadowLastChange), skipped %d entries.\n",
+                fixup_count, stale_count, fixup_count - stale_count, skip_count);
+    }
+    slapi_task_inc_progress(task);
+
+    slapi_free_search_results_internal(pb);
+    slapi_pblock_destroy(pb);
+    slapi_task_finish(task, result);
```
droideck, line 676: If one entry update fails and a later one succeeds, the task exits 0, so dsconf and Cockpit report success despite skipped failed updates. Just to clarify, is it by design?

ldap/servers/slapd/pw_mgmt.c
mreynolds389: The force option is designed to reset all entries regardless of current values:

https://www.port389.org/docs/389ds/design/shadow-fixup-design.html

What you are describing is how the default behavior already works.  It will fixup stale values if the passwwordExpirationtimte does not match up with shadowLastChange.

ldap/servers/slapd/pw_mgmt.c
mreynolds389: > If one entry update fails and a later one succeeds, the task exits 0, so dsconf and Cockpit report success despite skipped failed updates. Just to clarify, is it by design?

Yeah it should probably exit...

ldap/servers/slapd/pw_mgmt.c
droideck: Oh, right! I read the force path as "force reconciliation" but yeah, that's not the design here... Sorry for the noise!:)

ldap/servers/slapd/pw_mgmt.c
droideck: Yeah, I don't have a strong opinion here...
Personally, I'd use continue-on-error - so exit non-zero if any update failed.

This is a suffix-wide repair task, so fail-fast would be unfriendly, IMO: one bad entry or any odd issue with something would prevent fixing all later entries that are perfectly fixable. I think admin would want the task to repair as much as it can in one run and check what's wrong with the rest baddy-entries afterwards.

So yeah, a clean exit code would mean "everything that needed updating was updated." So probably, for that case, per-entry modify failures can be accumulated and reflected at the end.

My 5 cents)

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
```diff
@@ -823,6 +823,216 @@ def fin():
...
+    c_before = '20000'
+    log.info('Run fixup shadow attributes task (suffix only, no force; stale reconciliation)')
+    shadow_fixup_task = ShadowFixupTask(inst)
+    shadow_fixup_task.create(suffix=DEFAULT_SUFFIX, force=False)
+    shadow_fixup_task.wait(timeout=120)
+    assert shadow_fixup_task.get_exit_code() == 0
+
+    # User C should have ShadowLastChange updated to the current day
+    c_after_first = int(user_c.get_attr_val_utf8('shadowLastChange'))
+    assert c_after_first != int(c_before), (
```
droideck: IIUC, nothing stops us from deterministically asserting the exact value here: `c_after_first ==` the value of `(passwordExpirationTime - passwordMaxAge) / 86400`. And the same at `b_after_force`'s assert.
WDYT?

ldap/servers/slapd/pw_mgmt.c
```diff
@@ -335,4 +349,304 @@ pw_init(void)
...
+                          mytaskdata->suffix, mytaskdata->force ? " (force)" : "");
+    slapi_log_err(SLAPI_LOG_INFO, "shadow_fixup_task_thread",
+                  "ShadowLastChange fixup task starting (suffix: \"%s\"%s) ...\n",
+                  mytaskdata->suffix, mytaskdata->force ? " (force)" : "");
+
+    /* Search for entries that contain objectclass ShadowAccount */
+    slapi_search_internal_set_pb(pb, mytaskdata->suffix, LDAP_SCOPE_SUBTREE,
+                                 "(objectclass=ShadowAccount)", 0, 0, 0, 0,
+                                 (void *)pw_get_componentID(), 0);
+    slapi_search_internal_pb(pb);
```
droideck: Could we use `slapi_search_internal_callback_pb()` here instead of collecting `SLAPI_PLUGIN_INTOP_SEARCH_ENTRIES`? This task can run over a full suffix, so materializing every matching `shadowAccount` entry may hold a large result set in memory before processing starts...

And maybe request only the attrs needed by the fixup, for example:

```
"userpassword",
"passwordExpirationTime",
"shadowLastChange",
```

droideck (approved):
Besides the couple of comments, looks great!

ldap/servers/slapd/pw_mgmt.c
mreynolds389: I can do that, really this is a fixup task that should only ever need to run once.  I wish you would have brought this up sooner as these new requests are now going to cause this to miss this sprint's deadline.  Anyway I'll work on it...

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
mreynolds389: Sorry maybe I need some coffee, but I don't get the concern here.  I set a bad value, and then I  make sure it's updated after the fixup task.  This is not enough?

droideck: Yeah, fair. I agree it's enough to prove it changed.
I was thinking exact value would also prove it changed for the expected reason and not. For example, it should not accidentally fall back to "today". But not a blocker from my side.

ldap/servers/slapd/pw_mgmt.c
droideck: Sorry for the late ask, I haven't noticed it before...
Yeah, callback/smaller attrs would be nicer for large suffixes, so I'm fine with either fixing it now or doing it as a follow-up some time later. Up to you.

dirsrvtests/tests/suites/password/pwdPolicy_attribute_test.py
mreynolds389: I changed it as requested

ldap/servers/slapd/pw_mgmt.c
mreynolds389: All good, it's changed now.  Please review again @droideck

droideck (approved):
LGTM!

Issue #7500, opened 2026-05-18
Prevent integer underflow during stalled import


mreynolds389:
**Issue Description**

If an import becomes stalled for some reason the progress logging can underflow the unsigned integer showing a user progress rate:

```
[17/Mar/2026:18:18:19.512296196 +0000] - INFO - import_monitor_threads - import db1: Processed 1 entries (pass 2) – average rate 214748364.8/sec, recent rate 0.0/sec, hit ratio 0%
```

mreynolds389:
cdacf697c..777fa6816  389-ds-base-3.2 -> 389-ds-base-3.2
097f1f579..634e9f59b  389-ds-base-3.1 -> 389-ds-base-3.1
08bfb044a..2a81d9a6c  389-ds-base-3.0 -> 389-ds-base-3.0
a9fa73601..f719075de  389-ds-base-2.9 -> 389-ds-base-2.9
cabc38483..cca7d60ad  389-ds-base-2.8 -> 389-ds-base-2.8
ea9bfd121..0991e6baf  389-ds-base-2.7 -> 389-ds-base-2.7
6220437c8..fb7de0068  389-ds-base-2.6 -> 389-ds-base-2.6
cd76f4a30..d63f3819b  389-ds-base-2.4 -> 389-ds-base-2.4
acdbf1b46..8ddca11c6  389-ds-base-2.2 -> 389-ds-base-2.2

PR #7501, opened 2026-05-18
Issue 7500 - Prevent unsigned integer underflow during stalled import

mreynolds389:
Description:

During a stalled import the foreman's first ID could be greater than the history progress size which leads to underflowing the rate. Check if the foreman's first ID is greater than the progress size and just set it to zero.

Only affects BDB.

Relates: https://github.com/389ds/389-ds-base/issues/7500

## Summary by Sourcery

Prevent unsigned underflow in BDB import progress rate calculation and tighten MDB import worker slot iteration.

Bug Fixes:
- Guard BDB import progress rate calculation against cases where the foreman first ID exceeds the progress history size, avoiding unsigned underflow during stalled imports.

Enhancements:
- Use a size_t loop index when aggregating MDB worker slot entry counts to better match the slots array bounds type.


progier389 (approved):
LGTM

PR #7502, opened 2026-05-21
Issue 7468 - RFE - HIBP response caching

jchapma:
Description:
Add LRU cache for HIBP API responses to reduce latency and external API calls. The cache stores range query responses (prefix -> suffix list) rather than individual password verdicts, providing implicit negative caching since a cache hit returns the complete suffix list for matching.

- Hash table (256 buckets) plus a doubly linked list for LRU eviction
- Configurable size (default: 1000 entries) and TTL (default: 24 hours)
- Concurrency protection using NSPR locks
- Cache statistics tracking (hits/misses)
- Unit tests for cache operations

Dependencies:
- Issue 7468 - RFE - Add HIBP HTTP client
- Issue 7468 - RFE - HIBP password breach validation

Relates: https://github.com/389ds/389-ds-base/issues/7468

Reviewed by:

## Summary by Sourcery

Introduce an in-memory LRU cache for Have I Been Pwned (HIBP) range query responses and integrate it into the HIBP client lifecycle.

New Features:
- Add configurable, TTL-based LRU cache for HIBP prefix responses with hit/miss statistics.
- Cache HIBP API responses in hibp_query_api to reduce repeated external calls and latency.

Enhancements:
- Wire HIBP subsystem shutdown into the main server cleanup path to release cache resources cleanly.
- Refine HIBP public API header to separate core API, hash provider abstraction, and cache/test helper interfaces.

Tests:
- Add unit tests covering HIBP cache basic operations, LRU eviction behavior, TTL expiration, and LRU reordering under access patterns.

### DIFF
--- a/ldap/servers/slapd/hibp.h
+++ b/ldap/servers/slapd/hibp.h
@@ -11,39 +11,80 @@
 #include "slap.h"
 
 /*
- * Function pointer for pluggable SHA-1 implementation
- * Allows for different hash implementations (FIPS vs non FIPS)
+ * Core HIBP API
  */
-typedef int (*hibp_hash_fn)(const char *input, size_t len, unsigned char *digest);
-
-/* Set the hash provider, called during initialisation */
-void hibp_set_hash_provider(hibp_hash_fn fn);
 
 /* Initialise the HIBP subsystem */
 int hibp_init(void);
 
+/* Shutdown the HIBP subsystem */
+void hibp_shutdown(void);
+
 /*
  * Check if password appears in breach database
+ *
  * Returns: >0 = breach count, 0 = not found, -1 = error
  */
 int hibp_check_password(const char *password, passwdPolicy *pwpolicy);
 
 /*
+ * Hash provider abstraction
+ */
+
+/* Function pointer for pluggable SHA-1 implementation */
+typedef int (*hibp_hash_fn)(const char *input, size_t len, unsigned char *digest);
+
+/* Set the hash provider, called during initialisation */
+void hibp_set_hash_provider(hibp_hash_fn fn);
+
+/*
  * Convert password to uppercase SHA-1 hex string (40 chars + null terminator)
+ *
  * Returns: 0 on success, -1 on error
  */
 int hibp_sha1_hex(const char *password, char *hex_output);
 
-/* Test wrapper, expose hibp_parse_response for testing
- * Returns: breach count if found, 0 if not found, -1 on error
+/*
+ * Cache API
  */
-int hibp_parse_response_wrapper(char *response_data, const char *suffix);
 
-/* Test wrapper, expose hibp_write_callback for testing
- * Returns: buffer size on success, -1 on error
+/*
+ * Initialise the HIBP cache
+ *
+ * Returns: 0 on success, -1 on error
  */
-int hibp_write_callback_wrapper(const char *data, size_t data_len, size_t chunk_size, char **out_data);
+int hibp_cache_init(size_t max_size, int32_t ttl_seconds);
+
+/* Destroy cache and free all resources */
+void hibp_cache_destroy(void);
+
+/*
+ * Get cached response for prefix
+ *
+ * Returns: response data (caller must free), NULL if not cached
+ */
+char *hibp_cache_get(const char *prefix, size_t *response_size);
+
+/* Store response in cache */
+void hibp_cache_put(const char *prefix, const char *response_data, size_t response_size);
+
+/*
+ * Test helpers - expose internal functions for unit testing
+ */
 
 /* Set mock API response for testing */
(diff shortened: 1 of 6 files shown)


ldap/servers/slapd/hibp_client.c
```diff
@@ -392,6 +746,13 @@ hibp_query_api(const char *prefix, const char *api_url, HIBPResponse *response,
         return -1;
     }
 
+    /* Store successful response in cache */
+    if (g_hibp_cache && response->data && response->size > 0) {
+        hibp_cache_put(prefix, response->data, response->size);
```
droideck, line 751: Will you additionally modify the PR once another one is merged?
Currently, cache key ignores the configured breach DB URL which is understandable... But probably it should be handled as passwordBreachDbUrl can differ by policy, including local policies. A response from the public HIBP endpoint can be reused for an internal/custom endpoint with the same prefix, or vice versa, producing false "safe" or false "breached" decisions.

ldap/servers/slapd/hibp_client.c
```diff
@@ -80,6 +82,341 @@ typedef struct hibp_response
     size_t capacity;
 } HIBPResponse;
 
+/*
+ * LRU Cache for HIBP API responses
+ *
+ * Caches API responses by hash prefix.
+ * Uses double linked list for LRU ordering.
+ */
+#define HIBP_CACHE_DEFAULT_SIZE 1000    /* Default max cache entries */
```
droideck, line 91: IIUC, each accepted response can be up to HIBP_RESPONSE_MAX_SIZE of 1 MB. That makes the theoretical default cache size about 1 GB. That's a lot... At least, I think, we need to document it very explicitly. And maybe let it be configured? And even a way to disable the cache?

ldap/servers/slapd/hibp.h
```diff
@@ -11,39 +11,80 @@
...
+
+/* Destroy cache and free all resources */
+void hibp_cache_destroy(void);
+
+/*
+ * Get cached response for prefix
+ *
+ * Returns: response data (caller must free), NULL if not cached
+ */
+char *hibp_cache_get(const char *prefix, size_t *response_size);
```
droideck, line 66: Do we really need it to be public? 
Probably all of them can be left as static in hibp_client.c:
``` 
hibp_cache_get
hibp_cache_put
hibp_cache_init
hibp_cache_destroy
```

ldap/servers/slapd/hibp_client.c
```diff
@@ -322,6 +659,23 @@ hibp_query_api(const char *prefix, const char *api_url, HIBPResponse *response,
         return 0;
     }
 
+    /* Check cache first */
+    if (g_hibp_cache) {
+        size_t cached_size;
+        char *cached_data = hibp_cache_get(prefix, &cached_size);
+        if (cached_data) {
+            slapi_log_err(SLAPI_LOG_INFO, "hibp_query_api",
+                          "Cache HIT for prefix %s (size=%zu bytes)\n", prefix, cached_size);
```
droideck, line 668: A few slapi_log_err calls log every prefix hit/miss/store at INFO. Yeah, the prefix is intentionally partial, but it is still password-derived material... I’d downgrade this to debug/trace and avoid logging the prefix by default.

droideck (changes requested):
The new unit tests exercise bare cache operations, but not the real hibp_check_password() cache path because hibp_query_api() returns mock responses before cache lookup.
I think that would be nice to add tests for cache hit/miss through hibp_check_password, URL-scoped cache keys, duplicate prefix across two URLs, oversized response cache skipping/eviction, and concurrent miss storms?.. And run some ASAN with it, probably.

ldap/servers/slapd/hibp_client.c
jchapma: Good point, I made some updates so each cache entry stores the api url it came from, and lookups match on both (prefix, url). Now responses from different endpoints wont conflict with each other

ldap/servers/slapd/hibp_client.c
jchapma: Yes, in theory the default cache size could grow to 1GB, but in practice, real responses are ~40KB, giving a default size of ~40MB. But, I made some updates that will allow the admin tune or disable caching

ldap/servers/slapd/hibp.h
jchapma: We dont, I made all four static

ldap/servers/slapd/hibp_client.c
jchapma: Downgraded cache hit/miss/store logging from INFO to DEBUG and removed the prefix from all log messages

droideck:
In the comments, you said you made the changes but I can't see them... Were they pushed? Or we might have some weird GH bug)

Issue #7514, opened 2026-05-26
Crash when doing moddn on very large subtree


mreynolds389:
**Issue Description**

Attempting to move a subtree underneath another subtree leads to an assertion failure on debug builds

```
[26/May/2026:11:59:43.496465497 -0400] - ERR - cache - BAD CACHE ASSERTION at ../389-ds-base/ldap/servers/slapd/back-ldbm/cache.c/1394: e->ep_refcnt > 0


0x00007ff66efb3313 in cache_remove (cache=0x7ff66f303290, ptr=0x7ff12e321b60) at ../389-ds-base/ldap/servers/slapd/back-ldbm/cache.c:1394
1394	        ASSERT(e->ep_refcnt > 0);

#0  0x00007ff66efb3313 in cache_remove (cache=0x7ff66f303290, ptr=0x7ff12e321b60) at ../389-ds-base/ldap/servers/slapd/back-ldbm/cache.c:1394
#1  0x00007ff66effac13 in dsentrydn_moddn_rename (ptxn=0x7ff137ffc380, be=0x7ff66f4074c0, id=2, children=0x7ff12e234000, dn_parentdn=0x7ff12e21a0c0, 
    dn_newsuperiordn=0x7ff137ffc4f0, child_entries=0x7ff12e247000) at ../389-ds-base/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c:1665
#2  0x00007ff66eff918c in ldbm_back_modrdn (pb=0x7ff12e200000) at ../389-ds-base/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c:1132
#3  0x00007ff674289787 in op_shared_rename (pb=0x7ff12e200000, passin_args=0) at ../389-ds-base/ldap/servers/slapd/modrdn.c:652
#4  0x00007ff674288768 in do_modrdn (pb=0x7ff12e200000) at ../389-ds-base/ldap/servers/slapd/modrdn.c:225
#5  0x000000000040bbb8 in connection_dispatch_operation (conn=0x7ff66ba025b0, op=0x7ff128a00000, pb=0x7ff12e200000)
    at ../389-ds-base/ldap/servers/slapd/connection.c:711
#6  0x000000000040e9f9 in connection_threadmain (arg=0x7ff67257fe7c) at ../389-ds-base/ldap/servers/slapd/connection.c:2072
[...]

PR #7515, opened 2026-05-26
Issue 7514 - Crash when doing moddn on very large subtree

mreynolds389:
Description:

Attempting to move a subtree underneath another subtree leads to an assertion failure on debug builds.

relates: https://github.com/389ds/389-ds-base/issues/7514

## Summary by Sourcery

Bug Fixes:
- Fix an assertion/crash caused by improper cleanup of child entry copies during subtree modrdn operations.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
@@ -22,7 +22,7 @@ static int moddn_newrdn_mods(Slapi_PBlock *pb, const char *olddn, struct backent
 static IDList *moddn_get_children(back_txn *ptxn, Slapi_PBlock *pb, backend *be, struct backentry *parententry, Slapi_DN *parentdn, struct backentry ***child_entries, struct backdn ***child_dns, int is_resurect_operation);
 static int modrdn_rename_entry_update_indexes(back_txn *ptxn, Slapi_PBlock *pb, struct ldbminfo *li, struct backentry *e, struct backentry **ec, Slapi_Mods *smods1, Slapi_Mods *smods2, Slapi_Mods *smods3, Slapi_Mods *smods4);
 static void mods_remove_nsuniqueid(Slapi_Mods *smods);
-static int32_t dsentrydn_modrdn_update(backend *be, const char *newdn, struct backentry *e, struct backentry **ec, back_txn *txn);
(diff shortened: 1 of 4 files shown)


ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
```diff
@@ -1660,18 +1660,14 @@ dsentrydn_moddn_rename(
                                               newsuperiordncomps);
     }
 
-    if (0 == retval) {
-        for (size_t i = 0; child_entries[i]; i++) {
+    for (size_t i = 0; child_entry_copies[i]; i++) {
+        if (child_entry_copies[i]->ep_refcnt == 0) {
```
tbordaz: Is it the ep_refcnt that should be tested or ep_state or both.
I wonder if it could be
`if (ep_state & (ENTRY_STATE_NOT_IN_CACHE | ENTRY_STATE_DELETED | ENTRY_STATE_INVALID) &&
      ep_refcount == 0) {
    backentry_free()
} `

tbordaz:
I realize that child_entry_copy contains a duplication of all the targeted moving entry. That means it is unlimited. So if we are moving a large set of entries it will create a memory spike (in addition of a large set of internal updates), where it almost for free before storing dnentry.
Should we disable this attribute by default ? in another ticket.

ldap/servers/slapd/back-ldbm/ldbm_modrdn.c
progier389, line 1647: I am a bit puzzled by the logic of this code.
Why are we duplicating all children 
Then loop on renaming them
Then freed and remove from cache all duplicated entries. 

IMHO it will be more logical to:
 Iterate on all entries:
   duplicate the entry
    Inside dsentrydn_moddn_rename_child/dsentrydn_modrdn_update:
    do the CACHE_REMOVE/CACHE_RETURN  after successful cache_replace otherwise free the duplicated back entry

mreynolds389:
Currently this PR also has https://github.com/389ds/389-ds-base/pull/7517 applied (the changes around maxrecords) as these two issues work together.

mreynolds389:
> I realize that child_entry_copy contains a duplication of all the targeted moving entry. That means it is unlimited. So if we are moving a large set of entries it will create a memory spike (in addition of a large set of internal updates), where it almost for free before storing dnentry. Should we disable this attribute by default ? in another ticket.

I revised fix so we are not dealing with these copies.

progier389 (approved):
LGTM

Issue #7516, opened 2026-05-28
dblayer_bulk_nextdata should not return an error when maxrecords is hit


mreynolds389:
**Issue Description**

When iterating over a bulk data if we hit maxrecords (100) we return an error which prevents the loop logic from grabbing the next bulk chunk

mreynolds389:
This is most noticeable when doing a moddn and moving a subtree with 100+ entries to a new subtree.  The modrdn operation "succeeds", but it corrupts the database

mreynolds389:
e9d94d45a..467cd0bb3  389-ds-base-3.1 -> 389-ds-base-3.1
d20942284..b35ca0020  389-ds-base-3.0 -> 389-ds-base-3.0
c0d2b5c9e..bf9119020  389-ds-base-2.8 -> 389-ds-base-2.8
9ab79d496..e8553cc38  389-ds-base-2.7 -> 389-ds-base-2.7
5ea7d4015..3cc65c2e5  389-ds-base-2.6 -> 389-ds-base-2.6
c263c26b9..981a81073  389-ds-base-2.4 -> 389-ds-base-2.4
1fa7ffdb4..5b192fc18  389-ds-base-2.2 -> 389-ds-base-2.2

PR #7517, opened 2026-05-28
Issue 7516 - dblayer_bulk_nextdata should not return an error when maxrecords is hit

mreynolds389:
Description:

When doing a bulk read of an index if the candidates were more than 100 it would silently stop processing the indexi read. This would lead to database corruption. This is most easily reproduced by doing a moddn with a subtree that had a lot of child entries.

relates: https://github.com/389ds/389-ds-base/issues/7516

## Summary by Sourcery

Prevent bulk index reads from treating the max-records limit as an error and add regression coverage for modrdn operations with many child entries.

Bug Fixes:
- Fix bulk index iteration to stop returning errors when hitting the configured maxrecords limit, avoiding premature termination that could lead to database corruption during operations like modrdn.

Tests:
- Add a regression test that performs a modrdn of an organizational unit with over 100 child user entries to verify all children are correctly moved under the new superior.

### DIFF
--- a/dirsrvtests/tests/suites/basic/modrdn_bulk_children_test.py
+++ b/dirsrvtests/tests/suites/basic/modrdn_bulk_children_test.py
@@ -1,0 +1,96 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import os
+import pytest
+
+from lib389._constants import DEFAULT_SUFFIX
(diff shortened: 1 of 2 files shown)


mreynolds389:
After further investigation looks like we really don't need "maxrecords" at all.  Revising fix...

progier389:
Is not this change already part of PR #7515 ?

mreynolds389:
> Is not this change already part of PR #7515 ?

Sort of, I stated in #7515 that I had to merge them otherwise the CI test would fail.  So I'll merge this, and rebase the other PR.  I wanted to keep these separate as they are technically two different issues, although very closely related.

progier389 (approved):
Ok, so I have better approve that one first ! 😉

mreynolds389:
> Ok, so I have better approve that one first ! 😉

Yeah sorry I should have done a better job explaining the order when I put these out for review.  Anyway thanks for the reviews!

Issue #7521, opened 2026-06-01
UI - make changes for cockpit API updates


mreynolds389:
**Issue Description**

Previously we used cockpit.spawn superuser argument as:

    cockpit.spawn(cmd, { superuser: true, err: "message" })

But the API changed and now instead of "true" the value should be "require":

    cockpit.spawn(cmd, { superuser: "require", err: "message" })

mreynolds389:
39b2f419a..61b3b1c7f  389-ds-base-3.2 -> 389-ds-base-3.2
a52c7f4c3..44a96188c  389-ds-base-3.1 -> 389-ds-base-3.1
647f49042..b55cc4e27  389-ds-base-3.0 -> 389-ds-base-3.0

PR #7522, opened 2026-06-01
Issue 7521 - UI - make changes for cockpit API updates

mreynolds389:
Description:

Previously we used cockpit.spawn superuser argument as:

    cockpit.spawn(cmd, { superuser: true, err: "message" })

But the API changed and now instead of "true" the value should be "require":

    cockpit.spawn(cmd, { superuser: "require", err: "message" })

relates: https://github.com/389ds/389-ds-base/issues/7521

Authored-by: Cursor


droideck (approved):
LGTM!

Issue #7519, opened 2026-05-29
Ignore obsolete entrydn index when entryrdn is enabled


jchapma:
**Issue Description**
Upgraded instances can have a leftover cn=entrydn index configuration from pre-entryrdn deployments while nsslapd-subtree-rename-switch (entryrdn) is on. The server may still attempt to use or reindex entrydn, causing inconsistent search results, partially unindexed filters, and index task failures on BDB).

**Package Version and Platform:** 
 - Package and version: [e.g. <= 389-ds-base-2.8

**Steps to Reproduce**
1. Inject cn=entrydn into dse.ldif on an entryrdn-enabled backend
2. Restart
3. Run dsconf backend index reindex --attr entrydn
4. Error symptoms include:
5. LDAP filters on normal attributes return no entries even though data exists
6. Subtree or base scoped searches that depend on DN resolution behave incorrectlly
7. Slow queries on consumers
8. Reindexing entrydn fails with:
```
WARN  bdb_db2index - <backend>: Requested to index entrydn, but nsslapd-subtree-rename-switch is on
ERR   task_index_thread - Index failed (error -1 
```

jchapma:
9cab3383d..539cad435  389-ds-base-2.8 -> 389-ds-base-2.8
9cab3383d..539cad435  389-ds-base-2.9 -> 389-ds-base-2.9
45371051f..85b0be579  389-ds-base-2.7 -> 389-ds-base-2.7
99ba295fa..fabbeeb2b  389-ds-base-2.6 -> 389-ds-base-2.6

PR #7526, opened 2026-06-02
Issue 7519 - Ignore obsolete entrydn index when entryrdn is enabled

jchapma:
Description:
Upgraded instances can retain a leftover cn=entrydn index configuration from pre-entryrdn deployments while entryrdn is on. The server may still attempt to register or reindex entrydn, causing inconsistent search results, partially unindexed filters, and index task failures on BDB.

Fix:
When entryrdn is on, the server ignores a leftover entrydn index, logs a warning, skips registering the index on config add/modify (the DSE entry can remain), and skips entrydn during reindex. Healthcheck DSBLE0008 flags leftover cn=entrydn config or entrydn.db* and documents steps to cleanup.

Fixes: https://github.com/389ds/389-ds-base/issues/7519

Reviewed by:

## Summary by Sourcery

Detect and ignore obsolete entrydn indexes when entryrdn-based indexing is enabled, and surface this misconfiguration via healthcheck.

Bug Fixes:
- Prevent entrydn from being registered or reindexed when entryrdn is enabled to avoid inconsistent search results and index task failures on BDB and MDB backends.

Enhancements:
- Add a healthcheck lint rule and backend lint logic to detect leftover entrydn index configuration or on-disk entrydn.db files and provide guided remediation steps.
- Improve lint report generation to avoid mutating shared templates when reporting backend implementation issues.

Tests:
- Add healthcheck tests verifying that an obsolete entrydn index triggers DSBLE0008 and that the warning clears after the configuration is removed.

### DIFF
--- a/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
+++ b/dirsrvtests/tests/suites/healthcheck/health_system_indexes_test.py
@@ -1143,6 +1143,47 @@ def test_index_check_fixes_multiple_issues(topology_st):
     standalone.start()
 
 
+
+def test_ignore_entrydn_index(topology_st, log_buffering_enabled):
+    """Healthcheck reports DSBLE0008 when entrydn index is configured.
+
+    :id: 65abefe0-2e6b-45e0-8604-fb33d734e412
+    :setup: Standalone instance
+    :steps:
+        1. Add obsolete entrydn index to userroot
+        2. Run healthcheck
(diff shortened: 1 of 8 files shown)


progier389:
Sourcery A/I reports about breaking the loop is a false positive:
 The break exits out of the switch statment and not out of the for loop

progier389 (approved):
LGTM

tbordaz (approved):
LGTM

PR #7540, opened 2026-06-03
Issue 7539 - Server shutdown during online reindex may lead to data loss

### LINKED ISSUE #7539: Server shutdown during online reindex may lead to data loss
**Issue Description**
When an online reindex task is aborted, for example by shutting down the server with SIGTERM while the reindex is in progress, the database data is lost, id2entry contains only 1 entry - RUV tombstone.

2.6.x+ 

**Steps to Reproduce**
1. Create an instance with MDB backend, import large number of users (500k)
2. Enable replication
3. Do online reindex
4. Stop the server (send SIGTERM)
5. dbscan on id2entry shows only 1 entry - RUV tombstone with ID 1 -> complete data loss.

**Expected results**
After server restart all data is present, dbscan on id2entry should show all 500k entries.

vashirov:
Bug Description:

When an online reindex task is aborted, for example by shutting down the server with SIGTERM while the reindex is in progress, the database data is lost, id2entry contains only 1 entry - RUV tombstone. Additionally, a SIGSEGV in `dbmdb_public_db_op` is observed when replication plugin tries to save the RUV state.

Fix Description:
When a reindex fails, don't close or delete the db. Instead, bring the backend back online so the server can shut down cleanly. Log a CRIT message advising the admin that indexes may be incomplete and reindex should be re-run.

Fixes: https://github.com/389ds/389-ds-base/issues/7539

## Summary by Sourcery

Handle failed online reindex without deleting or closing the database instance and ensure the backend can return online cleanly, while validating behavior with a regression test.

Bug Fixes:
- Preserve id2entry and prevent data loss when an online reindex task fails or is aborted, such as during server shutdown, by not deleting the database instance for reindex operations.
- Avoid crashes during shutdown after reindex failures by keeping the backend available so replication and other components can complete cleanly.

Enhancements:
- Log a critical message when reindex fails to warn administrators that indexes may be incomplete and need to be rebuilt.

Tests:
- Add an LMDB-specific regression test that populates a large dataset, aborts an online reindex, verifies id2entry contents are preserved, and checks for the expected critical log message.

### DIFF
--- a/dirsrvtests/tests/suites/indexes/reindex_abort_test.py
+++ b/dirsrvtests/tests/suites/indexes/reindex_abort_test.py
@@ -1,0 +1,154 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import os
+
+import pytest
+
+from lib389._constants import DEFAULT_BENAME, DEFAULT_SUFFIX
+from lib389.cli_ctl.dblib import get_mdb_dbis
+from lib389.dbgen import dbgen_users
+from lib389.dirsrv_log import DirsrvErrorLog
+from lib389.properties import TASK_WAIT
+from lib389.replica import Replicas
+from lib389.tasks import Tasks
+from lib389.utils import get_default_db_lib
+from test389.topologies import topology_st as topo
+
+pytestmark = pytest.mark.tier2
+
+logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+NUM_USERS = 500000
+
+REINDEX_ATTRS = [
+    'uid', 'cn', 'sn', 'givenName', 'mail',
+    'telephoneNumber', 'objectclass',
+    'entryrdn', 'parentid', 'numsubordinates',
+    'ancestorid',
+]
+
+
+def _get_mdb_entry_count(inst, bename, dbiname):
+    """Get entry count for a specific DBI via dbscan.  Instance must be stopped."""
+    dbhome = inst.ds_paths.db_dir
+    dbis = get_mdb_dbis(dbhome)
+    be = bename.lower()
+    if be in dbis and dbiname in dbis[be]:
+        return int(dbis[be][dbiname]['nbentries'])
+    return -1
+
+
+@pytest.fixture(scope="function")
+def populated_instance(topo, request):
+    """Populate the default backend with 500k users and enable replication."""
+    inst = topo.standalone
+
+    # Generate and import users
+    import_ldif = os.path.join(inst.ds_paths.ldif_dir, 'reindex_abort_test.ldif')
+    dbgen_users(inst, NUM_USERS, import_ldif, DEFAULT_SUFFIX,
+                generic=True, parent=f'ou=people,{DEFAULT_SUFFIX}')
+
+    tasks = Tasks(inst)
+    assert tasks.importLDIF(
+        suffix=DEFAULT_SUFFIX,
+        input_file=import_ldif,
+        args={TASK_WAIT: True}
+    ) == 0
+
+    # Enable replication
+    replicas = Replicas(inst)
+    replicas.create(properties={
+        'cn': 'replica',
+        'nsDS5ReplicaRoot': DEFAULT_SUFFIX,
+        'nsDS5ReplicaId': '1',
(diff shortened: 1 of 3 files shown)


dirsrvtests/tests/suites/indexes/reindex_abort_test.py
```diff
@@ -0,0 +1,154 @@
...
+
+REINDEX_ATTRS = [
+    'uid', 'cn', 'sn', 'givenName', 'mail',
+    'telephoneNumber', 'objectclass',
+    'entryrdn', 'parentid', 'numsubordinates',
+    'ancestorid',
+]
+
+
+def _get_mdb_entry_count(inst, bename, dbiname):
```
tbordaz, line 39: get_mdb_dbis only works on 'mdb' , we should call this function only if database is mdb

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
```diff
@@ -2428,11 +2428,24 @@ bdb_public_bdb_import_main(void *arg)
...
+             * which is never modified during reindex. On failure we
+             * must NOT close or delete the instance, just bring the
+             * backend back online so the server can continue operating
+             * or shut down cleanly.
+             */
+            import_log_notice(job, SLAPI_LOG_CRIT, "bdb_public_bdb_import_main",
+                              "Reindex failed. Indexes may be incomplete."
+                              " Search results may be empty or missing entries"
+                              " until reindex is re-run.");
+            instance_set_not_busy(job->inst);
```
tbordaz: The reindex failed. We do not know exactly the status of the indexes. If we allow the use of the backend (not busy) while indexes are broken I am afraid it can go into unexpected paths.
Should not we better keep the backend busy until a successful reindex ?

tbordaz (commented):
Looks good. A question

dirsrvtests/tests/suites/indexes/reindex_abort_test.py
vashirov: It's a helper function used by the test, which is skipped on BDB.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_import.c
vashirov: Good point, it's safer than keeping it running in an inconsistent state.

dirsrvtests/tests/suites/indexes/reindex_abort_test.py
tbordaz: I meant the two tests using it, should not they be 'skipif bdb' ?

vashirov: There is only one test and it already guards it:
```
@pytest.mark.skipif(get_default_db_lib() != "mdb",
                      reason="LMDB-specific data loss bug")
```
What is the other test?

tbordaz: Humm.. I need a second pair of eyes :(
Forget my comment

tbordaz (approved):
LGTM

dirsrvtests/tests/suites/indexes/reindex_abort_test.py
```diff
@@ -0,0 +1,154 @@
...
+            replica.delete()
+        except Exception:
+            pass
+
+    request.addfinalizer(fin)
+    return inst, baseline_count
+
+
+@pytest.mark.skipif(get_default_db_lib() != "mdb",
+                    reason="LMDB-specific data loss bug")
```
progier389, line 99: A bit curious about why the test is limited to LMDB while the code in fixed the same way both in LMDB and BDB.
Is there really no data loss in bdb case ? (or was the limitation due to the fact it is harder to get  the entry count with BDB ?)

vashirov: BDB online reindex uses `bdb_db2index` and never calls `bdb_public_bdb_import_main`. While LMDB uses `dbmdb_db2index` -> `dbmdb_back_ldif2db` -> `dbmdb_public_dbmdb_import_main`. I updated BDB code for symmetry.

tbordaz: @progier389 my understanding is that we are in the same trouble BDB/LMDB but the testcase itself rely on a function that is only available in LMDB

progier389: @thierry, Wrong guess! :wink: 
  That was also what I was thinking but Victor answered and it is just the opposite:
  the problem does not occur on bdb because the way the function is called  is different.

progier389 (approved):
LGTM

Issue #7554, opened 2026-06-03
deref plugin null pointer dereference if ber_init fails


mreynolds389:
**Issue Description**

When parsing a deref control if ber_init() fails then we dereference a null pointer and crash

mreynolds389:
ccc9ab295..f21bc8899  389-ds-base-3.2 -> 389-ds-base-3.2
53924b67e..f9aad2743  389-ds-base-3.1 -> 389-ds-base-3.1
59bfb2bf0..ce027249a  389-ds-base-3.0 -> 389-ds-base-3.0
2be7df76a..71cda0874  389-ds-base-2.9 -> 389-ds-base-2.9
f87702052..e960a4cc7  389-ds-base-2.8 -> 389-ds-base-2.8
81e726f20..0379301a7  389-ds-base-2.7 -> 389-ds-base-2.7
5f68ea121..29f9291e8  389-ds-base-2.6 -> 389-ds-base-2.6
246993a06..0ea404c1e  389-ds-base-2.4 -> 389-ds-base-2.4
8af45d281..f53e120d4  389-ds-base-2.2 -> 389-ds-base-2.2
e4e710a31..fb4cc80b2  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7555, opened 2026-06-03
Issue 7554 - deref plugin null pointer dereference if ber_init fails

mreynolds389:
Description:

**CWE**: CWE-476 (NULL Pointer Dereference)

A flaw in the 389 Directory Server's dereference control plugin allows an unauthenticated attacker to crash the LDAP server when the system is under memory pressure(OOM). The deref plugin, enabled by default, fails to check for a memory allocation failure before using the result, causing the server process to terminate.

CI test 'test_deref_and_access_control' already covers this code.

relates: https://github.com/389ds/389-ds-base/issues/7554

## Summary by Sourcery

Bug Fixes:
- Prevent null pointer dereference in the dereference control plugin when BER initialization fails.

### DIFF
--- a/ldap/servers/plugins/deref/deref.c
+++ b/ldap/servers/plugins/deref/deref.c
@@ -357,6 +357,12 @@ deref_parse_ctrl_value(DerefSpecList *speclist, const struct berval *ctrlbv, int
     }
 
     ber = ber_init((struct berval *)ctrlbv);
+    if (!ber) {
+        *ldapcode = LDAP_UNWILLING_TO_PERFORM;
+        *ldaperrtext = "Deref control parsing failed to initialize BER element";
+        return;
+    }
+
     for (tag = ber_first_element(ber, &len, &last);
          (tag != LBER_ERROR) && (tag != LBER_END_OF_SEQORSET);
          tag = ber_next_element(ber, &len, last)) {


ldap/servers/plugins/deref/deref.c
```diff
@@ -357,6 +357,12 @@ deref_parse_ctrl_value(DerefSpecList *speclist, const struct berval *ctrlbv, int
     }
 
     ber = ber_init((struct berval *)ctrlbv);
+    if (!ber) {
+        *ldapcode = LDAP_NO_MEMORY;
```
tbordaz: It is a result code (not error code) would LDAP_OPERATIONS_ERROR be acceptable ?

tbordaz (approved):
LGTM

ldap/servers/plugins/deref/deref.c
mreynolds389: > It is a result code (not error code) would LDAP_OPERATIONS_ERROR be acceptable ?

Hahaha, that is what I first used, and Sorcery yelled at me saying there were other more appropriate error codes.  I'll return it operations error :-)

progier389: IMHO a better code is LDAP_UNWILLING_TO_PERFORM

mreynolds389: > IMHO a better code is LDAP_UNWILLING_TO_PERFORM

For OOM?  There is nothing wrong with the control we are just out of memory.  I do NOT at all have a strong opinion on this, but I just want everyone to agree on it :-)  

Honestly if we are OOM we will probably fail in many other places before this code is even hit.

progier389: > > IMHO a better code is LDAP_UNWILLING_TO_PERFORM
> 
> For OOM? There is nothing wrong with the control we are just out of memory. I do NOT at all have a strong opinion on this, but I just want everyone to agree on it :-)
> 
> Honestly if we are OOM we will probably fail in many other places before this code is even hit.

I agree: there is a fair chance that the server exited because slapi_ch_*alloc hit OOM 

But LDAP_NO_MEMORY in not an ldap result code,
and LDAP_OPERATIONS_ERROR in inappropriate ( cf https://ldapwiki.com/wiki/Wiki.jsp?page=LDAP_OPERATIONS_ERROR ) (even if we wrongly use it in 389ds). That is probably why Sourcery complained! :wink: 

So IMHO we should use LDAP_UNWILLING_TO_PERFORM which is the right code for generic errors

mreynolds389: > So IMHO we should use LDAP_UNWILLING_TO_PERFORM which is the right code for generic errors

Works for me, thanks for the review everyeone!

PR #7559, opened 2026-06-04
Issue 7558 - During online import, the IDL should be created with in-…

### LINKED ISSUE #7558: During online import, the IDL should be created with in-depth first approach
**Issue Description**
The online initialization requires that the supplier builds a sorted IDL. Sorted in the way that the parent entry appears before the children.
The way it is currently done does scale in all condition. It scales if ID are mostly consecutive/increasing/without holes. Else it does not scale and can lead to the failure of online init.

all versions since 1.3

**Steps to Reproduce**
No clear reproducer.
It happens when the IDs contains holes (that creates multiple ranges). Are not consecutive/increasing so finding the appropriate range in the list does not obey to any sorting. 

**Expected results**
The online initialisation should start more rapidly to prevent timeout

tbordaz:
…depth first approach

Bug description:
	The online initialization requires that the supplier builds a sorted IDL.
	It is sorted in the way that the parent entry appears before the children in the IDL.
	The current implementation goes through the parentid index from the first entry
	until the end (next). To make sure parent entry is already in IDL before adding a child
	it uses a list of ID ranges.
	This list works well if the ID are mostly consecutive and with limited number of holes,
	else the list grows (lot of singleton) and checking the list becomes costly as well.

Fix description:
	Instead of walking the parentid index from the first entry to the end it walks
	the parentid in depth first. So there is no need to check that the parent ID is already
	present in IDL

fixes: #7558

Reviewed by:

## Summary by Sourcery

Introduce a depth-first walk of the parentid index for bulk import to ensure parents precede children in ID lists and remove the previous leftover/range-based handling.

Bug Fixes:
- Ensure online/bulk import builds parentid-based ID lists with ancestors before descendants without relying on sparse ID range tracking, preventing performance degradation with non-consecutive IDs.

Enhancements:
- Add a dedicated depth-first parentid index traversal path for SLAPI_OP_RANGE_NO_IDL_SORT that builds candidate ID lists via direct key lookups, with limit enforcement and improved logging.
- Simplify range fetch and LMDB range fetch logic by removing leftover parent tracking, range helpers, and special-case sorting, always sorting final ID lists when not allids.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/idl_new.c
+++ b/ldap/servers/slapd/back-ldbm/idl_new.c
@@ -45,13 +45,6 @@ struct idl_private
     int dummy;
 };
 
-/* Used to store leftover parentid and entry ids */
-typedef struct _range_id_pair
-{
-    ID key;
-    ID id;
-} idl_range_id_pair;
-
 /* lmdb iterator callback context */
 typedef struct {
     backend *be;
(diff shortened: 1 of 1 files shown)


progier389 (approved):
LGTM.
Although this algorythm is slightly slower than the old one when the ID are rightly ordered, the time to get next entry is more reliable (it does not have the heavy coost of reodering the idlist with old database having higly disordered IDs) and makes the total init more robust by avoiding hitting the non response timeout

tbordaz:
Backport in 1.4.3
1a61a0ecc..05fb3b901  389-ds-base-1.4.3

PR #7568, opened 2026-06-08
Issue 7567 - Content Sync plugin unbounded queue growth and race cond…

### LINKED ISSUE #7567: Content Sync plugin unbounded queue growth and race conditions
**Issue Description**
The queue of each syncrepl persistent search is not limited. To prevent high memory consumption the number of entries in the queue should have a configurable maximum size.
There is a race condition in case of refresh error and sending entries function where the server can drop a request that is not yet processed.
Also for hardening parallel access move the read/write of 'thread_count' and 'plugin_closing' to atomic access.

 all versions

**Steps to Reproduce**
attached testcase

**Expected results**
Growth of the memory should be limited

tbordaz:
…itions

Bug description:
	Main issue is that the queue of each syncrepl persistent search is not limited.
	Also fixing some race condition cases

Fix description:
	Support of two new configuration attributes for the sync repl configuration
	entry (cn=Content Synchronization,cn=plugins,cn=config)
	- syncrepl-max-concurrent (default 10) that was previously configured with nsslapd-pluginarg0
	- syncrepl_queue_max-size (default 10000) that is the maximum size of the queue of the updated entries Also changing the access to 'thread_count' and 'plugin_closing'

fixes: #7567

Reviewed by:

## Summary by Sourcery

Limit and monitor content sync persistent search queues and harden thread lifecycle handling to prevent unbounded growth and race conditions.

Bug Fixes:
- Drop incoming sync entries when a per-request queue exceeds a configurable maximum size.
- Ensure sync request cleanup and thread counters are updated atomically to avoid race conditions during plugin shutdown and request termination.

Enhancements:
- Allow configuring maximum concurrent sync persistent searches and per-request queue size via dedicated plugin configuration attributes instead of plugin arguments.

### DIFF
--- a/dirsrvtests/tests/suites/upgrade/upgrade_plugin_attribute.py
+++ b/dirsrvtests/tests/suites/upgrade/upgrade_plugin_attribute.py
@@ -1,0 +1,64 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+from test389.topologies import topology_st as topo
+from lib389.dseldif import DSEldif
+
+log = logging.getLogger(__name__)
+
+
+def test_sync_plugin(topo):
+    """Test that the replication plugin name is updated to the new name at
+    server startup.
+
+    :id: c275f870-aa32-49bf-bfd3-d18b9c3ae250
+    :setup: Standalone Instance
+    :steps:
+        1. Stop Server
+        2. add nsslapd-pluginarg0 in content sync config entry
+        3. Start server
+        4. Verify nsslapd-pluginarg0 is removed
+        5. Verify syncrepl-max-concurrent is added
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Success
+    """
+
+    SYNC_PLUGIN_DN = "cn=Content Synchronization,cn=plugins,cn=config"
+    REPL_PLUGIN_NAME = "Multisupplier Replication Plugin"
+    SYNC_ATTR_OLD_NAME = "nsslapd-pluginarg0"
+    SYNC_ATTR_NEW_NAME = "syncrepl-max-concurrent"
+    SYNC_ATTR_VAL = "5"
+
+    # Stop the server
+    topo.standalone.stop()
+
+    # Edit sync plugin in dse.ldif
+    dse_ldif = DSEldif(topo.standalone)
+    dse_ldif.replace(SYNC_PLUGIN_DN, SYNC_ATTR_OLD_NAME, SYNC_ATTR_VAL)
+
+    topo.standalone.restart()
+    dse_ldif = DSEldif(topo.standalone)
+    pluginarg0 = dse_ldif.get(SYNC_PLUGIN_DN, SYNC_ATTR_OLD_NAME, single=True)
+    assert pluginarg0 == None
+    max_concurrent = dse_ldif.get(SYNC_PLUGIN_DN, SYNC_ATTR_NEW_NAME, single=True)
+    assert int(max_concurrent) == int(SYNC_ATTR_VAL)
+
+
+if __name__ == '__main__':
+    # Run isolated
+    # -s for DEBUG mode
+    CURRENT_FILE = os.path.realpath(__file__)
+    pytest.main(["-s", CURRENT_FILE])
+
--- a/ldap/servers/plugins/sync/sync_persist.c
+++ b/ldap/servers/plugins/sync/sync_persist.c
@@ -28,7 +28,7 @@ static SyncRequestList *sync_request_list = NULL;
  */
 #define SYNC_IS_INITIALIZED() (sync_request_list != NULL)
 
-static int plugin_closing = 0;
+static PRUint64 plugin_closing = 0;
 static PRUint64 thread_count = 0;
 static int sync_add_request(SyncRequest *req);
 static void sync_remove_request(SyncRequest *req);
@@ -595,8 +595,16 @@ sync_queue_change(OPERATION_PL_CTX_T *operation)
             }
             /* Put it on the end of the list for this sync search */
             PR_Lock(req->req_lock);
+            /* check if the queue max size is reached */
+            if (req->req_queue_count >= req->req_queue_max_size) {
+                slapi_log_err(SLAPI_LOG_WARNING, SYNC_PLUGIN_SUBSYSTEM, "sync_queue_change - queue max size reached, dropping entry \"%s\"\n", slapi_entry_get_dn_const(node->sync_entry));
+                PR_Unlock(req->req_lock);
+                sync_node_free(&node);
+                continue;
+            }
             pOldtail = req->ps_eq_tail;
             req->ps_eq_tail = node;
+            req->req_queue_count++;
             if (NULL == req->ps_eq_head) {
                 req->ps_eq_head = req->ps_eq_tail;
             } else {
@@ -631,7 +639,7 @@ sync_queue_change(OPERATION_PL_CTX_T *operation)
  * of established content sync persistent requests
  */
 int
-sync_persist_initialize(int argc, char **argv)
+sync_persist_initialize(int argc, char **argv, Slapi_Entry *config_entry)
 {
     if (!SYNC_IS_INITIALIZED()) {
         pthread_condattr_t sync_req_condAttr; /* cond var attribute */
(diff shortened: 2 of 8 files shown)


ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -777,16 +810,14 @@ sync_persist_terminate(PRThread *tid)
             if (cur->req_tid == tid) {
                 cur->req_active = PR_FALSE;
                 cur->req_complete = PR_TRUE;
+                sync_remove_request(cur);
```
droideck: Do I understand correctly that we call sync_remove_request(cur) while still holding SYNC_LOCK_READ()? And sync_remove_request() takes SYNC_LOCK_WRITE() hence a possible deadlock?

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -670,14 +678,39 @@ sync_persist_initialize(int argc, char **argv)
 
         sync_request_list->sync_req_head = NULL;
         sync_request_list->sync_req_cur_persist = 0;
-        sync_request_list->sync_req_max_persist = SYNC_MAX_CONCURRENT;
+        sync_request_list->sync_req_max_persist = SYNC_DEFAULT_MAX_CONCURRENT;
+        sync_request_list->sync_req_queue_max_size = SYNC_DEFAULT_QUEUE_MAX_SIZE;
         if (argc > 0) {
```
droideck, line 684: It looks like the new config-entry attributes are skipped whenever argc > 0. Is it the intention that we make syncrepl-queue-max-size ineffective on instances that still have legacy nsslapd-pluginarg0 configured?

src/lib389/lib389/cli_conf/plugins/contentsync.py
```diff
@@ -29,6 +37,8 @@ def contentsync_add(inst, basedn, log, args):
 def _add_parser_args(parser):
     parser.add_argument('--allow-openldap', choices=['on', 'off'], type=str.lower,
                         help='Allows openldap servers to act as read only consumers of this server via syncrepl')
+    parser.add_argument('--queue-max-size', default=10000, type=check_queue_size,
```
droideck: Unless I miss something, I don't think we should assign the default value here as it will overwrite existing values if the option is not specified.

src/lib389/lib389/cli_conf/plugins/contentsync.py
```diff
@@ -12,8 +12,16 @@
...
     'allow_openldap': 'syncrepl-allow-openldap',
+    'queue_max_size': 'syncrepl-queue-max-size',
 }
 
+def check_queue_size(value):
+    ivalue = int(value)
+    if 100 <= ivalue <= 100000:
+        return ivalue
+    else:
+        raise argparse.ArgumentTypeError(f"{value} should be in the range [100, 100000]")
```
droideck: Just to be on a safe side, I think, we need to import argparse.

droideck (changes requested):

ldap/servers/plugins/sync/sync_persist.c
tbordaz: If we are using legacy nsslapd-pluginarg0 (max_concurrent) then sync_request_list->sync_req_queue_max_size stays with the default queue_max_size (10K). So the queue size is bound.
I think it is acceptable to ignore mixed configuration tuning (argc and attr_name).
Now what we can do is to upgrade the tuning so that it replaces nsslapd-pluginarg0 with syncrepl-max-concurrent. WDYT ?

ldap/servers/plugins/sync/sync_persist.c
tbordaz: Good catch. Acquiring write lock while holding it in read leads to undefined behavior.
I will update the fix

src/lib389/lib389/cli_conf/plugins/contentsync.py
tbordaz: that is correct

src/lib389/lib389/cli_conf/plugins/contentsync.py
tbordaz: done

ldap/servers/plugins/sync/sync_persist.c
tbordaz: done

src/lib389/lib389/cli_conf/plugins/contentsync.py
```diff
@@ -6,14 +6,23 @@
...
```
droideck: You probably need to change the error type `argparse.ArgumentTypeError` -> `ArgumentTypeError`

src/lib389/lib389/cli_conf/plugins/contentsync.py
```diff
@@ -6,14 +6,23 @@
...
 
 arg_to_attr = {
     'allow_openldap': 'syncrepl-allow-openldap',
+    'queue_max_size': 'syncrepl-queue-max-size',
 }
 
+def check_queue_size(value):
+    ivalue = int(value)
+    if 100 <= ivalue <= 100000:
+        return ivalue
```
droideck: The generic dsconf edit path eventually passes this value into lib389 LDAP modify helpers, which expect string/bytes attribute values. With the current return value, `dsconf plugin contentsync set --queue-max-size 5000` fails with `"'int' object has no attribute 'encode'"`.
`str(ivalue)` will solve the issue:)

src/lib389/lib389/cli_conf/plugins/contentsync.py
```diff
@@ -29,6 +38,8 @@ def contentsync_add(inst, basedn, log, args):
 def _add_parser_args(parser):
     parser.add_argument('--allow-openldap', choices=['on', 'off'], type=str.lower,
                         help='Allows openldap servers to act as read only consumers of this server via syncrepl')
+    parser.add_argument('--queue-max-size', type=check_queue_size,
```
droideck, line 42: The commit adds/migrates `syncrepl-max-concurrent`, but dsconf does not expose it here. Should we add `max_concurrent: syncrepl-max-concurrent` to `arg_to_attr` and a `--max-concurrent` parser option?

ldap/servers/plugins/sync/sync_persist.c
droideck: Sure, fair point that the queue remains bounded at the 10K default!
The issue I'm pointing at is not unboundedness though, but that a legacy max-concurrent arg masks an independent queue-size attr because of the `else if`. I agree migrating `nsslapd-pluginarg0` to `syncrepl-max-concurrent` is a right thing to do...
That said, should we parse queue size independently from `config_entry` so mixed/first-upgrade startup honors explicit queue tuning?

droideck (changes requested):
While reviewing the PR, I created a couple of tests. Do you think you  can reuse some of them for the PR? these two for a new `clu` test suite:

```python
def _dsconf_contentsync(topo, *args):
    cmd = [
        '/usr/sbin/dsconf', topo.standalone.serverid,
        '-j', '-D', DN_DM, '-w', 'password',
        'plugin', 'contentsync',
        *args,
    ]
    return subprocess.run(cmd, text=True, capture_output=True)


def test_contentsync_queue_max_size_cli(topo):
    """Verify dsconf accepts and stores Content Sync queue max size

    :id: fceca807-bf01-466f-8a36-c65c4d2d18cd
    :setup: Standalone instance
    :steps:
        1. Enable Content Sync
        2. Set queue max size with dsconf
        3. Show plugin config
        4. Try an out-of-range queue size
    :expectedresults:
        1. Success
        2. Success
        3. syncrepl-queue-max-size is set as a string value
        4. dsconf rejects the invalid value cleanly
    """
    plugin = ContentSyncPlugin(topo.standalone)
    plugin.enable()

    result = _dsconf_contentsync(topo, 'set', '--queue-max-size', '5000')
    assert result.returncode == 0, result.stderr

    result = _dsconf_contentsync(topo, 'show')
    assert result.returncode == 0, result.stderr
    attrs = json.loads(result.stdout)['attrs']
    assert attrs['syncrepl-queue-max-size'] == ['5000']

    result = _dsconf_contentsync(topo, 'set', '--queue-max-size', '10')
    assert result.returncode != 0
    assert 'Traceback' not in result.stderr
    assert 'range' in result.stderr.lower()
```

possibly:

```python
def test_contentsync_max_concurrent_cli(topo):
    """Verify dsconf exposes the renamed max-concurrent setting

    :id: 7fbef7bf-3ac8-4925-a5aa-cd873c6f8e81
    :setup: Standalone instance
    :steps:
        1. Enable Content Sync
        2. Set max concurrent persistent searches with dsconf
        3. Show plugin config
    :expectedresults:
        1. Success
        2. Success
        3. syncrepl-max-concurrent is set
    """
    plugin = ContentSyncPlugin(topo.standalone)
    plugin.enable()

    result = _dsconf_contentsync(topo, 'set', '--max-concurrent', '12')
    assert result.returncode == 0, result.stderr

    result = _dsconf_contentsync(topo, 'show')
    assert result.returncode == 0, result.stderr
    attrs = json.loads(result.stdout)['attrs']
    assert attrs['syncrepl-max-concurrent'] == ['12']
```

And for the syncrepl_plugin basic:

```python
def test_syncrepl_queue_size_with_legacy_pluginarg(topology):
    """Verify queue size config is honored when legacy pluginarg0 is present

    :id: 83545ba4-782e-4326-9057-a90496ee82ad
    :setup: Standalone instance
    :steps:
        1. Enable Retro Changelog and Content Sync
        2. Set legacy nsslapd-pluginarg0 and new syncrepl-queue-max-size
        3. Restart the server
        4. Check startup logs for the configured queue size
    :expectedresults:
        1. Success
        2. Success
        3. Success
        4. The configured queue size is applied
    """
    st = topology.standalone
    st.config.loglevel(vals=(ErrorLog.DEFAULT, ErrorLog.PLUGIN))

    rcl = RetroChangelogPlugin(st)
    rcl.enable()

    csp = ContentSyncPlugin(st)
    csp.enable()
    csp.replace('nsslapd-pluginarg0', '5')
    csp.replace('syncrepl-queue-max-size', '5000')

    st.restart()

    assert st.ds_error_log.match(
        '.*sync_persist_initialize - Queue max size is set to 5000.*',
        after_pattern='.*starting up.*'
    )
```

src/lib389/lib389/cli_conf/plugins/contentsync.py
tbordaz: done

src/lib389/lib389/cli_conf/plugins/contentsync.py
tbordaz: right. I changed it to support separately max-concurrent/argc and max-queue

tbordaz:
@droideck thank you so much for your reviews and help to improve fix/tests

dirsrvtests/tests/suites/upgrade/upgrade_plugin_attribute.py
```diff
@@ -0,0 +1,64 @@
...
+import logging
+import pytest
+import os
+from test389.topologies import topology_st as topo
+from lib389.dseldif import DSEldif
+
+log = logging.getLogger(__name__)
+
+
+def test_sync_plugin(topo):
```
droideck, line 18: This new test is not collected by CI. It should be named like `*_test.py` and ideally we need to put it into one of the tiers (likely pytestmark = pytest.mark.tier1?)

ldap/servers/plugins/sync/sync_persist.c
```diff
@@ -595,8 +595,16 @@ sync_queue_change(OPERATION_PL_CTX_T *operation)
             }
             /* Put it on the end of the list for this sync search */
             PR_Lock(req->req_lock);
+            /* check if the queue max size is reached */
+            if (req->req_queue_count >= req->req_queue_max_size) {
```
droideck, line 599: Possible issue: when the queue is full, we drop this update but keep the persistent search alive. Is it okay for the session to continue at that point, or should we terminate/force a refresh so the client cannot miss changes silently?

droideck (approved):
Besides a couple of minor issues/a question, LGTM!

mreynolds389:
Any chance this new ticket https://github.com/389ds/389-ds-base/issues/7601 could be related to this?

tbordaz:
> Any chance this new ticket #7601 could be related to this?

I do not think so. #7601 is related to the spawning of a new persistent search thread while preparing the arguments for the it. So #7601 occurs before the persistent search thread is started. This ticket is about limiting the queue size with with a running persistent search thread.

PR #7572, opened 2026-06-09
Issue 7530 - CI - Stabilize DNA plugin replication tests timing out in CI

### LINKED ISSUE #7530: CI - Stabilize DNA plugin replication tests timing out in CI
## Problem

`plugins/dna_interval_test.py` is repeatedly erroring in both BDB and LMDB CI jobs. The failures appear tied to replication teardown/setup timing out.

Evidence:
- BDB plugins job: `78989710167`
- LMDB plugins job: `78996569550`

Observed failing tests:
- `dirsrvtests/tests/suites/plugins/dna_interval_test.py::test_dna_interval`
- `dirsrvtests/tests/suites/plugins/dna_interval_test.py::test_dna_max_value`
- `dirsrvtests/tests/suites/plugins/dna_interval_test.py::test_dna_interval_with_different_values[0]`
- `dirsrvtests/tests/suites/plugins/dna_interval_test.py::test_dna_interval_with_different_values[abc]`
- `dirsrvtests/tests/suites/plugins/dna_interval_test.py::test_dna_interval_with_different_values[2000]`

Representative log signature:

`NSMMReplicationPlugin - release_replica - agmt="cn=201" ... unable to receive endReplication extended operation response from the replica. Error -5 (Timed out)`

BDB also showed:

`dirsrvtests/tests/suites/plugins/managed_entry_test.py::test_managed_entry_delete_after_reimport`

- The same DNA interval errors and `release_replica ... Error -5 (Timed out)` signature repeated across the five latest sampled BDB `plugins` jobs and the five latest sampled LMDB `plugins` jobs.
- BDB repeatedly also fails `managed_entry_test.py::test_managed_entry_delete_after_reimport` in the same `plugins` job.
- Matrix frequency check found `plugins` failing in 20/20 sampled failed-main runs for both BDB and LMDB, so this is confirmed recurring rather than a one-off sampled failure.

## Proposed work

Investigate replication topology setup/teardown in the DNA plugin tests and why `release_replica` times out in CI. Confirm whether the DNA tests leave replication state behind, race with instance startup/shutdown, or need stronger waits/cleanup.

## Acceptance criteria

- The DNA interval tests pass reliably in both BDB and LMDB jobs.
- Replication timeout logs are no longer emitted for this suite during normal CI execution.
- If a fixture/topology cleanup issue is found, add targeted regression coverage or cleanup assertions.

jchapma:
Description:
dna_interval_test.py failed in teardown with NameError for DSEldif, leaving standalone1 stopped and causing later tests in the module to fail with SERVER_DOWN. Add pause replication around accpol_test restarts to avoid release_replica timeouts, and acknowledge endReplication on the consumer when no session is active.

Fixes: https://github.com/389ds/389-ds-base/issues/7530

Reviewed by:

## Summary by Sourcery

Stabilize replication-related plugin tests by improving replication control and teardown robustness.

Bug Fixes:
- Ensure DNA plugin teardown always restarts the standalone instance even if DSEldif handling fails.
- Handle EndReplication extended operations with no active session as successful so suppliers can shut down cleanly without timeouts.

Enhancements:
- Pause and resume replication around account policy plugin supplier/consumer restarts to prevent release_replica timeouts and reduce CI flakiness in replication tests.

### DIFF
--- a/ldap/servers/plugins/replication/repl_extop.c
+++ b/ldap/servers/plugins/replication/repl_extop.c
@@ -1376,8 +1376,9 @@ multisupplier_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
             /* Outbound replication agreements need to all be restarted now */
             /* XXXGGOOD RESTART REEPL AGREEMENTS */
         } else {
-            /* Unless bail out, we return uninitialized response */
-            goto free_and_return;
+            /* No active replication session on this connection, treat as
+             * already released so the supplier can shut down cleanly */
+            response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED;
         }
     }
 send_response:
--- a/dirsrvtests/tests/suites/plugins/accpol_test.py
+++ b/dirsrvtests/tests/suites/plugins/accpol_test.py
@@ -1350,6 +1350,11 @@ def test_acct_policy_consumer(topology_m1c1, request):
         except ldap.NO_SUCH_OBJECT:
             time.sleep(1)
 
+    # Pause replication before restarting supplier/consumer so release_replica
+    # can complete cleanly
+    topology_m1c1.pause_all_replicas()
+    time.sleep(5)
+
     # Configure lastLoginTime on supplier and consumer
     plugin = AccountPolicyPlugin(supplier)
     plugin.enable()
@@ -1370,6 +1375,10 @@ def test_acct_policy_consumer(topology_m1c1, request):
     accp.set('alwaysrecordlogin', 'yes')
     accp.set('stateattrname', 'lastLoginTime')
     consumer.restart(timeout=10)
+
+    # Allow replication to resume before bind tests
+    topology_m1c1.resume_all_replicas()
+    time.sleep(2)
 
     # On supplier
     # Do 3 binds with a delay to
@@ -1413,6 +1422,8 @@ def test_acct_policy_consumer(topology_m1c1, request):
 
     def fin():
         user_supplier.delete()
+        topology_m1c1.pause_all_replicas()
+        time.sleep(5)
         log.info('Disabling Global accpolicy plugin and removing pwpolicy attrs')
         try:
             plugin = AccountPolicyPlugin(supplier)
--- a/dirsrvtests/tests/suites/plugins/dna_interval_test.py
+++ b/dirsrvtests/tests/suites/plugins/dna_interval_test.py
@@ -11,6 +11,7 @@
 import logging
 import pytest
 from lib389._constants import DEFAULT_SUFFIX
+from lib389.dseldif import DSEldif
 from lib389.plugins import DNAPlugin, DNAPluginConfigs
 from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccounts
@@ -45,10 +46,13 @@ def dna_plugin(topology_st, request):
     inst.restart()
 
     def fin():
-        inst.stop()
-        dse_ldif = DSEldif(inst)
-        dse_ldif.delete_dn(f'cn=dna config,{plugin.dn}')
-        inst.start()
+        try:
+            inst.stop()
+            dse_ldif = DSEldif(inst)
+            dse_ldif.delete_dn(f'cn=dna config,{plugin.dn}')
+        finally:
+            if not inst.status():
+                inst.start()
     request.addfinalizer(fin)
 
     return dna_config


ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1376,8 +1376,9 @@ multisupplier_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
             /* Outbound replication agreements need to all be restarted now */
             /* XXXGGOOD RESTART REEPL AGREEMENTS */
         } else {
-            /* Unless bail out, we return uninitialized response */
-            goto free_and_return;
+            /* No active replication session on this connection, treat as
+             * already released so the supplier can shut down cleanly */
+            response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED;
```
droideck: Should we split the `connext == NULL` case from the "no replica acquired" case? The `else` above covers both but `connext == NULL` can mean we failed to acquire the connection extension because another start/end op is using it.
IIUC, in that case, returning `REPLICA_RELEASE_SUCCEEDED` may tell the supplier the release completed when we never inspected or changed the session state... I'm not that familiar with the area though, so I may miss something. WDYT?

jchapma: > Should we split the `connext == NULL` case from the "no replica acquired" case? The `else` above covers both but `connext == NULL` can mean we failed to acquire the connection extension because another start/end op is using it. IIUC, in that case, returning `REPLICA_RELEASE_SUCCEEDED` may tell the supplier the release completed when we never inspected or changed the session state... I'm not that familiar with the area though, so I may miss something. WDYT?

Good catch, I will split those cases. We keep RELEASE_SUCCEEDED for only when we could read connext and confirm there's nothing to release, not when acquire failed because another  op holds the extension.

ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1375,9 +1375,33 @@ multisupplier_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
...
+                    PR_Unlock(peek->lock);
+                    slapi_log_err(SLAPI_LOG_REPL, repl_plugin_name,
+                                  "multisupplier_extop_EndNSDS50ReplicationRequest - "
+                                  "conn=%" PRIu64 " op=%d Could not acquire connext; "
+                                  "held by op=%d\n",
+                                  connid, opid, peek->in_use_opid);
+                    response = NSDS50_REPL_INTERNAL_ERROR;
+                } else {
+                    PR_Unlock(peek->lock);
+                    response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED;
```
droideck: I think this may still race. This branch is reached after `consumer_connection_extension_acquire_exclusive_access()` failed, so we do not own `connext`. If the other op releases the extension before this peek, `in_use_opid` can be `-1` and we return `RELEASE_SUCCEEDED` without reacquiring and checking whether `replica_acquired` still needs to be released. 

Do you think does it make sense to retry/acquire the extension before returning success here? And only return `RELEASE_SUCCEEDED` after we actually hold `connext` and have inspected the state?

jchapma: Now that I know more about this area, yes it does make sense to retry acquiring the connext before we return success. I have made updates so when the connext is free we  acquire it by setting while holding the lock, then inspect replica_acquired and perform the full cleanup if needed. Thanks

ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1375,9 +1375,49 @@ multisupplier_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
...
+                                  "held by op=%d\n",
+                                  connid, opid, peek->in_use_opid);
+                    response = NSDS50_REPL_INTERNAL_ERROR;
+                } else {
+                    /* Extension is free now - acquire it properly and check state */
+                    peek->in_use_opid = opid;
+                    retry = peek;
+                    PR_Unlock(peek->lock);
+
+                    if (retry->replica_acquired != NULL) {
```
droideck: Nice! So we're mostly good, I think. My earlier concern about reacquiring connext before returning success is no more.

One remaining question, though: in the `retry->replica_acquired != NULL` case, we now own connext, but this branch only relinquishes the replica and clears session flags. It skips the protocol-specific cleanup from the normal EndReplication path above.
I wonder is this should call the same cleanup path... Or maybe, should this retry branch only return RELEASE_SUCCEEDED when replica_acquired is already NULL?

ldap/servers/plugins/replication/repl_extop.c
```diff
@@ -1375,9 +1375,49 @@ multisupplier_extop_EndNSDS50ReplicationRequest(Slapi_PBlock *pb)
...
             /* XXXGGOOD RESTART REEPL AGREEMENTS */
+        } else if (NULL != connext) {
+            /* Have connext but no replica lock. Prior End already
+             * released it; return success without repeating teardown. */
+            response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED;
         } else {
-            /* Unless bail out, we return uninitialized response */
-            goto free_and_return;
+            /* Could not acquire connext. Retry if it becomes free,
+             * otherwise another End op holds it. */
```
tbordaz: I am a bit lost here as it looks complex patch regarding the current state.
My understanding is that here we are on the consumer side and we received an end of replication session extop. The 'else' path is that the supplier did not hold the replica so IMHO we do not need much clean up. Just making sure we are responding to the supplier with slapi_send_ldap_result.

In short I have the feeling that the 'goto free_and_return' is weird and I would just replace it with 'response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED' and fall through the send_response.

Would it be enough for stabilizing the test ?

jchapma: My first commit (44800b16d86564dfd06b5dd81d4a7342516094d9) did exactly this. The retry block came from later feedback, to handle  a possible race condition where acquire_exclusive_access fails but connext is free again by the time we peek at it. 

The initial CI failure is the supplier hanging in release_replica because we call goto free_and_return and never call  send_response. I think the init commit is enough to stabilise this test. I can revert the retry block and go back to simply returning RELEASE_SUCCEEDED and fall through to send_response when we are not in the normal connext && replica_aquired path.

tbordaz: Ah okay I missed this original commit.
What looks weird is to try to acquire the replica at the time the supplier requests to end the session so when the consumer is asked to do cleanup.
Releasing the replica looks good if it was acquired. But I do not get the benefit of trying to acquire then release it at clean up time.
Sorry for this late feedback.

I remember having seen many error replication logs where the supplier logs that it did not receive response to a ending request. This fix (in replication plugin) may fix this issue.

jchapma: Ok, everyone happy if i revert back to the simple response = NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED in a non normal case ??

progier389: After the discussion in the internal technical discussion meeting, I finally think that sending directly NSDS50_REPL_REPLICA_RELEASE_SUCCEEDED response is indeed the right thing to do.
Anyway if the replica is not acquired we have nothing else to do.

droideck: Yeah, I'm okay with reverting the retry block too.
My earlier concern was basically that `connext == NULL` and "we got `connext`, but there is no `replica_acquired`" are not quite the same thing. If the acquire failed because another Start/End op had the extension, I was worried that returning success meant we had not actually looked at the session state.
But after seeing the retry version, I think it goes too far for this fix...
If we fail to acquire `connext` normally, trying to peek/reacquire and do cleanup from this path gets awkward fast, and the current version can end up releasing the replica while skipping the normal EndReplication cleanup above.
The bug we need to fix here is simpler.  So I'm good with the initial approach.

tbordaz: Reading your point I believe I see your point. The replica_acquire and connext are almost independent, only replica_acquire is stored in connext.
Connext is a connection extension used by replication. So if we have a connection used for replication from a supplier to a consumer we usually/always (not sure) have an connection extension. THe connection can be opened (and usually it remains open for a very long time), even if the supplier did not acquire exclusive access to the replica on the consumer.
A common workflow is the supplier open a connection to a consumer (create the connext on the consumer side). Periodically the supplier starts a replication session (extop) and read the consumer ruv. If the supplier has no update to send (comparing the RUV) it just ends the replication session (another extop) without closing the connection. When it has updates to send, it tries to acquire the replica on the consumer. It can fail if for example the replica is already updated by another supplier. When the supplier acquire the replica, the connext->replica_acquire is set (on the consumer).
In short we usually have connext set but only short time replica_acquire set

droideck: Thanks, that helps!
Now it's clearer:) So... The connext can outlive an active replication session, so `connext != NULL && replica_acquired == NULL` is a valid "nothing to release" state, and returning RELEASE_SUCCEEDED there fixes the missing-response hang.

So the only small edge case, not proven by any reproducer, is when we fail to acquire connext because another Start/End op owns it. But I agree that's separate from this fix and better left out rather than adding the retry cleanup path, unless we get a clear reproducer showing a real problem there.

jchapma: Ok, I have reverted the 2 latest commits, thank you all for your feedback

droideck (approved):
LGTM!

tbordaz (approved):
LGTM

PR #7575, opened 2026-06-11
Issue 7531 - Fix LMDB replication regression_m2 failures and core dumps

### LINKED ISSUE #7531: CI - Fix LMDB replication regression_m2 failures and core dumps
## Problem

`LMDB Test (replication/regression_m2_test.py)` repeatedly fails in CI and uploads core files such as `assets/cores/core.worker-*`.

Evidence:
- Job `78996570615`
- Job `79173525705`

Observed failing cases include:
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_moving_entry_make_online_init_fail`
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_keepalive_entries`
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_change_repl_passwd[using-bind-group]`
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_change_repl_passwd[using-bind-dn]`
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_repl_after_reindex`
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_rid_starting_with_0`
- errors in `test_suffix_entryid` and `test_bulk_import`

Representative logs include:
- `NSMMReplicationPlugin - release_replica ... Error -1 (Can't contact LDAP server)`
- `agmt_new_from_entry - Warning invalid value for nsds5ReplicaEnabled ... Ignoring this repl agreement`
- `prot_notify_agmt_changed - Replication agreement ... could not be updated`
- uploaded core files like `core.worker-4...` and `core.worker-8...`

- Latest sampled main run https://github.com/389ds/389-ds-base/actions/runs/26797037399, job `78996570615`, still shows `ldap.SERVER_DOWN` against `supplier1`, repeated replication agreement errors, and uploaded `assets/cores/core.worker-4.4882.752`.
- Related older issue https://github.com/389ds/389-ds-base/issues/6535 covers misleading CI core-file detection. It is related background, not a duplicate: this issue tracks the current LMDB `regression_m2_test.py` failures/crashes, while #6535 tracks core detection/reporting behavior.

## Proposed work

Investigate whether the failures are LMDB-specific replication bugs, test ordering/state leakage, fixture instability, or expected-negative agreement tests leaving replicas unhealthy. Analyze uploaded core files from failing runs where available.

## Acceptance criteria

- `replication/regression_m2_test.py` passes reliably under LMDB CI.
- No core files are produced during successful runs.
- If a server crash is confirmed, include a targeted regression test or diagnostic notes in the fix.

jchapma:
Description:
During a nested rename, _entryrdn_append_childidl walks the entryrdn index to collect descendant entry IDs. On LMDB it recursed into child subtrees inside the bulk_nextdata loop, which repositions the shared cursor before all duplicates under the parent key are read. That corrupts the affected IDList and crashes in idl_append_extend. The fix is to collect direct children into a temp LDList during the bulk loop, and recurse when it completes.

Fixes: https://github.com/389ds/389-ds-base/issues/7531

Reviewed by:

## Summary by Sourcery

Bug Fixes:
- Avoid crashes and corrupted IDLists in LMDB entryrdn processing during nested rename operations by collecting direct children separately before recursion.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
+++ b/ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
@@ -3320,17 +3320,17 @@ _entryrdn_append_childidl(entryrdn_db_ctx_t *ctx,
     /* E.g., C5 */
     char *keybuf = slapi_ch_smprintf("%c%u", RDN_INDEX_CHILD, id);
     dbi_val_t key = {0};
-    dbi_bulk_t data = {0};
-    char buffer[RDN_BULK_FETCH_BUFFER_SIZE];
+    dbi_val_t dataret = {0};
+    IDList *direct_children = NULL;
+    idl_iterator child_iter;
     int rc = 0;
 
     dblayer_value_set(ctx->be, &key, keybuf, strlen(keybuf) + 1);
-    /* Setting the bulk fetch buffer */
-    dblayer_bulk_set_buffer(ctx->be, &data, buffer, sizeof(buffer), DBI_VF_BULK_DATA);
+    dblayer_value_init(ctx->be, &dataret);
 
 /* Position cursor at the matching key */
 retry_get0:
-    rc = dblayer_cursor_bulkop(&ctx->cursor, DBI_OP_MOVE_TO_KEY, &key, &data);
+    rc = dblayer_cursor_op(&ctx->cursor, DBI_OP_MOVE_TO_KEY, &key, &dataret);
     if (rc) {
         if (DBI_RC_RETRY == rc) {
             slapi_log_err(ENTRYRDN_LOGLEVEL(rc), "_entryrdn_append_childidl",
@@ -3346,7 +3346,7 @@ retry_get0:
             rc = 0; /* okay not to have children */
         } else {
             _entryrdn_cursor_print_error("_entryrdn_append_childidl",
-                                         key.data, data.v.size, data.v.ulen, rc);
+                                         key.data, dataret.size, dataret.ulen, rc);
         }
         _ENTRYRDN_DEBUG_GOTO_BAIL();
         goto bail;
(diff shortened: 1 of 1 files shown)


progier389 (commented):
IMHO, for lmdb,  I think it is better to rewrite the algorithm avoiding using dblayer_cursor_bulkop. read bulk operations are not natively supported by lmdb so dblayer_cursor_bulkop is less efficient and more fragile than using plain dblayer_cursor_op(...,DBI_OP_NEXT_DATA,...) to iterate on duplicates

jchapma:
> IMHO, for lmdb, I think it is better to rewrite the algorithm avoiding using dblayer_cursor_bulkop. read bulk operations are not natively supported by lmdb so dblayer_cursor_bulkop is less efficient and more fragile than using plain dblayer_cursor_op(...,DBI_OP_NEXT_DATA,...) to iterate on duplicates

Just to make sure I understand you correctly: Instead of using the emulated bulk api, we should walk duplicates under the same parent key using dblayer_cursor_op api. But we still keep the deferred recursion, right ? IIUC the cursor is shared across recursive calls, so a nested entryrdn_append_childidl call would move the cursor with its own DBI_OP_MOVE_TO_KEY, which would break the parents duplicate walk if we recursed inside the loop. In the fix we store the direct children first, then we recurse once the walk is done ?

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -3370,28 +3373,46 @@ _entryrdn_append_childidl(entryrdn_db_ctx_t *ctx,
...
                 _ENTRYRDN_DEBUG_GOTO_BAIL();
                 goto bail;
             }
-            rc = _entryrdn_append_childidl(ctx, (const char *)myelem->rdn_elem_nrdn_rdn,
-                                           myid, affectedidl);
-            if (myelem != dataret.data) {
-                /* myelem was alloc by _entryrdn_resolve_redirect */
-                slapi_ch_free((void**)&myelem);
-            }
+            rc = idl_append_extend(&direct_children, myid);
```
droideck: Small cleanup: once direct_children has entries, later early goto bail paths can leak it. Could we move it to function scope and free it from bail?

ldap/servers/slapd/back-ldbm/ldbm_entryrdn.c
```diff
@@ -3370,28 +3373,46 @@ _entryrdn_append_childidl(entryrdn_db_ctx_t *ctx,
...
+         * causing an unbounded re-read of stale duplicate data. */
+        if (direct_children) {
+            child_iter = idl_iterator_init(direct_children);
+            for (;;) {
+                ID childid = idl_iterator_dereference_increment(&child_iter, direct_children);
+
+                if (childid == NOID) {
+                    break;
+                }
+                rc = _entryrdn_append_childidl(ctx, nrdn, childid, affectedidl);
```
droideck: IIUC, this looks oke for LMDB as bulk_nextdata drains entryrdn duplicates one at a time before recursion. But for BDB, can DB_MULTIPLE require another DBI_OP_NEXT_DATA buffer here? Maybe we can collect all direct children before recursing..

droideck (changes requested):
My 5 cent on the above discussion: I think Pierre's point is reasonable for a follow-up ticket. LMDB entryrdn values are variable length, so this bulk path does not get true MDB_GET_MULTIPLE; a plain dblayer_cursor_op duplicate loop may be simpler and less fragile.

jchapma:
> My 5 cent on the above discussion: I think Pierre's point is reasonable for a follow-up ticket. LMDB entryrdn values are variable length, so this bulk path does not get true MDB_GET_MULTIPLE; a plain dblayer_cursor_op duplicate loop may be simpler and less fragile.

I agree, I have updated _entryrdn_append_childidl to use dblayer_cursor_op instead of bulk ops, I will open a new ticket to cover the four entryrdn bulk loops.

droideck (approved):
LGTM!

Issue #7576, opened 2026-06-11
schema reload task leaks memory


mreynolds389:
**Issue Description**

Fix leak of temporary attribute syntax hash tables after schema reload.  After attr_syntax_swap_ht() promotes the tmp tables to live use and sets oid2asi_tmp and name2asi_tmp to NULL, attr_syntax_init() was recreating
them on every attr_syntax_read_lock() call (e.g. from LDAP searches via
slapi_filter_schema_check), leaking PL_NewHashTable allocations.

```
Direct leak of 392 byte(s) in 7 object(s) allocated from:
    #0 0x7f5039cef41f in malloc (/lib64/libasan.so.8+0xef41f) (BuildId: 5395ec74f54d9ec7bf97c06583dd39a96c230822)
    #1 0x7f503a304adc in PL_NewHashTable (/lib64/libplds4.so+0xadc) (BuildId: 9399ebd48a1620ba78055e592c96b7803d0aae85)
    #2 0x7f503961f05c in attr_syntax_init ldap/servers/slapd/attrsyntax.c:1470
    #3 0x7f503961f45c in attr_syntax_read_lock ldap/servers/slapd/attrsyntax.c:96
    #4 0x7f503979d238 in slapi_filter_schema_check (/usr/lib64/dirsrv/libslapd.so.0+0x19d238) (BuildId: d6fae35105d2777c7fa9f53c1c0b1f4ceb36aaca)
    #5 0x56326022f75f in do_search ldap/servers/slapd/search.c:223
    #6 0x56326020913e in connection_dispatch_operation ldap/servers/slapd/connection.c:740
    #7 0x56326020ad4f in connection_threadmain ldap/servers/slapd/connection.c:2070
    #8 0x7f503a3adc52 in _pt_root (/lib64/libnspr4.so+0x24c52) (BuildId: b2e27a52bd2230bc288993a204091f4ae9318dc0)
    #9 0x7f5039c2b606 in asan_thread_start(void*) (/lib64/libasan.so.8+0x2b606) (BuildId: 5395ec74f54d9ec7bf97c06583dd39a96c230822)
    #10 0x7f5039479d18 in start_thread (/lib64/libc.so.6+0x72d18) (BuildId: 18472003bbf1c5f098a09b5016b9b8bd4c7c59f0)
[...]

PR #7577, opened 2026-06-11
Issue 7576 - Fix leak of temporary attribute syntax hash tables after…

mreynolds389:
… schema reload

Description:

After attr_syntax_swap_ht() promotes the tmp tables to live use and sets oid2asi_tmp and name2asi_tmp to NULL, attr_syntax_init() was recreating them on every attr_syntax_read_lock() call (e.g. from LDAP searches via slapi_filter_schema_check), leaking PL_NewHashTable allocations.

Create the tmp tables lazily with attr_syntax_init_tmp() only during schema reload, and discard them with attr_syntax_destroy_tmp() on reload start, failure, or before rebuilding. attr_syntax_init() now initializes only the live oid2asi and name2asi tables.

relates: https://github.com/389ds/389-ds-base/issues/7576

co-authored by: Cursor

## Summary by Sourcery

Prevent leakage of temporary attribute syntax hash tables by managing their lifecycle explicitly during schema reloads.

Bug Fixes:
- Avoid leaking temporary attribute syntax hash tables that were being recreated after schema reloads.

Enhancements:
- Introduce dedicated helpers to lazily initialize and destroy temporary attribute syntax tables around schema reload operations.

### DIFF
--- a/ldap/servers/slapd/attrsyntax.c
+++ b/ldap/servers/slapd/attrsyntax.c
@@ -226,6 +226,8 @@ attr_syntax_add_by_oid(const char *oid, struct asyntaxinfo *a, PRUint32 schema_f
         return;
 
     if (schema_flags & DSE_SCHEMA_LOCKED) {
+        if (0 != attr_syntax_init_tmp())
+            return;
         PL_HashTableAdd(oid2asi_tmp, oid, a);
     } else {
         if (lock) {
@@ -416,6 +418,8 @@ attr_syntax_add_by_name(struct asyntaxinfo *a, PRUint32 schema_flags, int lock)
         return;
 
     if (schema_flags & DSE_SCHEMA_LOCKED) {
+        if (0 != attr_syntax_init_tmp())
+            return;
         /* insert the attr into the temp global linked list */
         attr_syntax_insert_tmp(a);
 
@@ -1465,13 +1469,6 @@ attr_syntax_init(void)
         }
     }
 
-    if (!oid2asi_tmp) {
-        /* temporary hash table for schema reload */
-        oid2asi_tmp = PL_NewHashTable(2047, hashNocaseString,
-                                      hashNocaseCompare,
-                                      PL_CompareValues, 0, 0);
-    }
-
     if (!name2asi) {
         name2asi = PL_NewHashTable(2047, hashNocaseString,
                                    hashNocaseCompare,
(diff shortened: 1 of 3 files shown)


ldap/servers/slapd/attrsyntax.c
```diff
@@ -226,6 +226,8 @@ attr_syntax_add_by_oid(const char *oid, struct asyntaxinfo *a, PRUint32 schema_f
         return;
 
     if (schema_flags & DSE_SCHEMA_LOCKED) {
+        if (0 != attr_syntax_init_tmp())
```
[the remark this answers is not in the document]
mreynolds389: The error is already logged in attr_syntax_init_tmp(), and we already use slapi_log_err() in this file without issue.

ldap/servers/slapd/schema.c
```diff
@@ -4941,6 +4941,14 @@ slapi_reload_schema_files(char *schemadir)
     slapi_be_Wlock(be); /* be lock must be outer of schemafile lock */
     reload_schemafile_lock();
     oc_delete_all_nolock();
+    attr_syntax_destroy_tmp();
+    if (0 != attr_syntax_init_tmp()) {
```
droideck, line 4944: Should we maybe do the destroy/init_tmp dance **before** oc_delete_all_nolock()? If init_tmp fails here we return with all the objectclasses already wiped and nothing re-parsed, so the server is left rejecting every add until someone reruns the task. I know it's OOM-only and the parse failure path below has the same problem anyway, but the reorder looks free since init_tmp doesn't touch the oc list.

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1492,16 +1489,75 @@ attr_syntax_init(void)
...
+
     if (!name2asi_tmp) {
-        /* temporary hash table for schema reload */
         name2asi_tmp = PL_NewHashTable(2047, hashNocaseString,
                                        hashNocaseCompare,
                                        PL_CompareValues, 0, 0);
+        if (!name2asi_tmp) {
+            slapi_log_err(SLAPI_LOG_ERR, "attr_syntax_init_tmp",
+                          "Failed to create name2asi_tmp hash table\n");
+            PL_HashTableDestroy(oid2asi_tmp);
```
droideck: Minor, but this destroys oid2asi_tmp even if it was already there before this call (with entries still sitting on global_at_tmp). Can't actually happen today since the two tables always come and go together, but maybe just create both and call attr_syntax_destroy_tmp() on failure? Keeps the "both or neither" invariant in one place and we don't need the hand-rolled cleanup.

ldap/servers/slapd/schema.c
mreynolds389: > Should we maybe do the destroy/init_tmp dance **before** oc_delete_all_nolock()? If init_tmp fails here we return with all the objectclasses already wiped and nothing re-parsed, so the server is left rejecting every add until someone reruns the task. I know it's OOM-only and the parse failure path below has the same problem anyway, but the reorder looks free since init_tmp doesn't touch the oc list.

Sounds good

mreynolds389:
@droideck changes made please review

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1492,16 +1489,80 @@ attr_syntax_init(void)
...
+            rc = 1;
+        }
+    }
+
+    if (rc) {
+        /* destroy everything on failure */
+        attr_syntax_destroy_tmp();
     }
 
     return 0;
```
droideck: Should this return `rc` instead? If either tmp hash table allocation fails, we log and destroy the partial tmp state above, but returning 0 makes `slapi_reload_schema_files()` continue as if initialization succeeded.

droideck (approved):
Besides the minor issue, LGTM!

Issue #7578, opened 2026-06-11
schema - attribute refcount is not maintained properly


mreynolds389:
**Issue Description**

The refcount mechanism for attributes is not correctly used, and this prevents its use for synchronizing hashtable access.  We increment the refcount, but we don't decrement it correctly.  This can potentially cause a rare race condition with schema reload task and heap-use-after-free with searches running at the same time as the reload task.

mreynolds389:
This also requires the fix for https://github.com/389ds/389-ds-base/issues/7576

PR #7582, opened 2026-06-15
Issue 7578 - schema - attribute refcount is not maintained properly

mreynolds389:
Description:

The refcount mechanism for attributes is not correctly used, and this prevents its use for synchronizing hashtable access. We increment the refcount, but we don't decrement it correctly. This can potentially cause a rare race condition with schema reload task and heap-use-after-free with searches running at the same time as the reload task.

**CWE**: CWE-416 (Use After Free) via CWE-362 (Race Condition)

CI test assisted by: Cursor

relates: https://github.com/389ds/389-ds-base/issues/7578

## Summary by Sourcery

Fix attribute syntax reference counting and cleanup during schema reload to prevent use-after-free race conditions, and add a stress test that exercises concurrent searches with schema reloads to detect crashes.

Bug Fixes:
- Correct attribute syntax reference counting by ensuring attr_syntax_return is called wherever references are taken and by guarding deletion on both refcount and deletion flags.
- Prevent premature freeing of global attribute syntax structures during hash table swaps by honoring reference counts and delete markers, avoiding use-after-free during schema reloads.
- Improve error logging in schema reload paths with clearer messages and ensure temporary DSE structures are destroyed on failure.

Tests:
- Add a multithreaded schema reload stress test that runs concurrent searches and schema reload tasks, with crash monitoring and detailed statistics, to expose heap-use-after-free or race-condition crashes in attribute syntax handling.

### DIFF
--- a/dirsrvtests/tests/suites/schema/schema_reload_heap_use_after_free_test.py
+++ b/dirsrvtests/tests/suites/schema/schema_reload_heap_use_after_free_test.py
@@ -1,0 +1,403 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+import threading
+import time
+import ldap
+from collections import Counter
+from lib389.schema import Schema
+from lib389.idm.domain import Domain
+from lib389.rootdse import RootDSE
+from lib389._constants import DEFAULT_SUFFIX
+from test389.topologies import topology_st as topo
+
+log = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Globals for cross-thread coordination
+# ---------------------------------------------------------------------------
+stop_event = threading.Event()
+crash_detected = threading.Event()
+reload_failed = threading.Event()
+stats_lock = threading.Lock()
+stats = Counter()
+
+ITERATIONS = 10
+SEARCH_THREADS = 12
+MAX_STRESS_TEST_DURATION = 300
+
+SEARCH_ATTRS = [
+    'cn', 'sn', 'givenName', 'mail', 'uid', 'userPassword',
+    'objectClass', 'description', 'telephoneNumber', 'l', 'st',
+    'postalCode', 'street', 'ou', 'title', 'manager', 'secretary',
+    'homeDirectory', 'loginShell', 'uidNumber', 'gidNumber',
+    'memberOf', 'nsAccountLock', 'createTimestamp', 'modifyTimestamp',
+    'entrydn', 'entryid', 'numSubordinates', 'hasSubordinates',
+]
+
+
+def _is_connection_error(exc):
+    """Return True if an LDAP error likely means the server crashed."""
+    if isinstance(exc, (ldap.SERVER_DOWN, ldap.CONNECT_ERROR, ldap.UNAVAILABLE)):
+        return True
+    if isinstance(exc, ldap.LDAPError):
+        msg = str(exc).lower()
+        return any(s in msg for s in (
+            "can't contact", "connection reset",
+            "server is unavailable", "connection refused",
+            "broken pipe",
+        ))
+    return False
+
+
+def check_server_alive(inst):
+    """Quick check: can we connect and read the root DSE?"""
+    try:
+        return RootDSE(inst).exists()
+    except ldap.LDAPError as e:
+        log.error(f"check_server_alive: check_server_alive failed: {str(e)}")
+        return False
+
+
+def search_worker(inst, _worker_id):
+    """Flood LDAP searches that exercise attribute syntax lookups.
+
+    Each search forces the server to look up attribute syntax for the
+    requested attributes (cn, sn, mail, etc.) via attr_syntax_find().
+    That function acquires the read lock, looks up the asyntaxinfo node,
+    bumps the refcount, and releases the lock -- creating the race window.
+
+    We search with many attributes to maximize the number of
+    attr_syntax_find() calls per operation, widening the window.
+    """
+    conn = inst.clone()
+    conn.rebind()
+    domain = Domain(conn, DEFAULT_SUFFIX)
+    domain._list_attrlist = SEARCH_ATTRS
+
+    try:
+        while not stop_event.is_set() and not crash_detected.is_set():
+            try:
+                domain.search(scope='subtree', filter='(objectclass=*)')
+                with stats_lock:
+                    stats['searches_ok'] += 1
+            except ldap.LDAPError as e:
+                with stats_lock:
+                    stats['searches_fail'] += 1
+                    if _is_connection_error(e):
+                        stats['connection_errors'] += 1
+            except Exception as e:
+                log.error(f"search_worker: failed with unknown exception: {str(e)}")
+                with stats_lock:
+                    stats['searches_error'] += 1
+    finally:
+        conn.close()
+
+
+def schema_reload_worker(inst):
+    """Repeatedly trigger schema reload to race against query threads.
+
+    Schema reload is triggered by adding a temporary attribute type to
+    cn=schema, then deleting it. Each add/delete causes the
+    server to call attr_syntax_swap_ht(), which frees all old asyntaxinfo
+    nodes without checking reference counts.
+    """
+    schema = Schema(inst)
+
+    for i in range(ITERATIONS):
+        if stop_event.is_set() or crash_detected.is_set():
+            break
+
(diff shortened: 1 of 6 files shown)


progier389 (approved):
LGTM

ldap/servers/slapd/schema.c
```diff
@@ -2859,11 +2861,12 @@ schema_replace_attributes(Slapi_PBlock *pb, LDAPMod *mod, char *errorbuf, size_t
                                        schema_errprefix_at, newasip->asi_name,
                                        "Could not be added (OID is \"%s\")",
                                        newasip->asi_oid);
-                attr_syntax_free(newasip);
+                attr_syntax_return(newasip);
                 goto clean_up_and_return;
             }
 
             newasip->asi_flags |= SLAPI_ATTR_FLAG_KEEP;
+            attr_syntax_return(newasip);
```
droideck: I think this return might be unsafe for this ownership path... `newasip` comes from `parse_at_str(_, &newasip, _)`, so it was newly allocated rather than acquired through an attr_syntax_get_*() lookup. IIUC, calling attr_syntax_return() here can decrement a refcount that was never incremented; after attr_syntax_add() succeeds, the schema table should own the object.

ldap/servers/slapd/schema.c
```diff
@@ -4214,6 +4219,7 @@ load_schema_dse(Slapi_PBlock *pb, Slapi_Entry *e, Slapi_Entry *ignored __attribu
                                                 primary_file /* force user defined? */,
                                                 schema_ds4x_compat, 0)) != 0)
                     break;
+                attr_syntax_return(tmpasip);
```
droideck: This looks like the same ownership mismatch as above. `tmpasip` is only a temporary parse result used for validation, not a refcounted lookup result, Should it be released directly with attr_syntax_free() without first calling attr_syntax_return()?

ldap/servers/slapd/schema.c
```diff
@@ -6047,6 +6055,7 @@ schema_berval_to_atlist(struct berval **at_berval)
                 at->asi_prev = at_list;
                 at_list = at;
             }
+            attr_syntax_return(at);
```
droideck: Probably, a similar issue here too? The parsed `at` node is linked into the private remote_at_list and later freed by schema_atlist_free(), so calling attr_syntax_return() while building the list might underflow the refcount

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1709,7 +1714,10 @@ attr_syntax_swap_ht()
     /* Free the global attr linked list */
     while (global_at) {
         next = global_at->asi_next;
-        attr_syntax_free(global_at);
+        global_at->asi_marked_for_delete = 1;
```
droideck, line 1717: Maybe I'm overthinking this, but I think deferred old nodes still keep their `asi_prev`/`asi_next` links into the old `global_at` list. If they are later freed via `attr_syntax_return()`, could `attr_syntax_remove()` follow stale links or disturb the new `global_at`? Should we detach them during the swap or skip unlinking swap-orphans?

dirsrvtests/tests/suites/schema/schema_reload_heap_use_after_free_test.py
```diff
@@ -0,0 +1,391 @@
...
+            with stats_lock:
+                stats['reloads_ok'] += 1
+        except ldap.LDAPError as e:
+            log.error(f"schema_reload_worker: remove_attributetype failed: {attr_name} - {str(e)}")
+            with stats_lock:
+                stats['reloads_fail'] += 1
+        except Exception:
+            pass
+
+        task = schema.reload()
```
droideck, line 176: This assertion runs inside the background reload thread, so pytest may not fail the test reliably if the reload task exits non-zero. Could we record reload task failures in stats and assert them from the main test thread, along with the existing crash/server-alive checks?

droideck (changes requested):

ldap/servers/slapd/attrsyntax.c
tbordaz, line 1717: I agree with that change but I would expect code review tools to report a risk of leak in case refcnt is not null.

ldap/servers/slapd/schema.c
```diff
@@ -2859,11 +2861,12 @@ schema_replace_attributes(Slapi_PBlock *pb, LDAPMod *mod, char *errorbuf, size_t
                                        schema_errprefix_at, newasip->asi_name,
                                        "Could not be added (OID is \"%s\")",
                                        newasip->asi_oid);
-                attr_syntax_free(newasip);
+                attr_syntax_return(newasip);
```
tbordaz: Sorry I can not find where the refcnt of 'newasip' is increased in attr_sync_add or before.
So attr_sync_return may set refcnt to -1. Why not just keeping attr_syntax_free ?

ldap/servers/slapd/schema.c
tbordaz: I think this one is useless as refcnt was not increase in that path

ldap/servers/slapd/schema.c
```diff
@@ -4387,15 +4393,16 @@ init_schema_dse_ext(char *schemadir, Slapi_Backend *be, struct dse **local_psche
             struct asyntaxinfo *tmpasip = NULL;
             rc = parse_at_str(attr_str, &tmpasip, errorbuf, sizeof(errorbuf),
                               DSE_SCHEMA_NO_GLOCK | schema_flags, 0, 0, 0);
+            attr_syntax_return(tmpasip);
```
tbordaz: same remark

ldap/servers/slapd/schema.c
```diff
@@ -6037,7 +6045,7 @@ schema_berval_to_atlist(struct berval **at_berval)
                 slapi_log_err(SLAPI_LOG_ERR, "schema_berval_to_atlist",
                               "parse_at_str(%s) failed - %s\n",
                               at_berval[i]->bv_val, errorbuf[0] ? errorbuf : "unknown");
-                attr_syntax_free(at);
+                attr_syntax_return(at);
```
tbordaz: Same remark I believe that parse_at_str does not increase refcnt

ldap/servers/slapd/schema.c
tbordaz: same remark

tbordaz (commented):
Few comments

mreynolds389:
@droideck @tbordaz thanks for the comments.  I need to run more tests to verify, but I recall that all these changes were needed to avoid ASAN errors and to fix the original heap corruption issue.  This patch was heavily tested without any issues, but perhaps some of the changes are not needed...

ldap/servers/slapd/attrsyntax.c
```diff
@@ -1709,7 +1714,10 @@ attr_syntax_swap_ht()
...
+        if (slapi_atomic_load_64(&global_at->asi_refcnt, __ATOMIC_ACQUIRE) == 0) {
```
droideck, line 1718: Should we also cover the nsslapd-schemamod=off configuration case?(`attr_syntax_init()` sets `asi_locking=0`)
Locking is disabled, but reload still seems allowed. Do you think it's possible to have a situation when lookup could race with hash-table destruction before the refcount is incremented?

droideck (changes requested):
And couple of other comments from the past review, IMO, still valid.

ldap/servers/slapd/attrsyntax.c
mreynolds389: > Should we also cover the nsslapd-schemamod=off configuration case?(`attr_syntax_init()` sets `asi_locking=0`) Locking is disabled, but reload still seems allowed. Do you think it's possible to have a situation when lookup could race with hash-table destruction before the refcount is incremented?

Sorry having a hard time trying to track down what the concern actually is.  All the schema/syntax locking is very complicated, and I do not change the locking behavior in this PR.  I have no problem rejecting the schema reload task is schemamod is "off" - does that address your concerns?

droideck: > Sorry having a hard time trying to track down what the concern actually is. All the schema/syntax locking is very complicated, and I do not change the locking behavior in this PR. I have no problem rejecting the schema reload task is schemamod is "off" - does that address your concerns?

Sorry, my original comment probably wasn't that clear... The concern is that the refcount only protects the ASI after the lookup and increment complete; the read lock protects that handoff. With `asi_locking=0`, reload could destroy the table between lookup and increment, IIUC.

So yes, rejecting schema reload when nsslapd-schemamod=off would address this, provided the check is server-side and the setting cannot change during the reload, hmm...
It maybe worth a small test, but it's not a blocker.

mreynolds389: Fixed and added test.

droideck (approved):
LGTM

Issue #7541, opened 2026-06-03
ASAN crashes in __aclp__normalize_acltxt()


IliaKash1:
Hello maintainers! I have found 3 related ASAN crashes `in __aclp__normalize_acltxt()`, all reprodicible via `acl_parse()`.

Steps to reproduce:

1) Build the project with ASAN;
```bash
export CC=clang-19
export CXX=clang++-19

autoreconf --force --install
export RUSTUP_TOOLCHAIN=stable
./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking --with-openldap --with-systemd --with-systemdsystemunitdir=/lib/systemd/system --with-systemdsystemconfdir=/etc/systemd/system --with-systemdgroupname=dirsrv.target --with-tmpfiles-d=/etc/tmpfiles.d --enable-autobind --enable-cmocka --enable-icu --enable-rust --enable-asan

make

ar rsc liball.a ldap/servers/slapd/.libs/*.o
ar rsc libsvrscore.a src/svrcore/src/.libs/*.o
ar rsc libutil.a lib/ldaputil/.libs/*.o
ar rsc libacl.a ./ldap/servers/plugins/acl/.libs/*.o
ar rsc libaccess.a lib/libaccess/.libs/*.o
ar rsc libbase.a lib/base/.libs/*.o
ar rsc libsi18.a lib/libsi18n/.libs/*.o
```
2) Build the reproducer:
```bash
[...]

PR #7591, opened 2026-06-18
Issue 7541 - Add invalid ACL text header regression test

droideck:
Description: Malformed ACI values with missing or too-short ACL text headers should be rejected during syntax validation.

Add parameterized coverage for empty ACL text headers, short headers, and headers without an ACL name. The test verifies these values fail with invalid syntax and ASAN doesn't crash.

Related: https://github.com/389ds/389-ds-base/issues/7541

Reviewed by: ?

## Summary by Sourcery

Tests:
- Add a parameterized test that verifies ACIs with empty, too-short, or unnamed ACL text headers fail with INVALID_SYNTAX.


mirielka (approved):
LGTM

PR #7594, opened 2026-06-18
Issue 7593 - Reject invalid SASL packet length values in sasl_io_start_packet

### LINKED ISSUE #7593: Reject invalid SASL packet length values in sasl_io_start_packet()
**Issue Description**
While processing SASL encrypted traffic, sasl_io_start_packet() reads a 4-byte length from the connection and adds sizeof(uint32_t) before resizing the read buffer.

jchapma:
Description:
While processing SASL encrypted traffic, sasl_io_start_packet() reads a 4-byte length from the connection and adds sizeof(uint32_t) before resizing the read buffer. Certain large length values can wrap in uint32_t, causing incorrect buffer sizing when malformed SASL data is received on an established connection.

Fixes: https://github.com/389ds/389-ds-base/issues/7593

Reviewed by:

## Summary by Sourcery

Validate SASL packet lengths to prevent overflow and add regression coverage for malformed encrypted packets.

Bug Fixes:
- Reject SASL packets whose reported length would overflow when accounting for the length header.
- Ensure packet length limit comparison uses a correctly sized unsigned value to avoid misinterpretation of large lengths.

Tests:
- Add a SASL DIGEST-MD5 regression test that sends a malformed length-prefixed packet and verifies the server remains running.

### DIFF
--- a/ldap/servers/slapd/sasl_io.c
+++ b/ldap/servers/slapd/sasl_io.c
@@ -16,6 +16,7 @@
 #include "fe.h"
 #include <sasl/sasl.h>
 #include <arpa/inet.h>
+#include <stdint.h>
 
 /*
  * I/O Shim Layer for SASL Encryption
@@ -371,6 +372,14 @@ sasl_io_start_packet(PRFileDesc *fd, PRIntn flags, PRIntervalTime timeout, PRInt
     /* Decode the length */
     packet_length = ntohl(*(uint32_t *)sp->encrypted_buffer);
     /* add length itself (for Cyrus SASL library) */
+    if (packet_length > (UINT32_MAX - sizeof(uint32_t))) {
+        slapi_log_err(SLAPI_LOG_ERR, "sasl_io_start_packet",
+                      "SASL packet length would overflow (%" PRIu32 ")\n",
+                      packet_length);
+        PR_SetError(PR_BUFFER_OVERFLOW_ERROR, 0);
+        *err = PR_BUFFER_OVERFLOW_ERROR;
+        return -1;
+    }
     packet_length += sizeof(uint32_t);
 
     slapi_log_err(SLAPI_LOG_CONNS, "sasl_io_start_packet",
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/sasl_io.c
```diff
@@ -371,6 +372,14 @@ sasl_io_start_packet(PRFileDesc *fd, PRIntn flags, PRIntervalTime timeout, PRInt
     /* Decode the length */
     packet_length = ntohl(*(uint32_t *)sp->encrypted_buffer);
     /* add length itself (for Cyrus SASL library) */
+    if (packet_length > (UINT32_MAX - sizeof(uint32_t))) {
```
tbordaz, line 375: I think we can also check config_get_maxsasliosize like
`
if (((saslio_limit != -1) && (packet_length > (uint32_t)saslio_limit)) ||
     packet_length > (UINT32_MAX - sizeof(uint32_t))) {`

tbordaz (commented):
LGTM

ldap/servers/slapd/sasl_io.c
jchapma: IIUC, If we used a combined if to check packet_length with  saslio_limit, the 4B header hasnt been added yet, we would be comparing the raw packet length and not the total buffer size.

ldap/servers/slapd/sasl_io.c
```diff
@@ -380,7 +389,7 @@ sasl_io_start_packet(PRFileDesc *fd, PRIntn flags, PRIntervalTime timeout, PRInt
     /* Check if the packet length is larger than our max allowed.  A
      * setting of -1 means that we allow any size SASL IO packet. */
     saslio_limit = config_get_maxsasliosize();
-    if ((saslio_limit != -1) && (packet_length > saslio_limit)) {
+    if ((saslio_limit != -1) && (packet_length > (uint32_t)saslio_limit)) {
```
progier389: Just curious: Why adding a cast here ? it seems useless as config_set_maxsasliosize ensures that saslio_limit >= 0 if it is not -1

ldap/servers/slapd/sasl_io.c
tbordaz: Thanks

tbordaz (approved):
LGTM

ldap/servers/slapd/sasl_io.c
jchapma: I added the cast to make the comparison explicitly unsigned, but as saslio_limit is already non negative this cat is redundant. Let me remove it.

Issue #7595, opened 2026-06-19
CI - Reduce GitHub Actions queue saturation from large pytest matrices


droideck:
## Problem

`Test` and `LMDB Test` each generate about 110 matrix jobs. When multiple PRs, branches, scheduled runs, or Renovate updates trigger CI at the same time, the repo can submit a large burst of GitHub-hosted runner jobs.

Some jobs then sit at:

`Waiting for a runner to pick up this job`

They never reach `Set up job`, never pull containers, and produce no logs.

This blocks PRs without actionable failure logs, wastes maintainer time on reruns, and makes real test failures harder to separate from CI infrastructure backlog.

It also lets unrelated branches and PRs delay each other when large matrices all fan out at once.

## Proposed Fix

Add `strategy.max-parallel` to `Test` and `LMDB Test` so each workflow run releases pytest jobs in smaller batches instead of all at once.

This preserves test coverage but reduces runner queue pressure.

## Acceptance Criteria

- `Test` and `LMDB Test` use `max-parallel`.
- Full matrix coverage is preserved.
- CI burst size is reduced.
- Image pull retry/fallback work remains tracked separately.

droideck:
Yeah, it gets a bit crazy... Started 16h ago, tests waiting but still blocking other jobs.

<img width="1108" height="557" alt="Image" src="https://github.com/user-attachments/assets/a314da14-60e9-44bb-9b5f-c435522611fc" />

<img width="1008" height="390" alt="Image" src="https://github.com/user-attachments/assets/28d1e9e3-f3eb-488d-a410-62e788d47051" />

Ace2932:
Max-parallel will soften the bursts and it doesn't cost any coverage, so it's the right first move. 
There's a second thing feeding the same queue that the issue doesn't mention though: both Test and LMDB Test trigger on unscoped `push:` as well as `pull_request:`, so every push to an in-repo branch with an open PR fires the whole ~110-job matrix twice for the same commit. 

In the last 100 runs of each workflow, 10 commits ran the full matrix once as a push event and again as a pull_request event, and most of them were Renovate branches, which the issue already calls out as a burst source. The concurrency group can't catch the pair either, because the push run keys on the ref while the PR run keys on the PR number, so they land in different groups, and cancel-in-progress only applies to the pull_request half. Scoping the push trigger to the branches you actually push to directly keeps all the coverage and halves those bursts:

```yaml
on:
  push:
    branches: [main, "389-ds-base-*"]
[...]

PR #7596, opened 2026-06-19
Issue 7595 - Reduce Actions queue saturation

droideck:
Description: Test and LMDB Test each generate large pytest matrices, which can submit hundreds of GitHub-hosted runner jobs when multiple PRs, branches, schedules, or Renovate updates trigger CI together.

Limit both pytest matrix workflows with strategy.max-parallel so each workflow run releases test jobs in smaller batches while preserving full matrix coverage. Also fix the GCC Strict compile matrix name and ignore cache/hidden suite directories when generating the pytest matrix.

Related: https://github.com/389ds/389-ds-base/issues/7595

Reviewed by: ?

## Summary by Sourcery

Limit CI test matrix concurrency and tighten matrix generation to avoid invalid suites and hidden directories.

Enhancements:
- Refine pytest matrix generation to ignore hidden, cache, and symlinked suite entries while handling replication tests via a shared suites path constant.
- Standardize the GCC Strict compile job name in the compile workflow.

CI:
- Set max-parallel to 6 for both pytest and LMDB pytest matrix workflows to throttle concurrent GitHub-hosted jobs.

### DIFF
--- a/.github/scripts/generate_matrix.py
+++ b/.github/scripts/generate_matrix.py
@@ -3,32 +3,49 @@ import sys
 import glob
 import json
 
+SUITES_PATH = "dirsrvtests/tests/suites"
+
+
+def is_valid_suite_dir(suite):
+    test_path = os.path.join(SUITES_PATH, suite)
+    return (
+        os.path.isdir(test_path)
+        and not os.path.islink(test_path)
+        and not suite.startswith(".")
+        and suite != "__pycache__"
+    )
+
+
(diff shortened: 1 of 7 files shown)


progier389 (commented):
Looks good nut you may perhaps take in account @Ace2932 comment: https://github.com/389ds/389-ds-base/issues/7595#issuecomment-4869667675

progier389 (approved):
LGTM

PR #7598, opened 2026-06-19
Issue 7535 - Fix race in test_schema_update_policy_reject

### LINKED ISSUE #7535: CI - Stabilize schema_update_policy_reject in BDB schema tests
## Problem

BDB `schema` tests are recurring in recent failed-main runs, with the representative failure in `schema_update_policy_test.py::test_schema_update_policy_reject`.

Recent signal from CI sweep:
- BDB `schema` failed in 8/20 sampled failed-main runs.
- Representative run/job: https://github.com/389ds/389-ds-base/actions/runs/26701053253, job `78694341731`.

Representative failure:

```text
dirsrvtests/tests/suites/schema/schema_update_policy_test.py::test_schema_update_policy_reject FAILED
dirsrvtests/tests/suites/schema/schema_update_policy_test.py:235: AssertionError
```

The failure appears to be that the reject policy is not rejecting the member attribute as expected.

## Proposed work

Investigate whether the schema update policy behavior changed, whether replication or test ordering leaves the policy in an unexpected state, or whether the test assertion needs to wait for/inspect a different state transition.

## Acceptance criteria

- `schema_update_policy_test.py::test_schema_update_policy_reject` passes reliably in BDB CI.
- The test confirms the reject policy through a deterministic observable state.
- Any product behavior change is covered by a focused regression check.

aadhikar:
Description: Removed temporary_oc2 fixture that added OC2 before the reject policy was active. Moved OC2 creation after the policy is set and supplier1 is restarted.

Fixes: #7535

Reviewed by: ???

Assisted by: Claude

## Summary by Sourcery

Adjust schema update rejection test to create and clean up the temporary object class after the reject policy is active, ensuring replication behavior is validated correctly.

Bug Fixes:
- Fix a race condition in the schema update rejection test where the temporary object class could be created before the reject policy was applied.

Tests:
- Update the schema update policy rejection test flow to verify the temporary object class is absent on the consumer before replication, and add explicit cleanup within the test.

### DIFF
--- a/dirsrvtests/tests/suites/schema/schema_update_policy_test.py
+++ b/dirsrvtests/tests/suites/schema/schema_update_policy_test.py
@@ -63,27 +63,6 @@ def _add_oc(instance, oid, name):
     schema.add_objectclass(params)
 
 
-@pytest.fixture
-def temporary_oc2(topology_m2):
-    """Fixture to create and automatically clean up OC2UpdatePolicy."""
-    supplier1 = topology_m2.ms["supplier1"]
-    supplier2 = topology_m2.ms["supplier2"]
-    name = "OC2UpdatePolicy"
-
-    log.info(f"Adding temporary objectclass: {name}")
-    _add_oc(supplier1, "1.2.3.4.5.6.7.8.9.10.3", name)
-
-    yield name
-
-    log.info(f"Cleaning up temporary objectclass: {name}")
-    for s in [supplier1, supplier2]:
-        try:
-            schema = Schema(s)
-            schema.remove_objectclass(name)
-        except (ldap.NO_SUCH_OBJECT, ValueError):
-            pass
-
-
 @pytest.fixture(scope="function")
 def setup_test_env(request, topology_m2):
     """Initialize the test environment"""
@@ -181,31 +160,34 @@ def test_schema_update_policy_allow(topology_m2, setup_test_env):
     assert entry.get_attr_val_utf8('description') == 'test_add'
 
 
-def test_schema_update_policy_reject(topology_m2, setup_test_env, temporary_oc2):
+def test_schema_update_policy_reject(topology_m2, setup_test_env):
     """Test that schema updates can be rejected based on policy
 
     :id: d9f4e3b5-6c8b-5b0e-0f3a-9b6c5d8e9g0
     :setup: Two supplier replication setup with test entries
     :steps:
-        1. Configure supplier1 to reject schema updates containing OC_NAME
-        2. Add a new objectclass on supplier1
-        3. Update an entry to trigger schema push
-        4. Verify schema was not pushed to supplier2
-        5. Remove reject policy
-        6. Update entry again
-        7. Verify schema is now pushed to supplier2
+        1. Configure supplier1 to reject schema updates containing OCUpdatePolicy
+        2. Add OC2UpdatePolicy objectclass on supplier1
+        3. Verify OC2UpdatePolicy is NOT on supplier2 before triggering replication
+        4. Update an entry to trigger schema push
+        5. Verify schema was not pushed to supplier2
+        6. Remove reject policy
+        7. Update entry again
+        8. Verify schema is now pushed to supplier2
     :expectedresults:
         1. Policy should be configured
         2. New objectclass should be added
-        3. Update should trigger replication
-        4. Schema should not be pushed due to policy
-        5. Policy should be removed
-        6. Update should trigger replication
-        7. Schema should now be pushed
+        3. OC2UpdatePolicy should not be on supplier2
+        4. Update should trigger replication
+        5. Schema should not be pushed due to policy
+        6. Policy should be removed
+        7. Update should trigger replication
+        8. Schema should now be pushed
     """
     supplier1 = topology_m2.ms["supplier1"]
     supplier2 = topology_m2.ms["supplier2"]
     repl = ReplicationManager(DEFAULT_SUFFIX)
+    oc2_name = "OC2UpdatePolicy"
 
     log.info("Configure supplier to reject schema updates for OCUpdatePolicy")
     policy_dn = f"cn=supplierUpdatePolicy,cn=replSchema,{supplier1.config.dn}"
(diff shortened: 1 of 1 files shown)


dirsrvtests/tests/suites/schema/schema_update_policy_test.py
```diff
@@ -247,9 +232,16 @@ def test_schema_update_policy_reject(topology_m2, setup_test_env, temporary_oc2)
...
     schema_attrs = supplier2.schema.get_objectclasses()
-    assert any('oc2updatepolicy' in (name.lower() for name in oc.names) for oc in schema_attrs)
+    assert any(oc2_name.lower() in (name.lower() for name in oc.names) for oc in schema_attrs)
+
+    log.info(f"Cleaning up {oc2_name}")
+    for s in [supplier1, supplier2]:
+        try:
+            Schema(s).remove_objectclass(oc2_name)
+        except (ldap.NO_SUCH_OBJECT, ValueError):
+            pass
```
droideck: Might worth placing it in `try/finally` block, IMO.

droideck (commented):
Thanks for putting this together, @aadhikar!
Since @mreynolds389 is assigned to #7535 and had also been looking into this, I'd leave the final review to him in case he has additional context or a preferred direction.
Besides the small cleanup/teardown comment:)

dirsrvtests/tests/suites/schema/schema_update_policy_test.py
aadhikar: @droideck Can you elaborate? If you mean wrapping the entire test logic in `try` with cleanup in `finally`, I can do that. If it's just the cleanup part, `finally` wouldn't add value since there's nothing to guard above it. Let me know what I am missing here!

mreynolds389:
I don't think the problem is the test.  I've been working on this the last few days and there is definitely is a race condition in DS (a bug).  When it fails, the schema reject policy is set the default policy for printer-uri-oid instead of the defined one for OCPolicyUpdate.  Since the policy set and the server is restarted it should be using the correct policy right away, but sometimes it still has the default policy set (I'm still working on it).

So I suspect if you run this test enough times it will still fail.

mreynolds389:
> I don't think the problem is the test. I've been working on this the last few days and there is definitely is a race condition in DS (a bug). When it fails, the schema reject policy is set the default policy for printer-uri-oid instead of the defined one for OCPolicyUpdate. Since the policy set and the server is restarted it should be using the correct policy right away, but sometimes it still has the default policy set (I'm still working on it).
> 
> So I suspect if you run this test enough times it will still fail.

Ok so I have verified this is a testing issue.  It's not a bug in DS.  What's happening is that the schema replication of "OCPolicyUpdate" is happening before we set it in the rejection policy.  I can not reproduce the problem using this patch.  I think the CI test could be hardened a bit more, but its been passing for me with both bdb and lmdb, so Ack!

mreynolds389 (approved):

tbordaz:
My understanding is the 'setup_test_env' fixture adds 'OCUpdatePolicy' in the topology (S1 and S2).
That means that before 'test_schema_update_policy_reject' starts 'OCUpdatePolicy' is already on S1 and S2.
So like @mreynolds389 said, rejecting OCUpdatePolicy is set too late.

The purpose of 'temporary_oc2' is just to update the schema and allow the replication of it. So IMHO we should not care about the presence of 'OC2UpdatePolicy' only 'OCUpdatePolicy' needs to be checked. This is the one we want to reject.

An option is

- reject OCUpdatePolicy on S1
- run 'temporary_oc2' to update the schema csn
- add 'OCUpdatePolicy' on S1
- update an entry (trigger replication)
- check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy' not on S2
- allow OCUpdatePolicy on S1
- update an entry (trigger replication)
- check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy'  on S2

aadhikar:
> My understanding is the 'setup_test_env' fixture adds 'OCUpdatePolicy' in the topology (S1 and S2). That means that before 'test_schema_update_policy_reject' starts 'OCUpdatePolicy' is already on S1 and S2. So like @mreynolds389 said, rejecting OCUpdatePolicy is set too late.
> 
> The purpose of 'temporary_oc2' is just to update the schema and allow the replication of it. So IMHO we should not care about the presence of 'OC2UpdatePolicy' only 'OCUpdatePolicy' needs to be checked. This is the one we want to reject.
> 
> An option is
> 
> * reject OCUpdatePolicy on S1
> * run 'temporary_oc2' to update the schema csn
> * add 'OCUpdatePolicy' on S1
> * update an entry (trigger replication)
> * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy' not on S2
> * allow OCUpdatePolicy on S1
> * update an entry (trigger replication)
> * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy'  on S2

Thanks for the suggestion @tbordaz! However, as per `schema.c` ([line 82-83](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/schema.c#L82-L83)), the supplier-side reject policy checks the consumer's schema, so OCUpdatePolicy must already be on S2 for the reject to trigger. Adding it only to S1 after the policy won't fire the reject, since S2 won't have it. This matches how your original `ticket47676_test.py` worked too. Let me know if I am missing something!

tbordaz:
> > My understanding is the 'setup_test_env' fixture adds 'OCUpdatePolicy' in the topology (S1 and S2). That means that before 'test_schema_update_policy_reject' starts 'OCUpdatePolicy' is already on S1 and S2. So like @mreynolds389 said, rejecting OCUpdatePolicy is set too late.
> > The purpose of 'temporary_oc2' is just to update the schema and allow the replication of it. So IMHO we should not care about the presence of 'OC2UpdatePolicy' only 'OCUpdatePolicy' needs to be checked. This is the one we want to reject.
> > An option is
> > 
> > * reject OCUpdatePolicy on S1
> > * run 'temporary_oc2' to update the schema csn
> > * add 'OCUpdatePolicy' on S1
> > * update an entry (trigger replication)
> > * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy' not on S2
> > * allow OCUpdatePolicy on S1
> > * update an entry (trigger replication)
> > * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy'  on S2
> 
> Thanks for the suggestion @tbordaz! However, as per `schema.c` ([line 82-83](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/schema.c#L82-L83)), the supplier-side reject policy checks the consumer's schema, so OCUpdatePolicy must already be on S2 for the reject to trigger. Adding it only to S1 after the policy won't fire the reject, since S2 won't have it. This matches how your original `ticket47676_test.py` worked too. Let me know if I am missing something!

I am afraid this comment is not inline with the code. Actually it looks to me that if a given attribute/objectclass in the schema is set to REPL_SCHEMA_UPDATE_REJECT_VALUE then the full update of the schema is abandoned independently if the consumer knows this attr/oc.
Regarding test_ticket47676_reject_action, the schema already contains OC_NAME. it  sets REPL_SCHEMA_UPDATE_REJECT_VALUE:OC_NAME. Then update the schema+OC2_NAME and force replication then check that the OC2_NAME has not been propagated to the consumer.

aadhikar:
> > > My understanding is the 'setup_test_env' fixture adds 'OCUpdatePolicy' in the topology (S1 and S2). That means that before 'test_schema_update_policy_reject' starts 'OCUpdatePolicy' is already on S1 and S2. So like @mreynolds389 said, rejecting OCUpdatePolicy is set too late.
> > > The purpose of 'temporary_oc2' is just to update the schema and allow the replication of it. So IMHO we should not care about the presence of 'OC2UpdatePolicy' only 'OCUpdatePolicy' needs to be checked. This is the one we want to reject.
> > > An option is
> > > 
> > > * reject OCUpdatePolicy on S1
> > > * run 'temporary_oc2' to update the schema csn
> > > * add 'OCUpdatePolicy' on S1
> > > * update an entry (trigger replication)
> > > * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy' not on S2
> > > * allow OCUpdatePolicy on S1
> > > * update an entry (trigger replication)
> > > * check 'OCUpdatePolicy' on S1 and 'OCUpdatePolicy'  on S2
> > 
> > 
> > Thanks for the suggestion @tbordaz! However, as per `schema.c` ([line 82-83](https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/schema.c#L82-L83)), the supplier-side reject policy checks the consumer's schema, so OCUpdatePolicy must already be on S2 for the reject to trigger. Adding it only to S1 after the policy won't fire the reject, since S2 won't have it. This matches how your original `ticket47676_test.py` worked too. Let me know if I am missing something!
> 
> I am afraid this comment is not inline with the code. Actually it looks to me that if a given attribute/objectclass in the schema is set to REPL_SCHEMA_UPDATE_REJECT_VALUE then the full update of the schema is abandoned independently if the consumer knows this attr/oc. Regarding test_ticket47676_reject_action, the schema already contains OC_NAME. it sets REPL_SCHEMA_UPDATE_REJECT_VALUE:OC_NAME. Then update the schema+OC2_NAME and force replication then check that the OC2_NAME has not been propagated to the consumer.

@tbordaz I ran your proposed flow as a quick experiment:

```
# Set reject policy BEFORE OC exists anywhere
policy_entry.add('schemaUpdateObjectclassReject', oc_name)
supplier1.restart()

# Bump schema CSN, then add the OC only on S1
_add_oc(supplier1, "...", oc2_name)
_add_oc(supplier1, "...", oc_name)

# Trigger replication
test_user.replace('description', 'experiment_test')
repl.wait_for_replication(supplier1, supplier2)

# Check
reject_policy = policy_entry.get_attr_val_utf8('schemaUpdateObjectclassReject')
s2_has_oc = Schema(supplier2).query_objectclass(oc_name) is not None

# pdb breakpoint:
# (Pdb) reject_policy
# 'OCTestReject'
# (Pdb) s2_has_oc
# True
```

The reject policy is set, but S2 still got the OC. It seems the reject only fires when the consumer already has the OC in its schema. Am I missing something?

dirsrvtests/tests/suites/schema/schema_update_policy_test.py
droideck: Yep, I think we need a finalizer/`try/finally` for the whole test clean up.

droideck (approved):
LGTM! Sorry for the late review! Somehow I missed it...

PR #7603, opened 2026-06-23
Issue 7602 - CI - lib389 user compare fails due to parentid mismatch

### LINKED ISSUE #7602: CI - lib389 user compare fails on replicated entry due to parentid mismatch
**Description**
The lib389 test user_compare_m2Repl fails even though replication succeeds. The only difference between the replicated users is parentid, for example supplier1 has b'4' and supplier2 has b'6'. Since parentid is backend-local metadata and may differ between suppliers, UserAccount.compare should ignore it or the test should exclude it.

droideck:
Description: Fix UserAccount.compare() failures between replicated entries where backend-local parentid values differ across suppliers. Treat parentid like entryid by excluding it from generic DSLdapObject compare attributes, since it is an internal database operational attribute generated per instance. Add regression coverage to ensure replicated user comparison does not include parentid.

Fixes: https://github.com/389ds/389-ds-base/issues/7602

Reviewed by: ?

## Summary by Sourcery

Exclude backend-local parentid from lib389 object comparison and add regression coverage for replicated user comparisons.

Bug Fixes:
- Prevent UserAccount.compare() from failing when replicated entries differ only in backend-local parentid values.

Enhancements:
- Treat parentid like other instance-local operational attributes in generic mapped object comparisons and clarify the compare() documentation comment.

Tests:
- Extend multi-supplier user comparison test to assert parentid is not part of compare attributes and provide detailed mismatch output on failure.

### DIFF
--- a/src/lib389/lib389/_mapped_object.py
+++ b/src/lib389/lib389/_mapped_object.py
@@ -158,8 +158,8 @@ class DSLdapObject(DSLogging, DSLint):
         self._create_objectclasses = []
         self._rdn_attribute = None
         self._must_attributes = None
-        # attributes, we don't want to compare
-        self._compare_exclude = ['entryid', 'modifytimestamp', 'nsuniqueid']
+        # Backend-generated attributes that are local to an instance.
+        self._compare_exclude = ['entryid', 'modifytimestamp', 'nsuniqueid', 'parentid']
         self._server_controls = None
         self._client_controls = None
         self._object_filter = '(objectClass=*)'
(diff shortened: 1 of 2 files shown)


src/lib389/lib389/_mapped_object.py
```diff
@@ -158,8 +158,8 @@ def __init__(self, instance, dn=None):
         self._create_objectclasses = []
         self._rdn_attribute = None
         self._must_attributes = None
-        # attributes, we don't want to compare
-        self._compare_exclude = ['entryid', 'modifytimestamp', 'nsuniqueid']
+        # Backend-generated attributes that are local to an instance.
+        self._compare_exclude = ['entryid', 'modifytimestamp', 'nsuniqueid', 'parentid']
```
progier389: I wonder if we should not also exclude some other operational attributes
( like the numsubordonates related one )

progier389 (approved):
LGTM

PR #7607, opened 2026-06-24
Issue 7147 - entrycache_eviction_test CI test is failing

progier389:
Fix entry cache LRU list corruption by preventing pinned_add to add the entry in the LRU if it is already there

Issue: #7147 

## Summary by Sourcery

Prevent LRU list corruption in the entry and DN caches and ensure cache entries are not added multiple times to the LRU.

Bug Fixes:
- Guard against adding cache entries to the LRU list when they are already present to avoid list corruption.
- Ensure entry and DN cache flush operations correctly maintain and clear LRU state to prevent inconsistent cache state.

Enhancements:
- Introduce explicit ENTRY_STATE_LRU tracking on cache entries and add assertions around LRU operations to detect misuse early.
- Add a CACHE_TYPE_UNKNOWN type and use it for newly allocated backentries until their concrete cache type is known.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/backentry.c
+++ b/ldap/servers/slapd/back-ldbm/backentry.c
@@ -25,7 +25,7 @@ backentry_free(struct backentry **bep)
     ep = *bep;
 
     PR_ASSERT(ep->ep_state & (ENTRY_STATE_DELETED | ENTRY_STATE_NOTINCACHE | ENTRY_STATE_INVALID));
-    PR_ASSERT((ep->ep_state & ENTRY_STATE_LRU) == 0);
+    PR_ASSERT((ep->ep_state & (ENTRY_STATE_LRU | ENTRY_STATE_PINNED)) == 0);
     if (ep->ep_entry != NULL) {
         slapi_entry_free(ep->ep_entry);
     }
--- a/ldap/servers/slapd/back-ldbm/backentry.c
+++ b/ldap/servers/slapd/back-ldbm/backentry.c
@@ -25,6 +25,7 @@ backentry_free(struct backentry **bep)
     ep = *bep;
 
     PR_ASSERT(ep->ep_state & (ENTRY_STATE_DELETED | ENTRY_STATE_NOTINCACHE | ENTRY_STATE_INVALID));
+    PR_ASSERT((ep->ep_state & ENTRY_STATE_LRU) == 0);
     if (ep->ep_entry != NULL) {
         slapi_entry_free(ep->ep_entry);
     }
@@ -42,7 +43,7 @@ backentry_alloc()
     struct backentry *ec;
     ec = (struct backentry *)slapi_ch_calloc(1, sizeof(struct backentry));
     ec->ep_state = ENTRY_STATE_NOTINCACHE;
-    ec->ep_type = CACHE_TYPE_ENTRY;
+    ec->ep_type = CACHE_TYPE_UNKNOWN;
 #ifdef LDAP_CACHE_DEBUG
     ec->debug_sig = 0x45454545;
 #endif
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -512,6 +512,7 @@ lru_delete(struct cache *cache, void *ptr)
     }
     e = (struct backcommon *)ptr;
     PR_ASSERT(e->ep_state & ENTRY_STATE_LRU);
+    PR_ASSERT(e->ep_state & ENTRY_STATE_PINNED);
 
 #ifdef LDAP_CACHE_DEBUG_LRU
     pinned_verify(cache, __LINE__);
--- a/ldap/servers/slapd/back-ldbm/cache.c
+++ b/ldap/servers/slapd/back-ldbm/cache.c
@@ -511,11 +511,13 @@ lru_delete(struct cache *cache, void *ptr)
         return;
     }
     e = (struct backcommon *)ptr;
+    PR_ASSERT(e->ep_state & ENTRY_STATE_LRU);
 
 #ifdef LDAP_CACHE_DEBUG_LRU
     pinned_verify(cache, __LINE__);
     lru_verify(cache, e, 1);
 #endif
+    e->ep_state &= ~ENTRY_STATE_LRU;
     if (e->ep_lruprev)
         e->ep_lruprev->ep_lrunext = e->ep_lrunext;
     else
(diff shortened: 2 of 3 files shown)


ldap/servers/slapd/back-ldbm/backentry.c
```diff
@@ -25,6 +25,7 @@ backentry_free(struct backentry **bep)
     ep = *bep;
 
     PR_ASSERT(ep->ep_state & (ENTRY_STATE_DELETED | ENTRY_STATE_NOTINCACHE | ENTRY_STATE_INVALID));
+    PR_ASSERT((ep->ep_state & ENTRY_STATE_LRU) == 0);
```
tbordaz: Could we test (ENTRY_STATE_LRU | ENTRY_STATE_PINNED) ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -511,11 +511,13 @@ lru_delete(struct cache *cache, void *ptr)
         return;
     }
     e = (struct backcommon *)ptr;
+    PR_ASSERT(e->ep_state & ENTRY_STATE_LRU);
```
tbordaz, line 514: PR_ASSERT((ep->ep_state & ENTRY_STATE_PINNED) == 0); ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -543,11 +545,14 @@ lru_add(struct cache *cache, void *ptr)
     e = (struct backcommon *)ptr;
     ASSERT(e->ep_refcnt == 0);
     ASSERT((e->ep_state & ENTRY_STATE_PINNED) == 0);
+    PR_ASSERT((e->ep_state & ENTRY_STATE_LRU) == 0);
```
tbordaz, line 549: Are we using ASSERT or PR_ASSERT ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -837,8 +842,10 @@ entrycache_flush(struct cache *cache)
                           "Unexpected NULL entry while flushing cache - LRU list may be corrupted\n");
             break;
         }
+        PR_ASSERT(e->ep_state & ENTRY_STATE_LRU);
```
tbordaz, line 846: PR_ASSERT(e->ep_type != CACHE_TYPE_ENTRY); ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -837,8 +842,10 @@ entrycache_flush(struct cache *cache)
...
         ASSERT(e->ep_refcnt == 0);
         e->ep_refcnt++;
+        e->ep_state &= ~ENTRY_STATE_LRU;
```
tbordaz, line 849: Shouldn't we drop the flag only after successful entrycache_remove ?

ldap/servers/slapd/back-ldbm/cache.c
```diff
@@ -1647,7 +1655,7 @@ entrycache_return(struct cache *cache, struct backentry **bep, PRBool locked)
                 backentry_free(bep);
             } else {
                 pinned_verify(cache, __LINE__);
-                if (!pinned_add(cache, e)) {
+                if (!pinned_add(cache, e) && ((e->ep_state & ENTRY_STATE_LRU) == 0)) {
```
tbordaz, line 1659: My understanding is that because of the previous assert this test is useless when this function is called from cache_return. So the only other path I found is from flush_hash() that is triggered during betxn plugin failure. That would explain why we did not see this bug before

tbordaz (approved):
LGTM
Minor comments

ldap/servers/slapd/back-ldbm/cache.c
progier389: The test is mostly doing searches after having setting up  the database so I doubt there is betxn failure involved.
According to Claude AI there is a a failure path involving pinned_add /pinned_flush / pinned_remove:

Here's what happens when the entry has ep_refcnt == 0:

1. **Thread calls entrycache_return()** with entry e (ep_refcnt=0, ep_state=0)
   - Holds cache lock throughout

2. **Calls pinned_add(cache, e)** at line 1658
   - pinned_add() tries to add entry to pinned list
   - Calls pinned_flush() at line 1311, 1342, or 1370

3. **pinned_flush() calls pinned_remove()** on some OTHER entry (cache->c_pinned_ctx->head)
   - pinned_remove() at line 1267 checks: if (e->ep_refcnt == 0)
   - **If the removed pinned entry has ep_refcnt == 0, it calls lru_add()!**
   - This sets ENTRY_STATE_LRU on the removed entry

4. **Now the problem**: If the entry being removed from pinned list happens to be
   THE SAME ENTRY that entrycache_return() is trying to return, then:
   - Entry already has ENTRY_STATE_LRU set (from step 3)
   - When pinned_add() returns false (line 1658)
   - entrycache_return() calls lru_add() AGAIN at line 1659
   - lru_add() asserts that ENTRY_STATE_LRU == 0
   - **ASSERTION FAILS** because the flag is already set!

ldap/servers/slapd/back-ldbm/cache.c
progier389: No: the entry get detached out of the LRU queue even if entrycache_remove_int fails

ldap/servers/slapd/back-ldbm/cache.c
progier389: We could add such assertions but that is a bit outside of this issue

ldap/servers/slapd/back-ldbm/cache.c
progier389: ASSERT is unconditional while PR_ASSERT is only tested in DEBUG build (on RHEL) 
Not sure we want the new assertion in production build that is why I used  PR_ASSERT
everywhere.
IMHO only e->ep_refcnt == 0 should really be in ASSERT

ldap/servers/slapd/back-ldbm/cache.c
tbordaz: Excellent !! Thanks for the explanation.
I think the code is confusing
`if (!pinned_add(cache, e) && ((e->ep_state & ENTRY_STATE_LRU) == 0)) {
  lru_add(cache, e);
}`
reading it I did not realized that pinned_add can update ep_state.
Would it be possible to add comment and split it in something like:
```
pinned_entry = pinned_add(cache, e);
/* In case the entry was not pinned then
   pinned_add may have already moved to entry to the LRU
   Check it is not already there before adding it
 */
if (!pinned_entry &&  ((e->ep_state & ENTRY_STATE_LRU) == 0) {
   lru_add(cache, e);
}
```

PR #7613, opened 2026-06-26
Issue 7611 - PBKDF2 password verification should reject invalid iteration count

### LINKED ISSUE #7611: PBKDF2 password verification should reject invalid iteration counts
**Issue Description**
The PBKDF2 password verification code accepts the iteration count embedded in the stored password hash without enforcing an upper bound.

 - Package and version: All versions

jchapma:
Description:
The C PBKDF2 password verification code does not validate the extracted iteration count before invoking PBKDF2 hash function.

Fix:
Reject hashes whose iteration count is outside the supported range before before performing password verification.

Fixes: https://github.com/389ds/389-ds-base/issues/7611

Reviewed by:

## Summary by Sourcery

Validate PBKDF2 password hashes before verification and add tests for invalid hash inputs.

Bug Fixes:
- Reject PBKDF2 password hashes whose iteration count falls outside the supported range during verification.
- Reject PBKDF2 password hashes whose decoded payload length is invalid or cannot be base64-decoded before verification.

Tests:
- Add unit tests covering PBKDF2 password comparison with out-of-range iterations, malformed payload lengths, and invalid base64 input.


ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -57,6 +57,8 @@
  * it's still better than ssha512.
  */
 #define PBKDF2_MINIMUM 2048
+/* Upper bound for iteration count */
+#define PBKDF2_MAX_ITERATIONS 1000000
```
tbordaz: The fix looks good to me but first thing I think 1M is too high. Also I wonder if we should not use PBKDF2_ITERATIONS as MAX_ITERATIONS.
To support higher iteration we may use a config parameter (disabled by default) allowing higher iterations than PBKDF2_ITERATIONS.

progier389: Current recommended number of iterations (According to https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html ) are:
  - 600K for PBKDF2-HMAC-SHA256:
  - 220K for PBKDF2-HMAC-SHA512
But they tend to increase every years ...

So I agree with Thierry that we should have a configurable limit rather than an hard one
 (maybe using PBKDF2_ITERATIONS as default (but that value is quite obsolete)
IMHO the maximum number of iteration should change whether we are using PBKDF2-HMAC-SHA512 or PBKDF2-HMAC-SHA256
So we should better have 2 configurables maximum values ...

jchapma: Thank you both for your comments. This PR is for the C {PBKDF2_SHA256} plugin, so I'll focus on that here. 

I agree the hardcoded 1M limit is not ideal. I'll drop it and use the auto tuned PBKDF2_ITERATIONS as the default compare limit. I'll also add optional nsslapd-pwdPBKDF2VerifyMaxIterations on the plugin entry so admins can raise the compare limit if they need to. I can mention the OWASP recommendation as guidance in this case.

The Rust pwdchan path ({PBKDF2-SHA*}) will be use a similar approach in a separate PR.

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -57,8 +57,10 @@
  * it's still better than ssha512.
  */
 #define PBKDF2_MINIMUM 2048
+#define PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR "nsslapd-pwdPBKDF2AcceptMaxIterations"
 
 static uint32_t PBKDF2_ITERATIONS = 8192;
+static uint32_t PBKDF2_ACCEPT_MAX_ITERATIONS = 0;
```
progier389: Using uppercase name for variable is misleading:
 Such names are usually used for constants (like #define) 
IMHO we should better use camelcase as for the other variables
 (Btw should probably replace PBKDF2_ITERATIONS too. I suspect it was originally a constant then moved as a variable when the number of iteration was made configurable)

progier389 (approved):

Issue #7505, opened 2026-05-21
RFE - CLI - add feature to determine which password policy applies to a user


mreynolds389:
**Issue Description**

Update CLI, either dsconf or dsidm, where you can enter a user's DN and find out what password policies are apply to them (global or local) and the policy settings

mreynolds389:
1a057485b..4e1c7a6f0  389-ds-base-3.2 -> 389-ds-base-3.2
5ebaf8d9a..bcc02c3ed  389-ds-base-3.1 -> 389-ds-base-3.1
16d814e12..a3584f06b  389-ds-base-3.0 -> 389-ds-base-3.0

PR #7614, opened 2026-06-26
Issue 7505 - RFE - add feature to determine which password policy applies to a user

mreynolds389:
Description:

Update CLI where you can enter a user's DN and find out what password policies are apply to them (global or local) and the policy settings.

Design doc:

https://www.port389.org/docs/389ds/design/dsidm-user-get-pwp-design.html

Relates: https://github.com/389ds/389-ds-base/issues/7505

co-authored-by: Cursor

## Summary by Sourcery

Add support for resolving and displaying the effective password policy that applies to a user, and expose it via CLI and Cockpit UI.

New Features:
- Introduce dsidm `user get-pwp` command to report the effective password policy for a user, with text and JSON output options.
- Add Cockpit UI modal to view a user's effective password policy from search results, tree view, and editor views, backed by a dsidm-based helper.
- Detect user entries and derive lookup parameters for password policy resolution based on entry objectClasses and attributes.

Enhancements:
- Extend password policy management library to compute effective policies including global, user, and subtree policies, handling inheritance of syntax settings and canonical attribute names.
- Refactor Cockpit LDAP search result parsing into a reusable helper and enrich it with user metadata for password policy lookups.
- Document the new `dsidm user get-pwp` capability in the lib389 README.

Tests:
- Add CLI tests covering dsidm `user get-pwp` behavior across user types, policy types, inheritance scenarios, JSON output, and error handling.
- Add unit tests for password policy target parsing, classification, attribute merging, and reporting order.

### DIFF
--- a/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
+++ b/src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
@@ -221,6 +221,89 @@ function getResourceLimits () {
     return limits;
 }
 
+function parseSearchResult(searchResult) {
+    const lines = searchResult.split('\n');
+    const allEntries = [];
+    let ldapsubentry = false;
+    let isRole = false;
+    let isLockable = false;
+    let objectclasses = [];
+    let dn = '';
+    let uid = '';
+    let cn = '';
+    let numSubordinates = '0';
+    let modifyTimestamp = '';
+    lines.map(currentLine => {
+        const accountObjectclasses = ['nsaccount', 'nsperson', 'simplesecurityobject',
+            'organization', 'person', 'account', 'organizationalunit',
+            'netscapeserver', 'domain', 'posixaccount', 'shadowaccount',
+            'posixgroup', 'mailrecipient', 'nsroledefinition'];
+        if (isAttributeLine(currentLine, 'dn:')) {
+            // Convert base64-encoded DNs
+            const pos = currentLine.indexOf(':');
+            if (currentLine.startsWith('dn::')) {
+                dn = b64DecodeUnicode(currentLine.substring(pos + 2).trim());
+            } else {
+                dn = currentLine.substring(pos + 1).trim();
+            }
+            ldapsubentry = false;
+            isRole = false;
+            isLockable = false;
+            objectclasses = [];
+            uid = '';
+            cn = '';
+        } else if (isAttributeLine(currentLine, 'numSubordinates:')) {
+            numSubordinates = (currentLine.split(':')[1]).trim();
+        } else if (isAttributeLine(currentLine, 'modifyTimestamp:')) {
+            modifyTimestamp = (currentLine.split(':')[1]).trim();
+        } else if (isAttributeLine(currentLine, 'uid:')) {
+            uid = currentLine.substring(currentLine.indexOf(':') + 1).trim();
+        } else if (isAttributeLine(currentLine, 'cn:')) {
+            cn = currentLine.substring(currentLine.indexOf(':') + 1).trim();
+        } else if (currentLine.toLowerCase().startsWith('objectclass:')) {
+            const ocVal = currentLine.substring(currentLine.indexOf(':') + 1).trim().toLowerCase();
+            objectclasses.push(ocVal);
+            if (ocVal === 'ldapsubentry') {
+                ldapsubentry = true;
+            } else if (ocVal === 'nsroledefinition') {
+                isRole = true;
+            }
+        }
+        for (const accountOC of accountObjectclasses) {
+            if (isAttributeLine(currentLine, `objectclass: ${accountOC}`)) {
+                isLockable = true;
+            }
+        }
+
+        if (currentLine === '' && dn !== '') {
+            const userPwpInfo = buildEntryUserPwpInfo(dn, objectclasses, uid, cn);
+            const result = JSON.stringify(
+                {
+                    dn,
+                    numSubordinates,
+                    modifyTimestamp: getModDate(modifyTimestamp, "utc"),
+                    modifyTimestampLocal: getModDate(modifyTimestamp, "local"),
+                    ldapsubentry,
+                    isRole,
+                    isLockable,
+                    isUser: userPwpInfo.isUser,
+                    userPwpLookup: userPwpInfo.userPwpLookup,
+                });
+            allEntries.push(result);
+
+            // Reset the variables:
+            dn = '';
+            numSubordinates = '0';
+            modifyTimestamp = '';
+            objectclasses = [];
+            uid = '';
+            cn = '';
+        }
+        return [];
+    });
+    return allEntries;
+}
+
 export function getSearchEntries (params, resultCallback) {
     /*
      params.serverId,
@@ -229,7 +312,7 @@ export function getSearchEntries (params, resultCallback) {
      params.searchScope
      params.sizeLimit,
      params.timeLimit
-  */
+    */
     const cmd = [
         'ldapsearch',
         '-LLL',
@@ -253,11 +336,7 @@ export function getSearchEntries (params, resultCallback) {
     ];
 
     log_cmd("getSearchEntries", "", cmd);
-    let dn = '';
-    let numSubordinates = '0';
-    let modifyTimestamp = '';
     let searchResult = null;
-    const allEntries = [];
     cockpit
             .spawn(cmd, { superuser: "require", err: 'message' }) // string.split("\n\r")
             .done(data => {
@@ -280,61 +359,7 @@ export function getSearchEntries (params, resultCallback) {
                 if (searchResult === null) {
                     return;
                 }
-                const lines = searchResult.split('\n');
-                let ldapsubentry = false;
-                let isRole = false;
-                let isLockable = false;
-                lines.map(currentLine => {
-                    const accountObjectclasses = ['nsaccount', 'nsperson', 'simplesecurityobject',
-                        'organization', 'person', 'account', 'organizationalunit',
-                        'netscapeserver', 'domain', 'posixaccount', 'shadowaccount',
-                        'posixgroup', 'mailrecipient', 'nsroledefinition'];
-                    if (isAttributeLine(currentLine, 'dn:')) {
-                        // Convert base64-encoded DNs
-                        const pos = currentLine.indexOf(':');
-                        if (currentLine.startsWith('dn::')) {
-                            dn = b64DecodeUnicode(currentLine.substring(pos + 2).trim());
-                        } else {
-                            dn = currentLine.substring(pos + 1).trim();
-                        }
-                        ldapsubentry = false;
-                        isRole = false;
-                        isLockable = false;
-                    } else if (isAttributeLine(currentLine, 'numSubordinates:')) {
-                        numSubordinates = (currentLine.split(':')[1]).trim();
-                    } else if (isAttributeLine(currentLine, 'modifyTimestamp:')) {
-                        modifyTimestamp = (currentLine.split(':')[1]).trim();
-                    } else if (isAttributeLine(currentLine, 'objectclass: ldapsubentry')) {
-                        ldapsubentry = true;
-                    } else if (isAttributeLine(currentLine, 'objectclass: nsroledefinition')) {
-                        isRole = true;
-                    }
-                    for (const accountOC of accountObjectclasses) {
-                        if (isAttributeLine(currentLine, `objectclass: ${accountOC}`)) {
-                            isLockable = true;
-                        }
-                    }
-
-                    if (currentLine === '' && dn !== '') {
-                        const result = JSON.stringify(
-                            {
-                                dn,
-                                numSubordinates,
-                                modifyTimestamp: getModDate(modifyTimestamp, "utc"),
-                                modifyTimestampLocal: getModDate(modifyTimestamp, "local"),
-                                ldapsubentry,
-                                isRole,
-                                isLockable,
-                            });
-                        allEntries.push(result);
-
-                        // Reset the variables:
-                        dn = '';
-                        numSubordinates = '0';
-                        modifyTimestamp = '';
-                    }
-                    return [];
-                });
+                const allEntries = parseSearchResult(searchResult);
                 // Process the list of entries.
                 resultCallback(allEntries, null);
             });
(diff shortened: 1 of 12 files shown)


src/lib389/README.md
```diff
@@ -86,5 +86,20 @@ Please see our [contributing guide](https://www.port389.org/docs/389ds/contribut
 The 389 Directory Server is subject to the terms detailed in the
 license agreement file called LICENSE in the main project directory.
 
+## dsidm: user get-pwp
```
progier389: I wonder if this file is the right place for this description (should not it be rather in some design document ?)

mreynolds389: > I wonder if this file is the right place for this description (should not it be rather in some design document ?)

Cursor added this change and I thought that perhaps we should have been updating this file as changes were made to lib389,  but I didn't realize how empty this file was :-)  I'm going to remove it as its definitely out of place

progier389 (approved):
LGTM

src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
```diff
@@ -1158,3 +1123,107 @@ export function getBaseDNFromTree (entrydn, treeViewRootSuffixes) {
...
+export function getBaseDnForEntry(entryDn, suffixList) {
+    if (!entryDn || !suffixList || suffixList.length === 0) {
+        return entryDn || "";
+    }
+    const dnLower = entryDn.toLowerCase();
+    let bestMatch = suffixList[0];
+    let found = false;
+    for (const suffix of suffixList) {
+        const suffixLower = suffix.toLowerCase();
+        if (dnLower.endsWith(suffixLower) && suffix.length > bestMatch.length) {
```
droideck: I think this misses the case where the matching suffix is the first or only suffix in `suffixList`. `bestMatch` starts as `suffixList[0]`, but `found` only becomes true when a later suffix is longer.

Image

src/lib389/lib389/pwpolicy.py
```diff
@@ -74,6 +251,172 @@ def __init__(self, instance):
...
+            else:
+                local.append(attr)
+        return local + inherited
+
+    def resolve_user_dn(self, basedn, user_type, selector):
+        """Resolve a dsidm user selector to an entry DN."""
+        if not selector:
+            raise ValueError('A user identifier is required')
+        accounts_class = USER_TYPE_ACCOUNTS.get(user_type or 'posix', nsUserAccounts)
+        accounts = accounts_class(self._instance, basedn)
```
droideck: Resolving through the type-specific `*UserAccounts` means `get-pwp` only works for entries under that type's default container, e.g. `ou=people` or `ou=Services`, so it will probably fail for any other.

droideck (changes requested):

src/lib389/lib389/pwpolicy.py
mreynolds389: > Resolving through the type-specific `*UserAccounts` means `get-pwp` only works for entries under that type's default container, e.g. `ou=people` or `ou=Services`, so it will probably fail for any other.

This works with all user "types" that dsidm supports.  Seems reasonable to me.  If you don't use dsidm to create users then you can not use it manage those "other" entries.  Am I missing a scenario where dsidm created users do not work this feature?

src/cockpit/389-console/src/lib/ldap_editor/lib/utils.jsx
mreynolds389: Weird it's completely broken now.  I have the same instance I used when developing the feature, and I only ever tested with a single suffix defined.   It worked fine last week, but now it's broken.  I wonder if a rebase broke it?  Ok let me investigate...

mreynolds389: > Weird it's completely broken now. I have the same instance I used when developing the feature, and I only ever tested with a single suffix defined. It worked fine last week, but now it's broken. I wonder if a rebase broke it? Ok let me investigate...

Oh I know what happened.  Sourcery gave me a suggestion to combine two duplicate functions.  Errr, ok working on it...

src/lib389/lib389/pwpolicy.py
droideck: Fair point! I'd narrow my concern to a Cockpit UX mismatch then: it may show "View Password Policy" for user-like entries outside dsidm-managed bases, where `dsidm user get-pwp` then correctly returns "not found"
Maybe we can disable the UI button for incorrect entries, WDYT?

droideck (approved):
Besides minor and non-blocking issue, LGTM!!

src/lib389/lib389/pwpolicy.py
mreynolds389: I think I can make some improvements here:

- UI created "basic" user does not show option for "Viewing pwp"
    - Need to improve UI's "user" detection (I can also apply your suggestion here)
- UI users created under non-standard locations shows option for "Viewing pwp", but it fails
    - Probably need new CLI/dsidm option to specify non-standard DIT location

jchapma (approved):
Tests run fine, UI changes look good, LGTM

mreynolds389:
@droideck ok I added a new "--parent-dn" argument to specify custom/non-default DIT locations.  UI is now using it.  I think everything is covered now, please review.

src/lib389/lib389/idm/user.py
```diff
@@ -95,7 +95,7 @@ class nsUserAccounts(DSLdapObjects):
     :type rdn: str
     """
 
-    def __init__(self, instance, basedn, rdn='ou=people'):
+    def __init__(self, instance, basedn, rdn=DEFAULT_BASEDN_RDN):
```
droideck: I think the change broke a couple of tests in GitHub CI (SSO token and some dn asserts)

src/lib389/lib389/pwpolicy.py
```diff
@@ -74,6 +254,175 @@ def __init__(self, instance):
...
+
+    def get_effective_policy(self, basedn, user_type, selector, parent_basedn=None):
+        """Resolve the effective password policy for a user account."""
+        user_dn = self.resolve_user_dn(basedn, user_type, selector, parent_basedn)
+        user_entry = Account(self._instance, user_dn)
+        user_attrs = user_entry.get_all_attrs_utf8()
+        if not user_attrs:
+            raise ValueError('User entry "{}" was not found'.format(user_dn))
+
+        pwp_subentry_vals = user_attrs.get('pwdpolicysubentry', [])
```
droideck: Just noticed, do I understand correctly that even if `nsslapd-pwpolicy-local` is off, we'll still report the local password policy if `pwdpolicysubentry` is present?

src/lib389/lib389/pwpolicy.py
```diff
@@ -74,6 +254,175 @@ def __init__(self, instance):
...
+                local.append(attr)
+        return local + inherited
+
+    def resolve_user_dn(self, basedn, user_type, selector, parent_basedn=None):
+        """Resolve a dsidm user selector to an entry DN."""
+        if not selector:
+            raise ValueError('A user identifier is required')
+        accounts_class = USER_TYPE_ACCOUNTS.get(user_type or 'posix', nsUserAccounts)
+        if parent_basedn is not None:
+            accounts = accounts_class(self._instance, parent_basedn, rdn=None)
```
droideck, line 290: I think ServiceAccount might have an issue with `rdn=None`... It will build a base like `None,<parent_dn>` because it handles it differently than other `accounts_class`es.

src/lib389/lib389/pwpolicy.py
mreynolds389: > Just noticed, do I understand correctly that even if `nsslapd-pwpolicy-local` is off, we'll still report the local password policy if `pwdpolicysubentry` is present?

It will, ok I'll look into that

src/lib389/lib389/pwpolicy.py
mreynolds389: > I think ServiceAccount might have an issue with `rdn=None`... It will build a base like `None,<parent_dn>` because it handles it differently than other `accounts_class`es.

There should be coverage for "service" accounts in the CI, but I'll double check it

src/lib389/lib389/idm/user.py
mreynolds389: > I think the change broke a couple of tests in GitHub CI (SSO token and some dn asserts)

Hmm the only change here is the case:  "ou=people" vs "ou=People".  I'll check those tests and hardened them up.

src/lib389/lib389/pwpolicy.py
droideck: I think the issue is still present, unless I miss something?

src/lib389/lib389/pwpolicy.py
```diff
@@ -74,6 +254,175 @@ def __init__(self, instance):
...
+            local_syntax_on = _attr_is_on(effective['passwordchecksyntax']['values'][0])
+        else:
+            local_syntax_on = False
+
+        inherit_global = _attr_is_on(
+            _first_attr_value(global_attrs_raw, 'nsslapd-pwpolicy-inherit-global', 'off'))
+        global_syntax_on = _attr_is_on(
+            _first_attr_value(global_attrs_raw, 'passwordchecksyntax', 'off'))
+
+        if not local_syntax_on and inherit_global and global_syntax_on:
```
droideck, line 385: IIUC, when the local policy does not enable syntax, while global syntax and inheritance are both enabled, the server replaces the local syntax fields with the global values. This merge currently preserves an explicitly stored local field because of the `attr in local_attrs` check. I think, for example, it can report a local minimum length of 12 while the server enforces the global value of 16.

Also, I think it's worth adding a test case that covers `not local_syntax_on and inherit_global and global_syntax_on` setup.

src/lib389/lib389/pwpolicy.py
droideck: I think we might still have the issue (note, it's for `--parent-dn` + `ServiceAccount` case, not for the default Service lookup)
And in the new Cockpit model it always supplies the parent DN. 
But yeah, adding one explicit-parent service test would be nice too:)

droideck (changes requested):
I see a couple of password policy tests fail. That might be related to this PR...

```
FAILED dirsrvtests/tests/suites/password/password_policy_test.py::test_duplicate_pwpolicy
FAILED dirsrvtests/tests/suites/password/password_policy_test.py::test_subtree_pwpolicy_cos_repair
```

Also, I see some `Validate tests and minimal Python version / validate (pull_request)` failures.

src/lib389/lib389/pwpolicy.py
mreynolds389: OK the fix will be in the next PR commit

mreynolds389:
> I see a couple of password policy tests fail. That might be related to this PR...
> 
> ```
> FAILED dirsrvtests/tests/suites/password/password_policy_test.py::test_duplicate_pwpolicy
> FAILED dirsrvtests/tests/suites/password/password_policy_test.py::test_subtree_pwpolicy_cos_repair
> ```
> 
> Also, I see some `Validate tests and minimal Python version / validate (pull_request)` failures.

The tests pass for me locally, and I did a fresh rebase.

src/lib389/lib389/pwpolicy.py
mreynolds389: > IIUC, when the local policy does not enable syntax, while global syntax and inheritance are both enabled, the server replaces the local syntax fields with the global values. This merge currently preserves an explicitly stored local field because of the `attr in local_attrs` check. I think, for example, it can report a local minimum length of 12 while the server enforces the global value of 16.
> 
> Also, I think it's worth adding a test case that covers `not local_syntax_on and inherit_global and global_syntax_on` setup.

Added a test and this does work correctly

src/lib389/lib389/pwpolicy.py
mreynolds389: Yeah there was a problem with service accounts.  Ok I think everything is fixed and I added more tests.  Please review....

droideck (approved):
LGTM!
One minor issue still remains, though: docstring validation tests still fail. 

```
dirsrvtests/tests/suites/password/pwpolicy_effective_test.py
============================================================

test_parse_policy_target_user:27
--------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_parse_policy_target_subtree:32
-----------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_parse_policy_target_invalid:37
-----------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_classify_user_policy:41
----------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_classify_subtree_policy:51
-------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_syntax_inheritance_merge:60
--------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_local_policy_only_explicit_attrs:89
----------------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_local_policy_camelcase_attrs:106
-------------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_inheritance_with_camelcase_global:123
------------------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

test_report_attribute_order_inherited_last:143
----------------------------------------------

* Missing docstring.
* Docstring should have at least id token(s)

dirsrvtests/tests/suites/clu/dsidm_user_get_pwp_test.py
=======================================================

test_get_pwp_global_policy_by_user_type:184
-------------------------------------------

* Docstring should have at least id token(s)

test_get_pwp_global_policy:201
------------------------------

* Docstring should have at least id token(s)

test_get_pwp_subtree_policy:214
-------------------------------

* Docstring should have at least id token(s)

test_get_pwp_user_policy:384
----------------------------

* Docstring should have at least id token(s)

test_get_pwp_user_policy_precedence:400
---------------------------------------

* Docstring should have at least id token(s)

test_get_pwp_syntax_inheritance_on:415
--------------------------------------

* Docstring should have at least id token(s)

test_get_pwp_syntax_inheritance_off:432
---------------------------------------

* Docstring should have at least id token(s)

test_get_pwp_json_output:500
----------------------------

* Docstring should have at least id token(s)

test_get_pwp_json_inherited_syntax:512
--------------------------------------

* Docstring should have at least id token(s)

test_get_pwp_user_not_found:530
-------------------------------

* Docstring should have at least id token(s)

test_get_pwp_broken_policy_reference:542
----------------------------------------

* Docstring should have at least id token(s)
```

Issue #7621, opened 2026-07-01
Stack Buffer Overflow in Password checkPrefix


mreynolds389:
**Issue Description**

A stack buffer overflow in `checkPrefix()` function allows a Directory Manager to crash the LDAP server by storing a reversible-encrypted attribute with an oversized algorithm ID.

Example

```
ldapmodify -H ldapi:// -Y EXTERNAL <<LDIF
   dn: cn=poc-003,cn=replica,cn=dc\3Dtest\2Cdc\3Dcom,cn=mapping tree,cn=config
   changetype: add
   objectclass: top
   objectclass: nsds5ReplicationAgreement
   cn: poc-003
   nsds5ReplicaHost: localhost
   nsds5ReplicaPort: 9999
   nsds5ReplicaRoot: dc=test,dc=com
   nsds5ReplicaBindDN: cn=replication manager,cn=config
   nsds5ReplicaCredentials: {AES-AAAA...<512 A's>...}dGVzdA==
   nsds5ReplicaBindMethod: SIMPLE
   nsds5ReplicaTransportInfo: LDAP
   LDIF
```

mreynolds389:
ce7c2d470..f06e40b24  389-ds-base-3.2 -> 389-ds-base-3.2
46b8a6967..11abd296d  389-ds-base-3.1 -> 389-ds-base-3.1
0f49ec55f..9423da425  389-ds-base-3.0 -> 389-ds-base-3.0
099c0e7ad..aa96d9ae5  389-ds-base-2.9 -> 389-ds-base-2.9
5fbae3279..77d5e59a8  389-ds-base-2.8 -> 389-ds-base-2.8
8ba1f42f8..0615c767b  389-ds-base-2.7 -> 389-ds-base-2.7
066573dbe..dfba58f42  389-ds-base-2.6 -> 389-ds-base-2.6
9eb3eb07b..dec7e63ae  389-ds-base-2.4 -> 389-ds-base-2.4
b0aa7434d..56b294703  389-ds-base-2.2 -> 389-ds-base-2.2
1cfcf5419..37ecdee4e  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7622, opened 2026-07-01
Issue 7621 - Stack Buffer Overflow in Password checkPrefix

mreynolds389:
Description:

A stack buffer overflow in checkPrefix() function allows a Directory Manager to crash the LDAP server by storing a reversible-encrypted attribute with an oversized algorithm ID.

Add a bounds check before the `memcpy` at `pw.c:466` to prevent the overflow.

relates: https://github.com/389ds/389-ds-base/issues/7621

## Summary by Sourcery

Prevent stack buffer overflow in password prefix handling and add a regression test to ensure the server remains stable when processing oversized algorithm IDs.

Bug Fixes:
- Add bounds checking in checkPrefix() to reject algorithm IDs that exceed the local buffer size instead of overflowing the stack.

Tests:
- Introduce an automated test that creates a replication agreement with an oversized password prefix and verifies the LDAP server continues running.


progier389 (approved):
LGTM

Issue #7623, opened 2026-07-01
Heap Buffer Overflow in 389-ds-base Audit Log Password Masking


mreynolds389:
**Issue Description**

When password storage scheme is set to CLEAR and you set a password that is less than 23-bytes then a heap overflow is possible.

mreynolds389:
7a0e8966f..ce7c2d470  389-ds-base-3.2 -> 389-ds-base-3.2
dbd0c5a5e..46b8a6967  389-ds-base-3.1 -> 389-ds-base-3.1
115fa2d13..0f49ec55f  389-ds-base-3.0 -> 389-ds-base-3.0
d2b95bf5b..099c0e7ad  389-ds-base-2.9 -> 389-ds-base-2.9
7459a0490..5fbae3279  389-ds-base-2.8 -> 389-ds-base-2.8
5138aa6d4..8ba1f42f8  389-ds-base-2.7 -> 389-ds-base-2.7
b8ef6ecc1..066573dbe  389-ds-base-2.6 -> 389-ds-base-2.6
1b1fda509..9eb3eb07b  389-ds-base-2.4 -> 389-ds-base-2.4
e8ac453cc..b0aa7434d  389-ds-base-2.2 -> 389-ds-base-2.2
d90d83715..1cfcf5419  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7624, opened 2026-07-01
Issue 7623  - Heap Buffer Overflow in 389-ds-base Audit Log Password Masking

mreynolds389:
Description:

When password storage scheme is set to CLEAR and you set a password that is less than 23-bytes then a possible heap overflow is possible.

Replace the unsafe `strcpy` with a bounded copy that respects the available space in the buffer

Severity: Low (CVSS 3.3)
CWE: CWE-122 (Heap-based Buffer Overflow)

relates: https://github.com/389ds/389-ds-base/issues/7623

## Summary by Sourcery

Prevent heap buffer overflow in audit log password masking and add regression coverage for ASAN-detected overflows.

Bug Fixes:
- Fix potential heap-based buffer overflow when masking short CLEAR passwords in audit log entries.

Tests:
- Add ASAN-gated audit log test to ensure password masking does not cause heap-buffer-overflow with short CLEAR passwords.

### DIFF
--- a/ldap/servers/slapd/auditlog.c
+++ b/ldap/servers/slapd/auditlog.c
@@ -106,7 +106,25 @@ create_masked_entry_string(Slapi_Entry *original_entry, int *len)
 
             /* Check if this is a password attribute that needs masking */
             if (is_password_attribute(line_start)) {
-                strcpy(colon_pos + 1, " **********************");
+                const char mask[] = " **********************";
+                size_t mask_len = (sizeof mask)-1;
+                size_t avail = 0;
+
+                /* Calculate available space from colon_pos+1 to end of value */
+                if (next_line != NULL) {
+                    avail = (size_t)(next_line - (colon_pos + 1));
+                } else {
+                    avail = strlen(colon_pos + 1);
+                }
+
+                if (mask_len <= avail) {
+                    memcpy(colon_pos + 1, mask, mask_len);
+                    /* Space-fill remaining space to avoid leaking original value */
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/auditlog.c
```diff
@@ -106,7 +106,25 @@ create_masked_entry_string(Slapi_Entry *original_entry, int *len)
...
+                /* Calculate available space from colon_pos+1 to end of value */
+                if (next_line != NULL) {
+                    avail = (size_t)(next_line - (colon_pos + 1));
+                } else {
+                    avail = strlen(colon_pos + 1);
+                }
+
+                if (mask_len <= avail) {
+                    memcpy(colon_pos + 1, mask, mask_len);
+                    /* NUL-fill remaining space to avoid leaking original value */
```
progier389: Comment should be: Space-fill instead of NUL-fill

ldap/servers/slapd/auditlog.c
```diff
@@ -106,7 +106,25 @@ create_masked_entry_string(Slapi_Entry *original_entry, int *len)
 
             /* Check if this is a password attribute that needs masking */
             if (is_password_attribute(line_start)) {
-                strcpy(colon_pos + 1, " **********************");
+                const char *mask = " **********************";
```
progier389: I wonder if the following would not be better:
```
  const char maskj[] = " **********************";
  size_t mask_len = (sizeof mask)-1;
```
( so the size get computed by the compiler rather than by strlen  )

progier389 (approved):
Looks good except 2 minor points

PR #7628, opened 2026-07-02
Issue 7627 - When it exists configured matching rule for an indexed a…

### LINKED ISSUE #7518: Replica install leaves parentid index in inconsistent state causing ONELEVEL searches to return no entries until reindex
# Replica install leaves parentid index in inconsistent state causing ONELEVEL searches to return no entries until reindex

### Source server

* RHEL 8.10
* ipa-server-4.9.13
* 389-ds-base-1.4.3.39

### Replica server

* RHEL 9.8
* ipa-server-4.13.1
* 389-ds-base-2.8.0

## Problem

Fresh `ipa-replica-install` consistently fails with:

```text
EmptyResult: no matching entry found
```

during:

```text
dns_update_system_records()
 -> server_find()
 -> server_role_find()
 -> ldap2.get_entries()
```

The traceback shows the failure occurring in:

```python
ipaserver/plugins/server.py
ipaserver/plugins/serverrole.py
ipaserver/servroles.py
```

while enumerating IPA masters.

---

## Observed behavior

Immediately after replica install failure:

### ONELEVEL search returns no entries

```bash
ldapsearch \
  -H ldapi://%2Frun%2Fslapd-EXAMPLE-COM.socket \
  -Y GSSAPI \
  -LLL \
  -b "cn=masters,cn=ipa,cn=etc,dc=example,dc=com" \
  -s one \
  "(objectClass=*)"
```

Returns:

```text
0 entries
```

### SUBTREE search returns expected entries

```bash
ldapsearch \
  -H ldapi://%2Frun%2Fslapd-EXAMPLE-COM.socket \
  -Y GSSAPI \
  -LLL \
  -b "cn=masters,cn=ipa,cn=etc,dc=example,dc=com" \
  -s sub \
  "(objectClass=*)"
```

Returns:

```text
cn=ipa1.example.com,...
cn=ipa2.example.com,...
cn=ipa3.example.com,...
...
```

### Source server behaves correctly

Running the same ONELEVEL search on the source server returns the expected child entries.

The problem is only observed on the newly created replica before reindexing.

---

## Entry relationships appear correct

Parent entry:

```text
dn: cn=masters,cn=ipa,cn=etc,dc=example,dc=com
entryid: 2008
numSubordinates: 3
```

Replica master entry:

```text
dn: cn=ipa3.example.com,cn=masters,cn=ipa,cn=etc,dc=example,dc=com
parentid: 2008
entryid: 2619
```

The parent-child relationship appears correct:

```text
child.parentid == parent.entryid
```

---

## Additional observations

389-ds reports:

```text
MISMATCH - parentid index has integerOrderingMatch configured,
but on-disk data uses lexicographic ordering.
This will cause searches to return incorrect or incomplete results.
```

Example:

```text
Backend 'userRoot': MISMATCH - parentid index has integerOrderingMatch configured,
but on-disk data uses lexicographic ordering.
This will cause searches to return incorrect or incomplete results.
Please reindex the parentid attribute.
```

---

## Critical observation

The parent entry reports:

```text
dn: cn=masters,cn=ipa,cn=etc,dc=example,dc=com
numSubordinates: 3
```

However, a ONELEVEL search against the same DN returns no entries:

```bash
ldapsearch \
  -H ldapi://%2Frun%2Fslapd-EXAMPLE-COM.socket \
  -Y GSSAPI \
  -LLL \
  -b "cn=masters,cn=ipa,cn=etc,dc=example,dc=com" \
  -s one \
  "(objectClass=*)"
```

Result:

```text
0 entries
```

At the same time:

* SUBTREE searches return the expected entries.
* Child entries exist and have the correct `parentid`.
* The parent entry reports `numSubordinates: 3`.

This appears inconsistent because the server reports existing children, yet a ONELEVEL search against the same entry returns no children.

---

## Workaround 1: Reindex parentid

Running:

```bash
dsconf INSTANCE backend index reindex \
  --wait \
  userRoot \
  --attr parentid
```

immediately restores ONELEVEL search behavior.

After reindex:

```bash
ldapsearch \
  -b "cn=masters,cn=ipa,cn=etc,dc=example,dc=com" \
  -s one \
  "(objectClass=*)"
```

returns the expected entries.

The replica installation can then proceed if the IPA code path is retried.

---

## Workaround 2: Change IPA search scope

The installer fails while enumerating IPA masters.

As an experiment, changing the LDAP search scope in IPA from:

```python
ldap.SCOPE_ONELEVEL
```

to:

```python
ldap.SCOPE_SUBTREE
```

allows the installation to complete successfully.

This strongly suggests that the installation failure is triggered by the incorrect result returned by the ONELEVEL search.

This workaround does not explain why the ONELEVEL search returns zero entries when:

* the entries exist,
* `numSubordinates` is non-zero,
* and SUBTREE searches return the expected results.

---

## Additional note

Reindexing restores ONELEVEL search behavior, but the following warning may still be reported after restart:

```text
MISMATCH - parentid index has integerOrderingMatch configured,
but on-disk data uses lexicographic ordering.
```

In other words:

```text
Reindex fixes the search behavior.
The MISMATCH warning may still remain.
```

---

## Reproducible sequence

```text
Fresh replica install
  ->
ONELEVEL search returns 0 entries
  ->
ipa-replica-install fails with EmptyResult
  ->
reindex parentid
  ->
ONELEVEL search immediately works
```

The LDAP entries themselves are present and searchable through SUBTREE scope.

The issue appears related to the parentid index state immediately after replica initialization and affects ONELEVEL search behavior.

### LINKED ISSUE #7627: When it exists configured matching rule for an indexed attribute it should be used if the schema does not define one
**Issue Description**
An attribute is defined in the schema. The matching rule is optional in the schema so the definition falls back to the default matching rule that is lexicographic order.

An attribute can be configured to be indexed and the configuration can define a matching rule 'nsMatchingRule".

If the schema does not define a matching rule, the server should fall back to the configured one ('nsMatchingRule') prior to default lexicographic one

All versions

**Steps to Reproduce**
Will be provided in a testcase

**Expected results**
Configured MR should be the default if the schema does not define one

tbordaz:
…ttribute it should be used if the schema does not define one

Bug description:
	An attribute is defined in the schema. The matching rule is optional in the schema
	so the definition falls back to the default matching rule that is lexicographic order.

	An attribute can be configured to be indexed and the configuration can define a
	matching rule 'nsMatchingRule".

	If the schema does not define a matching rule, the server should fall back to the
	configured one ('nsMatchingRule') prior to default lexicographic one

Fix description:
	At startup when binding the attribute their schema definitions
	if no matching rule is define and it existed a configured matching rule
	the default matching rule should be the configured one

fixes: #7627
fixes: #7518

Reviewed by:

## Summary by Sourcery

Honor configured ordering matching rules when schema definitions lack compatible rules and improve detection and coverage of index ordering behavior.

Bug Fixes:
- Use an indexed attribute’s configured ordering matching rule when the schema does not provide a compatible ordering rule, while preserving schema-compatible behavior.

Enhancements:
- Add index validation for detecting indexes configured with integer ordering whose on-disk ordering remains lexicographic.
- Expand regression coverage for configured matching-rule precedence, integer ordering, collation indexing, and range searches.

Tests:
- Add regression tests covering matching-rule fallback, index ordering mismatch detection, collation-based indexing, and lexicographic versus integer range behavior.

### DIFF
--- a/dirsrvtests/tests/suites/indexes/regression_test.py
+++ b/dirsrvtests/tests/suites/indexes/regression_test.py
@@ -13,6 +13,7 @@ import ldap
 import logging
 import glob
 import re
+import json
 from lib389.backend import Backend, Backends, DatabaseConfig
 from lib389.cli_ctl.dblib import DbscanHelper
 from lib389.config import LDBMConfig
@@ -23,6 +24,7 @@ from lib389.dirsrv_log import DirsrvErrorLog
 from lib389.idm.domain import Domain
 from lib389.idm.group import Groups
 from lib389.idm.nscontainer import nsContainer
+from lib389.idm.organizationalunit import OrganizationalUnits
 from lib389.idm.user import UserAccount, UserAccounts
 from lib389.index import Index
 from lib389._mapped_object import DSLdapObject, DSLdapObjects
@@ -37,6 +39,8 @@ pytestmark = pytest.mark.tier1
 SUFFIX2 = 'dc=example2,dc=com'
 BENAME2 = 'be2'
 CN_INDEX_DN = f'cn=cn,cn=index,cn={DEFAULT_BENAME},cn=ldbm database,cn=plugins,cn=config'
+PARENTID_INDEX_DN = f'cn=parentid,cn=index,cn={DEFAULT_BENAME},cn=ldbm database,cn=plugins,cn=config'
+UID_INDEX_DN = f'cn=uid,cn=index,cn={DEFAULT_BENAME},cn=ldbm database,cn=plugins,cn=config'
 
 DEBUGGING = os.getenv("DEBUGGING", default=False)
 logging.getLogger(__name__).setLevel(logging.INFO)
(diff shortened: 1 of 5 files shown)


ldap/servers/slapd/back-ldbm/ldbm_attr.c
```diff
@@ -940,6 +940,25 @@ attr_index_config(
         if (k > 0) {
             a->ai_index_rules = official_rules;
             a->ai_indexmask |= INDEX_RULES;
+            if (!slapi_matchingrule_is_ordering(official_rules[0], attrsyntax_oid) &&
+                slapi_matchingrule_can_use_compare_fn(attrValue->bv_val) &&
```
droideck: IIUC, after the loop `attrValue` holds the last nsMatchingRule value, while the other checks and `plugin_mr_find()` use `official_rules[0]`. With a multi-valued nsMatchingRule (say an ordering rule plus a substr spec) the fallback could silently skip depending on value order. Should this be `official_rules[0]` too?

Should we also check that `official_rules[0]` is actually an ordering rule? I think `!slapi_matchingrule_is_ordering(...)` is also true for plain equality rules, so on an attr with no schema ORDERING an equality RULE index (caseIgnoreMatch etc.) would get picked up as the key comparator and change the index ordering for configs that were fine before.

ldap/servers/slapd/back-ldbm/ldbm_attr.c
```diff
@@ -940,6 +940,25 @@ attr_index_config(
...
+                !a->ai_sattr.a_mr_ord_plugin) {
+                    /* the configured ordering matching rule (nsMatchingRule) is NOT compatible with the attribute syntax
+                     * In addition we have not identified a matching rule so we are falling back to the configured one
+                     */
+
+                    /* set the ordering matching rule */
+                    a->ai_sattr.a_mr_ord_plugin = plugin_mr_find(official_rules[0]);
```
droideck: Wondering about upgrades... An instance already running with this misconfig has the index built in lex order, and after upgrade the comparator changes with no config change, so lookups could silently miss keys until a reindex. Should we log a "reindex needed" warning may be?

dirsrvtests/tests/suites/indexes/regression_test.py
```diff
@@ -1398,6 +1401,213 @@ def test_large_multivalued_sn_attribute(topo):
...
+        assert any(k >= 10 for k in keys), (
+            f"Expected a multi-digit parentid key to distinguish integer vs "
+            f"lexicographic order, got {keys}"
+        )
+        integer_sorted = sorted(keys)
+        lex_sorted = sorted(keys, key=str)
+        assert keys == integer_sorted, (
+            f"parentid equality keys are not in integer order: {keys}"
+        )
+        assert integer_sorted != lex_sorted, (
```
droideck, line 1468: I think this can trip when the whole module runs: topo is module scoped, earlier tests use up a few hundred entry ids, so the OU ids likely all have the same digit count and integer order == lex order. Then this assert fails before the fix is exercised (and when they coincide, the order checks pass even without the fix). Maybe we can run it in a fresh backend (like BENAME2 earlier in the file) so ids start at 1 and the divergence is guaranteed?

droideck (changes requested):

ldap/servers/slapd/back-ldbm/ldbm_attr.c
tbordaz: Thanks @droideck for your careful review and nice comment.
Indeed, after a second look,  I think the I need to rewrite that part of code.

I think the design (current) is primary to support several indexing syntax (nsMatchingRule) but I have a doubt how it works for ordering as (IIUC) the index access is limited to NEXT

ldap/servers/slapd/match.c
```diff
@@ -224,6 +224,22 @@ slapi_matchingrule_unregister(char *oid __attribute__((unused)))
...
+    for (mrl = g_get_global_mrl(); mrl != NULL; mrl = mrl->mrl_next) {
+        if (mrl->mr_entry->mr_name && !strcasecmp(oid_or_name, mrl->mr_entry->mr_name)) {
+            return (mrl->mr_entry->mr_name &&
+                    PL_strcasestr(mrl->mr_entry->mr_name, "ordering"));
+        }
+        if (mrl->mr_entry->mr_oid && !strcmp(oid_or_name, mrl->mr_entry->mr_oid)) {
+            return (mrl->mr_entry->mr_name &&
+                    PL_strcasestr(mrl->mr_entry->mr_name, "ordering"));
+        }
+    }
```
droideck, line 241: I think we're missing a return 0; after the loop. If the name/OID isn't registered we fall off the end and the caller gets an undefined value. The old inline loop was protected by is_compat, but this one is called directly, so a typo in nsMatchingRule (or a collation language tag like fr) would reach it. Since it's exported in slapi-plugin.h, maybe a NULL guard on the argument as well?

ldap/servers/slapd/back-ldbm/ldbm_attr.c
```diff
@@ -886,6 +887,23 @@ attr_index_config(
                 do_continue = 1;                /* done with j - next j */
             }
 
+           if (!do_continue &&
+               slapi_matchingrule_is_ordering_only(attrValue->bv_val) &&
```
droideck: I believe this also catches collation rules. They're registered as `caseIgnoreOrderingMatch-<lang>`, so the "ordering" test passes, plugin_mr_find() returns NULL (the collation plugin has no plg_mr_names), and do_continue = 1 skips the RULE path. If I'm reading it right, nsMatchingRule: 2.16.840.1.113730.3.3.2.11.1 on cn/sn/uid would stop creating the index after upgrade. What do you think about only taking this branch when plugin_mr_find() returns a plugin with plg_mr_compare, and leaving do_continue at 0 otherwise so the RULE path still runs?

ldap/servers/slapd/back-ldbm/ldbm_attr.c
```diff
@@ -886,6 +887,23 @@ attr_index_config(
...
+               !a->ai_sattr.a_mr_ord_plugin) {
+                /* The ordering matching rule is not compatible with the attribute syntax
+                 * pick the plugin that supports this ordering matching rule and use it.
+                 * if there are multiple ordering matching rules for the same attribute,
+                 * use the last one that is found.
+                 */
+                mr_ordering_plugin = plugin_mr_find(attrValue->bv_val);
```
droideck: Small one: mr_ordering_plugin is overwritten on every value, including with NULL when the lookup fails. With `[integerOrderingMatch, <collation OID>]` on one index we'd end up with neither a comparator nor a RULE index. Maybe assign only on success?

ldap/servers/slapd/back-ldbm/ldbm_attr.c
```diff
@@ -985,6 +1003,25 @@ attr_index_config(
             a->ai_key_cmp_fn = NULL;
         }
     }
+    if (mr_ordering_plugin && (a->ai_sattr.a_mr_ord_plugin == NULL)) {
```
droideck: Two thoughts here.
This block doesn't check need_compare_fn, so with a compatible plus an incompatible ordering rule on the same index the incompatible one wins (uidNumber with integerOrderingMatch and caseIgnoreOrderingMatch would get a string comparator). Skipping the block when we already have a comparator would probably be enough.

Also, this is the upgrade case from my earlier comment and IMO it deserves a WARNING with a reindex hint rather than INFO. An existing lexicographic index gets opened with int_compare and equality lookups miss keys until db2index.
One more thing I noticed: for string syntax the keys aren't integer-normalized, so =007 and =7 compare equal and merge into one key, and (uid=7) would return the 007 entry because indexed equality skips the filter test. Would it make sense to keep the can_use_compare_fn gate and limit this to what parentid needs?
But I don't have a strong opinion here and maybe it's something we do intentionally.

src/lib389/lib389/cli_ctl/dbtasks.py
```diff
@@ -436,6 +436,31 @@ def dbtasks_index_check(inst, log, args):
...
+                    disk_ordering = _check_disk_ordering(db_dir, backend, index_name, dbscan_path, is_mdb, log)
+                    if disk_ordering == IndexOrdering.UNKNOWN:
+                        log.info("  %s - config: %s, disk: %s",
+                                 index_name, "integer", disk_ordering.value)
+                        log.info("  %s - could not determine disk ordering, skipping", index_name)
+                        continue
+                    if disk_ordering == IndexOrdering.LEXICOGRAPHIC:
+                        log.info("  %s - config: %s, disk: %s",
+                                 index_name, "integer", disk_ordering.value)
+                        log.warning("  %s - MISMATCH: config has integerOrderingMatch but disk is lexicographic (suggest redindex)", index_name)
```
droideck, line 455:
```suggestion
                        log.warning("  %s - MISMATCH: config has integerOrderingMatch but disk is lexicographic (suggest reindex)", index_name)
```

src/lib389/lib389/cli_ctl/dbtasks.py
```diff
@@ -436,6 +436,31 @@ def dbtasks_index_check(inst, log, args):
...
+                config_has_int_order = _has_integer_ordering_match(dse_ldif, backend, index_name)
+                if config_has_int_order:
+                    # Check disk ordering
+                    disk_ordering = _check_disk_ordering(db_dir, backend, index_name, dbscan_path, is_mdb, log)
+                    if disk_ordering == IndexOrdering.UNKNOWN:
+                        log.info("  %s - config: %s, disk: %s",
+                                 index_name, "integer", disk_ordering.value)
+                        log.info("  %s - could not determine disk ordering, skipping", index_name)
+                        continue
+                    if disk_ordering == IndexOrdering.LEXICOGRAPHIC:
```
droideck, line 452: Should we also set `all_ok = False` here when reporting the `MISMATCH`?

src/lib389/lib389/cli_ctl/dbtasks.py
```diff
@@ -436,6 +436,31 @@ def dbtasks_index_check(inst, log, args):
...
                     all_ok = False
 
+
+        for index_name in dse_ldif.get_indexes(backend):
+            # for indexes others than  parentid or ancestorid
+            if index_name != "parentid" and index_name != "ancestorid":
+                config_has_int_order = _has_integer_ordering_match(dse_ldif, backend, index_name)
+                if config_has_int_order:
+                    # Check disk ordering
+                    disk_ordering = _check_disk_ordering(db_dir, backend, index_name, dbscan_path, is_mdb, log)
```
droideck, line 446: _check_disk_ordering() only looks at the first 100 dbscan lines. For attributes with a sub index those are all */+ keys, so it returns UNKNOWN and we silently skip cn/uid on real deployments, IIUC.
Not new code, but now that it runs on user indexes it probably needs to count = keys instead of raw lines...

droideck (changes requested):

ldap/servers/slapd/back-ldbm/ldbm_attr.c
tbordaz: Good catch. collation rule can not be used for comparison. It it is a valid index rule. The patch prevents to use it as an index. I need to rework the patch.
I agree the problem here is that do_continue is set to 1. Now the collation plugin is not a valid comparison and the patch should prevent to use it via mr_ordering_plugin.

ldap/servers/slapd/back-ldbm/ldbm_attr.c
tbordaz: I think the an attribute should have only one comparison (ordering) nsMatchingRule. If multiple ordering are set, it looks to me an invalid config and taking the last value is acceptable.

ldap/servers/slapd/match.c
tbordaz: Agree

ldap/servers/slapd/back-ldbm/ldbm_attr.c
tbordaz: Right with multiple nsMatchingRule values we can have need_compare_fn and mr_ordering_plugin both set. I need to rework the patch for multiple ordering nsMatchingRule.

The upgrade I was thinking of heathcheck being responsible to report incoherent index ordering versus ordering config.

A string syntax attribute should not normalize '007' into '7'. So if the search filter is '(uid=007)' the equality index should not return the key '=7'. How did you get this behavior ?

droideck (approved):
LGTM! Thank you!

PR #7632, opened 2026-07-07
Issue 7611 - PBKDF2 password verification should reject invalid iteration counts

### LINKED ISSUE #7611: PBKDF2 password verification should reject invalid iteration counts

jchapma:
Description:
The Rust pwdchan PBKDF2 password verification code does not validate the extracted iteration count before invoking the PBKDF2 hash function.

Fix:
Add nsslapd-pwdPBKDF2AcceptMaxIterations configuration for each PBKDF2 plugin variant and reject stored hashes whose iteration count is below the minimum or above the configured accept maximum before password verification. Add lib389 and dsconf support to get, set, and delete the accept max setting.

Depends on the schema change adding nsslapd-pwdPBKDF2AcceptMaxIterations to pwdPBKDF2PluginConfig.

Fixes: https://github.com/389ds/389-ds-base/issues/7611

Assisted by: Cursor

Reviewed by:

## Summary by Sourcery

Enforce validation of PBKDF2 iteration counts during configuration and password verification and introduce a configurable upper bound for accepted iterations per PBKDF2 variant.

New Features:
- Add per-variant configuration of maximum accepted PBKDF2 iterations for stored hashes via the nsslapd-pwdPBKDF2AcceptMaxIterations attribute.
- Expose CLI and lib389 support to get, set, and delete the PBKDF2 accept-max-iterations setting for each PBKDF2 plugin variant.

Bug Fixes:
- Reject PBKDF2 password hashes whose stored iteration count is below the minimum or above the configured maximum during verification.

Enhancements:
- Refine PBKDF2 plugin configuration handling by sharing parsing/validation logic for numeric attributes and improving logging around invalid configuration and stored hashes.
- Extend unit tests to cover new accept-max-iterations behavior and validation of iteration bounds in PBKDF2 decomposition and comparison.

### DIFF
--- a/src/plugins/pwdchan/src/lib.rs
+++ b/src/plugins/pwdchan/src/lib.rs
@@ -26,6 +26,13 @@ const MAX_PBKDF2_ROUNDS: usize = 10_000_000;
 
 const PBKDF2_ROUNDS_ATTR: &str = "nsslapd-pwdPBKDF2NumIterations";
 const PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR: &str = "nsslapd-pwdPBKDF2AcceptMaxIterations";
+
+// Scheme identifiers, used for per-scheme config and atomics below.
+const SCHEME_PBKDF2: &str = "PBKDF2";
+const SCHEME_PBKDF2_SHA1: &str = "PBKDF2-SHA1";
+const SCHEME_PBKDF2_SHA256: &str = "PBKDF2-SHA256";
+const SCHEME_PBKDF2_SHA512: &str = "PBKDF2-SHA512";
+
 // Each algorithm gets its own atomic counter for thread-safe round and accept max iterations
 static PBKDF2_ROUNDS: AtomicUsize = AtomicUsize::new(DEFAULT_PBKDF2_ROUNDS);
 static PBKDF2_ROUNDS_SHA1: AtomicUsize = AtomicUsize::new(DEFAULT_PBKDF2_ROUNDS);
@@ -89,7 +96,7 @@ mod pbkdf2 {
 
     impl super::Pbkdf2Plugin for PwdChanPbkdf2 {
         fn digest_type() -> MessageDigest { MessageDigest::sha1() }
-        fn scheme_name() -> &'static str { "PBKDF2" }
+        fn scheme_name() -> &'static str { SCHEME_PBKDF2 }
     }
 }
 
@@ -101,7 +108,7 @@ mod pbkdf2_sha1 {
 
     impl super::Pbkdf2Plugin for PwdChanPbkdf2Sha1 {
         fn digest_type() -> MessageDigest { MessageDigest::sha1() }
-        fn scheme_name() -> &'static str { "PBKDF2-SHA1" }
+        fn scheme_name() -> &'static str { SCHEME_PBKDF2_SHA1 }
     }
 }
 
@@ -113,7 +120,7 @@ mod pbkdf2_sha256 {
 
     impl super::Pbkdf2Plugin for PwdChanPbkdf2Sha256 {
         fn digest_type() -> MessageDigest { MessageDigest::sha256() }
-        fn scheme_name() -> &'static str { "PBKDF2-SHA256" }
+        fn scheme_name() -> &'static str { SCHEME_PBKDF2_SHA256 }
     }
 }
 
@@ -125,7 +132,7 @@ mod pbkdf2_sha512 {
 
     impl super::Pbkdf2Plugin for PwdChanPbkdf2Sha512 {
         fn digest_type() -> MessageDigest { MessageDigest::sha512() }
-        fn scheme_name() -> &'static str { "PBKDF2-SHA512" }
+        fn scheme_name() -> &'static str { SCHEME_PBKDF2_SHA512 }
     }
 }
 
@@ -143,7 +150,7 @@ macro_rules! impl_slapi_pbkdf2_plugin {
 
             fn start(pb: &mut PblockRef) -> Result<(), PluginError> {
                 log_error!(ErrorLevel::Trace, "{} plugin start", Self::scheme_name());
-                PwdChanCrypto::handle_pbkdf2_config(pb, Self::digest_type())?;
+                PwdChanCrypto::handle_pbkdf2_config(pb, Self::scheme_name())?;
                 Ok(())
             }
 
@@ -161,11 +168,11 @@ macro_rules! impl_slapi_pbkdf2_plugin {
             }
 
             fn pwd_storage_encrypt(cleartext: &str) -> Result<String, PluginError> {
-                PwdChanCrypto::pbkdf2_encrypt(cleartext, Self::digest_type())
+                PwdChanCrypto::pbkdf2_encrypt(cleartext, Self::digest_type(), Self::scheme_name())
             }
 
             fn pwd_storage_compare(cleartext: &str, encrypted: &str) -> Result<bool, PluginError> {
-                PwdChanCrypto::pbkdf2_compare(cleartext, encrypted, Self::digest_type())
+                PwdChanCrypto::pbkdf2_compare(cleartext, encrypted, Self::digest_type(), Self::scheme_name())
             }
         }
     };
@@ -178,13 +185,13 @@ impl_slapi_pbkdf2_plugin!(pbkdf2_sha256::PwdChanPbkdf2Sha256);
 impl_slapi_pbkdf2_plugin!(pbkdf2_sha512::PwdChanPbkdf2Sha512);
 
 impl PwdChanCrypto {
-    fn validate_pbkdf2_config_rounds(rounds: usize) -> Result<(), PluginError> {
-        if rounds < MIN_PBKDF2_ROUNDS || rounds > MAX_PBKDF2_ROUNDS {
+    fn validate_pbkdf2_rounds(value: usize) -> Result<(), PluginError> {
+        if value < MIN_PBKDF2_ROUNDS || value > MAX_PBKDF2_ROUNDS {
             #[cfg(not(test))]
             log_error!(
                 ErrorLevel::Error,
-                "Invalid PBKDF2 number of iterations {}, must be between {} and {}",
-                rounds,
+                "Invalid PBKDF2 iteration count {}, must be between {} and {}",
+                value,
                 MIN_PBKDF2_ROUNDS,
                 MAX_PBKDF2_ROUNDS
             );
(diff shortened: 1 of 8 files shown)


src/plugins/pwdchan/src/lib.rs
```diff
@@ -288,75 +356,89 @@ impl PwdChanCrypto {
...
-            source
         );
+
+        let mut accept_max = rounds;
+
+        if entry.get_attr(PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR).is_some() {
+            accept_max = Self::parse_config_usize_attr(&entry, PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR)?;
+        }
+
+        Self::set_pbkdf2_accept_max(digest, accept_max)?;
```
droideck: I think we should validate `accept_max >= rounds` before storing here (and peobably cover that combination with a test)
pbkdf2_encrypt() embeds the configured rounds in new hashes, while validate_pbkdf2_stored_iterations() rejects counts above the cap. Allowing `rounds > accept_max` would therefore let the plugin create hashes that its own comparison path rejects.

src/plugins/pwdchan/src/lib.rs
```diff
@@ -288,75 +356,89 @@ impl PwdChanCrypto {
...
+            d if d == MessageDigest::sha1() => &PBKDF2_ACCEPT_MAX_SHA1,
+            d if d == MessageDigest::sha256() => &PBKDF2_ACCEPT_MAX_SHA256,
+            d if d == MessageDigest::sha512() => &PBKDF2_ACCEPT_MAX_SHA512,
+            _ => &PBKDF2_ACCEPT_MAX,
+        }
+    }
+
+    fn digest_name(digest: MessageDigest) -> &'static str {
+        match digest {
+            d if d == MessageDigest::sha1() => "PBKDF2-SHA1",
```
droideck: PBKDF2 and PBKDF2-SHA1 both return `sha1()` from digest_type(), so this function maps both configurations to the same atomic. As dsconf configures them independently, I think, we need to process the situation more explicitly for the user somehow

src/plugins/pwdchan/src/lib.rs
```diff
@@ -413,6 +495,73 @@ impl PwdChanCrypto {
         // If not in test mode or no thread-local value, use global
         Ok(Self::get_rounds_atomic(digest).load(Ordering::Relaxed))
     }
+
+    fn set_pbkdf2_accept_max(digest: MessageDigest, accept_max: usize) -> Result<(), PluginError> {
+        if accept_max < MIN_PBKDF2_ROUNDS {
```
droideck: I think we need to add an upper bound here that fits the OpenSSL API too.
The locked openssl 0.10.79 implementation converts iter from usize to c_int using try_into().unwrap(), while this setter currently accepts values outside that range. At minimum, IMO, we could validate against `i32::MAX`; reusing MAX_PBKDF2_ROUNDS may be simpler if that is the intended policy.

src/plugins/pwdchan/src/lib.rs
```diff
@@ -288,75 +356,89 @@ impl PwdChanCrypto {
...
-        } else {
-            "Unknown"
-        };
-
+        let digest_name = Self::digest_name(digest);
         log_error_ext!(
             ErrorLevel::Info,
             digest_name,
-            "Number of iterations set to {} from {}",
+            "Number of iterations set to {}",
```
droideck: Any particular reason to remove the `from {}",` % `default or configuration`  functionality?

droideck (changes requested):
Overall -- looks nice!!
A few clarifications before the ack:

src/plugins/pwdchan/src/lib.rs
jchapma: Good catch.  I will add  accept_max < rounds  validation in set_pbkdf2_accept_max and handle_pbkdf2_config and a unit test.

src/plugins/pwdchan/src/lib.rs
jchapma: Yes i agree, rounds and accept-max will be keyed by scheme_name() instead of digest_type(), so {PBKDF2} and {PBKDF2-SHA1} have independent config even though both use SHA-1.

src/plugins/pwdchan/src/lib.rs
jchapma: I agree, using MAX_PBKDF2_ROUNDS  is the way to go

src/plugins/pwdchan/src/lib.rs
jchapma: Initially i wanted the log to show whether the value was the default or from an updated config, but I changed my mind and removed it. I thought logging the effective value was enough.

droideck: I mean I added it in the initial implementation -- https://github.com/389ds/389-ds-base/commit/b8e442bc43975b7654596761a04b4f3a8279d5be
I'd rather keep the PR focused on the fix. And we can discuss the deprecation later, I think, if you want:)

src/plugins/pwdchan/src/lib.rs
```diff
@@ -274,144 +341,245 @@ impl PwdChanCrypto {
...
-                })?;
-                source = "configuration";
-            }
+        if entry.get_attr(PBKDF2_ROUNDS_ATTR).is_some() {
+            rounds = Self::parse_config_usize_attr(&entry, PBKDF2_ROUNDS_ATTR)?;
+            source = "configuration";
         }
 
-        Self::set_pbkdf2_rounds(digest, rounds)?;
+        let mut accept_max = rounds;
```
droideck: Perhaps we could follow the approach used in https://github.com/389ds/389-ds-base/commit/0e3b049653bcc0d9f4deecca5051a10b567f6a4d 
Keep the verification ceiling independent from generation rounds, fall back safely on invalid configuration, cap newly generated hashes accordingly, and throttle rejection logs. For Rust, we would adapt this per scheme using its own limits, of course, and mirror the validation and restart notice in lib389/dsconf. This should preserve existing hashes when rounds are lowered and avoid startup regressions (hence fixing some of the tests that currently fail in the GH CI pwp_storage test suites).

droideck (changes requested):

droideck:
But also yeah, a lot of other tests fail in the `pwp_storage` suite.

src/plugins/pwdchan/src/lib.rs
```diff
@@ -285,144 +438,286 @@ impl PwdChanCrypto {
...
+        let (mut rounds, source) = match configured_rounds {
+            Some(value) if Self::validate_pbkdf2_rounds(value).is_ok() => {
+                (value, "configuration")
+            }
+            _ => (DEFAULT_PBKDF2_ROUNDS, "default"),
+        };
+
+        let accept_max = match configured_accept_max {
+            Some(value) if Self::validate_pbkdf2_rounds(value).is_ok() => value,
+            _ => DEFAULT_PBKDF2_ACCEPT_MAX,
```
droideck, line 553: Some CI tests still fail. I think using the approach from 0e3b049653bcc0d9f4deecca5051a10b567f6a4d should help. It uses a fixed, bounded verification ceiling that covers the rounds already supported, and cap generation to that same ceiling? Falling back to 100,000 breaks existing 600,000-round configs and hashes after an upgrade. For ordinary clients, injecting a few-millions-round pre-hash is already gated by nsslapd-allow-hashed-passwords, which is off by default, and permission to write userPassword; root/password admins and replication are the expected exceptions.

src/plugins/pwdchan/src/lib.rs
```diff
@@ -213,12 +353,20 @@ impl PwdChanCrypto {
...
-        let (iter, salt, hash_expected) = Self::pbkdf2_decompose(encrypted).map_err(|e| {
-            // This means our DB content is flawed.
-            log_error!(ErrorLevel::Error, "invalid hashed pw -> {:?}", e);
-            e
-        })?;
+        let (iter, salt, hash_expected) = Self::pbkdf2_decompose(encrypted, scheme)
+            .map_err(|e| {
+                match e {
+                    // Iteration bounds are logged in validate_pbkdf2_stored_iterations.
+                    PluginError::InvalidConfiguration => e,
```
droideck: The throttled message is bypassed here. 0e3b0496 handles an out-of-range hash as a normal comparison failure after logging it. IIUC, here we propagate InvalidConfiguration, and the generated callback logs every Err at macros.rs:348. That means every rejected bind still adds a generic error line. Probably, we can make it return Ok(false) once the rejection has been logged. What do you think?

src/plugins/pwdchan/src/lib.rs
```diff
@@ -285,144 +438,286 @@ impl PwdChanCrypto {
...
+    }
+
+    fn handle_pbkdf2_config(pb: &mut PblockRef, scheme: &str) -> Result<(), PluginError> {
+        // Keep verification ceiling independent of generation rounds.
+        let mut configured_rounds = None;
+        let mut configured_accept_max = None;
+
+        if let Ok(entry) = pb.get_op_add_entryref() {
+            if entry.get_attr(PBKDF2_ROUNDS_ATTR).is_some() {
+                match Self::read_config_usize_attr(&entry, PBKDF2_ROUNDS_ATTR) {
```
droideck, line 574: Could we keep nsslapd-pwdPBKDF2NumIterations fail-closed here? 0e3b0496 only uses the soft fallback for the new accept-max setting. Before this change, an invalid generation setting prevented the plugin from loading; now it starts with 100,000 rounds while LDAP still contains the invalid value. Soft fallback makes sense for accept-max, but changing the existing generation setting's behavior seems risky...

droideck (changes requested):

src/plugins/pwdchan/src/lib.rs
jchapma: Thank you for your comments, but before another round of changes I would like to lock down the final policy as I feel feedback has shifted across reviews. 

Earlier feedback asked for hard accept_max >= rounds validation so we would not create hashes that compare would reject. Later feedback asked us to follow 0e3b049 instead: independent verification ceiling, soft fallback, cap generation, and throttled reject logs. That is what the current commits implement, which is why the hard cross-check was dropped. Soft-fallback on rounds also came from reading “fall back safely on invalid configuration” broadly; I am fine narrowing that to accept-max only if that is the intent.

Can you confirm that this is the best approach:
- Accept-max: independent of rounds; soft-fallback on invalid/missing; fixed default high enough for existing deployments. Which value should we use (600000 vs MAX_PBKDF2_ROUNDS)? The current 100000 default matches generation’s default, but as you noted it breaks existing 600k configs/hashes.
- Rounds (nsslapd-pwdPBKDF2NumIterations): fail-closed again on invalid config (plugin start fails).
- Encrypt: cap rounds to accept-max when higher.
- Compare: throttled reject log, then Ok(false) (not Err(InvalidConfiguration)).

Thank you

droideck: Yeah, it's fair. The initial comments were before I fixed the regression, so the feedback shifted a bit as we understood the problem better. Sorry about the extra round.
Okay, let's lock this down. In short, I agree with your list.

For a missing or invalid accept-max, let's use MAX_PBKDF2_ROUNDS (10,000,000), not 600,000. The 600,000 value only comes from the regression test, while 10m is the highest setting the Rust plugin already allowed for generated hashes. This avoids breaking hashes the plugin could previously create while still adding a hard upper limit. It doesn't mean 10m is recommended, and anything above it remains rejected.

For fresh installs, I'd set accept-max to 600,000, probably, explicitly in the plugin entries/templates. That keeps new installations at the tighter (but not-too-much) limit while allowing upgraded installations without the new attribute to continue working.
A regular user also can't just inject a 10m pre-hashed value by default: nsslapd-allow-hashed-passwords is off, and normal write permissions still apply when it's enabled. Root/password admins and replication are the exceptions.

For nsslapd-pwdPBKDF2NumIterations, a missing value still defaults to 100k, while an invalid configured value should fail plugin startup as before.

Encrypt should cap to accept-max and log it, IMO.

For comparison, an iteration count below the minimum or above accept-max should produce a rate-limited rejection message and then be treated as a normal password mismatch with Ok(false). Returning Err for this case would make the callback log an additional generic error on every bind, bypassing the rate limit. Malformed hashes and internal/OpenSSL failures should still return Err and follow the normal error-logging path; they are not covered by this throttle.

Please also cover the missing accept-max + existing 600k hash case and the 10m boundary in the tests.

jchapma: No worries, I completely understand. 

Perfect, I will update the PR accordingly. Thank you

droideck (approved):
Looks great! Ack

P.S. there were some PR CI execution issues recently... Hopefully, now it's fixed after the rebase.

Once the CI finishes and it's clean, feel free to merge!

jchapma:
> Looks great! Ack
> 
> P.S. there were some PR CI execution issues recently... Hopefully, now it's fixed after the rebase.
> 
> Once the CI finishes and it's clean, feel free to merge!

Will do, thank you for your feedback.

PR #7634, opened 2026-07-08
Issue 7633 - RFE - Add offline diagnostics for thread pool saturation

### LINKED ISSUE #7633: RFE - Add offline diagnostics for thread pool saturation
**Description**
When the worker thread pool is saturated, admins lose the usual visibility into what the server is doing because `cn=monitor` also needs a worker thread to respond. The access log does not fully solve this, since long-running or stuck operations only get their `RESULT` line after they finish. This makes thread pool exhaustion hard to diagnose in real time and often leaves admins guessing whether the server is overloaded, stalled, or already dead.

We need an offline way to inspect the thread pool state without relying on LDAP request processing.

droideck:
Description: When the worker pool is fully saturated, cn=monitor cannot be used for diagnostics because the monitor search itself needs a worker thread.

Add offline thread-pool status reporting by publishing pool gauges and per-worker activity into a hardened memory-mapped file under the instance run directory. The new dsctl thread-pool-status command reads this file directly, without an LDAP connection, so admins can inspect the pool even when worker threads are exhausted.

cn=monitor also exposes a sanitized threadpoolworker attribute backed by the same data source. The feature is enabled by default and can be disabled with the new nsslapd-thread-pool-stats cn=config attribute. Changing this setting requires a restart.

Fixes: https://github.com/389ds/389-ds-base/issues/7633

Reviewed by: ?

## Summary by Sourcery

Add offline thread-pool saturation diagnostics backed by a hardened mmap file and expose sanitized per-worker status via both dsctl and cn=monitor.

New Features:
- Introduce a memory-mapped thread-pool status file under the instance run directory to expose pool gauges and per-worker activity without requiring LDAP worker threads.
- Add a dsctl thread-pool status CLI command (with text and JSON output and an optional --file input) to read and interpret thread-pool status files offline.
- Expose a sanitized threadpoolworker attribute on cn=monitor that reports per-worker state without revealing connection or operation identifiers.
- Add a configurable nsslapd-thread-pool-stats cn=config switch to enable or disable thread-pool diagnostics at startup.

Enhancements:
- Track per-worker lifecycle and operation activity in the connection threads and feed it into the shared thread-pool status mmap.
- Publish thread-pool metrics into cn=monitor by backing them with the shared status data source instead of only volatile in-memory counters.
- Persist and prune crash-time thread-pool status archives on startup to aid post-mortem diagnostics while bounding archive count.

Tests:
- Add an extensive monitor/threadpool_status_test.py suite covering mmap file lifecycle, dsctl output (including saturation and crash scenarios), symlink and corruption hardening, heartbeat staleness, configuration toggling, and archive pruning behavior.

### DIFF
--- a/ldap/servers/slapd/threadpool_stats.c
+++ b/ldap/servers/slapd/threadpool_stats.c
@@ -34,6 +34,15 @@
 #define TP_STATS_HEARTBEAT_INTERVAL_MS 1000
 #define TP_STATS_ARCHIVE_KEEP 5
 
+/*
+ * Crash archives are named <file>.YYYYMMDD-HHMMSS; the suffix is produced
+ * by strftime(TP_STATS_ARCHIVE_TIME_FMT). The lib389 reader matches the
+ * same pattern (\.\d{8}-\d{6}$).
+ */
+#define TP_STATS_ARCHIVE_TIME_FMT "%Y%m%d-%H%M%S"
+#define TP_STATS_ARCHIVE_DATE_LEN (sizeof("YYYYMMDD") - 1)
+#define TP_STATS_ARCHIVE_STAMP_LEN (sizeof("YYYYMMDD-HHMMSS") - 1)
+
 static tp_stats_header_t *tp_stats_header = NULL;
 static int tp_stats_fd = -1;
 static size_t tp_stats_len = 0;
@@ -61,11 +70,11 @@ tp_load_32(uint32_t *ptr, int memorder)
 }
 
 static tp_worker_slot_t *
-tp_stats_slot_at(tp_stats_header_t *header, uint32_t idx)
+tp_stats_slot_at(tp_stats_header_t *header, size_t idx)
 {
     return (tp_worker_slot_t *)((uint8_t *)header +
                                 TP_STATS_HEADER_SIZE +
-                                ((size_t)idx * TP_STATS_WORKER_SLOT_SIZE));
+                                (idx * TP_STATS_WORKER_SLOT_SIZE));
 }
 
 static tp_worker_slot_t *
@@ -213,11 +222,13 @@ tp_stats_name_cmp(const void *a, const void *b)
 static bool
 tp_stats_is_archive_suffix(const char *suffix)
 {
-    if (strlen(suffix) != 15 || suffix[8] != '-') {
+    if (suffix == NULL || strlen(suffix) != TP_STATS_ARCHIVE_STAMP_LEN ||
+        suffix[TP_STATS_ARCHIVE_DATE_LEN] != '-') {
         return false;
     }
-    for (int i = 0; i < 15; i++) {
-        if (i != 8 && !isdigit((unsigned char)suffix[i])) {
+    for (size_t i = 0; i < TP_STATS_ARCHIVE_STAMP_LEN; i++) {
+        /* Every character is a digit except the '-' after the date part */
+        if (i != TP_STATS_ARCHIVE_DATE_LEN && !isdigit((unsigned char)suffix[i])) {
             return false;
         }
     }
@@ -323,7 +334,7 @@ tp_stats_archive_crash_file(const char *path)
 
     now = slapi_current_utc_time();
     if (localtime_r(&now, &tms) == NULL ||
-        strftime(tbuf, sizeof(tbuf), "%Y%m%d-%H%M%S", &tms) == 0) {
+        strftime(tbuf, sizeof(tbuf), TP_STATS_ARCHIVE_TIME_FMT, &tms) == 0) {
         return;
     }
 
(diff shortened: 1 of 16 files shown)


droideck:
Design: https://github.com/389ds/389ds.github.io/pull/30

jchapma (approved):
This is a really nice feature, I like the offline mmap approach. 

A minor comment you can ignore, do we mention anywhere that a restart is required for nsslapd-thread-pool-stats to take effect. The value updates in cn=config immediately, but the running server keeps the diagnostics active until restart.

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,750 @@
...
+static int
+tp_stats_name_cmp(const void *a, const void *b)
+{
+    return strcmp(*(char *const *)a, *(char *const *)b);
+}
+
+static bool
+tp_stats_is_archive_suffix(const char *suffix)
+{
+    if (strlen(suffix) != 15 || suffix[8] != '-') {
```
tbordaz: May be check suffix not NULL and strlen>8.
I think 8 and 15 should be computed.

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,750 @@
...
+        return false;
+    }
+    for (int i = 0; i < 15; i++) {
+        if (i != 8 && !isdigit((unsigned char)suffix[i])) {
```
tbordaz: Do you mean i>=8 ?

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,750 @@
...
+    char *configdir = config_get_configdir();
+    char *instname = NULL;
+    char *path = NULL;
+
+    if (configdir != NULL) {
+        instname = PL_strrstr(configdir, "slapd-");
+    }
+
+    if (rundir != NULL && instname != NULL) {
+        path = slapi_ch_smprintf("%s/%s%s/%s", rundir, instname,
```
tbordaz, line 159: Creating a subdirectory, does it require a new selinux rule ?

tbordaz (approved):
LGTM (very impressive)
few comments

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,750 @@
...
+    return strcmp(*(char *const *)a, *(char *const *)b);
+}
+
+static bool
+tp_stats_is_archive_suffix(const char *suffix)
+{
+    if (strlen(suffix) != 15 || suffix[8] != '-') {
+        return false;
+    }
+    for (int i = 0; i < 15; i++) {
```
mreynolds389: Also please use size_t for "for" loops.

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,750 @@
...
+    vals[0] = &val;
+    vals[1] = NULL;
+    attrlist_delete(&e->e_attrs, TP_STATS_ATTR_THREADPOOL_WORKER);
+
+    if (header == NULL) {
+        return;
+    }
+
+    now_ns = tp_stats_mono_ns();
+    for (uint32_t i = 0; i < header->max_workers; i++) {
```
mreynolds389: size_t here was well - it should be compatible with max_workers

ldap/servers/slapd/threadpool_stats.c
droideck: Good point! Replaced the magic numbers with derived constants. The strftime call that produces the suffix now also uses the same TP_STATS_ARCHIVE_TIME_FMT define, so the producer and this validator can't drift apart. Added the NULL guard too...

ldap/servers/slapd/threadpool_stats.c
droideck: `i != 8` is intentional: the suffix is `YYYYMMDD-HHMMSS`, so index 8 is the `-` separator and every other position must be a digit. So `i >= 8` would stop validating the time digits and would reject the dash itself. But I replaced the magic number and added a comment.

ldap/servers/slapd/threadpool_stats.c
droideck: No new rule needed. `dirsrv.fc` labels the run directory recursively `/run/dirsrv(/.*)? ... dirsrv_var_run_t`. So the new `.monitor` subdirectory and the file inside it inherit dirsrv_var_run_t at creation time. 
Added a paragraph on this to the design doc.

ldap/servers/slapd/threadpool_stats.c
```diff
@@ -0,0 +1,761 @@
...
+
+static int
+tp_stats_name_cmp(const void *a, const void *b)
+{
+    return strcmp(*(char *const *)a, *(char *const *)b);
+}
+
+static bool
+tp_stats_is_archive_suffix(const char *suffix)
+{
```
tbordaz, line 225: Just to help maintenance/readability add a comment stating that the expected suffix is 'YYYYMMDD-HHMMSS'

tbordaz (approved):
Many thanks for you updates

PR #7640, opened 2026-07-10
Issue 7558 - Total init sends the suffix entry twice

### LINKED ISSUE #7558: During online import, the IDL should be created with in-depth first approach

### LINKED ISSUE #7604: CI - BDB replication regression_m2 count mismatch in test_moving_entry_make_online_init_fail
## Problem

`BDB Test (replication/regression_m2_test.py)` failed in main-branch CI with an entry-count mismatch after the nested move / replication sequence in `test_moving_entry_make_online_init_fail`.

Evidence:
- Run https://github.com/389ds/389-ds-base/actions/runs/27997232667
- Job https://github.com/389ds/389-ds-base/actions/runs/27997232667/job/82885887407

Observed failing case:
- `dirsrvtests/tests/suites/replication/regression_m2_test.py::test_moving_entry_make_online_init_fail`

Representative logs include:
- `assert 60 == 61`
- failure at `dirsrvtests/tests/suites/replication/regression_m2_test.py:888`
- `m1entries` length: 60
- `m2entries` length: 61
- the pytest repr for `m2entries` begins with two `dn: dc=example,dc=com` entries, suggesting the extra returned entry may be a duplicate suffix entry

The failing assertion is:

```python
m1entries = M1.search_s(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE,
                        '(|(objectclass=ldapsubentry)(objectclass=nstombstone)(nsuniqueid=*))')
m2entries = M2.search_s(DEFAULT_SUFFIX, ldap.SCOPE_SUBTREE,
                        '(|(objectclass=ldapsubentry)(objectclass=nstombstone)(nsuniqueid=*))')

assert len(m1entries) == len(m2entries)

droideck:
Description: Exclude the suffix entry from the depth-first parentid walk because total init sends it separately.
Preserve NEW_IDL_NO_ALLID across all parentid fetches and honor it in the LMDB fetch path to maintain parent-first ordering. Update LMDB reindex handling so entries with an explicit parentid are treated as regular entries when their RDN matches the suffix.

Add a regression test with a wide moved subtree that verifies entry order and ensures the suffix is sent only once.

Fixes: https://github.com/389ds/389-ds-base/issues/7558
Fixes: https://github.com/389ds/389-ds-base/issues/7604

Reviewed by: ?

## Summary by Sourcery

Ensure total initialization preserves tree order while avoiding duplicate suffix entries and maintains parent-first ordering across back-ldbm and LMDB paths.

Bug Fixes:
- Prevent the suffix entry from being sent and stored twice during online total initialization.
- Avoid collapsing parentid-based candidate lists to ALLIDS in LMDB when NEW_IDL_NO_ALLID is requested.
- Treat non-root entries with explicit parentid as regular entries during LMDB reindexing even if their RDN matches the suffix.

Enhancements:
- Refine parentid depth-first walks to exclude the root suffix entry from the bulk candidate list while still traversing all descendants.

Tests:
- Add a regression test that performs online init with a wide moved subtree and verifies the suffix is sent only once and entry counts match between suppliers.

### DIFF
--- a/dirsrvtests/tests/suites/replication/regression_m2_test.py
+++ b/dirsrvtests/tests/suites/replication/regression_m2_test.py
@@ -19,6 +19,7 @@ import time
 import random
 import string
 from shutil import rmtree
+from lib389.backend import DatabaseConfig
 from lib389.dbgen import dbgen_users
 from lib389.idm.user import TEST_USER_PROPERTIES, UserAccount, UserAccounts
 from lib389.pwpolicy import PwPolicyManager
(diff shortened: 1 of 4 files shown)


progier389 (approved):
LGTM

Issue #7639, opened 2026-07-09
Move log compression outside of global write lock


mreynolds389:
**Issue Description**

Current log compression holds the write lock, blocking all worker threads that are trying to write to that log type.

mreynolds389:
1f7cfe83a..a20b5fc79  389-ds-base-3.2 -> 389-ds-base-3.2
f29452d3b..44b81e61f  389-ds-base-3.1 -> 389-ds-base-3.1
bca1414f7..21766f918  389-ds-base-3.0 -> 389-ds-base-3.0

PR #7642, opened 2026-07-10
Issue 7639 - Move log compression outside of global write lock

mreynolds389:
Description:

Create a new log maintenance thread that handles log compression and deletion outside of the write lock. This prevents potentially long file operations from blocking worker threads from updating the active log file.

Design doc: https://www.port389.org/docs/389ds/design/log-async-compress-delete-design.html

relates: https://github.com/389ds/389-ds-base/issues/7639

Co-authored-by: Cursor

## Summary by Sourcery

Introduce asynchronous log maintenance to perform compression and retention outside of log write locks for all log streams, updating rotation metadata and shutdown handling accordingly.

Enhancements:
- Add a dedicated log maintenance worker thread and per-stream job queues to handle background log compression and retention sweeps.
- Distinguish rotation-time deletion from retention passes in log deletion routines, and refresh rotationinfo files and compressed size metadata whenever logs are compressed or removed.
- Update log lock/unlock macros so pending maintenance jobs are submitted after write locks are released, ensuring worker operations do not block logging.
- Adjust copyright years in logging-related sources and headers to reflect 2026.

Tests:
- Extend logging compression tests to account for asynchronous compression and deletion behavior, including waiting for background maintenance and verifying rotated/compressed log lifecycle.

### DIFF
--- a/dirsrvtests/tests/suites/logging/async_log_compress_chain_uaf_test.py
+++ b/dirsrvtests/tests/suites/logging/async_log_compress_chain_uaf_test.py
@@ -1,0 +1,107 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+import glob
+import logging
+import os
+import time
+
+import pytest
+
+from lib389._constants import DEFAULT_SUFFIX
+from lib389.idm.domain import Domain
+from test389.topologies import topology_st as topo
+
+log = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.tier2
+
+
+def _generate_access_log_load(suffix, iterations=100):
+    for i in range(iterations):
+        suffix.replace('description', f'compress_chain_uaf_{i}')
+        suffix.get_attr_val('description')
+
+
+def _assert_no_empty_compressed_logs(log_dir, log_type='access'):
+    for gz_path in glob.glob(f'{log_dir}/{log_type}.*.gz'):
+        assert os.path.getsize(gz_path) > 0, f'Empty compressed log: {gz_path}'
+
+
+def _assert_accesslog_list_matches_disk(inst, log_dir):
+    accesslog_list = inst.config.get_attr_vals_utf8('nsslapd-accesslog-list')
+    if len(accesslog_list) == 1 and accesslog_list[0] == '':
+        accesslog_list = []
+    disk_files = glob.glob(f'{log_dir}/access.2*')
+
+    disk_files_for_compare = set()
+    for fpath in disk_files:
+        if fpath.endswith('.gz'):
+            disk_files_for_compare.add(fpath[:-3])
+        else:
+            disk_files_for_compare.add(fpath)
+
+    missing_from_disk = set(accesslog_list) - disk_files_for_compare
+    assert not missing_from_disk, (
+        'nsslapd-accesslog-list references files missing from disk: '
+        f'{missing_from_disk}'
+    )
+
+
+def test_async_compress_survives_chain_node_reuse(topo):
+    """Regression: compress jobs must not dereference freed rotation-chain nodes.
+
+    Background compress jobs used to keep a raw LogFileInfo pointer into the
+    rotation chain. Retention trims, logdir changes, and disk-monitor purges
+    can free those nodes while jobs are still queued, causing use-after-free
+    when the worker updates l_compressed.
+
+    :id: d8b693be-860e-47d8-a9cf-bf0ba6f8dabf
+    :setup: Standalone Instance
+    :steps:
+        1. Enable access log compression with aggressive rotation limits
+        2. Rapidly rotate logs to queue compress jobs while retention trims
+        3. Change the access log path to discard the in-memory chain
+        4. Rotate again on the new path
+        5. Verify LDAP still works and accesslog-list matches disk
+    :expectedresults:
+        1. Success
+        2. Success
+        3. Success
+        4. Success
+        5. Server remains responsive; list matches files on disk
+    """
+    inst = topo.standalone
+    suffix = Domain(inst, DEFAULT_SUFFIX)
+    log_dir = inst.get_log_dir()
+
+    inst.config.set('nsslapd-accesslog-logbuffering', 'off')
+    inst.config.set('nsslapd-accesslog-compress', 'on')
+    inst.config.set('nsslapd-accesslog-maxlogsize', '1')
+    inst.config.set('nsslapd-accesslog-maxlogsperdir', '2')
+    inst.config.set('nsslapd-accesslog-logexpirationtime', '-1')
+    inst.config.set('nsslapd-accesslog-logrotationsync-enabled', 'off')
+
+    for cycle in range(12):
+        log.info('Rapid rotation cycle %d/12', cycle + 1)
+        _generate_access_log_load(suffix, iterations=120)
+
+    original_accesslog = inst.config.get_attr_val_utf8('nsslapd-accesslog')
+    alt_accesslog = os.path.join(log_dir, 'access_chain_uaf')
+    inst.config.set('nsslapd-accesslog', alt_accesslog)
+
+    for cycle in range(6):
+        log.info('Post-logdir-change rotation cycle %d/6', cycle + 1)
+        _generate_access_log_load(suffix, iterations=120)
+
+    time.sleep(2)
+    suffix.get_attr_val('description')
+    _assert_no_empty_compressed_logs(log_dir)
+    _assert_accesslog_list_matches_disk(inst, log_dir)
+
+    inst.config.set('nsslapd-accesslog', original_accesslog)
--- a/dirsrvtests/tests/suites/logging/logging_compression_test.py
+++ b/dirsrvtests/tests/suites/logging/logging_compression_test.py
@@ -1,5 +1,5 @@
 # --- BEGIN COPYRIGHT BLOCK ---
-# Copyright (C) 2025 Red Hat, Inc.
+# Copyright (C) 2026 Red Hat, Inc.
 # All rights reserved.
 #
 # License: GPL (version 3 or any later version).
@@ -39,7 +39,17 @@ def log_rotated_count(log_type, log_dir, check_compressed=False):
     return len(log_files)
 
 
-def update_and_sleep(inst, suffix, sleep=True):
+def wait_for_compressed_logs(log_type, log_dir, min_count=1, timeout=120):
+    """Wait for background log compression to finish."""
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        if log_rotated_count(log_type, log_dir, check_compressed=True) >= min_count:
+            return
+        time.sleep(1)
+    assert log_rotated_count(log_type, log_dir, check_compressed=True) >= min_count
+
+
+def update_and_sleep(inst, suffix, sleep=61):
     for loop in range(2):
         for count in range(10):
             suffix.replace('description', str(count))
@@ -48,13 +58,10 @@ def update_and_sleep(inst, suffix, sleep=True):
                 suffix.add('doesNotExist', 'error')
             # For security log we need binds to populate the log
             DirectoryManager(inst).bind()
+            time.sleep(0.5)
 
-        if sleep:
-            # log rotation smallest unit is 1 minute
-            time.sleep(61)
-        else:
-            # should still sleep for a little bit
-            time.sleep(1)
+        # log rotation smallest unit is 1 minute
+        time.sleep(sleep/(loop+1))
 
 
 def test_logging_compression(topo):
@@ -102,6 +109,7 @@ def test_logging_compression(topo):
         inst.config.set('nsslapd-' + ds_log + '-logrotationtimeunit', timeunit)
         inst.config.set('nsslapd-' + ds_log + '-maxlogsize', '1')
         inst.config.set('nsslapd-' + ds_log + '-maxlogsperdir', '3')
+        inst.config.set('nsslapd-' + ds_log + '-compress', 'on')
 
     # Perform ops that will write to each log
     update_and_sleep(topo.standalone, suffix)
@@ -110,21 +118,28 @@ def test_logging_compression(topo):
     for log_type in ['access', 'audit', 'auditfail', 'errors', 'security']:
         assert log_rotated_count(log_type, log_dir) > 0
 
-    # Enable log compression on all logs
-    for ds_log in ['accesslog', 'auditlog', 'auditfaillog', 'errorlog', 'securitylog']:
-        inst.config.set('nsslapd-' + ds_log + '-compress', 'on')
-
     # Perform ops that will write to each log
-    update_and_sleep(topo.standalone, suffix)
+    update_and_sleep(topo.standalone, suffix, sleep=30)
 
-    # Make sure all logs were rotated again and are compressed
+    # Make sure all logs were rotated again and are compressed (async maintenance)
     for log_type in ['access', 'audit', 'auditfail', 'errors', 'security']:
-        assert log_rotated_count(log_type, log_dir, check_compressed=True) > 0
+        wait_for_compressed_logs(log_type, log_dir, min_count=2)
+        assert log_rotated_count(log_type, log_dir, check_compressed=True) == 2
+
+    # Take note of the current compressed log names
+    log_file = f'{log_dir}/access.2*.gz'
+    log_files_before = glob.glob(log_file)
 
     # Make sure log deletion is working
-    update_and_sleep(topo.standalone, suffix, sleep=False)
+    update_and_sleep(topo.standalone, suffix)
     for log_type in ['access', 'audit', 'auditfail', 'errors', 'security']:
         assert log_rotated_count(log_type, log_dir) == 2
+
+    # Grab a fresh list of rotated files
+    log_files_after = glob.glob(log_file)
+
+    # The logs should be different if deletion worked
+    assert log_files_before != log_files_after, "Log files should be different after deletion"
 
 
 if __name__ == '__main__':
--- a/ldap/servers/slapd/log.c
+++ b/ldap/servers/slapd/log.c
@@ -1,7 +1,7 @@
 /** BEGIN COPYRIGHT BLOCK
  * Copyright (C) 2001 Sun Microsystems, Inc. Used by permission.
- * Copyright (C) 2025 Red Hat, Inc.
  * Copyright (C) 2010 Hewlett-Packard Development Company, L.P.
+ * Copyright (C) 2026 Red Hat, Inc.
  * All rights reserved.
  *
  * License: GPL (version 3 or any later version).
@@ -106,17 +106,24 @@ static int log__open_errorlogfile(int logfile_type, int locked);
 static int log__open_auditlogfile(int logfile_type, int locked);
 static int log__open_auditfaillogfile(int logfile_type, int locked);
 static int log__needrotation(LOGFD fp, int logtype);
-static int log__delete_access_logfile(void);
-static int log__delete_security_logfile(void);
-static int log__delete_error_logfile(int locked);
-static int log__delete_audit_logfile(void);
-static int log__delete_auditfail_logfile(void);
+static int log__delete_access_logfile(bool at_rotation);
+static int log__delete_security_logfile(bool at_rotation);
+static int log__delete_error_logfile(int locked, bool at_rotation);
+static int log__delete_audit_logfile(bool at_rotation);
+static int log__delete_auditfail_logfile(bool at_rotation);
+static int log__write_rotationinfo(int logtype);
+static void log__update_logentry_compressed_size(const char *log_file, PRInt64 maxlogsize,
+                                                 LogFileInfo *logp);
+static LogFileInfo *log__find_chain_entry(LogFileInfo *chain, time_t ctime);
+static void log__purge_rotated_logchain(LogFileInfo **logchain, unsigned int *numof_logs,
+                                        const char *log_file, int logtype);
+static void log_maint_after_compress(int stream_idx, time_t log_ctime);
 static int log__access_rotationinfof(char *pathname);
 static int log__security_rotationinfof(char *pathname);
 static int log__error_rotationinfof(char *pathname);
 static int log__audit_rotationinfof(char *pathname);
 static int log__auditfail_rotationinfof(char *pathname);
-static int log__extract_logheader(FILE *fp, long *f_ctime, PRInt64 *f_size, PRBool *compressed);
+static int log__extract_logheader(FILE *fp, long *f_ctime, PRInt64 *f_size, bool *compressed);
 static int log__check_prevlogs(FILE *fp, char *filename);
 static PRInt64 log__getfilesize(LOGFD fp);
 static PRInt64 log__getfilesize_with_filename(char *filename);
@@ -141,7 +148,7 @@ static void vslapd_log_emergency_error(LOGFD fp, const char *msg, int locked);
 static int get_syslog_loglevel(int loglevel);
 static void log_external_libs_debug_openldap_print(char *buffer);
 static int log__fix_rotationinfof(char *pathname);
-static int log__validate_rotated_logname(const char *timestamp_str, PRBool *is_compressed);
+static int log__validate_rotated_logname(const char *timestamp_str, bool *is_compressed);
 
 static int
 get_syslog_loglevel(int loglevel)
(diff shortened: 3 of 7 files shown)


ldap/servers/slapd/log.c
```diff
@@ -235,6 +240,460 @@ compress_log_file(char *log_name, int32_t mode)
...
+logs_maintenance_shutdown(void)
+{
+    if (!log_maint_initialized) {
+        return;
+    }
+
+    for (size_t i = 0; i < LOG_MAINT_NUM_STREAMS; i++) {
+        log_maint_submit_pending(log_maint_streams[i].logtype);
+    }
+
```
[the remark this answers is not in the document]
mreynolds389: We need all compression jobs to finish before shutdown, because the log was already rotated and we do not check for uncompressed logs at server startup.

ldap/servers/slapd/log.c
```diff
@@ -235,6 +240,469 @@ compress_log_file(char *log_name, int32_t mode)
...
+    stream = &log_maint_streams[idx];
+    if (log_maint_path_pending(stream, path)) {
+        return;
+    }
+
+    job = (LogMaintPendingJob *)slapi_ch_calloc(1, sizeof(LogMaintPendingJob));
+    job->job_type = LOG_MAINT_JOB_COMPRESS;
+    job->path = slapi_ch_strdup(path);
+    job->mode = mode;
+    job->log_entry = entry;
```
droideck, line 574: IIUC, the job keeps a raw pointer into the rotation chain, but the chain doesn't know about it.
Unless I missed something, three things free chain nodes while a job waits: the next rotation's trim, log_update_accesslogdir(), and the disk-monitor purge. The worker then writes entry->l_compressed (log.c:392) into freed memory.

Fix idea: let the job own copies of what it needs (path, mode) and update the chain node under the lock only if it still exists, or refcount the nodes. To check: ASAN build, maxlogsperdir: 2, tiny maxlogsize, steady write load so a rotation lands while a gzip is still running. ASAN should flag the write, IMO

ldap/servers/slapd/log.c
```diff
@@ -235,6 +240,469 @@ compress_log_file(char *log_name, int32_t mode)
...
+    if (idx < 0 || !log_maint_initialized) {
+        return;
+    }
+
+    stream = &log_maint_streams[idx];
+    if (stream->pending_head == NULL) {
+        return;
+    }
+
+    pending = stream->pending_head;
```
droideck, line 605: Staging onto this list happens under the stream lock, but draining happens after PR_Unlock (the new macros), and this function takes no lock. IIUC, two threads unlocking back to back can both walk the list and free the same nodes. This is true for all five streams.
I think we can put the pending list under a lock, or drain before unlocking. TSAN build with several client threads and 1 minute rotation should flag the race on pending_head.

ldap/servers/slapd/log.c
```diff
@@ -488,6 +956,8 @@ g_log_init()
         exit(-1);
     }
     CFG_UNLOCK_READ(cfg);
+
+    logs_maintenance_init();
```
droideck, line 960: This runs before detach() forks, and threads don't survive a fork. main.c has a comment warning exactly about this.
https://github.com/389ds/389-ds-base/blob/3e4308488fc4223b455089124a1db38ded0c993b/ldap/servers/slapd/main.c#L803-L809
Under systemd nothing forks, so it looks fine... So it's a minor correctness issue, but I think still worth fixing. A plain ns-slapd start loses the worker: compression never happens and shutdown joins a thread that only existed in the parent. So on with a start without systemd and without -d, and trying to rotate with compression on. No .gz will appear.
So maybe we can start the worker after detach, where the other server threads start.

ldap/servers/slapd/log.c
```diff
@@ -235,6 +240,469 @@ compress_log_file(char *log_name, int32_t mode)
...
+    int idx = log_maint_stream_index(logtype);
+    LogMaintPendingJob *job = NULL;
+    LogMaintenanceStream *stream = NULL;
+
+    if (idx < 0 || path == NULL || !log_maint_initialized) {
+        return;
+    }
+
+    stream = &log_maint_streams[idx];
+    if (log_maint_path_pending(stream, path)) {
```
droideck, line 566: I think we may have a tricky race here...
This line only checks the per-stream pending list, not the global queue. Two rotations in the same second produce the same archive name, so the second job opens the finished .gz with O_TRUNC (log.c:205) and then fails on the missing source. The good archive becomes a zero-byte file. 
So, IMO, we need also check the global queue, or skip compression when the target .gz already exists.

droideck (changes requested):

droideck:
Regarding the `Non-constant format string` CodeQL alerts on log.c's `slapi_log_err` are pre-existing, not from this change. There are nine calls in this file that omit the subsystem argument, so the intended format string sits in the subsystem slot and a runtime value (pathname or loginfo.log_error_file) becomes the format string. They predate this PR and only surfaced because it touches the file...

ldap/servers/slapd/log.c
mreynolds389: Sounds good

ldap/servers/slapd/log.c
mreynolds389: Yeah there is definitely an issue there.  New commit should fix all of this.

ldap/servers/slapd/log.c
mreynolds389: Yeah we needed a pending list lock.  This is now fixed.

mreynolds389:
@droideck ok made more changes and ran a lot of tests (including ASAN).  Sorry the patch is getting larger, but it's ready for review.

ldap/servers/slapd/log.c
```diff
@@ -4756,117 +5568,110 @@ log__delete_access_logfile(void)
...
-                          "Unable to remove file: %s - error %d (%s)\n",
-                          buffer, prerr, slapd_pr_strerror(prerr));
-            logp = logp->l_next;
-            continue;
+            if (PR_FILE_NOT_FOUND_ERROR == prerr && !logp->l_compressed) {
+                PR_snprintf(buffer, sizeof(buffer), "%s.%s.gz", log_file, tbuf);
+                if (PR_Delete(buffer) != PR_SUCCESS) {
+                    prerr = PR_GetError();
+                    if (PR_FILE_NOT_FOUND_ERROR != prerr) {
+                        slapi_log_err(SLAPI_LOG_ERR, "log__delete_rotated_logs",
```
droideck: We're holding the error log write lock here, and slapi_log_err() grabs that same lock again (the nested wrlock just returns EDEADLK, which we ignore). IIUC, the nested call ends up releasing it, so the rest of the purge runs unlocked and the final unlock corrupts the rwlock. Only happens when PR_Delete fails with something other than "not found", but when it does the error log lock is toast:D
Possibly we can duplicate how log__delete_error_logfile() handles a similar issue

ldap/servers/slapd/log.c
```diff
@@ -4756,117 +5568,110 @@ log__delete_access_logfile(void)
...
+                                      "Unable to remove file: %s - error %d (%s)\n",
+                                      buffer, prerr, slapd_pr_strerror(prerr));
+                    }
+                }
+            } else if (PR_FILE_NOT_FOUND_ERROR != prerr) {
+                slapi_log_err(SLAPI_LOG_ERR, "log__delete_rotated_logs",
```
droideck: Same here

ldap/servers/slapd/log.c
droideck, line 5369: This can now be called from the maintenance worker with `at_rotation == false`, but this `maxnumlogs == 1` block still assumes we're in rotation where the caller reopens the log right after. From the worker nobody reopens, so we close and delete the live log.

And a bit lower, `log__getfilesize()` at 5460 gets log_access_fdes with no NULL check, which crashes if the open failed during rotation. Possibly one guard at the top covers both: `if (!at_rotation && (loginfo.log_access_maxnumlogs == 1 || loginfo.log_access_fdes == NULL)) return 0;` The other four log__delete*logfile() functions would need the same, I think

ldap/servers/slapd/log.c
```diff
@@ -202,37 +220,698 @@ compress_log_file(char *log_name, int32_t mode)
...
+        PR_Lock(log_maint_global.lock);
+        while (log_maint_global.queue_head == NULL && !log_maint_global.shutdown) {
+            PR_WaitCondVar(log_maint_global.cvar, PR_INTERVAL_NO_TIMEOUT);
+        }
+        if (log_maint_global.shutdown && log_maint_global.queue_head == NULL) {
+            PR_Unlock(log_maint_global.lock);
+            break;
+        } else if (log_maint_global.shutdown && !shutting_down) {
+            shutting_down = true;
+            slapi_log_err(SLAPI_LOG_NOTICE, "log_maint_worker",
```
droideck: We log here while still holding log_maint_global.lock, and slapi_log_err() can trigger an error log rotation, which calls back into log_maint_compress_path_scheduled() and tries to take the same lock at 380. Needs errorlog compression on, but then the worker deadlocks on its own lock and shutdown hangs. Possibly just move the message after the PR_Unlock at 610.

ldap/servers/slapd/log.c
```diff
@@ -202,37 +220,698 @@ compress_log_file(char *log_name, int32_t mode)
...
-        }
-        bytes_read = fread(buf, 1, LOG_CHUNK, source);
+    log_maint_free_global_queue();
+    for (size_t i = 0; i < LOG_MAINT_NUM_STREAMS; i++) {
+        log_maint_free_stream_pending(&log_maint_streams[i]);
     }
-    gzclose(outfile);
-    fclose(source);
-    PR_Delete(log_name); /* remove the old uncompressed log */
+    log_maint_destroy_sync();
```
droideck: Small ordering thing: we destroy the locks here but log_maint_initialized only flips to false at 914.
IIUC, a thread logging in that window passes the flag check in log_maint_submit_pending() and takes a destroyed lock. Flipping the flag before the frees closes it, I think...

ldap/servers/slapd/log.c
```diff
@@ -202,37 +220,698 @@ compress_log_file(char *log_name, int32_t mode)
...
+/*
+ * Async log maintenance: background gzip and retention for all log streams.
+ * A single worker thread (log_maint) drains one global FIFO queue. During
+ * rotation, jobs are staged on a per-stream pending list under the write lock
+ * and handed to the worker when LOG_*_UNLOCK_WRITE() runs.
+ */
+
+typedef enum {
+    LOG_MAINT_JOB_COMPRESS = 1,        /* gzip a rotated log file */
+    LOG_MAINT_JOB_RETENTION_SWEEP = 2  /* run expiration/disk retention only */
```
droideck: Non-blocking: nothing ever queues LOG_MAINT_JOB_RETENTION_SWEEP, so log_maint_run_retention_policies() is dead code. Was the sweep meant to be scheduled somewhere?

ldap/servers/slapd/log.c
mreynolds389: > Small ordering thing: we destroy the locks here but log_maint_initialized only flips to false at 914. IIUC, a thread logging in that window passes the flag check in log_maint_submit_pending() and takes a destroyed lock. Flipping the flag before the frees closes it, I think...

Agreed, I'll switch this.

ldap/servers/slapd/log.c
mreynolds389: > We log here while still holding log_maint_global.lock, and slapi_log_err() can trigger an error log rotation, which calls back into log_maint_compress_path_scheduled() and tries to take the same lock at 380. Needs errorlog compression on, but then the worker deadlocks on its own lock and shutdown hangs. Possibly just move the message after the PR_Unlock at 610.

We don't need to log anything here so I'll remove those logging lines.

ldap/servers/slapd/log.c
mreynolds389: > We're holding the error log write lock here, and slapi_log_err() grabs that same lock again (the nested wrlock just returns EDEADLK, which we ignore). IIUC, the nested call ends up releasing it, so the rest of the purge runs unlocked and the final unlock corrupts the rwlock. Only happens when PR_Delete fails with something other than "not found", but when it does the error log lock is toast:D Possibly we can duplicate how log__delete_error_logfile() handles a similar issue

Ok we only call this function with the write lock taken, so I added an unlock -> write error -> lock

ldap/servers/slapd/log.c
mreynolds389: I fixed both

ldap/servers/slapd/log.c
```diff
@@ -4756,117 +5516,161 @@ log__delete_access_logfile(void)
...
+                              "Unable to remove file: %s - error %d (%s)\n",
+                              buffer, prerr, slapd_pr_strerror(prerr));
+                log_write_lock(logtype);
+            }
         }
-        prev_log = logp;
-        loginfo.log_numof_auditfail_logs--;
-        logp = logp->l_next;
-        slapi_ch_free((void **)&prev_log);
+        slapi_ch_free((void **)&logp);
```
droideck: I'm not sure it's the best option... The loop frees each node without unlinking it, and *logchain = NULL only happens after the loop. So IIUC, while we're unlocked here, the chain head points at freed nodes, and anyone grabbing the lock in that window (a rotation, an accesslog-list read, the worker's chain lookup) walks freed memory... Or do I miss something?
Why not use the established pattern from `log__delete_error_logfile` as I suggested before? It reports failures via log__error_emergency(buffer, 0, locked) (syslog only, doesn't touch the lock), so we'd never need to unlock at all. And for the non-error streams the unlock isn't needed anyway since slapi_log_err only grabs the error lock.
But may be I miss something...

ldap/servers/slapd/log.c
```diff
@@ -4582,23 +5340,23 @@ log__needrotation(LOGFD fp, int logtype)
...
+    PRInt64 f_size = 0;
     int numoflogs = loginfo.log_numof_access_logs;
     int rv = 0;
     char *logstr;
     char buffer[BUFSIZ];
     char tbuf[TBUFSIZE];
 
     /* If we have only one log, then  will delete this one */
-    if (loginfo.log_access_maxnumlogs == 1) {
+    if (!at_rotation && loginfo.log_access_maxnumlogs == 1) {
```
droideck: I think the condition got inverted here. My suggestion was to bail out early in retention mode (`if (!at_rotation && ...) return 0;`), but this gates the close-and-delete branch with `!at_rotation`, so it now runs only from the maintenance worker, the exact case where nobody reopens the log.
So access logging silently stops after this (slapi_log_access checks fdes for NULL), and a second sweep hits PR_Close(NULL). At rotation the branch can't be reached anyway since log__needrotation bails when nlogs == 1, so I'd just make retention return 0 here as mentioned in the previous comment. Or am I missing something about why you went this way?..

PS: same in the other four delete functions.

dirsrvtests/tests/suites/logging/async_log_compress_chain_uaf_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_async_compress_survives_chain_node_reuse(topo):
+    """Regression: compress jobs must not dereference freed rotation-chain nodes.
+
+    Background compress jobs used to keep a raw LogFileInfo pointer into the
+    rotation chain. Retention trims, logdir changes, and disk-monitor purges
+    can free those nodes while jobs are still queued, causing use-after-free
+    when the worker updates l_compressed.
+
+    :id: async-log-compress-chain-uaf-001
```
droideck: Probably AI misunderstanding the UUID stuff:)

dirsrvtests/tests/suites/logging/logging_compression_test.py
```diff
@@ -110,22 +118,32 @@ def test_logging_compression(topo):
...
-
     # Perform ops that will write to each log
-    update_and_sleep(topo.standalone, suffix)
+    update_and_sleep(topo.standalone, suffix, sleep=30)
 
-    # Make sure all logs were rotated again and are compressed
+    # Make sure all logs were rotated again and are compressed (async maintenance)
     for log_type in ['access', 'audit', 'auditfail', 'errors', 'security']:
-        assert log_rotated_count(log_type, log_dir, check_compressed=True) > 0
+        wait_for_compressed_logs(log_type, log_dir)
```
droideck: Possibly, we should wait for min_count=2 as we assert `== 2` below.

ldap/servers/slapd/log.c
droideck: This was fine before since compress_log_file() had already run by this point, but now the compress is deferred, so the .gz name written here doesn't exist yet on disk, IIUC. It self-corrects when the worker rewrites rotationinfo, but access/security avoid the mismatch entirely by gating on logp->l_compressed... Can we do the same here (and in audit/auditfail)?

droideck (changes requested):

ldap/servers/slapd/log.c
mreynolds389: > I'm not sure it's the best option... The loop frees each node without unlinking it, and *logchain = NULL only happens after the loop. So IIUC, while we're unlocked here, the chain head points at freed nodes, and anyone grabbing the lock in that window (a rotation, an accesslog-list read, the worker's chain lookup) walks freed memory... Or do I miss something? 
Why not use the established pattern from `log__delete_error_logfile` as I suggested before? It reports failures via log__error_emergency(buffer, 0, locked) (syslog only, doesn't touch the lock), so we'd never need to unlock at all. And for the non-error streams the unlock isn't needed anyway since slapi_log_err only grabs the error lock. But may be I miss something...

@droideck I think perhaps I need a different approach here and completely scrap this PR and start over.  It keeps getting tweaked via your suggestions, and you keep finding issues that were there from the very beginning that you are only now mentioning, and I'm losing track of what's going on.  WDYT?

mreynolds389: > Why not use the established pattern from `log__delete_error_logfile` as I suggested before? It reports failures via log__error_emergency(buffer, 0, locked) (syslog only, doesn't touch the lock), so we'd never need to unlock at all. And for the non-error streams the unlock isn't needed anyway since slapi_log_err only grabs the error lock. But may be I miss something...

Actually I did look at that function and it checks if it's locked or not before logging anything when it fails to delete a file.  That's why I added those lock functions as I didn't want to pass yet another parameter to the function and have to refactor a bunch of code.  Then after you said this ^^^ I checked again and at the very bottom of the function it uses log__error_emergency().  So I originally thought you wanted me to pass a new lock parameter which why I did what I did.

ldap/servers/slapd/log.c
mreynolds389: I thought I followed your exact recommendation, but I was a bit hasty when I looked it over.

ldap/servers/slapd/log.c
mreynolds389: > I'm not sure it's the best option... The loop frees each node without unlinking it, and *logchain = NULL only happens after the loop. So IIUC, while we're unlocked here, the chain head points at freed nodes, and anyone grabbing the lock in that window (a rotation, an accesslog-list read, the worker's chain lookup) walks freed memory... Or do I miss something?

I think this was correctly unlinking the logchain, the "for" loop was unconventional so I changed it to a easier to follow while() loop.  Also we do hold the write lock during this entire process.  I merged these changes, but I'm running through ASAN right now...

dirsrvtests/tests/suites/logging/logging_compression_test.py
droideck: I think we still need the `min_count=2` parameter to wait for two logs instead of default one? The next line after that asserts two logs.

droideck (approved):
> @droideck I think perhaps I need a different approach here and completely scrap this PR and start over. It keeps getting tweaked via your suggestions, and you keep finding issues that were there from the very beginning that you are only now mentioning, and I'm losing track of what's going on. WDYT?

Please don't scrap it! The architecture is right, and your latest push fixed the big ones cleanly.

About "issues that were there from the beginning" - fair, and partly it's how I review... I also use AI to sweep a patch this big, but then I hand-verify every candidate finding and throw away most of them, so the real ones surface in batches as I get through the layers. It takes hours per round honestly. But every comment I left has a concrete failure path behind it.

One thing I noticed in this PR that I think worth mentioning, because I keep hitting the same in my own work too. All the mechanical parts here are right: the queue, the worker, the staging lists, the tests... But every bug we found was an invariant violation. Lock dropped while the structure is inconsistent, node freed but still linked, branch that assumes its caller reopens the file.
IMO, that split is not a coincidence, it's just how current AI tooling works: really good at mechanics, confidently wrong about invariants. And both come out looking equally clean... Which kind of breaks the old heuristic we all relied on for years - hand-written code that reads clean usually WAS careful, because the same attention produces both. Generated code reads clean no matter what:) So "looks right and tests pass" verifies much less than our instincts tell us. And sync-to-async conversion is probably the worst case input for these models.

And there's the other side of the same coin that worries me more... The same tooling is now pointed at projects like ours from the outside. People literally run AI over public diffs of security-sensitive daemons hunting for exactly these windows - a UAF behind a dropped lock that would sit unnoticed for years in the pre-AI era now can get found in days, and may be not by the good guys first. We're already seeing the CVE flood across the ecosystem from this. So IMO the pre-push bar for memory/concurrency code has to be higher now than it ever was, for all of us. Not about this PR specifically, it's just the water we swim in now... Oh well.

What changed for me in practice because of all that: the effort I used to put into writing this kind of code, now goes into writing down the invariants first (who holds which lock when, who owns each chain node, what every exit path must restore), and then running deep AI review rounds on my own patch before it goes public - with the model told to attack those invariants, like "find any path where the write lock is dropped while the chain is inconsistent". It catches most of this class pre-push, where fixes are cheap and nobody has to track the review waves. Your ASAN/TSAN runs under rotation load are the other half of exactly that.

Anyway, still open from my side: `min_count=2` in the test (one-liner)
And once ASAN comes back clean, I'm good!

droideck:
One caveat on that btw - I don't think AGENTS.md style guidance can fully save us here, even a good one. Those files can only describe the invariants that already exist in the tree. But a patch like this one literally rewrites them: LOG*UNLOCK_WRITE now means "unlock and submit jobs", the delete functions got two modes with different rules, chain nodes got a new concurrent reader. No pre-written doc could contain the invariants of code that didn't exist yet. So this is exactly the place where the developer has to be opinionated and bring the domain understanding himself - define what must stay true in YOUR new design, write it down, and only then let the model check the code against it. If we let the model both invent the invariants and verify them, it just grades its own homework :)

To be fair, sometimes we get lucky and the model picks the right direction on its own, especially the top-tier ones - they do surprisingly well. But "usually right" is exactly the dangerous property for invariant stuff... When we leave it a lot of freedom, the failures don't disappear, they just go hidden: the code works, tests pass, and the one interleaving nobody stated blows up half a year later at a customer under real load. That kind of bite costs way more than any review round would.

And yeah, throwing the most expensive model at everything is an option too, but IMO a wrong one, for two reasons - it's expensive, and it's outsourcing the muscle. Our domain knowledge is exactly what lets us spend the tokens where they pay (mechanics, sweeps, adversarial checks) and keep the invariant thinking to ourselves. Which conveniently is also how the brain stays in shape :) Use it or lose it applies here too, heh.

Honestly, that's also why I'm not worried about good engineers going extinct. The typing got cheap, but somebody still has to know what must stay true.

dirsrvtests/tests/suites/logging/logging_compression_test.py
mreynolds389: Sorry should be fixed now

mreynolds389:
> One caveat on that btw - I don't think AGENTS.md style guidance can fully save us here, even a good one. 

I agreed, but it should improve it somewhat.

> And yeah, throwing the most expensive model at everything is an option too, but IMO a wrong one, for two reasons - it's expensive, and it's outsourcing the muscle. Our domain knowledge is exactly what lets us spend the tokens where they pay (mechanics, sweeps, adversarial checks) and keep the invariant thinking to ourselves. Which conveniently is also how the brain stays in shape :) Use it or lose it applies here too, heh.

Absolutely :-)

mreynolds389:
> > @droideck I think perhaps I need a different approach here and completely scrap this PR and start over. It keeps getting tweaked via your suggestions, and you keep finding issues that were there from the very beginning that you are only now mentioning, and I'm losing track of what's going on. WDYT?
> 
> Please don't scrap it! The architecture is right, and your latest push fixed the big ones cleanly.
> 
> About "issues that were there from the beginning" - fair, and partly it's how I review... I also use AI to sweep a patch this big, but then I hand-verify every candidate finding and throw away most of them, so the real ones surface in batches as I get through the layers. It takes hours per round honestly. But every comment I left has a concrete failure path behind it.
...
> What changed for me in practice because of all that: the effort I used to put into writing this kind of code, now goes into writing down the invariants first (who holds which lock when, who owns each chain node, what every exit path must restore), and then running deep AI review rounds on my own patch before it goes public - with the model told to attack those invariants, like "find any path where the write lock is dropped while the chain is inconsistent". It catches most of this class pre-push, where fixes are cheap and nobody has to track the review waves. Your ASAN/TSAN runs under rotation load are the other half of exactly that.
> 
> Anyway, still open from my side: `min_count=2` in the test (one-liner) And once ASAN comes back clean, I'm good!

Well said!

PR #7646, opened 2026-07-13
Issue 7645 - Add runtime LeakSanitizer leak check

### LINKED ISSUE #7645: Add runtime LeakSanitizer leak check
**Issue Description**
If the server is built with ASAN or `LD_PRELOAD`'ed with `libasan`, a recoverable leak check can be triggered at runtime using `__lsan_do_recoverable_leak_check` via SIGUSR1 which is currently unused https://github.com/389ds/389-ds-base/blob/4ce7a49576f84c1d8bcbf8635fdb6d18b31db771/ldap/servers/slapd/daemon.c#L2460

vashirov:
Description:
If the server is built with ASAN or LD_PRELOADed with libasan, a recoverable leak check can be triggered at runtime by sending SIGUSR1 to ns-slapd.
Before calling LeakSanitizer, a fork+ptrace probe verifies that ptrace is permitted. Without this check, LeakSanitizer calls exit(1) when ptrace is denied by SELinux or suid_dumpable restrictions.

Fixes: https://github.com/389ds/389-ds-base/issues/7645

## Summary by Sourcery

Add a runtime LeakSanitizer leak-check mechanism triggered by SIGUSR1, with safety checks to avoid server termination when ptrace is restricted.

New Features:
- Allow triggering a recoverable LeakSanitizer leak check in ns-slapd at runtime via SIGUSR1 when AddressSanitizer is available.

Bug Fixes:
- Prevent LeakSanitizer from terminating the server by probing ptrace availability before running the leak check and skipping it when ptrace is denied.

Enhancements:
- Integrate leak-check logging and signal handling so SIGUSR1 is repurposed from a no-op to a controlled diagnostic action on all platforms.

### DIFF
--- a/ldap/servers/slapd/daemon.c
+++ b/ldap/servers/slapd/daemon.c
@@ -32,6 +32,9 @@
 #include <time.h>
 #include <signal.h>
 #include <fcntl.h>
+#include <dlfcn.h>
+#include <sys/ptrace.h>
+#include <sys/wait.h>
 #define TCPLEN_T int
 #ifdef NEED_FILIO
 #include <sys/filio.h>
@@ -141,6 +144,8 @@ static PRFileDesc **createprlistensockets(unsigned short port,
                                           int local);
 static const char *netaddr2string(const PRNetAddr *addr, char *addrbuf, size_t addrbuflen);
 static void set_shutdown(int);
+static void set_lsan_check(int);
+static volatile sig_atomic_t lsan_check_requested = 0;
 static void setup_pr_ct_firsttime_pds(Connection_Table *ct);
 #ifdef ENABLE_EPOLL
 static PRIntn setup_pr_accept_pds(PRFileDesc **n_tcps, PRFileDesc **s_tcps, PRFileDesc **i_unix, int epoll_fd);
@@ -1385,7 +1390,9 @@ slapd_daemon(daemon_ports_t *ports)
 #endif /* !ENABLE_EPOLL */
     /* The meat of the operation is in a loop on a call to select */
     while (!g_get_shutdown()) {
-
+        if (lsan_check_requested) {
+            slapd_lsan_check();
+        }
         usleep(500 * 1000);
     }
     /* We get here when the server is shutting down */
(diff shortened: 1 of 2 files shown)


vashirov:
To trigger, run `kill -SIGUSR1 $(pidof ns-slapd)`:
```
Jul 13 14:14:02 fedora ns-slapd[4486]: [13/Jul/2026:14:14:02.787881509 +0000] - INFO - slapd_lsan_check - Leak check completed - 1 leak(s) detected
```
*.asan file will be generated at `/var/run/dirsrv/`.

If ptrace is not available (SELinux in Enforcing mode or suid_dumpable!=1), it prints an error without killing the server:
```
Jul 13 14:13:15 fedora ns-slapd[4375]: [13/Jul/2026:14:13:15.435819366 +0000] - ERR - slapd_lsan_check - Cannot perform leak check - ptrace is not permitted. Check: SELinux 'allow dirsrv_t self:process ptrace' (ausearch -m avc -c ns-slapd), and /proc/sys/fs/suid_dumpable (must be 1)
```

progier389 (approved):
LGTM

ldap/servers/slapd/daemon.c
```diff
@@ -2593,26 +2595,88 @@ set_shutdown(int sig __attribute__((unused)))
...
+    child = fork();
+    if (child == -1) {
+        return 0;
+    }
+
+    if (child == 0) {
+        /* Child: try to ptrace the parent */
+        if (ptrace(PTRACE_ATTACH, getppid(), NULL, NULL) == 0) {
+            /* Attached - detach and exit success */
+            waitpid(getppid(), NULL, 0);
```
droideck, line 2636: I think it's possible for a server to hang on a shutdown in the ASAN build (even though rarely).
IIUC, the child inherits the SIGCHLD handler. After PTRACE_ATTACH the parent's thread is the child's tracee, so if the inherited handler runs first, it eats the stop event. The child then waits here forever, and the parent's main thread stays stuck at `ret = waitpid(child, &status, 0);` a bit below on :2636. That thread is the one polling shutdown, so systemctl stop hangs until the kill timeout. 
Maybe it makes sense to reset SIGCHLD to default in the child before ptrace, and report the probe result over a pipe instead of the exit status. WDYT?

ldap/servers/slapd/daemon.c
```diff
@@ -2593,26 +2595,88 @@ set_shutdown(int sig __attribute__((unused)))
...
+                      "(ausearch -m avc -c ns-slapd), "
+                      "and /proc/sys/fs/suid_dumpable (must be 1)\n");
+        return;
+    }
+
+    slapi_log_err(SLAPI_LOG_INFO, "slapd_lsan_check",
+                  "SIGUSR1 signal received - performing recoverable leak check\n");
+    num_leaks = lsan_check_fn();
+    slapi_log_err(SLAPI_LOG_INFO, "slapd_lsan_check",
+                  "Leak check completed - %d leak(s) detected\n", num_leaks);
```
droideck: IIUC, `lsan_check_fn` resolves to `__lsan_do_recoverable_leak_check`, and per sanitizer/lsan_interface.h it "returns zero if no leaks have been found or if leak detection is disabled, non-zero otherwise". So the message is incorrect, as it may find 7 leaks but lsan_check_fn will return 1 because if its nature (I checked locally).
I think we should just log "leaks detected, see report" vs "no leaks detected", and mention where the report was written.

droideck (changes requested):

droideck (approved):
LGTM!

Issue #7466, opened 2026-04-30
Local password policies can be created with invalid input


mirielka:
**Issue Description**
When creating a local password policy, a letter can be input into a field that should only accept non-negative integer and no error is displayed.

**Package Version and Platform:**
cockpit-389-ds-3.2.0-4.el10dsrv.noarch

**Steps to Reproduce**
1. Database -> Password Policies -> Local Policies -> Create a new policy
2. Insert a letter into any field that should only accept non-negative integer
3. Confirm by clicking "Create New Policy"

**Expected results**
An error should be displayed under the affected field and it should not be possible to create the policy.

**Actual results**
No error or warning is displayed, "Create New Policy" button becomes clickable, the policy is created. All the values that had the letter input now have value of 0.

**Screenshots**

<img width="1401" height="1295" alt="Image" src="https://github.com/user-attachments/assets/a8d92618-979e-40aa-9c7d-74d1fb125fa4" />

**Additional context**
This only occurs when creating local policy, not in editing.
Does not affect global policies.

mreynolds389:
This seems to be a limitation with TextInput type "number".  We should just switch to a NumberInput.  Will work on this after https://github.com/389ds/389-ds-base/pull/7524 is merged.

mreynolds389:
4e1c7a6f0..44ec05706  389-ds-base-3.2 -> 389-ds-base-3.2

PR #7648, opened 2026-07-13
Issue 7466 - UI - Refactor all TextInput number types to NumberInput

mreynolds389:
Description:

TextInput "number" types allow for invalid values, while NumberInput has built-in functionality to prevent this.

relates: https://github.com/389ds/389-ds-base/issues/7466

Assisted-by: Cursor

### DIFF
--- a/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
+++ b/src/cockpit/389-console/src/lib/database/databaseConfig.jsx
@@ -28,6 +28,7 @@ import {
 import PropTypes from "prop-types";
 import { SyncAltIcon } from '@patternfly/react-icons';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
+import { BanIcon } from '@patternfly/react-icons/dist/js/icons/ban-icon';
 
 const _ = cockpit.gettext;
 
@@ -1838,7 +1839,7 @@ export class GlobalDatabaseConfigMDB extends React.Component {
                                         title={_("Database maximum size in megabytes. The practical maximum size of an LMDB database is limited by the system's addressable memory (nsslapd-mdb-max-size).")}
                                         className="ds-margin-top-xlg"
                                     >
-                                        <GridItem className="ds-label" span={4}>
+                                        <GridItem className="ds-label" span={3}>
                                             {_("Database Maximum Size")}
                                         </GridItem>
                                         <GridItem span={8}>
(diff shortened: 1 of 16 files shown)


src/cockpit/389-console/src/lib/database/databaseConfig.jsx
```diff
@@ -28,6 +28,7 @@ import {
 import PropTypes from "prop-types";
 import { SyncAltIcon } from '@patternfly/react-icons';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
+import { BanIcon } from `@patternfly/react-icons/dist/js/icons/ban-icon`;
```
droideck: I think we should use quotes here instead of backticks. Something may fail...

src/cockpit/389-console/src/lib/dsNumberInput.jsx
```diff
@@ -0,0 +1,151 @@
...
+            max: range.max !== undefined ? range.max : INT32_MAX,
+        };
+    }
+    return { min: 0, max: INT32_MAX };
+}
+
+function isAllowedInput(raw, min) {
+    if (raw === "") {
+        return true;
+    }
```
droideck, line 39: Just to be safe, I'd validate typed values against min/max here too.
PatternFly only bounds the `+/-` buttons. We probably want to suggest rejecting or clamping out-of-range input before onChange.

droideck (changes requested):
Image

DN Cache Size can't be edited.

mreynolds389:
> DN Cache Size can't be edited.

DN cache is autosized - so it should not be editable

mreynolds389:
> > DN Cache Size can't be edited.
> 
> DN cache is autosized - so it should not be editable

Wait it should not be blank, or selectable

mreynolds389:
@droideck ok I think I fixed everything, please review...

src/cockpit/389-console/src/lib/dsNumberInput.jsx
```diff
@@ -0,0 +1,157 @@
...
+    if (raw === "-" && min < 0) {
+        return true;
+    }
+    if (max !== undefined && Number(raw) > max) {
+        return false;
+    }
+    if (min !== undefined && Number(raw) < min) {
+        return false;
+    }
```
droideck: If I select a field with min=2 (for example passwordminlength) and type 12, the first 1 is rejected, so the controlled input resets and I have 82 instead of 12.

droideck (approved):
Besides a few small min/max issues -- LGTM!

- Some new ranges do not match backend validation.
PBKDF2 iterations at globalPwp.jsx (line 1608) default to 0..INT32_MAX, while lib389 accepts only 10,000..10,000,000 password_plugins.py (line 63)
- SASL buffer size at sasl.jsx (line 632) uses min=1; the server silently raises anything below 2,097,152 to that default.

src/cockpit/389-console/src/lib/dsNumberInput.jsx
mreynolds389: Ok I think everything is fixed, please review @droideck

mreynolds389:
@droideck  - I addressed you private comments in this PR, please review.

droideck (approved):
LGTM!

PR #7649, opened 2026-07-15
Issue 7611 - Preserve legacy PBKDF2 hash compatibility

droideck:
Description: Use a fixed 50,000-round verification ceiling so legacy C PBKDF2_SHA256 hashes continue to authenticate across upgrades and restarts.
Cap generated legacy hashes at the configured ceiling, fall back safely on invalid configuration, and keep the optional object-class upgrade non-fatal.
Exercise migration from the legacy underscore scheme to a separately configured modern PBKDF2-SHA256 scheme at 600,000 rounds, without changing the legacy verification ceiling.
Keep the regression coverage self-contained and restore modified server state.

Relates: https://github.com/389ds/389-ds-base/issues/7611

Reviewed by: ?

## Summary by Sourcery

Preserve compatibility with legacy C PBKDF2-SHA256 password hashes while enforcing a sane verification ceiling and supporting migration to a modern PBKDF2-SHA256 scheme.

Bug Fixes:
- Ensure legacy PBKDF2_SHA256 hashes with 50,000 rounds continue to authenticate across restarts and upgrades under a fixed verification ceiling.
- Prevent misconfigured PBKDF2 iteration ceilings from breaking startup by falling back to a safe default and logging an error.
- Maintain valid pre-hashed passwords across plugin restarts and ensure they cannot be reused once moved into password history.

Enhancements:
- Introduce an upgrade step that adds the optional PBKDF2 configuration object class to existing C PBKDF2-SHA256 plugin entries without failing the upgrade if it cannot be applied.
- Set a default PBKDF2 accept-max-iterations limit, decouple it from the dynamic work-factor calculation, and cap the effective rounds at this ceiling.
- Refine PBKDF2 configuration parsing to validate iteration ceilings, clamp invalid values, and log clearer diagnostics.
- Expand unit and integration tests to cover PBKDF2 iteration ceiling behavior, legacy-to-modern PBKDF2-SHA256 migration, and safe handling of malformed hashes.

### DIFF
--- a/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
+++ b/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
@@ -58,11 +58,15 @@
  * it's still better than ssha512.
  */
 #define PBKDF2_MINIMUM 2048
-#define PBKDF2_ACCEPT_MAX_ITERATIONS_DEFAULT 50000
 #define PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR "nsslapd-pwdPBKDF2AcceptMaxIterations"
 
 static uint32_t pbkdf2Iterations = 8192;
 static uint32_t pbkdf2AcceptMaxIterations = PBKDF2_ACCEPT_MAX_ITERATIONS_DEFAULT;
+
+/* Rejections of out-of-range iteration counts are logged at most once per interval */
+#define PBKDF2_REJECT_LOG_INTERVAL 300
+static uint64_t pbkdf2RejectLastLogTime = 0;
+static uint64_t pbkdf2RejectSuppressed = 0;
 
 static const char *schemeName = PBKDF2_SHA256_SCHEME_NAME;
 static const uint32_t schemeNameLength = PBKDF2_SHA256_NAME_LEN;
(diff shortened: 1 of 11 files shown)


ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -278,15 +280,10 @@ pbkdf2_sha256_pw_cmp(const char *userpwd, const char *dbpwd)
...
-
-    if (accept_max == 0) {
-        accept_max = pbkdf2Iterations;
-    }
-    if (iterations < PBKDF2_MINIMUM || iterations > accept_max) {
+    if (iterations < PBKDF2_MINIMUM || iterations > pbkdf2AcceptMaxIterations) {
         slapi_log_err(SLAPI_LOG_ERR, (char *)schemeName,
                       "PBKDF2 iteration count %" PRIu32 " is out of range (max %" PRIu32 ")\n",
-                      iterations, accept_max);
+                      iterations, pbkdf2AcceptMaxIterations);
```
tbordaz: If the stored hash has a higher number of iterations than the default/configured one then authentication will fail. Shouldn't we suggest the command to launch to allow successful auth ?

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
```diff
@@ -385,21 +382,22 @@ pbkdf2_sha256_get_accept_max_iterations(Slapi_PBlock *pb, uint32_t default_max)
...
+                          " and %d; using the default %" PRIu32 "\n",
+                          PBKDF2_ACCEPT_MAX_ITERATIONS_ATTR, configured,
+                          PBKDF2_MINIMUM, INT_MAX, default_max);
+        } else {
+            accept_max = (uint32_t)configured;
         }
-        accept_max = (uint32_t)configured;
     }
 
     pbkdf2AcceptMaxIterations = accept_max;
```
tbordaz: To be honest it looks surprising to me to set pbkdf2AcceptMaxIterations in a _get_ callback

tbordaz (commented):
minor comments

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
droideck: Sounds good! Done.
Added `dsconf ... set-accept-max-iterations` command and improved the logging message. Also, throttled repeated messages a bit to prevent log flooding.

ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c
droideck: Yeah, it was already in the helper's code, I just moved it around a bit. But yeah, I agree it looks strange here.
I moved the part into startup callback. The renamed read helper now only returns the configured value; the global assignment happens explicitly during startup.

Basic idea about the assignment -- we never generate hashes that verification would reject. And I've added logging message so it's not silent.

tbordaz (approved):
LGTM

Issue #7652, opened 2026-07-16
Change bind result with NULL dn and non NULL password


progier389:
**Issue Description**
We should change the simple bind result when no DN is specified but a  password is specified.
The behavior of this case is not specified by RFC 4513
So actual result  (handling it as an anonymous bind) is not really a bug, but returning INVALID_CREDENTIAL would be better as it would avoid buggy application to consider that it authenticates successfully while it just bind anonymously.

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: Any version
 - Browser: N/A

**Steps to Reproduce**
1. Use python ldap to perform a simple bind with no DN but a password
2. Result is SUCCESS instead of INVALID_CREDENTIAL

**Expected results**
Should get INVALID_CREDENTIAL and be bound anonymously (as usual after a bind failure)

**Additional context**
Jira issue: IDM-7181

PR #7653, opened 2026-07-16
Issue 7652 - change bind result with password but no dn

progier389:
Reject bind operation with password and no dn whose behavior is not defined in RFC 4513

Issue: #7652

Reviewed by: @tbordaz , @mreynolds389 (Thanks!)

## Summary by Sourcery

Reject LDAP bind operations that provide a password with a null DN and add coverage to ensure anonymous binds remain allowed while password-without-DN binds are rejected.

Bug Fixes:
- Prevent undefined LDAP bind behavior by rejecting binds that use a null DN with a non-empty password.

Tests:
- Add a basic suite test verifying that binds with no DN succeed only when no password is supplied and fail with INAPPROPRIATE_AUTH when a password is provided.

### DIFF
--- a/ldap/servers/slapd/bind.c
+++ b/ldap/servers/slapd/bind.c
@@ -475,6 +475,15 @@ do_bind(Slapi_PBlock *pb)
 
         /* accept null binds */
         if (dn == NULL || *dn == '\0') {
+            if (cred.bv_len > 0) {
+                /* RFC 4513 does not specify this case. Let's reject it. */
+                send_ldap_result(pb, LDAP_INAPPROPRIATE_AUTH, NULL,
+                                 "Bind with a null DN and non-empty password is rejected", 0, NULL);
+                /* increment BindSecurityErrorcount */
+                slapi_counter_increment(g_get_per_thread_snmp_vars()->ops_tbl.dsBindSecurityErrors);
+                slapi_log_security(pb, SECURITY_BIND_FAILED,  SECURITY_MSG_INVALID_PASSWD);
+                goto free_and_return;
+            }
             slapi_counter_increment(g_get_per_thread_snmp_vars()->ops_tbl.dsAnonymousBinds);
             /* by definition anonymous is also unauthenticated so increment
                that counter */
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/bind.c
```diff
@@ -475,6 +475,14 @@ do_bind(Slapi_PBlock *pb)
 
         /* accept null binds */
         if (dn == NULL || *dn == '\0') {
+            if (cred.bv_len > 0) {
+                /* RFC 4513 does not specify this case. Lets reject it. */
+                send_ldap_result(pb, LDAP_INAPPROPRIATE_AUTH, NULL,
```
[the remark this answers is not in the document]
progier389: LDAP_INAPPROPRIATE_AUTH is also returned by the server when attempting anonymous bind if they are disabled, so it seems a reasonable error code

ldap/servers/slapd/bind.c
```diff
@@ -475,6 +475,14 @@ do_bind(Slapi_PBlock *pb)
...
+                                 "bind with Null DN and a non NULL password is rejected", 0, NULL);
+                /* increment BindSecurityErrorcount */
+                slapi_counter_increment(g_get_per_thread_snmp_vars()->ops_tbl.dsBindSecurityErrors);
```
mreynolds389, line 483: Should we add this to the Security log?  

`slapi_log_security(pb, SECURITY_BIND_FAILED, SECURITY_MSG_NO_ENTRY | SECURITY_MSG_INVALID_PASSWD);`  

If we update snmp we should probably update our log.  Not sure if SECURITY_MSG_NO_ENTRY or SECURITY_MSG_INVALID_PASSWD is better, or we create a new message.

ldap/servers/slapd/bind.c
```diff
@@ -475,6 +475,14 @@ do_bind(Slapi_PBlock *pb)
 
         /* accept null binds */
         if (dn == NULL || *dn == '\0') {
+            if (cred.bv_len > 0) {
+                /* RFC 4513 does not specify this case. Lets reject it. */
```
tbordaz: Minor comment. 
Should we add the rejection after increasing dsAnonymousBinds and dsUnAuthBinds like it is done when config_get_anon_access_switch=off.

tbordaz (approved):
LGTM

ldap/servers/slapd/bind.c
progier389: I have no strong opinion about that one. 
The only point is that if we decide to log that one in security log, we should also add a log about the 
config_get_anon_access_switch() == SLAPD_ANON_ACCESS_OFF  error case 
which is very similar

ldap/servers/slapd/bind.c
progier389: I wondered the same when I wrote the fix and decided not to do it because it is neither an anonymous/unauthenticated bind but rather a plain invalid one

PR #7656, opened 2026-07-21
Issue 7655 - Slow search when using many DN syntax components and lot…

### LINKED ISSUE #7655: Slow search when using many DN syntax components and lot of candidate entries
**Issue Description**
The filter creates a candidate list of entry_IDs. The list can contain entryID that actually do not match the filter so before returning a given entry (from the list) the server check if the entry matches the filter.
An entry with large valueset (like static group) sorts the valueset.
The RFE #6172 allows to use sorted valueset to match a filter component containing a DN syntax attribute. (e.g. (uniquemember=cn=foo,dc=example))

If the entry does not contain a large valueset the call to sorted valueset is slighly slower.

If the filter contains many components '(attr=value)' with attr syntax being DN. Then the evaluation of the full filter is slower.
In addition if the candidate list is long the overall SRCH is much slower than if bypass the sorted valueset.

Tests with 300+ components and 600 IDs shows that the SRCH is twice slower

All versions  #6172 is present in 1.4.3 and later

**Steps to Reproduce**
1.Create large filter with 300+ OR components with (attr=value) and attr is DN type. Make sure that the filter is indexed but will return no entry (to avoid sending time) like adding a AND NOT (attr=value)
2. create 1000 entries matching one component (attr=value)
4. run SRCH (directory manager for the ACL) and check etime

tbordaz:
… of candidate entries

Bug description:
	If the entry does not contain a large valueset the call to sorted valueset is
	slighly slower
	If the filter contains many components '(attr=value)' with attr syntax being DN.
	Then the evaluation of the full filter is slower.
	In addition if the candidate list is long the overall SRCH is much slower than
	if bypass the sorted valueset.

	Tests with 300+ components and 600 IDs shows that the SRCH is twice slower

Fix description:
	Prefer direct call to plugin_call_syntax_filter_ava
	when the valueset is not sorted

fixes: #7655

Assisted by: Cursor

Reviewed by:

## Summary by Sourcery

Improve performance of DN syntax attribute equality filters by avoiding the sorted valueset path when the valueset is not sorted.

Bug Fixes:
- Fix slow search evaluations for filters with many DN syntax components over large candidate sets by selectively using direct syntax filter evaluation.

Enhancements:
- Add an API to check whether a Slapi_ValueSet maintains a sorted index and use it to guard the DN-specific optimized filter path.

### DIFF
--- a/ldap/servers/slapd/valueset.c
+++ b/ldap/servers/slapd/valueset.c
@@ -595,6 +595,15 @@ slapi_valueset_isempty(const Slapi_ValueSet *vs)
 }
 
 int
+slapi_valueset_is_sorted(const Slapi_ValueSet *vs)
+{
+    if (NULL != vs) {
+        return (vs->sorted != NULL);
+    }
+    return 0;
+}
+
+int
 valueset_isempty(const Slapi_ValueSet *vs)
 {
     if (NULL == vs) {
--- a/ldap/servers/slapd/filterentry.c
+++ b/ldap/servers/slapd/filterentry.c
@@ -333,6 +333,7 @@ test_ava_filter(
         for (; a != NULL; a = a->a_next) {
             if (slapi_attr_type_cmp(ava->ava_type, a->a_type, SLAPI_TYPE_CMP_SUBTYPE) == 0) {
                 if ((ftype == LDAP_FILTER_EQUALITY) &&
+                    (slapi_valueset_is_sorted(&a->a_present_values)) &&
                     (slapi_attr_is_dn_syntax_type(a->a_type))) {
                     /* This path is for a performance improvement */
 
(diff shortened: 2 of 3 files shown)


ldap/servers/slapd/valueset.c
```diff
@@ -594,6 +594,15 @@ slapi_valueset_isempty(const Slapi_ValueSet *vs)
     return 1;
 }
 
+int
```
progier389, line 597: Should better use bool/true/false rather than int in that case

progier389 (approved):
LGTM except a minor comment

droideck (approved):
LGTM!
Also, note that performance measurements in https://github.com/389ds/389ds.github.io/pull/31 are done with the PR in the branch (lookup off column in the Performance section's table)

PR #7657, opened 2026-07-21
Backport Issue 7519 — ignore obsolete entrydn when entryrdn is in use

### LINKED ISSUE #7654: Backport Issue 7519 — ignore obsolete entrydn when entryrdn is in use
**Issue Description**
Upgraded instances can keep leftover cn=entrydn / entrydn.db*. With entryrdn in use, that can break searches and reindex. 
Fixed on 2.8 by 539cad435 (#7526). This issue tracks adapting that fix to main. 3.2, 3.1 (no switch branches).

A straight cherry-pick to 3.1+/main conflicts because Issue 6639 removed nsslapd-subtree-rename-switch / entryrdn_get_switch(), which the original patch still uses.

Backport approach, two variants:
- With switch (2.8, 3.0, …): cherry-pick original; ignore entrydn only when switch is on
- Without switch (3.1, main): resolve/adapt conflicts; always ignore when index name is entrydn (no switch APIs)
Include DSBLE0008 + CI test on both.

 - Package and version: Branches main, 3.2, 3.1

jchapma:
Description:
Upgraded instances can retain leftover cn=entrydn while entryrdn is in use, causing bad searches and reindex failures. Fixed on 2.8 by 539cad4 (#7519/#7526). This adapts that fix for branches where nsslapd-subtree-rename-switch was removed (Issue #6639).

Fix:
Always ignore leftover entrydn by index name, skip reindex and add healthcheck DSBLE0008 + CI test.

Relates: https://github.com/389ds/389-ds-base/issues/7519
Fixes: https://github.com/389ds/389-ds-base/issues/7654

Reviewed by:

## Summary by Sourcery

Backport healthcheck and index handling changes to ignore obsolete entrydn indexes when entryrdn is in use, and surface them as a dedicated healthcheck issue.

Bug Fixes:
- Prevent db2index and index configuration operations from attempting to use obsolete entrydn indexes when entryrdn-based indexing is active, avoiding inconsistent searches and reindex failures.

Enhancements:
- Add a new DSBLE0008 lint definition and backend healthcheck to detect obsolete entrydn index configuration and on-disk entrydn database files.
- Extend healthcheck test helpers and add a new system index healthcheck test to verify DSBLE0008 is reported when entrydn is configured and no longer reported once it is removed.

Tests:
- Enhance the healthcheck test harness to assert both presence and absence of specific healthcheck codes, and add coverage for obsolete entrydn index detection.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
@@ -1493,11 +1493,11 @@ bdb_db2index(Slapi_PBlock *pb)
                     slapi_log_err(SLAPI_LOG_INFO, "bdb_db2index", "%s: Indexing: %s\n",
                                   inst->inst_name, LDBM_ENTRYRDN_STR);
                     index_ext |= DB2INDEX_ENTRYRDN;
-                } else if (strcasecmp(attrs[i] + 1, LDBM_ENTRYDN_STR) == 0) {
+                } else if (ldbm_index_entrydn_should_ignore(attrs[i] + 1)) {
                     slapi_task_log_notice(task, "%s: Requested to index %s, but the index is no longer applicable",
                                           inst->inst_name, LDBM_ENTRYDN_STR);
                     slapi_log_err(SLAPI_LOG_WARNING,
-                                  "bdb_db2index", "%s: Requested to index %s,but the index is no longer applicable\n",
+                                  "bdb_db2index", "%s: Requested to index %s, but the index is no longer applicable\n",
                                   inst->inst_name, LDBM_ENTRYDN_STR);
                     goto err_out;
                 } else {
--- a/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
+++ b/ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
@@ -1494,12 +1494,13 @@ bdb_db2index(Slapi_PBlock *pb)
                                   inst->inst_name, LDBM_ENTRYRDN_STR);
                     index_ext |= DB2INDEX_ENTRYRDN;
                 } else if (ldbm_index_entrydn_should_ignore(attrs[i] + 1)) {
+                    /* Warn and skip obsolete entrydn*/
                     slapi_task_log_notice(task, "%s: Requested to index %s, but the index is no longer applicable",
                                           inst->inst_name, LDBM_ENTRYDN_STR);
                     slapi_log_err(SLAPI_LOG_WARNING,
                                   "bdb_db2index", "%s: Requested to index %s, but the index is no longer applicable\n",
                                   inst->inst_name, LDBM_ENTRYDN_STR);
-                    goto err_out;
+                    break;
                 } else {
                     if (strcasecmp(attrs[i] + 1, SLAPI_ATTR_OBJECTCLASS) == 0) {
                         index_ext |= DB2INDEX_OBJECTCLASS;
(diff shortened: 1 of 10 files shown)


ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -1240,6 +1240,20 @@ process_db2index_attrs(Slapi_PBlock *pb, ImportCtx_t *ctx)
             if (pt != NULL) {
                 *pt = '\0';
             }
+            if (ldbm_index_entrydn_should_ignore(attrname)) {
```
droideck, line 1243: IIUC, if entrydn was the only requested attribute, the list stays NULL after this skip, and is_reindexed_attr() treats an empty list as "reindex everything". So `--attr entrydn` becomes a full rebuild of every index and VLV, while the log says the index was skipped. This comes from 539cad435, but  I think it's worth fixing here...
Unless I miss something, probably worth adding a test too (instance with the leftover entrydn config, run reindex --attr entrydn, check the task log if it rebuilds every index)
A possible fix: if the list ends up empty after the skip, end the task as a no-op.

ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
```diff
@@ -1493,11 +1493,11 @@ bdb_db2index(Slapi_PBlock *pb)
...
                     index_ext |= DB2INDEX_ENTRYRDN;
-                } else if (strcasecmp(attrs[i] + 1, LDBM_ENTRYDN_STR) == 0) {
+                } else if (ldbm_index_entrydn_should_ignore(attrs[i] + 1)) {
                     slapi_task_log_notice(task, "%s: Requested to index %s, but the index is no longer applicable",
                                           inst->inst_name, LDBM_ENTRYDN_STR);
                     slapi_log_err(SLAPI_LOG_WARNING,
-                                  "bdb_db2index", "%s: Requested to index %s,but the index is no longer applicable\n",
+                                  "bdb_db2index", "%s: Requested to index %s, but the index is no longer applicable\n",
                                   inst->inst_name, LDBM_ENTRYDN_STR);
                     goto err_out;
```
droideck: If we have a BDB instance with leftover entrydn config, and we run a plain full reindex. Task aborted partway -> the earlier index files are gone.

IIUC, a plain full reindex enumerates every configured index, including the leftover entrydn (both the online task and offline db2index do this). Attributes processed before the abort were already set offline and their files erased, and err_out restores nothing. After restart those indexes are empty, which means silently wrong search results. Upgraded BDB installs have the this leftover, I think.

Should we warn and continue on BDB, same as the LMDB side?

droideck (changes requested):

ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
jchapma: Good catch, yes on bdb, aborting mid pass after earlier indexes were removed leave them empty, I will change bdb to warn and continue, like lmdb.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
droideck: I think this is still relevant. Or do I miss something? We can fix it in a separate PR, though.

droideck (approved):
LGTM!

ldap/servers/slapd/back-ldbm/db-bdb/bdb_ldif2db.c
jchapma: Ah yes, sorry, I thought I had  already fixed this. You’re right, empty lists after the skip were treated as reindex everything. Fixed this for mdb and bdb and add CI test.

Issue #7658, opened 2026-07-21
Potential sasl io overflow


mreynolds389:
**Issue Description**

Heap Buffer Overflow in sasl_io_recv() via Padded SASL UNBIND

In sasl_io.c, function sasl_io_recv(), when sasl_io_start_packet() returns
SASL_IO_BUFFER_NOT_ENCRYPTED (triggered by sending an unencrypted LDAP UNBIND
after SASL layer setup) there is no check that sp->encrypted_buffer_count <= len before the memcpy

mreynolds389:
e63592d5e..f15ffd63e  389-ds-base-3.2 -> 389-ds-base-3.2
3fc51b255..c7a9b25b0  389-ds-base-3.1 -> 389-ds-base-3.1
2f58e9355..433239293  389-ds-base-3.0 -> 389-ds-base-3.0
cdc57272d..4cbfeb1e4  389-ds-base-2.9 -> 389-ds-base-2.9
a23a87926..b9404b3c7  389-ds-base-2.8 -> 389-ds-base-2.8
bd9bf109f..1c39bf98a  389-ds-base-2.7 -> 389-ds-base-2.7
b160731bd..adfb2a269  389-ds-base-2.6 -> 389-ds-base-2.6
1fb7e6f83..624418967  389-ds-base-2.4 -> 389-ds-base-2.4
3f300f544..735ae70bf  389-ds-base-2.2 -> 389-ds-base-2.2
40ed5a73a..949f3d64c  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7659, opened 2026-07-21
Issue 7658 - Heap Buffer Overflow in sasl_io_recv() via Padded SASL UNBIND

mreynolds389:
Description:

In sasl_io.c, function sasl_io_recv(), when sasl_io_start_packet() returns SASL_IO_BUFFER_NOT_ENCRYPTED (triggered by sending an unencrypted LDAP UNBIND after SASL layer setup) there is no check that

sp->encrypted_buffer_count <= len before the memcpy

relates: https://github.com/389ds/389-ds-base/issues/7658

## Summary by Sourcery

Prevent heap buffer overflow when handling unencrypted SASL UNBIND packets and add an ASAN regression test for the scenario.

Bug Fixes:
- Clamp the number of bytes copied from the SASL encrypted buffer to the caller buffer length when processing unencrypted UNBIND messages to avoid heap buffer overflow.

Tests:
- Add an ASAN-based SASL I/O test that sends a padded LDAP UNBIND over a raw socket to verify the server does not crash or report a heap-buffer-overflow.

### DIFF
--- a/dirsrvtests/tests/suites/sasl/io_overflow_asan_test.py
+++ b/dirsrvtests/tests/suites/sasl/io_overflow_asan_test.py
@@ -1,0 +1,248 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import ldap
+import logging
+import pytest
+import os
+import socket
+import struct
+import sys
+import time
+from ldap import sasl as ldap_sasl
+from lib389._constants import DEFAULT_SUFFIX, PASSWORD
+from lib389.idm.user import UserAccounts
+from lib389.utils import check_asan_report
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/sasl/io_overflow_asan_test.py
```diff
@@ -0,0 +1,366 @@
...
+            conn.unbind_s()
+        except ldap.LDAPError:
+            pass
+
+    assert inst.status(), "Server crashed while processing padded UNBIND"
+    assert inst.get_pid() == server_pid, "Server restarted while processing padded UNBIND"
+
+    # Check ASAN report
+    log.info("[*] Checking ASAN report")
+    assert not check_asan_report(inst, 'heap-buffer-overflow'), "Heap-buffer-overflow found in ASAN report"
```
droideck: I think it should be used similar how the function used in dirsrvtests/tests/suites/logging/auditlog_pwd_mask_overflow_test.py 

`assert` seems incorrect for this case, or do I miss something?

droideck (approved):
Besides the minor issue, LGTM!

Issue #6275, opened 2024-07-25
Search request executes slowly, if filter contains many values (more than 100 in 'or' list). 


OGSmv:
**Issue Description**
Filter evaluation (or matching) non optimal. This causes slow execution time on search request with big filter.
One search request with 1000 uids in filter works slower than 2 requests by 500 uids in filter.

**Package Version and Platform:**
 - Platform: Debian (Buster)
 - Package and version: 389ds/dirsrv:2.4; 389ds/dirsrv:3.0

**Steps to Reproduce**
1. Start 389ds with simple structure
2. Create test organizational unit
3. Create many user accounts (for example 100000 accounts) in test ou.
4. Execute search requests for 1000 accounts with filter by uids.
5. Split request on two request (by 500 accounts) and execute search 2 times (by 500) + concatenation results. This works faster than one search for 1000 accounts 

**Expected results**
One search slower than several.
On screenshots we can see the result of reading 1000 entities in several strategies (Time spent in processing).

**Screenshots**
![image](https://github.com/user-attachments/assets/577e5857-c7cf-4ec8-9cb7-f17e0e9e5d57)
![image](https://github.com/user-attachments/assets/f385baaa-fa89-4756-a9ef-41c6166da111)

**Additional context**
[Python script](https://gist.github.com/OGSmv/de6b5a41dc308a78a90b14900491c95a) to reproduce bug.
Command:
`python3 read_test.py --host <ip> --port <port> --user "cn=Directory Manager" --pass <password> --root <root_dn>`

progier389:
Hi, 
Interesting study.
I am not surprised:  IMHO that test is showing the impact of the candidates list merge in the or operation (the one that  ensures that search operation returns an entry only once). As it is the worst case merging 1000 time a single candidate the algorithm cannot be very efficient. I think that complexity is something like  ( n  * n ) as we store candidates in an array
 
I do not think that we will try to improve that  (anyway, we do not want to encourage  developer to use complex filters)
At least this clearly  shows that whwn writing ldap application, you has better use a reasonable  number of sub-filters in OR filter.

OGSmv:
Hi @progier389,

Do you mean that it is normal case to use many simple requests to obtain 1000+ unique entities, than use one request ?

Firstyear:
@progier389 See:

https://github.com/389ds/389-ds-base/blob/main/ldap/servers/slapd/back-ldbm/back-ldbm.h#L270
[...]

PR #7665, opened 2026-07-23
Issue 7664 - RFE - Add equality lookup tables for large OR filters

### LINKED ISSUE #7664: Per-entry filter test is O(candidates x branches) for large equality OR filters
**Description**
A search whose filter is a large OR of equality tests on one attribute,

    (|(uid=user1)(uid=user2) ... (uid=user1000))

is slow even when every value is indexed, and the time grows with result size times branch count.
The cost is not in candidate generation. Before returning an entry the server verifies it against the filter, and that test walks the OR branch by branch; every branch rescans the entry's attributes and re-normalizes its values through the syntax plugin.
Measured with the same 1000-branch uid OR against the same database:

- 999 values absent from the database: 7 ms
- all 1000 values present: 1.07 s

Index lookups and the candidate-list union are identical in both runs (each uid branch matches at most one entry, so the union input is about a thousand IDs); only the number of live entry-times-branch evaluations differs.

#6275 reports this slowness, but the analysis there pointed at the O(n^2) IDL merge in idl_set.c. For this filter shape the merge input is
tiny and the per-entry filter test dominates, so this issue tracks the filter-test fix separately.

droideck:
Description:
The per-entry filter test walks every OR branch for every candidate entry, and each branch re-normalizes the entry's values through the syntax plugin, so a filter like (|(uid=v1)...(uid=v1000)) costs candidates x branches even when fully indexed. The time is there, not in candidate generation: the same 1000-branch OR costs 7 ms when 999 values are absent and over a second when all are live.

Build a sorted table of the branch values once per search, on the backend's private normalized filter copy, and let each entry find its branch with one normalization and a binary search. The table only selects which branch to test: a hit still runs the same access check and match call as the classic walk, all-miss entries are decided directly where false and undefined are indistinguishable (no NOT ancestor, not VLV), and anything the table cannot decide exactly falls back to the untouched walk.

Eligibility is narrow: at least 16 equality branches on one attribute, a string-family or DN syntax whose equality is byte equality of the normalized forms, a standard matching rule, and no attribute options. The largest qualifying family wins, first occurrence breaking ties. nsslapd-enable-or-filter-lookup (default on) turns the feature off.

On a 100,000-entry database 10,000-candidate search drops from 23 s to 0.4 s, with no regression beyond noise on ineligible shapes.

Also fix a latent double free in list_candidates: idl_set_insert_idl frees an ALLIDS list it is handed, so stop keeping a local alias to it.

Coverage: result parity across matching rules, ACLs, paging, and VLV; strict log-based feature contracts; an ASan filter-lifecycle module.

Fixes: https://github.com/389ds/389-ds-base/issues/7664
Relates: https://github.com/389ds/389-ds-base/issues/6275

Assisted by: Claude (investigation and tests)

Reviewed by: ?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1134,6 +1134,7 @@ libslapd_la_SOURCES = ldap/servers/slapd/add.c \
 	ldap/servers/slapd/features.c \
 	ldap/servers/slapd/fileio.c \
 	ldap/servers/slapd/filter.c \
+	ldap/servers/slapd/filter_or_lookup.c \
 	ldap/servers/slapd/filtercmp.c \
 	ldap/servers/slapd/filterentry.c \
 	ldap/servers/slapd/generation.c \
--- a/dirsrvtests/tests/suites/filter/filter_large_filter_interaction_test.py
+++ b/dirsrvtests/tests/suites/filter/filter_large_filter_interaction_test.py
@@ -1,0 +1,248 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+
+"""Functional interactions between large equality ORs and costly reads.
+
+4,200 entries share the substring and approximate keys, an indexed AND
+selects exactly 612 of them, and a 355-branch DN equality family appears
+in every filter.
+"""
+
+import ldap
+import logging
+import os
+import pytest
+
+from ldap.controls import SimplePagedResultsControl
+from lib389.utils import ensure_str
+from test389.topologies import topology_st as topo
+
+from .filter_large_filter_support import (
+    APPROX_ATTR,
+    EXCLUDED_ONE,
+    EXCLUDED_TWO,
+    FALLBACK_ATTR,
+    GUARD_ATTR,
+    GUARD_VALUE,
+    HIT_POSITIONS,
+    OUTER_A,
+    OUTER_B,
+    OUTER_VALUE,
+    SUBSTRING_ATTR,
+    TEST_BASE,
+    combined_assertions,
+    combined_filter,
+    dn_or,
+    hit_assertions,
+    interaction_data,
+    lookup_disabled,
+    miss_assertions,
+    search_dns,
+)
+
+
+pytestmark = pytest.mark.tier1
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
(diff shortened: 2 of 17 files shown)


tbordaz:
I find the idea/implementation excellent. However I wonder if it could not be extended.
If I am correct, it builds an accelerator for large OR (only if ava are equality and same attribute) so that normalized filter component assertion (ava) are stored in an sorted array, so that matching of candidate attribute values does not require to normalize again and again (at each candidate) the assertion and to accelerate the finding of the possible matching assertion.

Something we could extend is that assuming that there are very often a very large set of candidate (e.g. >1000) we could systematically normalize the assertion (only for equality ava) in the filter. Even for AND, NOT lists. So that we normalize filter assertion once instead of 1000 times.
Then we can add the accelerator for large OR (equality ava and same attribute) that sort the ava values.

I think it could be extended in a separated ticket, no need to make this PR too complex. But this PR should be compatible with future extension. WDYT

droideck:
@tbordaz, agreed, it's worth exploring!
 
One correction though... Assertion normalization is already done once per search, for the whole filter tree, not just the OR. `ldbm_back_search` runs `slapi_filter_normalize` with value normalization on the operation-private copy (AND/NOT lists included), and `ldbm_search_compile_filter` flags each ava so `plugin_call_syntax_filter_ava` skips re-normalizing the assertion at test time.
The lookup table actually depends on that: its keys are borrowed pointers to the already-normalized values. That also means the PR is compatible with the extension as is when we'll look into it.

But I think the idea is sound and we still have things to do, possibly.
The DN equality path constructs a Slapi_Value from the assertion on every per-entry call (`slapi_value_new_berval` in `test_ava_filter`) just to feed `slapi_valueset_find`. IIUC, those bytes never change during the search, so we could pre-build that value once at compile time and cache it on the node. Only wrinkle: ava_private is already taken by the flags pointer, so the cache needs a small struct or a new field.
Another point where we can do something: `plugin_call_syntax_filter_ava_sv allocates and frees a pblock per compare, per candidate. 

And one huge thing which I'd keep as a separate optimization, I think... Filters outside the search path (ACI targetfilter, roles, psearch, plugin configs) really do renormalize assertions per evaluation, because those trees never get the normalize pass. Extending normalize-once there would be nice! But those trees are shared, long-lived, and evaluated concurrently (and normalization mutates the value in place) so each consumer needs its own ownership story (ACL normalizing at ACI-parse time would be the first candidate, IMO).

So yeah, I think it's certainly a separate ticket territory. This PR should be okay as is... The lookup table consumes the normalize pass instead of doing its own (it borrows the already-normalized bytes and doesn't touch the tree after the build), and it only ever lives on the backend's per-search filter copy. So whatever we end up changing about where normalization happens, the table just picks up whatever values are there, and it can't collide with the shared trees we'd need to touch for the ACI/roles part.

progier389 (approved):
LGTM

PR #7667, opened 2026-07-24
Issue 7666 - Replication performance degradation during total init on…

### LINKED ISSUE #7666: Replication performance degradation during total init on high-latency storage
**Issue Description**
During replication total init the consumer’s LMDB writer thread commits transactions with fsync on each commit. This causes progressive performance degradation as the database grows. This is especially noticeable on network-attached storage, for example AWS EBS, where fsync latency is high and grows under sustained write load.

The offline LDIF import sets MDB_NOSYNC on the LMDB environment during import, which skips per-commit fsync, and only flushes at the end via mdb_env_sync. But the online total init doesn’t do that.

 - Package and version: 389-ds-base-3.2.x. Affects all versions with LMDB.

**Steps to Reproduce**
1. Set up 2 instances with replication and run total init

**Expected results**
Steady import rate

**Actual results**
Import rate degrades over time
```
Processed 18123 entries -- average rate 906.1\/sec,recent rate 906.1\/sec,
Processed 31704 entries -- average rate 792.6\/sec,recent rate 792.6\/sec,
Processed 40936 entries -- average rate 682.3\/sec,recent rate 570.3\/sec,
Processed 49426 entries -- average rate 610.2\/sec,recent rate 432.2\/sec,
Processed 56427 entries -- average rate 558.7\/sec,recent rate 377.8\/sec,
Processed 62493 entries -- average rate 516.5\/sec,recent rate 326.7\/sec,
Processed 70093 entries -- average rate 497.1\/sec,recent rate 341.6\/sec,
Processed 75642 entries -- average rate 469.8\/sec,recent rate 328.7\/sec,
Processed 82006 entries -- average rate 453.1\/sec,recent rate 297.8\/sec,
Processed 89158 entries -- average rate 443.6\/sec,recent rate 337.9\/sec,
Processed 95724 entries -- average rate 433.1\/sec,recent rate 342.9\/sec,
```

**Additional context**
With MDB_NOSYNC:
```
[2026-07-24T12:23:02.320097] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 39647 entries -- average rate 1982.3/sec, recent rate 1982.3/sec,
[2026-07-24T12:23:22.389082] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 86119 entries -- average rate 2153.0/sec, recent rate 2152.9/sec,
[2026-07-24T12:23:42.418362] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 127510 entries -- average rate 2125.2/sec, recent rate 2196.6/sec,
[2026-07-24T12:24:02.507150] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 172510 entries -- average rate 2129.8/sec, recent rate 2107.1/sec,
[2026-07-24T12:24:22.595750] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 216437 entries -- average rate 2142.9/sec, recent rate 2169.0/sec,
[2026-07-24T12:24:42.624749] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 261083 entries -- average rate 2157.7/sec, recent rate 2214.3/sec,
[2026-07-24T12:25:02.705228] - INFO - dbmdb_import_monitor_threads - import userroot: Processed 282635 entries -- average rate 2004.5/sec, recent rate 1655.0/sec,
```

vashirov:
… high-latency storage

Bug Description:
During replication total init the consumer's LMDB writer thread commits
transactions with fsync on each commit. This causes progressive
performance degradation as the database grows. This is especially
noticeable on network-attached storage, for example AWS EBS, where fsync
latency is high and grows under sustained write load.

The offline LDIF import sets MDB_NOSYNC on the LMDB environment during
import, which skips per-commit fsync, and only flushes at the end via
mdb_env_sync. But the online total init doesn't do that.

Additionally, the nsslapd-db-durable-transactions config parameter was
not honored by the LMDB backend (only BDB checked it).

Fix Description:
- Add nsslapd-mdb-online-import-nosync config parameter (default: off)
  that sets MDB_NOSYNC on the LMDB environment during online import, same
  as offline LDIF import. The final mdb_env_sync() in the writer thread
  ensures all data is flushed when import completes.

- Honor nsslapd-db-durable-transactions for the LMDB backend by
  setting MDB_NOSYNC when the environment is opened if this parameter
  is set to off. Previously this config was only effective for BDB.

Fixes: https://github.com/389ds/389-ds-base/issues/7666

## Summary by Sourcery

Bug Fixes:
- Prevent replication total initialization from suffering performance degradation on high-latency storage by disabling per-commit fsync during LMDB bulk import and re-enabling it when the instance returns online.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
+++ b/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
@@ -781,23 +781,24 @@ dbmdb_import_all_done(ImportJob *job, int ret)
             index->ai->ai_indexmask &= ~INDEX_OFFLINE;
             index = index->next;
         }
-        /* Re-enable fsync before going back online */
-        {
+        /* Re-enable fsync before going back online, but only if we set it */
+        if (job->nosync_set) {
             struct ldbminfo *li = (struct ldbminfo *)inst->inst_be->be_database->plg_private;
             dbmdb_ctx_t *ctx = MDB_CONFIG(li);
             if (ctx->env) {
-                unsigned int env_flags = 0;
-                mdb_env_get_flags(ctx->env, &env_flags);
-                if (env_flags & MDB_NOSYNC) {
-                    int nosync_rc = mdb_env_set_flags(ctx->env, MDB_NOSYNC, 0);
-                    if (nosync_rc != 0) {
-                        slapi_log_err(SLAPI_LOG_ERR, "dbmdb_import_all_done",
-                                      "Failed to clear MDB_NOSYNC flag "
-                                      "(error %d: %s)\n",
-                                      nosync_rc, mdb_strerror(nosync_rc));
-                    }
+                int nosync_rc = mdb_env_set_flags(ctx->env, MDB_NOSYNC, 0);
+                if (nosync_rc != 0) {
+                    slapi_log_err(SLAPI_LOG_ERR, "dbmdb_import_all_done",
+                                  "Failed to clear MDB_NOSYNC flag "
+                                  "(error %d: %s)\n",
+                                  nosync_rc, mdb_strerror(nosync_rc));
+                } else {
+                    slapi_log_err(SLAPI_LOG_INFO, "dbmdb_import_all_done",
+                                  "MDB_NOSYNC cleared. "
+                                  "Per-commit fsync re-enabled.\n");
                 }
             }
+            job->nosync_set = 0;
         }
         /* start up the instance */
         rc = dbmdb_instance_start(job->inst->inst_be, DBLAYER_NORMAL_MODE);
@@ -1171,6 +1172,30 @@ error:
         priv->dblayer_auto_tune_fn(li);
     }
 
+    /*
+     * If NOSYNC was set by us and not yet cleared (e.g. import thread failed
+     * before reaching dbmdb_import_all_done), clear it now to restore
+     * durability for all backends.
+     */
+    if (job->nosync_set) {
+        struct ldbminfo *li = inst->inst_li;
+        dbmdb_ctx_t *mdb_ctx = MDB_CONFIG(li);
+        if (mdb_ctx->env) {
+            int nosync_rc = mdb_env_set_flags(mdb_ctx->env, MDB_NOSYNC, 0);
+            if (nosync_rc != 0) {
+                slapi_log_err(SLAPI_LOG_ERR, "dbmdb_public_dbmdb_import_main",
+                              "Failed to clear MDB_NOSYNC after import failure "
+                              "(error %d: %s)\n",
+                              nosync_rc, mdb_strerror(nosync_rc));
+            } else {
+                slapi_log_err(SLAPI_LOG_INFO, "dbmdb_public_dbmdb_import_main",
+                              "MDB_NOSYNC cleared after import failure. "
+                              "Per-commit fsync re-enabled.\n");
+            }
+        }
+        job->nosync_set = 0;
+    }
+
     /* Re-enable the ndn cache */
     ndn_cache_dec_import_task();
     dbmdb_clear_dirty_flags(be);
@@ -1431,7 +1456,6 @@ dbmdb_bulk_import_start(Slapi_PBlock *pb)
     backend *be = NULL;
     PRThread *thread = NULL;
     int ret = 0;
-    int nosync_set = 0;  /* Track whether we set MDB_NOSYNC */
 
     slapi_pblock_get(pb, SLAPI_BACKEND, &be);
     if (be == NULL) {
@@ -1522,7 +1546,11 @@ dbmdb_bulk_import_start(Slapi_PBlock *pb)
                               "Import will continue with fsync enabled.\n",
                               ret, mdb_strerror(ret));
             } else {
-                nosync_set = 1;
+                job->nosync_set = 1;
+                slapi_log_err(SLAPI_LOG_INFO, "dbmdb_bulk_import_start",
+                              "MDB_NOSYNC enabled for online import "
+                              "(nsslapd-mdb-online-import-nosync: on). "
+                              "Per-commit fsync is disabled until import completes.\n");
             }
         }
     }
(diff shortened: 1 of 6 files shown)


progier389:
Although it provides better performance this change as a major drawback:
  The environment flag is global and impacts all backends so setting it to MDB_NOSYNC disables the durability for all the backends 
 This is not really an issue with the off line import because only the imported backend is impacted and it needs to be reimported if the import aborts.  
But with on line way if there are several backends, so a txn may be completed (on the other backends) without being properly written on the disk because of the import) which could be a problem for some people in case of system crash. 
(The probability is low because the txn need to occurs while the import writer thread does not held the write txn, but it exists)
  So I wonder if we should not rather have a parameter to allow/disable this behavior

vashirov:
@progier389, I didn't consider multiple backends, that's a very good point!

> So I wonder if we should not rather have a parameter to allow/disable this behavior

Do you mean something specific for bulk import like `nsslapd-bulk-import-nosync`? 
We also have `nsslapd-db-durable-transactions`, but it's completely ignored on LMDB. Perhaps we should fix it too.

progier389:
> Do you mean something specific for bulk import like `nsslapd-bulk-import-nosync`?

Yes

I agree: It would be nice to fix the nsslapd-db-durable-transactions on lmdb too.

vashirov:
@progier389, should this new parameter affect both offline and online import or only online? I'm thinking of the interaction with `nsslapd-db-durable-transactions` when it's off.

progier389:
> @progier389, should this new parameter affect both offline and online import or only online? I'm thinking of the interaction with `nsslapd-db-durable-transactions` when it's off.

For a technical point of view, it should only affect the online import (including replication bulk import).
It seems better to not have txn durability for off line import : (Better performance and no drawback to postpone the actual write until the end of the import, (anyway if the import is aborted for whatever reason, the backend will then be broken and re imported before being usable))
That said for parameter description clarity we may choose otherwise ...

jchapma (approved):
LGTM, the only comment i have is the same as  sourcery, do we need t check the return value of mdb_env_set_flags

vashirov:
Ok, changes applied, new parameter is nsslapd-mdb-online-import-nosync (default: off), added a test to measure the improvement.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -1484,6 +1503,30 @@ dbmdb_bulk_import_start(Slapi_PBlock *pb)
...
+     * durability is not needed. The writer thread calls mdb_env_sync() at the end.
+     * Note: MDB_NOSYNC is environment-level and affects all backends.
+     */
+    if (MDB_CONFIG(li)->dsecfg.online_import_nosync) {
+        dbmdb_ctx_t *ctx = MDB_CONFIG(li);
+        unsigned int env_flags = 0;
+
+        mdb_env_get_flags(ctx->env, &env_flags);
+        if (!(env_flags & MDB_NOSYNC)) {
+            ret = mdb_env_set_flags(ctx->env, MDB_NOSYNC, 1);
```
droideck, line 1542: If we enable the knob, start a total init, kill the supplier mid-stream, then check the env flags, NOSYNC will be still set, IIUC.

So the clear covers failures before the import thread starts. Failures after that (disk full, bad entry, dropped supplier connection) go to the else if (0 != ret) arm at
https://github.com/389ds/389-ds-base/blob/65a1afcdee407b85d6d1720560836f0a5f37ebfc/ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c#L1129-L1131
which never clears the flag. So it's possible that we'll leave the whole env non-durable, including other suffixes and the changelog, until something else happens to clear it.
I think we should add a clear on the import-thread failure path too, keyed off nosync_set.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
```diff
@@ -781,6 +781,24 @@ dbmdb_import_all_done(ImportJob *job, int ret)
...
             index = index->next;
         }
+        /* Re-enable fsync before going back online */
+        {
+            struct ldbminfo *li = (struct ldbminfo *)inst->inst_be->be_database->plg_private;
+            dbmdb_ctx_t *ctx = MDB_CONFIG(li);
+            if (ctx->env) {
+                unsigned int env_flags = 0;
+                mdb_env_get_flags(ctx->env, &env_flags);
+                if (env_flags & MDB_NOSYNC) {
```
droideck: Do I understand correctly that it clears NOSYNC whenever it sees it?
So even if `nsslapd-db-durable-transactions=off` (which sets NOSYNC). So finishing any online import quietly turns fsync back...

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import.c
droideck, line 1542: I wonder if we should log something (probably an INFO, but I don't have a strong opinion) on engage and on clear. Also, in this case, it'll be nice to have a test that asserts the change. I think it's a valuable info for the administrator.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_instance.c
```diff
@@ -717,6 +717,12 @@ int dbmdb_make_env(dbmdb_ctx_t *ctx, int readOnly, mdb_mode_t mode)
     if (readOnly) {
         flags = MDB_RDONLY;
     }
+    if (!ctx->dsecfg.durable_transactions) {
```
droideck: IIUC, this will be run on any call where `dbmdb_make_env` are involved (dbscan-style things). So it will rebuild `dsecfg` from the info file, whose table has no durable_transactions entry. 
I think we should apply this only when the config was actually loaded (dsecfg.dseloaded), or persist the setting into the info file alongside MAXSIZE and friends.

droideck (changes requested):

droideck (approved):
LGTM!

PR #7125, opened 2025-12-02
Issue 7124 - BDB cursor race condition with transaction isolation

vashirov:
Bug Description:
ASAN reported crashes in `__db_ditem_nolog()` with negative-size-param errors. Cursor operations without transaction isolation allowed concurrent page modifications to corrupt cursor state, leading to invalid memory access.

The race condition occurs when:
1. T1 opens a cursor without transaction protection
2. T2 modifies the same index page
3. T1 cursor operates on stale page metadata
4. `__db_ditem_nolog()` calculates negative size for `memmove()`
5. Crash: `AddressSanitizer: negative-size-param: (size=-8)`

Reproducer: dirsrvtests/tests/stress/backend/bdb_cursor_race_test.py Crash under ASAN usually happens within 10-30 minutes, but sometimes it can run for hours without any crash.

Fix Description:
Implement transaction isolation for cursors in `idl_new_fetch()` and
`idl_new_range_fetch()` by always calling `dblayer_read_txn_begin()`.

In `bdb_txn_begin()` verify if the environment supports transactions
(has DB_INIT_TXN flag) before attempting to begin a transaction.
This prevents errors during offline import which uses a private
environment without transaction support.

In `bdb_public_new_cursor()` skip transaction usage when the database's
environment doesn't support transactions.

Fixes: https://github.com/389ds/389-ds-base/issues/7124

## Summary by Sourcery

Ensure BDB cursor operations run under transaction isolation to prevent race conditions and crashes, and add a stress test that exercises high-contention search/modify/delete workloads.

Bug Fixes:
- Guard BDB cursor-based index fetch operations with a dedicated or caller-provided read transaction to avoid cursor state corruption and negative-size memory operations.

Enhancements:
- Adjust transaction-begin logic to respect explicitly zeroed transaction structures so callers can force isolated, non-nested transactions for cursor operations.

Tests:
- Introduce a long-running stress test that drives concurrent searches, modifications, deletions, and add/delete churn to reproduce and guard against BDB cursor race conditions.

Issue #7124, opened 2025-12-02
BDB cursor race condition


vashirov:
**Issue Description**
ASAN reported crashes in `__db_ditem_nolog()` with negative-size-param errors:

```
=================================================================
==109094==ERROR: AddressSanitizer: negative-size-param: (size=-8)
    #0 0x7fadfdce4329 in memmove (/usr/lib64/libasan.so.8+0xe4329) (BuildId: 0adabddcb77130fc2ea3840d060eb4e8a9ae0c85)
    #1 0x7badf6dcb1e8 in __db_ditem_nolog (/lib64/libdb-5.3.so+0xe51e8) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #2 0x7badf6dcc3db in __db_ditem (/lib64/libdb-5.3.so+0xe63db) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #3 0x7badf6d020f7 in __bamc_physdel.lto_priv.0 (/lib64/libdb-5.3.so+0x1c0f7) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #4 0x7badf6cfc654 in __bamc_close (/lib64/libdb-5.3.so+0x16654) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #5 0x7badf6dc063c in __dbc_close (/lib64/libdb-5.3.so+0xda63c) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #6 0x7badf6dd6c7e in __dbc_close_pp (/lib64/libdb-5.3.so+0xf0c7e) (BuildId: 8d20c7e9865bc743660a55af97fad4f11e754da6)
    #7 0x7badf6f69979 in bdb_public_cursor_op ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c:6883
    #8 0x7badf6ec70b1 in dblayer_cursor_op (/usr/lib64/dirsrv/plugins/libback-ldbm.so+0xf0b1) (BuildId: d6d3dd3b3eb64308d950b7f200a3e0c0af52c064)
    #9 0x7badf6ed59c4 in idl_new_fetch ldap/servers/slapd/back-ldbm/idl_new.c:316
    #10 0x7badf6ed8955 in index_read_ext_allids ldap/servers/slapd/back-ldbm/index.c:1074
[...]

PR #7671, opened 2026-07-27
Issue 7670 - BDB range searches intermittently fail with err=1 under write load

### LINKED ISSUE #7670: BDB range searches intermittently fail with err=1 under write load
**Description:**
On a busy server, searches that use a range filter on an indexed attribute intermittently fail with err=1 (Operations error) while the server is processing concurrent writes. The most common case is SSSD's "(entryUSN>=N)" refresh search. Every occurrence logs:

    ERR - idl_new_range_fetch - Failed to build range candidate list on entryusn index. Error is -12795
    ERR - build_candidate_list - Database error -12795

On write-heavy deployments these messages flood the errors log.

The impact is most visible on IPA. SSSD picks up new and changed entries through exactly these searches, so newly created users, hosts, groups and policies are slow to reach clients, while existing entries keep working from client caches. The failures are transient: the same search succeeds when retried. There is no data corruption and reindexing does not help.

BDB backend only.

**Steps to Reproduce:**
1. BDB backend with the USN plugin enabled (IPA default).
2. Run a continuous modify load while several clients issue "(entryUSN>=N)" searches.
3. Watch the search results and the errors log.

**Expected results:**
Searches succeed. Internal lock contention between searches and writes is handled by the server without failing client operations or flooding the errors log.

**Actual results:**
Individual searches fail with err=1 and the two ERR lines above are logged for every occurrence.

droideck:
Description: Since #7125 index cursor walks run inside DB_TXN_NOWAIT transactions, so under concurrent write load a range scan such as SSSD's "(entryUSN>=N)" fails immediately with DB_LOCK_DEADLOCK, surfaced as DBI_RC_RETRY (-12795). The equality lookup path retries since #7126, but the range path did not: every conflict failed the whole search with err=1 and flooded the errors log.

Retry the range fetch with the same bounded exponential backoff as the equality path, skipping the retry in both paths when the fetch runs inside a caller transaction that must be retried as a whole. Log transient attempts at debug level and emit a single ERR line only when the retry budget is exhausted and the search really fails. Also document why serializable cursor isolation must not be weakened, fix a NULL dereference on the idl_lmdb_range_fetch error path, add the missing backoff to the old-idl and ldbm_back_seq loops, and add a stress suite reproducing the entryusn workload.

Fixes: https://github.com/389ds/389-ds-base/issues/7670
Relates: #7124 #7126 #7462

Reviewed by: ?

## Summary by Sourcery

Handle transient BDB lock conflicts during indexed range reads by adding bounded retry-with-backoff logic and enhance logging and tests to validate range search robustness under write-heavy load.

Bug Fixes:
- Ensure BDB index equality and range reads under NOWAIT transactions treat DBI_RC_RETRY as a transient error with bounded exponential backoff instead of failing searches immediately.
- Prevent NULL dereferences and spurious errors in LMDB range fetch error paths when cursor operations fail with transient conflicts.

Enhancements:
- Factor the DBI_RC_RETRY backoff policy into a reusable helper and apply it consistently across index read, range read, and sequential cursor code paths.
- Clarify documentation around the need for serializable cursor isolation in new IDL range walks to avoid BDB cursor crashes.
- Improve error logging so transient range read conflicts are logged at debug level and hard failures emit a single clear ERR message when retry budgets are exhausted.

Tests:
- Add a slow stress test suite that simulates SSSD-style entryUSN range searches under heavy write load to verify that transient BDB lock conflicts do not cause range search failures and to measure collision rates.
- Extend pytest markers with a 'slow' marker to categorize long-running stress or soak tests.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/index.c
+++ b/ldap/servers/slapd/back-ldbm/index.c
@@ -1098,7 +1098,6 @@ index_read_ext_allids(
     }
     for (retry_count = 0; retry_count < IDL_FETCH_RETRY_COUNT; retry_count++) {
         *err = NEW_IDL_DEFAULT;
-        PRIntervalTime interval;
         idl_free(&idl);
         idl = idl_fetch_ext(be, db, &key, db_txn, ai, err, allidslimit);
         if (*err == DBI_RC_RETRY) {
@@ -1106,23 +1105,13 @@ index_read_ext_allids(
             slapi_log_err(SLAPI_LOG_BACKLDBM, "index_read_ext_allids",
                           "DBI_RC_RETRY on retry %d/%d for %s\n",
                           retry_count + 1, IDL_FETCH_RETRY_COUNT, basetype);
-#ifdef FIX_TXN_DEADLOCKS
-#error can only retry here if txn == NULL - otherwise, have to abort and retry txn
-#endif
-            /* Exponential backoff with jitter, capped at 200ms */
-            {
-                PRUint32 backoff_ms = 10;
-                int i;
-                for (i = 0; i < retry_count && backoff_ms < 200; i++) {
-                    backoff_ms *= 2;
-                }
-                if (backoff_ms > 200) {
-                    backoff_ms = 200;
-                }
-                backoff_ms += slapi_rand() % (backoff_ms + 1);
-                interval = PR_MillisecondsToInterval(backoff_ms);
+            if (NULL != db_txn) {
+                /* Only the caller can retry its own transaction */
+                break;
             }
-            DS_Sleep(interval);
+            if (retry_count + 1 < IDL_FETCH_RETRY_COUNT) {
+                ldbm_fetch_retry_sleep(retry_count);
+            }
             continue;
         } else if (*err == DBI_RC_NOTFOUND) {
             /* Key not found in index - this is normal, not an error */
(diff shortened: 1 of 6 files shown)


dirsrvtests/pytest.ini
```diff
@@ -4,3 +4,4 @@ markers =
     tier1: mark a test as part of tier1
     tier2: mark a test as part of tier2
     tier3: mark a test as part of tier3
+    slow: mark a test as a long-running stress or soak test
```
vashirov: This is what tier3 tests are for

vashirov (approved):
LGTM

progier389 (approved):
LGTM

ldap/servers/slapd/back-ldbm/index.c
```diff
@@ -1694,20 +1688,53 @@ index_range_read_ext(
...
+            if (db_txn != NULL) {
+                /* Only the caller can retry its own transaction */
+                break;
+            }
+            ldbm_nasty("index_range_read_ext", "Retrying range fetch", 1091, *err);
+            slapi_log_err(SLAPI_LOG_BACKLDBM, "index_range_read_ext",
+                          "DBI_RC_RETRY on range fetch retry %d/%d for %s\n",
+                          retry_count + 1, IDL_FETCH_RETRY_COUNT, type);
+            if (retry_count + 1 < IDL_FETCH_RETRY_COUNT) {
+                ldbm_fetch_retry_sleep(retry_count);
```
tbordaz, line 1730: I am a bit concern with this exponential sleep on a range search.
IIUC before that fix, the range search failed immediately and it is the RC of the bug.
Now it will retry with a sleep.
A concern is that the txn being opened, if the READER thread acquired multiple pages (read) before hitting that retry it will potentially make WRITE thread sleep as well if the  writer needs a page that the READER already acquired. In short I wonder if that sleep, in read path, may slow down a writer thread.

tbordaz (commented):
It is a very nice fix/test. Overall looks good to me, I have a minor question.

ldap/servers/slapd/back-ldbm/index.c
progier389: Unless mistaken there is no active txn associated with the current thread when calling ldbm_fetch_retry_sleep 
 (txn has been aborted because of the retry error and new one is not yet restarted)
 So the sleep will delay the current thread but not the other ones ...

tbordaz: That is correct. The txn is open/abort/commit at lower level (idl_new_range_fetch).
Forget my comment there is no risk of sleep while holding a page under a transaction

tbordaz (approved):

Issue #7673, opened 2026-07-29
Avoid using hardcoded IV for attribute encryption


progier389:
**Issue Description**
A flaw was found in 389-ds-base where the LDBM backend attribute encryption uses a hardcoded static initialization vector for AES-CBC and 3DES-CBC operations, allowing an attacker with privileged filesystem access to detect plaintext equality across encrypted entries by comparing ciphertext blocks.

**Package Version and Platform:**
 - Platform: Fedora
 - Package and version: all versions 

**Steps to Reproduce**
1.  pytest testcase test_attr_encryption_compare_values

**Expected results**
Different encrypted attributes should have different encrypted values
even if the clear value is the same.

**Additional context**
https://access.redhat.com/security/cve/cve-2026-14969

PR #7674, opened 2026-07-29
Issue 7673 - Avoid using hardcoded IV for attribute encryption

progier389:
When adding a encrypted attribute config entry, a random Initialization Vector is added
(dsInitializationVector attribute) in the entry and its associated attrinfo
Note: To be able to decrypt attributes encoded before this fix, legacy  config entry are kept without dsInitializationVector
Later on to crypt/decrypt attribute, if attrinfo has an iv, it is used otherwise the old hardcoded value is used.

Note: changelog is still encrypted with the hardcoded IV (because there is NULL attrinfo for changelog encryption)
but anyway all changelog records are different so there is no way to compare them

Issue: #7673

Reviewed by: ?

## Summary by Sourcery

Generate per-attribute initialization vectors for encrypted attributes while preserving compatibility with legacy configurations and ensure encryption only runs when a cipher is configured.

Enhancements:
- Use a per-attribute initialization vector stored in configuration entries instead of relying on a single hardcoded IV, falling back to the legacy IV only for existing entries and changelog data.
- Guard attribute and index encryption/decryption paths so they only execute when an encryption cipher is actually configured, preventing unintended cryptographic operations.
- Ensure attrinfo structures properly manage and free initialization vector memory when attributes are deleted to avoid leaks.

Tests:
- Add an attribute encryption test that verifies different encrypted attributes with identical cleartext values yield distinct ciphertext while decrypting back to the original data.


ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
```diff
@@ -100,15 +111,78 @@ ldbm_instance_attrcrypt_enable(struct attrinfo *ai, int cipher)
...
+    if (bv) {
+        bv = slapi_ch_bvdup(bv);
+    } else {
+        if (is_in_dse(e)) {
+            /* The entry exists but is a legacy entry.
+             * ==> Keep it as is and use hardcoded IV.
+             */
+            return bv;
+        }
+        /* Generate random Initialization Vector */
```
Firstyear, line 166: This seems to just be reintroducing the same bug.

The IV needs to be unique *every time an encryption operation is called*.

However here it seems that you are making the IV once and storing it into dse.ldif, leaving us open to the same vulnerability but with a few more layers in between. 

It should be that every time you call `slapd_pk11_ParamFromIV` that a unique IV is set into the params. 

The challenge here is that you now need to *store* this IV into the entry data too. 

An alternative is that you use a confounder pattern. This is where you use a static or all 0 IV, and then the first block you encrypt is a randomised value. This way when you decrypt you skip the first block on read and can read the entry. But the confounder has the same effect as an IV since it then causes the subsequent blocks to be unique despite the same IV being present (see AES-CTS for reference). 

Either way, this PR doesn't resolve the issue I'm sorry.

Firstyear (changes requested):

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
progier389: > The IV needs to be unique every time an encryption operation is called.

Sorry but what you propose is just not doable with 389 ds architecture because the encrypted values are also used as index keys and the index database relies on the fact that a given value of a given attribute is always encrypted the same way (otherwise we would not be able to find the value in the index.)

What the proposed fix is doing is to ensure that different attributes have different encrypted values 
even if clear value is the same so that we cannot compare the values between different attributes (that is what the testcase do)
(There is no way we can prevent comparing indexed attribute of entries because the index provides the data directly)

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
```diff
@@ -100,15 +111,81 @@ ldbm_instance_attrcrypt_enable(struct attrinfo *ai, int cipher)
...
+    Slapi_Value *va[2];
+
+    slapi_entry_attr_find(e, IV_ATTRIBUTE_NAME, &attr);
+    slapi_attr_first_value(attr, &sval);
+    if (sval) {
+        bv = (struct berval*) slapi_value_get_berval(sval);
+    }
+
+    if (bv) {
+        bv = slapi_ch_bvdup(bv);
```
tbordaz, line 158: As it was generated by the server, we could check the bv->bv_len == 16 before using it.

tbordaz (approved):
LGTM

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
```diff
@@ -100,15 +111,81 @@ ldbm_instance_attrcrypt_enable(struct attrinfo *ai, int cipher)
...
+}
+
+static bool
+is_in_dse(Slapi_Entry *e)
+{
+    const Slapi_DN *sdn = slapi_entry_get_sdn_const(e);
+    int rc = slapi_search_internal_get_entry((Slapi_DN *)sdn, NULL, NULL, plugin_get_default_component_id());
+    if (rc != LDAP_SUCCESS && rc != LDAP_NO_SUCH_OBJECT) {
+        slapi_log_err(SLAPI_LOG_WARNING, "ldbm_instance_attrcrypt_init_iv",
+                      "Failed to determine whether the entry %s is in dse.ldif. Attribute may not be decrypted properly. Error: %d:%s",
```
jchapma, line 131: Missing \n

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
```diff
@@ -100,15 +111,81 @@ ldbm_instance_attrcrypt_enable(struct attrinfo *ai, int cipher)
...
+    slapi_attr_first_value(attr, &sval);
+    if (sval) {
+        bv = (struct berval*) slapi_value_get_berval(sval);
+    }
+
+    if (bv) {
+        bv = slapi_ch_bvdup(bv);
+    } else {
+        if (is_in_dse(e)) {
+            /* The entry exists and has not IV. So it is a legacy entry\
```
jchapma, line 161: typo: has no IV

jchapma (approved):
IIUC, deterministic encryption is required for indexed attributes.

Possibly a crazy idea, but would it make sense to differentiate based on index status in the future? Non-indexed encrypted attributes could use a unique IV per value, while indexed ones use per-attribute IV. Not for this PR, just a thought for later.

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
Firstyear: > Sorry but what you propose is just not doable with 389 ds architecture because the encrypted values are also used as index keys and the index database relies on the fact that a given value of a given attribute is always encrypted the same way (otherwise we would not be able to find the value in the index.)

Then you haven't solved the CVE / vulnerability. You have moved from one static IV, to another form of static IV. 

You will likely need to re-do the index to use a HMAC for the attribute lookup since you can store HMAC outputs safely as they do not need an IV and don't leak information about the value (since HMAC is keyed, but also doesn't have an IV). But the actual attribute storage itself needs to have a unique IV each time encryption is called or you need to introduce a confounder.

Firstyear:
> unique IV per value, while indexed ones use per-attribute IV. Not for this PR, just a thought for later.

Still would be vulnerable. IV's must not be reused. I have suggested an alternative using HMAC outputs that would be secure and avoid the problem.

ldap/servers/slapd/back-ldbm/ldbm_attrcrypt_config.c
tbordaz: The concern raised by @Firstyear looks valid. Now the PR  looks enough to fix the CVE because it limits the security concern to comparing values of a same attribute. In short it improves the security regarding this use case.

A same attribute value needing the same encrypted value (for index purpose) I am not sure how to address the raised concern. Possibly changing the way we index encrypted attributes.

Firstyear: Yes. Instead of encrypting the index key, you should use a HMAC-S256 instead. It still provides unique value per attribute, without needing you to change anything else, or have IV's or whatever.

progier389: @Firstyear,   @tbordaz discussed this issue with our prodsec architect. His answer was:

In my opinion, the current PR is sufficient to consider CVE-2026-14969 fixed. The reported vulnerability is about a single hardcoded IV shared globally across all attributes and entries. Pierre's fix replaces that with a random IV per encrypted attribute, which eliminates the global IV and prevents cross-attribute ciphertext comparison.

Firstyear's suggestion of using HMAC-based outputs is a valid hardening improvement that would further strengthen the design, but it's a larger architectural change. I'd recommend tracking that separately as a follow-up enhancement rather than blocking this CVE fix on it.

----
So I propose that we create a new issue about using HMAC-S256 and unblock the current PR so that we can go ahead.

Firstyear: Lets review the code.

The IV you are generating here is performed once  https://github.com/389ds/389-ds-base/pull/7674/changes#diff-ff73395b9b52d1e223d2ebb993add39fd812e2a3dcdad05757a6f0bdef09bb7fR185

This then sets a SINGLE IV value into IV_ATTRIBUTE_NAME.

This value is retrieved https://github.com/389ds/389-ds-base/pull/7674/changes#diff-ff73395b9b52d1e223d2ebb993add39fd812e2a3dcdad05757a6f0bdef09bb7fR154

Again, this indicates that it is a single, staticly encoded IV value.

This function is called here: https://github.com/389ds/389-ds-base/pull/7674/changes#diff-ff73395b9b52d1e223d2ebb993add39fd812e2a3dcdad05757a6f0bdef09bb7fR241

Which sets the IV into `ai->ai_attrcrypt->iv`. Again, it is still the single value that was set via IV_ATTRIBUTE_NAME. No diversification has occurred.

Finally the encryption occurs https://github.com/389ds/389-ds-base/pull/7674/changes#diff-8ada7dc22eb4255d3cf41bb032f315f17842038a6049818f77d122f12c579fc0R1623

This accesses  `ai->ai_attrcrypt->iv` and passes it directly into `slapd_pk11_ParamFromIV`

So your comment: "Pierre's fix replaces that with a random IV per encrypted attribute, which eliminates the global IV and prevents cross-attribute ciphertext comparison." is provably incorrect. You do not have a PER ATTRIBUTE IV. You have an IV that is just set in IV_ATTRIBUTE_NAME. This means you again, have a hardcoded IV being used during attribute encryption.

As a result, this *does not resolve the CVE* from any cryptographic viewpoint. The product architect is likely not a cryptographer. I suspect you and Thierry are not either. So I do not believe you can simply wave this away and say "oh it's too hard to do it properly, this fix is good enough".

It is *not*. It is not a fix. It is the same vulnerability, dressed up in a pretty facade. 

__ I will not accept this as a temporary fix __

To reiterate, you do have some options:

* Replace attribute encryption in the indexes with a HMAC-SHA256 instead which avoids the need for an IV.
* Change the on disk format of attribute indexes to allow serialisation of the IV with the encrypted values.
* Retain the current new "IV_ATTRIBUTE_NAME" code and rename it to "IV_KDF_SECRET" or similar. Then use kbkdf (aka nist sp100-108) here https://github.com/389ds/389-ds-base/pull/7674/changes#diff-8ada7dc22eb4255d3cf41bb032f315f17842038a6049818f77d122f12c579fc0R1626  using IV_KDF_SECRET as the secret value, the attribute *value* as the context, and the attribute name as label. This then allows you to derive a per-attribute IV, which improves things. 

And the reality is that: "Firstyear's suggestion of using HMAC-based outputs is a valid hardening improvement that would further strengthen the design, but it's a larger architectural change. I'd recommend tracking that separately as a follow-up enhancement rather than blocking this CVE fix on it." will never happen. I've been around too long to know that this as a temporary fix, will never be revisited to actually fix the problem properly in future.

tbordaz: @Firstyear, you are right I am not a cryptographer, so sorry in advance for this novice question

- HMAC-SHA256 requires a key. I guess it is what you named 'IV_KDF_SECRET'. Where is it stored ?

Firstyear: You are confusing two of my suggestions @tbordaz which I think is the problem here. 

* HMAC-SHA256 requires a key, and that key would be stored in nss as the current AES encryption key is.
* When using KBKDF (nist sp100-108) then you can keep this value in dse.ldif via an attribute, the same way the current PR uses IV_ATTRIBUTE_NAME to reference that value in dse.ldif. 

This is because these two primitives behave in different ways, and resolve the CVE by different means.

When you use HMAC-SHA256 you have two inputs. The secret and the value to HMAC. This means disclosure of the secret reveals the content of all values. This is the same as AES/encryption however the process is one-way instead of reversible.  In the 389-ds example, the secret would be from the nss pkcs11 store, and the value is the attribute-value that we wish to "encrypt" or in this case, HMAC.  This is important as given the same secret and value, HMAC-SHA256 will always give the same result, allowing you to then do equality comparisons and lookups in the DB.

When you are using KBKDF the goal is different. In this case you are still using AES-256-CBC to encrypt the values.  In this situation you would be using KBKDF to derive unique *IVs* for each attribute/backend which avoids the hardcoding issue you have. However you still need some variance in each deviation for each 389-ds instance. This is why KBKDF requires three inputs. A secret, a context, and a label. Given these three values, you will always get the same output. However a change to any of these values *changes* the output. Here I have chosen that the secret is stored in dse.ldif in the attribute IV_KDF_SECRET, and the context and label to be the attributes value and name. This causes two identical attribute/values to derive the same IVs, which then results in the correct deterministic output of the AES-256-CBC encrypted value. 

The important point is to use KBKDF as a method of deriving unique IVs. Critically IVs are not considered "secret" data. Because of this the secret value can be stored in dse.ldif because disclosure of that secret does not reveal the content of the AES encryption key (which is in nss pkcs11). Importantly though it *does* mean that the IV will become unique for each unique encryption operation performed which does resolve this vulnerability**. 


** - provided that label and context are changed correctly during each encryption operation.

progier389:
Updated the test case (that will be needed for whatever solution we implement) to check that index keys are différents among different attributes even if the value is the same.

Then moving the PR to draft until it is reworked

PR #7678, opened 2026-07-30
Issue 7652 - Fix acl_test after fixing anonymous bind with password

mirielka:
Description:
Fixing acl test after bind result with NULL dn and non NULL password behavior was changed. The test now also sets password to NULL in order to perform its actions as intended.

Relates: #7652
Author: Lenka Doudova
Reviewer: ???

## Summary by Sourcery

Tests:
- Update acl_test anonymous bind setup to explicitly clear the bind password so the test runs as intended under the new bind semantics.


droideck (approved):
LGTM!

PR #7687, opened 2026-08-05
Issue 7685 - BUG - Avoid ioblocktimeout

### LINKED ISSUE #7685: connection_read_operation waits 100 ms in PR_Poll before checking new_operation
## Bug description

389 Directory Server occasionally adds approximately 100 ms of latency to an LDAP operation when reading the next operation from an already established connection.

The delay occurs in connection_read_operation() when the socket read returns EAGAIN / WOULD_BLOCK.

The current code performs:

ret = PR_Poll(&pr_pd, 1, timeout);

with:

CONN_TURBO_TIMEOUT_INTERVAL = 100

before checking whether this is the first read of a new LDAP operation:

if (new_operation) {
    ret = CONN_TIMEDOUT;
    goto done;
}

GDB showed that the 100 ms polls always occurred with:

new_operation: 1
waits_done: 0
remaining_data: 0

Therefore, no partial LDAP PDU had been received. The worker was speculatively waiting for the beginning of a new operation.

### Environment

Initially reproduced with:

Operating system: Debian 13.6
389 Directory Server: 3.2.1
Build configuration:
  --with-openldap
  --enable-gcc-security
  --enable-epoll

Also observed with:

Operating system: Rocky Linux 10.2
Package: 389-ds-base-3.2.0-8.el10_2.x86_64

### Reproducer

A single LDAP connection is opened and bound. The client then performs 1,000 sequential indexed LDAP searches over the same connection.

The test is repeated multiple times.

Typical searches complete in less than 1 ms, but some operations take approximately 100–101 ms.

Example:

search124=100.819
search817=100.853
search403=101.011

### strace evidence

strace shows that the delay corresponds to a full timeout on the LDAP socket:

poll([{fd=64<TCP...>, events=POLLIN}], 1, 100)
    = 0 (Timeout) <0.100197>

### GDB backtrace

```text
#0 poll()
#1 libnspr4
#2 connection_read_operation()
   at ldap/servers/slapd/connection.c
#3 connection_threadmain()
#4 libnspr4
#5 libc
```

Variables captured immediately before PR_Poll():

new_operation: 1
waits_done: 0
remaining_data: 0

### Access log correlation

For affected operations:

wqtime ≈ 0
wtime  ≈ 0
optime ≈ 0.1005
etime  ≈ 0.1006

This indicates that the delay occurs while processing/reading the operation, not in the operation queue or database backend.

### Experimental fix

Moving the new_operation check before PR_Poll() eliminates the speculative wait:

if (new_operation) {
    ret = CONN_TIMEDOUT;
    goto done;
}

struct PRPollDesc pr_pd;
PRIntervalTime timeout =
    PR_MillisecondsToInterval(CONN_TURBO_TIMEOUT_INTERVAL);

Partially received PDUs continue using the existing PR_Poll() and nsslapd-ioblocktimeout logic.

### Test result after the change

Before the change, sporadic searches took approximately 100–104 ms.

After rebuilding 389 DS 3.2.1 with the change:

Total LDAP searches: 10,000
Searches >= 20 ms: 0

The server remained functional and normal search latency remained below approximately 1 ms.

### Expected behavior

When no bytes have been read for a new LDAP operation, the connection should be returned immediately to the connection event loop instead of making an operation worker wait speculatively for 100 ms.

### Additional checks already performed

The delay was not caused by:

- Nagle, because TCP_NODELAY was enabled.
- CPU cgroup throttling.
- Access logging.
- MDB processing or LDAP indexes.
- Multiple operation workers per connection.
- Turbo Mode.
- A partially received LDAP PDU.

I can provide the complete test script, GDB scripts, strace traces, access-log excerpts, and the experimental patch.

Firstyear:
Bug Description: In some cases, the fd will trigger on poll which causes it to move to the active list. This then leads to connection_read_operation attempting a PR_Poll and then assuming that the ioblock might be occuring even if no data was actually received.

Fix Description: Instead of calling a second PR_Poll with the ioblocktimeout immediately, if we are a new operation and there is no buffered data available then we immediately shortcut out to prevent the extra delay.

fixes: https://github.com/389ds/389-ds-base/issues/7685

Review by: ???

## Summary by Sourcery

Bug Fixes:
- Avoid spurious ioblock timeout handling by immediately timing out new operations that would otherwise block with no buffered data.

### DIFF
--- a/ldap/servers/slapd/connection.c
+++ b/ldap/servers/slapd/connection.c
@@ -1442,6 +1442,18 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
             syserr = PR_GetOSError();
             if (SLAPD_PR_WOULD_BLOCK_ERROR(err) ||
                 SLAPD_SYSTEM_WOULD_BLOCK_ERROR(syserr)) {
+
+                /*
+                 * We blocked, is this the first read in a PDU? We don't want to
+                 * wait here on a new operation, but we do if there is actually data coming
+                 * in and getting queued up.
+                 */
+                if (new_operation && !conn_buffered_data_avail_nolock(conn, &conn_closed)) {
+                    /* If so, we return */
+                    ret = CONN_TIMEDOUT;
+                    goto done;
+                }
+
                 struct PRPollDesc pr_pd;
                 PRIntervalTime timeout = PR_MillisecondsToInterval(CONN_TURBO_TIMEOUT_INTERVAL);
                 pr_pd.fd = (PRFileDesc *)conn->c_prfd;
@@ -1458,29 +1470,22 @@ connection_read_operation(Connection *conn, Operation *op, ber_tag_t *tag, int *
                         ret = CONN_SHUTDOWN;
                         goto done;
                     }
-                    /* We timed out, is this the first read in a PDU ? */
-                    if (new_operation) {
-                        /* If so, we return */
-                        ret = CONN_TIMEDOUT;
+                    /* Loop, unless we exceeded the ioblock timeout */
+                    if (waits_done > ioblocktimeout_waits) {
+                        slapi_log_err(SLAPI_LOG_CONNS, "connection_read_operation",
+                                      "ioblocktimeout expired on connection %" PRIu64 "\n", conn->c_connid);
+                        disconnect_server_nomutex(conn, conn->c_connid, -1,
+                                                  SLAPD_DISCONNECT_IO_TIMEOUT, 0);
+                        ret = CONN_DONE;
                         goto done;
                     } else {
-                        /* Otherwise we loop, unless we exceeded the ioblock timeout */
-                        if (waits_done > ioblocktimeout_waits) {
-                            slapi_log_err(SLAPI_LOG_CONNS, "connection_read_operation",
-                                          "ioblocktimeout expired on connection %" PRIu64 "\n", conn->c_connid);
-                            disconnect_server_nomutex(conn, conn->c_connid, -1,
-                                                      SLAPD_DISCONNECT_IO_TIMEOUT, 0);
-                            ret = CONN_DONE;
-                            goto done;
-                        } else {
 
-                            /* The turbo mode may cause threads starvation.
-                                  Do a yield here to reduce the starving.
-                            */
-                            PR_Sleep(PR_INTERVAL_NO_WAIT);
+                        /* The turbo mode may cause threads starvation.
+                              Do a yield here to reduce the starving.
+                        */
+                        PR_Sleep(PR_INTERVAL_NO_WAIT);
 
-                            continue;
-                        }
+                        continue;
                     }
                 }
                 if (-1 == ret) {


Firstyear:
@israel-villar can you test with this PR?

Firstyear:
Additionally can you test with -DENABLE_EPOLL?

israel-villar:
Hi @Firstyear,

Yes, I have tested PR #7687, both with and without `ENABLE_EPOLL`.

For reproducibility, these are the exact source revisions and test environment used.

## Test environment

- OS: Debian 13, x86_64
- 389 Directory Server built directly from source
- All four builds were tested on the same host
- All variants used the same LDAP database, schema, indexes and server configuration

## Source revisions

### Base / upstream master

- Commit: `00c4bb8ac6181b1240de2264f5fe91e75a4115d6`
- 389 DS version: `389-Directory/3.3.0.202608091713git00c4bb8ac`
- Both the EPOLL and non-EPOLL builds were produced from this same commit.

### PR #7687

- Commit: `b1f81c7de0489d8f983bea5c2753086b725da04b`
- 389 DS version: `389-Directory/3.3.0.202608091719gitb1f81c7de`
- Both the EPOLL and non-EPOLL builds were produced from this same PR commit.

## Test matrix

I tested the following 2x2 matrix:

| Variant | Source commit | PR #7687 | EPOLL | Result |
|---|---|---:|---:|---|
| Base / no EPOLL | `00c4bb8ac6181b1240de2264f5fe91e75a4115d6` | No | No | 25 latency spikes / 4000 searches, all around 250 ms |
| Base / EPOLL | `00c4bb8ac6181b1240de2264f5fe91e75a4115d6` | No | Yes | 26 latency spikes / 4000 searches, all around 100–105 ms |
| PR #7687 / no EPOLL | `b1f81c7de0489d8f983bea5c2753086b725da04b` | Yes | No | 16 latency spikes / 4000 searches, all around 250 ms |
| PR #7687 / EPOLL | `b1f81c7de0489d8f983bea5c2753086b725da04b` | Yes | Yes | **0 latency spikes / 4000 searches** |

The normal search latency was essentially the same in all four cases, around 0.65–0.70 ms median. The difference is in the latency tail.

## PR #7687 + ENABLE_EPOLL results

For the `PR #7687 + ENABLE_EPOLL` build:

- 4000 searches
- Median: 0.680 ms
- p99: 1.962 ms
- p99.9: 4.547 ms
- Maximum: 9.682 ms
- Searches >= 50 ms: **0**
- ~100 ms spikes: **0**
- ~250 ms spikes: **0**

I repeated this over four independent server starts, with 1000 searches per run, and got 0/1000 spikes in all four runs.

## Interpretation

This is particularly interesting because it matches the issue we had previously traced in `connection_read_operation()` in `connection.c`, where `PR_Poll()` could introduce a wait when a complete PDU had not yet arrived.

With the unmodified base commit:

- Without EPOLL we see the characteristic ~250 ms delays.
- With EPOLL enabled, the events are still present at approximately the same frequency, but their duration changes to ~100–105 ms.

With PR #7687 but without EPOLL, the ~250 ms events are still present.

With **PR #7687 + ENABLE_EPOLL**, they disappear completely in our tests.

So, based on these results, PR #7687 appears to solve the latency problem we were seeing when used together with EPOLL.

## Next validation

As soon as we can use the RPM build from `packit/389ds-389-ds-base-7687`, we plan to repeat the same tests on a RHEL-compatible operating system.

This will allow us to verify that the behaviour is also reproduced with the packaged RPM build, rather than only with our source builds on Debian 13.

Thank you for pointing me to the PR and specifically asking me to test it with `ENABLE_EPOLL`.

progier389 (approved):
LGTM

israel-villar:
Hi @Firstyear,

I tried to install the Packit build on our Rocky Linux 10.2 test system, but the COPR project currently has no EL10/EPEL10 build:

```text
Error: It wasn't possible to enable this project.
Repository 'epel-10-x86_64' does not exist in project 'packit/389ds-389-ds-base-7687'.
Available repositories: 'fedora-43-x86_64', 'fedora-44-x86_64', 'fedora-rawhide-x86_64'
```

The Rocky Linux system is currently running:

```text
Rocky Linux 10.2 (Red Quartz)
x86_64
389-ds-base-3.2.0-8.el10_2.x86_64
```

Would it be possible to enable an `epel-10-x86_64` build for this Packit/COPR project?

That would allow us to repeat the latency tests using the packaged PR #7687 RPM on a RHEL-compatible system, as planned.

israel-villar:
Hi @Firstyear,

I initially tried to install the Packit build directly on our Rocky Linux 10.2 test system, but the COPR project currently has no EL10/EPEL10 repository:

```text
Error: It wasn't possible to enable this project.
Repository 'epel-10-x86_64' does not exist in project 'packit/389ds-389-ds-base-7687'.
Available repositories: 'fedora-43-x86_64', 'fedora-44-x86_64', 'fedora-rawhide-x86_64'
```

Rather than installing a Fedora RPM on Rocky, I downloaded the Packit SRPM and rebuilt it natively on Rocky Linux 10.2.

## Test system

```text
Rocky Linux 10.2 (Red Quartz)
x86_64
kernel: 6.12.0-211.42.1.el10_2.x86_64
```

The system was originally running:

```text
389-ds-base-3.2.0-8.el10_2.x86_64
```

The Packit SRPM used for the rebuild was:

```text
389-ds-base-3.3.0.202608050449gitb1f81c7de-pr7687.fc43.src.rpm
```

The SRPM identifies the PR commit as:

```text
b1f81c7de
```

which matches PR #7687 commit:

```text
b1f81c7de0489d8f983bea5c2753086b725da04b
```

One EL10-specific rebuild detail was required: the Fedora-generated Packit SRPM did not contain the bundled Berkeley DB source required by the RHEL conditional build. I reused the corresponding source tarball from the official Rocky Linux 389-ds-base SRPM and rebuilt the packages natively for EL10.

I built two RPM variants from the same Packit source:

```text
PR #7687 / no EPOLL
389-ds-base-3.3.0.202608050449gitb1f81c7de-pr7687.el10.x86_64

PR #7687 / EPOLL
389-ds-base-3.3.0.202608050449gitb1f81c7de-pr7687.epoll.el10.x86_64
```

The only functional build difference between these two variants is `--enable-epoll`.

For the EPOLL build, the resulting `ns-slapd` contains the expected dynamic references:

```text
epoll_create1
epoll_ctl
epoll_wait
```

## Rocky Linux 10.2 latency results

Each variant was tested with four blocks of 1000 LDAP searches, for a total of 4000 searches.

### PR #7687 / no EPOLL

```text
Searches:     4000
Average:      1.203 ms
Maximum:      249.997 ms
>= 50 ms:     10
80-120 ms:    0
220-280 ms:   10
```

All 10 slow operations were tightly clustered around 250 ms:

```text
249.777
249.892
249.997
249.671
249.666
249.671
249.866
249.765
249.656
249.748 ms
```

### PR #7687 / EPOLL

```text
Searches:     4000
Average:      0.666 ms
Maximum:      5.879 ms
>= 50 ms:     0
80-120 ms:    0
220-280 ms:   0
```

So on Rocky Linux 10.2 the result is:

```text
PR #7687 / no EPOLL : 10 / 4000 ~250 ms spikes
PR #7687 / EPOLL    :  0 / 4000 ~250 ms spikes
```

This reproduces the same behavior we previously observed on Debian:

```text
Debian base / no EPOLL     : 25 / 4000
Debian base / EPOLL        : 26 / 4000
Debian PR #7687 / no EPOLL : 16 / 4000
Debian PR #7687 / EPOLL    :  0 / 4000

Rocky PR #7687 / no EPOLL  : 10 / 4000
Rocky PR #7687 / EPOLL     :  0 / 4000
```

So far, our results strongly suggest that **PR #7687 together with EPOLL is what eliminates the latency spikes**.

## Rocky Linux 10.2 systemd / SELinux compatibility note

This topic has been quite a challenge.There was one additional integration issue when running the rebuilt 3.3.0 packages on Rocky Linux 10.2.

With the systemd unit shipped by this 3.3.0 build, `ns-slapd` initially failed to start under SELinux enforcing mode.

The relevant unit hardening settings included:

```ini
PrivateDevices=true
ProtectHostname=true
ProtectClock=true
NoNewPrivileges=yes
```

SELinux reported a denied `nnp_transition` while systemd attempted the transition from `init_t` to `dirsrv_t`.

As a consequence, `ns-slapd` could not start correctly in its expected SELinux domain and was subsequently denied access to its configuration.

For the Rocky Linux validation we therefore used this instance-specific systemd compatibility override:

```ini
# /etc/systemd/system/dirsrv@{instance-name}.service.d/selinux-nnp-compat.conf

[Service]
PrivateDevices=no
ProtectHostname=no
ProtectClock=no
NoNewPrivileges=no
```

After applying this override, `ns-slapd` starts in the expected SELinux domain:

```text
system_u:system_r:dirsrv_t:s0
```

SELinux remained in `Enforcing` mode throughout the tests.

Importantly, this same override was used for **both Rocky Linux test variants**:

```text
PR #7687 / no EPOLL
PR #7687 / EPOLL
```

so both latency tests were performed under the same systemd execution environment.

This compatibility issue appears to be independent of PR #7687 and independent of the EPOLL latency behavior. It appears to result from running the newer 3.3.0 Fedora-oriented package/systemd configuration after rebuilding it for Rocky Linux 10.2.

## Additional EPOLL-specific SELinux finding

There was a second, separate SELinux issue that affects the EPOLL build.

With EPOLL enabled, `Frontend_config_init()` reads:

```text
/proc/sys/fs/epoll/max_user_watches
```

On Rocky Linux 10.2, SELinux initially denied this access for `ns-slapd` running in `dirsrv_t`.

The AVC was:

```text
scontext=system_u:system_r:dirsrv_t:s0
tcontext=system_u:object_r:sysctl_fs_t:s0
tclass=dir
denied { search }
```

The corresponding syscall failed with:

```text
openat(...) = -1 EACCES (Permission denied)
```

We instrumented the Packit source temporarily to identify the exact startup failure, and confirmed that `ns-slapd` was exiting here:

```c
if ((f = fopen("/proc/sys/fs/epoll/max_user_watches", "r")) == NULL) {
    slapi_log_err(SLAPI_LOG_ERR, "Frontend_config_init",
                  "Unable to open file \"/proc/sys/fs/epoll/max_user_watches\". errno=%d\n",
                  errno);
    exit(-1);
}
```

A minimal local SELinux policy allowing `dirsrv_t` read access to `sysctl_fs_t` was therefore added for the EPOLL validation.

After adding that policy, the **original, non-instrumented EPOLL RPM** starts normally with:

```text
SELinux: Enforcing
ns-slapd domain: system_u:system_r:dirsrv_t:s0
```

The final latency results above were obtained using the original EPOLL RPM, not the diagnostic build.

So there appear to be two separate EL10 integration issues:

1. A systemd/SELinux `NoNewPrivileges` / domain-transition issue affecting both rebuilt variants.
2. An additional SELinux permission requirement for reading `/proc/sys/fs/epoll/max_user_watches` when EPOLL is enabled.

Neither changes the comparison between the two Rocky latency tests, since both were run with the same systemd compatibility override and with SELinux still enforcing.

At this point we have reproduced the EPOLL result on both Debian and Rocky Linux 10.2.

PR #7692, opened 2026-08-07
Issue 7605 - Harden CI test ports against ephemeral allocation

### LINKED ISSUE #7605: CI - topology_i2 fails creating standalone2 because port 38902 is already in use
## Problem

`LMDB Test (replication/single_master_test.py)` failed during `topology_i2` setup while creating `standalone2`.

Evidence:
- Run https://github.com/389ds/389-ds-base/actions/runs/27999932263
- Job https://github.com/389ds/389-ds-base/actions/runs/27999932263/job/82884788185

Observed failing case:
- `dirsrvtests/tests/suites/replication/single_master_test.py::test_lastupdate_attr_before_init`

Representative logs include:
- `ERROR at setup of test_lastupdate_attr_before_init`
- `dirsrvtests/lib/test389/topologies.py:458: in topology_i2`
- `Command '['systemctl', 'start', 'dirsrv@standalone2']' returned non-zero exit status 1`
- `PR_Bind() on All Interfaces port 38902 failed: Netscape Portable Runtime error -5982 (Local Network address is in use.)`

`topology_i2` creates two standalone instances via `create_topology({ReplicaRole.STANDALONE: 2}, request=request)`.

The topology helper generates standalone ports from `38900 + inst_num`, so `standalone2` is expected to use port `38902`.

The log proves that something was already bound to `38902`. It does not identify the process, so the root cause is still uncertain.

One code-path concern: `create_topology()` registers cleanup only after `_create_instances()` returns. If instance creation fails partway through, partially created instances may not get the normal topology finalizer.

## Proposed work

Investigate why `38902` was occupied when `topology_i2` attempted to create `standalone2`.

At minimum, improve failure diagnostics so this is actionable next time:
- log whether `dirsrv@standalone2` already exists before creation
- log what process owns `38902` when instance startup fails
- check whether partially created topology instances are cleaned up when `_create_instances()` fails

## Acceptance criteria

- `topology_i2` setup no longer fails in CI because `standalone2` cannot bind to `38902`, or
- if it fails again, the CI log identifies what owned `38902` and whether a partial topology cleanup path was involved.

droideck:
Bug Description: Test jobs sometimes fail at topology setup because ns-slapd can't bind one of the fixed lib389 test ports (38902, 39002, 39004, 39202 seen): PR_Bind returns EADDRINUSE and the whole module errors out. Those ports are inside the kernel's default ephemeral range (32768-60999), while the secure ports at 636xx sit outside it and have never failed, so an outgoing connection briefly grabbing a test port is the likely cause.

Fix Description: Pass --sysctl net.ipv4.ip_local_reserved_ports=38900-39399 to the test container, covering all four lib389 role ranges. Reserved ports are skipped when the kernel picks a source port but can still be bound explicitly, so the listeners are unaffected.

Fixes: https://github.com/389ds/389-ds-base/issues/7605

Reviewed by: ?

## Summary by Sourcery

CI:
- Update lmdbpytest and pytest GitHub workflows to set net.ipv4.ip_local_reserved_ports=38900-39399 on test containers, avoiding kernel ephemeral allocation of lib389 test ports.

### DIFF
--- a/.github/workflows/lmdbpytest.yml
+++ b/.github/workflows/lmdbpytest.yml
@@ -126,7 +126,7 @@ jobs:
           echo "Pull failed, retrying in $((i * 15))s..."
           sleep $((i * 15))
         done
-        CID=$(sudo docker run -d -h server.example.com --ulimit core=-1 --cap-add=SYS_PTRACE --privileged --rm --shm-size=4gb -v ${PWD}:/workspace quay.io/389ds/ci-images:test)
+        CID=$(sudo docker run -d -h server.example.com --ulimit core=-1 --cap-add=SYS_PTRACE --privileged --rm --shm-size=4gb --sysctl net.ipv4.ip_local_reserved_ports=38900-39399 -v ${PWD}:/workspace quay.io/389ds/ci-images:test)
         until sudo docker exec $CID sh -c "systemctl is-system-running"
         do
           echo "Waiting for container to be ready..."
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -126,7 +126,7 @@ jobs:
           echo "Pull failed, retrying in $((i * 15))s..."
           sleep $((i * 15))
         done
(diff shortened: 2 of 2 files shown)


progier389 (approved):
LGTM

Issue #7698, opened 2026-08-10
LMDB bulk import can silently drop entries


droideck:
**Issue Description**
An online total init into an LMDB backend can finish successfully while the consumer ends up with fewer entries than the supplier sent. Both servers report success and nothing is logged on either side; the consumer is simply short.

Only tombstone entries whose parent still exists can be affected (live user entries are always imported), but the loss is not bounded to one entry and leaves the replicas silently diverged in a way ds-replcheck can count but not explain.

**Package Version and Platform:**
Platform: LMDB backend
Package and version: main

**What failed**
- A replication regression test doing an online total init between two suppliers failed intermittently on LMDB with an entry count mismatch (assert 166 == 165): https://github.com/389ds/389-ds-base/actions/runs/30322914484/job/90163746504
  The consumer accepted every entry off the wire (each returned success), the import summary reported one entry fewer than was queued with zero skipped, and there was no warning, error, or core file.
- A targeted regression test confirms the failure principle: when a total update delivers a tombstone that gets processed before its still-existing parent, the tombstone is silently dropped. The import completes, reports success to the supplier, and the entry is missing from the consumer.

**Steps to Reproduce**
1. Set up two suppliers on the LMDB backend.
2. Populate a suffix that contains tombstones whose parent entries still exist.
3. Run an online total init from supplier1 to supplier2.
4. Compare entry counts on both sides. The consumer can come up short while both servers report success. (Timing-dependent over a real agreement.)

**Expected results**
Either every entry that was sent gets imported, or the import fails and says what it dropped.

droideck:
ce52c159a..dbb69001b 389-ds-base-3.2 -> 389-ds-base-3.2
9df98e758..7cadfa113 389-ds-base-3.1 -> 389-ds-base-3.1
2b914bd9f..0a0410fdf 389-ds-base-3.0 -> 389-ds-base-3.0
edd43cda6..9f8a2d4b8 389-ds-base-2.9 -> 389-ds-base-2.9
e2dd6260a..545a0f8a4 389-ds-base-2.8 -> 389-ds-base-2.8

PR #7699, opened 2026-08-11
Issue 7698 - Fix silent entry loss in LMDB bulk import waiter handling

droideck:
Bug Description: During an online total init the LMDB bulk producer parks an entry processed before its parent on waitingq. A tombstone waits on its parent's nsuniqueid, but the release loop only matched the parent's ndn, so such a tombstone was never released and was freed silently at producer exit: the init reported success while the entry was missing on the consumer.

Fix Description: Release waiters by the processed entry's nsuniqueid too, using the same guard and key format as the private-db registration in dbmdb_import_entry_info_by_param(), so a parked tombstone is recognized when its parent is processed.
On a clean producer exit with a non-empty waitingq, log a bounded sample of the stranded DNs with a total count and abort, so the loss is visible instead of silent.
Also fix a pre-existing leak of a released waiter's key buffer. The regression test drives the total update extended operations directly, which is the only way to control the nsuniqueid that a plain add discards.

Relates: https://github.com/389ds/389-ds-base/issues/7698

Reviewed by: ?

## Summary by Sourcery

Prevent silent LMDB bulk-import entry loss by matching waiters through parent unique IDs and surfacing unresolved entries as import failures.

Bug Fixes:
- Release LMDB bulk-import waiters when their parent is identified by nsuniqueid, preventing tombstones from being silently dropped.
- Abort bulk imports with stranded waiting entries and log a bounded sample with the total count instead of completing silently.

Enhancements:
- Clean up waiter key buffers when entries are released or requeued.

Tests:
- Add LMDB bulk-import regression coverage for tombstones released by a live parent and for reporting imports with missing parents.

### DIFF
--- a/dirsrvtests/tests/suites/import/bulk_import_test.py
+++ b/dirsrvtests/tests/suites/import/bulk_import_test.py
@@ -1,0 +1,234 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import os
+import time
+import ldap
+import pytest
+from ldap.extop import ExtendedRequest
+from lib389._constants import *
+from lib389.idm.organizationalunit import OrganizationalUnits
+from lib389.replica import Replicas, BootstrapReplicationManager
+from lib389.tombstone import Tombstones
+from lib389.utils import get_default_db_lib
+from test389.topologies import topology_st as topo
+
+pytestmark = [pytest.mark.tier1,
+              pytest.mark.skipif(get_default_db_lib() != "mdb",
+                                 reason="Exercises the LMDB bulk import consumer")]
+
+DEBUGGING = os.getenv("DEBUGGING", default=False)
+if DEBUGGING:
+    logging.getLogger(__name__).setLevel(logging.DEBUG)
+else:
+    logging.getLogger(__name__).setLevel(logging.INFO)
+log = logging.getLogger(__name__)
+
+START_OID = "2.16.840.1.113730.3.5.3"   # StartNSDS50ReplicationRequest
+ENTRY_OID = "2.16.840.1.113730.3.5.6"   # NSDS50ReplicationEntry
+END_OID = "2.16.840.1.113730.3.5.5"     # EndNSDS50ReplicationRequest
+TOTAL_PROTOCOL_OID = "2.16.840.1.113730.3.6.2"
+
+REPLICA_READY = 0x00
+RELEASE_SUCCEEDED = 0x09
+
+REPL_MANAGER_PW = 'replmanagerpw'
+FAKE_SUPPLIER_RID = 400
+
+SUFFIX_UUID = "aaaaaaaa-aaaaaaaa-aaaaaaaa-aaaaaaaa"
+ANCESTOR_UUID = "bbbbbbbb-bbbbbbbb-bbbbbbbb-bbbbbbbb"
+PARENT_UUID = "11111111-11111111-11111111-11111111"
+CHILD_UUID = "22222222-22222222-22222222-22222222"
+
+ANCESTOR_DN = f"ou=init-ancestor,{DEFAULT_SUFFIX}"
+PARENT_DN = f"ou=init-parent,{ANCESTOR_DN}"
+TOMBSTONE_DN = f"nsuniqueid={CHILD_UUID},uid=init-orphan,{PARENT_DN}"
+
+
+# Minimal DER encoders for the total update payloads.  The grammar mirrors
+# the consumer's decoders: decode_startrepl_extop()/decode_endrepl_extop()
+# in repl_extop.c and decode_total_update_extop()/my_ber_scanf_attr()/
+# my_ber_scanf_value() in repl5_total.c.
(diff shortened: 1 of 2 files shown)


dirsrvtests/tests/suites/import/bulk_import_test.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def _send_entry(conn, uniqueid, dn, attrs):
+    try:
+        conn.extop_s(ExtendedRequest(ENTRY_OID, _entry_payload(uniqueid, dn, attrs)))
+    except ldap.LDAPError as exc:
+        raise AssertionError(f"consumer rejected entry {dn}: {exc}") from exc
+
+
+def test_lmdb_bulk_import_tombstone_released_by_live_parent(topo):
```
tbordaz: Something not clear to me. You can reproduce the error while acting as a replication agreement and sending the appropriate entries/order that trigger the bug.
Can it be reproduced with a regular replication session ? For example creating a tombsone under a parent then move the parent under an entry E that is younger than than the parent.
I would expect the new in-depth first algo in the online total init supplier to play a role to send the tombstone only after the  'E' was sent.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
```diff
@@ -4257,6 +4277,32 @@ dbmdb_bulk_producer(void *param)
...
+            }
+            waiting_count++;
+        }
+        import_log_notice(job, SLAPI_LOG_ERR, "dbmdb_bulk_producer",
+                          "Aborting bulk import: %lu entries had no parent in the import stream; "
+                          "logged %lu and omitted %lu additional entries.",
+                          (long unsigned int)waiting_count,
+                          (long unsigned int)waiting_logged,
+                          (long unsigned int)(waiting_count - waiting_logged));
+        thread_abort(info);
```
progier389, line 4304: I wonder if we should abort or not:
  in online/offline import we just skip the entries without parent without aborting
That said for bulk import producer we may want to abort (because something is seriously amiss on the replica)

dirsrvtests/tests/suites/import/bulk_import_test.py
```diff
@@ -0,0 +1,234 @@
...
+    # the cause before the assertions.
+    for pattern in ('.*was never imported.*', '.*Aborting bulk import.*'):
+        for line in inst.ds_error_log.match(pattern):
+            log.info("error log: %s", line.strip())
+
+    assert OrganizationalUnits(inst, ANCESTOR_DN).get('init-parent') is not None
+    found = Tombstones(inst, DEFAULT_SUFFIX).filter(f'(nsUniqueId={CHILD_UUID})')
+    log.info("%d matching tombstones found after total update", len(found))
+    assert len(found) == 1
+
```
progier389, line 238: test_lmdb_bulk_import_tombstone_released_by_live_parent checks that tombstone 
 is added when entries are received out of order. ( This test the nsuniqueid part of the fix )

I think it would be nice to have a second test to check the log/abort part ( by sending a tombstone without parent )

dirsrvtests/tests/suites/import/bulk_import_test.py
progier389: As you said, the new in-depth first algo does prevent hitting the issue. But we still need to fix the issue because it will happen when initializing  from a replica with an older version (i.e the upgrade scenario)
So that is why the test should mimic a replication bulk import rather than relying on replication

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
droideck: Yeah, that's a trade-off I was thinking about too...
And I think I'm closer to your second pint. The ldif import a skip is visible in the "Processed X, skipped Y" summary. Bulk import has no such channel: the DONE path reports success to the supplier unconditionally, so a skip here means a replica that installs the supplier RUV and claims to be in sync while silently missing entries.

A total init stream is complete by construction, IIUC, so a waiter still parked at a clean end of stream means the parent never arrived. So supplier inconsistency or another key-pairing bug... And the producer already aborts on DNRC_DUP, DNRC_NOPARENT_ID, DNRC_BAD_TOMBSTONE; this closes the silent exit.

With the matching fix the abort should be unreachable in practice; so yeah, it's mostly a safety net.
Unless I miss something..

dirsrvtests/tests/suites/import/bulk_import_test.py
droideck: Added test_lmdb_bulk_import_aborts_on_stranded_tombstone.
It sends a tombstone whose nsparentuniqueid never appears in the stream, so it parks unconditionally, no ordering tricks. And then checks for the messages.

ldap/servers/slapd/back-ldbm/db-mdb/mdb_import_threads.c
tbordaz: During a bulk import if an entry is missing that means that a later replicated update targeting this entry will fail and break replication. The breakage can happen much after the bulk import and it will be difficult to diagnose. IMHO I would vote to abort the here.

progier389 (approved):
LGTM

progier389:
Note: That is a bit weird that the password test are failing both on LMDFB and BDB while it is pretty stable in the CI dashboard.
Anyway it seems unrelated to the PR (probably some test that was fixed after last PR branch rebase )

PR #7706, opened 2026-08-12
Issue 7705 - With memberOfEntryScope set, deferred memberOf skips MODIFY operations

### LINKED ISSUE #7705: With memberOfEntryScope set, deferred memberOf skips MODIFY operations
**Issue Description**
With `memberOfDeferredUpdate: on` and `memberOfEntryScope` configured, the deferred replay skips the memberOf fanout of every MODIFY that touches a grouping attribute. A plain MOD_ADD of a member never produces the member's memberOf value, on the origin server or on any replica, and nothing is logged.
Only a fixup task repairs the affected entries.

Cause: `deferred_mod_func` runs its entry scope check on a pre-op entry it never reads from the task pblock. The entry pointer stays NULL, so the check sees no DN and `slapi_sdn_issuffix()` matches no subtree. `memberOfEntryScope` is an allowlist: an entry must positively match one of the listed subtrees or `memberof_entry_in_scope()` returns out of scope, so with the scope set every target fails the check and the task bails with a success return code. The exclude setting is a denylist (no match keeps the entry in scope) and the default config has no scope at all, which is why only `memberOfEntryScope` configurations are affected and the bug stays silent everywhere else.

The enqueue side already duplicates the pre and post op entries into the task pblock, and the ADD, DELETE and MODRDN replay functions read their entries correctly; only the MODIFY replay is affected.

Possibly, a regression from the issue #7035 scoping refactor. FreeIPA always sets `memberOfEntryScope`, so any IPA deployment that enables `memberOfDeferredUpdate` loses all membership MODIFY updates.

**Package Version and Platform**
2.8.0 and later, including current main. BDB backend (deferred update is inert on LMDB).

**Steps to Reproduce**
1. Enable memberOf, set `memberOfDeferredUpdate: on` and `memberOfEntryScope` to the suffix, restart.
2. Create a group and a user, add the user as a member with a normal modify.
3. Read the user's memberOf.

**Expected results**
memberOf contains the group DN, the same as with deferred update off or with no entry scope configured.

**Actual results**
memberOf is never added. No error or warning is logged.

droideck:
Description: deferred_mod_func checks the entry scope against a pre-op entry it never reads from the task pblock. With memberOfEntryScope set the check always fails, so the memberOf fanout of every grouping attribute MODIFY is silently skipped, for direct and replicated operations alike. Possibly, a regression from the issue 7035 scoping refactor.

Read the post-op entry the way the direct modify path does, and add a regression test.

Fixes: https://github.com/389ds/389-ds-base/issues/7705

Reviewed by: ?

## Summary by Sourcery

Ensure deferred memberOf updates respect configured entry scope by using the post-operation entry for scope evaluation and add regression coverage for scoped and excluded subtree behavior.

Bug Fixes:
- Fix deferred memberOf fanout being skipped when memberOfEntryScope is set due to incorrect use of the pre-operation entry for scope checks.

Tests:
- Add memberOf plugin regression tests verifying deferred updates apply within memberOfEntryScope and are skipped for groups under memberOfEntryScopeExcludeSubtree.

### DIFF
--- a/dirsrvtests/tests/suites/memberof_plugin/memberof_deferred_scope_test.py
+++ b/dirsrvtests/tests/suites/memberof_plugin/memberof_deferred_scope_test.py
@@ -1,0 +1,129 @@
+# --- BEGIN COPYRIGHT BLOCK ---
+# Copyright (C) 2026 Red Hat, Inc.
+# All rights reserved.
+#
+# License: GPL (version 3 or any later version).
+# See LICENSE for details.
+# --- END COPYRIGHT BLOCK ---
+#
+import logging
+import pytest
+import os
+import time
+from lib389._constants import DEFAULT_SUFFIX
+from test389.topologies import topology_st as topo
+from lib389.plugins import MemberOfPlugin
+from lib389.idm.user import UserAccounts
+from lib389.idm.group import Groups
+from lib389.idm.nscontainer import nsContainers
+from lib389.utils import get_default_db_lib
+
+log = logging.getLogger(__name__)
+
+EXCLUDED_SUBTREE = f'cn=excluded,{DEFAULT_SUFFIX}'
+
+
+def configure_memberof(inst, scope=None, exclude=None):
+    memberof = MemberOfPlugin(inst)
+    memberof.enable()
+    memberof.set_autoaddoc('nsMemberOf')
+    memberof.set_memberofdeferredupdate('on')
+    memberof.remove_all('memberOfEntryScope')
+    memberof.remove_all('memberOfEntryScopeExcludeSubtree')
+    if scope:
+        memberof.set('memberOfEntryScope', scope)
+    if exclude:
+        memberof.set('memberOfEntryScopeExcludeSubtree', exclude)
+    inst.restart()
+
+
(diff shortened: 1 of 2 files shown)


tbordaz (approved):
Nice finding !!
LGTM

tbordaz:
I would just change the summary with something like "with memberOfEntryScope set, Deferred memberOf skip MODIFY operations"

droideck:
dbb69001b..4b41542b2 389-ds-base-3.2 -> 389-ds-base-3.2
7cadfa113..0827770b8 389-ds-base-3.1 -> 389-ds-base-3.1
0a0410fdf..e2ce43b64 389-ds-base-3.0 -> 389-ds-base-3.0
9f8a2d4b8..ffcd77c6b 389-ds-base-2.9 -> 389-ds-base-2.9
545a0f8a4..ea83db3d4 389-ds-base-2.8 -> 389-ds-base-2.8
85f569ae8..a7dcb653c 389-ds-base-2.7 -> 389-ds-base-2.7
4395554c0..01edf5c3a 389-ds-base-2.6 -> 389-ds-base-2.6

Issue #7710, opened 2026-08-12
MemberOf deferred update - Use condvar instead of sleep loop


mreynolds389:
**Issue Description**

Currently we use a while/sleep loop to check if a deferred update is complete.  Better to use a condition variable instead.

mreynolds389:
ede8e0bce..e63592d5e  389-ds-base-3.2 -> 389-ds-base-3.2
fdff9eb39..3fc51b255  389-ds-base-3.1 -> 389-ds-base-3.1
969796da8..2f58e9355  389-ds-base-3.0 -> 389-ds-base-3.0
08abe0bb3..cdc57272d  389-ds-base-2.9 -> 389-ds-base-2.9
e8c9a7095..a23a87926  389-ds-base-2.8 -> 389-ds-base-2.8
3e9164312..bd9bf109f  389-ds-base-2.7 -> 389-ds-base-2.7
b6b3c7744..b160731bd  389-ds-base-2.6 -> 389-ds-base-2.6
bad040e16..40ed5a73a  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7712, opened 2026-08-12
Issue 7710 - MemberOf deferred update - Use condvar instead of sleep loop

mreynolds389:
Description:

Currently we use a while/sleep loop to check if a deferred update is complete. Better to use a condition variable instead.

Deferred updates also needs to skip special backends:

- cn=config
- cn=schema
- cn=changelog

relates: https://github.com/389ds/389-ds-base/issues/7710

Assisted-by: Cursor

## Summary by Sourcery

Introduce per-operation synchronization for deferred memberOf updates using a pblock-based condition variable and replace backend sleep polling with this wait mechanism, while excluding special containers and RUV updates from deferred processing.

New Features:
- Add a pblock-level API to wait for completion of deferred memberOf work using a condition variable.
- Skip deferred memberOf processing for cn=config, cn=schema, cn=changelog, and replication RUV operations.

Enhancements:
- Replace backend while/sleep polling for deferred memberOf completion with a condition-variable-based wait tied to the operation pblock.
- Manage lazy initialization and teardown of deferred memberOf synchronization primitives within the pblock lifecycle.

### DIFF
--- a/ldap/servers/plugins/memberof/memberof.c
+++ b/ldap/servers/plugins/memberof/memberof.c
@@ -37,6 +37,7 @@
 #include <ctype.h>
 #include <time.h>
 #include <unistd.h>
+#include <stdbool.h>
 #include "slapi-plugin.h"
 #include "string.h"
 #include "nspr.h"
@@ -51,6 +52,7 @@ static void *_PluginID = NULL;
 static Slapi_DN *_ConfigAreaDN = NULL;
 static Slapi_RWLock *config_rwlock = NULL;
 static Slapi_DN* _pluginDN = NULL;
+static Slapi_DN** ignored_containers_sdn = NULL;
 static int usetxn = 0;
 static int premodfn = 0;
 static PRLock *fixup_lock = NULL;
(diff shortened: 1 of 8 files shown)


ldap/servers/plugins/memberof/memberof.c
```diff
@@ -948,8 +951,6 @@ deferred_add_func(MemberofDeferredAddTask *task)
     pb = task->pb;
 
     slapi_pblock_get(pb, SLAPI_TARGET_SDN, &sdn);
-    slapi_log_err(SLAPI_LOG_PLUGIN, MEMBEROF_PLUGIN_SUBSYSTEM,
-                  "deferred_mod_func: target %s\n", slapi_sdn_get_dn(sdn));
```
tbordaz, line 952: The name of the function is wrong (deferred_mod_func instead of deferred_add_func). Is it the reason why you remove this log ?

tbordaz (commented):
The patch looks good to me. Let's wait test completion.

ldap/servers/plugins/memberof/memberof.c
mreynolds389: >  Is it the reason why you remove this log ?

No reason to remove the logging - not sure how it happened haha.  I'll fix it after the tests finish.

tbordaz (approved):
LGTM. Thanks Mark
Minor comment regarding a log erroneously removed.

mreynolds389:
> LGTM. Thanks Mark Minor comment regarding a log erroneously removed.

 It was removed by mistake :-)  I added it back (and fixed the logging name issues)

Issue #7709, opened 2026-08-12
Add EPEL 10 target to Packit COPR builds


israel-villar:
**Is your feature request related to a problem? Please describe.**

During the validation of #7687 on Rocky Linux 10.2, we wanted to test the PR using RPM packages built for an Enterprise Linux 10 environment.

The Packit COPR project only provided builds for:

```text
fedora-43-x86_64
fedora-44-x86_64
fedora-rawhide-x86_64
```

There was no EPEL 10 / EL10 build available.

Because of this, we had to download the Packit SRPM and rebuild it manually on Rocky Linux 10.2 in order to complete the validation.

This worked, but it adds extra manual work and makes validation on RHEL-compatible systems less straightforward and less reproducible.

**Describe the solution you'd like**

It would be very useful if the Packit configuration also generated an EPEL 10 target for pull requests.

For example:

```yaml
targets:
  - fedora-all
  - epel-10
```

This would allow changes to be tested directly in an EL10-compatible environment, alongside the existing Fedora builds.

If generating an EPEL 10 build for every pull request would add too much CI overhead, a manually triggered EPEL 10 Packit/COPR build would also be a very useful option.

**Describe alternatives you've considered**

For #7687, we worked around the limitation by downloading the Fedora-generated Packit SRPM and rebuilding it manually on Rocky Linux 10.2.

That allowed us to complete the validation, but required additional EL10-specific packaging work.

In particular, the Fedora-generated SRPM did not include the bundled Berkeley DB source required by the RHEL conditional build, so we had to obtain the corresponding source tarball from the official Rocky Linux `389-ds-base` SRPM before rebuilding the packages.

This approach works, but having Packit generate the EL10 build directly would be much simpler and would also validate the EL10 packaging path itself.

**Additional context**

#7687 is a concrete example where an EL10 Packit build would have been useful.

We performed the final validation on Rocky Linux 10.2 with SELinux Enforcing and compared two native EL10 RPM builds of the PR, with and without EPOLL.

The results and the EL10-specific integration details are documented in #7687.

Adding an EPEL 10 Packit target would make similar validation much easier for future changes affecting RHEL-compatible systems.

mreynolds389:
fcc5ef898..1b1982c1c  389-ds-base-3.2 -> 389-ds-base-3.2

israel-villar:
[...]

PR #7716, opened 2026-08-12
Issue 7709 - Add EPEL 10 target to Packit COPR builds

mreynolds389:
Description:

Add epel-10 to Packit copr build targets

relates: https://github.com/389ds/389-ds-base/issues/7709

## Summary by Sourcery

Build:
- Include epel-10 in Packit COPR build targets for both pull request and main branch commit triggers.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -21,11 +21,12 @@ srpm_build_deps:
 actions:
   post-upstream-clone:
     - cargo install cargo-license
-    - make -f rpm.mk rpmspec
+    - make -f rpm.mk rpmroot rpmspec
   create-archive:
-    - make -f rpm.mk download-cargo-dependencies tarballs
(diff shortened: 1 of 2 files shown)


vashirov (approved):
LGTM

mreynolds389:
@vashirov with your change to always bundle libdb do we still need the changes to the packit yaml file?

vashirov:
With my change libdb tarball is always bundled during srpm creation. The issue was that depending on where (Fedora, RHEL) you create srpm, different sources were included. Packit/copr use Fedora, so it would omit the libdb tarball from srpm, but it would be required for compilation on RHEL.

`epel-10:` section is required, and `without_opts:` and `with_opts:` can be kept to make it more obvious what options are used, the actual decision to build with or without bundled libdb is handled elsewhere.

vashirov:
FWIW, it builds on EPEL also with these minimal changes: https://github.com/vashirov/389-ds-base/pull/17/changes

PR #7725, opened 2026-08-18
Issue 3082 - Add test389.topologies compatibility shim for backports

vashirov:
Bug Description:
Backported tests from main use 'from test389.topologies import ...' which fails on older branches where test389 doesn't exist.

Fix Description:
Add a shim that re-exports lib389.topologies as test389.topologies so backports work without per-test fixups.

Relates: https://github.com/389ds/389-ds-base/issues/3082

## Summary by Sourcery

Add compatibility support for backported topology tests by exposing `lib389.topologies` through the `test389.topologies` namespace.

Bug Fixes:
- Enable backported tests that import `test389.topologies` on branches where only `lib389.topologies` is available.

Enhancements:
- Add a compatibility re-export for topology helpers and make the test library path available during test configuration.

### DIFF
--- a/dirsrvtests/conftest.py
+++ b/dirsrvtests/conftest.py
@@ -12,6 +12,7 @@ import pytest
 import shutil
 import glob
 import ldap
+import sys
 import os
 import gzip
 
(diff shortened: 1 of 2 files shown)


vashirov:
Note: this is opened against 3.1 on purpose, the same change needs to go to other versioned branches.

mirielka (approved):
LGTM!

PR #7729, opened 2026-08-18
bdb: Fix nsslapd-db-compactdb-interval: 0 not disabling auto compaction

### LINKED ISSUE #1: pre compile and normalize search filter
Cloned from Pagure issue: https://pagure.io/389-ds-base/issue/1
- Created at 2011-12-17 07:47:22 by [rmeggins](https://pagure.io/user/rmeggins) (@richm)
- Closed as Fixed
- Assigned to [rmeggins](https://pagure.io/user/rmeggins) (@richm)
- Associated bugzillas
  - https://bugzilla.redhat.com/show_bug.cgi?id=772777

---

When processing large search filters which are applied to every entry in the search result set, the filter is normalized anew each time a new entry is tested.  For substring filters, a regular expression must be created, compiled, and freed each time the substring filter is tested, in addition to normalizing the values.  For example, if the search filter contains 1000 substring sub-filters, for each entry tested with the filter, this will require 1000 filter normalizations followed by 1000 regex creation, compilation, and cleanup.  If there are 1000 entries in the search result set, this will require a million such operations.

### LINKED ISSUE #7728: nsslapd-db-compactdb-interval: 0 does not disable automatic database compaction
## Issue Description

  Setting `nsslapd-db-compactdb-interval: 0` is documented to disable automatic
  database compaction, but it does not. Auto-compaction keeps running once a day
  at the configured `nsslapd-db-compactdb-time`, and `nsslapd-db-compactdb-starttime`
  keeps drifting on every server restart even though the feature has been disabled.

  **Environment:** 389-ds-base-2.8.0 (also reproduced on main and the 3.0/3.1/3.2
  release branches). No workaround other than setting the interval to an absurdly
  large nonzero value (e.g. 999999999) instead of 0.

  ## Root Cause

  In `ldap/servers/slapd/back-ldbm/db-bdb/bdb_layer.c`, `bdb_checkpoint_threadmain()`
  never checks for `interval == 0` before scheduling or executing compaction:

  1. **At startup**, if `nsslapd-db-compactdb-starttime` already exists from a prior
     run, the check `if (compactdb_interval < (curr_time - compactdb_start_time))`
     is always true when `interval == 0`, forcing the internal timer to expire
     almost immediately on every restart.
  2. **In the main loop**, the scheduling condition has no `interval == 0` branch,
     so once the bogus timer expires it unconditionally schedules `bdb_compact` via
     `slapi_eq_once_rel()` and rewrites `nsslapd-db-compactdb-starttime`.
  3. **Self-rearming**: `bdb_compact()` resets `compaction_scheduled = PR_FALSE`
     after every run, re-arming the same broken condition - the cycle repeats
     forever, once a day, regardless of `interval: 0`.

 There is even a code comment acknowledging this ("this could have been a bug in
  fact, where compactdb_interval was 0...") but the guard was never added.

  The existing regression test added in #4854 (for #4778) only sleeps ~3 seconds
  after setting `nsslapd-db-compactdb-time` to a value usually hours away, so it
  never exercises the actual scheduled-compaction path and doesn't catch this.

  A fix and additional regression tests will be submitted in a linked PR.

apeddire-wq:
## Description
  
  Fixes #7728
  
  `nsslapd-db-compactdb-interval: 0` is documented to disable automatic database
  compaction, but `bdb_checkpoint_threadmain()` / `bdb_compact()` never checked
  for `interval == 0`, so auto-compaction kept running once a day and
  `nsslapd-db-compactdb-starttime` kept drifting on every restart.
  
   ## Fix

  Four changes, all in `bdb_layer.c`:

  1. `bdb_compact()` re-checks the live config when its scheduled callback fires
     and no-ops if `interval == 0`. This closes the race where compaction was
     already queued via `slapi_eq_once_rel()` before the admin disabled it.
     (`bdb_compact()` has exactly one call site in the codebase; the manual
     on-demand `compact db` task path never calls it, so this has no effect on
     manual compaction.)
  2. Guard the pre-loop timer-adjustment math with `interval > 0` to avoid
     computing a negative `time_t`.
  3. Guard the in-loop timer-adjustment math the same way, for the case where
     the admin changes the interval while the server is running.
  4. **Primary fix** - the scheduling block is now gated on
     `compactdb_interval_update == 0`, skipping scheduling entirely while
     keeping `compactdb_interval_orig` in sync, so re-enabling compaction later
     takes effect immediately without a server restart.

  `bdb_write_compact_start_time()` itself is intentionally left unguarded, since
  it's shared with the manual compact-task path; fix #1 already prevents the
  automatic scheduler from calling it while disabled.

  
  ## Tests
  
  Two new tests in `dirsrvtests/tests/suites/config/compact_test.py`:
  
  - `test_no_compaction_near_scheduled_time` - sets interval=0 with
    `compactdb-time` ~90s in the future, confirms no scheduling/compaction log
    lines appear, and confirms `nsslapd-db-compactdb-starttime` is unchanged
    across a restart. 
  - `test_disable_compaction_after_already_scheduled` - schedules a compaction
    with a short nonzero interval, then disables it (interval=0) before it
    fires, and confirms it never actually runs.

  Verified with `python3 -m py_compile`.

## Summary by Sourcery

Ensure setting the automatic BDB compaction interval to zero reliably disables scheduled compaction without affecting manual compaction.

Bug Fixes:
- Prevent automatic BDB compaction when `nsslapd-db-compactdb-interval` is set to 0, including compactions that were already scheduled before the setting changed.
- Prevent the compaction start time from being rewritten while automatic compaction is disabled and allow compaction to resume promptly when re-enabled.

Tests:
- Add coverage for disabling compaction near its scheduled time and after a compaction event has already been queued.

### DIFF
--- a/dirsrvtests/tests/suites/config/compact_test.py
+++ b/dirsrvtests/tests/suites/config/compact_test.py
@@ -14,6 +14,7 @@ import datetime
 from lib389.utils import get_default_db_lib
 from lib389.tasks import DBCompactTask
 from lib389.backend import DatabaseConfig
+from lib389._mapped_object import DSLdapObject
 from test389.topologies import topology_m1 as topo
 from lib389.utils import ldap, ds_is_older
 from lib389.idm.user import UserAccounts
@@ -22,6 +23,8 @@ from lib389._constants import DEFAULT_SUFFIX
 
 pytestmark = pytest.mark.tier2
 log = logging.getLogger(__name__)
+
+BDB_CONFIG_DN = "cn=bdb,cn=config,cn=ldbm database,cn=plugins,cn=config"
 
 
 def test_compact_db_task(topo):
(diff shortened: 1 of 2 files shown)


apeddire-wq:
Addressed both points in 8918590f5:

  - Lowered the disabled-compaction log message in `bdb_compact()` from
    NOTICE to DEBUG, since it fires during expected admin behavior
    (intentionally disabling auto-compaction), not an error condition.

  - Replaced the long blind `time.sleep()` calls with small polling
    helpers that fail fast as soon as an unexpected log line appears,
    instead of waiting out the full duration before checking. Also
    shrank the `compactdb-time` offsets from 90s/60s down to 15s —
    the checkpoint thread actually re-evaluates compaction scheduling
    roughly every 2.5s (`DBLAYER_SLEEP_INTERVAL * 10` in bdb_layer.c),
    independent of any configured interval, so the original margins
    were far more conservative than necessary.

progier389 (approved):
LGTM

Issue #7735, opened 2026-08-19
Heap buffer overflow in schema objectclass superior


mreynolds389:
**Issue Description**

When adding/parsing an objectclass the superior or parent objectclass is not properly calculated into the buffer size and can lead to a buffer overflow if the supperior value is too large.

mreynolds389:
e9950ea56..85839dc0a  389-ds-base-3.2 -> 389-ds-base-3.2
1acfa3889..ab44ca468  389-ds-base-3.1 -> 389-ds-base-3.1
e486c8bee..d45b5d171  389-ds-base-3.0 -> 389-ds-base-3.0
01fd78656..2859a4ac2  389-ds-base-2.9 -> 389-ds-base-2.9
9e288e451..cbfac3488  389-ds-base-2.8 -> 389-ds-base-2.8
0afa2e38b..a36df48a0  389-ds-base-2.7 -> 389-ds-base-2.7
f4a653be0..1dd3687a6  389-ds-base-2.6 -> 389-ds-base-2.6
4c5c481be..1e5fb4817  389-ds-base-2.4 -> 389-ds-base-2.4
2d1421818..0b0f22a82  389-ds-base-2.2 -> 389-ds-base-2.2
9a0245b84..43b384286  389-ds-base-1.4.3 -> 389-ds-base-1.4.3

PR #7736, opened 2026-08-19
Issue 7735 - Heap overflow when parsing objectclass superior

mreynolds389:
Description:

When adding/parsing an objectclass the superior or parent objectclass is not properly calculated into the buffer size and can lead to a buffer overflow if the superior value is too large.

Relates: https://github.com/389ds/389-ds-base/issues/7735

Reviewed by: progier & spichugi(Thanks!!)

## Summary by Sourcery

Account for object class superior names when calculating schema serialization buffers to safely handle oversized definitions.

Bug Fixes:
- Prevent heap-buffer overflows when serializing object classes with large superior definitions.

Tests:
- Add a Tier 1 regression test covering object classes with oversized superior and attribute definitions.


mirielka (approved):
LGTM!

PR #7717, opened 2026-08-13
Issue 7595 - Skip redundant CI runs to relieve the Actions queue

droideck:
Description:
The org's free plan allows 20 concurrent Actions jobs, while each push/PR event schedules ~222 (111 suites x 2 backends). Full runs take 40+ hours; backport sweeps take days. Remove the runs whose result is already known:

- Drop push triggers from the test workflows: backports already pass full PR CI. A nightly dispatcher instead runs each release branch (X.Y and X.Y.Z) that changed in the last 30 days, once per head commit. (Push runs use the workflow on the pushed ref, so this needs backporting to active branches to fully take effect.)
- Skip main's nightly when HEAD already passed the previous one.
- Ignore doc-only paths (markdown, docs/agents, docs/design, LICENSE, .packit.yaml) in the test and CodeQL workflows. docs/slapi.doxy.in and the doxygen assets stay CI-visible because make all consumes them.
- Run only the webui suite for PRs that touch nothing outside src/cockpit/389-console; renamed files count both their old and new path.
- Fix the pytest_tests default: "false" made bare manual dispatches run an empty matrix. The dispatcher also passes explicit empty inputs so older branch workflows run in full.

Gates fail open, including on gate job failure, and there are no required status checks, so skipped workflows cannot block merges.

Relates: https://github.com/389ds/389-ds-base/issues/7595

Reviewed by: ?

## Summary by Sourcery

Reduce redundant CI workload while preserving coverage for relevant code changes, nightly validation, and recently updated release branches.

New Features:
- Add nightly dispatching for recently changed release branches and reduce pull-request test matrices to the Web UI suite when changes are isolated to the Cockpit console.

Bug Fixes:
- Correct manual test-dispatch defaults so unspecified test selections run the full suite, including on older branch workflows.

Enhancements:
- Eliminate redundant push-triggered test runs and skip unchanged nightly test runs while keeping gating fail-open.
- Exclude documentation- and metadata-only changes from test and CodeQL workflows while retaining build-relevant documentation inputs.

CI:
- Reduce GitHub Actions load by consolidating release-branch coverage into scheduled dispatches and avoiding duplicate CI execution.

PR #7753, opened 2026-08-21
Issue 7595 - Skip redundant CI runs to relieve the Actions queue

droideck:
Bug Description:
Every push to this branch schedules the full test matrix against the org's 20-job concurrency limit, so a backport sweep saturates the queue for days. The results are already known: each backport passed full PR CI before it landed. A manual workflow_dispatch on this branch is broken as well - pytest_tests defaults to the string "false", which generate_matrix.py rejects, so the run reports success having executed zero tests.

Fix Description:
Backport the main-branch change. Drop the push triggers from the test workflows; main's nightly dispatcher runs this branch once per head commit instead. Narrow the push triggers of the remaining workflows to main and the release branches, and ignore doc-only paths. Fix the pytest_tests default to "".

The gate jobs are inert here, since schedule only fires on the default branch, and they fail open, so they cannot block a run. Where the surrounding lines conflicted, the pick resolves to main's side, which carries earlier queue-relief work along: the job timeout caps from issue 7296, and, on branches that lacked them, the PR-cancel concurrency groups and read-only permissions from issue 6841. The runner pins and container images are unchanged.

codeql.yml and nightly-dispatch.yml are deliberately not backported. CodeQL's only release-branch trigger is pull_request, which would add jobs instead of removing them, and the dispatcher is a single-instance orchestrator that belongs on main.

Assisted by: Claude

(cherry picked from commit 4731986532fde46136f60dc1356ad77a67adb73c)

## Summary by Sourcery

Reduce CI queue pressure by avoiding redundant backport runs while retaining pull-request, release, and nightly validation.

Bug Fixes:
- Fix manual test workflow dispatches so an omitted test-suite selection runs the full test matrix.

Enhancements:
- Reduce redundant CI workload by limiting push-triggered workflows to main and release branches, ignoring documentation-only changes, and skipping unchanged scheduled test runs.
- Preserve reliable nightly execution with fail-open gates and add execution time limits to key build and test jobs.

### DIFF
--- a/.github/workflows/cargotest.yml
+++ b/.github/workflows/cargotest.yml
@@ -2,7 +2,25 @@ name: Rust Tests
 
 on:
   push:
+    branches:
+      - main
+      - "389-ds-base-*"
+    paths-ignore:
+      - '**.md'
+      # not docs/**: slapi.doxy.in and the doxygen assets are build inputs
+      - 'docs/agents/**'
+      - 'docs/design/**'
+      - 'LICENSE*'
+      - '.packit.yaml'
+      - '.github/renovate.json'
   pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/agents/**'
+      - 'docs/design/**'
+      - 'LICENSE*'
+      - '.packit.yaml'
+      - '.github/renovate.json'
   schedule:
     - cron: '0 0 * * *'  # Run daily at midnight UTC
   workflow_dispatch:  # Allow manual triggering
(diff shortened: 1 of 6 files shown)


droideck:
Backport of #7717. Push runs use the workflow on the pushed ref, so this must land on the branch to take effect.

Adaptations: the build runner stays on ubuntu-22.04, and pr_scope.py is omitted since generate_matrix.py here still excludes webui.
codeql.yml and nightly-dispatch.yml stay main-only. Gate shows Skipped on this PR (it fails open).
After merge, pushes trigger no test run; the nightly dispatcher covers the branch.

progier389 (approved):
LGTM

PR #7773, opened 2026-08-26
Issue 7772 - double-free when processing an invalid Session Tracking …

### LINKED ISSUE #7772: double-free when processing an invalid Session Tracking control
**Issue Description**
The Session Tracking control is invalid if its criticality is true. In such case the control is rejected and the operation fails.
The control is freed when it is reject and freed again when the operation returns because the control was stored in the pblock

 - since 3.1

**Steps to Reproduce**
testcase provided

**Expected results**
The double free should not occur on failure

tbordaz:
…control

Bug description:
	The Session Tracking control is invalid if its criticality is true.
	In such case the control is rejected and the operation fails.
	The control is freed when it is rejected and freed again when
	the operation returns because the control was stored in the pblock

Fix description:
	When parsing the control, if it is rejected (and freed) then
	remove the control from the pblock

fixes: #7772

Reviewed by:

## Summary by Sourcery

Prevent rejected Session Tracking controls from remaining attached to request state after parsing failures.

Bug Fixes:
- Prevent double-free crashes when invalid critical Session Tracking controls are rejected during request processing.

Tests:
- Add a regression test that repeatedly submits critical Session Tracking controls and verifies rejected requests do not affect subsequent searches.

### DIFF
--- a/dirsrvtests/tests/suites/session_tracking/session_test.py
+++ b/dirsrvtests/tests/suites/session_tracking/session_test.py
@@ -31,6 +31,86 @@ SESSION_TRACKING_FORMAT_OID = SESSION_TRACKING_CONTROL_OID + ".1234"
 
 pytestmark = pytest.mark.tier0
 
+from ldap.controls import RequestControl
+from pyasn1.type import namedtype,univ
+from pyasn1.codec.ber import encoder
+from pyasn1_modules.rfc2251 import LDAPString,LDAPOID
+
+
+# OID constants
+SESSION_TRACKING_CONTROL_OID = "1.3.6.1.4.1.21008.108.63.1"
+
+
+class SessionTrackingControlCritical(RequestControl):
+  """
+  Class for Session Tracking Control
+  With a criticality set to TRUE
+  """
+
+  class SessionIdentifierControlValue(univ.Sequence):
+    componentType = namedtype.NamedTypes(
+      namedtype.NamedType('sessionSourceIp',LDAPString()),
+      namedtype.NamedType('sessionSourceName',LDAPString()),
+      namedtype.NamedType('formatOID',LDAPOID()),
+      namedtype.NamedType('sessionTrackingIdentifier',LDAPString()),
+    )
+
+  controlType = SESSION_TRACKING_CONTROL_OID
+
+  def __init__(self,sessionSourceIp,sessionSourceName,formatOID,sessionTrackingIdentifier):
+    # criticality set to TRUE for test robustness
+    self.criticality = True
+    self.sessionSourceIp,self.sessionSourceName,self.formatOID,self.sessionTrackingIdentifier = \
+      sessionSourceIp,sessionSourceName,formatOID,sessionTrackingIdentifier
+
+  def encodeControlValue(self):
+    s = self.SessionIdentifierControlValue()
+    s.setComponentByName('sessionSourceIp',LDAPString(self.sessionSourceIp))
+    s.setComponentByName('sessionSourceName',LDAPString(self.sessionSourceName))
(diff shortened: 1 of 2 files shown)


ldap/servers/slapd/control.c
```diff
@@ -557,6 +557,7 @@ get_ldapmessage_controls_ext(
 
 free_and_return:;
     ldap_controls_free(ctrls);
+    slapi_pblock_set(pb, SLAPI_REQCONTROLS, NULL); /* make sure controls are removed from the pblock */
```
vashirov, line 562: Should we also NULLify SLAPI_SESSION_TRACKING?

ldap/servers/slapd/control.c
vashirov: This overwrites the session id set in the if-block.

dirsrvtests/tests/suites/session_tracking/session_test.py
```diff
@@ -31,6 +31,86 @@
 
 pytestmark = pytest.mark.tier0
 
+from ldap.controls import RequestControl
```
vashirov, line 34: Imports should be moved to the top

dirsrvtests/tests/suites/session_tracking/session_test.py
```diff
@@ -31,6 +31,86 @@
...
+from pyasn1.type import namedtype,univ
+from pyasn1.codec.ber import encoder
+from pyasn1_modules.rfc2251 import LDAPString,LDAPOID
+
+
+# OID constants
+SESSION_TRACKING_CONTROL_OID = "1.3.6.1.4.1.21008.108.63.1"
```
vashirov, line 41: This redefines the constant from lib389

dirsrvtests/tests/suites/session_tracking/session_test.py
```diff
@@ -31,6 +31,86 @@
...
+      SESSION_TRACKING_IDENTIFIER
+    )
+
+    # 1000 attempts is enough to trigger SIGSEV
+    for i in range(0, 1000):
+        try:
+            topology_st.standalone.search_ext_s(DEFAULT_SUFFIX,
+                                                ldap.SCOPE_SUBTREE,
+                                                '(uid=*)',
+                                                serverctrls=[st_ctrl])
```
vashirov, line 102: We should add `escapehatch='i am sure'` to prevent warnings

vashirov (commented):
I suspect SLAPI_SESSION_TRACKING has the same bug as SLAPI_REQCONTROLS

ldap/servers/slapd/control.c
tbordaz: Not necessarily.  pb_session_tracking_id from pblock is freed at pblock_done

ldap/servers/slapd/control.c
tbordaz: Indeed it is bad programming. session_tracking_id is NULL in case of error or no control.
In case of valid control, it set it twice. 
I will change that.

dirsrvtests/tests/suites/session_tracking/session_test.py
tbordaz: +1

dirsrvtests/tests/suites/session_tracking/session_test.py
tbordaz: +1

dirsrvtests/tests/suites/session_tracking/session_test.py
tbordaz: +1

mreynolds389 (approved):

PR #7792, opened 2026-09-02
Issue 7595 - Remove the nightly dedup gate and fix dispatched test runs

droideck:
Description: The Gate job skipped main's nightly at an unchanged head, which is when a nightly is useful... (since the test image is rebuilt from Fedora every night)
It never fired for the pytest workflows and switched the cargo nightly off entirely. Every dispatched release-branch run that got past the build also failed in the tmate step, because the dispatcher's empty debug_enabled input resolved to the string "false", which is true in the step condition.

Drop the gate from all three workflows, require debug_enabled == 'true', let the dispatcher rely on each branch's workflow defaults, fail the Build job on an unknown pytest_tests value, and document the dispatch inputs and the unconditional nightly in building.md.

Reviewed by: ?

## Summary by Sourcery

Run nightly test coverage reliably across current and release branches while making dispatch inputs and invalid test selections fail safely.

Bug Fixes:
- Ensure pytest and Cargo nightly workflows run unconditionally instead of skipping unchanged heads.
- Prevent dispatched test runs from incorrectly opening tmate sessions when debug mode is not enabled.
- Fail pytest builds when the requested test suites do not resolve to any valid tests.

Enhancements:
- Allow nightly dispatches to use each release branch's workflow defaults and skip branches without a dispatchable workflow copy.
- Remove obsolete workflow gate dependencies and simplify test job execution conditions.

CI:
- Update nightly dispatching to validate branch workflow triggers before dispatching test workflows.

Documentation:
- Document manual dispatch inputs, nightly environment coverage, and workflow concurrency behavior.


vashirov (approved):
LGTM

PR #7807, opened 2026-09-02
Issue 7806 - Preserve dsEntryDN case on modify

### LINKED ISSUE #7806: Preserve dsEntryDN case on modify
**Issue Description**

Bug Description:
`id2entry_add_ext()` unconditionally overwrites dsEntryDN every time an entry is written to the database. 
When an entry is loaded from disk with `nsslapd-return-original-entrydn` is off, the DN is rebuilt from the entryrdn index in normalized form. A plain modify then permanently replaces the original-case dsEntryDN with this normalized value.

vashirov:
Bug Description:
`id2entry_add_ext()` unconditionally overwrites dsEntryDN every time an entry is written to the database. When an entry is loaded from disk with `nsslapd-return-original-entrydn` is off, the DN is rebuilt from the entryrdn index in normalized form. A plain modify then permanently replaces the original-case dsEntryDN with this normalized value.

Fix Description:
Compare the existing dsEntryDN with the current entry DN before overwriting.
If they refer to the same entry, keep the stored value. If the DN changed (MODRDN) or is not set yet (ADD), update it.

Fixes: https://github.com/389ds/389-ds-base/issues/7806

## Summary by Sourcery

Preserve dsEntryDN casing across modifications while correctly reflecting DN changes.

Bug Fixes:
- Preserve the original-case dsEntryDN during ordinary entry modifications while still updating it when an entry is renamed or newly created.

Enhancements:
- Ensure case-only MODRDN operations update dsEntryDN to the new DN.
- Allow persisted attribute values with CSN state suffixes to be recognized by their base attribute type.

Tests:
- Add tier-1 coverage for dsEntryDN preservation across restart and modification, including case-only renames.

### DIFF
--- a/ldap/servers/slapd/back-ldbm/id2entry.c
+++ b/ldap/servers/slapd/back-ldbm/id2entry.c
@@ -68,8 +68,20 @@ id2entry_add_ext(backend *be, struct backentry *e, back_txn *txn, int encrypt, i
         Slapi_Entry *entry_to_use = encrypted_entry ? encrypted_entry->ep_entry : e->ep_entry;
         memset(&data, 0, sizeof(data));
         entrydn = slapi_entry_get_dn(entry_to_use);
-        slapi_entry_attr_set_charptr(entry_to_use, SLAPI_ATTR_DS_ENTRYDN,
-                entrydn);
+        {
+            char *old_ds_entrydn = slapi_entry_attr_get_charptr(
+                    entry_to_use, SLAPI_ATTR_DS_ENTRYDN);
+            if (old_ds_entrydn == NULL ||
+                slapi_UTF8CASECMP(old_ds_entrydn, entrydn) != 0)
+            {
+                /* ADD: not set yet */
+                /* MODRDN: DN changed, update it */
+                slapi_entry_attr_set_charptr(entry_to_use,
+                        SLAPI_ATTR_DS_ENTRYDN, entrydn);
+            }
+            /* MODIFY: same DN, keep the original case */
+            slapi_ch_free_string(&old_ds_entrydn);
+        }
 
         struct backdn *oldbdn = NULL;
         Slapi_DN *sdn =
--- a/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
+++ b/dirsrvtests/tests/suites/basic/ds_entrydn_test.py
@@ -17,6 +17,8 @@ from test389.topologies import topology_st as topo
 
 log = logging.getLogger(__name__)
 
+pytestmark = pytest.mark.tier1
+
 SUFFIX = "dc=Example,DC=COM"
 SUBTREE = "ou=People,dc=Example,DC=COM"
 NEW_SUBTREE = "ou=humans,dc=Example,DC=COM"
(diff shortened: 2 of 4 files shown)


ldap/servers/slapd/back-ldbm/id2entry.c
```diff
@@ -68,8 +68,20 @@ id2entry_add_ext(backend *be, struct backentry *e, back_txn *txn, int encrypt, i
         Slapi_Entry *entry_to_use = encrypted_entry ? encrypted_entry->ep_entry : e->ep_entry;
         memset(&data, 0, sizeof(data));
         entrydn = slapi_entry_get_dn(entry_to_use);
-        slapi_entry_attr_set_charptr(entry_to_use, SLAPI_ATTR_DS_ENTRYDN,
-                entrydn);
+        {
+            char *old_ds_entrydn = slapi_entry_attr_get_charptr(
+                    entry_to_use, SLAPI_ATTR_DS_ENTRYDN);
+            if (old_ds_entrydn == NULL ||
+                slapi_UTF8CASECMP(old_ds_entrydn, entrydn) != 0)
```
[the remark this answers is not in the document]
vashirov: Both values were already normalized for spacing and excaping. `slapi_sdn_compare` normalizes the case, this is what we're trying to avoid.

mreynolds389 (approved):

ldap/servers/slapd/back-ldbm/id2entry.c
droideck, line 75: I think this breaks case-only renames (uid=foo -> uid=FOO)... IIUC, we support it (worth a test).
Fixing capitalisation of a cn is a pretty normal admin thing.

The compare itself is fine, both sides are already normalized as you noted so slapi_sdn_compare would add nothing, and folding case is required or the modify case would not work. The catch is that folding also makes uid=foo and uid=FOO look the same here. ec still carries the old dsEntryDN and nothing in ldbm_back_modrdn sets a new one for the target, so the old value stays on disk. With return-original-entrydn on (the default) the rename comes back after a restart, and db2ldif writes the old dn too, IIUC.

A possible fix: set dsEntryDN on ec right after slapi_entry_set_sdn(ec->ep_entry, &dn_newdn) in ldbm_modrdn.c (line 662), same as dsentrydn_modrdn_update already does for children. Then this compare is a no-op for modrdn and still protects modify. And yeah, we might need a test for that.

droideck (changes requested):

ldap/servers/slapd/back-ldbm/id2entry.c
vashirov: Indeed, case-only rename was not covered. Changes were made.

droideck (approved):
LGTM!

ldap/servers/slapd/back-ldbm/id2entry.c
vashirov: This change broke dirsrvtests/tests/suites/ds_tools/replcheck_test.py::test_inconsistencies test, investigating...

vashirov:
BDB Test replication/repl_log_monitoring_test.py passes for me locally, looks like. a race condition.

Issue #7815, opened 2026-09-03
Support nsslapd-attribute-name-exceptions when adding entries


msolberg:
We are working on a migration of a 400K object database which was created prior to RFC 4512, when underscores were removed from the list of allowed characters in DNs. To import the database, we need to disable syntax validation, even though nsslapd-attribute-name-exceptions is turned on in cn=config.

We are proposing a patch to keystring_validate() in ldap/servers/plugins/syntaxes/validate.c which will allow underscores when nsslapd-attribute-name-exceptions is set.

mreynolds389:
8c2711bd6..c2789766c  389-ds-base-3.3 -> 389-ds-base-3.3
5ce6c1715..63ee78f7d  389-ds-base-3.2 -> 389-ds-base-3.2
fb0d6b742..8d7f08548  389-ds-base-3.1 -> 389-ds-base-3.1
2846d5288..fd5784575  389-ds-base-3.0 -> 389-ds-base-3.0
9f6f80cc4..6e40f8e3e  389-ds-base-2.9 -> 389-ds-base-2.9
c525b7530..74ead7388  389-ds-base-2.8 -> 389-ds-base-2.8
50910b918..b22916e9d  389-ds-base-2.7 -> 389-ds-base-2.7
f658aef86..5a13ce61b  389-ds-base-2.6 -> 389-ds-base-2.6

PR #7816, opened 2026-09-03
Issue 7815 - Support nsslapd-attribute-name-exceptions when adding entries

msolberg:
Problem Description:

When nsslapd-attribute-name-exceptions is set to on, the schema loader (schema_check_name in schema.c) correctly allows underscores in attribute names. However, the syntax validation plugins used during LDIF import and entry operations do not honor this setting. The keystring_validate function in validate.c uses the IS_KEYCHAR macro, which only permits [a-zA-Z0-9-] — no underscore. This means:

  - DN-valued attributes (like seeAlso, member) whose values contain DNs with underscored attribute types are rejected by slapi_entry_syntax_check
  - The rejection chain: slapi_entry_syntax_check → dn_validate → distinguishedname_validate → rdn_validate → keystring_validate → IS_KEYCHAR('_') = false → error

Fix Description:

ldap/servers/plugins/syntaxes/validate.c — Modified keystring_validate to check config_get_attrname_exceptions() and allow underscores when the setting is enabled, matching the behavior of schema_check_name in schema.c. Since rdn_validate, oid_validate (cis.c), and guide syntax validators all delegate to keystring_validate, they all inherit the fix automatically.

dirsrvtests/tests/suites/syntax/acceptance_test.py — Added regression test test_attrname_exceptions_dn_syntax that enables the exception, creates a schema attribute with an underscore, adds an entry using it, and verifies a seeAlso value containing a DN with the underscored attribute type is accepted.

relates: https://github.com/389ds/389-ds-base/issues/7815

Author: Michael Solberg <msolberg@redhat.com>

Assisted-by: Claude

## Summary by Sourcery

Allow syntax validation to accept underscored attribute names when nsslapd-attribute-name-exceptions is enabled.

Bug Fixes:
- Honor nsslapd-attribute-name-exceptions during syntax validation so DN-valued attributes can reference attribute types containing underscores.

Tests:
- Add a regression test covering underscored attribute types in DN values when attribute-name exceptions are enabled.

### DIFF
--- a/ldap/servers/plugins/syntaxes/validate.c
+++ b/ldap/servers/plugins/syntaxes/validate.c
@@ -29,6 +29,7 @@ keystring_validate(
 {
     int rc = 0; /* assume the value is valid */
     const char *p = begin;
+    int allow_exceptions = config_get_attrname_exceptions();
 
     if ((begin == NULL) || (end == NULL)) {
         rc = 1;
@@ -38,10 +39,14 @@ keystring_validate(
     /* Per RFC4512:
      *
      *   keystring = leadkeychar *keychar
+     *
+     * When nsslapd-attribute-name-exceptions is enabled,
+     * underscores are also permitted (for compatibility
+     * with legacy directory servers such as Sun ONE DS).
      */
     if (IS_LEADKEYCHAR(*p)) {
         for (p++; p <= end; p++) {
-            if (!IS_KEYCHAR(*p)) {
+            if (!IS_KEYCHAR(*p) && !(allow_exceptions && *p == '_')) {
                 rc = 1;
                 goto exit;
             }
--- a/ldap/servers/plugins/syntaxes/validate.c
+++ b/ldap/servers/plugins/syntaxes/validate.c
@@ -29,7 +29,6 @@ keystring_validate(
 {
     int rc = 0; /* assume the value is valid */
     const char *p = begin;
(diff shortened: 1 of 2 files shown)


progier389:
The new testcase: test_attrname_exceptions_dn_syntax failed on lmdb:
https://github.com/389ds/389-ds-base/actions/runs/33803009808/job/100814095277?pr=7816

msolberg:
> The new testcase: test_attrname_exceptions_dn_syntax failed on lmdb: https://github.com/389ds/389-ds-base/actions/runs/33803009808/job/100814095277?pr=7816

Working on it. Thanks.

ldap/servers/plugins/syntaxes/validate.c
```diff
@@ -29,6 +29,7 @@ keystring_validate(
 {
     int rc = 0; /* assume the value is valid */
     const char *p = begin;
+    int allow_exceptions = config_get_attrname_exceptions();
```
mreynolds389: I know it's minor, but to avoid unnecessary locking of the config can you move config_get_attrname_exceptions() to line 47 after "if (IS_LEADKEYCHAR..." ?

msolberg: Sure! Makes sense.

mreynolds389 (approved):

PR #7824, opened 2026-09-08
Issue 7823 - CI: stabilize test_controltype_expired_grace_limit

### LINKED ISSUE #7823: CI: stabilize test_controltype_expired_grace_limit
**Issue Description**
This test runs the teardown while the clean connection is not yet properly authenticated

 - Platform: all
 - Package and version: 389-ds-base

**Steps to Reproduce**
1. run CI password/pwdPolicy_controls_sequence_test.py::test_controltype_expired_grace_limit
it sometime fails

**Expected results**
It should not fail

tbordaz:
Bug description:
	bind_and_get_control authenticates the connection to
	directory manager before returning. The problem is
	that authentication may still pending upon return
	and teardown is called with the previous auth resulting
	in err=50 (insufficient access)

Fix description:
	Force a synchrone call (whoami_s) after DM authentication

fixes: #7823

Reviewed by:

## Summary by Sourcery

Bug Fixes:
- Stabilize the expired grace-limit control test by ensuring directory manager authentication completes before teardown.

### DIFF
--- a/dirsrvtests/tests/suites/password/pwdPolicy_controls_sequence_test.py
+++ b/dirsrvtests/tests/suites/password/pwdPolicy_controls_sequence_test.py
@@ -68,7 +68,7 @@ def bind_and_get_control(topo):
         res_ctrls = ast.literal_eval(str(e))
         pass
 
-    topo.standalone.simple_bind(DN_DM, PASSWORD)
+    topo.standalone.simple_bind_s(DN_DM, PASSWORD)
     return res_ctrls
 
 


dirsrvtests/tests/suites/password/pwdPolicy_controls_sequence_test.py
```diff
@@ -69,6 +69,8 @@ def bind_and_get_control(topo):
         pass
 
     topo.standalone.simple_bind(DN_DM, PASSWORD)
+    whoami = topo.standalone.whoami_s()
```
progier389: IMHO  it would be clearer to use topo.standalone.simple_bind_s(DN_DM, PASSWORD) 
so the bind is synchronous

progier389 (changes requested):

dirsrvtests/tests/suites/password/pwdPolicy_controls_sequence_test.py
tbordaz: Indeed much simpler and valid. The fix with whoami is failing when bind is slow so it should have loop.
I changed the fix

progier389 (approved):
LGTM

PR #7831, opened 2026-09-09
Issue 7830 - Instance creation fails when `/usr/lib` is read-only

### LINKED ISSUE #7830: Instance creation fails when `/usr/lib` is read-only
**Issue Description**
After 989e13895684ff0df2e7ee848e348a43ba9e5f14 instance creation fails when `/usr/lib` is read-only. This is caused by the change in tmpfiles installation. Previously, tmpfiles configs were installed in `/etc/tmpfiles.d`, which is writeable. But after the change `dscreate` would install it in `/usr/lib`. If there are multiple instances installed, duplicate entries are created for common directories like `/run/dirsrv` and `/run/lock/dirsrv`.

tmpfiles should be split into system-wide:
```
d /run/dirsrv 0770 dirsrv dirsrv
d /run/lock/dirsrv/ 0770 dirsrv dirsrv
```
and instance specific:
```
d /run/lock/dirsrv/slapd-localhost 0770 dirsrv dirsrv
```

Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=2530878

vashirov:
Bug Description:
After the tmpfiles installation change, instance creation fails when `/usr/lib` is read-only. Configs moved from writable `/etc/tmpfiles.d` to `/usr/lib`. Multiple instances also duplicate common entries such as `/run/dirsrv` and `/run/lock/dirsrv`.

Fix Description:
Split tmpfiles into system-wide config and instance-specific configs.

Fixes: https://github.com/389ds/389-ds-base/issues/7830

## Summary by Sourcery

Separate system-wide and per-instance tmpfiles configuration so instance lifecycle operations work with read-only vendor directories and avoid duplicated shared runtime entries.

Bug Fixes:
- Fix instance creation when the vendor `/usr/lib` tmpfiles directory is read-only by placing instance-specific tmpfiles configuration in the writable administrator directory.
- Prevent duplicate system-wide tmpfiles entries when multiple instances are created.

Enhancements:
- Support separate system-wide and runtime-generated tmpfiles directories, including configurable runtime tmpfiles paths and cleanup from either location.

Build:
- Install the system-wide DirSrv tmpfiles configuration separately from generated per-instance configuration.

Documentation:
- Document the additional runtime tmpfiles path.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -309,9 +309,15 @@ gdbautoloaddir = $(prefixdir)/share/gdb/auto-load$(sbindir)
 cockpitdir = $(prefixdir)/share/cockpit@cockpitdir@
 metainfodir = $(prefixdir)/share/metainfo/389-console
 tmpfiles_d = @tmpfiles_d@
+tmpfiles_ddir = @tmpfiles_d@
+runtime_tmpfiles_d = @runtime_tmpfiles_d@
 
 sysctldir = @sysctl_d@
 sysusersdir = @sysusers_d@
+
+if INSTALL_TMPFILES
+dist_tmpfiles_d_DATA = ldap/admin/src/dirsrv.conf
+endif
 
 defaultuser=@defaultuser@
 defaultgroup=@defaultgroup@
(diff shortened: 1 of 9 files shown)


progier389 (approved):
LGTM

PR #7844, opened 2026-09-15
Issue 7843 - SyntaxWarning in lib389/lint.py

### LINKED ISSUE #7843: SyntaxWarning in lib389/lint.py
**Issue Description**
Test cases output shows: 

../../../../src/lib389/lib389/lint.py:29
  /home/progier/sb/i7227/tst/source/389-ds-base/src/lib389/lib389/lint.py:29: SyntaxWarning: "\=" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\="? A raw string is also an option.
    cn: dc\=example\,dc\=com

 - Platform: Fedora
 - Package and version:main branch

**Steps to Reproduce**
1. Run pytest and check result
2. pytest dirsrvtests/tests/suites/tls/ecdsa_test.py
3. Show that the log displays a SyntaxWarning

**Expected results**
No SyntaxWarning

xhon-pelushi:
Fixes #7843.

The `DSBLE0001` `detail` text is a plain (non-raw) string, and the example mapping tree entry inside it writes the DN escapes as single backslashes:

```
cn: dc\=example\,dc\=com
```

`\=` and `\,` are not valid Python escape sequences. They still evaluate to a backslash followed by the character, so the help text renders correctly today, but Python 3.12 emits a `SyntaxWarning` for each and the sequences are slated to become errors in a later release. The warning appears in the output of any test run that imports lib389, which is what #7843 reports.

This writes them as `\\=` and `\\,`, matching the `\\3D` and `\\2C` already used two lines above, so the whole example is now consistent about escaping.

**The rendered text is unchanged.** Compiling the module before and after produces byte-identical string constants, and the help output still reads:

```
cn=dc\3Dexample\2Cdc\3Dcom,cn=mapping tree,cn=config
cn: dc=example,dc=com
cn: dc\=example\,dc\=com
```

After the change the module compiles and executes cleanly under `python3 -W error::SyntaxWarning`.

### One other thing, if it is useful

Compiling every `.py` in the tree with warnings captured turns up **21 more invalid escape sequences**, all in the test suite, so `pytest` output stays noisy even with this fixed:

```
dirsrvtests/tests/suites/acl/globalgroup_part2_test.py:123,179,235     '\ '
dirsrvtests/tests/suites/acl/globalgroup_test.py:371,399,427          '\ '
dirsrvtests/tests/suites/acl/syntax_test.py:70                        '\('
dirsrvtests/tests/suites/acl/syntax_test.py:118                       '\{'
dirsrvtests/tests/suites/ds_logs/ds_logs_test.py:897,1112             '\(' '\_'
dirsrvtests/tests/suites/fourwaymmr/fourwaymmr_test.py:228            '\='
dirsrvtests/tests/suites/monitor/monitor_test.py:164                  '\|'
dirsrvtests/tests/suites/replication/acceptance_test.py:849           '\('
dirsrvtests/tests/suites/replication/virtual_attribute_replication_test.py:145,203  '\('
dirsrvtests/tests/suites/roles/basic_test.py:378,441,497,814          '\('
dirsrvtests/tests/tickets/ticket48973_test.py:102                     '\('
dirsrvtests/tests/tickets/ticket49471_test.py:68                      '\*'
```

Most are regex or ACI strings that want to be raw strings, but a few are not mechanical — `syntax_test.py:118` is `f'(target = ldap:\\\{DEFAULT_SUFFIX})'`, and the `\ ` cases in the ACI tests may or may not be intentional. I have deliberately left them out of this PR because I cannot run the test suite here to confirm the strings still behave the same, and I would rather not change assertion text I cannot exercise. Happy to do them as a follow-up if you would like, either all at once or split per suite.

## Summary by Sourcery

Bug Fixes:
- Remove invalid Python escape sequences from the lib389 lint help text to eliminate SyntaxWarnings while preserving the rendered example.


progier389 (approved):
LGTM

PR #7845, opened 2026-09-15
Issue 7842 - epoll used incorrect timeout

### LINKED ISSUE #7842: Epoll + Multi listener updates
**Issue Description**
The epoll connection handling has some issues when the server is configured with multiple connection table listeners.

Multiple listeners do not provide a performance benefit when using epoll. However, existing deployments may already be configured with the number of listeners greater than one, and enabling epoll must not break those configurations.

- Connections are not correctly distributed across the connection-table lists, resulting in uneven lists.
- Connections removed from epoll while being processed by a worker are not always added back once the worker has finished.
- Closing connections released by workers need to wake the owning connection table listener so they can be reaped from the active list.
- The epoll signal pipe needs to be drained when processing its event to prevent level triggered epoll from repeatedly waking on the same signal.

 - Package and version: Main branch

**Steps to Reproduce**
1. Build server with epoll support enabled.
2. Configure the server with multiple listeners.
3. Generate activity across the connections so they are dispatched to worker threads.
4. Observe the unbalanced listenrs
5. After a while the server runs out of free slots in the connection table, as they are never removed from the freelist.

Firstyear:
Bug Description: epoll_wait was using pr_timeout which is a platform dependent interval. epoll_wait however takes milliseconds.

Fix Description: Use milliseconds directly in epoll_wait. Extend the timeout time briefly when epoll is enabled to prevent excess hot looping in the wait function.

fixes: https://github.com/389ds/389-ds-base/issues/7842

Review by: ???

## Summary by Sourcery

Fix epoll timeout handling and reduce excessive wakeups in connection-table wait loops.

Bug Fixes:
- Correct epoll wait timeouts by passing millisecond values directly instead of platform-dependent intervals.
- Increase epoll wakeup intervals to reduce unnecessary hot looping during waits.


Firstyear:
@jchapma I don't think this fixes the problem, but it certainly relates to it.

Firstyear:
Something else @jchapma is we are setting epoll_wait maxevents to list_size which could cause it's own set of issues, perhaps we should cap that value or experiment with it?

 I am wondering if what's happening is that because we use level triggered fd reads then every single active connection will trigger in every epoll_wait, which then leads to every active con being triggered into connection_activity(). So as we add more tables, we have more epoll_wait events, which leads to more hammering of connection_activity queuing up the events (sometimes for an already queued conn/fd).

Consider have a look at this btw it was really good for me to read. 
[lusp_epoll_slides-mkerrisk-man7.org.pdf](https://github.com/user-attachments/files/32220433/lusp_epoll_slides-mkerrisk-man7.org.pdf)

tbordaz (approved):
LGTM
A question regarding the change of timer values. Did you increase timeout to save CPU consumption ? Was it significant ?
the impact being only on shutdown detection I am fine increasing it.

jchapma:
> Something else @jchapma is we are setting epoll_wait maxevents to list_size which could cause it's own set of issues, perhaps we should cap that value or experiment with it?
> 
> I am wondering if what's happening is that because we use level triggered fd reads then every single active connection will trigger in every epoll_wait, which then leads to every active con being triggered into connection_activity(). So as we add more tables, we have more epoll_wait events, which leads to more hammering of connection_activity queuing up the events (sometimes for an already queued conn/fd).
> 
> Consider have a look at this btw it was really good for me to read. [lusp_epoll_slides-mkerrisk-man7.org.pdf](https://github.com/user-attachments/files/32220433/lusp_epoll_slides-mkerrisk-man7.org.pdf)

That is a good point, defo worth experimenting with. My understanding is that list_size doesn't generate more events, it just allows epoll_wait to return more ready fds in one call. But with multiple ct lists we have multiple threads feeding connection_activity concurrently, so capping maxevents and comparing the performance could be interesting.

Firstyear:
> LGTM A question regarding the change of timer values. Did you increase timeout to save CPU consumption ? Was it significant ? the impact being only on shutdown detection I am fine increasing it.

It's nothing to do with CPU consumption - it's just so that we aren't pinging the shutdown or cert refresh atomic values as frequently in a loop. 

> That is a good point, defo worth experimenting with. My understanding is that list_size doesn't generate more events, it just allows epoll_wait to return more ready fds in one call. But with multiple ct lists we have multiple threads feeding connection_activity concurrently, so capping maxevents and comparing the performance could be interesting.

Correct, but there is a catch - if you epoll_wait with list_size you get all the events back and then we have to loop pretty aggressively to queue them all up. There might be some trades there depending on how active the connections are etc. 

If we use epoll_wait with a smaller size, epoll actually round-robins/randomises which fd's are yielded to us to queue up. So it effectively load-balances for us in a way. 

This could have some interesting interactions. First as mentioned, since we level trigger we have to consider some odd possibilities. Lets think about two FD's for now. 

```
FD 1 <- recieves data
FD 2 <- recieves data
epoll_wait ( ..., X)

Conncurrent - ( FD 2 recieves more data, queue FD 1 to do work)
queue FD 2 to do work

epoll_wait( ..., X)
```

So here because we level trigger, after the second call to epoll_wait, since FD2 got more data it will be queued again. If we limit X (the max events size) to a small number, then in this example if we limited it to 1, then we wouldn't have double checked FD2 for work. (I think FD2 won't be queued again due to the `c->c_gettingber` check on line 1960 of ldap/servers/slapd/daemon.c ). So that means when we "loop" were looping over "actually interesting activity" rather than "something we already know is in the pipeline".

jchapma:
> If we use epoll_wait with a smaller size, epoll actually round-robins/randomises which fd's are yielded to us to queue up. So it effectively load-balances for us in a way.
> 
> This could have some interesting interactions. First as mentioned, since we level trigger we have to consider some odd possibilities. Lets think about two FD's for now.
> 
> ```
> FD 1 <- recieves data
> FD 2 <- recieves data
> epoll_wait ( ..., X)
> 
> Conncurrent - ( FD 2 recieves more data, queue FD 1 to do work)
> queue FD 2 to do work
> 
> epoll_wait( ..., X)
> ```
> 
> So here because we level trigger, after the second call to epoll_wait, since FD2 got more data it will be queued again. If we limit X (the max events size) to a small number, then in this example if we limited it to 1, then we wouldn't have double checked FD2 for work. (I think FD2 won't be queued again due to the `c->c_gettingber` check on line 1960 of ldap/servers/slapd/daemon.c ). So that means when we "loop" were looping over "actually interesting activity" rather than "something we already know is in the pipeline".

I like the idea of limiting the number of events, currently that number is very large. I will start by setting it to 1 and rerun the perf tests.

jchapma:
Regarding this PR, I changed the epoll_wait timeout in the other epoll PR, but when you merge this I can remove that. I would liek to see the effect of changing the epoll_wait wakeup interval. Let me add it to my list.

LGTM

Firstyear:
@jchapma If you want I'm happy to close this in favour of your PR since you're testing more there anyway.

Issue #5181, opened 2022-02-25
Add epoll support to 389-ds-base


llile:
Add epoll support to 389-ds-base

I tried not to deviate too far from the existing code at this point, the major changes at this point are:

Listeners moved to a listen_table (setup_epoll_listen_pds)
    listen_table is a list of Connection's so the can be handled in the same way as a client Connection
    Differentiated by Connection->conn_state = CONN_STATE_LISTEN
    Connection_Table->listen_count is no longer maintained
    Eliminates listener pd handing from setup_pr_read_pds

epoll_arm_listen_pds
    Adds or removes all listener pds from epoll
    Triggered from main event loop based on connection count limits

Connection_Table->fd is currently only maintained for listeners
    This could likely be fully eliminated, but I'm not sure what to do with signalpipe to accomplish this end

Connection_Table->epollfd has been added to hold the epoll fd set

handle_new_connection
    Adds descriptors to epoll immediately
    Eliminates the need for setup in setup_pr_read_pds

epoll_pr_idle_pds ( timeout related section of setup_pr_read_pds )
    Should only handle client timeouts or special cases for re-adding a descriptor to epoll
    Eliminates the needs from the remainder of setup_pr_read_pds

setup_pr_read_pds is not used with epoll

llile:
Hi,

Can someone link this to PR 5182?  I can't seem to work out how to do that.

Thanks.

--Larry

PR #7846, opened 2026-09-15
Issue 7842 - Epoll + Multi listener updates

### LINKED ISSUE #7842: Epoll + Multi listener updates

jchapma:
Description:
The epoll connection handling has some issues when the server is configured with multiple connection table listeners.

Multiple listeners do not provide a performance benefit when using epoll. However, existing deployments may already be configured with the number of listeners greater than one, and enabling epoll must not break those configurations

Fix:
Ensure each connection is handled by the connection table listener it is assigned to. Register the connection socket and idle timer with that listener's epoll instance, and use the same epoll instance when removing them.

Closing connections are reaped by their owning connection table listener. When a worker thread marks a connection for closure, signal the appropriate listener so it wakes from epoll_wait and can safely remove the connection.

Also handle cases where both the connection socket and idle timer generate events for the same connection in a single epoll_wait cycle, avoiding duplicate processing of connections that have already been closed or removed from the active list.

Fixes: https://github.com/389ds/389-ds-base/issues/7842
Relates: https://github.com/389ds/389-ds-base/issues/5181

Reviewed by:

## Summary by Sourcery

Ensure epoll remains reliable with multiple connection listeners by routing events and cleanup through each connection’s owning listener.

Bug Fixes:
- Fix epoll connection handling when multiple connection-table listeners are configured by keeping each connection and its idle timer on its owning listener’s epoll instance.
- Ensure closing connections are reaped reliably and prevent duplicate or stale processing when multiple epoll events reference the same connection.

Enhancements:
- Use one-shot epoll monitoring to coordinate socket readiness with worker-thread processing and re-arm connections as capacity becomes available.

Build:
- Enable epoll by default when no explicit setting is provided.

Chores:
- Correct signal-pipe nonblocking setup and add safer handling for already-removed epoll descriptors.


jchapma:
Epoll is currently enabled by default so the CI framework can detect any regressions.

ldap/servers/slapd/daemon.c
```diff
@@ -1577,27 +1646,33 @@ ct_list_thread(uint64_t threadnum)
...
-            select_return = epoll_wait(the_connection_table->epoll_fd[threadid], events, the_connection_table->list_size, pr_timeout);
+         struct epoll_event events[the_connection_table->list_size];
+         select_return = epoll_wait(the_connection_table->epoll_fd[threadid], events, the_connection_table->list_size, slapd_ct_thread_wakeup_timer);
 #else /* !ENABLE_EPOLL */
          num_poll = setup_pr_read_pds(the_connection_table, threadid);
          select_return = POLL_FN(the_connection_table->fd[threadid], num_poll, pr_timeout);
 #endif /* ENABLE_EPOLL */
          switch (select_return) {
              case 0: /* Timeout */
+#ifdef ENABLE_EPOLL
```
Firstyear, line 1665: This could create a starvation situation where if every ct list entry is filled, but a small number of connections are highly active, then we'll never hit the timeout condition to reap the closed conns.

So perhaps in the timeout situation, we do the full reap, during a new-data ready (or just before the call to epoll_wait) we do a call to reap_closing_connections but we cap how many ct entries we examine to minimise latency in the hot path.

ldap/servers/slapd/daemon.c
```diff
@@ -1953,6 +2085,21 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
         if (c->c_state != CONN_STATE_FREE) {
             /* this check can be done without acquiring the mutex */
             if (c->c_gettingber) {
+#ifdef ENABLE_EPOLL
+                /*
+                 * A worker thread is reading from this socket. Poll omits it from the
+                 * next wait set. Epoll is level triggered, so we must delete it until
+                 * the worker thread is done and adds it back, or this thread busy loops.
+                 */
+                epoll_ctl_del_fd(ct->epoll_fd[list_num], c->c_sd);
```
Firstyear, line 2094: So I'm not sure if we actually want this here because the call to epoll_ctl_del_fd is a syscall which does make it quite costly. Given that we have the c->gettingber check in the pr_read_ready path, we just skip already active conns anyway.

ldap/servers/slapd/daemon.c
```diff
@@ -1963,23 +2110,39 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
...
-                    out_flags = events[i].events;
-                } else {
-                    out_flags = 0;
+            /*
+             * Poll omits this fd when at max threads per conn. Epoll must delete
+             * it until a worker drops below the cap and adds it back, or
+             * connection_activity will increment past the cap.
+             */
+            if (c->c_threadnumber >= c->c_max_threads_per_conn) {
+                epoll_ctl_del_fd(ct->epoll_fd[list_num], c->c_sd);
```
Firstyear, line 2120: As above.

Firstyear (changes requested):
Just remember, every call to `epoll_ctl` is a syscall which may introduce a lot of latency, so it may be better to avoid putting the sockets into/out of the epoll table without good cause.

ldap/servers/slapd/daemon.c
```diff
@@ -1963,23 +2110,39 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
             if (pthread_mutex_trylock(&(c->c_mutex)) == EBUSY) {
                 continue;
             }
-            if (connection_is_active_nolock(c) && c->c_gettingber == 0) {
```
tbordaz, line 1966: We just acquired the c_mutex we must check again the condition gettingber==0 before doing the rest.

ldap/servers/slapd/connection.c
```diff
@@ -2213,11 +2234,23 @@ connection_threadmain(void *arg)
                     }
                     conn->c_threadnumber--;
                     connection_release_nolock(conn);
+#ifdef ENABLE_EPOLL
+                    /*
+                    * Retry adding the socket after decrementing c_threadnumber.
+                    * The earlier attempt may have been skipped if at the thread limit.
+                    */
+                    connection_epoll_add_socket(conn);
```
tbordaz, line 2242: Shouldn't we skip this call in case of CONN_FLAG_CLOSING

ldap/servers/slapd/daemon.c
```diff
@@ -1567,6 +1607,35 @@ ct_thread_cleanup(void)
...
+ * calls this, so there is no poll array to rebuild. We need a way to detect
+ * closing connections and unlink them from the active list. The worker threads
+ * mark the connection as closing when its finished with it.
+ */
+static void
+reap_closing_connections(Connection_Table *ct, int list_num)
+{
+    Connection *c;
+    Connection *next;
+
```
tbordaz, line 1622: We should hold ct->table_mutex before iterating.
any risk of deadlock with c_mutex if the locks are taken in the opposite order ?

tbordaz (commented):
few questions

tbordaz:
Hi @jchapma ,
my understanding is that this PR is fixing several issues (perf using one epoll_wait cycle for sd/idle, closed connections not returned, ...). Should they be fixed at the same time or is it possible to do separate PR. For example exhausting connection looks a blocker while perf improvement has a lower prio.

Firstyear:
> my understanding is that this PR is fixing several issues (perf using one epoll_wait cycle for sd/idle, closed connections not returned, ...). Should they be fixed at the same time or is it possible to do separate PR. For example exhausting connection looks a blocker while perf improvement has a lower prio.

They all kind of interlink - sure you could split it to multiple patches, but I don't think that helps with the review much. 

Additionally, some of these changes to improve the connection exhaust issue may regress performance anyway ....

ldap/servers/slapd/daemon.c
jchapma: > This could create a starvation situation where if every ct list entry is filled, but a small number of connections are highly active, then we'll never hit the timeout condition to reap the closed conns.
> 
> So perhaps in the timeout situation, we do the full reap, during a new-data ready (or just before the call to epoll_wait) we do a call to reap_closing_connections but we cap how many ct entries we examine to minimise latency in the hot path.

Thats a good point, doing a reap on timeout if the epoll_wait continuously returns ready events would be a problem.

Regarding the cap on how many ct entries we check, how about we use a small fixed batch size first, just to see if it helps, we could then tune the batch size based on performance testing. We would also need to keep track of where the previous reap stopped, if each ct list tracked its last reap position, the next reap could continue from there.

jchapma: Thats a good point, doing a reap on timeout if the epoll_wait continuously returns ready events would be a problem.

Regarding the cap on how many ct entries we check, how about we use a small fixed batch size first, just to see if it helps, we could then tune the batch size based on performance testing. We would also need to keep track of where the previous reap stopped, if each ct list tracked its last reap position, the next reap could continue from there.

ldap/servers/slapd/daemon.c
jchapma: Yes, checking c->gettingber would prevent the connection from being dispatched to a worker again. The reason I added the del/add cycle was to avoid a potential busy loop that could happen if the socket remains readable while the worker is processing it, epoll_wait would keep returning the same fd, only for us to skip it because c->c_gettingber is set.

Saying that, i agree the del/add introduces additional syscalls, maybe I could remove it and see if t he repeated events actually cause an issue in practice.

ldap/servers/slapd/daemon.c
jchapma: We check the connection flags and gettingber a few lines later. We only add the conn op to the work_q is gettingber == 0

ldap/servers/slapd/connection.c
jchapma: connection_epoll_add_socket already bails out when CONN_FLAG_CLOSING is set,  so this call is a no-op for a closing connection. The flag check below is  to wake the ct-list thread so it can reap the connection.

ldap/servers/slapd/daemon.c
jchapma: IIUC we should take c_mutex and then table_mutex (fe.h+77). reap_closing_connections only trylocks c_mutex, the table_mutex is taken inside connection_table_move_connection_out_of_active_list

jchapma:
> Hi @jchapma , my understanding is that this PR is fixing several issues (perf using one epoll_wait cycle for sd/idle, closed connections not returned, ...). Should they be fixed at the same time or is it possible to do separate PR. For example exhausting connection looks a blocker while perf improvement has a lower prio.

The connection leak is the main issue, I agree. The rest of this change is kind of on the same path (socket and idle timer are two fds for one conn, so one epoll_wait can return both, and we have to skip the stale event after the first has already unlinked or closed it). Multi listener is the same class of bug (event on the wrong epoll instance / wrong ct-list thread). They all live in handle_pr_read_ready and the worker cleanup, so splitting would be two PRs rewriting the same functions. 

I get what you are saying though. lets get it stable then work on performance, I feel the changes here are functional. I value your opinion, so I am happy to split if you would prefer a smaller first PR.

Firstyear:
@jchapma I have the opposite opinion - I think the PR can stay as one, I don't see benefit to splitting it up since today it's not even the default network path. I think we can review it well enough as is, and go from there. It makes it easier for all of us.

jchapma:
For evaluation I used two workloads:
- ltest with 10k open connections running sparse lightweight base level searches. This targest the connection table listener with a large fd set and the overhead of rebuilding the fd list each cycle, compared with the persistent epoll interest set.
- dsrate with 32 client threads running indexed searches over a small  number of busy connections. This targets the listener/worker handoff.

I initially used an add/del fd cycle to prevent the listener receiving events while a worker was processing the connection. However, dsrate showed a perf regression due to the additional epoll_ctl() calls. I replaced this with EPOLLONESHOT, which disabled the fd after am event is delivered, the fd is "re-enabled" with EPOLL_CTL_MOD when the worker is finished. 

I also tested different maxevent values of 1, 32 and 64. maxevents=1  did improved ltest latency but reduced dsrate throughput. 32 and 64 produced similar results, so I stuck with 32 as a small fixed batch rather than using the full connection table list size.

Final results with EPOLLONESHOT and maxevents=32 (2 runs)

image

In the single listener case, epoll reduces ltest latency by ~70% while dsrate is effectively unchanged. Multiple listeners remain functional, although dsrate decreases as the listener count increases.

tbordaz:
> For evaluation I used two workloads:
> 
>     * ltest with 10k open connections running sparse lightweight base level searches. This targest the connection table listener with a large fd set and the overhead of rebuilding the fd list each cycle, compared with the persistent epoll interest set.
> 
>     * dsrate with 32 client threads running unindexed searches over a small  number of busy connections. This targets the listener/worker handoff.
> 
> 
> I initially used an add/del fd cycle to prevent the listener receiving events while a worker was processing the connection. However, dsrate showed a perf regression due to the additional epoll_ctl() calls. I replaced this with EPOLLONESHOT, which disabled the fd after am event is delivered, the fd is "re-enabled" with EPOLL_CTL_MOD when the worker is finished.
> 
> I also tested different maxevent values of 1, 32 and 64. maxevents=1 did improved ltest latency but reduced dsrate throughput. 32 and 64 produced similar results, so I stuck with 32 as a small fixed batch rather than using the full connection table list size.
> 
> Final results with EPOLLONESHOT and maxevents=32 (2 runs)
> image
> 
> In the single listener case, epoll reduces ltest latency by ~70% while dsrate is effectively unchanged. Multiple listeners remain functional, although dsrate decreases as the listener count increases.

It is always interesting to have access logs samples along the overall metric (latency/throughput).

is it a typo that dsrate used unindexed searches ? Why unindexed, it will put the pressure on the worker rather that the listener that is where the change is done ? you may use indexed search each of them returning few entries and just DN.

jchapma:
> It is always interesting to have access logs samples along the overall metric (latency/throughput).
> 
> is it a typo that dsrate used unindexed searches ? Why unindexed, it will put the pressure on the worker rather that the listener that is where the change is done ? you may use indexed search each of them returning few entries and just DN.

Sorry yes its an indexed equality search, let me update the message. Also I forgot to capture the access logs, let me do so. Thanks

ldap/servers/slapd/daemon.c
Firstyear: Yeah, that seems reasonable to me :)

ldap/servers/slapd/daemon.c
Firstyear: Yeah, I think that combined with returning a smaller set of the active entries will help too since that naturally will create a distribution of the events too.

Firstyear:
> dsrate with 32 client threads running indexed searches over a small number of busy connections. This targets the listener/worker handoff.

I think in the paper that I sent you recently it noted that epoll helps a lot with high numbers of active connections, but may not be as effective with a small number of very active connections. So this explains the degredation as you increase listeners as you are inversely decreasing the number of active connections on each epoll interest list.

ldap/servers/slapd/connection.c
```diff
@@ -2213,11 +2234,24 @@ connection_threadmain(void *arg)
                     }
                     conn->c_threadnumber--;
                     connection_release_nolock(conn);
+#ifdef ENABLE_EPOLL
+                    /*
+                     * Re-arm the ONESHOT socket after c_threadnumber--.
+                     * The socket may have been left disabled while the
+                     * connection was at the thread limit.
+                     */
+                    connection_epoll_add_socket(conn);
```
tbordaz, line 2243: Is it required to call it at the end of each operation (it calls epoll_ctl). I would expect the worker to have already called connection_make_readable_nolock/connection_epoll_add_socket after reading each operation. If it is to make it readable only after thread limit, we may add it in the (conn->c_threadnumber == maxthreads) few lines above.

ldap/servers/slapd/connection.c
```diff
@@ -2213,11 +2234,24 @@ connection_threadmain(void *arg)
...
+                    if (conn->c_flags & CONN_FLAG_CLOSING) {
+                        need_wakeup = 1;
+                    }
+#endif
                     /* If need_wakeup, call signal_listner once.
                      * Need to release the connection (refcnt--)
                      * before that call.
                      */
-                    if (need_wakeup) {
+                    if (need_wakeup && conn->c_ct_list >= 0) {
```
tbordaz, line 2254: When conn->c_ct_list >= 0, does it mean 'conn' is on the active list ? We may use a macro (IS_ACTIVE(conn)) to make the code easier to read

ldap/servers/slapd/daemon.c
tbordaz: It looks to me that ct->table_mutex is protecting the access to active/inactive list and freelist.
At the same time handle_pr_read_ready access active/inactive without this lock so I should be wrong.

Note you may also use the generic function connection_table_iterate_active_connections with a callback.

ldap/servers/slapd/daemon.c
```diff
@@ -1577,27 +1659,33 @@ ct_list_thread(uint64_t threadnum)
...
+         select_return = epoll_wait(the_connection_table->epoll_fd[threadid], events, EPOLL_MAX_EVENTS, slapd_ct_thread_wakeup_timer);
 #else /* !ENABLE_EPOLL */
          num_poll = setup_pr_read_pds(the_connection_table, threadid);
          select_return = POLL_FN(the_connection_table->fd[threadid], num_poll, pr_timeout);
 #endif /* ENABLE_EPOLL */
          switch (select_return) {
              case 0: /* Timeout */
+#ifdef ENABLE_EPOLL
+            /* No fds ready. Scan for closing conns the pipe missed. */
+             reap_closing_connections(the_connection_table, (int)threadid);
```
tbordaz, line 1680: A listener thread can be busy for a long time reaping the closing connection. So there is a risk of not listening established connection.
What about reaping with another thread ?

ldap/servers/slapd/daemon.c
```diff
@@ -1942,6 +2043,50 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
...
+            * A worker may have signaled this ct-list thread after releasing a
+            * closing connection, so also reap connections that can now be moved
+            * from the active list.
+            */
+            char buf[200];
+            while (read(signalpipes[list_num].readsignalpipe, buf, sizeof(buf)) > 0) {
+                /* Drain */
+            }
+
+            reap_closing_connections(ct, list_num);
```
tbordaz, line 2060: similar remark. The listener may be busy for a long reap waiting too long for new incoming operations/events

ldap/servers/slapd/daemon.c
```diff
@@ -1961,25 +2117,51 @@ handle_pr_read_ready(Connection_Table *ct, int list_num, PRIntn num_poll __attri
              * if connection mutex is held for a long time
              */
             if (pthread_mutex_trylock(&(c->c_mutex)) == EBUSY) {
+#ifdef ENABLE_EPOLL
+                /*
+                * EPOLLONESHOT disabled the fd when the event was delivered.
+                * Re-arm it since the connection could not be processed while
+                * c_mutex is held.
+                */
+                if (!c->c_gettingber) {
```
tbordaz, line 2126: If we are not holding c_mutex we should not test/set the connection.

ldap/servers/slapd/daemon.c
Firstyear: This is the timeout path, eg no activity on any sockets at all.

ldap/servers/slapd/connection.c
jchapma: Yes, you are correct, it should not be called every operation, the worker would have re-armed in make_readable unless we were at the per conn thread limit. If we move it to the if (conn->c_threadnumber == maxthreads) block, it still has the old count so it would not take effect, it has to be run after conn->c_threadnumber--

ldap/servers/slapd/connection.c
jchapma: Yes, conn->c_ct_list >= 0 means the conn is still assigned to a ct_list, set when the conn is moved onto the active list and cleared to -1 when the conn if moved off. This check happens in a few places, so yes I will add a macro or a helper function.

ldap/servers/slapd/daemon.c
jchapma: Yes table_mutex is the "list" lock, The walk in handle_pr_read_ready() and reap_closing_connections()  is used by the ct_list, which should be the only walker of that list.

Thats am interestng idea, let me look into it.

PR #7548, opened 2026-06-03
Issue 7547 - Heap buffer overflow in ldap_utf8prev()

IliaKash1:
Bug description:

Heap buffer overflow in ldap_utf8prev() can be triggered via str2simple if '=' is not preceded by proper symbols.

Fix description:

Additional checks are added to account for '=' being preceded by nothing or by non-ASCII bytes.

Fixes: #7547

Author: Ilia Kashintsev

## Summary by Sourcery

Bug Fixes:
- Reject filter strings where '=' is the first character or appears in the middle of a UTF-8 multibyte sequence to prevent heap buffer overreads in ldap_utf8prev().

PR #7847, opened 2026-09-15
Issue 7758 - heap-buffer-overflow caused by default_dirstring_normalize_int()

### LINKED ISSUE #7758: heap-buffer-overflow caused by default_dirstring_normalize_int()
Hello maintainers! There is an OOB read caused by another unconditional call to `ldap_utf8prev()`, now in `default_dirstring_normalize_int()`.

In this case `default_dirstring_normalize_int()` can call `ldap_utf8prev(d)` after `d` is already equal to `head` while stripping trailing blanks:
https://github.com/389ds/389-ds-base/blob/b23597b7219e994ab04dd89873f6a421ee922646/ldap/servers/slapd/attrsyntax.c#L741-L750

Steps to reproduce:

1) Build the project with ASAN;
```bash
export CC=clang-19
export CXX=clang++-19

autoreconf --force --install
export RUSTUP_TOOLCHAIN=stable
./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking --with-openldap --with-systemd --with-systemdsystemunitdir=/lib/systemd/system --with-systemdsystemconfdir=/etc/systemd/system --with-systemdgroupname=dirsrv.target --with-tmpfiles-d=/etc/tmpfiles.d --enable-autobind --enable-cmocka --enable-icu --enable-rust --enable-asan

make

ar rsc liball.a ldap/servers/slapd/.libs/*.o
ar rsc libsvrscore.a src/svrcore/src/.libs/*.o
```
2) Build the example with ASAN, launch, and get the ASAN report:
```bash
$CC -fsanitize=address -g -O0 -I./ldap/include -I./ldap/servers/slapd $(pkg-config --cflags nspr nss openssl libsasl2) ./ex3.c liball.a libavl.a libsvrscore.a ./.libs/librslapd.a ./rs/rslapd/release/librslapd.a -lldap -llber -lpcre2-8 -lz $(pkg-config  --libs nspr nss openssl libsasl2) -lcrack -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -lkrb5support -ldl -pthread -lresolv -ljson-c -Wl,--allow-multiple-definition -o ex3
```
ex3.c:
```c
#include <stdio.h>

#include <slap.h>
#include <slapi-plugin.h>
#include <proto-slap.h>

int
main(void)
{
    Slapi_Attr attr = {0};
    Slapi_Value *in_vals[2] = {0};
    Slapi_Value **out_vals = NULL;
    struct berval bad = {
        .bv_len = 2,
        .bv_val = "\xB0 "
    };

    attr_syntax_read_lock();
    attr_syntax_unlock_read();

    slapi_attr_init(&attr, "cn");
    in_vals[0] = slapi_value_new_berval(&bad);

    slapi_attr_values2keys_sv(&attr, in_vals, &out_vals, LDAP_FILTER_EQUALITY);
}

```
```bash
./ex
==1189413==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000004d6f at pc 0x638a94d79a82 bp 0x7ffe7c80af40 sp 0x7ffe7c80af38
READ of size 1 at 0x502000004d6f thread T0
    #0 0x638a94d79a81 in ldap_utf8prev /389-ds-base/ldap/servers/slapd/utf8.c:110:14
    #1 0x638a94c80633 in default_dirstring_normalize_int /389-ds-base/ldap/servers/slapd/attrsyntax.c:747:18
    #2 0x638a94c8028c in default_dirstring_values2keys /389-ds-base/ldap/servers/slapd/attrsyntax.c:655:13
    #3 0x638a94d1d773 in slapi_attr_values2keys_sv /389-ds-base/ldap/servers/slapd/plugin_syntax.c:644:14
    #4 0x638a94c78590 in main /389-ds-base/./ex3.c:24:5
    #5 0x71dfcdbab249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #6 0x71dfcdbab304 in __libc_start_main csu/../csu/libc-start.c:360:3
    #7 0x638a94b98c80 in _start (/389-ds-base/ex3+0xf0c80) (BuildId: 97a86bebe764d14d77327157f55e27243a4ef539)

0x502000004d6f is located 1 bytes before 3-byte region [0x502000004d70,0x502000004d73)
allocated by thread T0 here:
    #0 0x638a94c1f9ea in strdup (/389-ds-base/ex3+0x1779ea) (BuildId: 97a86bebe764d14d77327157f55e27243a4ef539)
    #1 0x638a94c832a7 in slapi_ch_strdup /389-ds-base/ldap/servers/slapd/ch_malloc.c:211:14
    #2 0x638a94c8027c in default_dirstring_values2keys /389-ds-base/ldap/servers/slapd/attrsyntax.c:654:17
    #3 0x638a94d1d773 in slapi_attr_values2keys_sv /389-ds-base/ldap/servers/slapd/plugin_syntax.c:644:14
    #4 0x638a94c78590 in main /389-ds-base/./ex3.c:24:5
    #5 0x71dfcdbab249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

SUMMARY: AddressSanitizer: heap-buffer-overflow /389-ds-base/ldap/servers/slapd/utf8.c:110:14 in ldap_utf8prev
Shadow bytes around the buggy address:
  0x502000004a80: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fd
  0x502000004b00: fa fa fd fd fa fa fd fd fa fa fd fd fa fa 06 fa
  0x502000004b80: fa fa 05 fa fa fa 03 fa fa fa 04 fa fa fa 07 fa
  0x502000004c00: fa fa 05 fa fa fa 06 fa fa fa 07 fa fa fa 03 fa
  0x502000004c80: fa fa 04 fa fa fa 05 fa fa fa 06 fa fa fa 04 fa
=>0x502000004d00: fa fa 03 fa fa fa 03 fa fa fa 00 00 fa[fa]03 fa
  0x502000004d80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x502000004e00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x502000004e80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x502000004f00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x502000004f80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==1189413==ABORTING

```

xhon-pelushi:
Fixes #7758.

`ldap_utf8prev()` always reads the byte before its argument, and then keeps walking back for as long as it sees UTF-8 continuation bytes:

```c
register unsigned char *prev = (unsigned char *)s;
while (((*--prev & 0xC0) == 0x80) && (prev != limit)) {
    ;
}
```

Neither is safe at the start of a value, and `default_dirstring_normalize_int()` is only guarded by testing the pointer that comes back:

```c
nd = ldap_utf8prev(d);
while (nd && nd >= head && ldap_utf8isspace(nd)) {
    d = nd;
    nd = ldap_utf8prev(d);
    *d = '\0';
}
```

The read has already happened by the time `nd` is compared against `head`, so the check cannot prevent it.

### Reaching it

A value of `"\xA0 "` is enough. `0xA0` is a stray continuation byte and is not a space, so the leading-blank strip leaves it alone and it gets copied to the output; the trailing space sets `prevspace`. Walking back over that space leaves `d` at `head + 1`, and `ldap_utf8prev(head + 1)` reads `head[0]`, finds a continuation byte, and keeps going past the start of the buffer.

```
    #0 ldap_utf8prev
    #1 default_dirstring_normalize_int
```

So it is not only the `d == head` case: the walk back over continuation bytes starts from `d` and crosses `head` on its own.

### The fix

Walk back with a helper that takes the start of the value and never reads below it, returning `NULL` when there is no preceding character. The `nd >= head` test in the loop then becomes redundant and is dropped.

**This is why the guard used in `value.c` is not enough here.** `value.c` (in commit b3b672b, issue #7400) has the same loops with a `d <= head` break added before the next `ldap_utf8prev(d)`. That stops `d` from reaching `head`, but not the backward scan from crossing it. I ran that exact form of the guard against `"\xA0 "` and it still overflows, which is why this uses a bounded walk instead. I have not exercised `value.c`'s own `value_normalize_ext()` — it has quite a few more branches — so I am not claiming it is reachable there, only that the guard shape would not stop it if it is. Happy to look at that separately if you would like.

### Verification

I could not build the full server here, so the checked-in function was verified by extracting it together with the real `UTF8len`, `ldap_utf8next`, `ldap_utf8len`, `ldap_utf8prev` and `ldap_utf8isspace` bodies and exercising it directly:

| check | result |
| --- | --- |
| differential, current vs patched, over 541370 inputs built from ASCII, blanks, lead bytes and continuation bytes (exhaustive to length 4, then random to length 16) | **0 differences in output** |
| the same inputs under ASAN, value placed flush against the start of its own allocation | **no overflow** with the patch, overflow without it |
| `gcc -fsyntax-only -Wall -Wextra` on `attrsyntax.c` | clean, and identical to the result on pristine `main` |

The reproducer is two files and I am glad to attach it, or to fold it into a test if you have somewhere it belongs — I could not see an existing unit test target that covers these normalisers.

## Summary by Sourcery

Prevent directory string normalization from reading before the start of a value during backward UTF-8 traversal.

Bug Fixes:
- Prevent heap-buffer-overflow during directory string normalization when trimming trailing whitespace from values beginning with UTF-8 continuation bytes.

Enhancements:
- Bound backward UTF-8 character traversal to the start of the value and preserve existing normalization behavior.


xhon-pelushi:
A correction to my own note above: I said I could not see an existing unit test target covering these normalisers. That was wrong — `test_slapd` (`check_PROGRAMS`/`TESTS` in `Makefile.am`) is a cmocka harness that links `libslapd.la`, and `attrsyntax.c` is part of libslapd, so `test/libslapd/` is the natural home for a case like this.

I am happy to add one following the pattern of `test/libslapd/filter/optimise.c`, asserting that a value of `"\xA0 "` normalises to `"\xA0"` without reading before the buffer. One caveat so you can decide: I cannot build the server here, so I would not be able to run `make check` on it myself and would be relying on your CI to confirm it compiles and passes. Say the word if you would like it in this PR or as a separate one.

xhon-pelushi:
Pushed an update. The review comment on the helper is correct and I have fixed it.

The point was that

```c
```

```c
```

Re-ran the differential from the PR description with the clamp in place: **541370 inputs, 0
in the sibling PRs #7858 and #7860, which carried the identical line.

xhon-pelushi:
Separately from the review fix above — a question about how you want the rest of this handled,
because I would rather ask than keep opening issues.

**The call-site list in this PR was incomplete, including my own.** I grepped for
`ldap_utf8prev`. `slapi-plugin.h` also wraps it in two macros:

```c
#define LDAP_UTF8PREV(s) ldap_utf8prev(s)
#define LDAP_UTF8DEC(s) (s = ldap_utf8prev(s))
```

Counting those, it is not four call sites. It is twenty.

**`ldap_utf8prevn()` is one of them, and it is wrong in the same way.** I said in #7858 that it
looked bounded by construction. It is not — it has the floor in hand and checks it one call too
late:

```c
for (; n > 0; --n) {
    prev = ldap_utf8prev(prev);          /* the read happens here */
    if ((n > 0) && (prev < s)) {         /* the bound is checked here */
        return NULL;
    }
}
```

`check_trivial_words()` (`pw.c:1873`) reaches it. It `slapi_ch_strdup()`s a value of `uid`, `cn`,
`sn`, `givenname`, `ou`, `mail` or any `passwordUserAttributes` attribute and walks back
`passwordMinTokenLength` characters from the end. `ldap_utf8characters()` counts a stray
continuation byte as one character, so `"\xA0AA"` passes the length gate at the default token
length of 3:

```
  #0 ldap_utf8prev  #1 ldap_utf8prevn  #2 check_trivial_words
0x50200000002f is located 1 bytes before 4-byte region
```

Same `pw_syntax == LDAP_ON` gate as #7857. I have a fix for it, bounded inside `utf8.c` so it
returns NULL for exactly the cases where the unbounded walk would have ended below the value —
which is what `ldap_utf8prevn()` already reported, only without reading out of bounds to find out.
Differential over 207370 (value, token-length) pairs: no change to `ldap_utf8prevn()`, no change to
`check_trivial_words()`'s answer, and the current code's output never depends on the bytes it
reads out of bounds.

**Three more in the ACL plugin**, all ASAN-reproduced:

- `aclparse.c:1956` `__acl_strip_trailing_space()` — an ACI component that is only whitespace.
- `aclparse.c:640` `__acl_strip_auth_str()` — an ACI whose text begins with `authenticate`; `s--`
  then `*s`.
- `acllas.c:825` `DS_LASGroupDnEval()` — a `groupdn` beginning with `||`. This one has no floor at
  all (`LDAP_UTF8DEC(t); while (' ' == *t || '\t' == *t) { LDAP_UTF8DEC(t); }`) and then does
  `LDAP_UTF8INC(t); *t = '\0';` — so the terminator can be **written** below the value. Measured
  at 8 bytes below with spaces in front of it, and it would go further given more.

These need an administrator-supplied ACI, so I am not calling them remotely triggerable.

I also checked the other eight `LDAP_UTF8DEC` sites in the ACL plugin individually — they are
safe, because their leading-space skip runs first and guarantees the trailing loop stops before
reaching the start.

**The question.** #7548 (July) fixed this same root cause by guarding a caller in `str2filter.c`.
With #7847, #7858 and #7860 that is four caller-side patches, and I now have four more sites. I do
not think opening four more issues at you is the right move — it is the same bug each time, and
`ldap_utf8prev()` and `ldap_utf8prevn()` are both exported in `slapi-plugin.h`, so plugins outside
this tree inherit the same trap.

So, your call, and I will do whichever you prefer:

1. One `utf8.c` change adding a bounded backward walk, with every call site converted, superseding
   these three PRs.
2. Keep these three as they are, and add one further PR for `ldap_utf8prevn()` and the ACL sites.
3. Separate issues and PRs per site, the way the first three went.

I have the fixes and the differentials for all of it either way — I would just rather not fill your
queue with the shape you do not want. There is also one item I would rather not discuss in a public
thread; I will use the repository's private vulnerability reporting for that one.

PR #7858, opened 2026-09-16
Issue 7857 - heap-buffer-overflow in check_pw_syntax_ext()

### LINKED ISSUE #7857: heap-buffer-overflow in check_pw_syntax_ext() via ldap_utf8prev()
While looking at #7758 I checked the other `ldap_utf8prev()` call sites. The one in `check_pw_syntax_ext()` has the same problem, and this one is reached with a user-supplied password.

`ldap/servers/slapd/pw.c`, the character-class loop:

```c
/* check for repeating characters. If this is the
   first char of the password, no need to check */
if (pwd != p) {
    int len = ldap_utf8len(p);
    char *prev_p = ldap_utf8prev(p);
```

The `pwd != p` test stops the call being made on the first character, but it does not stop the walk itself from leaving the buffer. `ldap_utf8prev()` reads the byte before its argument and then keeps walking back while it sees continuation bytes:

```c
register unsigned char *prev = (unsigned char *)s;
while (((*--prev & 0xC0) == 0x80) && (prev != limit)) {
    ;
}
```

So a password whose **first** byte is a stray continuation byte is enough. `ldap_utf8next()` treats `0xA0` as one byte, so after the first iteration `p` is `pwd + 1`, `pwd != p` passes, and `ldap_utf8prev(pwd + 1)` reads `pwd[0]`, sees `0xA0`, and keeps going below the start of the value.

Password `"\xA0a"`:

```
    #0 ldap_utf8prev
    #1 check_pw_syntax_ext character-class loop
```

I reproduced that by extracting the loop together with the real `UTF8len`, `ldap_utf8next`, `ldap_utf8len` and `ldap_utf8prev` bodies, since I cannot build the server here. The `is*()` helpers were stubbed out because they do not move any pointer.

### Severity

Low, as far as I can tell, and I would rather state that plainly than overstate it. It is a read of at most 6 bytes before the value, the bytes only feed `ldap_utf8len()` and a `memcmp()` that adjust a repeated-character counter, and nothing is returned to the client. The realistic worst case is a crash if the allocation happens to start a page. Happy to be told it deserves different handling.

### Suggested fix

The same bounded walk proposed in #7847 — a helper that takes the start of the value and never reads below it.

I also tried the alternative of remembering the previous position while iterating forward, which removes the backward walk entirely. Over 22620 inputs built from ASCII, blanks, lead bytes and continuation bytes:

| variant | differs from current |
|---|---|
| bounded walk | **0** |
| track previous position | 405, **all of them malformed UTF-8** (0 on well-formed input) |

So the bounded walk is the behaviour-preserving option, and that is what I would send unless you prefer the forward-tracking version — it is arguably the more correct reading of "the previous character", but it does change the repeat count for malformed input.

xhon-pelushi:
Fixes #7857. Same root cause as #7847, different call site — this one is reached with a user-supplied password.

The character-class loop in `check_pw_syntax_ext()` looks back one character to count repeats:

```c
if (pwd != p) {
    int len = ldap_utf8len(p);
    char *prev_p = ldap_utf8prev(p);
```

`pwd != p` stops the call being made on the first character, but it does not stop the walk from leaving the value — `ldap_utf8prev()` reads the byte before its argument and then keeps walking back while it sees continuation bytes.

A password whose **first** byte is a stray continuation byte is enough: `ldap_utf8next()` treats `0xA0` as one byte, so after the first iteration `p` is `pwd + 1`, the test passes, and `ldap_utf8prev(pwd + 1)` reads `pwd[0]`, finds a continuation byte, and carries on below the start of the value.

```
    #0 ldap_utf8prev
    #1 check_pw_syntax_ext character-class loop
0x50200000002f is located 1 bytes before 3-byte region
```

### Severity, stated plainly

It stays a read of at most six bytes, the bytes only feed `ldap_utf8len()` and a `memcmp()` that adjust a repeated-character counter, and nothing goes back to the client. The realistic worst case is a crash when the value happens to start a page. I would rather say that than dress it up.

### Why this shape of fix

Two options, measured over 22620 inputs built from ASCII, blanks, lead bytes and continuation bytes:

| variant | differs from current |
|---|---|
| bounded walk (this PR) | **0** |
| remember the previous position while iterating forward | 405, **all malformed UTF-8** (0 on well-formed input) |

Forward tracking removes the backward walk altogether and is arguably the more correct reading of "the previous character", but it changes the repeat count for malformed passwords. I have kept this a fix rather than a behaviour change; say the word if you would prefer the other one.

### Verification

I cannot build the server here, so the loop was extracted with the real `UTF8len`, `ldap_utf8next`, `ldap_utf8len` and `ldap_utf8prev` bodies and exercised directly (the `is*()` helpers do not move any pointer, so they are stubbed). The ASAN trace above and the 22620-input comparison both come from that. `gcc -fsyntax-only -Wall -Wextra` on the real `pw.c` is clean and identical to the result on pristine `main`.

If you take #7847 as well, the two helpers are the same code in two files and would be better as one shared bounded `prev` in `utf8.c` — happy to do that instead if you would rather have it in one place.

## Summary by Sourcery

Bound UTF-8 predecessor lookup during password syntax validation to prevent reads before the start of user-supplied passwords.

Bug Fixes:
- Prevent heap-buffer-overflow when password syntax validation examines repeated characters in malformed UTF-8 values.

Enhancements:
- Preserve existing repeat-counting behavior while safely handling character lookups at password boundaries.


progier389:
safe remark than for #7847 (better have a single ldap_utf8prev_safe function rather than duplicating for each kind of string )

xhon-pelushi:
Pushed an update. A review comment on the sibling PR #7847 applies to this helper too — the same line is here — so I have fixed it in all three.

The point was that

```c
```

```c
```

Re-ran the differential from the PR description with the clamp in place: **22620 inputs, 0
in the sibling PRs #7847 and #7860, which carried the identical line.

PR #7860, opened 2026-09-16
Issue 7859 - heap-buffer-overflow in value_normalize_ext()

### LINKED ISSUE #7859: heap-buffer-overflow in value_normalize_ext() — the b3b672b guard is not sufficient
Following #7758 and #7857, this is the third and last `ldap_utf8prev()` call site with the same defect. This one matters a bit more than the other two because `value.c` already carries a guard for it that turns out not to be enough.

b3b672b (issue #7400) added a `d <= head` break to both trailing-blank loops in `value_normalize_ext()`:

```c
nd = ldap_utf8prev(d);
while (nd && nd >= head && utf8isspace_fast(nd)) {
    d = nd;
    *d = '\0';
    if (d <= head) {
        break;
    }
    nd = ldap_utf8prev(d);
}
```

That stops `d` from *reaching* `head`, but the walk itself still starts at `d` and crosses `head` on its own: `ldap_utf8prev()` reads the byte before its argument and then keeps walking back while it sees continuation bytes. With `d` at `head + 1` and `head[0]` a stray continuation byte, it goes straight past the start of the value.

A CIS value of `"\xA0 "` with `TRIM_TRAILING_BLANK` is enough — `0xA0` is not a space, so the leading strip leaves it, it is copied to the output, and the trailing space sets `prevspace`:

```
    #0 ldap_utf8prev
    #1 value_normalize_ext trailing-blank loop
```

Reproduced by extracting the CIS path and both trailing-blank blocks verbatim, together with the real `UTF8len`, `ldap_utf8next`, `ldap_utf8len`, `ldap_utf8prev` and `ldap_utf8isspace` bodies, since I cannot build the server here. The `SHRINK_TRAILING_BLANK` block has the same shape and the same exposure.

Severity looks the same as the other two: a read of at most six bytes before the value, feeding a space test, with nothing returned to the client.

### Suggested fix

The bounded walk from #7847/#7858. After three sites needing the identical helper, it would be better as one shared bounded `prev` in `utf8.c` rather than a third static copy — I am happy to send it either way, or to consolidate all three into a single PR if you would prefer to review it in one place. Just say which you want and I will reshape them.

xhon-pelushi:
Fixes #7859. Third and last `ldap_utf8prev()` call site, after #7847 and #7858.

This one is worth a closer look than the other two, because `value.c` already carries a guard for exactly this and it turns out not to be enough. b3b672b (issue #7400) added a `d <= head` break to both trailing-blank loops:

```c
nd = ldap_utf8prev(d);
while (nd && nd >= head && utf8isspace_fast(nd)) {
    d = nd;
    *d = '\0';
    if (d <= head) {
        break;
    }
    nd = ldap_utf8prev(d);
}
```

That stops `d` from *reaching* `head`, but the walk still starts at `d` and crosses `head` by itself — `ldap_utf8prev()` reads the byte before its argument and then keeps walking back while it sees continuation bytes. With `d` at `head + 1` and `head[0]` a stray continuation byte, it goes straight past.

A CIS value of `"\xA0 "` with `TRIM_TRAILING_BLANK`:

```
    #0 ldap_utf8prev
    #1 value_normalize_ext trailing-blank loop
0x50200000002f is located 1 bytes before 3-byte region
```

The `SHRINK_TRAILING_BLANK` block has the same shape and the same exposure.

### Verification

I cannot build the server here, so the CIS path and both trailing-blank blocks were extracted verbatim together with the real `UTF8len`, `ldap_utf8next`, `ldap_utf8len`, `ldap_utf8prev` and `ldap_utf8isspace` bodies and exercised directly.

| check | result |
| --- | --- |
| differential, current vs patched, **66660** input and trim-flag combinations (exhaustive to length 4 over ASCII, blanks, lead bytes and continuation bytes, across six flag sets) | **0 differences in output** |
| the same set under ASAN with the value flush against the start of its allocation | **no overflow** with the patch, overflow without |
| `gcc -fsyntax-only -Wall -Wextra` on the real `value.c` | clean |

The differential matters here because I did not just add a guard — the `nd >= head` tests and `d <= head` breaks become redundant once the walk is bounded, so the loops are restructured, and I wanted to be sure the shrink block in particular still preserves exactly one trailing space.

### On the duplication

This is the third file needing the identical helper. Three static copies is the wrong end state — one shared bounded `prev` in `utf8.c` would be better, and I am happy to send that instead and rebase #7847 and #7858 onto it, or to fold all three into a single PR if that is easier to review. Whichever you prefer.

## Summary by Sourcery

Prevent value normalization from reading before the start of malformed UTF-8 values.

Bug Fixes:
- Prevent heap-buffer-overflow during trailing-blank normalization when malformed UTF-8 causes backward traversal before the value buffer.

Enhancements:
- Preserve trailing-space normalization behavior while safely bounding UTF-8 reverse traversal within the value.


progier389:
The idea of using a safe utf8prev function looks good but since it is a generic issue related to ldap_utf8prev (not limited to the syntax plugin, 
IMHO you should rather define an  ldap_utf8prev_safe function in utf8.c  reusing the ldap_utf8prev code with an extra test:  if (head > limit) { limit = head; }
And we should better replace all ldap_utf8prev occurences by ldap_utf8prev_safe

xhon-pelushi:
Pushed an update. The review comment on the helper is correct and I have fixed it.

The point was that

```c
```

```c
```

Re-ran the differential from the PR description with the clamp in place: **66660 (input, flags) pairs, 0
in the sibling PRs #7847 and #7858, which carried the identical line.

PR #7868, opened 2026-09-18
Fix: DEREF_AFTER_NULL

### LINKED ISSUE #7867: NULL pointer dereference in snmp_update_ops_table() / snmp_update_entries_table() / snmp_update_interactions_table() when stats is not initialized
In snmp_collator_update() (ldap/servers/slapd/snmp_collator.c), the
global pointer `stats` is checked for NULL before one dereference:

    if (stats != NULL) {
        stats->hdr_stats.updateTime = time(0);
    }

However, the calls immediately after it dereference the same global
pointer unconditionally, and none of the called functions check it
either:

    snmp_update_ops_table();
    snmp_update_entries_table();
    snmp_update_interactions_table();

snmp_update_ops_table() alone dereferences `stats->ops_stats.*` in
~27 places with no NULL check inside the function. `stats` is
declared as `struct agt_stats_t *stats = NULL;`, so if this callback
fires while `stats` is still NULL (e.g. before initialization
completes, or during/after snmp_collator_stop()), the process will
crash with a NULL pointer dereference.

This was found via static analysis (a DEREF_AFTER_NULL-class
checker), not a runtime reproduction — but the missing guard is
visible directly by inspection: the existing `if (stats != NULL)`
around the updateTime write shows the code already anticipates this
pointer being NULL at this point, it's just not applied consistently
to the calls that follow.

Suggested fix: extend the existing NULL check to cover the three
calls as well, e.g.:

    if (stats != NULL) {
        stats->hdr_stats.updateTime = time(0);

        /* update the mmap'd tables */
        snmp_update_ops_table();
        snmp_update_entries_table();
        snmp_update_interactions_table();
    }

Found on current main (checked 18.09.2026).

ushevchenko:
Problem:
The global pointer 'stats' is checked for NULL before one dereference in snmp_collator_update(), but the subsequent calls to snmp_update_ops_table(), snmp_update_entries_table() and snmp_update_interactions_table() dereference the same pointer unconditionally.
Solution: extended the existing NULL check to cover all four uses of 'stats' in this function.
Signed-off-by: u.shevchenko@fobos-nt.ru
Signed-off-by: gerben@altlinux.org

Fixes #7867

## Summary by Sourcery

Bug Fixes:
- Prevent SNMP table updates when the shared statistics pointer is unavailable, avoiding null-pointer dereferences.


jchapma (approved):
LGTM

PR #7876, opened 2026-09-21
Issue 6645 - Fix backend state toggle attribute

### LINKED ISSUE #6645: dsconf fails to set the backend/mapping tree state
**Issue Description**
The command 'dsconf instance backend suffix set --disable userroot' sets the attribute 'nsslapd-nsstate' in 'cn=<suffix>,cn=mapping tree,cn=config.
This attribute is not handled by mapping_tree code. There is a typo in src/lib389/lib389/backend.py that should set 'nsslapd-state' rather than 'nsslapd-nsstate'

 - since 1.4.3

**Steps to Reproduce**
1. enable audit log
2. run the command 'dsconf instance backend suffix set --disable userroot'
3. run the command 'dsconf instance backend suffix get userroot' and verify that 'nsslapd-state: backend'

**Expected results**
the mapping tree entry should contain 'nsslapd-state: disabled'

dajiaohuang:
Bug Description: Backend.disable() and Backend.enable() write the misspelled nsslapd-nsstate attribute, so callers can report success without changing the mapping tree state.

Fix Description: Write the canonical nsslapd-state values and add a mapping-tree regression test covering both state transitions.

Fixes: https://github.com/389ds/389-ds-base/issues/6645

## Summary by Sourcery

Fix backend state toggling so enable and disable operations correctly update the mapping tree state.

Bug Fixes:
- Correct backend enable and disable helpers to update the canonical mapping tree state attribute and values.

Tests:
- Add a mapping-tree regression test covering backend disable and enable state transitions.


dajiaohuang:
Implemented the review fix in 099e31f: the disabled-state assertion now runs inside a try/finally, so backend.enable() always restores userRoot even when the assertion or mapping-tree read fails. Validation: py_compile, git diff --check, duplicate-ID scan (1,889 IDs, no duplicates), and vermin Python 3.8 checks for src/lib389 and dirsrvtests passed. The Linux integration suite was not run here because no repository build/test environment is available; CI is pending.

